PR #535, opened 2021-10-27
Functional: Beautified iterator

havogt:
only for embedded


havogt:
Adding support for non-embedded (i.e. IR generation) requires type deduction as we need to distinguish function calls from `deref(shift(...))`

havogt:
outdated, could be supported now, by using type deduction.

PR #537, opened 2021-10-29
Add FrozenStencil

jdahm:
## Description

Add FrozenStencil and add a few files to pre-commit.


src/gt4py/definitions.py
```diff
@@ -21,22 +21,20 @@
...
-from gt4py.utils.attrib import List as ListOf
-from gt4py.utils.attrib import Tuple as TupleOf
 from gt4py.utils.attrib import attribclass, attribkwclass, attribute
 
 
 class CartesianSpace:
     @enum.unique
     class Axis(enum.Enum):
-        I = 0
+        I = 0  # noqa: E741
```
egparedes, line 37: I suggest to add a short description of QA error codes, specially for the not-so-common ones concrete 
```suggestion
        I = 0  # noqa: E741  # Do not use variables named 'I', 'O', or 'l' 
```

src/gt4py/definitions.py
```diff
@@ -513,8 +510,8 @@ def shape(self):
 
 
 class Extent(FrameTuple):
-    """Stencil support: region defined by the smallest and the largest offsets in
-     a stencil pattern computation.
+    """
```
egparedes, line 513: Comestic detail: I think in this codebase docstrings start always (or most of the times) in the triple-quoted line

src/gt4py/definitions.py
```diff
@@ -616,8 +613,8 @@ def _apply(self, other, left_func, right_func=None):
 
 
 class CenteredExtent(Extent):
-    """Stencil support: region defined by the largest negative offset (or zero) and
-    the largest positive offset (or zero) in a stencil pattern computation.
+    """
```
egparedes, line 616: Same as before

src/gt4py/stencil_object.py
```diff
@@ -1,27 +1,57 @@
...
-    Shape,
-)
+from gt4py.definitions import DomainInfo, FieldInfo, Index, ParameterInfo, Shape
+
+
+FieldType = Union[gt_storage.storage.Storage, np.ndarray]
+OriginType = Union[Tuple[int, int, int], Dict[str, Tuple[int, ...]]]
+
+
+class FrozenStencil:
```
egparedes, line 36: I would make `FrozenStencil` immutable (raising an error in `__setattr__`) to make sure that the state is not corrupted by mistake after creation

src/gt4py/stencil_object.py
```diff
@@ -143,7 +173,8 @@ def _get_max_domain(
         *,
         squeeze: bool = True,
     ) -> Shape:
-        """Return the maximum domain size possible
+        """
```
egparedes, line 176: Same comment as before about docstrings

src/gt4py/stencil_object.py
```diff
@@ -188,8 +212,15 @@ def _get_max_domain(
         else:
             return max_domain
 
-    def _validate_args(self, field_args, param_args, domain, origin) -> None:
-        """Validate input arguments to _call_run.
+    def _validate_args(  # noqa: C901
```
egparedes, line 215: Please add C901 error description

src/gt4py/stencil_object.py
```diff
@@ -277,7 +316,7 @@ def _validate_args(self, field_args, param_args, domain, origin) -> None:
                         f"Origin for field {name} too small. Must be at least {min_origin}, is {field_domain_origin}"
                     )
 
-                spatial_domain = domain.filter_mask(field_domain_mask)
+                spatial_domain = Shape(domain).filter_mask(field_domain_mask)
```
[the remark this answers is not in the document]
egparedes: Domain was originally typed as a tuple and later modified. It's obvious for humans that after the try-block it cannot be anything else than a `Shape` but not for mypy. The 100% right coding here would be to either use a different variable name for the domain parameter and the _'shapified'_ domain, or to use a user-defined typeguard (https://www.python.org/dev/peps/pep-0647/).

In any case, the current fix should use `typing.cast()` instead of creating a new Shape object
```suggestion
                spatial_domain = typing.cast(Shape, domain).filter_mask(field_domain_mask)
```

src/gt4py/stencil_object.py
```diff
@@ -267,6 +297,15 @@ def _validate_args(self, field_args, param_args, domain, origin) -> None:
...
 
+                if (
+                    isinstance(field, gt_storage.storage.Storage)
+                    and tuple(field.mask)[:domain_ndim] != field_domain_mask
+                ):
+                    raise ValueError(
+                        f"Storage for '{name}' has domain mask '{field.mask}' but the API signature "
+                        f"expects '[{', '.join(field_info.axes)}]'"
+                    )
+
```
egparedes, line 308: Are there tests to exercise this new condition? Isn't it redundant after the checks in `# assert compatibility of fields with stencil` (line 253 ...)?

src/gt4py/stencil_object.py
```diff
@@ -352,48 +433,14 @@ def _call_run(
...
-        domain_ndim = self.domain_info.ndim
-        origin = self._make_origin_dict(origin)
-        all_origin = origin.get("_all_", None)
-
-        # Set an appropriate origin for all fields
-        for name, field_info in self.field_info.items():
-            if field_info is not None:
-                assert name in field_args, f"Missing value for '{name}' field."
-                field_origin = origin.get(name, None)
+        origin = self._normalize_origins(field_args, origin)
```
egparedes, line 439: I have two questions here:

1. This assert has been removed. Can you explain why? Is it not needed?
```
        assert (
            len(domain) == domain_ndim
        ), f"Provided domain '{domain}' is not {domain_ndim}-dimensional."
```

2. The code in `_normalize_origins()` was previously inlined here instead of declared in a different function to avoid an extra Python function call at run-time in this performance-critical part of the code. It could have been overkill, but the truth is that the code is not used in any other part anyway, so moving it to a different function does not provide any reusability (at least for now

egparedes (changes requested):
It looks good, just a couple of cosmetic issues and some additional comments open to discussion.

src/gt4py/stencil_object.py
jdahm: 1. The check is the second one in `self._validate_args`, so I didn't feel like it was needed here. I could be convinced otherwise if you feel that it is needed.
2. That was a good thing to do, but I'd like to try it out this way and see if there's a significant performance impact. I think the readability that is gained in `_call_run`, together with the ability to re-use the origin normalization call when using `FrozenStencil` could outweigh the performance impact. For example, although ugly, my test for the `FrozenStencil` so far has been
```python
field_args = {"field_in": in_storage, "field_out": out_storage}
origin = definition._normalize_origins(field_args, None)
domain = definition._get_max_domain(field_args, origin)
frozen = definition.freeze(origin, domain)
exec_info = {"__aggregate_data": True}
for i in range(num_iter):
    frozen(**field_args, exec_info=exec_info)
```

I much prefer the cached version to frozen for this reason.

src/gt4py/stencil_object.py
jdahm: The test was moved over from `_get_max_domain`, but you're right that I don't think both that and the test for number of dimensions is needed.
```python
                if field.ndim != field_domain_ndim + len(field_info.data_dims):
                    raise ValueError(
                        f"Storage for '{name}' has {field.ndim} dimensions but the API signature "
                        f"expects {field_domain_ndim + len(field_info.data_dims)} ('{field_info.axes}[{field_info.data_dims}]')"
                    )

                if (
                    isinstance(field, gt_storage.storage.Storage)
                    and tuple(field.mask)[:domain_ndim] != field_domain_mask
                ):
                    raise ValueError(
                        f"Storage for '{name}' has domain mask '{field.mask}' but the API signature "
                        f"expects '[{', '.join(field_info.axes)}]'"
                    )
```
I suggest we remove the one for the number of dimensions. That causes one test in `test_call_interface` to fail, but it triggers on the other exception so I can fix it.

jdahm: If `in_storage = gt_storage.ones(shape=(4, 4, 2), backend="gtc:numpy", default_origin=(0, 0, 0), dtype=(np.float_, (3, 3)))` then the dimension exception is raised.

jdahm: I opened up #542 to address this.

src/gt4py/stencil_object.py
egparedes: Regarding the point 1. (the removed `assert`), I'm fine to get it removed, but note that it was not redundant since `validate_args` is not always called, and in that case it would have provided more meaningful error messages

egparedes: Regarding the second point, I'm fine with factorizing out the code in `_normalize_origins()` (although it good be good if you can run a couple of microbenchmarks with `timeit` to see if there is a performance regression).

I don't see the use-case to make the methods you mention public. For testing, it shouldn't be a problem to call these methods and if we make them public, it could be a bit confusing for users to expose this functionality as part of the intended interface of `StencilObject`.

src/gt4py/stencil_object.py
egparedes: Ok, it makes sense now. Thanks for the clarification.

egparedes (approved):
It looks ready to go. I'll leave it open for now to make sure we agree in the open discussions, but plan to merge it tomorrow.

egparedes:
bors try

src/gt4py/stencil_object.py
```diff
@@ -316,7 +312,7 @@ def _validate_args(  # noqa: C901
                         f"Origin for field {name} too small. Must be at least {min_origin}, is {field_domain_origin}"
                     )
 
-                spatial_domain = Shape(domain).filter_mask(field_domain_mask)
+                spatial_domain = cast(Shape, domain).filter_mask(field_domain_mask)
```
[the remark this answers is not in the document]
jdahm: This is much cleaner style indeed. I'll make that change.

src/gt4py/stencil_object.py
jdahm: > (although it good be good if you can run a couple of microbenchmarks with timeit to see if there is a performance regression

Seems to have no effect. I picked a single-statement short test stencil I already wrote and wrapped it with timeit:
```python
import timeit
from gt4py import gtscript
import numpy as np


@gtscript.stencil(backend="gtc:numpy", rebuild=False)
def stencil(
    field_in: gtscript.Field[gtscript.IJK, (np.float_, (2,))],
    field_out: gtscript.Field[gtscript.IJK, np.float_],
):
        field_out = field_in[0, 0, 0][0]


setup = """
from __main__ import stencil
from gt4py import storage as gt_storage
import numpy as np

in_storage = gt_storage.ones(
    shape=(4, 4, 2), backend="gtc:numpy", default_origin=(0, 0, 0), dtype=(np.float_, (2,))
)
out_storage = gt_storage.ones(
    shape=(4, 4, 2),
    backend="gtc:numpy",
    default_origin=(0, 0, 0),
    dtype=np.float_,
)
"""

print(
    timeit.timeit(
        setup=setup,
        stmt="stencil(in_storage, out_storage)",
        number=10000,
    )
)
```

On master this returns ~3.78, and on the `frozen-stencil` branch it's actually a hair faster: 3.70.

I found out that `timeit` isn't super compatible with the way we read the AST in gtscript, so I had to keep that outside the setup code above. I pasted it in case others come across this issue in the future.

egparedes: :+1:

egparedes:
Thanks for all the fixes and improvements. I'll merge as soon as you solve the conflicts.

src/gt4py/stencil_object.py
jdahm: > Regarding the point 1. (the removed assert), I'm fine to get it removed, but note that it was not redundant since validate_args is not always called, and in that case it would have provided more meaningful error messages

Definitely. In this sense I agree it was not redundant. However, if I'm a user calling a stencil with `validate_args=False`, I probably don't want that check slowing it down.

jdahm:
@egparedes Good thing I realized that the code was calling `StencilObject.__call__` when it should have been calling `StencilObject.run()`. Can you check that and verify I have things correct now?

gronerl:
bors try

gronerl:
bors trz

gronerl:
bors try

egparedes:
bors try

PR #539, opened 2021-10-29
Cache origin and domain normalization

### LINKED ISSUE #102: Stencil launch overhead
The launch overhead of gt4py stencils seems to be on the order of 10ms, which is a _lot_. The whole timestep of an atmospheric model should typically take O(100ms) and is launching O(100) stencils, if we are aiming for a throughput of 1 SYPD at 1 km resolution. The plot shows launch overheads of the same stencil running in a) plain python b) gt4py with numpy backend c) gt4py with gtx86 bakckend and d) Fortran.

image

jdahm:
## Description

This caches the stencil pre-computation normalization. The unit test stencil overhead is sped up by 25x on subsequent calls with the same origin and domain.

Builds on #537.

Resolves #102.


jdahm:
@egparedes perhaps it should be a wrapper over StencilObject that is set with `cached=True` on the `gtscript.stencil` call?

src/gt4py/stencil_object.py
```diff
@@ -1,27 +1,47 @@
...
+
+def _compute_cache_key(field_args, parameter_args, domain, origin) -> int:
+    field_hashes = tuple(
+        hash(name) + hash(field.shape) + hash(field.default_origin)
+        for name, field in field_args.items()
+    )
+    parameter_hashes = tuple(hash(name) + hash(param) for name, param in parameter_args.items())
+    domain_hash = hash(domain)
+    origin_hash = hash(origin)
+    return hash((*field_hashes, *parameter_hashes, domain_hash, origin_hash))
```
egparedes, line 44: `hash()` should only be called **_once_** with an argument tuple containing all the information needed to create the key.

src/gt4py/stencil_object.py
```diff
@@ -352,50 +420,25 @@ def _call_run(
...
-                if field_origin is not None:
-                    field_origin_ndim = len(field_origin)
-                    if field_origin_ndim != field_info.ndim:
-                        assert (
-                            field_origin_ndim == field_info.domain_ndim
-                        ), f"Invalid origin specification ({field_origin}) for '{name}' field."
-                        origin[name] = (*field_origin, *((0,) * len(field_info.data_dims)))
+        cache_key = _compute_cache_key(field_args, parameter_args, domain, origin)
+        if cache_key != self._last_key:
+            origin = self._normalize_origins(field_args, origin)
```
egparedes, line 428: I understand that checking if the current call matches the last one is the most common case, but I would add a dict here  `{cache_key: (origin, domain)}` (maybe limited in size?) to make sure that information from any previous call is reused, since the cost of computing the cache key is already much larger than a (small) dict lookup.

egparedes (changes requested):
There is a lot of overlap with #537. I would wait until that one is merged before keep working on this. Besides that, I have a couple of comments regarding the implementation.

src/gt4py/stencil_object.py
```diff
@@ -352,50 +417,23 @@ def _call_run(
...
-        assert (
-            len(domain) == domain_ndim
-        ), f"Provided domain '{domain}' is not {domain_ndim}-dimensional."
-
-        if validate_args:
-            self._validate_args(field_args, parameter_args, domain, origin)
+            self._seen_domain_origin[cache_key] = (domain, origin)
+        else:
+            domain = self._seen_domain_origin[cache_key][0]
+            origin = self._seen_domain_origin[cache_key][1]
```
jdahm, line 436: Combine.

egparedes:
#537 has been already merged. I can review this PR again once you have merged master into it and solved the conflicts.

src/gt4py/stencil_object.py
```diff
@@ -118,17 +128,14 @@ class StencilObject(abc.ABC):
...
     def __new__(cls, *args, **kwargs):
         if getattr(cls, "_instance", None) is None:
             cls._instance = object.__new__(cls)
         return cls._instance
 
-    def __setattr__(self, key, value) -> None:
-        raise AttributeError("Attempting a modification of an attribute in a frozen class")
-
-    def __delattr__(self, item) -> None:
-        raise AttributeError("Attempting a deletion of an attribute in a frozen class")
```
jdahm, line 130: This is implemented at the expense of freezing the StencilObject class. Does anyone see this as a problem? We could potentially do this in a wrapper class a la FrozenStencil.

src/gt4py/stencil_object.py
```diff
@@ -467,13 +474,19 @@ def _call_run(
...
-        if domain is None:
-            domain = self._get_max_domain(field_args, origin)
+            if validate_args:
+                self._validate_args(field_args, parameter_args, domain, origin)
 
-        if validate_args:
-            self._validate_args(field_args, parameter_args, domain, origin)
+            self._seen_domain_origin[cache_key] = (domain, origin)
+        else:
+            domain, origin = self._seen_domain_origin[cache_key]
```
egparedes, line 489:
```suggestion
            domain, origin = type(self)._domain_origin_cache[cache_key]
```

egparedes (changes requested):
I think there are some conceptual and implementation errors in the PR and thus I suggest changes to overcome them.
Most of the errors should have been caught by proper tests, so I suggest to also add new tests for the new functionality (showing that it works with and without cache hit, and that there is performance benefit for the cache hits).

egparedes:
> @egparedes perhaps it should be a wrapper over StencilObject that is set with `cached=True` on the `gtscript.stencil` call?

Ooops, I had missed this comment, @jdahm. I like the idea and it shouldn't be hard to implement. Maybe we can fix first the current issues and then refactor the code following your idea.

src/gt4py/stencil_object.py
```diff
@@ -35,6 +35,16 @@
 OriginType = Union[Tuple[int, int, int], Dict[str, Tuple[int, ...]]]
 
 
+def _compute_cache_key(field_args, parameter_args, domain, origin) -> int:
+    field_data = tuple(
+        name + str(field.shape) + str(field.default_origin)
```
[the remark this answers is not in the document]
jdahm: Yes, it seems like the storage implementation forces it to be a `Tuple[int, ...]` so that's fine. Thanks for the comment!

src/gt4py/stencil_object.py
```diff
@@ -118,17 +128,14 @@ class StencilObject(abc.ABC):
     _gt_id_: str
     definition_func: Callable[..., Any]
 
+    _seen_domain_origin: Dict[int, Tuple[Tuple[int, ...], Dict[str, Tuple[int, ...]]]] = {}
+    """Stores domain/origin pairs that have been used by hash."""
+
```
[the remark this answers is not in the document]
jdahm: I like your naming scheme. I just chose whatever came into my head first, but I hadn't really reviewed this PR myself much yet. Thanks!

src/gt4py/stencil_object.py
```diff
@@ -35,6 +36,15 @@
 OriginType = Union[Tuple[int, int, int], Dict[str, Tuple[int, ...]]]
 
 
+def _compute_cache_key(field_args, parameter_args, domain, origin) -> int:
+    field_data = tuple(
+        (name, field.shape, field.default_origin)
+        for name, field in field_args.items()
+        if field is not None
+    )
+    return hash((*field_data, *parameter_args.keys(), dumps(domain), dumps(origin)))
```
egparedes, line 45: No need to unpack `field_data` tuple since it is hashable.
```suggestion
    return hash((field_data, *parameter_args.keys(), dumps(domain), dumps(origin)))
```

tests/test_integration/test_stencil_object.py
```diff
@@ -0,0 +1,67 @@
...
+    in_storage = gt_storage.ones(
+        backend=backend, default_origin=(0, 0, 0), shape=shape, dtype=float
+    )
+    out_storage = gt_storage.ones(
+        backend=backend, default_origin=(0, 0, 0), shape=shape, dtype=float
+    )
+
+    def runit(*args, **kwargs) -> float:
+        exec_info: Dict[str, Any] = {}
+        start_time: float = time.perf_counter()
```
egparedes, line 52: It think the `time.perf_counter()` calls for `start_time` and `end_time` variables are not needed here since the same information is available in the `exec_info` dict as `exec_info["call_run_start_time"] ` and `exec_info["call_run_end_time"] `

egparedes (changes requested):
It looks good, but there are some bugs yet. I outlined possible fixes in my review.

src/gt4py/stencil_object.py
```diff
@@ -35,6 +36,15 @@
 OriginType = Union[Tuple[int, int, int], Dict[str, Tuple[int, ...]]]
 
 
+def _compute_cache_key(field_args, parameter_args, domain, origin) -> int:
+    field_data = tuple(
+        (name, field.shape, field.default_origin)
```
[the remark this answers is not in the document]
jdahm: Thanks for looking into that!

tests/test_integration/test_stencil_object.py
```diff
@@ -0,0 +1,65 @@
...
+from gt4py import storage as gt_storage
+from gt4py.gtscript import PARALLEL, Field, computation, interval
+
+
+@pytest.fixture
+def backend() -> str:
+    return "gtc:numpy"
+
+
+def test_stencil_object_cache(backend: str):
```
egparedes, line 33: I think creating a fixture here is a bit of an overkill, so I suggest to use a parametrize argument:

```suggestion
@pytest.mark.parametrize("backend", ["gtc:numpy"])
def test_stencil_object_cache(backend: str):
```

egparedes (approved):
It looks good.

egparedes:
bors try

egparedes (changes requested):
This PR is now solid and it could be merged already, but I would like to ask for one last missing feature: let users clean the domain/origin cache on request. This extra feature could be used for reproducibilty or benchmarking purposes, as well as to free up some memory in case the cache grows too large (although it seems unlikely).

I propose to add a new class method (e.g. `clean_call_args_cache()`) to `StencilObject` which will just call to `cls._domain_origin_cache.clear()`

tests/test_integration/test_stencil_object.py
```diff
@@ -0,0 +1,61 @@
...
+    base_time = runit(in_storage, out_storage, offset=1.0)
+    fast_time = runit(in_storage, out_storage, offset=1.0)
+    assert fast_time < base_time
+
+    # When an origin changes, it needs to recompute more, so the time should increase
+    out_storage = gt_storage.ones(
+        backend=backend, default_origin=(1, 0, 0), shape=shape, dtype=float
+    )
+    third_time = runit(in_storage, out_storage, offset=1.0)
+    assert third_time > fast_time
```
egparedes, line 61: Add small test case for cache cleaning method and minor code style changes.

```suggestion
    # When an origin changes, it needs to recompute more, so the time should increase
    other_out_storage = gt_storage.ones(
        backend=backend, default_origin=(1, 0, 0), shape=shape, dtype=float
    )
    other_origin_time = runit(in_storage, other_out_storage, offset=1.0)
    assert other_origin_time > fast_time
    
    # When the cache is cleared everything is recomputed again, so the time will increase
    cleaned_cache_time = runit(in_storage, out_storage, offset=1.0)
    assert cleaned_cache_time > fast_time    
    
```

egparedes (commented):
It looks very good. I'll wait for all the tests to pass and I will merge afterwards.

egparedes (approved):
Good to go.

PR #542, opened 2021-11-01
Check data dimensions shape

jdahm:
## Description

Adds a check for data dimensions shape and a test.

### DIFF
--- a/src/gt4py/stencil_object.py
+++ b/src/gt4py/stencil_object.py
@@ -267,6 +267,15 @@ class StencilObject(abc.ABC):
                         f"expects {field_domain_ndim + len(field_info.data_dims)} ('{field_info.axes}[{field_info.data_dims}]')"
                     )
 
+                # Check: data dimensions shape
+                if not all(
+                    field.shape[field_domain_ndim + i] == dim_size
+                    for i, dim_size in enumerate(field_info.data_dims)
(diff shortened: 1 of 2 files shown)


src/gt4py/stencil_object.py
```diff
@@ -267,6 +267,15 @@ def _validate_args(self, field_args, param_args, domain, origin) -> None:
                         f"expects {field_domain_ndim + len(field_info.data_dims)} ('{field_info.axes}[{field_info.data_dims}]')"
                     )
 
+                # Check: data dimensions shape
+                if not all(
+                    field.shape[field_domain_ndim + i] == dim_size
+                    for i, dim_size in enumerate(field_info.data_dims)
+                ):
```
egparedes, line 274: Maybe this check can be simplified like this:
```suggestion
                if field.shape[field_domain_ndim:] != field_info.data_dims:
```
Or am I missing something? Can `field.shape` have more dimensions?

egparedes:
bors try

egparedes (commented):
It looks good to me, just a single comment

src/gt4py/stencil_object.py
jdahm: No it cannot as far as I know. This is a nice simplification!

PR #546, opened 2021-11-03
add a static single assignment pass with tests

DropD:
## Description

Add a single static assign node transformer for python AST + unit tests

### DIFF
--- a/src/functional/ffront/single_static_assign.py
+++ b/src/functional/ffront/single_static_assign.py
@@ -49,12 +49,12 @@ class SingleStaticAssignPass(ast.NodeTransformer):
         """
         Rename right hand side names.
 
-        Only read from parent visitor state, can not modify.
+        Only read from parent visitor state, should not modify.
         """
 
         def __init__(self, state):
             super().__init__()
-            self.state = SingleStaticAssignPass.State(name_counter=state.name_counter.copy())
+            self.state = state
 
         def visit_Name(self, node: ast.Name) -> ast.Name:
             if node.id in self.state.name_counter:
--- a/src/functional/ffront/single_static_assign.py
+++ b/src/functional/ffront/single_static_assign.py
@@ -18,7 +18,7 @@ class SingleStaticAssignPass(ast.NodeTransformer):
 
         def visit_Name(self, node: ast.Name) -> ast.Name:
             if node.id in self.state.name_counter:
-                node.id = f"{node.id}_{self.state.name_counter[node.id]}"
+                node.id = f"{node.id}${self.state.name_counter[node.id]}"
             return node
 
     @dataclass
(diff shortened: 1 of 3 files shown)


src/functional/ffront/single_static_assign.py
```diff
@@ -0,0 +1,46 @@
...
+        # first update rhs names to reference the latest version
+        node.value = self.RhsRenamer(self.state).visit(node.value)
+        # then update lhs to create new names
+        node.targets = [self.visit(target) for target in node.targets]
+        return node
+
+    def visit_Name(self, node: ast.Name) -> ast.Name:
+        if node.id in self.state.name_counter:
+            self.state.name_counter[node.id] += 1
+            node.id = f"{node.id}_{self.state.name_counter[node.id]}"
```
egparedes, line 42: To avoid collisions with actual variable names, I would use "$" as separator:
```suggestion
            node.id = f"{node.id}${self.state.name_counter[node.id]}"
```

src/functional/ffront/single_static_assign.py
```diff
@@ -0,0 +1,46 @@
...
+        else:
+            self.state.name_counter[node.id] = 0
+            node.id = f"{node.id}_0"
```
egparedes, line 45: Here maybe it's not needed to rename the symbol, but if it's renamed, it should use the same separator (e.g.  "$") as above
```suggestion
            node.id = f"{node.id}$0"
```

src/functional/ffront/single_static_assign.py
```diff
@@ -0,0 +1,46 @@
...
+    class RhsRenamer(ast.NodeTransformer):
+        """
+        Rename right hand side names.
+
+        Only read from parent visitor state, can not modify.
+        """
+
+        def __init__(self, state):
+            super().__init__()
+            self.state = SingleStaticAssignPass.State(name_counter=state.name_counter.copy())
```
egparedes, line 17: I don't see the point of cloning the state here if it's never going to be modified. If you do it just for protection, you could define `name_counter` in `State` as a generic `Mapping` and here create a new `State` using `MappingProxyType` as a wrapper around the input state, which cannot be update it (but as I said, it might be a bit of overkill)

src/functional/ffront/single_static_assign.py
```diff
@@ -0,0 +1,46 @@
...
+
+        def visit_Name(self, node: ast.Name) -> ast.Name:
+            if node.id in self.state.name_counter:
+                node.id = f"{node.id}_{self.state.name_counter[node.id]}"
```
egparedes, line 21: To avoid collisions with actual variable names, I would use "$" as separator (which is also common practice in compilers):
```suggestion
                node.id = f"{node.id}${self.state.name_counter[node.id]}"
```

tests/functional_tests/ffront_tests/test_single_static_assign.py
```diff
@@ -0,0 +1,93 @@
...
+        tmp = 1
+        tmp = 2
+        tmp = 3
+        """
+    )
+
+    lines = ast.unparse(ssa_ast).split("\n")
+    assert lines[0].strip() == "tmp_0 = 1"
+    assert lines[1].strip() == "tmp_1 = 2"
+    assert lines[2].strip() == "tmp_2 = 3"
```
egparedes, line 26: Update tests with the new separator

egparedes (changes requested):
I like it. Just have a request and a minor suggestion

src/functional/ffront/single_static_assign.py
DropD: good point

src/functional/ffront/single_static_assign.py
DropD: I will just pass the state through. I agree that making mutation impossible is over-engineering.

egparedes (approved):
It looks good.

PR #547, opened 2021-11-04
Ff unpack

DropD:
## Description

Add passes to linearize python assignments:

```
a = [b, c] = foo()
```

can be turned into

```
a = foo()
b = foo()[0]
c = foo()[1]
```

So consecutive passes only have to deal with one assignment target. Only preserves semantics if run after the single static assign pass.


src/functional/ffront/parsing/single_static_assign.py
```diff
@@ -76,6 +95,12 @@ def visit_Assign(self, node: ast.Assign) -> ast.Assign:
         node.targets = [self.visit(target) for target in node.targets]
         return node
 
+    def visit_AnnAssign(self, node: ast.AnnAssign) -> ast.AnnAssign:
+        if node.value:
+            node.value = self.RhsRenamer(self.state).visit(node.value)
+        node.target = self.visit(node.target)
```
egparedes, line 101: Shouldn't this line be part of the previous `if`? I mean, if there is no value, it's just a variable type declaration, right?
```suggestion
            node.target = self.visit(node.target)
```

src/functional/ffront/parsing/single_static_assign.py
```diff
@@ -65,6 +80,10 @@ def visit_Name(self, node: ast.Name) -> ast.Name:
     class State:
         name_counter: dict[str, int] = field(default_factory=dict)
 
+    @classmethod
+    def mutate_ast(cls, node: ast.AST) -> ast.AST:
```
egparedes, line 84: I would standardize the name of the classmethod meant to be used as entry point for all visitor-like classes (parsers, passes, etc.) as `apply()`, as I think it is done right in GT4Py & GTC

src/functional/ffront/parsing/linear_assign.py
```diff
@@ -0,0 +1,140 @@
...
+                get_ast_from_func(foo)
+            )
+        ))
+
+        # will print out:
+
+        def foo():
+            a = 1
+            b = 1
+
```
egparedes, line 67: I would use doctest syntax in the docstring to make sure it never gets out of sync and it is nicely formatted in the docs. (See several examples here: https://github.com/GridTools/gt4py/blob/master/src/eve/utils.py)

src/functional/ffront/parsing/single_static_assign.py
```diff
@@ -17,7 +32,7 @@ class SingleStaticAssignPass(ast.NodeTransformer):
     Function ``foo()`` in the following example keeps overwriting local variable ``a``
 
         import ast
-        from functional.ffront.parsers import get_ast_from_func
+        from functional.ffront.parsing.parsing import get_ast_from_func
```
egparedes, line 35: Use doctest syntax for the examples

src/functional/ffront/parsing/foir_to_itir.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import List, Optional
+
+from eve import NodeTranslator
+from functional.ffront import field_operator_ir as foir
+from functional.iterator import ir as iir
+
+
+class SymExprResolver(NodeTranslator):
+    @classmethod
+    def parse(cls, nodes: List[foir.Expr], *, params: Optional[list[iir.Sym]] = None) -> foir.Expr:
```
egparedes, line 27: I would standardize the name of the classmethod meant to be used as entry point for all visitor-like classes (parsers, passes, etc.) as `apply()`, as I think it is done right in GT4Py & GTC

src/functional/ffront/parsing/linear_assign.py
```diff
@@ -0,0 +1,140 @@
...
+        # will print out:
+
+        def foo():
+            a0 = 1
+            b0 = 5
+            a1 = (b0, a0)[0]
+            b1 = (b0, a0)[1]
+            return a1, b1
+
+        # which would not have been equivalent had the input AST not been in SSA form.
```
egparedes, line 114: Use doctest syntax

src/functional/ffront/parsing/linear_assign.py
```diff
@@ -0,0 +1,140 @@
...
+        if len(node.targets) != 1:
+            raise ValueError(
+                "AST contains multi target assigns. Please run `SingleAssignTargetPass` first."
+            )
+        if isinstance(target := node.targets[0], (ast.Tuple, ast.List)):
+            yield from self.unpack_assignment(node, targets=target.elts)
+        else:
+            yield node
+
+    def unpack_assignment(
```
egparedes, line 127: Just a question (not important): would it make sense to make this method private just to keep the intended API simpler?

```suggestion
    def _unpack_assignment(
```

src/functional/ffront/parsing/linear_assign.py
```diff
@@ -0,0 +1,140 @@
...
+        self, node: ast.Assign, *, targets: list[ast.expr]  # targets passed here for typing
+    ) -> Iterator[ast.Assign]:
+        for index, subtarget in enumerate(targets):
+            new_assign = copy.copy(node)
+            new_assign.targets = [subtarget]
+            new_assign.value = ast.Subscript(
+                ctx=ast.Load(),  # <- for some reason this must be given for astpretty to work
```
egparedes, line 134: Well, `ctx` is not an optional attribute so it makes sense. In this case needs to be a `Load` context because it is used in the right hand side of the assignment, otherwise it would be an invalid AST... 
```suggestion
                ctx=ast.Load(),  # <- Load context in the right hand side of an assignment
```

egparedes (changes requested):
It looks good. I mostly have comments about two code style issues: standardize method names and use of doctest syntax in the docstrings.

src/functional/ffront/parsing/single_static_assign.py
DropD: True, it is a bit weird, that
```
a: int
a = 5
```

would turn into

```
a: int
a$0 = 5
```

But better than losing the value like in the following case:

```
a = 5
a: int
```

would have been

```
a$0 = 5
a$1: int
```

DropD: actually I found a way to not lose the value and only lose the type annotation in the case of a dangling empty AnnAssign.

src/functional/ffront/parsing/linear_assign.py
DropD: I will rephrase to emphasize more that `ctx` is not optional for Subscript, which strikes me as unusual.

src/functional/ffront/ast_passes/simple_assign.py
```diff
@@ -0,0 +1,145 @@
...
+            elif isinstance(old_value, ast.AST):
+                new_node, *_ = list(self.visit(old_value)) or (None,)
+                if new_node is None:
+                    delattr(node, field)
+                else:
+                    setattr(node, field, new_node)
+        yield node
+
+
+class SingleAssignTargetPass(NodeYielder):
```
egparedes, line 53: Minor comment here: maybe `SSAPass` would be actually easier to read...

egparedes (changes requested):
It looks ready to go except for a couple of minor details that should be very easy to fix.

src/functional/ffront/ast_passes/simple_assign.py
```diff
@@ -0,0 +1,145 @@
...
+        return result[0]
+
+    def visit(self, node: ast.AST) -> Iterator[ast.AST]:
+        result = super().visit(node)
+        if isinstance(result, ast.AST):
+            yield result
+        else:
+            yield from result
+
+    def generic_visit(self, node: ast.AST) -> Iterator[ast.AST]:  # type: ignore
```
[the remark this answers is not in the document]
DropD: good point!

src/functional/ffront/ast_passes/simple_assign.py
DropD: We can always rename later.

egparedes (approved):
It looks good to me.

PR #555, opened 2021-11-11
Complete foast

DropD:
## Description

Add missing nodes to FOAST and integrate them into lowerings.

Also:
 * rename FOIR -> FOAST
 * introduce source locations into FOAST nodes and pass locations through from Python AST


src/functional/ffront/field_operator_ast.py
```diff
@@ -20,48 +20,108 @@
...
+class UnaryOp(Expr):
+    op: UnaryOperator
+    operand: Expr
+
+
+class BinaryOperator(StrEnum):
+    ADD = "plus"
+    SUB = "minus"
+    MULT = "multiplies"
+    DIV = "divides"
```
egparedes, line 74: I think we also need to add `&` and `|` as logical and and or for fields (see next comment)

src/functional/ffront/field_operator_ast.py
```diff
@@ -20,48 +20,108 @@
...
+
+class BoolOperator(StrEnum):
+    AND = "and_"
+    OR = "or_"
+
+
+class BoolOp(Expr):
+    op: BoolOperator
+    left: Expr
+    right: Expr
```
egparedes, line 103: I don't think these nodes are needed until we introduce scalars. To be coherent with Python semantics and NumPy API, the logical `and` and `or` between fields should be represented using the `&` and `|` operators.

tests/functional_tests/ffront_tests/test_interface.py
```diff
@@ -128,3 +162,278 @@ def copy_field(inp):
...
+            itir.FunCall(fun=DEREF, args=[itir.SymRef(id="a")]),
+            itir.FunCall(
+                fun=AND,
+                args=[
+                    itir.FunCall(fun=DEREF, args=[itir.SymRef(id="b")]),
+                    itir.FunCall(fun=DEREF, args=[itir.SymRef(id="c")]),
+                ],
+            ),
+        ],
+    )
```
egparedes, line 439: These should be changed to used the proper operators for fields (`&` and `|`)

src/functional/ffront/func_to_foast.py
```diff
@@ -0,0 +1,307 @@
...
+            target=foast.Name(id=target.id, location=self._getloc(target)),
+            value=self.visit(node.value),
+            location=self._getloc(node),
+        )
+
+    def visit_AnnAssign(self, node: ast.AnnAssign) -> foast.Assign:
+        # TODO (ricoh): type checking
+        #
+        # if the annotation does not match the inferred type of value
+        # then raise raise an exception
```
egparedes, line 129:
```suggestion
        # then raise an exception
```

egparedes (changes requested):
It looks good. The only relevant issue I see is related to the boolean operators with fields.

src/functional/ffront/field_operator_ast.py
DropD: Should we remove them for now?

egparedes: Yes, I would remove them or comment them out.

egparedes (approved):
All fine. Just remove/comment out the unneeded bool operations nodes in FOAST before merging.

PR #562, opened 2021-11-23
Horizontal executions merging fix

### LINKED ISSUE #559: OnTheFlyMerging duplicates LocalScalar definitions
The OnTheFlyMerging pass folds up HorizontalExecutions on the OIR. There are cases when it inlines a horizontal execution that already has LocalScalars in the `declarations` list for the `HorizontalExecution`, and when that's the case then the resulting top-level symbol table sees multiple definitions of the same `LocalScalar`.

The resolution to this to ensure that every `LocalScalar.name` is unique to every `HorizontalExecution`. To make this happen, the `OnTheFlyMerging` pass needs to add a step that looks for existing `LocalScalars` in the `HorizontalExecution` that it is merging in and map these to new names, with new `LocalScalars`.

jdahm:
## Description

Fix #559.

### DIFF
--- a/src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
+++ b/src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
@@ -15,12 +15,50 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from dataclasses import dataclass
-from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Set, Tuple, Union
 
 from eve import NodeTranslator, SymbolTableTrait
 from gtc import common, oir
 
 from .utils import AccessCollector, symbol_name_creator
+
+
+class NameUniqifier(NodeTranslator):
+    @dataclass
+    class Context:
+        used_names: Set[str]
+
+    @classmethod
+    def apply(
+        cls, vertical_loops: List[oir.VerticalLoop], skip_names: Sequence[str]
+    ) -> List[oir.VerticalLoop]:
+        return cls().visit(vertical_loops, ctx=cls.Context(used_names=skip_names))
+
+    def visit_HorizontalExecution(
+        self, node: oir.HorizontalExecution, *, ctx: "Context"
+    ) -> oir.HorizontalExecution:
+        new_symbol_name = symbol_name_creator(ctx.used_names)
+        name_map = {
+            d.name: new_symbol_name(d.name) for d in node.declarations if d.name in ctx.used_names
+        }
+
+        ctx.used_names |= {d.name for d in node.declarations}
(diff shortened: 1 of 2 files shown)


tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py
```diff
@@ -160,14 +160,20 @@ def test_on_the_fly_merging_repeated():
...
+                body=[AssignStmtFactory(left__name="out", right__name="tmp_offset")]
             ),
-            HorizontalExecutionFactory(body=[AssignStmtFactory(left__name="tmp")]),
             HorizontalExecutionFactory(
-                body=[AssignStmtFactory(left__name="out2", right__name="tmp")]
+                body=[AssignStmtFactory(left__name="out2", right__name="tmp_offset")]
             ),
         ],
-        declarations=[TemporaryFactory(name="tmp")],
+        declarations=[TemporaryFactory(name="tmp"), TemporaryFactory(name="tmp_offset")],
```
gronerl, line 176: I propose to make this a new separate test, the old one relied on the particular read-write pattern that was there and I'm not sure if the new test would capture the old test properly.

src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
```diff
@@ -23,6 +23,44 @@
...
+
+    def visit_HorizontalExecution(
+        self, node: oir.HorizontalExecution, *, ctx: "Context"
+    ) -> oir.HorizontalExecution:
+        new_symbol_name = symbol_name_creator(ctx.used_names)
+        name_map = {
+            d.name: new_symbol_name(d.name) for d in node.declarations if d.name in ctx.used_names
+        }
+
+        ctx.used_names |= {d.name for d in node.declarations}
```
gronerl, line 45: shouldn't this add the new names instead of the old ones?

gronerl (changes requested):

tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py
jdahm: Sounds good. The problem was that this test didn't even claim to do what it said it was doing. It does "multiple" applications of the pass, but there is not a nested `_merge` call. I added this test back and renamed both.

src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
jdahm: Good catch! It should have a combination of the new (if collision in used_names) and old names (if no collision detected). I fixed this.

jdahm:
@gronerl Take a look at #564. That might be the way to go instead of this PR. Thanks!

jdahm:
I believe I now prefer #562 over this. Closing in favor of that one.

PR #563, opened 2021-11-23
Add tests to known_first_party isort config

jdahm:
## Description

- Make tests `known_first_party` since they are included in the repo.
- Remove `seed-isort-config` as it is deprecated since isort v5 (see https://github.com/asottile-archive/seed-isort-config).


egparedes (approved):
It looks good. Thanks for the contribution!

PR #564, opened 2021-11-24
Make HorizontalExecution a SymbolTableTrait

### LINKED ISSUE #559: OnTheFlyMerging duplicates LocalScalar definitions

jdahm:
## Description

Makes the `oir.HorizontalExecution` a `SymbolTableTrait`, which allows multiple `oir.HorizontalExecution` to define and use `declarations` `Decl`s with the same name. Follow-ons from this change:
- Makes a change to the horizontal execution merging pass to handle merging executions that have local scalars with the same name: in this case one needs to be re-named.
- Adds existing scalars to the symbol name mapper in order to ensure that all new temporaries have unique names.
- Adds a test for this new behavior.

Fixes #559. Alternative to #562; might be cleaner.

### DIFF
--- a/src/eve/traits.py
+++ b/src/eve/traits.py
@@ -68,10 +68,6 @@ class SymbolTableTrait(concepts.Model):
         values["symtable_"] = cls._collect_symbols(values)
         return values
 
-    def collect_symbols(self) -> None:
-        self.symtable_ = dict()
-        self.symtable_ = self._collect_symbols(self)
-
     @staticmethod
     @contextlib.contextmanager
     def symtable_merger(
--- a/src/gtc/oir.py
+++ b/src/gtc/oir.py
@@ -159,7 +159,7 @@ class Interval(LocNode):
         return cls(start=AxisBound.start(), end=AxisBound.end())
 
 
-class HorizontalExecution(LocNode):
+class HorizontalExecution(LocNode, SymbolTableTrait):
     body: List[Stmt]
     declarations: List[LocalScalar]
 
--- a/src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
+++ b/src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
@@ -64,7 +64,7 @@ class OnTheFlyMerging(NodeTranslator):
 
     def visit_ScalarAccess(
         self, node: oir.ScalarAccess, *, scalar_map: Dict[str, str], **kwargs: Any
-    ):
+    ) -> oir.ScalarAccess:
         if node.name in scalar_map:
             name = scalar_map[node.name]
         else:
@@ -160,27 +160,31 @@ class OnTheFlyMerging(NodeTranslator):
                 (name, o): new_symbol_name(name) for name in writes for o in read_offsets
             }
 
-            # Build up the declarations map of the new horizontal execution
-            from_later = [
+            # 4 contributions to the new declarations list
+            decls_from_later = [
                 d for d in horizontal_execution.declarations if d.name not in duplicated_locals
             ]
-            from_first = [
+            decls_from_first = [
                 d
                 for d in first.declarations
                 if d not in horizontal_execution.declarations or d.name in duplicated_locals
             ]
-            renamed_locals_in_later = [
+            decls_renamed_locals_in_later = [
                 oir.LocalScalar(
                     name=new_name, dtype={**symtable, **first.symtable_}[old_name].dtype
                 )
                 for old_name, new_name in scalar_map.items()
             ]
-            new_locals = [
+            new_decls = [
                 oir.LocalScalar(
                     name=new_name, dtype={**symtable, **first.symtable_}[old_name].dtype
                 )
                 for (old_name, _), new_name in offset_symbol_map.items()
             ]
+
+            declarations = (
+                decls_from_later + decls_from_first + decls_renamed_locals_in_later + new_decls
+            )
 
             merged = oir.HorizontalExecution(
                 body=self.visit(
@@ -188,7 +192,7 @@ class OnTheFlyMerging(NodeTranslator):
                     offset_symbol_map=offset_symbol_map,
                     scalar_map=scalar_map,
                 ),
-                declarations=from_later + from_first + renamed_locals_in_later + new_locals,
+                declarations=declarations,
             )
             for offset in read_offsets:
                 merged.body = (
--- a/src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
+++ b/src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
@@ -20,7 +20,7 @@ from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
 from eve import NodeTranslator, SymbolTableTrait
 from gtc import common, oir
 
-from .utils import AccessCollector, symbol_name_creator
+from .utils import AccessCollector, collect_symbol_names, symbol_name_creator
 
 
 @dataclass
@@ -61,6 +61,15 @@ class OnTheFlyMerging(NodeTranslator):
             if key in offset_symbol_map:
                 return oir.ScalarAccess(name=offset_symbol_map[key], dtype=node.dtype)
         return self.generic_visit(node, **kwargs)
+
+    def visit_ScalarAccess(
+        self, node: oir.ScalarAccess, *, scalar_map: Dict[str, str], **kwargs: Any
+    ):
+        if node.name in scalar_map:
+            name = scalar_map[node.name]
+        else:
+            name = node.name
+        return oir.ScalarAccess(name=name, dtype=node.dtype)
 
     def _merge(
         self,
@@ -122,6 +131,7 @@ class OnTheFlyMerging(NodeTranslator):
         ):
             return [first] + self._merge(others, symtable, new_symbol_name, protected_fields)
 
+        first_scalars = {decl.name for decl in first.declarations}
         writes = first_accesses.write_fields()
         others_otf = []
         for horizontal_execution in others:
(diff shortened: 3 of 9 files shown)


tests/test_unittest/test_gtc/common_utils.py
```diff
@@ -35,8 +35,7 @@ def visit_Node(self, node: concepts.Node, *, symbols: Set[str], refs: Set[str])
                     elif issubclass(type_, type_definitions.SymbolRef):
                         refs.add(getattr(node, name))
 
-            if not isinstance(node, traits.SymbolTableTrait):
```
havogt, line 38: What's this about?

src/gtc/oir.py
```diff
@@ -159,7 +159,7 @@ def full(cls):
         return cls(start=AxisBound.start(), end=AxisBound.end())
 
 
-class HorizontalExecution(LocNode):
```
havogt, line 162: Why does this fix #559?

tests/test_unittest/test_gtc/common_utils.py
jdahm: The point of `undefined_symbol_list` is to create a list of results of calls (usually `Decl`s) for each ref name without a decl iterated over by `CollectSymbolsAndRefs`. The problem when making `HorizontalExecution` a `SymbolTableTrait` is that the collector stops at the `VerticalLoopSection`, so it never sees even a single ref to any API field, so each test ends up with errors about undefined references from the validator.

src/gtc/oir.py
jdahm: The problem is currently that the horizontal execution merging when applied in a nested way results in multiple `HorizontalExecutions` having the same `oir.LocalScalar` (a decl). Therefore it ends up with multiple declarations of the same symbol name. The symbol table collects for the sub-tree, skipping iterating into new symbol tables. Therefore making `HorizontalExecution` a `SymbolTableTrait` removes the duplicate decls.

havogt: I agree that it makes sense to open a scope at the HorizontalExecution level, however it feels like it is solving the issue by side-effect of how the merging is implemented. I think merging should respect that there are 2 different LocalScalar with the same name. Also reusing the same declaration creates a fake dependency which might be confusing even if it might never be a problem for correctness.
Does it make sense?

tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py
```diff
@@ -172,3 +172,28 @@ def test_on_the_fly_merging_repeated():
     transformed = OnTheFlyMerging().visit(testee)
     hexecs = transformed.vertical_loops[0].sections[0].horizontal_executions
     assert len(hexecs) == 2
+
+
+def test_on_the_fly_merging_nested():
```
havogt, line 177: Is this testing the LocalScalar declaration problem? How?

havogt (changes requested):
General question, do you know why this pass is active in the default pipeline? Shouldn't the corresponding DaCe optimization do the same?

tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py
jdahm: Yes, it is a nested merging. The first merge step merges the first horizontal iteration into 2-4, then the second merges what was the second into the remaining two. It ends up with `tmp0` defined in two `HorizontalExecution`s.

src/gtc/oir.py
havogt: After slack chat: I misunderstood the problem. This PR is solving the issue of not being allowed to have LocalScalars with the same name in 2 different HorizontalExecutions. But this fix introduces the problem I was talking about: now if you have 2 distinct HorizontalExecutions with LocalScalars with the same name, that get merged into a single HorizontalExecution, you will still have a name clash.

jdahm: Yes. What I'm understanding is that we need a small change to `OnTheFlyMerging` in this PR that checks if there are `LocalScalars` with the same name, and if so renames one set of them to avoid the clash.

jdahm: Turns out after hours of investigation that it's not so easy... The pass very much assumes that the symbol table comes from the `symtable` keyword-argument from `visit_VerticalLoopSection`. Instead it needs to keep merging in the `horizontal_execution.symtable_` for each one it merges. The new `HorizontalExecution` needs to have a combination of symbols from that as well as the mapped scalars for any conflicted symbols. It's really rather tricky.

havogt: What's still missing? Your test with the LocalScalar seems to pass?

jdahm: What I mentioned above still isn't implemented. It's creating one horizontal execution, but the code is wrong inside it.

src/gtc/passes/oir_optimizations/temporaries.py
```diff
@@ -78,12 +78,13 @@ def visit_VerticalLoop(
...
     def visit_Stencil(self, node: oir.Stencil, **kwargs: Any) -> oir.Stencil:
         tmps_to_replace = kwargs["tmps_to_replace"]
+        all_names = collect_symbol_names(node)
         return oir.Stencil(
             name=node.name,
             params=node.params,
             vertical_loops=self.visit(
                 node.vertical_loops,
-                new_symbol_name=symbol_name_creator(set(kwargs["symtable"])),
+                new_symbol_name=symbol_name_creator(all_names),
```
jdahm, line 87: This change is necessary when making HorizontalExecution a SymbolTableTrait. Temporaries could be made that conflicted with existing local names.

src/gtc/passes/oir_optimizations/utils.py
```diff
@@ -161,7 +162,7 @@ def symbol_name_creator(used_names: Set[str]) -> Callable[[str], str]:
 
     def increment_string_suffix(s: str) -> str:
         if not s[-1].isnumeric():
-            return s + "0"
+            return s + "_gen_0"
```
jdahm, line 165: This is a bit clearer name to grep for in the generated code. Sometimes frontend users would put digits after the name and it became unclear.

fthaler: Maybe even replace the if by `not re.match(r".+_gen_[0-9]+")` to have a stricter distinction between generated symbols and user symbols?

tests/test_unittest/test_gtc/common_utils.py
fthaler: Yes, that’s perfect. I left that check there to make sure that we update the handling of nested symbol tables correctly once we introduce some. But as `undefined_symbol_list` is only used at the stencil level, this new behavior should probably be always correct.

fthaler (approved):

src/gtc/passes/oir_optimizations/utils.py
jdahm: Oops, yes I meant to add something like that before. Thanks for catching that!

src/eve/traits.py
```diff
@@ -68,10 +68,6 @@ def _collect_symbols_validator(  # type: ignore  # validators are classmethods
         values["symtable_"] = cls._collect_symbols(values)
         return values
 
-    def collect_symbols(self) -> None:
-        self.symtable_ = dict()
-        self.symtable_ = self._collect_symbols(self)
-
```
jdahm, line 74: FYI, this was just unused duplicated logic from elsewhere.

jdahm:
I opened up #572 to address the clean-up of this pass, but I propose we do that in a separate step.

havogt (changes requested):
Please summarize the state of the current PR. At some point you added a test to check behavior if 2 horizontal executions are merged which have a LocalScalar of the same name. Why did you remove it? Does that work now?

tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py
```diff
@@ -172,3 +175,43 @@ def test_on_the_fly_merging_repeated():
     transformed = OnTheFlyMerging().visit(testee)
     hexecs = transformed.vertical_loops[0].sections[0].horizontal_executions
     assert len(hexecs) == 2
+
+
+def test_on_the_fly_merging_localscalars():
```
jdahm, line 180: @havogt the test is still here. I think I had an older version of this test too, but it wasn't really testing what it claimed to.

jdahm:
@havogt I updated the description at the top.

src/gtc/passes/oir_optimizations/utils.py
```diff
@@ -160,8 +161,8 @@ def symbol_name_creator(used_names: Set[str]) -> Callable[[str], str]:
     """
 
     def increment_string_suffix(s: str) -> str:
-        if not s[-1].isnumeric():
-            return s + "0"
+        if not re.match(r".+_gen_[0-9]+", s):
```
egparedes, line 164: I would compile this regular expression at module level since it's constant. It won't probably make such a big difference but just in case

src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
```diff
@@ -199,12 +236,13 @@ def visit_VerticalLoop(self, node: oir.VerticalLoop, **kwargs: Any) -> oir.Verti
     def visit_Stencil(self, node: oir.Stencil, **kwargs: Any) -> oir.Stencil:
         vertical_loops: List[oir.VerticalLoop] = []
         protected_fields = set(n.name for n in node.params)
+        all_names = collect_symbol_names(node)
         for vl in reversed(node.vertical_loops):
```
egparedes, line 240: I don't know if it is actually faster, but this for loop can be easily rewritten as a list comprehension and it's (maybe ?) simpler to understand

```python
vertical_loops = [
    *reversed(
        self.visit(
            vl,
            new_symbol_name=symbol_name_creator(all_names),
            protected_fields=protected_fields,
            **kwargs,
        )
        for vl in reversed(node.vertical_loops)
    )
]
```

src/eve/traits.py
egparedes: Just a clarification here: this is not necessarily meant to be used, it's just an additional function that the trait adds to the node class, in case the user wants to explicitly trigger a re-collection of symbols. It can be deleted if you think it's useless though

src/gtc/passes/oir_optimizations/utils.py
jdahm: Good idea. Implemented!

src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
jdahm: It turns out that `reversed` requires a Sequence or at least Reversible type (which the generator is not) so I had to make an inner list as well, but it should work.

egparedes: I realized my suggestion was completely wrong and it should be reverted. First, it requires creating an extra temporary list for `reversed` to work as @jdahm mentioned, and second, and more importantly, the other two statements inside the loop have been removed in the list comprehension:
```
access_collection = AccessCollector.apply(vl)
protected_fields |= access_collection.fields()
```
The funny thing is that tests are still passing even without these two statements, so I think the tests should be improved to catch this error.

src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
```diff
@@ -199,18 +236,20 @@ def visit_VerticalLoop(self, node: oir.VerticalLoop, **kwargs: Any) -> oir.Verti
...
+                        new_symbol_name=symbol_name_creator(all_names),
+                        protected_fields=protected_fields,
+                        **kwargs,
+                    )
+                    for vl in reversed(node.vertical_loops)
+                ]
             )
-            access_collection = AccessCollector.apply(vl)
-            protected_fields |= access_collection.fields()
+        ]
```
egparedes, line 252: Why are the tests passing if `protected_fields` is not longer updated inside the loop with the results of the `AccessCollector` pass? Either the rests are not detailed enough or the logic in the loop was previously wrong.

egparedes (changes requested):

src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
jdahm: Earlier the pass was using protected fields as a way of not creating temporaries that already existed as named fields. Now that's being done automatically in a different way.

jdahm: I did take note of that, but thanks for thinking of it!

egparedes: Ok, it makes sense. Thanks for the explanation.

egparedes (approved):
It looks good to me. Should this be merged or do you still have questions @havogt ?

PR #567, opened 2021-12-01
Field View: (very) basic type propagation and checking.

DropD:
## Description

Propagate symbol annotations into FOAST and type check / deduce in a separate pass.


src/functional/ffront/ast_passes/single_static_assign.py
```diff
@@ -108,7 +110,9 @@ def visit_AnnAssign(self, node: ast.AnnAssign) -> ast.AnnAssign:
         return node
 
     def visit_Name(self, node: ast.Name) -> ast.Name:
-        if node.id in self.state.name_counter:
+        if node.id.capitalize() in _BUILTINS:
```
egparedes, line 113: I don't understand why variable names are explicitly capitalized here instead of using lower case versions the list of builtins and compare by value without modifications. Could you elaborate on it?

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python
```
egparedes, line 1: This module is not expected to be used as executable
```suggestion
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def is_complete_type(fo_type: foast.DataType) -> bool:
+    if fo_type is None:
+        return False
+    elif isinstance(fo_type, foast.DeferredSymbolType):
+        return False
+    elif isinstance(fo_type, foast.SymbolType):
+        return True
+    return False
```
egparedes, line 31: Any reason to not to use the new pattern matching syntax here? Additionally, this function looks a lot like a custom typeguard (https://www.python.org/dev/peps/pep-0647/) to me, would it make sense to explicitly define it like one?

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -0,0 +1,95 @@
...
+        if node.id not in symtable:
+            warnings.warn(  # TODO (ricoh): raise this instead (requires externals)
+                FieldOperatorTypeDeductionError.from_foast_node(
+                    node, msg=f"Undeclared symbol {node.id}"
+                )
+            )
+            return node
+
+        symbol = symtable[node.id]
+        return foast.Name(id=node.id, type=symbol.type, location=node.location)
```
egparedes, line 61: I'm not 100% sure but it looks like this Visitor should use the `SymbolTableTrait.symtable_merger` as `context`:

```suggestion
class FieldOperatorTypeDeduction(NodeTranslator):
    """Deduce and check types of FOAST expressions and symbols."""
    
    contexts = (SymbolTableTrait.symtable_merger,)
    
    @classmethod
    def apply(cls, node: foast.FieldOperator) -> foast.FieldOperator:
        return cls().visit(node)

    def visit_FieldOperator(self, node: foast.FieldOperator, **kwargs) -> foast.FieldOperator:
        return foast.FieldOperator(
            id=node.id,
            params=self.visit(node.params, **kwargs),
            body=self.visit(node.body, **kwargs),
            location=node.location,
        )

    def visit_Name(self, node: foast.Name, **kwargs) -> foast.Name:
        symtable = kwargs["symtable"]
        if node.id not in symtable:
            warnings.warn(  # TODO (ricoh): raise this instead (requires externals)
                FieldOperatorTypeDeductionError.from_foast_node(
                    node, msg=f"Undeclared symbol {node.id}"
                )
            )
            return node

        symbol = symtable[node.id]
        return foast.Name(id=node.id, type=symbol.type, location=node.location)
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class FieldOperatorTypeDeductionError(SyntaxError, SyntaxWarning):
```
egparedes, line 64:
```suggestion
class FieldOperatorTypeDeductionError(GTSyntaxError, SyntaxWarning):
```

src/functional/ffront/type_parser.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env python
```
egparedes, line 1: Not meant to be executed directly.
```suggestion
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
egparedes, line 2: Not needed, default encoding in python 3.x
```suggestion
```

src/functional/ffront/type_parser.py
```diff
@@ -0,0 +1,167 @@
...
+# -*- coding: utf-8 -*-
```
egparedes, line 2:
```suggestion
```

src/functional/ffront/type_parser.py
```diff
@@ -0,0 +1,167 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import ast
+from typing import Any, Optional, Union
+
+from eve import NodeTranslator
+from functional.ffront import field_operator_ast as foast
+
+
+class FieldOperatorTypeError(SyntaxError):
```
egparedes, line 25:
```suggestion
class FieldOperatorTypeError(GTSyntaxError):
```

egparedes (changes requested):
It looks good, just some comments regarding style issues and simple improvements.

src/functional/ffront/foast_passes/type_deduction.py
DropD: I was not aware that black had already fixed the problem with parsing `match`.

DropD: it is not strictly a type guard (it excludes also DeferredSymbolType which is included in SymbolType), but for practical purposes we can think of it as one.

src/functional/ffront/ast_passes/single_static_assign.py
egparedes: > Or even better in a third module where it is obvious that this should be the definitive list of names that the fieldview considers reserved for types.

I like this option better.

Additionally, I don't think the names need to be capitalized at all, even if they are used for dispatching in the `make_` methods of the TypeParser. I mean, `make_int32()` looks perfectly fine to me.

src/functional/ffront/foast_passes/type_deduction.py
egparedes: > it is not strictly a type guard (it excludes also DeferredSymbolType which is included in SymbolType), but for practical purposes we can think of it as one.

Ok, thanks for the explanation, it make sense.

src/functional/ffront/ast_passes/single_static_assign.py
DropD: I was thinking the same

src/functional/ffront/builtins.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+
+from numpy import float32, float64, int32, int64
```
egparedes, line 3: We would need to implement our own definitions for these types but I think we can keep it like that for now

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -0,0 +1,143 @@
...
+        symtable = kwargs["symtable"]
+        if refine_type:
+            check_type_refinement(node, refine_type)
+            new_node = foast.FieldSymbol(id=node.id, type=refine_type, location=node.location)
+            symtable[new_node.id] = new_node
+            return new_node
+        return node
+
+
+class FieldOperatorTypeDeductionError(GTSyntaxError, SyntaxWarning):
```
egparedes, line 112: I would create a new `GTSyntaxWarning` class and inherit from it

```suggestion
class FieldOperatorTypeDeductionError(GTSyntaxWarning):
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -0,0 +1,143 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+import warnings
+from typing import Optional, TypeGuard
+
+import functional.ffront.field_operator_ast as foast
+from eve import NodeTranslator, SymbolTableTrait
+from functional.common import GTSyntaxError
+
+
+def is_complete_type(fo_type: foast.SymbolType) -> TypeGuard[foast.SymbolType]:
```
egparedes, line 24: Maybe `is_complete_symbol_type()` would be a better name here since the function is now a `SymbolType` typeguard

egparedes (commented):
Almost ready to go, just minor comments.

src/functional/ffront/builtins.py
DropD: agreed.

src/functional/ffront/foast_passes/type_deduction.py
DropD: I would not, because this inheriting from `SyntaxWarning` is only a temporary measure until external symbols can be type checked.

DropD: Other than that I don't have any concrete plans for a syntax warning and although I suspect it will come up in the future, let's put it in only when it is used somewhere.

egparedes: Ok, I understand, agreed.

egparedes (approved):
It looks good to me.

PR #462, opened 2021-07-14
Fix HorizontalExecution merging transformations

gronerl:
* prevent OTF to merge if intermediate is used later or is signature parameter
* prevent GraphMerging if iteration spaces don't match and any output is API
* remove GreedyMerging, since not used anywhere and above issue can't be fixed there easily
* also applies OTF repeatedly, which can increase the number of valid matches.

PR #571, opened 2021-12-02
Adds GTC HorizontalExecution merging

### LINKED ISSUE #558: DaCe horizontal merging pass recurses seemingly forever
This pass should be temporarily turned off and fixed. At the same time `dace` should be made an optional extra for `gt4py`.

jdahm:
## Description

Disables DaCe-based HorizontalExecution (graph_merge_horizontal_executions) merging by default.
Resolves #558 and #579.

To do:
- [x] Add unit tests


src/gtc/passes/oir_pipeline.py
```diff
@@ -70,11 +67,10 @@ def __init__(self, *, skip: Optional[Sequence[PassT]] = None):
     @staticmethod
     def all_steps() -> Sequence[PassT]:
         return [
-            graph_merge_horizontal_executions,
             AdjacentLoopMerging,
+            OnTheFlyMerging,
             LocalTemporariesToScalars,
             WriteBeforeReadTemporariesToScalars,
-            OnTheFlyMerging,
```
jdahm, line 77: Performing adjacent loop merging makes sense before the OnTheFlyMerging because that pass only looks within each `VerticalLoopSection`. Then, more temporaries can be made into scalars once horizontal executions are merged.

jdahm:
Removing the the DaCe pass from the default list uncovered the bugs with the OTF merging pass that we're addressing in #564. Let's merge that one before this. I tested that fix with this branch and it solves the unit test failures.

havogt:
@gronerl Could you comment here on the assumption that the DaCe merging made? I already forgot what it was about.

gronerl:
You probably refer to an issue that was fixed in #462. Basically, if two HorizontalExecutions are merged which write to different grid points due to their extents, they can only be merged if none of the changed fields are API and the difference therefore observable. This extent information was not available in the original list-based GreedyMerging and therefore that was disabled permanently.

jdahm:
@gronerl The dace backend is failing for the `vertical_advection_dycore` without a merging pass here (the OTF merging pass is doing nothing since this is a vertical stencil). Could we make an issue to tackle that separately, I'll disable that backend for that test, and focus on adding back the GreedyMerging?

jdahm:
Scary - the `gtc:dace` backend works again on `vertical_advection_dycore` after applying the GreedyMerging pass 🪄 .

gronerl (commented):
I'm surprised that you had you arrived at something so different from the original fusion. Looks good to me though. 

I'd like, however, to bring up the question of the name of the pass:
This is no more or less Greedy than e.g. OnTheFlyFusion, all the other passes are named according to what they do rather than the heuristic they apply. @havogt

Just `HorizontalExecutionMerging` ?

src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
```diff
@@ -18,9 +18,94 @@
...
+        **kwargs: Any,
+    ) -> oir.VerticalLoopSection:
+        horizontal_executions = [node.horizontal_executions[0]]
+        new_block_extents = [block_extents[id(horizontal_executions[-1])]]
+
+        for this_hexec in node.horizontal_executions[1:]:
+            last_extent = new_block_extents[-1]
+
+            last_writes = (
+                AccessCollector.apply(horizontal_executions[-1]).cartesian_accesses().write_fields()
```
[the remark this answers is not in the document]
jdahm: Oh, true! I had `cartesian_accesses` from a previous iteration that I hadn't removed yet. Shouldn't be necessary, but then again all writes are on-center away which are cartesian accesses so it's the same both ways. I'll make the change.

src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
```diff
@@ -18,9 +18,94 @@
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
 
 from eve import NodeTranslator, SymbolTableTrait
+from gt4py.definitions import Extent
```
jdahm, line 21: It's unfortunate that we keep needing/using this type in GTC. Would be good to port over into a separate utility to break the circular dependency (though it's read-only).

jdahm:
> I'm surprised that you had you arrived at something so different from the original fusion.

Yes, I tried to write one without looking at yours at first. It's different because we now have to map over the local scalars because they can have the same name in different horizontal executions.

PR is ready to go again!

src/gtc/passes/oir_optimizations/horizontal_execution_merging.py
jdahm: I added an issue to track this #592.

havogt:
bors try

havogt (approved):

PR #574, opened 2021-12-04
Replace eve.concepts.AnyNode by eve.concepts.BaseNode

### LINKED ISSUE #565: ClassVar parameters cannot include any type variables
[PEP-526](https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations) states that 

> Note that a ClassVar parameter cannot include any type variables, regardless of the level of nesting: ClassVar[T] and ClassVar[List[Set[T]]] are both invalid if T is a type variable.

Unfortunately the new `contexts` feature of the `NodeVisitors` use this implicitly because `eve.concepts.TreeNode = Union[AnyNode, CollectionNode]` and

```python
AnyNode = TypeVar("AnyNode", bound="BaseNode")
```

Need to find a way around this or reimplement the feature.

cc: @egparedes

jdahm:
## Description

This removes the TypeVar from concepts, which means these can be used in `ClassVar` instances.

Resolves #565.

### DIFF
--- a/src/eve/concepts.py
+++ b/src/eve/concepts.py
@@ -37,6 +37,7 @@ from .typingx import (
     Set,
     Tuple,
     TypedDict,
+    TypeVar,
     Union,
     no_type_check,
 )
@@ -115,6 +116,7 @@ class FrozenModel(pydantic.BaseModel):
 _EVE_NODE_INTERNAL_SUFFIX = "__"
 _EVE_NODE_IMPL_SUFFIX = "_"
 
+NodeT = TypeVar("NodeT", bound="BaseNode")
 ValueNode = Union[bool, bytes, int, float, str, IntEnum, StrEnum]
 LeafNode = Union["BaseNode", ValueNode]
 CollectionNode = Union[List[LeafNode], Dict[Any, LeafNode], Set[LeafNode]]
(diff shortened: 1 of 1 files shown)


src/eve/concepts.py
```diff
@@ -116,11 +115,10 @@ class Config:
 _EVE_NODE_INTERNAL_SUFFIX = "__"
 _EVE_NODE_IMPL_SUFFIX = "_"
 
-AnyNode = TypeVar("AnyNode", bound="BaseNode")
```
egparedes, line 119: I'm not sure if we should delete the `AnyNode` TypeVar definition, since it has a different meaning than `BaseNode`. Example:
```
def f1(a: BaseNode) -> BaseNode:
    ...

def f2(a: AnyNode) -> AnyNode:
    ...
```
`f2` always returns the same BaseNode subclass it got as an argument, while `f1` may return a different subclass. Maybe it could be renamed to `NodeT` to make clear that is a TypeVar.

egparedes (changes requested):
The tests should be fixed. Other than that it looks fine, although I would like to discuss if `AnyNode` should be removed as you propose or just renamed to `NodeT`

jdahm:
@egparedes Apologies, I meant to convert this to a draft before I left for the day on Friday. There many cascading type issues that mypy catches that need to be solved first. Your comment might help with that.

jdahm:
Closing this for inactivity. Will re-examine when I have time.

PR #575, opened 2021-12-06
Ffront externals

egparedes:
## Description

Add external symbols handling to new functional frontend


src/eve/typingx.py
```diff
@@ -218,3 +222,133 @@ def resolve_type(
                 raise e
 
     return actual_type
+
+
+def _collapsable_type_args(*args: Any) -> tuple[bool, tuple]:
+    if len(args) and all(args[0] == a for a in args):
```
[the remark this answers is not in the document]
DropD: Possibly you meant `if len(args) > 1`, that would make sense.

src/eve/typingx.py
```diff
@@ -218,3 +222,134 @@ def resolve_type(
...
+        return value
+
+    elif value in (None, types.NoneType):
+        return None
+
+    elif isinstance(value, tuple):
+        unique_type, args = _collapsable_type_args(*(recursive_get(item) for item in value))
+        if unique_type and len(args) > 1:
+            return tuple[args[0], ...]
+        elif len(args):
```
DropD, line 307:
```suggestion
        elif args:
```

DropD, line 307:
```suggestion
        elif args:
```

src/eve/typingx.py
```diff
@@ -218,3 +222,134 @@ def resolve_type(
...
+            return tuple[args]
```
DropD, line 308: My intuition would have been that it should be `tuple[*args]`

src/functional/ffront/ast_passes/stringify_annotations.py
```diff
@@ -0,0 +1,82 @@
...
+        a$0 = 1
+        a$1 = 2 + a$0
+        a$2 = 3 + a$1
+        return a$2
+
+    Note that each variable name is assigned only once and never updated / overwritten.
+
+    Note also that after parsing, running the pass and unparsing we get invalid but
+    readable python code. This is ok because this pass is not intended for
+    python-to-python translation.
```
DropD, line 55: Seems to be accidentally still left over from copy pasting the SSA pass.

src/functional/ffront/ast_passes/stringify_annotations.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+
+    @classmethod
+    def apply(cls, node: ast.AST) -> ast.AST:
+        return cls().visit(node)
+
+    def visit_arg(self, node: ast.arg) -> ast.arg:
+        node.annotation = self._stringify_annotation(node.annotation)
+        result = self.generic_visit(node)
+        assert isinstance(result, ast.arg)
```
DropD, line 65: are these asserts really necessary or are they leftovers from debugging?

src/eve/typingx.py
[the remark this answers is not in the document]
egparedes: Yes, I think it was a leftover from a previous version. I'll fix it.

src/functional/ffront/func_to_foast.py
```diff
@@ -16,23 +16,153 @@
...
+    param_names = set()
+    imported_names = set()
+    local_names = set()
+    for name in func_st.get_locals():
+        if (s := func_st.lookup(name)).is_imported():
+            imported_names.add(name)
+        elif s.is_parameter:
+            param_names.add(name)
+        else:
+            local_names.add()
```
DropD, line 101: is this `.add()` empty on purpose? If so, why is it there?

src/functional/ffront/func_to_foast.py
```diff
@@ -16,23 +16,153 @@
...
+@dataclass(frozen=True)
+class SourceDefinition:
+    source: str
+    filename: str = MISSING_FILENAME
+    starting_line: int = 1
+
+    def __iter__(self) -> Iterator[tuple[str, ...]]:
+        yield from iter((self.source, self.filename, self.starting_line))
+
+    from_function = staticmethod(_make_source_definition_from_function)
```
DropD, line 126: What's the thought process behind this pattern?

src/eve/typingx.py
egparedes: Fixed

src/functional/ffront/func_to_foast.py
```diff
@@ -133,35 +261,79 @@ def apply(
...
+        if node.module not in fbuiltins.MODULE_BUILTIN_NAMES:
+            self._make_syntax_error(
+                node, f"Only 'from' imports from {fbuiltins.MODULE_BUILTIN_NAMES} are supported"
+            )
+
+        symbols = []
+        for alias in node.names:
+            if alias.name not in self.externals_defs:
+                raise self._make_syntax_error(
+                    node, message="Missing imported symbol '{alias.name}'"
```
DropD, line 316: Do I understand this correctly: this error message is triggered when the user forgets to supply `X` in the externals and then tries to `from externals import X`?

src/eve/typingx.py
egparedes: That would be a syntax error since you cannot unpack inside a `[]` operator expression. I guess it's because `getitem` only accepts a single argument (although you can use unpacking if you call `__getitem__` directly, but that would cause a run-time error)

src/functional/ffront/func_to_foast.py
egparedes: Yes, that's correct.

src/functional/ffront/symbol_makers.py
```diff
@@ -0,0 +1,192 @@
...
+            id=name,
+            type=symbol_type,
+            namespace=namespace,
+            location=location,
+        )
+    else:
+        raise common.GTTypeError(f"Impossible to map '{value}' value to a Symbol")
+
+
+class FieldOperatorTypeError(common.GTTypeError):
```
DropD, line 183: Will it be possible to recover the ability to tell the user where (source location) the type error comes from?

tests/functional_tests/ffront_tests/test_interface.py
```diff
@@ -229,37 +227,37 @@ def clashing(inp: Field[..., "float64"]):
...
-    with pytest.raises(
-        FieldOperatorSyntaxError,
-        match=r"functions can only be called directly! \(.*\)",
-    ):
-        _ = FieldOperatorParser.apply_to_func(call_expr)
+#     with pytest.raises(
+#         FieldOperatorSyntaxError,
+#         match=r"functions can only be called directly! \(.*\)",
+#     ):
+#         _ = FieldOperatorParser.apply_to_function(call_expr)
```
DropD, line 260: Is this commented out because we want to recover this test in changed form later?

DropD (changes requested):
Looks ok, except for the few comments which are already there. Pity about the loss of source location in type errors though.

src/functional/ffront/func_to_foast.py
egparedes: Good catch. It's a typo

src/functional/ffront/func_to_foast.py
egparedes: The `__iter__` method is added here just to support convenient unpacking of the `SourceDefinition` class. Example:

```python
srd_def = SourceDefinition(source, filename, starting_line)
source, filename, starting_line = src_def
```

DropD: I was referring more to the

```python
from_function = staticmethod(...)
```

pattern, rather than 

```python
@staticmethod
def from_function(...):
    ...
```

tests/functional_tests/ffront_tests/test_interface.py
egparedes: I don't think it is actually required. I will delete it.

src/functional/ffront/func_to_foast.py
egparedes: Ah, ok, sorry about the confusion. I decided to separate the implementation of the creation method from the class to enhance readability, since methods are not very short and inlining them in the class body polluted the simplicity of the dataclass definition with code that is only related to its creation, not in how the object behaves.

DropD: That's fine with me. The only thing that might cause some confusion is that we are creating a protected function and then we are "leaking" it into the public API. The uninitiated dev might assume that the signatures there are implementation details.

DropD (commented):
👍  the only changes remaining are to make the from_function functions public and the todo for source location in typing exceptions

src/functional/ffront/func_to_foast.py
egparedes: You're right, I made them public to avoid inconsistencies.

src/functional/ffront/symbol_makers.py
egparedes: It should be possible and we should do it. I added a TODO to do it in a following PR.

DropD (approved):

Issue #31, opened 2020-05-27
Slices of gt4py storages cannot be copied


mcgibbon:
I have a piece of code which extracts variables (kind of like tracer variables) out of a larger array. After this, I needed to copy the array, and ran into this issue (shown as a MCVE):

```python
>>> arr = gt4py.storage.empty("numpy", default_origin=[0, 0, 0], shape=[10, 10, 10], dtype=float)
>>> s = arr[:, :, 0]
>>> copy.deepcopy(s)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.7/copy.py", line 161, in deepcopy
    y = copier(memo)
  File "/usr/src/gt4py/src/gt4py/storage/storage.py", line 350, in __deepcopy__
    res = super().__deepcopy__(memo=memo)
  File "/usr/src/gt4py/src/gt4py/storage/storage.py", line 186, in __deepcopy__
    managed_memory=not isinstance(self, ExplicitlySyncedGPUStorage),
  File "/usr/src/gt4py/src/gt4py/storage/storage.py", line 38, in empty
    shape=shape, dtype=dtype, backend=backend, default_origin=default_origin, mask=mask
  File "/usr/src/gt4py/src/gt4py/storage/storage.py", line 141, in __new__
    shape = storage_utils.normalize_shape(shape, mask)
  File "/usr/src/gt4py/src/gt4py/storage/utils.py", line 49, in normalize_shape
    "len(shape) must be equal to len(mask) or the number of 'True' entries in mask."
ValueError: len(shape) must be equal to len(mask) or the number of 'True' entries in mask.
```

I may find a way to work around the issue, but copy.deepcopy should probably work when used on slices?

egparedes:
Related to (and probably fixed by): #28 #29

jdahm:
This triggers an interesting new error:
```python
>>> s = arr[:, :, 0]
>>> copy.deepcopy(s)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/nix/store/4i5w9iv836cg6hgr02g6phds6br8nas0-python3-3.8.9/lib/python3.8/copy.py", line 153, in deepcopy
    y = copier(memo)
  File "/Users/johannd/gt4py/src/gt4py/storage/storage.py", line 181, in __deepcopy__
    return self.copy()
  File "/Users/johannd/gt4py/src/gt4py/storage/storage.py", line 381, in copy
    res = super().copy()
[...]

PR #576, opened 2021-12-07
Disallow dimension reducing slicing of storages 

tehrengruber:
## Description

Disallow dimension reducing slicing of storage to prevent inconsistent metadata.

__Problem__
Taking a slice of a gt storage that results in a storage of lower dimension has incorrect metadata, e.g. origin, mask are just the "old" values. Fixing this properly is non trivial as it requires recognizing which dimension was collapsed. However the slices given to `__getitem__` are not accessible anymore in `__array_finalize__`, i.e. the point where the metadata is updated. Even in case we would write the machinery to do this, we would need to update all metadata such that it is equal to creating a new storage with the same data. Something that is error prone and time consuming given the current state of the storages.

Example from #31:
```
arr = gt4py.storage.empty("numpy", default_origin=[0, 0, 0], shape=[10, 10, 10], dtype=float)
s = arr[:, :, 0]
```

__Proposed solution__

In the light of advances to simplify storages in the past and the aforementioned problems I propose to disallow dimension reducing slicing of storages all together. For users that require this functionality using ~~`Storage.data`~~ `Storage.to_numpy()`, slicing and creating a new storage from the result provides a clean workaround. If this use case occurs frequently in user code we could provide a utility function or similar.

@jdahm What is your take on this? Does this happen frequently in FV3?

## Requirements


jdahm:
Hi @tehrengruber, I like this idea. We discussed it this morning in the dev meeting and it looks as though fv3 uses a wrapper around the storage data anyway when it slices like that.

tehrengruber:
bors try

docs/gt4py/quickstart.rst
```diff
@@ -207,6 +207,10 @@ The ``default_origin`` parameter plays two roles:
 #. If when calling the stencil, no other `origin` is specified, this value is where the `iteration domain` begins, i.e.
    the grid point with the lowest index where a value is written.
 
+.. note::
+    Slicing a storage with reduced dimension, e.g. `field_a[:, :, 0]`, is not supported. Use `Storage.data` to retrieve
+    a slicable array, e.g. `field_a.data[:, :, 0]`, and create a new storage from it if needed.
```
jdahm, line 212: 👍

tehrengruber:
bors try

jdahm (approved):
Looks right to me!

tehrengruber:
bors try

src/gt4py/storage/storage.py
```diff
@@ -207,6 +207,11 @@ def __array_finalize__(self, obj):
                     raise RuntimeError(
                         "Meta information can not be inferred when creating Storage views from other classes than Storage."
                     )
+                if self.ndim != obj.ndim and self.ndim != 0:
```
gronerl, line 210: What case does the self.ndim != 0 cover? So it's still allowed to access scalar elements?

src/gt4py/storage/storage.py
```diff
@@ -427,6 +437,18 @@ def _construct(cls, backend, dtype, default_origin, shape, alignment, layout_map
     def data(self):
         return self._device_field
 
+    def to_numpy(self, copy=False):
+        self.synchronize()
```
gronerl, line 441:
```suggestion
        self.device_to_host()
```

src/gt4py/storage/storage.py
```diff
@@ -427,6 +437,18 @@ def _construct(cls, backend, dtype, default_origin, shape, alignment, layout_map
...
+        if copy:
+            return copy.deepcopy(np.asarray(self))
+        return np.asarray(self)
+
+    def to_copy(self, copy=False):
+        self.synchronize()
```
gronerl, line 447:
```suggestion
    def to_cupy(self, copy=False):
        self.host_to_device()
```
note also the typo in the method name

src/gt4py/storage/storage.py
```diff
@@ -427,6 +437,18 @@ def _construct(cls, backend, dtype, default_origin, shape, alignment, layout_map
     def data(self):
         return self._device_field
 
+    def to_numpy(self, copy=False):
+        self.synchronize()
+        if copy:
+            return copy.deepcopy(np.asarray(self))
```
gronerl, line 443:
```suggestion
            return np.array(self, subok=False)
```
Is the numpy way of doing copies. Maybe needs double checking if this doesn't result in a storage with our numpy patch-hack.

src/gt4py/storage/storage.py
```diff
@@ -427,6 +437,18 @@ def _construct(cls, backend, dtype, default_origin, shape, alignment, layout_map
...
+        return np.asarray(self)
+
+    def to_copy(self, copy=False):
+        self.synchronize()
+        if copy:
+            return copy.deepcopy(cp.asarray(self))
```
gronerl, line 449:
```suggestion
            return cp.array(self)
```

tests/test_unittest/test_storage.py
```diff
@@ -944,7 +944,7 @@ def test_sum_gpu():
         shape=shape,
     )
 
-    q1[i1 : i2 + 1, jslice, 0] = cp.sum(q2[i1 : i2 + 1, jslice, :], axis=2)
+    q1[i1 : i2 + 1, jslice, 0] = cp.sum(q2.data[i1 : i2 + 1, jslice, :], axis=2)
```
gronerl, line 947: Looks like this test is supposed testing suming over storages (which is a ufunc) if we just use .data this test tests nothing - i.e. we should probably remove the test as it's just no longer supported to do sums.

gronerl (changes requested):

src/gt4py/storage/storage.py
```diff
@@ -427,6 +437,18 @@ def _construct(cls, backend, dtype, default_origin, shape, alignment, layout_map
...
+        return self.view(cp.asarray(self))
```
gronerl, line 450:
```suggestion
        return cp.asarray(self)
```
self.view() will only work to upcast to the ndarray baseclass or possibly other np.ndarray subclasses, which cp.ndarray is not.

src/gt4py/storage/storage.py
tehrengruber: This is not 100% clean, but I would say the benefits of allowing scalar storages (with incorrect metadata) overweight here. For example `numpy.allclose` relies on reducing an array/storage of booleans to a scalar boolean, so disallowing this case would mean require all occurences.

gronerl:
bors try

gronerl (approved):

Issue #570, opened 2021-12-01
Failure of `graph_merge_horizontal_executions`


fthaler:
The Dace-pass `graph_merge_horizontal_executions` fails on stencils that write to a different set of fields in different intervals. E.g.:

```python
from gt4py.gtscript import PARALLEL, Field, IJK, computation, interval, stencil


@stencil(backend="gtc:gt:cpu_ifirst")
def foo(a: Field[IJK, float], b: Field[IJK, float]):
    with computation(PARALLEL):
        with interval(0, 1):
            a = 42
        with interval(1, None):
            a = 42
            b = 42
```

Ther error message in the particular example is `networkx.exception.NodeNotFound: Either source a or target VerticalLoop_* is not in G`.

PR #578, opened 2021-12-09
Fix bug where node was added to SDFG too late if no read acces was there

gronerl:
Presumably a fix for #570 @fthaler can you verify that this fixes your original issue?

You wrote that it "fails on stencils that write to a different set of fields in different intervals" but this fix for your reproducer is not related to that reason.


fthaler (approved):
Thansk! Looks good to me, reasons for failures seem to be identical.

gronerl:
bors try

PR #583, opened 2021-12-16
Update compilation flag -std=c++14 to -std=c++17

gronerl:


gronerl:
bors try

gronerl:
Looks like the requirement of c++17 in gridtools forces us to either use an old commit on gridtools 2 or drop support for cuda 10 (and therefore have to update the CI pipeline) @DropD @havogt

havogt:
I think we should update CI and make GT4Py ready for C++17.

jdahm:
This wouldn't have suddenly been a problem if we pinned the GridTools version... Just sayin' ;)

havogt:
bors try

havogt (approved):

PR #473, opened 2021-07-16
Pin the GridTools C++ versions used

jdahm:
## Description

Currently uses the latest versions of `release_v1.1` and does not check if the code is out of date. This new version pins the version. It allows commits that can descendants of the pinned commit, but it will flag installs when the pinned version is later than the one on disk.

Resolves #457.

PR #584, opened 2021-12-16
Freeze GridTools versions to releases

havogt:
As we are decoupling the development of the functional GT4Py from the Cartesian, I am fine with setting GridTools to a release version instead of master.
Fixes #473.


havogt:
bors try

egparedes (commented):
We could also add a check when loading `gtcpp.backend` to raise a warning if the current GridTools C++ installed version does not match the expected version. Just a warning to make sure users are aware that something may break when using other versions (and maybe also how to update/download the right version using `gt_src_manager`).

PR #586, opened 2021-12-17
Add source information to GT4Py IRs

tbennun:
## Description

This PR adds and propagates source line information to the GT4Py IRs:

- [x] Python AST -> Definition IR
- [x] Definition IR -> GTIR
- [x] GTIR -> OIR
- [x] OIR -> SDFG -> OIR


src/gtc/dace/dace_to_oir.py
```diff
@@ -74,7 +74,7 @@ def sdfg_arrays_to_oir_decls(sdfg: dace.SDFG) -> Tuple[List[oir.Decl], List[oir.
     return params, decls
 
 
-def convert(sdfg: dace.SDFG) -> oir.Stencil:
+def convert(sdfg: dace.SDFG, stencil: Optional[oir.Stencil] = None) -> oir.Stencil:
```
jdahm, line 77: This does not actually use stencil, only the attribute on stencil. Before merging, could you convert this to something like`*, loc: Optional[common.SourceLocation] = None`?

jdahm:
Before we merge I'd like to see if we could implement an idea from #587 to simplify this.

havogt:
@jdahm Are you actively working on the idea of #587?

src/gtc/dace/dace_to_oir.py
tbennun: sounds good 👍 will do

DropD:
I think #587 would indeed simplify this as it would many other things. However, I also think this is useful in the meantime even if it will be made obsolete relatively soon.

DropD (approved):
Looks good and I suppose it will help when inspecting SDFGs, I wish I had the source locations there while working on the halo exchanges.

tbennun:
> Looks good and I suppose it will help when inspecting SDFGs, I wish I had the source locations there while working on the halo exchanges.

yes, that is exactly the intent :) it will also help name the GPU kernels after their respective stencil names for clarity in profiles

havogt:
bors try

PR #590, opened 2021-12-17
GTScript frontend check to ensure consistency of data dimensions.

### LINKED ISSUE #588: GTScript does not error when omitting a data index
The frontend (or somewhere else in the toolchain) should omit an error if the variable is supposed to be indexed with data dimensions, and they are omitted.

A further nicety would be to bounds-check the indices.

jdahm:
## Description

Checks that the number of data dims matches and that no accesses are out of bounds.

Resolves #588.


tehrengruber:
The errors could be more expressive, but it's fine.

PR #603, opened 2022-01-05
Fix issue uncovered with new pydantic version, and update pre-commit and hook versions

### LINKED ISSUE #551: New python version introduces bug with pydantic
Python 3.9.8 and Python 3.10.1 introduce an API change in `typing.ForwardRef` that introduces a bug in `pydantic` (for details see https://github.com/samuelcolvin/pydantic/issues/3401 and https://github.com/samuelcolvin/pydantic/pull/3403).

In order to get around this for the moment we downgraded the Python 3.9 version used on CI to 3.9.7, but we should prioritize fixing https://github.com/GridTools/gt4py/issues/548 and get the latest version with the fix ASAP.

jdahm:
## Description

Updates pydantic and a few pre-commit hook libraries, but does not unpin python 3.9 because that triggers another issue.

### DIFF
--- a/.github/workflows/gt4py-tox.yml
+++ b/.github/workflows/gt4py-tox.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: [3.8, 3.9.7]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


jdahm:
@egparedes @havogt I thought we could remove the pin in `gt4py-tox.yml` to allow later versions of `3.9` but doing so causes a check in the standard library's `typing.py` to catch (what I think is) a design issue in eve.

egparedes (approved):
It looks good to me.

havogt:
bors try

havogt (approved):

PR #604, opened 2022-01-06
CI: run tests daily

### LINKED ISSUE #602: Nightly tests to catch issues
There is no ideal solution to dependency management and testing:
1. If packages are pinned in testing, then either (a) the package becomes uninstallable along with other packages that have version clashes, or (b) a user comes along with a new installation with broken tests, since they are using newer versions than those pinned.
2. If packages are not pinned (our current approach), we currently do not catch issues with gt4py until we work on another feature and launch CI. Additionally, users might catch issues before we do.

To make (2) less frustrating for development, we should consider adding nightly build tests somewhere to catch errors. There is a GitHub Action to support this: https://github.com/marketplace/actions/deploy-nightly

havogt:
Fixes #602

### DIFF
--- a/.github/workflows/gt4py-tox.yml
+++ b/.github/workflows/gt4py-tox.yml
@@ -1,6 +1,8 @@
 name: Tox (CPU only)
 
 on:
+  schedule:
+    - cron: '30 5 * * *'
   push:
     branches:
     - master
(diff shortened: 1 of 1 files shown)


jdahm:
18:30 Zurich / 9:30 Seattle sounds good to me.

jdahm (approved):

PR #608, opened 2022-01-11
Fortran bindings prototype

havogt:
This implements a very simple prototype for getting Fortran bindings via the command line interface (`gtpyc`). It uses cpp_bindgen to actually generate the Fortran (and C) files. Most likely this is not the best choice longer term (e.g. there is an extra codegen step involved by executing a binary; you don't get nice names for the Fortran functions; ...).

To test, do the following
1. `gtpyc gen --backend=gtc:gt:cpu_kfirst src/gscp_graupel.py --output-path=./generated_stencils --bindings fortran`  will generate a c++ file that instantiates the stencil, and a CMakeLists.txt for compilation.
2. Go to the generated code directory where the CMakeLists.txt is located and do, e.g. `mkdir build && cd build && cmake .. -DCMAKE_INSTALL_PREFIX=$(pwd)/../install && make install`, which will generate the bindings and compile the C++ library.
3. Now compile the Fortran module that was generated and link it to your driver, e.g. something like `gfortran install/src/cpp_bindgen/handle.f90 install/src/cpp_bindgen/array_descriptor.f90 <the_generated_module.f90> <my_driver.f90> install/lib/libdemo_copy_module.a -fopenmp -lstdc++ install/lib/libcpp_bindgen_handle.a install/lib/libcpp_bindgen_generator.a`

TODO:
- the fix for the CLI for gtc is super hacky, most likely I broke the normal mode

### DIFF
--- a/src/gt4py/backend/gtc_backend/gtcpp/backend.py
+++ b/src/gt4py/backend/gtc_backend/gtcpp/backend.py
@@ -164,7 +164,7 @@ class FortranBindingsCodegen(codegen.TemplatedGenerator):
                     name=node.name, sid_ndim=sid_ndim, dtype=dtype, stride_kind=stride_kind
                 )
             else:
-                return node.name
+                return f"gridtools::sid::shift_sid_origin({node.name}, gridtools::array<int, 3>{{offset_x, offset_y,offset_z}})"
 
     def visit_ScalarDecl(self, node: gtir.ScalarDecl, **kwargs):
         if "external_arg" in kwargs:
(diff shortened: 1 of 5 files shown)


src/gt4py/backend/gtc_backend/gtcpp/backend.py
```diff
@@ -134,14 +142,139 @@ def apply(cls, root, *, module_name="stencil", **kwargs) -> str:
...
+
+FetchContent_Declare(
+  cpp_bindgen
+  GIT_REPOSITORY https://github.com/GridTools/cpp_bindgen.git
+  GIT_TAG        master # consider replacing master by a tagged version
+)
+FetchContent_MakeAvailable(cpp_bindgen)
+
+bindgen_add_library(${libname}_module SOURCES ${filename})
+target_link_libraries(${libname}_module PRIVATE GridTools::stencil_cpu_kfirst)
```
havogt, line 239: TODO: here is some hard-coded cpu_kfirst dependency

PR #615, opened 2022-01-14
Fix bug in computing block extents

jdahm:
## Description

When looping over the stencil and computing block field and block extents, only the vertical loop sections were visited in reverse order, whereas it should have been the horizontal executions.


havogt (commented):
As you already indicated in the slack message, we should add tests here.

jdahm:
bors try

tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py
```diff
@@ -35,6 +37,22 @@
...
+                body=[AssignStmtFactory(left__name="tmp", right__name="input")]
+            ),
+            HorizontalExecutionFactory(
+                body=[AssignStmtFactory(left__name="out", right__name="tmp", right__offset__i=1)]
+            ),
+        ]
+    )
+    hexecs = stencil.vertical_loops[0].sections[0].horizontal_executions
+    block_extents = compute_horizontal_block_extents(stencil)
+    assert block_extents[id(hexecs[0])] != Extent.zeros(ndims=2)
```
fthaler, line 53: Better check what the value should be than what it should not be.

tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py
```diff
@@ -35,6 +37,22 @@
 )
 
 
+def test_horiz_exec_extents():
+    stencil = StencilFactory(
+        vertical_loops__0__sections__0__horizontal_executions=[
+            HorizontalExecutionFactory(
+                body=[AssignStmtFactory(left__name="tmp", right__name="input")]
+            ),
```
fthaler, line 45: Just having `HorizontalExecutionFactory(body__0__left__name="tmp")` should also work. You do not have to specify the `right__name`, a unique value is generated automatically (and omission helps the reader to understand that its value is irrelevant).

tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py
```diff
@@ -35,6 +37,22 @@
...
+            HorizontalExecutionFactory(
+                body=[AssignStmtFactory(left__name="out", right__name="tmp", right__offset__i=1)]
```
fthaler, line 47: Same here.

fthaler (changes requested):
Looks good, just added some minor suggestions about the provided test. A bit surprising that there weren’t more errors because of this bug.

jdahm:
bors try

fthaler (approved):

PR #616, opened 2022-01-15
Remove dawn backends

### LINKED ISSUE #595: Remove dawn integration
The dawn integration is only present in the legacy backends which are going to be removed. `dawn` will also need to be removed as an extra dependency, and the results involving the integration also removed.

jdahm:
## Description

Remove the deprecated Dawn backends. Resolves #595.


havogt:
bors try

havogt (approved):

PR #617, opened 2022-01-15
Make DaCe an extra

### LINKED ISSUE #594: Make dace an optional dependency
Once #571 is merged, we can make dace an optional dependency. This would make `dace` an extra for gt4py, and if it's not present the `gtc:dace` backend will not be registered.

jdahm:
## Description

Make dace an optional dependency. Resolves #594.


docs/gt4py/conf.py
```diff
@@ -101,7 +101,7 @@
     "special-members": "__call__",
     "undoc-members": True,
 }
-autodoc_mock_imports = []
+autodoc_mock_imports = ["dace"]
```
havogt, line 104: @egparedes any better ways to make this work?

egparedes: I don't think so. It looks to me like this is the proper way to make it work

havogt:
bors try

jdahm:
bors try

havogt:
bors try

tox.ini
```diff
@@ -20,6 +20,7 @@ whitelist_externals =
     g++
     ldd
 extras =
```
egparedes, line 22: With  the current settings `dace` is always installed, which means that a vanilla `gt4py` installation is never tested.
I would suggest to add a new factor to the matrix to run either vanilla or dace-flavored gt4py (and also add a similar change in  `.github/workflows/gt4py-tox.yml` to run both branches in parallel)

```
envlist =
    py{38,39}-{internal,dace}-{cpu,cuda,cuda90,cuda91,cuda92,cuda100,cuda101}

extras =
    testing
    dace: dace
    cuda: cuda
    cuda90: cuda90
    cuda91: cuda91
    cuda92: cuda92
    cuda100: cuda100
    cuda101: cuda101

[testenv:py{38,39}-{internal,dace}-cpu]
...

[testenv:py{38,39}-{internal,dace}-{cuda,cuda90,cuda91,cuda92,cuda100,cuda101}]
...

```

egparedes (commented):
It looks good to me, but CI settings should be modified if we want to test a gt4py installation without dace, because right now is always installed anyway,

tox.ini
havogt: I think Enrique proposes internal backends only and internal **plus** dace

havogt:
bors try

tox.ini
egparedes: > I think Enrique proposes internal backends only and internal **plus** dace
Yes, that was my proposal.

tox.ini
```diff
@@ -3,7 +3,7 @@
 
 [tox]
 envlist =
-    py{38,39}-internal-{cpu,cuda,cuda90,cuda91,cuda92,cuda100,cuda101}
+    py{38,39}-{internal,all}-{cpu,cuda,cuda90,cuda91,cuda92,cuda100,cuda101}
```
egparedes, line 6: Everywhere else in this PR the new factor is called `with-externals` instead of `all`, so it should be updated accordingly here. However, `with-externals` contains a dash and most likely is going to mess up with `tox` matrix/factors system, so it should be replaced with something else (ideally a single word, but probably underscores would be also fine).

egparedes (changes requested):
The new factor is used with different names in two places and it should not contain dashes for compatibility with tox. Other than that, it looks good.

.github/workflows/gt4py-tox.yml
```diff
@@ -20,6 +20,11 @@ jobs:
     strategy:
       matrix:
         python-version: [3.8, 3.9.7]
+        backends: [internal-cpu, with-externals-cpu]
+        exclude:
+        - python-version: 3.8
+          backends: internal-cpu
```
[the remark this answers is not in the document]
havogt: the other 2 are running **all** backends, here we are only testing if all works in case external backends are not there. Unless you expect that there could be a case where
- 3.9 with only internal works
- 3.8 with all works
- 3.8 with only internal breaks
That would be a python version dependent code that relies on dace being installed.

tox.ini
havogt: I don't remember why I switched from all to with-externals (even it was only 2 hours ago). I'll go back to all.

havogt:
bors try

.github/workflows/gt4py-tox.yml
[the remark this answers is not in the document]
egparedes: Ok, I see. It makes sense to save CI hours.

.github/workflows/gt4py-tox.yml
```diff
@@ -45,4 +50,4 @@ jobs:
         pyversion_no_dot=${pyversion_no_dot//./}
         pyversion_no_dot=${pyversion_no_dot:0:2}
         pip install tox clang-format
```
egparedes, line 52: Minor issue unrelated to this PR: I realized that installing `clang-format` does not make sense since tox will create a new venv to run the tests. Most likely the correct way would be adding `clang-format` to the `extras` section in tox.ini

egparedes (approved):
It looks good.

.github/workflows/gt4py-tox.yml
havogt: Ok, let's do separate.

tox.ini
jdahm: I read it that way too, but then I thought of all the CI time we already spend 😆

PR #625, opened 2022-01-21
Ff typing

DropD:
Add type deduction and autolifting for some common cases.


src/functional/ffront/field_operator_ast.py
```diff
@@ -62,21 +69,42 @@ class ScalarType(DataType):
...
+        return f"tuple{self.types}"
+
 
 class FieldType(DataType):
     dims: Union[list[Dimension], Literal[Ellipsis]]  # type: ignore[valid-type,misc]
     dtype: ScalarType
 
+    def __str__(self):
+        dims = (
+            "..." if self.dims is Ellipsis else "[" + ", ".join(dim.name for dim in self.dims) + "]"
```
egparedes, line 92:
```suggestion
            "..." if self.dims is Ellipsis else f"[{', '.join(dim.name for dim in self.dims)}]"
```

src/functional/ffront/field_operator_ast.py
```diff
@@ -165,7 +201,7 @@ class Compare(Expr):
 
 
 class Call(Expr):
-    func: Name
+    func: Expr
```
egparedes, line 204: Why is this an expression and not just a name?

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
+            ):
+                if dtype_kind is foast.ScalarKind.BOOL:
+                    return True
+        return False
+
+    def can_be_refined_to(self, other: "TypeInfo") -> bool:
+        if self.is_any_type:
+            return True
+        if self.is_complete:
+            return False
```
egparedes, line 120: Shouldn't it be true in this case if the passed type is the same?
```suggestion
            return self.type == other.type
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
 
 
+@dataclass
+class TypeInfo:
+
+    type: foast.SymbolType  # noqa: A003
+
+    @property
+    def is_complete(self) -> bool:
+        return is_complete_symbol_type(self.type)
```
egparedes, line 69: Not related to this PR, but I wonder if it makes sense to define `is_complete_symbol_type()` as a `TypeGuard[foast.SymbolType]` if it returns `False` for `foast.DeferredSymbolType`, which is a subclass of `foast.SymbolType`

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
+        if self.is_complete:
+            return False
+        return (self.type is None) or (self.type.constraint is None)
+
+    @property
+    def constraint(self) -> Optional[foast.SymbolType]:
+        if self.is_complete:
+            return self
+        elif not self.is_any:
+            return self.constraint
```
egparedes, line 82: I would add an explicit `return None` either in an extra `else` branch or directly in the body of the function

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
+
+    @property
+    def is_complete(self) -> bool:
+        return is_complete_symbol_type(self.type)
+
+    @property
+    def is_any_type(self) -> bool:
+        if self.is_complete:
+            return False
+        return (self.type is None) or (self.type.constraint is None)
```
egparedes, line 75: Just a comment: have you thought about merging both conditions into a single `and` expression?

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
+
+    @property
+    def constraint(self) -> Optional[foast.SymbolType]:
+        if self.is_complete:
+            return self
+        elif not self.is_any:
```
egparedes, line 81: Maybe...
```suggestion
        elif not self.is_any_type:
```

src/functional/ffront/foast_passes/type_deduction.py
egparedes, line 82: Doesn't this create a infinite recursive call? Did you mean
```suggestion
            return self.type.constraint
```?

File: src/functional/ffront/foast_passes/type_deduction.py [hunk 8]
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
+        return is_complete_symbol_type(self.type)
+
+    @property
+    def is_any_type(self) -> bool:
+        if self.is_complete:
+            return False
+        return (self.type is None) or (self.type.constraint is None)
+
+    @property
+    def constraint(self) -> Optional[foast.SymbolType]:
```
egparedes, line 78: When an actual constraint is returned, I think the type hint is wrong..
```suggestion
    def constraint(self) -> Optional[foast.SymbolType, Type[foast.SymbolType]]:
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
+        if self.is_complete:
+            return isinstance(self.type, foast.ScalarType)
+        else:
+            return isinstance(self.constraint, foast.ScalarType)
+        return False
+
+    @property
+    def is_arithmetic_compatible(self) -> bool:
+        match self.type:
+            case foast.FieldType(dtype=foast.ScalarType(kind=dtype_kind)) | foast.ScalarType(
```
egparedes, line 99: Where is `dtype_kind` defined?

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
+                kind=dtype_kind
+            ):
+                if dtype_kind is not foast.ScalarKind.BOOL:
+                    return True
+        return False
+
+    @property
+    def is_logics_compatible(self) -> bool:
+        match self.type:
+            case foast.FieldType(dtype=foast.ScalarType(kind=dtype_kind)) | foast.ScalarType(
```
egparedes, line 109: Where is `dtype_kind` defined?

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
+    def can_be_refined_to(self, other: "TypeInfo") -> bool:
+        if self.is_any_type:
+            return True
+        if self.is_complete:
+            return False
+        if self.constraint:
+            if other.is_complete:
+                return isinstance(other.type, self.constraint)
+            elif other.constraint:
+                return self.constraint is other.constraint
```
egparedes, line 125: Do `self.constraint` and `other.constraint` need to match exactly or one of them could be a subclass of the other?

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
+        return False
+
+
+def dims_promotable(left: TypeInfo, right: TypeInfo) -> bool:
```
egparedes, line 129: It's not critical but I think it's more appropriate use  the "broadcast" instead of "promotion" when talking about dimensions and not types.

egparedes (commented):
Just a first set of comments about the actual code. I will continue with the tests next week.

src/functional/ffront/field_operator_ast.py
DropD: This was a temporary workaround for shifting temporaries:

```python
temp = inp_a + inp_b
return temp(Offset[value])
```

Gets inlined to `(inp_a + inp_b)(Offset[value])`. The longterm solution is to create a `Shift` node in FOAST.

src/functional/ffront/foast_passes/type_deduction.py
DropD: I suppose it would be less error-prone to do it that way and I will change it.

The reason it wasn't already so is that technically a complete type can not be refined anymore and there is no need to call this method if you have two equal types already.

src/functional/ffront/foast_passes/type_deduction.py
DropD: I was wondering that too when I followed your suggestion to do so.

src/functional/ffront/foast_passes/type_deduction.py
DropD: It even looks to me like the first check is completely unnecessary.

DropD: It isn't but I got everything on one line nonetheless

src/functional/ffront/foast_passes/type_deduction.py
DropD: It did, but I solved the problem by having `.constraint` not reference `.is_any_type`

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -58,6 +59,93 @@ def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
...
+    @property
+    def is_field_type(self) -> bool:
+        return isinstance(self.type, foast.FieldType) or self.constraint is foast.FieldType
+
+    @property
+    def is_scalar(self) -> bool:
+        if self.is_complete:
+            return isinstance(self.type, foast.ScalarType)
+        else:
+            return isinstance(self.constraint, foast.ScalarType)
```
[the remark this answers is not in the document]
DropD: Correct

src/functional/ffront/foast_passes/type_deduction.py
DropD: it is the capture variable (one could also write `kind as dtype_kind`)

src/functional/ffront/foast_passes/type_deduction.py
DropD: see below

src/functional/ffront/foast_passes/type_deduction.py
DropD: In case anyone ever creates a subclass hierarchy with intricate intercompatibilities, it will be up to them to sort out the mess they created here.

src/functional/ffront/foast_passes/type_deduction.py
DropD: agreed.

src/functional/ffront/foast_passes/type_deduction.py
egparedes: Yes, I remember the conversation but either I hadn't realized  that `DeferredSymbolType` was a subclass of `SymbolType` or I was just wrong.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -107,8 +260,177 @@ def visit_FieldSymbol(
...
+            case foast.BinaryOperator.SUB:
+                return "-"
+            case foast.BinaryOperator.MULT:
+                return "*"
+            case foast.BinaryOperator.DIV:
+                return "/"
+            case foast.BinaryOperator.BIT_AND:
+                return "&"
+            case foast.BinaryOperator.BIT_OR:
+                return "|"
```
egparedes, line 341: This looks very much like the definition of the `foast.BinaryOperator.__str__()` function, I'd move it there.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -107,8 +260,177 @@ def visit_FieldSymbol(
...
+            return operand_ti.is_logics_compatible
+
+    def _unaryop_to_str(self, op: foast.UnaryOperator):
+        match op:
+            case foast.UnaryOperator.UADD:
+                return "+"
+            case foast.UnaryOperator.USUB:
+                return "-"
+            case foast.UnaryOperator.NOT:
+                return "not"
```
egparedes, line 414: Missing return type hint. I would additionally move it to `foast.UnaryOperator.__str__`

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -107,8 +260,177 @@ def visit_FieldSymbol(
...
+        if not self._unaryop_type_compatible(op=node.op, operand_type=new_operand.type):
+            raise FieldOperatorTypeDeductionError.from_foast_node(
+                node,
+                msg=f"Incompatible type for unary operator '{self._unaryop_to_str(node.op)}': {new_operand.type}!",
+            )
+        return foast.UnaryOp(
+            op=node.op, operand=new_operand, location=node.location, type=new_operand.type
+        )
+
+    def _unaryop_type_compatible(
```
egparedes, line 398: A predicate-like name seems always clearer to me.
```suggestion
    def _is_unaryop_type_compatible(
```

tests/functional_tests/ffront_tests/test_interface.py
```diff
@@ -246,6 +243,25 @@ def call(inp: Field[..., "float64"]) -> Field[..., "float64"]:
     )
 
 
+def test_temp_tuple():
+    """returning a temp tuple should work."""
```
egparedes, line 247:
```suggestion
    """Returning a temp tuple should work."""
```

tests/functional_tests/ffront_tests/test_type_deduction.py
```diff
@@ -0,0 +1,246 @@
...
+                "is_field_type": True,
+                "is_scalar": False,
+                "is_arithmetic_compatible": False,
+                "is_logics_compatible": True,
+            },
+        ),
+    ]
+
+
+@pytest.mark.parametrize("symbol_type,expected", type_info_cases())
```
egparedes, line 101: :+1:  Please remind other people to follow the same pattern when the number of cases in `parametrize` grows too much. I should've refactor the `symbol_makers` tests in this way (but the reviewer didn't remind me about it :rofl: )

tests/functional_tests/ffront_tests/test_type_deduction.py
```diff
@@ -0,0 +1,246 @@
...
+            ),
+            foast.FieldType(
+                dims=Ellipsis, dtype=foast.ScalarType(kind=foast.ScalarKind.INT64, shape=None)
+            ),
+        ]
+    )
+
+
+def test_adding_bool():
+    """Expect an error (or at least a warnign) when using arithmetic on bools."""
```
egparedes, line 180:
```suggestion
    """Expect an error (or at least a warning) when using arithmetic on bools."""
```

egparedes (commented):
Second round of comments. It looks almost ready to go, only a few comments about style, naming and typos.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -30,36 +32,189 @@ def is_complete_symbol_type(fo_type: foast.SymbolType) -> TypeGuard[foast.Symbol
...
-                    node,
-                    msg=(
-                        "type inconsistency: expression was deduced to be "
-                        f"of type {new}, instead a {type(old)} type was expected."
-                    ),
-                )
+    """
+    Check if an expressions's foast type can be updated to the new type safely.
+
+    Raises a warning if safety can not be guaranteed. This warning will be upgraded to an error in the future.
```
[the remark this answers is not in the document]
DropD: Actually, there is nothing in the way of the transformation anymore.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -30,36 +32,189 @@ def is_complete_symbol_type(fo_type: foast.SymbolType) -> TypeGuard[foast.Symbol
...
+            return self.type == other.type
+        if self.constraint:
+            if other.is_complete:
+                return isinstance(other.type, self.constraint)
+            elif other.constraint:
+                return self.constraint is other.constraint
+        return False
+
+
 def check_type_refinement(node: foast.Expr, new: foast.DataType) -> None:
```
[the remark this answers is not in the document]
DropD: That predicate already exists, this was just for code reuse. However, it was used only in two places anymore.

tests/functional_tests/ffront_tests/test_type_deduction.py
DropD: I only came up with this pattern afterwards. At the time I didn't see a way around it but it started me thinking...

egparedes: :+1:  We used it already in the past (with eve, think) but then kind of forgot about it.

egparedes (approved):
Looks good.

PR #626, opened 2022-01-24
functional: sparse field as iterator of iterator

havogt:


src/functional/iterator/embedded.py
```diff
@@ -382,22 +379,119 @@ def deref(self):
...
+
+    def max_neighbors(self):
+        assert self._max_neighbors is not None
+        return self._max_neighbors
+
+    def is_none(self):
+        return self.index < self._max_neighbors
+
+
+class SparseIterator:
```
havogt, line 405: cleanup code duplication

src/functional/iterator/embedded.py
```diff
@@ -382,21 +407,44 @@ def deref(self):
...
+
+    def deref(self):
+        return self.tup[self.index]
+
+    def max_neighbors(self):
+        assert self._max_neighbors is not None
+        return self._max_neighbors
+
+    def is_none(self):
+        return self.index < self._max_neighbors
```
fthaler, line 428: Shouldn’t this be the other way around?

fthaler (changes requested):

PR #627, opened 2022-01-25
Add concurrency to GitHub Actions

jdahm:
## Description

This should cancel expensive continuous integration when new commits are added to PRs. See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

### DIFF
--- a/.github/workflows/gt4py-gitlab-ci.yml
+++ b/.github/workflows/gt4py-gitlab-ci.yml
@@ -8,7 +8,7 @@ on:
     - staging
 
 concurrency:
-  group: ${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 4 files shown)


jdahm:
Not quite there. We still want the build tests to run on the later commit. Hmm...

jdahm:
bors try

.github/workflows/gt4py-gitlab-ci.yml
```diff
@@ -1,10 +1,16 @@
...
   push:
     branches:
     - master
     - trying
     - staging
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
havogt, line 13: Let's remove it from here. This workflow runs only for a few seconds anyway.

havogt, line 13:
```suggestion
```

havogt (approved):
Let's remove it from the gitlab job.

PR #628, opened 2022-01-25
Add mask_acc to NamedScalars and VectorTemps

jdahm:
## Description

Adds mask accesses in the `gtc:numpy` backend to scalars referenced inside conditional blocks. Not sure why this has not triggered an error yet.

Without the change to the code generation, `gtc:numpy` backend raises an exception at run-time in the changed stencil. The problem is unrelated to lower dimensional storages, but that test was convenient to change.
```
            _mask_0_ = np.full((I - i, J - j, 1), mask_7638408784_gen_0_)
>           outp_[i:I, j:J, k_ : (k_ + 1)][_mask_0_] = tmp_gen_0_
E           TypeError: NumPy boolean array indexing assignment requires a 0 or 1-dimensional input, input has 3 dimensions
```
This happens because `[_mask_0_]` is not added on the RHS access.

### DIFF
--- a/src/gtc/numpy/npir_codegen.py
+++ b/src/gtc/numpy/npir_codegen.py
@@ -228,8 +228,8 @@ class NpirCodegen(TemplatedGenerator):
     def visit_FieldSlice(
         self,
         node: npir.FieldSlice,
-        mask_acc: str = "",
         *,
+        mask_acc: str = "",
         is_serial: bool = False,
         is_rhs: bool = False,
         **kwargs: Any,
@@ -267,9 +267,11 @@ class NpirCodegen(TemplatedGenerator):
 
     EmptyTemp = FormatTemplate("ShimmedView(np.zeros({shape}, dtype={dtype}), {origin})")
 
-    NamedScalar = FormatTemplate("{name}")
+    def visit_NamedScalar(self, node: npir.NamedScalar, *, mask_acc: str = "", **kwargs):
+        return f"{node.name}{mask_acc}"
 
-    VectorTemp = FormatTemplate("{name}_")
+    def visit_VectorTemp(self, node: npir.VectorTemp, *, mask_acc: str = "", **kwargs):
+        return f"{node.name}_{mask_acc}"
 
     def visit_MaskBlock(self, node: npir.MaskBlock, **kwargs) -> Union[str, Collection[str]]:
         if isinstance(node.mask, npir.FieldSlice):
--- a/tests/test_integration/test_code_generation.py
+++ b/tests/test_integration/test_code_generation.py
@@ -247,8 +247,9 @@ def test_lower_dimensional_masked_2dcond(backend):
         outp: gtscript.Field[gtscript.IJK, np.float_],
     ):
         with computation(FORWARD), interval(...):
+            tmp = inp
             if cond > 0.0:
-                outp = inp
+                outp = tmp
 
     inp = np.random.randn(10, 10)
     outp = np.random.randn(10, 10, 10)


jdahm:
bors try

jdahm:
The daint tests passed, there was just a problem propagating that back to GitHub here.

src/gtc/numpy/oir_to_npir.py
```diff
@@ -14,264 +14,207 @@
...
-        return npir.Computation(
-            field_decls=field_decls,
-            field_params=field_names,
-            params=[decl.name for decl in node.params],
-            vertical_passes=vertical_passes,
+    # --- Decls ---
+    def visit_FieldDecl(
+        self, node: oir.FieldDecl, *, extents: StencilExtentComputer.Context, **kwargs: Any
+    ) -> npir.FieldDecl:
+        extent = cast(npir.HorizontalExtent, extents.fields.get(node.name, ((0, 0), (0, 0))))
```
egparedes, line 37: I really think that using `typing.cast()` increases readability of the sources.

src/gtc/numpy/npir.py
```diff
@@ -22,147 +22,164 @@
...
 
-class AxisName(eve.StrEnum):
-    I = "I"  # noqa: E741 (ambiguous variable name)
-    J = "J"
-    K = "K"
+class TemporaryDecl(Decl):
+    # Temporary field shared across HorizontalBlocks
+    data_dims: Tuple[int, ...] = eve.field(default_factory=tuple)
+    offset: Tuple[int, int]
+    boundary: Tuple[int, int]
```
egparedes, line 60: What's the actual meaning of the numbers here?

src/gtc/numpy/npir_codegen.py
```diff
@@ -14,56 +14,62 @@
...
-    delta = str(value)
-    if value == 0:
-        delta = ""
-    elif value < 0:
-        operator = " - "
-        delta = str(abs(value))
-    else:
-        operator = " + "
-    return operator, delta
+def _paren_wrap(s: str) -> str:
```
egparedes, line 30: Just a suggestion: it looks like this function is always used to wrap the expression coming from dumping some interator's content, so you could also add the logic for dumping the iterator here. Something like:

```python
def dump_sequence(sequence, *, separator=", ", start="(", end=")"):
    return f"{start}{separator.join(sequence)}{end}"
```

src/gtc/numpy/npir_codegen.py
```diff
@@ -14,56 +14,62 @@
...
+    return f"{ch}{offset:+d}:{ch.upper()}{offset:+d}" if offset != 0 else f"{ch}:{ch.upper()}"
 
 
 ORIGIN_CORRECTED_VIEW_CLASS = textwrap.dedent(
     """\
-    class ShimmedView:
-        def __init__(self, field, offsets):
-            # use a numpy array here to avoid dimension reducing slicing of storages, which is prohibited and not needed
-            self.field = field.view(np.ndarray)
+    class Field:
```
egparedes, line 40: Using the name `Field` here is a bit confusing for me, specially if it has an attribute `self.field`. I would change it (and maybe also rename `self.field` to `self.field_view`

src/gtc/numpy/npir_codegen.py
```diff
@@ -87,189 +93,105 @@ def __setitem__(self, key, value):
...
-        )
-        origin = ", ".join(
-            [
-                f"_origin_['{node.name}'][{next(origin_idx)}]" if dim else "0"
-                for dim in node.dimensions
-            ]
+        dims = _paren_wrap(", ".join(["True" if d else "False" for d in node.dimensions]))
+        return self.generic_visit(node, dims=dims, **kwargs)
+
+    FieldDecl = FormatTemplate("{name} = Field({name}, _origin_['{name}'], {dims})")
```
egparedes, line 110: Just a suggestion, it should be possible to inline the dims part in the template:

```suggestion
    FieldDecl = FormatTemplate("{name} = Field({name}, _origin_['{name}'], ({', '.join('True' if d else 'False' for d in dimensions)}))")
```

src/gtc/numpy/npir_codegen.py
```diff
@@ -87,189 +93,105 @@ def __setitem__(self, key, value):
...
+                + ["_dK_"]
+            )
         )
-        return self.generic_visit(node, shape=shape, origin=origin, **kwargs)
+        return self.generic_visit(node, shape=shape, **kwargs)
 
-    FieldDecl = FormatTemplate(
-        "{name} = np.reshape({name}.view(np.ndarray), ({shape}))\n_origin_['{name}'] = [{origin}]"
+    TemporaryDecl = FormatTemplate(
+        "{name} = Field.empty({shape}, ({', '.join(str(off) for off in offset)}, 0))"
```
egparedes, line 125: Same here (it should be possible inline the `shape` part)

src/gtc/numpy/npir_codegen.py
```diff
@@ -87,189 +93,105 @@ def __setitem__(self, key, value):
...
-        if node.data_index:
-            offset_str += ", " + ", ".join(self.visit(x, **kwargs) for x in node.data_index)
+    LocalScalarAccess = FormatTemplate("{name}")
 
-        if is_rhs and mask_acc and any(off is None for off in offset):
-            k_size = "1" if is_serial else "K - k"
-            arr_expr = f"np.broadcast_to({node.name}_[{offset_str}], (I - i, J - j, {k_size}))"
-        else:
-            arr_expr = f"{node.name}_[{offset_str}]"
+    ParamAccess = FormatTemplate("{name}")
```
egparedes, line 169: Just for fun, I think this could also work...
```python
LocalScalarAccess = ParamAccess = FormatTemplate("{name}")
```

egparedes (commented):
I have some comments but it looks good to me. Additionally, I think someone else with a deeper knowledge of the gtc toolchain should review the PR.

src/gtc/numpy/npir.py
```diff
@@ -22,147 +22,164 @@
...
+class Decl(eve.Node):
+    name: eve.SymbolName
+    dtype: common.DataType
 
 
-class Cast(common.Cast[Expr], Expr):
+class ScalarDecl(Decl):
+    # Scalar per grid point
+    # Locals never have data_dims
+    # Used for API scalar parameters and oir local scalars
```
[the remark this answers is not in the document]
jdahm: Thanks!

src/gtc/numpy/npir_codegen.py
jdahm: Nice idea!

src/gtc/numpy/npir.py
jdahm: Offset is the origin of the temporary field, and boundary (changed to padding) is the extra padding needed compared to the compute domain.

egparedes (commented):
Almost ready from my side, just a couple of small issues.

jdahm:
bors try

egparedes (approved):
It looks good from my side. It would be great if someone else with a deeper knowledge of GTC would have time to review it today, but it's not a strict requirement so I'm approving the PR. @jdahm can merge in the evening if there are no comments from anyone else.

jdahm:
bors try

PR #629, opened 2022-01-26
functional: Tuple support in embedded mode

havogt:
Add tuple support in embedded iterator view.

Includes fix bad design in the original field view design: stencil closure should have only one output symbol (which can refer to a tuple of fields).

Implemented semantics:
For output fields, fields of tuples or tuples of fields both work for an applied stencil that returns tuple.
For input fields, fields of tuples or tuples of fields are both treated as iterator of tuples. If you need separate iterators, pass the fields as separate arguments, not as tuple.

See ADRs for details.


docs/functional/architecture/0004-Lifted_Stencils_with_Tuple_Return.md
```diff
@@ -0,0 +1,88 @@
...
+
+```python
+def uniform(inp0,inp1,inp2):
+    res = lift(foo)(inp0,inp1,inp2)
+    shifted = shift(O)(res)
+    return deref(shifted)
+
+def non_uniform(inp0,inp1,inp2):
+    res_it = lift(foo)(inp0,inp1,inp2)
+    shifted_vals_O = deref(shift(O)(res_it)) # everything is shifted
```
DropD, line 55: Couldn't this also be written
```python
shifted_vals_0 = shift(0)(lift(lambda x: tuple_get(1, deref(x)))(res_it))
...
```

Which might not even require optimization and for which it is easy to provide a shorthand if needed?

docs/functional/architecture/0004-Lifted_Stencils_with_Tuple_Return.md
```diff
@@ -0,0 +1,88 @@
...
+
+Example
+
+```python
+def stencil():
+    return 42, 43
+
+# What is the type of
+lift(stencil)()
+```
```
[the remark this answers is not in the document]
havogt: I am not super up-to-date how the current lowering is implemented, but I see the following way:

from field view
```python
def fieldop(a: Field[...]):
    return a + 42, a + 43
```
to iterator ir:
```python
def fieldop(a_it):
    return make_tuple(lift(lambda a: deref(a)+42)(a), lift(lambda a: deref(a)+43)(a))
```
Note that this is not a stencil as it does not return value. It's also not a lifted stencil because it returns tuple of iterator. But it is a legal function in iterator IR.

Does it make sense?

docs/functional/architecture/0004-Lifted_Stencils_with_Tuple_Return.md
havogt: Makes sense. I'll add it as a note to the document.

havogt: Actually, that wouldn't change anything, because lift is implemented as applying shift to the arguments of the function. It would be executed as

```python
shifted_vals_0 = (lambda x: tuple_get(1, deref(x)))(shift(0)(res_it)
```

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -45,47 +45,46 @@ def make_domain(dim_name: str, lower: int, upper: int) -> itir.FunCall:
...
-        fencil_definitions=[fencil_from_fop(node, out_names=out_names, domain=domain)],
+        fencil_definitions=[fencil_from_fop(node, out_name=out_name, domain=domain)],
         setqs=[],
     )
 
 
 # todo(tehrengruber): dim and size are implicitly given bys out_names. Get values from there
 def program_from_function(
-    func, out_names: list[str], dim: CartesianAxis, size: int
+    func, dim: CartesianAxis, size: int, out_name: str = "foo"
```
havogt, line 83: out_name is not relevant for the test and is distracting, therefore I added this default.

docs/functional/architecture/0004-Lifted_Stencils_with_Tuple_Return.md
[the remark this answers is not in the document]
DropD: Looks reasonable and also looks like the details of how to use this in the field view wrappers will be easy to work out.

docs/functional/architecture/0004-Lifted_Stencils_with_Tuple_Return.md
DropD: right.

PR #630, opened 2022-01-26
Fixes #220: Add definitive assignment analysis

### LINKED ISSUE #220: Conditional temporary creation with consumption outside
We should warn or error if a temporary is created inside only one branch of a conditional and used outside it. Created from discussion in #193.

tehrengruber:
## Description

If a variable is assigned to before it is used is currently not checked. This PR adds a simple static pass to validate this and warns the user if not (see [#220](https://github.com/GridTools/gt4py/issues/220#issuecomment-988887788) for examples).

__Excerpt from the documentation__
> A symbol is said to be defined if it was assigned to in the code-path of its usage. In other words: If a symbol is defined in both branches of an if-statement it may be used outside. If a symbol is only defined in a single branch it may only be used inside that branch unless it was already defined previously (as this is equal to assigning to the previous value again). Note that whether a symbols is defined is independent of the actual result of the condition.

__Notes on condition evaluation__
In my first implementation I started symbolically evaluating the conditions for all possible values of the mask to also cover cases like these:
```python
def daa_4(in_field: Field[float], mask: bool, out_field: Field[float]):
    with computation(PARALLEL):
        with interval(...):
            if mask:
                tmp = in_field
            if not mask:
                tmp = in_field + 1
            out_field = tmp
```
I have seen such cases (usually more complicated) occasionally occur in numerical code and given that the parallel model defines the semantics of an if-else statement exactly that way this appeared like the way to go. However I abandoned this approach at some point for several reasons:
- Many special cases like the following will not work and it is hard both to document and to convey to the user what triggers a warning and what not:
```python
def daa_4(in_field: Field[float], mask: Field[bool], out_field: Field[float]):
    with computation(PARALLEL):
        with interval(...):
            if mask:
                tmp = in_field
            mask = True
            if not mask:
                tmp = in_field + 1
            out_field = tmp
```
- If evaluating the condition is required to understand whether a variable is defined it should be best practice to refactor the code anyway.
- It makes sense to do these checks in a way that also applies to the new model. However there stencils can be composed increasing the amount of involved variables. Consequently the runtime would explode or if one limits the amount of variables the analysis would become useless.

Nonetheless such analysis might still be useful for other things in the future (e.g. generating specialized code for boundary conditions). Here are some insights and links in case I/someone revisits this again:
- As a rule of thumb until 5 variables it is faster to just try all conditions (german: https://de.wikipedia.org/wiki/Verfahren_nach_Quine_und_McCluskey#Anmerkung)
- https://en.wikipedia.org/wiki/Quine%E2%80%93McCluskey_algorithm
- https://second.wiki/wiki/resolventenmethode

## Requirements

### DIFF
--- a/src/gt4py/stencil_builder.py
+++ b/src/gt4py/stencil_builder.py
@@ -18,6 +18,7 @@ import pathlib
 from typing import TYPE_CHECKING, Any, Dict, Optional, Type, Union
 
 import gt4py.caching
+import gt4py.frontend
 from gt4py.backend.gtc_backend.defir_to_gtir import DefIRToGTIR
 from gt4py.definitions import BuildOptions, StencilID
 from gt4py.type_hints import AnnotatedStencilFunc, StencilFunc
--- a/src/gtc/passes/gtir_definitive_assignment_analysis.py
+++ b/src/gtc/passes/gtir_definitive_assignment_analysis.py
@@ -42,9 +42,9 @@ class DefinitiveAssignmentAnalysis(NodeVisitor):
             invalid_accesses.append(node)
 
     @classmethod
-    def apply(cls, gtir_stencil_expr: gtir.Stencil) -> Set[gtir.FieldAccess]:
+    def apply(cls, gtir_stencil_expr: gtir.Stencil) -> List[gtir.FieldAccess]:
         """Execute analysis and return all accesses to undefined symbols."""
-        invalid_accesses: Set[gtir.FieldAccess] = set()
+        invalid_accesses: List[gtir.FieldAccess] = []
         DefinitiveAssignmentAnalysis().visit(
             gtir_stencil_expr,
             alive_vars=set(gtir_stencil_expr.param_names),
--- a/src/gtc/passes/gtir_definitive_assignment_analysis.py
+++ b/src/gtc/passes/gtir_definitive_assignment_analysis.py
@@ -24,7 +24,9 @@ class DefinitiveAssignmentAnalysis(NodeVisitor):
         self.visit(node.false_branch, alive_vars=false_branch_vars, **kwargs)
         alive_vars.update(true_branch_vars & false_branch_vars)
 
-    def visit_ParAssignStmt(self, node: gtir.ParAssignStmt, *, alive_vars: Set[str], **kwargs) -> None:
+    def visit_ParAssignStmt(
+        self, node: gtir.ParAssignStmt, *, alive_vars: Set[str], **kwargs
+    ) -> None:
         self.visit(node.right, alive_vars=alive_vars, **kwargs)
         alive_vars.add(node.left.name)
 
--- a/src/gtc/passes/gtir_definitive_assignment_analysis.py
+++ b/src/gtc/passes/gtir_definitive_assignment_analysis.py
@@ -17,14 +17,14 @@ class DefinitiveAssignmentAnalysis(NodeVisitor):
     result of the condition.
     """
 
-    def visit_IfStmt(self, node: gtir.FieldIfStmt, *, alive_vars: Set[str], **kwargs):
+    def visit_IfStmt(self, node: gtir.FieldIfStmt, *, alive_vars: Set[str], **kwargs) -> None:
         true_branch_vars = {*alive_vars}
         false_branch_vars = {*alive_vars}
         self.visit(node.true_branch, alive_vars=true_branch_vars, **kwargs)
         self.visit(node.false_branch, alive_vars=false_branch_vars, **kwargs)
         alive_vars.update(true_branch_vars & false_branch_vars)
 
-    def visit_ParAssignStmt(self, node: gtir.ParAssignStmt, *, alive_vars: Set[str], **kwargs):
+    def visit_ParAssignStmt(self, node: gtir.ParAssignStmt, *, alive_vars: Set[str], **kwargs) -> None:
         self.visit(node.right, alive_vars=alive_vars, **kwargs)
         alive_vars.add(node.left.name)
 
@@ -35,14 +35,14 @@ class DefinitiveAssignmentAnalysis(NodeVisitor):
         alive_vars: Set[str],
         invalid_accesses: List[gtir.FieldAccess],
         **kwargs,
-    ):
+    ) -> None:
         if node.name not in alive_vars:
             invalid_accesses.append(node)
 
     @classmethod
-    def apply(cls, gtir_stencil_expr: gtir.Stencil):
+    def apply(cls, gtir_stencil_expr: gtir.Stencil) -> Set[gtir.FieldAccess]:
         """Execute analysis and return all accesses to undefined symbols."""
-        invalid_accesses: List[gtir.FieldAccess] = []
+        invalid_accesses: Set[gtir.FieldAccess] = set()
         DefinitiveAssignmentAnalysis().visit(
             gtir_stencil_expr,
             alive_vars=set(gtir_stencil_expr.param_names),
@@ -54,7 +54,7 @@ class DefinitiveAssignmentAnalysis(NodeVisitor):
 analyze = DefinitiveAssignmentAnalysis.apply
 
 
-def check(gtir_stencil_expr: gtir.Stencil):
+def check(gtir_stencil_expr: gtir.Stencil) -> gtir.Stencil:
     """Execute definitive assignment analysis and warn on errors."""
     invalid_accesses = analyze(gtir_stencil_expr)
     for invalid_access in invalid_accesses:
--- a/src/gtc/passes/gtir_definitive_assignment_analysis.py
+++ b/src/gtc/passes/gtir_definitive_assignment_analysis.py
@@ -16,6 +16,7 @@ class DefinitiveAssignmentAnalysis(NodeVisitor):
     previous value again). Note that whether a symbols is defined is independent of the actual
     result of the condition.
     """
+
     def visit_IfStmt(self, node: gtir.FieldIfStmt, *, alive_vars: Set[str], **kwargs):
         true_branch_vars = {*alive_vars}
         false_branch_vars = {*alive_vars}
@@ -31,7 +32,7 @@ class DefinitiveAssignmentAnalysis(NodeVisitor):
         self,
         node: gtir.FieldAccess,
         *,
-        alive_vars: [str],
+        alive_vars: Set[str],
         invalid_accesses: List[gtir.FieldAccess],
         **kwargs,
     ):
@@ -54,9 +55,7 @@ analyze = DefinitiveAssignmentAnalysis.apply
 
 
 def check(gtir_stencil_expr: gtir.Stencil):
-    """
-    Execute definitive assignment analysis and warn on errors.
-    """
+    """Execute definitive assignment analysis and warn on errors."""
     invalid_accesses = analyze(gtir_stencil_expr)
     for invalid_access in invalid_accesses:
         warnings.warn(f"`{invalid_access.name}` may be uninitialized.")
--- a/src/gtc/passes/gtir_definitive_assignment_analysis.py
+++ b/src/gtc/passes/gtir_definitive_assignment_analysis.py
@@ -1,0 +1,63 @@
+import copy
+import warnings
+from typing import List, Set
+
+from eve import NodeVisitor
+from gtc import gtir
+
+
+class DefinitiveAssignmentAnalysis(NodeVisitor):
+    def visit_IfStmt(self, node: gtir.FieldIfStmt, *, alive_vars: Set[str], **kwargs):
+        true_branch_vars = copy.copy(alive_vars)
+        false_branch_vars = copy.copy(alive_vars)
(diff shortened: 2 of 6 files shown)


tehrengruber:
bors try

src/gt4py/stencil_builder.py
```diff
@@ -18,6 +18,7 @@
 from typing import TYPE_CHECKING, Any, Dict, Optional, Type, Union
 
 import gt4py.caching
+import gt4py.frontend
```
egparedes, line 21: Good catch! I know it's outside of the scope of this PR, but if everything was working before, is this really needed? And if it is needed, why is not tested?

src/gtc/passes/gtir_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+def analyse(gtir_stencil_expr: gtir.Stencil):
+    """
+    Analyse gtir stencil expression for access to undefined symbols.
+
+    A symbol is said to be defined if it was assigned to in the code-path of its usage. In other words: If a symbol is
+    defined in both branches of an if-statement it may be used outside. If a symbol is only defined in a single branch
+    it may only be used inside that branch unless it was already defined previously (as this is equal to assigning to
+    the previous value again). Note that whether a symbols is defined is independent of the actual result of the
+    condition.
+    """
```
egparedes, line 42: Please shorten the length of the docstrings (around 80/90 chars)

src/gt4py/stencil_builder.py
egparedes, line 21: Good catch! Out of the scope of this PR, but I wonder if it is actually needed since the error hasn't been found earlier, and if it is indeed needed, why is not tested...

tests/test_unittest/test_gtc/test_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,129 @@
...
+        global test_data
+        test_data.append((definition, {"valid": valid}))
+        return definition
+
+    return _wrapper
+
+
+# test cases
+@register_test_case(valid=False)
+def daa_0(in_field: Field[float], mask: Field[bool], out_field: Field[float]):
```
egparedes, line 31: It would help to have a more meaningful name for the test cases.

tests/test_unittest/test_gtc/test_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,129 @@
...
+from gt4py.stencil_builder import StencilBuilder
+from gtc.passes import gtir_definitive_assignment_analysis as daa
+
+
+class TestData(TypedDict):
+    valid: bool
+
+
+# A list of dictionaries containing a stencil definition and the expected test case outputs
+test_data: List[Tuple[Callable, TestData]] = []
```
egparedes, line 17: Why not just this?
```suggestion
test_data: List[Tuple[Callable, bool]] = []
```

tests/test_unittest/test_gtc/test_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,129 @@
...
+import pytest
+
+from gt4py.backend import from_name
+from gt4py.gtscript import PARALLEL, Field, computation, interval, stencil
+from gt4py.stencil_builder import StencilBuilder
+from gtc.passes import gtir_definitive_assignment_analysis as daa
+
+
+class TestData(TypedDict):
+    valid: bool
```
egparedes, line 13: This looks redundant to me.

tests/test_unittest/test_gtc/test_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,129 @@
...
+def daa_7(in_field: Field[float], mask: Field[bool], out_field: Field[float]):
+    with computation(PARALLEL):
+        with interval(...):
+            if in_field > 0:
+                tmp = in_field
+            else:
+                tmp = in_field + 1
+            out_field = tmp
+
+
```
egparedes, line 113: I miss some test case using (nested) `elif`s

src/gtc/passes/gtir_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+        *,
+        alive_vars: Set,
+        invalid_accesses: List[gtir.FieldAccess],
+        **kwargs,
+    ):
+        if node.name not in alive_vars:
+            invalid_accesses.append(node)
+
+
+def analyse(gtir_stencil_expr: gtir.Stencil):
```
egparedes, line 33: It's not really documented but the (implicit) convention is to declare an `apply()` class method in the visitor as the public interface (creating an instance and executing it with the provided data). If a free function API is also desired, it's just an alias to apply, like:

```python
analyse = DefinitiveAssignmentAnalysis.apply
```
On a different topic, I think we use US spelling everywhere in the code, and in that case `analyze` fits better.

src/gtc/passes/gtir_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def visit_ParAssignStmt(self, node: gtir.ParAssignStmt, *, alive_vars: Set[str], **kwargs):
+        self.visit(node.right, alive_vars=alive_vars, **kwargs)
+        alive_vars.add(node.left.name)
+
+    def visit_FieldAccess(
+        self,
+        node: gtir.FieldAccess,
+        *,
+        alive_vars: Set,
```
egparedes, line 25:
```suggestion
        alive_vars: Set[str],
```

src/gtc/passes/gtir_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+from eve import NodeVisitor
+from gtc import gtir
+
+
+class DefinitiveAssignmentAnalysis(NodeVisitor):
+    def visit_IfStmt(self, node: gtir.FieldIfStmt, *, alive_vars: Set[str], **kwargs):
+        true_branch_vars = copy.copy(alive_vars)
+        false_branch_vars = copy.copy(alive_vars)
+        self.visit(node.true_branch, alive_vars=true_branch_vars, **kwargs)
+        self.visit(node.false_branch, alive_vars=false_branch_vars, **kwargs)
```
egparedes, line 14: We can just create new set instances without using the `copy` module
```suggestion
        self.visit(node.true_branch, alive_vars={*alive_vars}, **kwargs)
        self.visit(node.false_branch, alive_vars={*alive_vars}, **kwargs)
```

egparedes (commented):
It looks good. Just some minor comments

jdahm:
I noticed that `TestConstantFolding` matches this pattern. Does the gtir pass catch that one?

tests/test_unittest/test_gtc/test_definitive_assignment_analysis.py
tehrengruber: The names become way too long when meaningful. I've added docstrings instead.

tests/test_unittest/test_gtc/test_definitive_assignment_analysis.py
tehrengruber: Done.

src/gt4py/stencil_builder.py
tehrengruber: I have to admit I do not understand why this did not surface elsewhere/before. The circular import surfaced in `conftest.py` which should be run everytime tests are executed. Maybe another import avoiding this happens before occasionally? Anyway running `python ./tests/conftest.py` also fails on the current master.

tehrengruber: Yes I think this is needed. Check the error message above for the circular import error.

tehrengruber:
@jdahm Good catch, yes this is covered (see the first test [here](https://github.com/GridTools/gt4py/pull/630/files#diff-4facc2f958b2065aa19bb0675c4be0a6cd0b4128e4bca2f10d4ff4f14ba44e74R26)). I removed `TestConstantFolding`.
@egparedes Should be fine for another review round now.

src/gt4py/stencil_builder.py
egparedes: I don't think it is ever expected to run `conftest.py` as a script, it's mostly just a container of pytest settings.

src/gtc/passes/gtir_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+        alive_vars.add(node.left.name)
+
+    def visit_FieldAccess(
+        self,
+        node: gtir.FieldAccess,
+        *,
+        alive_vars: Set[str],
+        invalid_accesses: List[gtir.FieldAccess],
+        **kwargs,
+    ):
```
egparedes, line 38:
```suggestion
    ) -> None:
```

src/gtc/passes/gtir_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+
+    def visit_IfStmt(self, node: gtir.FieldIfStmt, *, alive_vars: Set[str], **kwargs):
+        true_branch_vars = {*alive_vars}
+        false_branch_vars = {*alive_vars}
+        self.visit(node.true_branch, alive_vars=true_branch_vars, **kwargs)
+        self.visit(node.false_branch, alive_vars=false_branch_vars, **kwargs)
+        alive_vars.update(true_branch_vars & false_branch_vars)
+
+    def visit_ParAssignStmt(self, node: gtir.ParAssignStmt, *, alive_vars: Set[str], **kwargs):
```
egparedes, line 27:
```suggestion
    def visit_ParAssignStmt(self, node: gtir.ParAssignStmt, *, alive_vars: Set[str], **kwargs) -> None:
```

src/gtc/passes/gtir_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+
+    A symbol is said to be defined if it was assigned to in the code-path of its usage. In
+    other words: If a symbol is defined in both branches of an if-statement it may be used
+    outside. If a symbol is only defined in a single branch it may only be used inside that
+    branch unless it was already defined previously (as this is equal to assigning to the
+    previous value again). Note that whether a symbols is defined is independent of the actual
+    result of the condition.
+    """
+
+    def visit_IfStmt(self, node: gtir.FieldIfStmt, *, alive_vars: Set[str], **kwargs):
```
egparedes, line 20:
```suggestion
    def visit_IfStmt(self, node: gtir.FieldIfStmt, *, alive_vars: Set[str], **kwargs) -> None:
```

src/gtc/passes/gtir_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+        *,
+        alive_vars: Set[str],
+        invalid_accesses: List[gtir.FieldAccess],
+        **kwargs,
+    ):
+        if node.name not in alive_vars:
+            invalid_accesses.append(node)
+
+    @classmethod
+    def apply(cls, gtir_stencil_expr: gtir.Stencil):
```
egparedes, line 43:
```suggestion
    def apply(cls, gtir_stencil_expr: gtir.Stencil) -> Set[gtir.FieldAccess]:
```

src/gtc/passes/gtir_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+        """Execute analysis and return all accesses to undefined symbols."""
+        invalid_accesses: List[gtir.FieldAccess] = []
```
egparedes, line 45:
```suggestion
        invalid_accesses: Set[gtir.FieldAccess] = set()
```

src/gtc/passes/gtir_definitive_assignment_analysis.py
```diff
@@ -0,0 +1,63 @@
...
+            alive_vars=set(gtir_stencil_expr.param_names),
+            invalid_accesses=invalid_accesses,
+        )
+        return invalid_accesses
+
+
+analyze = DefinitiveAssignmentAnalysis.apply
+
+
+def check(gtir_stencil_expr: gtir.Stencil):
```
egparedes, line 57:
```suggestion
def check(gtir_stencil_expr: gtir.Stencil) -> gtir.Stencil:
```

tehrengruber:
bors try

egparedes (commented):
Just some missing type hints. Check the suggestions

src/gt4py/stencil_builder.py
tehrengruber: I see. For me the error occured when executing the tests with PyCharm

tehrengruber:
bors try

tests/test_unittest/test_gtc/test_definitive_assignment_analysis.py
egparedes: Agree

egparedes (approved):
It looks ready to go.

tehrengruber:
bors try

PR #635, opened 2022-01-29
Do not k-cache fields referenced in masks

jdahm:
## Description

Skips k-caching fields under two additional criteria:
1. The field is not referenced in a mask, and
2. Any offsets match the loop order.

This is a leftover effort from Eddie, which I wanted to PR before we forget about it.

@fthaler Could you take a look and assess if this makes sense to add? Thanks!

To do:
- [ ] Determine case when this is necessary
- [ ] Add a test


fthaler:
Thanks for the PR, but actually I don’t understand why those changes should be required or desired.

I think that the loop order check would actually disable k-caches in many loops (for example within the COSMO dycore) where k-caching is actually very helpful.

I neither understand the idea behind the masked writes. There are for sure cases where k-caching _can_ slow things down if there are masked writes (for example if a mask is used to enable writes only on a small subset of vertical levels). But generally disabling k-caches on masked writes is probably not the right strategy around this, as this would probably also disable k-caches in undesired cases. Maybe Eddie also found some cases where the current strategy produces wrong results? But in this case it would be helpful to have test cases to actually check that we only disable the caches when required.

jdahm:
> Maybe Eddie also found some cases where the current strategy produces wrong results? But in this case it would be helpful to have test cases to actually check that we only disable the caches when required.

Yes, precisely. I was hoping we could create tests cases from the failures to incrementally improve what exists in the code. On your assessment, I'll close this PR for now but save the branch. We'll see what the status is again when we get a new version working with the dycore.

PR #636, opened 2022-02-02
Functional: Unify all Symbol nodes in foast

tehrengruber:
## Description

The FOAST currently has various nodes subclassing `Symbol`, i.e. `DataSymbol`, `FieldSymbol`, `TupleSymbol`, `Function`, `OffsetSymbol`. As such the decision on what type a symbol has needs to be done when creating the symbol, i.e. at the syntax level before the type inference/deduction is run. Let's look at how this works __currently__:

- Assign: `Symbol` can be either a `Field` or a `Tuple`. By default everything is a `Field` unless there is a `TupleExpr` on the rhs. This is problematic as soon as one tries to add more types, e.g. scalars, as we don't know what the type of the right hand side is until we run type inference. As an example consider `some_scalar = 1+some_external` or a function call returning a scalar or tuple.
- Field operator parameters: Everything is a field. If we want scalar parameters in the future, the Python -> FOAST parsing needs to understand the annotation and emit a different symbol. By our guiding principles we want lean visitors and the type of a symbol already has all information.
- External from value: Since the value is known so is its type -> works

This PR removes all of the various types of `Symbol` nodes and uses only `Symbol` itself with the respective type information. This solves the aforementioned problems, reduces the complexity of the FOAST and avoids duplicating type information when expressed both in the node type and the `type` attribute of a Symbol. It comes at the additional cost of needing validators when only certain types are allowed (see changes in `FieldOperator` node).

Side note: This PR originated from my PRGAST work where I recognized that I need to duplicate all of the different types of Symbols and generalize `make_symbol_from_value`. After this PR I only need to duplicate one node and generalizing`make_symbol_from_value` is straightforward.

### DIFF
--- a/src/functional/ffront/field_operator_ast.py
+++ b/src/functional/ffront/field_operator_ast.py
@@ -171,4 +171,4 @@ class FieldOperator(LocatedNode, SymbolTableTrait):
     id: SymbolName  # noqa: A003
     params: list[DataSymbol]
     body: list[Stmt]
-    closure: list[DataSymbol]
+    closure: list[Symbol]
--- a/src/functional/ffront/field_operator_ast.py
+++ b/src/functional/ffront/field_operator_ast.py
@@ -14,7 +14,7 @@
 
 
 import re
-from typing import Optional, Union
+from typing import Generic, Optional, TypeVar, Union
 
 from pydantic import validator
 
@@ -39,9 +39,12 @@ class SymbolName(eve.traits.SymbolName):
     regex = re.compile(r"^[a-zA-Z_][\w$]*$")
 
 
-class Symbol(LocatedNode):
+SymbolT = TypeVar("SymbolT", bound=common_types.SymbolType)
+
+
+class Symbol(eve.GenericNode, LocatedNode, Generic[SymbolT]):
     id: SymbolName  # noqa: A003
-    type: common_types.SymbolType  # noqa A003
+    type: Union[SymbolT, common_types.DeferredSymbolType]  # noqa A003
     namespace: Namespace = Namespace(Namespace.LOCAL)
 
 
@@ -158,14 +161,6 @@ class Return(Stmt):
 
 class FieldOperator(LocatedNode, SymbolTableTrait):
     id: SymbolName  # noqa: A003
-    params: list[Symbol]
+    params: list[Symbol[common_types.DataType]]
     body: list[Stmt]
     closure: list[Symbol]
-
-    @validator("params")
-    def validate_params_type(cls, params: list[Symbol]):
-        assert all(
-            isinstance(param.type, (common_types.DeferredSymbolType, common_types.DataType))
-            for param in params
-        )
-        return params
--- a/src/functional/ffront/field_operator_ast.py
+++ b/src/functional/ffront/field_operator_ast.py
@@ -16,8 +16,6 @@
 import re
 from typing import Generic, Optional, TypeVar, Union
 
-from pydantic import validator
-
 import eve
 from eve import Node
 from eve.traits import SymbolTableTrait
--- a/src/functional/ffront/field_operator_ast.py
+++ b/src/functional/ffront/field_operator_ast.py
@@ -155,7 +155,7 @@ class Stmt(LocatedNode):
 
 
 class ExternalImport(Stmt):
-    symbols: list[DataSymbol]
+    symbols: list[Symbol]
 
 
 class Assign(Stmt):
--- a/src/functional/ffront/field_operator_ast.py
+++ b/src/functional/ffront/field_operator_ast.py
@@ -55,6 +55,8 @@ FieldSymbol = Symbol[FieldTypeT]
 ScalarTypeT = TypeVar("ScalarTypeT", bound=common_types.ScalarType)
 ScalarSymbol = Symbol[ScalarTypeT]
 
+TupleTypeT = TypeVar("TupleTypeT", bound=common_types.TupleType)
+TupleSymbol = Symbol[TupleTypeT]
 
 class Expr(LocatedNode):
     type: Optional[common_types.SymbolType] = None  # noqa A003
@@ -159,7 +161,7 @@ class ExternalImport(Stmt):
 
 
 class Assign(Stmt):
-    target: FieldSymbol
+    target: Union[FieldSymbol, TupleSymbol]
     value: Expr
 
 
--- a/src/functional/ffront/field_operator_ast.py
+++ b/src/functional/ffront/field_operator_ast.py
@@ -46,6 +46,16 @@ class Symbol(eve.GenericNode, LocatedNode, Generic[SymbolT]):
     namespace: Namespace = Namespace(Namespace.LOCAL)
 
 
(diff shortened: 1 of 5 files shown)


src/functional/ffront/field_operator_ast.py
```diff
@@ -176,6 +158,14 @@ class Return(Stmt):
...
     body: list[Stmt]
     closure: list[Symbol]
+
+    @validator("params")
+    def validate_params_type(cls, params: list[Symbol]):
+        assert all(
+            isinstance(param.type, (common_types.DeferredSymbolType, common_types.DataType))
+            for param in params
+        )
+        return params
```
egparedes, line 171: Could this custom validator be replaced by a generic definition of the `Symbo` node (as `Symbol[T]`)?

src/functional/ffront/symbol_makers.py
```diff
@@ -168,22 +168,16 @@ def make_symbol_type_from_typing(
...
-    elif isinstance(symbol_type, common_types.FunctionType):
-        return foast.Function(
-            id=name,
-            type=symbol_type,
-            namespace=namespace,
-            location=location,
-        )
-    elif isinstance(symbol_type, common_types.OffsetType):
-        return foast.OffsetSymbol(id=name, type=symbol_type, namespace=namespace, location=location)
+    """Make a symbol node from a python value."""
```
egparedes, line 171:
```suggestion
    """Make a symbol node from a Python value."""
```

src/functional/ffront/symbol_makers.py
```diff
@@ -168,22 +168,16 @@ def make_symbol_type_from_typing(
...
+    assert not isinstance(value, type) and type(value).__module__ != "typing"
```
egparedes, line 172: If passing types-like args is forbidden, I think it'd better to raise a `ValueError` exception with a meaningful message.

egparedes (commented):
Looks good in general, but it could be better to try to to use a generic `Symbol` class as discussed offline.

src/functional/ffront/field_operator_ast.py
```diff
@@ -176,6 +159,6 @@ class Return(Stmt):
 
 class FieldOperator(LocatedNode, SymbolTableTrait):
     id: SymbolName  # noqa: A003
-    params: list[DataSymbol]
+    params: list[Symbol[common_types.DataType]]
```
egparedes, line 162: This works but it is probably more correct to use a `DataSymbolT` `TypeVar` bound to `common_types.DataType`

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -224,12 +224,12 @@ def visit_Assign(self, node: foast.Assign, **kwargs) -> foast.Assign:
...
 
-    def visit_FieldSymbol(
+    def visit_Symbol(
         self,
-        node: foast.FieldSymbol,
+        node: foast.Symbol,
         refine_type: Optional[common_types.FieldType] = None,
         **kwargs,
-    ) -> foast.FieldSymbol:
+    ) -> foast.Symbol:
```
egparedes, line 232: I guess for static type checkers it would be better to define two overloads here, one when `refine_type` is `None` which returns a parameterized `foast.Symbol[T]` using the same `TypeVar` than the argument, and another with non-optional `refine_type` `TypeVar("RefinedT", common_types.FieldType, common_types.TupleType)` value which returns a parameterized `foast.Symbol[RefinedT]`.

I'm not sure mypy will be able to deal with all the Nodes/pydantic magic anyway, so maybe we can skip it for now, but at least I want to document the fact.

src/functional/ffront/func_to_foast.py
```diff
@@ -362,14 +362,14 @@ def visit_ImportFrom(self, node: ast.ImportFrom, **kwargs) -> None:
 
         return foast.ExternalImport(symbols=symbols, location=self._make_loc(node))
 
-    def visit_arguments(self, node: ast.arguments) -> list[foast.FieldSymbol]:
+    def visit_arguments(self, node: ast.arguments) -> list[foast.Symbol]:
```
egparedes, line 365: Shouldn't return a list of concrete `foast.Symbol` for the correct types (DataType ?) instead of a generic one?

src/functional/ffront/func_to_foast.py
```diff
@@ -362,14 +362,14 @@ def visit_ImportFrom(self, node: ast.ImportFrom, **kwargs) -> None:
...
         return [self.visit_arg(arg) for arg in node.args]
 
-    def visit_arg(self, node: ast.arg) -> foast.FieldSymbol:
+    def visit_arg(self, node: ast.arg) -> foast.Symbol:
```
egparedes, line 368: Same comment here. Also, shouldn't it be checked that `new_type` is a `DataType`?

src/functional/ffront/symbol_makers.py
```diff
@@ -168,22 +168,18 @@ def make_symbol_type_from_typing(
...
+        # we don't have types of types so disallow this
+        raise ValueError("The type of a symbol can not be a type itself.")
+    type_ = typingx.get_typing(value, annotate_callable_kwargs=True)
+
+    symbol_type = make_symbol_type_from_typing(type_)
+
+    if isinstance(
+        symbol_type, (common_types.DataType, common_types.FunctionType, common_types.OffsetType)
+    ):
+        return foast.Symbol(id=name, type=symbol_type, namespace=namespace, location=location)
```
egparedes, line 182: A different concrete specialization of Symbol should be returned in each one of these cases.

egparedes (commented):
Again, it looks good, but if the `GenericNode` approach is followed, then it should be used completely, instantiating the right concrete subclass whenever is possible instead of always using the generic version.

src/functional/ffront/field_operator_ast.py
tehrengruber: You are absolutely right (would be great if this would throw an error).

src/functional/ffront/func_to_foast.py
tehrengruber: If `new_type` is not a DataType pydantic will already throw an error. If added an additional check that gives a more readable message.

src/functional/ffront/field_operator_ast.py
tehrengruber: Fixed. I've changed `FieldOperator.{params, closure}` and `ExternalImport.symbols` to contain `DataSymbols` only.

src/functional/ffront/foast_passes/type_deduction.py
tehrengruber: I agree. Skipping for now.

src/functional/ffront/field_operator_ast.py
tehrengruber: Fixed. We also discussed offline to change closure and external import to only allow `DataType`, but I recognized this is not sufficient as we also need `OffsetType`.

tehrengruber:
@egparedes ready for re-review

egparedes (approved):
It looks good to me.

PR #640, opened 2022-02-04
Functional: Add Reduction to FieldView

mroethlin:
## Description

This PR adds a built in function `nbh_sum` to the FieldView. This includes
* lowering of `nbh_sum` to the appropriate `reduce` primitive in the iterator IR
* syntax checking of `nbh_sum`
* type checking of the shifts within the `nbh_sum`

~~**NOTE**: this currently breaks 3 structured tests~~

## Requirements

- [ ] All the authors are covered by a valid contributor assignment agreement,
- [ ] The names of all the new contributors have been added to an updated


src/functional/ffront/fbuiltins.py
```diff
@@ -15,13 +15,22 @@
...
+from functional.ffront import common_types as ct
 
 
-__all__ = ["Field", "float32", "float64", "int32", "int64"]
+__all__ = ["Field", "float32", "float64", "int32", "int64", "nbh_sum"]
 
 TYPE_BUILTINS = [Field, float, float32, float64, int, int32, int64, bool, tuple]
 TYPE_BUILTIN_NAMES = [t.__name__ for t in TYPE_BUILTINS]
 
+nbh_sum = ct.FunctionType(
```
egparedes, line 26: `nbh_sum` is a bit too cryptic. I'd suggest to change it. Additionally, I now think it would be better to define the function as an actual (empty) function with the `__gt_type_` attribute set to the type definition. This change would be needed anyway for embedded execution and it feels like a cleaner design.

src/functional/ffront/itir_makers.py
```diff
@@ -73,6 +73,12 @@ def ensure_expr(literal_or_expr: Union[str, int, itir.Expr]) -> itir.Expr:
     return literal_or_expr
 
 
+def ensure_offset(str_or_offset: Union[str, itir.OffsetLiteral]) -> itir.OffsetLiteral:
+    if isinstance(str_or_offset, str):
```
egparedes, line 77: Missing docstring

src/functional/ffront/symbol_makers.py
```diff
@@ -147,31 +147,35 @@ def make_symbol_type_from_typing(
...
-        raise ValueError("The type of a symbol can not be a type itself.")
-    # TODO(tehrengruber): What we expect here currently is a GTCallable. Maybe
-    #  we should check for the protocol in the future?
-    if hasattr(value, "__gt_type__"):
-        symbol_type = value.__gt_type__()
+def make_symbol_type_from_value(value: Any) -> ct.SymbolType:
+    """Make a symbol node from a Python value."""
+    if isinstance(value, ct.SymbolType):
+        # Temporary hack for type deductions with shifted fields
+        symbol_type = value
```
egparedes, line 164: I think this first workaround is only required for the `neighbor_sum` function so if the function is implemented as in my previous suggestion, this can be removed because the `__gt_type__` will cover the case.

src/functional/ffront/symbol_makers.py
```diff
@@ -147,31 +147,35 @@ def make_symbol_type_from_typing(
...
+    elif isinstance(value, runtime.Offset):
+        # Temporary hack for type deductions with shifted fields
+        symbol_type = ct.OffsetType(source=value.source, target=value.target)
```
egparedes, line 167: This other workaround can also be handled using the  `__gt_type__` attribute. Most likely it is only needed to add this line to the `runtime.Offset` class:
```python
class Offset:
  ...

    def __gt_type__(self):
        return ct.OffsetType(source=self.source, target=self.target)
```
I would also add a TODO comment mentioning that a cleaner solution will be to use the encode the `OffsetType` as an actual type also valid for the iterator prototype

src/functional/iterator/runtime.py
```diff
@@ -12,10 +12,12 @@
 @dataclass(frozen=True)
 class Offset:
     value: Optional[Union[int, str]] = None
+    source: Optional[common.Dimension] = None
+    target: Optional[Tuple[common.Dimension, common.Dimension]] = None
```
egparedes, line 16: Is there any reason why this needs to be a pair or can it be a unrestricted tuple?
```suggestion
    target: Optional[Tuple[common.Dimension, ...] = None
```

egparedes (commented):
I suggest a couple of changes to improve some of the workarounds required for this PR.

src/functional/iterator/runtime.py
DropD: Works just as well with an unrestricted tuple.

DropD: Works just as well with an unrestricted tuple.

src/functional/ffront/symbol_makers.py
DropD: Actually, it would probably be better to eventually lift both `runtime.Offset` and `common_types` into `functional`.

src/functional/ffront/itir_makers.py
```diff
@@ -73,6 +73,23 @@ def ensure_expr(literal_or_expr: Union[str, int, itir.Expr]) -> itir.Expr:
     return literal_or_expr
 
 
+def ensure_offset(str_or_offset: Union[str, itir.OffsetLiteral]) -> itir.OffsetLiteral:
+    """
+    Convert python literals into an OffsetLiteral and let OffsetLiterals pass unchanged.
```
egparedes, line 78:
```suggestion
    Convert Python literals into an OffsetLiteral and let OffsetLiterals pass unchanged.
```

src/functional/ffront/fbuiltins.py
```diff
@@ -12,19 +12,65 @@
...
+    source: Optional[Dimension] = None
+    target: Optional[tuple[Dimension, ...]] = None
+
+    def __gt_type__(self):
+        return ct.OffsetType(source=self.source, target=self.target)
+
+
+# TODO(ricoh): same applies as for FVOffset above
+def fvoffset(value, source=None, target=None):
+    return FVOffset(value, source, target)
```
egparedes, line 76: Is the function really needed? I think `runtime` implements the same pattern, but I think it is basically useless to define a function which basically calls the constructor with the same arguments.

src/functional/ffront/fbuiltins.py
```diff
@@ -12,19 +12,65 @@
...
 ALL_BUILTIN_NAMES = TYPE_BUILTIN_NAMES + MODULE_BUILTIN_NAMES
 
 BUILTINS = {name: globals()[name] for name in __all__}
+
+
+# TODO(ricoh): This should probably be reunified with ``iterator.runtime.Offset``
+# potentially in ``functional.common``, which requires lifting of
+# ``ffront.common_types`` into ``functional``.
+@dataclass(frozen=True)
+class FVOffset(runtime.Offset):
```
egparedes, line 66: I understand this class is a temporary solution and fully agree with your comment about how to improve it in the future. However, I still don't like the name of the class. What about `OffsetLiteral` or `FieldOffset` or something more readable which doesn't clash with `runtime.Offset`?

egparedes (commented):
Basically ready to go. Just a couple of questions and suggestions

egparedes (approved):
It looks good. It is only needed to fix the unrelated CI errors by updating `black` version.

PR #644, opened 2022-02-04
Use pathlib to determine location of module

jdahm:
## Description

Allows for relocatable caches.

Thanks to @mcgibbon for the contribution!

Blocked by: #647.

### DIFF
--- a/src/gt4py/backend/module_generator.py
+++ b/src/gt4py/backend/module_generator.py
@@ -466,14 +466,14 @@ class PyExtModuleGenerator(BaseModuleGenerator):
         source = ["from gt4py import utils as gt_utils"]
         if self._is_not_empty():
             assert self.pyext_file_path is not None
-            file_path = "f{{pathlib.Path(__file__).parent.resolve()}}/{}".format(
+            file_path = 'f"{{pathlib.Path(__file__).parent.resolve()}}/{}"'.format(
                 os.path.basename(self.pyext_file_path)
             )
             source.append(
                 textwrap.dedent(
                     f"""
                 pyext_module = gt_utils.make_module_from_file(
-                    "{self.pyext_module_name}", "{file_path}", public_import=True
+                    "{self.pyext_module_name}", {file_path}, public_import=True
                 )
                 """
                 )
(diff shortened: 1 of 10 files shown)


src/gt4py/backend/module_generator.py
```diff
@@ -465,11 +465,15 @@ def _is_not_empty(self) -> bool:
     def generate_imports(self) -> str:
         source = ["from gt4py import utils as gt_utils"]
         if self._is_not_empty():
+            assert self.pyext_file_path is not None
+            file_path = "f{{pathlib.Path(__file__).parent.resolve()}}/{}".format(
+                os.path.basename(self.pyext_file_path)
+            )
```
egparedes, line 471: It looks like there is an error here. Should this be something like:
```suggestion
            file_path = "{{pathlib.Path(__file__).parent.resolve()}}/{base}".format(
                base=os.path.basename(self.pyext_file_path)
            )
```
?

egparedes (commented):
The idea looks good, but the implementation contains some kind of bug.

src/gt4py/backend/module_generator.py
jdahm: The `f` has to be pulled outside the quotes. I think I got it now. The latest commit I pushed works locally.

egparedes:
borg try

egparedes:
bors try

jdahm:
bors try

src/gt4py/backend/module_generator.py
```diff
@@ -465,11 +465,15 @@ def _is_not_empty(self) -> bool:
     def generate_imports(self) -> str:
         source = ["from gt4py import utils as gt_utils"]
         if self._is_not_empty():
+            assert self.pyext_file_path is not None
+            file_path = 'f"{{pathlib.Path(__file__).parent.resolve()}}/{}"'.format(
+                os.path.basename(self.pyext_file_path)
+            )
```
egparedes, line 471: Mixing escaped and non-escaped placeholders is a bit confusing. I propose to replace the `format()`  call by a simple string concatenation:
```suggestion
            file_path = 'f"{{pathlib.Path(__file__).parent.resolve()}}/"' + str(
                os.path.basename(self.pyext_file_path)
            )
```

jdahm:
This build passed, but it failed to propagate the success back to github. It just needs your re-review to merge.

PR #651, opened 2022-02-07
Always return FieldInfo or ParameterInfo for GTC backends

jdahm:
## Description

Now that the legacy backends are disabled, we can correct the odd behavior of leaving values as `None` in the `field_info` and `parameter_info` of the `StencilObject`.

I added a TODO to update the type hint once we actually remove the IIR.

PR #653, opened 2022-02-08
Remove test_make_args_data_from_iir and add checks to gtir version

### LINKED ISSUE #652: Test make_args_data_from_gtir independent from iir
Testing this in isolation will add robustness and remove another another key dependency on the legacy toolchain.

jdahm:
## Description

Removes another dependency on the legacy toolchain by testing the module generator args_data from gtir directly without comparing it to the iir version.

Blocked by #651.
Resolves #652.

### DIFF
--- a/src/gt4py/backend/module_generator.py
+++ b/src/gt4py/backend/module_generator.py
@@ -15,6 +15,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import abc
+import itertools
 import numbers
 import os
 import textwrap
@@ -47,13 +48,13 @@ class ModuleData:
 
     @property
     def field_names(self) -> Set[str]:
-        """Set of field names that are referenced."""
-        return {name for name in self.field_info if name not in self.unreferenced}
+        """Set of all field names."""
+        return set(self.field_info.keys())
 
     @property
     def parameter_names(self) -> Set[str]:
-        """Set of parameter names that are referenced."""
-        return {name for name in self.parameter_info if name not in self.unreferenced}
+        """Set of all parameter names."""
+        return set(self.parameter_info.keys())
 
 
 def make_args_data_from_iir(implementation_ir: gt_ir.StencilImplementation) -> ModuleData:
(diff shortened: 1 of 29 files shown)


jdahm:
bors try

jdahm:
bors try

src/gt4py/backend/module_generator.py
```diff
@@ -78,24 +81,23 @@ def make_args_data_from_iir(implementation_ir: gt_ir.StencilImplementation) -> M
...
-            if arg.name not in implementation_ir.unreferenced:
-                data.parameter_info[arg.name] = ParameterInfo(
-                    dtype=implementation_ir.parameters[arg.name].data_type.dtype
-                )
-            else:
-                data.parameter_info[arg.name] = None
+            data.parameter_info[arg.name] = ParameterInfo(
+                access=(
+                    AccessKind.NONE
+                    if arg.name in implementation_ir.unreferenced
```
tehrengruber, line 96: Since you are not using `implementation_ir.unreferenced` in the above branch anymore I would suggest to also remove the dependency on it here:
```suggestion
                    if arg.name not in in_fields and arg.name not in out_fields
```

src/gt4py/backend/module_generator.py
```diff
@@ -166,15 +172,23 @@ def make_args_data_from_gtir(pipeline: GtirPipeline, legacy=False) -> ModuleData
     ]
     for name in sorted(referenced_scalar_params):
         data.parameter_info[name] = ParameterInfo(
-            dtype=numpy.dtype(node.symtable_[name].dtype.name.lower())
+            access=AccessKind.READ, dtype=numpy.dtype(node.symtable_[name].dtype.name.lower())
         )
 
     unref_params = get_unused_params_from_gtir(pipeline)
     for param in sorted(unref_params, key=lambda decl: decl.name):
```
tehrengruber, line 179: It's somewhat strange that there are two seperate loops for the referenced and unreferenced params. If you are touching this part anyway you could combine the two and only use `get_unused_params_from_gtir`, but it's optional (as unrelated to the actual PR).

src/gt4py/definitions.py
```diff
@@ -700,8 +700,15 @@ def ndim(self):
 
 @dataclass(frozen=True)
 class ParameterInfo:
+    # NOTE: Parameters are read-only, so the possible values are AccessType.NONE and AccessType.READ.
+    access: AccessKind
```
tehrengruber, line 704:
```suggestion
    access: Union[Literal[AccessKind.None], Literal[AccessKind.Read]]
```

tehrengruber (approved):
Looks good. I assume the next step is to remove `make_args_data_from_iir`?

jdahm:
> Looks good. I assume the next step is to remove make_args_data_from_iir?

Yes. Thank you, these are nice suggestions on the PR, and gave me some ideas. I haven't pushed the code yet, so let's hold off on merging until I can make the changes.

jdahm:
@tehrengruber this is ready for a final review. The `AccessKindComputer` is a big improvement on the existing analysis, because it catches the case where a read happens after a write, but it's not perfect because it's not taking into account intervals and accumulations in the vertical. I'll put something together to extend it to handle those cases correctly in a future PR. For now I think it can stay as-is.

jdahm:
bors try

PR #654, opened 2022-02-09
Ff lowering

DropD:
Use let-expressions to lower from field view to iterator view.
 
Also:
* Introduce an actually readable way to create Iterator IR snippets (in `mockitir.py`).
* Separate parsing tests from lowering tests
* Add an ADR explaining the lowering algorithm and why it was chosen


src/functional/ffront/mockitir.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+def tuple_get(tuple_expr, index):
+    """Create a tuple_get FunCall, shorthand for ``call("tuple_get")(tuple_expr, index)``."""
+    return call("tuple_get")(tuple_expr, index)
+
+
+def lift(expr):
+    """Create a lift FunCall, shorthand for ``call("lift")(expr)``."""
+    return call(call("lift")(expr))
```
egparedes, line 177: The docstring and the code don't match.

src/functional/ffront/mockitir.py
```diff
@@ -0,0 +1,215 @@
...
+# version. See the LICENSE.txt file at the top-level directory of this
+# distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from typing import Union
+
+from functional.iterator import ir as itir
+
+
```
egparedes, line 19: I think the functions in this module could be added to the `foast_to_itir` module, since they are basically makers to reduce boilerplate code in the creation of iterator IR. If you think they should go in a different module because it is easy to share them for different purposes (like testing), I think the current name is confusing and would suggest something more explicit like `itir_makers`.

Additionally, to make clear that they are makers for itir nodes, I think it'd be worth to use a naming convention like `make_<itir_node>()` or `_<itir_node>()` or `<itir_node>_()` and get rid of the one-line makers for trivial calls to builtin functions, since they are not really needed and in the lowering code it would look like that there are nodes in the iterator IR with those names.

tests/functional_tests/ffront_tests/test_lowering.py
```diff
@@ -0,0 +1,382 @@
...
+IDim = CartesianAxis("IDim")
+
+
+def debug_itir(tree):
+    import black
+    from devtools import debug
+
+    from functional.iterator.backends.roundtrip import EmbeddedDSL
+
+    debug(black.format_str(EmbeddedDSL.apply(tree), mode=black.Mode()))
```
egparedes, line 33: Just a question here: are you aware of `eve.codegen.format_source`? Do you think it is better to directly use `black` for tests?

egparedes (commented):
It looks good. I added some comments and questions, mostly about style.

src/functional/ffront/mockitir.py
DropD: I can agree to adding one or maximum two characters to the makers to distinguish them from iterator builtins.
Currently this is achieved by how they are imported (`mi.<itir_node>`).

I can also agree to renaming the module they live in. I do think however that they merit their own module because they are useful beyond the lowering (and even testing).

The following example illustrates why I can not support dropping the builtins shortcuts.

```python
snippet = lambda_("a", "b", "c", ... )(
    plus_(deref_("a"), plus_(deref_("b"), plus_(...)))
)(...)

# vs

snippet = call_("lambda")(
    ["a", "b", "c", ...], 
    call_("plus")(call_("deref")("a"), call_("plus")(call_("deref")("b"), call_("plus")(...))))
)(...)
```

The former allows you to express (**precisely** I might argue) that this snippet creates the IR for a fundef that looks exactly the same, except that variable names wouldn't be strings.
There is a reason why all design notes and discussion notes put field view code next to iterator view fundefs **and not** iterator IR.

The latter is on one hand a more accurate description of the IR nodes, but on the other hand it is also simply a more verbose and harder to read version of the former (do the parentheses actually match? - who can tell without reformatting it or using an IDE?)

Since this seems to become obvious only after trying different ways for 3-4 days maybe it requires an ADR?

tests/functional_tests/ffront_tests/test_lowering.py
DropD: I had simply forgotten about it.

src/functional/ffront/mockitir.py
```diff
@@ -0,0 +1,215 @@
...
+    return literal_or_expr
+
+
+class lambda_:
+    """
+    Create a lambda from params and an expression.
+
+    Examples
+    --------
+    >>> lambda_("a")(deref("a"))  # doctest: +ELLIPSIS
```
tehrengruber, line 82: Not sure I'm a fan of this staggered creation. I don't really see the need to do that for a lambda.

docs/functional/architecture/0002-Field_View_Lowering.md
```diff
@@ -0,0 +1,118 @@
...
+    )
+  )
+```
+
+This was not a major goal but a happy accident, since it is expected that a common subexpression elimination optimization will be run on the Iterator IR anyway.
+
+#### Subexpression Lifting
+
+A consequence of this algorithm is that all field view expressions must be lowered to iterator expressions. This may lead to some not strictly necessary
+lifting and dereferencing but is in line with the intuition that in field view every expression is a field expression (even scalar literals, which are not implemented yet at the time of writing).
```
tehrengruber, line 98: Line break as above

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -126,10 +36,10 @@ def are_broadcast_compatible(left: TypeInfo, right: TypeInfo) -> bool:
 
     Examples:
     ---------
-    >>> int_scalar_t = TypeInfo(common_types.ScalarType(kind=common_types.ScalarKind.INT64))
+    >>> int_scalar_t = TypeInfo(ScalarType(kind=ScalarKind.INT64))
```
tehrengruber, line 39: For modules containing a set of similar objects I always prefix with modules. If you change this here we should at least do it consistently everywhere in a separate PR.

src/functional/ffront/foast_to_itir.py
```diff
@@ -160,93 +59,93 @@ def visit_FieldOperator(self, node: foast.FieldOperator, **kwargs) -> itir.Funct
...
-        params: Optional[list[itir.Sym]] = None,
-        **kwargs,
-    ) -> itir.Expr:
-        return self.visit(AssignResolver.apply(exprs), **kwargs)
+        self, body: list[foast.Stmt], params: Optional[list[itir.Sym]] = None, **kwargs
+    ) -> itir.FunCall:
+        *assigns, return_stmt = body
+        current_expr = self.visit(return_stmt, **kwargs)
+
+        for assign in assigns[-1::-1]:
```
tehrengruber, line 67: Suggestion for readability:
```suggestion
        for assign in reverse(assigns):
```

src/functional/ffront/foast_to_itir.py
```diff
@@ -160,93 +59,93 @@ def visit_FieldOperator(self, node: foast.FieldOperator, **kwargs) -> itir.Funct
...
+    def visit_Name(self, node: foast.Name, **kwargs) -> itir.SymRef:
+        return mi.ref(node.id)
 
     def visit_Subscript(self, node: foast.Subscript, **kwargs) -> itir.FunCall:
-        return ItirFunCallFactory(
-            name="tuple_get",
-            args=[self.visit(node.value, **kwargs), itir.IntLiteral(value=node.index)],
-        )
+        param_names = self._make_lambda_param_names(node, **kwargs)
+        return mi.lift(
```
tehrengruber, line 97: The lifted-lambda-call is such a common pattern in this lowering I would create a helper for it (including the param name extraction).

src/functional/ffront/foast_to_itir.py
```diff
@@ -160,93 +59,93 @@ def visit_FieldOperator(self, node: foast.FieldOperator, **kwargs) -> itir.Funct
...
+            mi.lambda_(*param_names)(
+                mi.tuple_get(mi.deref(self.visit(node.value, **kwargs)), node.index)
```
tehrengruber, line 99: I suggest adding a check that node.type is a TupleType.

src/functional/ffront/foast_to_itir.py
```diff
@@ -160,93 +59,93 @@ def visit_FieldOperator(self, node: foast.FieldOperator, **kwargs) -> itir.Funct
...
 
-    def _gen_shift_args(
-        self, args: list[foast.Subscript], **kwargs
-    ) -> Iterator[Union[itir.OffsetLiteral, itir.IntLiteral]]:
-        for arg in args:
-            name, offset = self._make_shift_args(arg)
-            yield name
-            yield offset
+    def visit_shift(self, node: foast.Call, **kwargs) -> itir.FunCall:
+        return mi.shift(node.args[0].value.id, node.args[0].index)(self.visit(node.func, **kwargs))
```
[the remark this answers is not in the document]
DropD: Wrong place, syntax checking is only wanted in python -> FOAST. We can create an issue for it, but basically it's out of scope for this PR.

src/functional/ffront/foast_to_itir.py
DropD: Will change to `reversed(assigns)`

src/functional/ffront/foast_to_itir.py
DropD: The check happens in type deduction (which is the proper place for it).

src/functional/ffront/mockitir.py
egparedes: Ok, I understand your reasons for keeping the one-liners and it's fine for me. It would be enough with adding just a single line/paragraph to the existing ADR mentioning that makers exist for itir symbols to simplify the understanding of the itir expressions (specially in tests, which I think it's the only place where heavily nested code can appear).

src/functional/ffront/foast_passes/type_deduction.py
DropD: Is it acceptable to follow the convention, but for example `import common_types as ct` to improve readability?

src/functional/ffront/mockitir.py
DropD: I am convinced, if you spend some days building complex enough snippets, you will appreciate it as much as I do.

src/functional/ffront/foast_to_itir.py
```diff
@@ -160,93 +59,93 @@ def visit_FieldOperator(self, node: foast.FieldOperator, **kwargs) -> itir.Funct
...
+        param_names = self._make_lambda_param_names(node, **kwargs)
+        return mi.lift(
+            mi.lambda_(*param_names)(
+                mi.tuple_get(mi.deref(self.visit(node.value, **kwargs)), node.index)
+            )
+        )(*param_names)
 
     def visit_TupleExpr(self, node: foast.TupleExpr, **kwargs) -> itir.FunCall:
-        return ItirFunCallFactory(name="make_tuple", args=self.visit(node.elts, **kwargs))
+        args = [mi.deref(arg) for arg in self.visit(node.elts, **kwargs)]
```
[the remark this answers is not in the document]
DropD: This means there is currently no way of returning two fields from a field operator by our chosen path of creating a stencil and then lifting it, since

```python
return field1, field2
```

becomes
```python
return deref(make_tuple(iter1, iter2))
```

Which is invalid.

src/functional/ffront/foast_passes/type_deduction.py
tehrengruber: Yes this makes sense, I think we should use `ct` for `common_types` everywhere then.

src/functional/ffront/mockitir.py
DropD: I don't mind changing it if you still think it's better after considering the following:

They are both not an exact match.
```python
(lambda a, b, c: expr)(1, 2, 3)
call(lambda_(a, b, c)(expr))(1, 2, 3)
call(lambda_([a, b, c], expr))(1, 2, 3)
```
However the former captures that python syntax is not `(lambda [a, b, c], expr)(1, 2, 3)` for good reason. The `:` helps visually.

What it gives you when you actually use it is

```python
call(
    lambda(*vars)(
        (
            super_long_expression
            ...
        )
    )
)(*values)
```
as opposed to
```python
call(
    lambda(
        vars, 
        (
            super_long_expression
            ...
        )
    )
)(*values)
```

Which may not seem like much in such a simple example, but note how the `vars` line up with the values they are given above, and how they don't below. This slight advantage is multiplied by the complexity of the snippet.

egparedes (commented):
Second round of review with minor comments.

src/functional/ffront/ast_passes/single_static_assign.py
```diff
@@ -63,55 +62,60 @@ class RhsRenamer(ast.NodeTransformer):
...
-    @dataclass
-    class State:
-        name_counter: dict[str, int] = field(default_factory=dict)
-
     @classmethod
     def apply(cls, node: ast.AST) -> ast.AST:
         return cls().visit(node)
 
-    def __init__(self):
+    def __init__(self, separator="__"):
```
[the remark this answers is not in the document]
DropD: The renamed variables now end up in the let expressions so they have to be valid python names.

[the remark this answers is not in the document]
DropD:
```python
def foo(...):
    tmp = a
    tmp = tmp+1
    ...
```
becomes
```
let tmp__0 = a in (
    let tmp__1 = tmp__0 + 1 in (
         ... 
    )
)

File: src/functional/ffront/mockitir.py (hunk 4 above)
Comment (egparedes) [reply src/functional/ffront/mockitir.py]: Just to clarify, I'm fine with both the original proposal and Till's proposal, so for me this is not a blocker. Iif you don't manage to agree here, I'd then suggest to go with the original proposal since it's already implemented and it was explicitly designed that way.

File: src/functional/ffront/ast_passes/single_static_assign.py (hunk 12 above)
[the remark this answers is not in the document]
Comment (egparedes) [reply src/functional/ffront/ast_passes/single_static_assign.py:83]: Ok, I see. In principle this is fine for me too, but this change makes possible again to have name collisions with user-declared variables. The right way to do it would be to check for possible collisions before renaming the symbols, but it feels a bit overkill here. Another possibility to reduce the collision chance would be to use a very uncommon unicode character as separator (see complete list here: https://www.asmeurer.com/python-unicode-variable-names/start-characters.html) like runnic letters (e.g. tmpᛄ1) although it also has some inconveniences. Anyway, the only real action point I have here is to add a comment or a point in the ARDs to mention this issue.

File: tests/functional_tests/ffront_tests/test_lowering.py [hunk 13]
```diff
@@ -0,0 +1,367 @@
...
+from functional.ffront.fbuiltins import float64, int64
+from functional.ffront.foast_to_itir import FieldOperatorLowering
+from functional.ffront.func_to_foast import FieldOperatorParser
+from functional.iterator.runtime import CartesianAxis, offset
+
+
+IDim = CartesianAxis("IDim")
+
+
+def debug_itir(tree):
```
egparedes, line 27: Just a suggestion: this function looks like it's only useful for debugging purposes if something goes wrong, so maybe a comment can be added here to explicitly state it.

egparedes (approved):
It looks good to me. I left a couple of non-essential suggestions, but they are fully optional.

src/functional/ffront/ast_passes/single_static_assign.py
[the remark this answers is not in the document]
DropD: will do

tests/functional_tests/ffront_tests/test_lowering.py
DropD: good idea

PR #656, opened 2022-02-09
functional: Refactor utilities from python function to foast into seperate module

tehrengruber:
Some utilities in the python to foast parsing can be reused in the python to past parsing. This PR refactors them into a seperate module `py_lowering_utils` (suggestions for a better naming schema are welcome). 

Question to the reviewer: The utilities are used in some other places, e.g. tests, doc tests. Is an indirect import from the `func_to_foast` module fine or should one directly import from `py_lowering_utils`?


DropD:
I would prefer direct imports in the tests because they double as examples.

egparedes (approved):
It looks good.

PR #657, opened 2022-02-10
Add pass to re-use temporaries

### LINKED ISSUE #648: Re-use temporaries in OIR when possible
It would be convenient to have an optimization pass on OIR that reduces the number of temporary fields, as we have seen a number of kernels with very high shared memory usage, leading to either (1) not being able to run the kernel or (2) in the best cases very low occupancy.

Reducing the number of temporaries should be possible by using a [live variable analysis](https://en.wikipedia.org/wiki/Live_variable_analysis).

jdahm:
## Description

Adds a pass that runs over the OIR and tries to remap temporaries to reduce the set of fields allocated as temporaries in order to reduce memory pressure. Scalar temporaries are not considered here as they do not contribute to shared memory pressure on GPU backends.

Resolves #648.


fthaler:
I did a first pass through the code and have some questions.

First, your tests all work within a single vertical loop and a single horizontal execution. These situations should be already addressed by existing optimization passes, shouldn’t they? That is, the local temporaries should be replaced by scalars and no allocation should take place at all. So I wonder in which cases you actually want to achieve improvements? I would have guessed that in the of single-loop stencils we don’t do too badly as we already try to inline stuff, replace temporaries by scalars, and introduce caches if possible.

For the case of larger multi-loop stencils, I can broadly see two strategies for reducing the allocations. The first one is by looking at the vertical loops as black boxes and only check if a temporary is accessed in any way. The second one involves checking the orders of reads and writes within vertical loops. The latter is extremely tricky to do accurately in the current IR. For example, it’s pretty clear that just traversing the vertical loop sections in order without checking offsets is not enough to do any accurate decisions as soon as we have more than one section within a loop. So I wonder if the first one might already help in your cases?

For example, in the following case:
|       | VL0 | VL1 | VL 2 | VL3 |
|------|-------|------|-----|--|
|tmp0| access|access | | | 
|tmp1|access | | |access |
|tmp2| | | access|access | 

we could rename tmp2 to tmp0, leading to:

|       | VL0 | VL1 | VL 2 | VL3 |
|------|-------|------|-----|--|
|tmp0| access|access | access| access | 
|tmp1|access | | |access |

(this problem can be solved optimally using an interval graph and greedy coloring, which is a quite simple algorithm).

src/gtc/passes/oir_optimizations/temporaries.py
```diff
@@ -150,3 +151,141 @@ def write_before_read(tmp: str) -> bool:
...
+                unused_allocated.remove(lval_temp)
+                in_use_allocated.add(lval_temp)
+                symbol_to_temp[lval_symbol] = lval_temp
+                lhs.name = lval_temp
+            else:
+                in_use_allocated.add(lval_symbol)
+                symbol_to_temp[lval_symbol] = lval_symbol
+                # lhs.name is up to date
+        else:
+            pass
```
fthaler, line 253: ?

jdahm: Leftover from converting pseudo-code into Python. Will remove later.

jdahm:
> First, your tests all work within a single vertical loop and a single horizontal execution. These situations should be already addressed by existing optimization passes, shouldn’t they?

Yes, the idea is to replace true temporaries across horizontal executions. You're right that if those tests as written ran through other optimizations that those temporaries would be converted to scalars and the pass itself would do nothing. Definitely before settling on a final algorithm we would want to write better and more complicated tests. This was just a proof of concept to try out one algorithm.

> For example, it’s pretty clear that just traversing the vertical loop sections in order without checking offsets is not enough to do any accurate decisions as soon as we have more than one section within a loop.

Yes, in this version of the algorithm we've taken the safest approach and ignored vertical loop sections entirely, so a read on any part of the data is a "field read". It could be much more aggressive if done carefully.

> For example, in the following case:

I believe the algorithm I presented here actually works in the this case. If `tmp0` is not accessed after `VL1` then it is released back to the `unused_allocated` and therefore `tmp2` will be renamed to `tmp0`, and in the final step the declaration will be removed.

havogt:
Closing for now, as we don't have resources to continue this work.

PR #666, opened 2022-02-13
functional: refactor common types utils into separate module

tehrengruber:
The utilities in the FOAST type deduction are independent of the FOAST and only require the symbol types in `ffront.common_types`. This PR refactors them into a separate module so that I can use them in the `PAST` type deduction.

Note that the mypy version that we are currently using does not support pythons `match` statement used in the code. As such mypy fails and according to https://github.com/python/mypy/issues/11829 there doesn't seem to be a way to suppress the warning. Support is expected to land in the next release 0.940 and is already on master: https://github.com/python/mypy/pull/10191


egparedes:
Since these functions fully depend on the `common_types` module, I think it would make more sense to just add them there instead of creating a new separate module.

egparedes (commented):
Since these functions fully depend on the `common_types` module, I think it would make more sense to just add them there instead of creating a new separate module.

egparedes:
> Shall we just ignore the mypy errors? They will eventually disappear anyway.

Maybe add the files to the ignore list in the pre-commit settings for mypy. When the next mypy release comes out in a couple of weeks, we can remove all files from the ignore list.

egparedes:
You need to rebase/merge the new `functional` branch here and it is ready to go.

tehrengruber:
I tried just to recognize that we apparently both missed that @DropD  did almost the same change in https://github.com/GridTools/gt4py/pull/654. Well the utils landed in a new file [`type_info.py`](https://github.com/GridTools/gt4py/blob/functional/src/functional/ffront/type_info.py). We decided to put it in `common_types` to avoid a new module, but let's leave it that way for now. Otherwise merging my PAST PR becomes even more complex.

PR #368, opened 2021-03-02
Rewrite DemoteLocalTemporariesToVariablesPass

jdahm:
## Description

Cleans up the logic in the `DemoteLocalTemporariesToVariablesPass`.

PR #667, opened 2022-02-15
Eve: Add basic pattern matching capabilities

tehrengruber:
While testing the PAST nodes I became tired of always writing out the entire nodes and the resulting tests were often testing more then necessary. This PR introduces basic pattern matching facilities that allow comparing two values and showing the differences between the two. Note that no capturing is supported, but this could be added at a later point.

Example:
```python
import foo_ir
foo_ir_ = ModuleWrapper(foo_ir)
assert foo_ir_.Foo(bar="baz").matches(foo_ir.Foo(bar="baz", foo="bar"))
assert not foo_ir_.Foo(bar="bar").matches(foo_ir.Foo(bar="baz", foo="bar"))
```


src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+    def matches(self, other: Any, raise_: bool = False) -> bool:
+        """Return if object pattern matches `other` using :func:`get_differences`.
+
+        If `raise_` is specified raises an exception with all differences found.
+        """
+        diffs = list(get_differences(self, other))
+        if raise_ and len(diffs) != 0:
+            diffs_str = "\n  ".join([f"  {self.cls.__name__}{path}: {msg}" for path, msg in diffs])
+            raise ValueError(f"Object and pattern don't match:\n  {diffs_str}")
+        return len(diffs) == 0
```
egparedes, line 34: It doesn't make sense to iterate the generators with all the errors if they haven't been requested by the user. We could just return `False` as soon as the first error is found:

```suggestion
        if not raise_exception:
            return next(get_differences(self, other), None) is None
        else:
            diffs = [*get_differences(self, other)]
            if len(diffs) == 0:
                return True
            else:
                diffs_str = "\n  ".join([f"  {self.cls.__name__}{path}: {msg}" for path, msg in diffs])
                raise ValueError(f"Object and pattern don't match:\n  {diffs_str}")
```

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def get_differences(a: Any, b: Any, path: str = "") -> Iterator[tuple[str, str]]:
+    """Compare two objects and return a list of differences.
+
+    If the arguments are lists or dictionaries comparison is recursively per item. Objects are compared
+    using equality operator or if the left-hand-side is an `ObjectPattern` its type and attributes
+    are compared to the right-hand-side object. Only the attributes of the `ObjectPattern` are used
+    for comparison, disregarding potential additional attributes of the right-hand-side.
+    """
```
egparedes, line 107: Please, make lines shorter

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+    >>> assert not foo_ir_.Foo(bar="bar").matches(foo_ir.Foo(bar="baz", foo="bar"))  # doctest: +SKIP
+    """
+
+    module: ModuleType
+
+    def __getattr__(self, item: str) -> ObjectPatternConstructor:
+        val = getattr(self.module, item)
+        if not inspect.isclass(val):
+            raise ValueError("Only classes allowed.")
+        return ObjectPatternConstructor(val)
```
egparedes, line 139: I find this interface confusing and error-prone. I would rather use a more explicit pattern creator syntax, for example using __class_getitem__ to get something like `Pattern[SomeClass](foo="foo", bar="bar")` or `Matcher[SomeClass](foo="foo", bar="bar")`

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+            path,
+            f"Expected an instance of class {a.cls.__name__}, but got {type(b).__name__}",
+        )
+    else:
+        for k in a.attrs.keys():
+            if not hasattr(b, k):
+                yield (path, f"Value has no attribute {k}.")
+            else:
+                for diff in get_differences(a.attrs[k], getattr(b, k), path=f"{path}.{k}"):
+                    yield diff
```
egparedes, line 63: The pythonic way would be:
```suggestion
                yield from get_differences(a.attrs[k], getattr(b, k), path=f"{path}.{k}")
```

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        diffs = list(get_differences(self, other))
+        if raise_ and len(diffs) != 0:
+            diffs_str = "\n  ".join([f"  {self.cls.__name__}{path}: {msg}" for path, msg in diffs])
+            raise ValueError(f"Object and pattern don't match:\n  {diffs_str}")
+        return len(diffs) == 0
+
+    def __str__(self) -> str:
+        attrs_str = ", ".join([f"{str(k)}={str(v)}" for k, v in self.attrs.items()])
+        return f"{self.cls.__name__}({attrs_str})"
```
egparedes, line 38:
```suggestion
        fields_str = ", ".join([f"{str(k)}={str(v)}" for k, v in self.fields.items()])
        return f"{self.cls.__name__}({fields_str})"
```

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+
+def _get_differences_list(a: list, b: Any, path: str = "") -> Iterator[tuple[str, str]]:
+    if not isinstance(b, list):
+        yield (path, f"Expected list, but got {type(b).__name__}")
+    elif len(a) != len(b):
+        yield (path, f"Expected list of length {len(a)}, but got length {len(b)}")
+    else:
+        for i, (el_a, el_b) in enumerate(zip(a, b)):
+            for diff in get_differences(el_a, el_b, path=f"{path}[{i}]"):
+                yield diff
```
egparedes, line 74:
```suggestion
        for i, (el_a, el_b) in enumerate(zip(a, b)):
            yield from get_differences(el_a, el_b, path=f"{path}[{i}]")
```
It might be even possible to make it a one-liner using `itertools.chain.from_iterable()` but I think it would harder to read.

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+import inspect
+from dataclasses import dataclass
+from types import ModuleType
+from typing import Any, Iterator
+
+
+@dataclass(frozen=True)
+class ObjectPattern:
```
egparedes, line 21: Missing docstring.

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+            yield (
+                path,
+                f"Expected dictionary with keys `{'`, `'.join(map(str, a.keys()))}`, but the following keys are extra: {extra_keys_str}",
+            )
+    else:
+        for k, v_a, v_b in zip(a.keys(), a.values(), b.values()):
+            yield from get_differences(v_a, v_b, path=f'{path}["{k}"]')
+
+
+def get_differences(a: Any, b: Any, path: str = "") -> Iterator[tuple[str, str]]:
```
egparedes, line 100: I would protect the `path` argument since it's an internal argument not meant to be used in the public interface of the function.

```suggestion
def get_differences(a: Any, b: Any, *, __path: str = "") -> Iterator[tuple[str, str]]:
```

tests/eve_tests/unit_tests/test_pattern_matching.py
```diff
@@ -0,0 +1,81 @@
...
+    a: Any
+    b: Any
+    expected_diff: list[tuple[str, str]]
+
+
+test_data: list[tuple[str, TestData]] = []
+
+
+def register_test_case(name, a, b, expected_diff):
+    test_data.append((name, {"a": a, "b": b, "expected_diff": expected_diff}))
```
egparedes, line 21: This is more convoluted than needed and not idiomatic (the TypedDict definition is also wrong: should not contain `name`). The simple way to implement the same workflow using regular `pytest` style would be something like:

```
# Each sample contains: (case_name, a, b, expected_diff)
test_data: list[tuple[str, Any, Any, list[tuple[str, str]]]] = [
    ("int_equal", 1, 1, [])
]

@pytest.mark.parametrize("name,a, b, expected_diff", test_data)
def test_all(name,a, b, expected_diff):
    diff = list(get_differences(a, b, path="a"))
    assert diff == expected_diff
```

Additionally, I miss some test cases dealing with nested data dict/lists/objects

egparedes (changes requested):
I'm not fully convinced of the scalability/composability of using `get_differences()` as main dispatcher, but it could be good enough for now. I left some suggestions to enhance the current implementation.

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+        attrs_str = ", ".join([f"{str(k)}={str(v)}" for k, v in self.attrs.items()])
+        return f"{self.cls.__name__}({attrs_str})"
+
+
+@dataclass(frozen=True)
+class ObjectPatternConstructor:
+    cls: type
+
+    def __call__(self, **kwargs: Any) -> ObjectPattern:
+        return ObjectPattern(self.cls, kwargs)
```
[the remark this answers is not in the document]
tehrengruber: I deliberately created this class instead of using `functools.partial` to have a good error message and to have access to `cls` during debugging when one accidentally tries to use the `ObjectPatternConstructor` instead of the `ObjectPattern` itself.

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+import inspect
+from dataclasses import dataclass
+from types import ModuleType
+from typing import Any, Iterator
+
+
+@dataclass(frozen=True)
+class ObjectPattern:
+    cls: type
+    attrs: dict[str, Any]
```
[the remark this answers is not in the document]
tehrengruber: I'm not too much into using `fields` as what is stored here applies to general objects for which the common name is attribute (correct me if I'm wrong here). How about `attributes`?

src/eve/pattern_matching.py
tehrengruber: True, I've implemented a slightly different, but similar version to what you proposed.

src/eve/pattern_matching.py
tehrengruber: See comment at the top.

src/eve/pattern_matching.py
tehrengruber: It is explicitly meant as a public argument. This is actually used in one of the tests:
```
@pytest.mark.parametrize("name,test_data", test_data)
def test_all(name, test_data):
    a, b, expected_diff = test_data["a"], test_data["b"], test_data["expected_diff"]
    diff = list(get_differences(a, b, path="a"))
```
The resulting messages contain valid python expressions to access the respective value and are very handy to be used inside the debugger.

tests/eve_tests/unit_tests/test_pattern_matching.py
tehrengruber: True, I've simplified as proposed and added more tests for nested things. Actually nesting was already tested before as the test for equality is already a recursive call.

src/eve/pattern_matching.py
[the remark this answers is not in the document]
egparedes: I proposed `fields` because it is now a widespread way to call plain data attributes of value classes (both  `dataclasses` and `pydantic` models use this convention), but `attributes` is also fine for me. The main point is to not to use `attrs`

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,159 @@
...
+    ...    def __init__(self):
+    ...        self.bar = 1
+    ...        self.baz = 2
+    >>> assert ObjectPattern(Foo, {"bar": 1}).match(Foo())
+    """
+
+    cls: type
+    attrs: dict[str, Any]
+
+    def match(self, other: Any, raise_exception: bool = False) -> bool:
```
egparedes, line 39: I really think that `raise_exception` should be a keyword-only argument here.

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,159 @@
...
+
+@singledispatch
+def get_differences(a: Any, b: Any, path: str = "") -> Iterator[tuple[str, str]]:
+    """Compare two objects and return a list of differences.
+
+    If the arguments are lists or dictionaries comparison is recursively per item. Objects are compared
+    using equality operator or if the left-hand-side is an `ObjectPattern` its type and attributes
+    are compared to the right-hand-side object. Only the attributes of the `ObjectPattern` are used
+    for comparison, disregarding potential additional attributes of the right-hand-side.
+    """
```
egparedes, line 83: Please make docstrings lines always around 80 chars, not longer.

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,159 @@
...
+    if type(a) != type(b):
```
egparedes, line 84: Types are singletons, so this could be a better way to check:

```suggestion
    if type(a) is not type(b):
```

src/eve/pattern_matching.py
egparedes: Why not using the suggested `yield from` ?

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,159 @@
...
+
+    A pattern matches an object if it is an instance of the specified
+    class and all attributes of the pattern (recursively) match the
+    objects attributes.
+
+    >>> class Foo:
+    ...    def __init__(self):
+    ...        self.bar = 1
+    ...        self.baz = 2
+    >>> assert ObjectPattern(Foo, {"bar": 1}).match(Foo())
```
egparedes, line 33: Slightly more realistic example:
```suggestion
    >>> class Foo:
    ...    def __init__(self, bar, baz):
    ...        self.bar = bar
    ...        self.baz = baz
    >>> assert ObjectPattern(Foo, {"bar": 1}).match(Foo(1, 2))
```

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,159 @@
...
+    >>> assert not foo_ir_.Foo(bar="bar").match(foo_ir.Foo(bar="baz", foo="bar"))  # doctest: +SKIP
+    """
+
+    module: ModuleType
+
+    def __getattr__(self, item: str) -> _ObjectPatternConstructor:
+        val = getattr(self.module, item)
+        if not inspect.isclass(val):
+            raise ValueError("Only classes allowed.")
+        return _ObjectPatternConstructor(val)
```
egparedes, line 159: As I said in my previous review, I think this util causes more harm than good since makes the creation of pattern objects look exactly the same as instantiation actual objects. I think a more explicit approach should be used here

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,159 @@
...
+    ...        self.bar = 1
+    ...        self.baz = 2
+    >>> assert ObjectPattern(Foo, {"bar": 1}).match(Foo())
+    """
+
+    cls: type
+    attrs: dict[str, Any]
+
+    def match(self, other: Any, raise_exception: bool = False) -> bool:
+        """Return if object pattern matches `other` using :func:`get_differences`.
```
egparedes, line 40:
```suggestion
        """Return ``True`` if object pattern matches ``other`` using :func:`get_differences`.
```

egparedes (changes requested):
Thanks for the changes. Except for the ModuleWrapper issue, It looks almost ready to go

src/eve/pattern_matching.py
tehrengruber: I missed 3 comments from your last review including this one. As discussed I see your point, but still like this solution better. I'll let this rest for a couple of days to see what I think and ask Rico and then we can revisit. Until then there is a comment and we should merge if the rest is sorted out.

src/eve/pattern_matching.py
tehrengruber: Mh, not a fan of this, but I've changed it. If I would be nit picking what you are proposing here is not as precise because it is unspecified what happens if the pattern does not match.

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,162 @@
...
+        for k, v_a, v_b in zip(a.keys(), a.values(), b.values()):
+            yield from get_differences(v_a, v_b, path=f'{path}["{k}"]')
+
+
+# TODO(tehrengruber): The ModuleWrapper has some magicness with it that could be confusing to the user. Do we really
+#  want this? Should we maybe use something more in the line of `P[foo_ir.Foo](...)`
+@dataclass(frozen=True)
+class ModuleWrapper:
+    """
+    Small wrapper to conveniently create `ObjectPattern`s for classes of a module.
```
DropD, line 147:
```suggestion
    """Small wrapper to conveniently create `ObjectPattern`s for classes of a module.
```

One single convention is still better than mixing, even if it is the ugliest convention.

egparedes (commented):
Minor style issues and the `ModuleWrapper` issue, which in my opinion is wrong and needs to be changed. I would only agree to merge this PR in its current state to unblock the merge of #368 if that PR is ready to go before this one.

src/eve/pattern_matching.py
egparedes: It's a minor thing. You can change it back if you prefer, but please use proper restructured text markups.

egparedes (approved):
It looks good to me.

PR #668, opened 2022-02-18
Functional: Field view vertical slice

tehrengruber:
Field view vertical slice
- new Program AST (PAST)
- Python to PAST parser
- PAST to ITIR lowering
- @program and @field_operator decorators
- several new tests

Blocked by: https://github.com/GridTools/gt4py/pull/667


src/functional/ffront/decorator.py
```diff
@@ -0,0 +1,227 @@
...
+        #  since ClosureRefs is immutable we have to resort to this rather ugly way of doing a copy
+        closure_refs = ClosureRefs(
+            **{
+                **{
+                    field.name: getattr(self.closure_refs, field.name)
+                    for field in dataclasses.fields(self.closure_refs)
+                },
+                "globals": {**self.closure_refs.globals, name: self},
+            }
+        )
```
egparedes, line 211: Did you try with `dataclasses.replace()` https://devdocs.io/python~3.10/library/dataclasses#dataclasses.replace ?

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -0,0 +1,143 @@
...
+        yield f"Function takes {len(func_type.args)} arguments, but {len(args)} were given."
+    for i, (a_arg, b_arg) in enumerate(zip(func_type.args, args)):
+        if a_arg != b_arg:
+            yield f"Argument {i} expected to be of type {a_arg}, but got {b_arg}."
+
+    # check for missing or extra keyword arguments
+    kw_a_m_b = set(func_type.kwargs.keys()).difference(set(kwargs.keys()))
+    if len(kw_a_m_b) > 0:
+        yield f"Missing required keyword argument(s) `{'`, `'.join(kw_a_m_b)}`."
+    kw_b_m_a = set(kwargs.keys()).difference(set(func_type.kwargs.keys()))
```
egparedes, line 40:
```suggestion
    kw_b_m_a = set(kwargs.keys()) - set(func_type.kwargs.keys())
```

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -0,0 +1,143 @@
...
+        return past.TupleExpr(elts=elts, type=ct.TupleType(types=[el.type for el in elts]))
+
+    def visit_Call(self, node: past.Call, **kwargs):
+        func = self.visit(node.func, **kwargs)
+        args = self.visit(node.args, **kwargs)
+        kwargs = self.visit(node.kwargs, **kwargs)
+
+        func_type = func.type
+        # functions returning fields in a program are implicitly converted into stencil closures. Change function
+        #  signature accordingly
```
egparedes, line 76: Make lines shorter

src/functional/ffront/decorator.py
DropD: Also, is the reason for ClosureRefs being immutable really stronger than the inconvenience it creates here?  Is it really difficult to avoid changing it when it shouldn't be in the normal course of things?

src/functional/ffront/decorator.py
```diff
@@ -0,0 +1,227 @@
...
+            definition=definition,
+        )
+
+    def __gt_type__(self) -> ct.FunctionType:
+        type_ = symbol_makers.make_symbol_type_from_value(self.definition)
+        assert isinstance(type_, ct.FunctionType)
+        return type_
+
+    def __gt_itir__(self):
+        return FieldOperatorLowering.apply(self.foast_node)
```
DropD, line 161: I should think these two deserve docstrings. Ideally referring to an explanation of how they can be used on custom classes to incorporate any IteratorIR snippet elsewhere.

tests/functional_tests/ffront_tests/test_program.py
```diff
@@ -0,0 +1,291 @@
...
+    fencil_pattern = itir_.FencilDefinition(
+        id="copy_restrict_program",
+        params=[
+            itir_.Sym(id="in_field"),
+            itir_.Sym(id="out_field"),
+            itir_.Sym(id="__in_field_size_0"),
+            itir_.Sym(id="__out_field_size_0"),
+        ],
+        closures=[closure_pattern],
+    )
```
DropD, line 218: This goes for all Iterator IR patterns: They all seem complete, as in every attribute is given (ITIR being very minimal this is not surprising). Is there any reason for using the pattern matching here, except to avoid false negatives in the future if things like source locations should be added to ITIR?

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+@dataclass(frozen=True)
+class ObjectPattern:
+    cls: type
+    attrs: dict[str, Any]
+
+    def matches(self, other: Any, raise_: bool = False) -> bool:
+        """Return if object pattern matches `other` using :func:`get_differences`.
+
+        If `raise_` is specified raises an exception with all differences found.
+        """
```
DropD, line 29:
```suggestion
        """
        Return if object pattern matches `other` using :func:`get_differences`.

        If `raise_` is specified raises an exception with all differences found.
        """
```

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+import inspect
+from dataclasses import dataclass
+from types import ModuleType
+from typing import Any, Iterator
+
+
+@dataclass(frozen=True)
+class ObjectPattern:
```
DropD, line 21: Ideally, a docstring with a (running) doctest example would be nice!

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+            raise ValueError(f"Object and pattern don't match:\n  {diffs_str}")
+        return len(diffs) == 0
+
+    def __str__(self) -> str:
+        attrs_str = ", ".join([f"{str(k)}={str(v)}" for k, v in self.attrs.items()])
+        return f"{self.cls.__name__}({attrs_str})"
+
+
+@dataclass(frozen=True)
+class ObjectPatternConstructor:
```
DropD, line 42: Again, a docstring with example would help.

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+@dataclass(frozen=True)
+class ModuleWrapper:
+    """
+    Small wrapper to conveniently create `ObjectPattern`s for classes of a module.
+
+    Example:
+    >>> import foo_ir  # doctest: +SKIP
+    >>> foo_ir_ = ModuleWrapper(foo_ir)  # doctest: +SKIP
+    >>> assert foo_ir_.Foo(bar="baz").matches(foo_ir.Foo(bar="baz", foo="bar"))  # doctest: +SKIP
+    >>> assert not foo_ir_.Foo(bar="bar").matches(foo_ir.Foo(bar="baz", foo="bar"))  # doctest: +SKIP
```
DropD, line 130: Why no running docstrings?

DropD (changes requested):

src/functional/ffront/program_ast.py
```diff
@@ -0,0 +1,101 @@
...
+
+class Assign(Stmt):
+    target: Symbol
+    value: Expr
+
+
+class Program(LocatedNode, SymbolTableTrait):
+    id: SymbolName  # noqa: A003
+    params: list[Symbol[common_types.DataType]]
+    body: list[Union[Expr, Stmt]]
```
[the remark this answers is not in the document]
tehrengruber: Indeed, only calls are allowed now.

src/functional/ffront/symbol_makers.py
```diff
@@ -165,28 +164,27 @@ def make_symbol_type_from_typing(
...
-        return foast.Symbol[type(symbol_type)](
-            id=name, type=symbol_type, namespace=namespace, location=location
-        )
+        return symbol_type
     else:
         raise common.GTTypeError(f"Impossible to map '{value}' value to a Symbol")
 
 
 # TODO(egparedes): Add source location info (maybe subclassing FieldOperatorSyntaxError)
+# TOD(tehrengruber): Find a better name as everything here is not tied to field operators
```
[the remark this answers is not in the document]
tehrengruber: This error is only ever used for type annotations. I renamed it to `TypingError` in the style of the `typing` module.

src/functional/ffront/past_to_itir.py
```diff
@@ -0,0 +1,194 @@
...
+from typing import Union
+
+from eve import NodeTranslator, SymbolTableTrait
+from functional.common import GTTypeError
+from functional.ffront import common_types
+from functional.ffront import program_ast as past
+from functional.iterator import ir as itir
+
+
+def _size_arg_from_field(field_name: str, dim: int):
```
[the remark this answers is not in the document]
tehrengruber: It is used in multiple places and I try to keep machinery involved in generating symbol names isolated such that it is easy to refactor them when conventions change.

src/functional/ffront/past_to_itir.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def _visit_stencil_call_out_arg(
+        self, node: past.Expr, **kwargs
+    ) -> tuple[list[itir.SymRef], itir.FunCall]:
+        # as the ITIR does not support slicing a field we have to do a deeper inspection of the PAST to emulate
+        #  the behaviour
+        if isinstance(node, past.Subscript):
+            out_field_name: past.Name = node.value
+            if isinstance(node.slice_, past.TupleExpr):
+                out_field_slice_: list[past.Slice] = node.slice_.elts
```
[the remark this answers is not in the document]
tehrengruber: No, actually mypy is disabled for most of the modules in ffront due to the issue with python's match statement. I narrowed the condition to only allow past.Slice nodes.

src/functional/ffront/past_to_itir.py
```diff
@@ -0,0 +1,194 @@
...
+                    )
+                    if slice_.upper.value < 0:
+                        upper = itir.FunCall(
+                            fun=itir.SymRef(id="plus"),
+                            args=[dim_size, self.visit(slice_.upper, **kwargs)],
+                        )
+                    else:
+                        upper = self.visit(slice_.upper, **kwargs)
+                else:
+                    raise AssertionError()
```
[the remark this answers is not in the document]
tehrengruber: Both parts were simple in the beginning without a need to place them into a separate function. I factored them out new that they grow in complexity.

src/functional/ffront/decorator.py
```diff
@@ -0,0 +1,227 @@
...
+            closure_refs=closure_refs,
+            externals={} if externals is None else externals,
+            backend=backend,
+            definition=definition,
+        )
+
+    def __post_init__(self):
+        if self.backend is None:
+            self.backend = default_backend
+
```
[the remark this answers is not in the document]
tehrengruber: I just forgot to remove this function (and to make `Program` frozen). The backend argument to the decorator is `None` by default and I didn't want to have multiple places where the backend is set to `DEFAULT_BACKEND`. However there seems to be no good way to transform a field of a frozen dataclass (here from `None` -> `DEFAULT_BACKEND`) and I resorted to using the default value at the only place where it used.

src/functional/ffront/decorator.py
tehrengruber: @egparedes That is exactly what I was looking for. Works great.
@DropD Good point. IMO it makes sense for ClosureRefs to be immutable and with `replace` the inconvenience is managable.

src/functional/ffront/common_types.py
```diff
@@ -39,6 +45,13 @@ class SymbolTypeVariable(SymbolType):
     bound: typing.Type[SymbolType]
 
 
+@dataclass(frozen=True)
+class VoidType(SymbolType):
+    """Return type of a function without return values."""
```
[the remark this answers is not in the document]
tehrengruber: There is one place where the DSL is not functional and that is the program part (fencil in ITIR linguee):
```python
# functional part
@field_operator
def foo(...):
   ...

# stateful part
@program
def foo(in_field, out_field):
  stencil(in_field, out=out_field)
```
The syntax we settled on in the design process is not optimal, as the signature of the field operator changes depending on context so we might want to make state more visible in the future, but this is a different story:
```python
# functional part
@field_operator
def foo(...):
   ...

# stateful part
@program
def foo(in_field, out_field):
  out_field = stencil(in_field)
```

src/functional/ffront/symbol_makers.py
```diff
@@ -165,28 +164,27 @@ def make_symbol_type_from_typing(
...
-    """Make a symbol node from a Python value."""
+def make_symbol_type_from_value(value: Any) -> common_types.SymbolType:
+    """Make a symbol type from a Python value."""
     if isinstance(value, type) or type(value).__module__ == "typing":
         # we don't have types of types so disallow this
         raise ValueError("The type of a symbol can not be a type itself.")
-    type_ = typingx.get_typing(value, annotate_callable_kwargs=True)
-
-    symbol_type = make_symbol_type_from_typing(type_)
+    if hasattr(value, "__gt_type__"):
```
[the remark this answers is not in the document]
tehrengruber: I have added a Protocol `GTCallable` with a more elaborate description.

src/functional/ffront/past_to_itir.py
```diff
@@ -0,0 +1,194 @@
...
+                    )
+                    if slice_.lower.value < 0:
+                        lower = itir.FunCall(
+                            fun=itir.SymRef("plus"),
+                            args=[dim_size, self.visit(slice_.lower, **kwargs)],
+                        )
+                    else:
+                        lower = self.visit(slice_.lower, **kwargs)
+                else:
+                    raise AssertionError()
```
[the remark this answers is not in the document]
tehrengruber: If I remember correctly, this was not exactly the error message, but yes the linter complained about `assert False` (for good reason).

tests/functional_tests/ffront_tests/test_program.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+def invalid_call_sig_program(
+    in_field: Field[[IDim], "float64"], out_field: Field[[IDim], "float64"]
+):
+    identity(in_field, in_field, out=out_field)
+
+
+def invalid_slice_program(in_field: Field[[IDim], "float64"], out_field: Field[[IDim], "float64"]):
+    identity(in_field, out=out_field[1:2, 3:4])
```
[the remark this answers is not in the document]
tehrengruber: You are absolutely right. All field operators are now inside a fixture, but decorated as this is required before being used inside a program (at least with respect to being executable). It is possible to avoid this, but not without duplicating programs for parsing+lowering and execution. Programs on the other hand are undecorated now and only decorated where needed. In total I must say the resulting fixture code doesn't feel elegant, so I am open for suggestions on improvement.

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -0,0 +1,143 @@
...
+    if len(kw_b_m_a) > 0:
+        yield f"Got unexpected keyword argument(s) `{'`, `'.join(kw_b_m_a)}`."
+
+    for kwarg in set(func_type.kwargs.keys()) & set(kwargs.keys()):
+        if func_type.kwargs[kwarg] != kwargs[kwarg]:
+            yield f"Expected keyword argument {kwarg} to be of type {func_type.kwargs[kwarg]}, but got {kwargs[kwarg]}."
+
+
+class ProgramTypeDeduction(NodeTranslator):
+    contexts = (SymbolTableTrait.symtable_merger,)  # type: ignore[assignment]
```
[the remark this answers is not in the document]
tehrengruber: Yes they are and correctly as far as I can tell. I have no clue why mypy fails here, it is used everywhere else just as it is here. I just ignored the error, but I have to admit I was not overly curious to find its origin after the annoyance with the erratic mypy failures on `match` statements in files that were already excluded.

tests/eve_tests/unit_tests/test_pattern_matching.py
```diff
@@ -0,0 +1,81 @@
...
+    name: str
+    a: Any
+    b: Any
+    expected_diff: list[tuple[str, str]]
+
+
+test_data: list[tuple[str, TestData]] = []
+
+
+def register_test_case(name, a, b, expected_diff):
```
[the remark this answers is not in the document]
tehrengruber: Yes, I agree. I wondered why there is no tooling in pytest for this (maybe I'm missing it?).

src/functional/ffront/past_to_itir.py
```diff
@@ -0,0 +1,194 @@
...
+        elif isinstance(idx, past.Constant):
+            if idx.value < 0:
+                return itir.FunCall(
+                    fun=itir.SymRef("minus"), args=[dim_size, self.visit(idx, **kwargs)]
+                )
+            else:
+                return self.visit(idx, **kwargs)
+        raise RuntimeError("Expected `None` or `past.Constant` node.")
+
+    def _visit_stencil_call_out_arg(
```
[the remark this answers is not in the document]
tehrengruber: You are absolutely right, but reducing the depth doesn't really help. The function itself is already a very narrow case, it is only responsible for lowering the output argument of a stencil call. Adding another layer just means jumping around in code for understanding what conceptually belongs together. Nonetheless I hope we eventually get rid of this code, but not needing to emulate this behavior.

src/functional/ffront/decorator.py
tehrengruber: There is a Protocol `GTCallable` now with some explanations. Examples on how to really use it will be added later when the concept matured. As you might have seen in the raw iterator decorator PR, where signature and definition don't play nice together, there is still room for improvement

tests/functional_tests/ffront_tests/test_program.py
tehrengruber: False negatives in the future were one reason, but the main reason is that the `StencilClosure` is not complete here. There is now an additional test, that is actually testing each and every attribute, but I still use the pattern matching to avoid having a different way of comparing there (the pattering matching could be improved in that respect).

tehrengruber:
Ready for another round of reviews. I have very raw and incomplete notes for an ADR, but didn't find time to actually write one. Will not optimal I think we should merge before the ADR as the PR is already needed for Ricos project.
@DropD I'll address your comments with respect to pattern matching in #667.

src/eve/pattern_matching.py
tehrengruber: Creating the `foo_ir` dynamically makes the doctest much harder to read and distracts from what is meant to be documented here. If you want an example that is standalone executable I can add one, but I personally like shortness over completeness more here.

src/eve/pattern_matching.py
tehrengruber: This is just a partial function application, I've made it private as it should have been from the beginning.

src/eve/pattern_matching.py
tehrengruber: At least in eve this is the convention used for docstrings.

src/functional/ffront/past_passes/type_deduction.py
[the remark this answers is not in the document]
DropD: I think what @egparedes  meant was

```suggestion
    contexts: ClassVar[Optional[Tuple[ContextCallable, ...]]] = (SymbolTableTrait.symtable_merger,)  # type: ignore[assignment]
```
where

```python
ContextCallable = Callable[["NodeVisitor", concepts.TreeNode, Dict[str, Any]], ContextManager[None]]
```

like in the `eve.NodeVisitor` class.

src/functional/ffront/past_to_itir.py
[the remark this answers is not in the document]
DropD: The problem of jumping around in code only happens when the extracted parts are not logically cohesive and / or not clearly named. But this is probably "good enough for now".

src/eve/pattern_matching.py
```diff
@@ -0,0 +1,139 @@
...
+        if type(a) != type(b):
+            yield (path, f"Expected a value of type {type(a).__name__}, but got {type(b).__name__}")
+        elif a != b:
+            yield (path, f"Values are not equal. `{a}` != `{b}`")
+
+
+@dataclass(frozen=True)
+class ModuleWrapper:
+    """
+    Small wrapper to conveniently create `ObjectPattern`s for classes of a module.
```
DropD, line 124:
```suggestion
    """Small wrapper to conveniently create `ObjectPattern`s for classes of a module.
```

src/eve/pattern_matching.py
DropD: In that case we have separate conventions for eve and gt4py.

src/functional/ffront/dialect_parser.py
```diff
@@ -0,0 +1,166 @@
...
+from eve.type_definitions import SourceLocation
+from functional import common
+from functional.ffront.source_utils import ClosureRefs, SourceDefinition, SymbolNames
+
+
+DialectRootT = TypeVar("DialectRootT")
+
+
+@dataclass(frozen=True, kw_only=True)
+class DialectParser(ast.NodeVisitor, Generic[DialectRootT]):
```
egparedes, line 31: Shouldn't it also be generic with respect to the syntax error type?

src/functional/ffront/dialect_parser.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # 'SymbolNames.from_source()' uses the symtable module to analyze the isolated source
+        # code of the function, and thus all non-local symbols are classified as 'global'.
+        # However, 'closure_refs' comes from inspecting the live function object, which might
+        # have not been defined at a global scope, and therefore actual symbol values could appear
+        # in both 'closure_refs.globals' and 'self.closure_refs.nonlocals'.
+        if not set(closure_refs.globals) | set(closure_refs.nonlocals.keys()) == (
+            global_names | nonlocal_names
+        ):
+            raise AssertionError("ClosureRefs do not agree with information from symtable module.")
```
egparedes, line 95: I would move this function out of the class definition

src/functional/ffront/dialect_parser.py
```diff
@@ -0,0 +1,166 @@
...
+            output_ast = cls._postprocess_dialect_ast(
+                cls(
+                    source=source,
+                    filename=filename,
+                    starting_line=starting_line,
+                    closure_refs=closure_refs,
+                    externals_defs=externals or {},
+                ).visit(definition_ast)
+            )
+            cls._assert_source_invariants(source_definition, closure_refs)
```
egparedes, line 61: I would call it like:
```
if __debug__:
    assert_source_invariants(source_definition, closure_refs)
```
Check here for meaning of `__debug__`, https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement
See other comment about moving this function out of the class.

src/functional/ffront/dialect_parser.py
```diff
@@ -0,0 +1,166 @@
...
+        except SyntaxError as err:
+            if not err.filename:
+                err.filename = filename
+            if not isinstance(err, DialectSyntaxError):
+                err.lineno = (err.lineno or 1) + starting_line - 1
```
egparedes, line 66: Add TODO about idea of preprocessing pass fixing the source locations.

src/functional/ffront/decorator.py
```diff
@@ -0,0 +1,294 @@
...
+
+@typing.runtime_checkable
+class GTCallable(Protocol):
+    """Protocol (abstract base class) defining the interface for callables.
+
+    Direct subclasses of this base class only need to implement the
+    abstract methods to be callable from inside a program or from other
+    callables.
+    """
```
egparedes, line 47:
```suggestion
    """Typing Protocol (abstract base class) defining the interface for subroutines.

    Any class implementing the methods defined in this protocol can be
    called from ``ffront``  programs or operators.
    """
```

src/functional/ffront/decorator.py
```diff
@@ -0,0 +1,294 @@
...
+
```
egparedes, line 48: I suggest to add a generic `__call__` method without a specific signature (to avoid mypy errors in subclasses)
```suggestion

__call__: Callable
```

src/functional/ffront/decorator.py
```diff
@@ -0,0 +1,294 @@
...
+    def __gt_type__(self) -> ct.FunctionType:
+        """Return symbol type, i.e. signature and return type.
+
+        The type is used internally to populate the closure vars of the
+        various dialects root nodes (i.e. FOAST Field Operator, PAST Program)
+        """
+        ...
+
+    def __gt_itir__(self) -> itir.FunctionDefinition:
```
egparedes, line 57: Either the protocol is renamed to GTOperator or the return type hint is updated, or a TODO node is added.

src/functional/ffront/symbol_makers.py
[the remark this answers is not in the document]
egparedes: I feel it is better for performance and stability to use `hasattr(...,  __gt_type__)` here, but at the same type it's a bit weird that there is a `runtime_checkable` protocol exist and it's not used ...

egparedes (commented):
It looks good. This round contains much fewer and shorter comments.

src/functional/ffront/decorator.py
```diff
@@ -0,0 +1,294 @@
...
+    @classmethod
+    def from_function(
+        cls,
+        definition: types.FunctionType,
+        externals: Optional[dict] = None,
+        backend: Optional[str] = None,
+    ):
+        closure_refs = ClosureRefs.from_function(definition)
+        foast_node = FieldOperatorParser.apply_to_function(definition)
+        return FieldOperator(
```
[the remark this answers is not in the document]
tehrengruber: That's not nitpicking, but plainly wrong... Fixed

src/functional/ffront/type_info.py
```diff
@@ -22,6 +23,33 @@ def is_complete_symbol_type(fo_type: SymbolType) -> TypeGuard[SymbolType]:
     return False
 
 
+def check_signature(
+    func_type: FunctionType, args: list[SymbolType], kwargs: dict[str, SymbolType]
+) -> Iterator[str]:
+    """Check if a function can be called given arguments and keyword arguments types.
+
+    All types must be concrete/complete.
+    """
```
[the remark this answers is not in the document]
tehrengruber: I refactored the function into a new function `function_signature_incompatibilities` returning all incompatibilities for a call with given arguments and added a new method `is_callable_for_args` to `TypeInfo`. `function_signature_incompatibilities` still uses generators to allow short-circuting.

src/functional/ffront/dialect_parser.py
tehrengruber: No, wouldn't that be wrong actually? One could argue that the syntax error type is a property of the dialect in which case it should be derived from that type, but you should not change the syntax error type while keeping the dialect and vice-versa and thus only the dialect is generic. I made the syntax error type a class var assigning the responsibility to decide what type of error is to be thrown to the concrete parser implementation (which is the only place syntax errors should arise from).

src/functional/ffront/decorator.py
tehrengruber: As it is right now a `GTCallable` does not need to be callable directly, so I think we should skip this for now. The fact that the signature of `__call__` to a field operator does not match it's `__gt_type__` signature (we should fix this at some point) puts further insult to inquiry.

src/functional/ffront/decorator.py
tehrengruber: A `GTCallable` doesn't need to be an operator it can also be a regular function in terms of the ITIR. Only when it appears inside a stencil closure it needs to be an operator.

src/functional/ffront/symbol_makers.py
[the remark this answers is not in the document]
tehrengruber: Yes this is not optimal. Another problem is that the protocol currently lives in the `decorator.py` file for the lack of a better place, but it doesn't really belong there. I have added a todo here for the time being.

src/functional/ffront/past_passes/type_deduction.py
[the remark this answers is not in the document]
tehrengruber: This shouldn't be needed. I checked and mypy uses the type information of the parent class. Additionally this is exactly the same as everywhere else in the code

src/functional/ffront/decorator.py
```diff
@@ -0,0 +1,294 @@
...
+from functional.ffront import common_types as ct
+from functional.ffront import field_operator_ast as foast
+from functional.ffront import program_ast as past
+from functional.ffront import symbol_makers
+from functional.ffront.foast_to_itir import FieldOperatorLowering
+from functional.ffront.func_to_foast import FieldOperatorParser
+from functional.ffront.func_to_past import ProgramParser
+from functional.ffront.past_passes.type_deduction import ProgramTypeDeduction
+from functional.ffront.past_to_itir import ProgramLowering
+from functional.ffront.source_utils import ClosureRefs
```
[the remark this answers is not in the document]
tehrengruber: This is the convention we used elsewhere and having another convention just because the amount of imports exceeds a limit doesn't make sense to me (I don't like this style of importing and naming anyway). The problem is another one here: This file contains too many different components and should be split up. I've added a comment.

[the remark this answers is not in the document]
egparedes: I'm not aware of any convention about directly importing objects instead of modules (except for typing classes) but you're right this is not  documented anywhere yet.

src/functional/ffront/type_info.py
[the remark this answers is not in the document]
egparedes: Ok, that works for me. However, I don't really see the point of allowing short-circuiting a function that it is part of the frontend, not performance-critical and most likely should always be fully evaluated for good error messages.

src/functional/ffront/decorator.py
egparedes: It is impossible to define a generic field-operator or protocol whose `__call__` signature matches the `__gt_type__` signature of the object without dependent types, so we can already forget about having a fully correct solution here.
However, I think the protocol definition should state clearly that this object has to be callable (for embedded execution) even if right now.actual implementations raise a `NotImplementedError`

src/functional/ffront/dialect_parser.py
egparedes: >   One could argue that the syntax error type is a property of the dialect in which case it should be derived from that type
In that case, the DialectError should be parameterized with the Dialect itself, like:

```python
class DialectSyntaxError(common.GTSyntaxError, Generic[DialectRootT]):
    ...

class DialectParser(ast.NodeVisitor, Generic[DialectRootT]):
    ...
    syntax_error_cls: ClassVar[Type[DialectSyntaxError[DialectRootT]]]
```
But anyway this is not supported in Python because `ClassVar`s cannot hold `TypeVar`s, so I guess the current solution is the best one.

egparedes (approved):
It looks good to me. The only extra change I would like to have is adding __call__ to the GTProtocol as I said in the comment. If my arguments are not convicing, at least it should be added as a comment with a TODO explaining that it will be required for embedded execution.

src/functional/ffront/decorator.py
tehrengruber: I'm not completely sure I correctly understand what you mean here, but my point is that a field operator has a `__gt_type__` signature whose return type is a field, while it's `__call__` signature has return type void. Solving this is a question of using a different syntax inside programs. I've added a comment with respect to `__call__`.

PR #670, opened 2022-02-18
Functional: @raw_itir_stencil decorator

tehrengruber:
Some features, e.g. reductions, are still in the working or missing in the frontend. This PR adds an additional decorator `@raw_itir_stencil` that may be used inside programs or field operators to inject raw iterator IR nodes. The signature of the function needs to follow the rules of how FieldOperators are typed for the time beeing.

Example:
```python
@raw_itir_stencil
def add_self(arg: Field[[IDim], "float64"]) -> Field[[IDim], "float64"]:
    return itir.FunCall(fun=itir.SymRef(id="plus"), args=[
        itir.FunCall(fun=itir.SymRef(id="deref"), args=[arg]),
        itir.FunCall(fun=itir.SymRef(id="deref"), args=[arg])])

@program
def raw_copy_program(in_field: Field[[IDim], "float64"], out_field: Field[[IDim], "float64"]):
    add_self(in_field, out=out_field)
```

### DIFF
--- a/src/functional/ffront/decorator.py
+++ b/src/functional/ffront/decorator.py
@@ -36,10 +36,9 @@ from functional.ffront.past_to_itir import ProgramLowering
 from functional.ffront.source_utils import ClosureRefs
 from functional.iterator import ir as itir
 from functional.iterator.backend_executor import execute_program
-
+from functional.iterator.runtime import Offset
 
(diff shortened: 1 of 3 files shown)


havogt:
outdated, we could instead support the pretty-printed format if we want still want this feature

PR #671, opened 2022-02-21
Unify all extent computation paths on OIR

jdahm:
## Description

A step in merging horizontal regions and removing the legacy backends is to unify the places that extents are calculated. These are:

1. `make_args_from_gtir` for the field_info
2. oir horizontal execution merging
3. all GTC backends: most are doing their own analysis


jdahm:
bors try

src/gtc/passes/oir_optimizations/utils.py
```diff
@@ -201,35 +202,59 @@ class Context:
...
+            ctx.fields[name] |= total_extent
+
+        for name in results.write_fields():
+            ctx.fields.setdefault(name, horizontal_extent)
 
 
-def compute_horizontal_block_extents(node: oir.Stencil) -> Dict[int, Extent]:
-    ctx = StencilExtentComputer().visit(node)
+def compute_horizontal_block_extents(node: oir.Stencil, **kwargs: Any) -> Dict[int, Extent]:
+    ctx = StencilExtentComputer(**kwargs).visit(node)
```
gronerl, line 247: It doesn't have to be in this PR, but I can't find in the stack of horizontal region PR's where this gets extended to the case of regions. Did I miss that?

gronerl (commented):
I only have one suggestion and a question.

src/gtc/passes/oir_optimizations/utils.py
jdahm: Ah ha! That means the change is well hidden in layers of abstraction. The specific change is separating the concepts of offsets and extents. There is a method that converts between them given the extent for the horizontal execution (`horizontal_extent`). See [here](https://github.com/GridTools/gt4py/pull/678/files#diff-1ad76d78dea66f59e27bce7379282bc29ce0633d4fa3274c1b9773882d4bcfd7R252) for the call.

jdahm:
bors try

src/gtc/passes/oir_optimizations/utils.py
gronerl: :+1: thanks for the explanation

gronerl (approved):

PR #672, opened 2022-02-22
functional/eve: Refactor eve.utils.UIDGenerator as an instantiable object.

egparedes:
Refactor eve.utils.UIDGenerator as an instantiable object instead of a pure singleton class.


DropD (approved):
Thanks, this should be useful in the reduction lowering now.

Issue #566, opened 2021-11-30
Remove deprecated legacy backends


jdahm:
The GTC toolchain is now very mature and has only a few minor bugs to iron out. It implements the parallel model very well. The legacy toolchain is no longer maintained, and does not clearly implement the parallel model. In the next release of GT4py let's remove the old code.

When doing this, we need to also remove the `dawn4py` integration since it was only ever integrated with the legacy backends.

Steps:
- [x] Disable tests and unregister backends (#620)
- [x] Remove unused backend files
- [x] Remove unused tests
- [x] Move references to `gt_backends.py` and other legacy files to better locations
- [x] Remove GridTools1 from the source manager and update regression tests to use GridTools2 sources
- [x] Remove `gtc:` prefix on GTC backends.
- [x] Move definition IR into frontend directory.

gronerl:
I propose we also remove the `gtc:` prefix from all backend names.

jdahm:
Yes, agreed 🎉 ! I was going to do that in a separate branch after this one.

jdahm:
Yes, agreed 🎉 ! I added it to the list above.

jdahm:
This is done!

PR #673, opened 2022-02-22
Make GTScript Frontend output GTIR

### LINKED ISSUE #501: Bijection between defir and gtir
I noticed the other day that translating defir to gtir loses some information that is required to generate the stencil module. Before removing the old backends we need to make sure that all this information is captured in gtir.

jdahm:
## Description

This makes the gtscript frontend directly output gtir, and removes definition IR from everywhere except the frontend itself, which still writes definition IR internally before converting to gtir.

- Make gtir contain all the info (sources, externals, api_signature) that definition ir contains.
- Remove all references to definition ir except those inside the gtscript frontend.
- Moved several Python extension generation classes into `gt4py/backend/gtc_backend/base.py` to isolate `gtc_backend` from legacy modules in `gt4py.backend`, since these will be removed in the next step.

Part of #566.

Resolves #501.
Blocked by #671.


src/gt4py/backend/gt_backends.py
```diff
@@ -846,6 +856,29 @@ def generate_extension(self, **kwargs: Any) -> Tuple[str, str]:
         return self.make_extension(uses_cuda=False)
 
 
+class CUDAPyExtModuleGenerator(PyExtModuleGenerator):
```
jdahm, line 859: Use `CUDAPyExtModuleGenerator` from `.gtc_backend.base` and delete this.

src/gt4py/backend/gtc_backend/cuda/backend.py
```diff
@@ -19,43 +19,44 @@
...
+class GTCCudaExtGenerator(GeneratorClass):
     def __init__(self, class_name, module_name, backend):
         self.class_name = class_name
         self.module_name = module_name
         self.backend = backend
 
-    def __call__(self, definition_ir) -> Dict[str, Dict[str, str]]:
-        gtir = GtirPipeline(DefIRToGTIR.apply(definition_ir)).full()
-        base_oir = gtir_to_oir.GTIRToOIR().visit(gtir)
+    def __call__(self, ir: gtir.Stencil) -> Dict[str, Dict[str, str]]:
```
jdahm, line 52: Find a better naming scheme for the `gtir` here and elsewhere.

jdahm: the name `gtir` was shadowing the imported module in a few spots. Since it's the "default" IR now I thought perhaps `ir` was best.

tests/test_unittest/test_gtc/test_gtir.py
```diff
@@ -53,69 +41,18 @@
...
-        body=[copy_assign],
-        temporaries=[],
-    )
-
-
-@pytest.fixture
-def copy_computation(copy_v_loop):
-    yield Stencil(
+def test_copy():
+    copy_computation = StencilFactory(
```
tehrengruber, line 45: I'm not sure it is a good idea to rely on the factory for this test. The whole purpose of the test is to test construction and some basic properties of the node, but by using the factory a test failure will occur inside the factory, making a fix rather inconvenient. I think the original test is fine, is there a reason to change it here?

src/gt4py/backend/module_generator.py
```diff
@@ -260,23 +258,18 @@ def generate_domain_info(self) -> str:
...
+        NDIM = 3
+        return repr(
             DomainInfo(
-                parallel_axes=tuple(ax.name for ax in parallel_axes),
-                sequential_axis=sequential_axis,
+                parallel_axes=("I", "J"),
+                sequential_axis="K",
                 min_sequential_axis_size=min_sequential_axis_size,
-                ndim=len(parallel_axes) + (1 if sequential_axis else 0),
+                ndim=NDIM,
```
tehrengruber, line 270:
```suggestion
                ndim=3,
```
Just a suggestion

src/gtc/gtir.py
```diff
@@ -234,14 +234,23 @@ def _no_write_and_read_with_horizontal_offset(
...
+    name: Str
+    is_keyword: bool
+    default: Str
+
+
 class Stencil(LocNode, SymbolTableTrait):
     name: Str
-    # TODO(havogt) deal with gtscript externals
+    api_signature: List[Argument]
     params: List[Decl]
```
tehrengruber, line 246: I think the information of the `api_signature` should be integrated into params at some point. Either by reflecting the structure of the signature in the values or in the structure of the member (list of arguments + dict of keyword arguments). I would leave a small todo comment.

src/gt4py/backend/gtc_backend/defir_to_gtir.py
```diff
@@ -132,13 +133,28 @@ def visit_StencilDefinition(self, node: StencilDefinition) -> gtir.Stencil:
...
+            externals = {name: str(value) for name, value in node.externals.items()}
+        else:
+            externals = {}
         return gtir.Stencil(
             name=node.name,
+            api_signature=[
+                gtir.Argument(
+                    name=f.name,
+                    is_keyword=f.is_keyword,
+                    default=str(f.default) if not isinstance(f.default, type(Empty)) else "",
```
tehrengruber, line 146: Isn't `type(Empty) == type`? Not sure what the motive is here or is it a type?

src/gt4py/backend/gtc_backend/defir_to_gtir.py
```diff
@@ -132,13 +133,28 @@ def visit_StencilDefinition(self, node: StencilDefinition) -> gtir.Stencil:
         field_params = {f.name: self.visit(f) for f in node.api_fields}
         scalar_params = {p.name: self.visit(p) for p in node.parameters}
         vertical_loops = [self.visit(c) for c in node.computations if c.body.stmts]
+        if node.externals is not None:
+            externals = {name: str(value) for name, value in node.externals.items()}
```
tehrengruber, line 137:
```suggestion
            externals = {name: repr(value) for name, value in node.externals.items()}
```
Not completely sure, but at least this was the previous strategy used in `generate_constants`.

tehrengruber (changes requested):
Looks good, only small things.

src/gtc/passes/gtir_prune_unused_parameters.py
```diff
@@ -33,6 +35,7 @@ def prune_unused_parameters(node: gtir.Stencil) -> gtir.Stencil:
         .to_list()
     )
     used_params = list(filter(lambda param: param.name in used_variables, node.params))
-    return gtir.Stencil(
-        name=node.name, params=used_params, vertical_loops=node.vertical_loops, loc=node.loc
-    )
+
+    new_stencil = copy.deepcopy(node)
```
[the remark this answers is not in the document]
jdahm: Good find! I changed it to copy and use the `update` keyword.

src/gtc/gtir.py
```diff
@@ -234,14 +234,23 @@ def _no_write_and_read_with_horizontal_offset(
...
     vertical_loops: List[VerticalLoop]
+    externals: Dict[str, str]
```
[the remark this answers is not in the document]
jdahm: Ah yes here is a problem: sometimes externals are not literals. They can be `gtscript.AxisIndex` or even (I think) `gtscript.AxisInterval` for horizontal regions. I guess we could use a `Union`. The other option is to filter these out entirely and only keep the plain-old-data types (the isinstance `numbers.Number`).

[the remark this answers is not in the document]
jdahm: I added `_make_literal` to convert. Let me know what you think.

src/gt4py/backend/module_generator.py
```diff
@@ -236,10 +234,10 @@ def generate_constants(self) -> Dict[str, str]:
 
         This is unlikely to require overriding.
         """
-        if self.builder.definition_ir.externals:
+        if self.builder.gtir.externals:
             return {
                 name: repr(value)
-                for name, value in self.builder.definition_ir.externals.items()
+                for name, value in self.builder.gtir.externals.items()
                 if isinstance(value, numbers.Number)
```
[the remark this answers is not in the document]
jdahm: Now we need to add a repr method on literals.

src/gt4py/stencil_builder.py
```diff
@@ -255,22 +253,11 @@ def pkg_qualname(self) -> str:
...
-
-    @property
-    def implementation_ir(self) -> "StencilImplementation":
-        return self._build_data.get("iir") or self._build_data.setdefault(
-            "iir", gt4py.analysis.transform(self.definition_ir, self.options)
-        )
-
     @property
     def gtir_pipeline(self) -> GtirPipeline:
         return self._build_data.get("gtir_pipeline") or self._build_data.setdefault(
```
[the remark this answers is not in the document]
jdahm: I elected to put the check in the frontend itself because that outputs gtir directly.

tests/test_unittest/test_gtc/test_gtir.py
jdahm: I agree that the error is poorly formatted when using the factories, but this is also really fragile because if any attribute changes, the tests break. I'm also not sure this is the use-case for pytest fixtures. I suppose I thought this was written before the factories were added.

That said, would you rather I switch it back? I'm OK doing that if you think the semantic of creating nodes without factories is critical.

src/gtc/gtir.py
jdahm: I'll do that in a follow-up: https://github.com/GridTools/gt4py/issues/710.

src/gt4py/backend/gtc_backend/defir_to_gtir.py
jdahm: I'll try removing it!

src/gt4py/backend/gtc_backend/defir_to_gtir.py
jdahm: Solved a different way by using `gtir.Literal` instead of strings.

src/gt4py/backend/gtc_backend/cuda/backend.py
```diff
@@ -49,19 +49,19 @@ def __init__(self, class_name, module_name, backend):
...
-        oir = oir_pipeline.run(base_oir)
-        oir = FillFlushToLocalKCaches().visit(oir)
-        cuir = OIRToCUIR().visit(oir)
-        cuir = kernel_fusion.FuseKernels().visit(cuir)
-        cuir = extent_analysis.CacheExtents().visit(cuir)
+        o_ir = oir_pipeline.run(base_o_ir)
+        o_ir = FillFlushToLocalKCaches().visit(o_ir)
+        cu_ir = OIRToCUIR().visit(o_ir)
+        cu_ir = kernel_fusion.FuseKernels().visit(cu_ir)
+        cu_ir = extent_analysis.CacheExtents().visit(cu_ir)
```
tehrengruber, line 62: I don't think this is a good solution to the name clash problem.

src/gt4py/backend/gtc_backend/defir_to_gtir.py
```diff
@@ -53,6 +53,21 @@
 from gtc.common import ExprKind
 
 
+def _make_literal(v: numbers.Number) -> gtir.Literal:
+    value: Union[BuiltinLiteral, str]
+    if issubclass(type(v), bool):
```
tehrengruber, line 58:
```suggestion
    if isinstance(v, bool):
```

src/gt4py/backend/gtc_backend/defir_to_gtir.py
```diff
@@ -53,6 +53,21 @@
...
+        dtype = common.DataType.BOOL
+        value = common.BuiltInLiteral.TRUE if v else common.BuiltInLiteral.FALSE
+    else:
+        dtype = (
+            common.DataType.INT64
+            if (cast(float, v) / 1.0).is_integer()
+            else common.DataType.FLOAT64
+        )
```
tehrengruber, line 66: I'm not convinced that this is a task that should be carried out in the upper part of the toolchain. Additionally what happens if you get neither float nor int? I would have expected a dict from python dtype to common.DataType here.

src/gtc/common.py
```diff
@@ -317,6 +317,9 @@ class Literal(Node):
     dtype: DataType
     kind: ExprKind = ExprKind.SCALAR
 
+    def __repr__(self) -> str:
+        return str(self.value)
```
tehrengruber, line 321: Just a question: Where is this used?

tehrengruber (changes requested):
The naming scheme for IR nodes you have used is different to everything we have used so far and I think adding an underscore inside an established name to distinguish it is not a good solution to a name clash. Beside that it looks good.

src/gt4py/backend/gtc_backend/cuda/backend.py
jdahm: Fixed. Thanks!

src/gt4py/backend/gtc_backend/defir_to_gtir.py
jdahm: This will be removed once the frontend writes to gtir directly. The `common.DataType` enumeration doesn't support anything else besides boolean, integer, and floating point types. If you can think of a way to check that I'd like to put it in.

src/gtc/common.py
jdahm: I can't tell 😆 . I'll try disabling it run the tests.

src/gt4py/backend/gtc_backend/defir_to_gtir.py
tehrengruber: Not sure I understand the second sentence. What else could one need? I would propose something along these lines:
```
if isinstance(v, int):
  # note: could be extended to support 32 bit integers by checking the values magnitude
  dtype = common.DataType.INT64
elif isinstance(v, float):
  dtype = common.DataType.FLOAT64
else:
  print(f"Warning: Only INT64 and FLOAT64 literals are supported currently. Implicitly upcasting `{v}` to FLOAT64")
  dtype = common.DataType.FLOAT64
```

jdahm:
bors try

jdahm:
bors try-
bors try

jdahm:
bors try

tehrengruber (approved):

PR #675, opened 2022-02-23
Add Gitpod setup

havogt:


.gitpod/aliases.txt
```diff
@@ -0,0 +1 @@
+alias pre-commit="PIP_USER=no pre-commit"
```
egparedes, line 1: I would add a brief comment explaining why this is needed

.gitpod.yml
```diff
@@ -0,0 +1,33 @@
...
+        # enable for pull requests coming from forks (defaults to false)
+    pullRequestsFromForks: true
+        # add a check to pull requests (defaults to true)
+    addCheck: true
+        # add a "Review in Gitpod" button as a comment to pull requests (defaults to false)
+    addComment: false
+        # add a "Review in Gitpod" button to the pull request's description (defaults to false)
+    addBadge: false
+        # add a label once the prebuild is ready to pull requests (defaults to false)
+    addLabel: false
```
egparedes, line 33: Just a suggestion: wouldn't it be nice to activate one of these?

egparedes (approved):

.gitpod.yml
havogt: It is visible under checks. I thought since this is probably only me using it, I didn't want to spam the PR with messages.

PR #683, opened 2022-03-01
Functional minor cosmetics

tehrengruber:
Minor cosmetics I found while working on RawIterator PR

- ~show filename and line number on syntax error~
- improve error message if Python ast node is not supported by dialect
- fix wrong base class of `ProgramSyntaxError`
- fix missing format string prefix

### DIFF
--- a/src/functional/ffront/dialect_parser.py
+++ b/src/functional/ffront/dialect_parser.py
@@ -116,7 +116,7 @@ class DialectParser(ast.NodeVisitor, Generic[DialectRootT]):
     def generic_visit(self, node: ast.AST) -> None:
         raise self._make_syntax_error(
             node,
-            message=f"Nodes of type {type(node).__module__}.{type(node).__name__} not supported in dialect.",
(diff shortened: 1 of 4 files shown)


src/functional/ffront/dialect_parser.py
```diff
@@ -113,6 +113,12 @@ def apply_to_function(
         closure_refs = ClosureRefs.from_function(func)
         return cls.apply(source_definition, closure_refs, externals)
 
+    def generic_visit(self, node: ast.AST) -> None:
+        raise self._make_syntax_error(
+            node,
+            message=f"Nodes of type {type(node).__module__}.{type(node).__name__} not supported in dialect.",
```
egparedes, line 119: See: https://docs.python.org/3.10/glossary.html#term-qualified-name

```suggestion
            message=f"Nodes of type {type(node).__module__}.{type(node).__qualname__} not supported in dialect.",
```

egparedes (approved):
It looks good, except for the suggested change

PR #684, opened 2022-03-03
Refactor exception handling in dialect parsers

tehrengruber:
Mostly as we discussed last week:
- New `RewriteLineNumbers` pass that preprocesses a Python AST Node offseting line numbers or adding them from the parent if not there (e.g. `ast.POW`)
- The PAST and FOAST parser now just throw their respective `SyntaxError`s directly, which will be enhanced by the filename in the `apply` method.


src/functional/ffront/ast_passes/rewrite_lineno.py
```diff
@@ -0,0 +1,72 @@
...
+            node.lineno = self._parent_node.lineno
+            node.col_offset = self._parent_node.col_offset
+            # the end positions are optional according to
+            #  https://docs.python.org/3/library/ast.html#ast.AST.end_col_offset
+            if hasattr(node, "end_lineno"):
+                node.end_lineno = self._parent_node.end_lineno
+            if hasattr(node, "end_col_offset"):
+                node.end_col_offset = self._parent_node.end_col_offset
+
+        return super().generic_visit(node)
```
egparedes, line 72: Can't this Transformer be fully replaced by a simple function using `ast.fix_missing_locations()` (https://docs.python.org/3/library/ast.html#ast.fix_missing_locations) and `ast.ast.increment_lineno()` (https://docs.python.org/3/library/ast.html#ast.increment_lineno)

src/functional/ffront/ast_passes/rewrite_lineno.py
```diff
@@ -0,0 +1,72 @@
...
+    _parent_node: ast.AST
+
+    @classmethod
+    def apply(
+        cls,
+        node: ast.AST,
+        starting_line: int,
+        *,
+        inherit_from_parent: bool = True,
+        inplace: bool = False,
```
egparedes, line 40: Just a suggestion, since utilities in `ast` always mutate nodes in place, I don't think the `inplace` option here is needed. In the rare cases where the caller does not want modifications in place, he can call directly to `copy.deepcopy()`.

src/functional/ffront/dialect_parser.py
```diff
@@ -67,30 +66,27 @@ def apply(
...
         except SyntaxError as err:
-            # TODO(tehrengruber): Instead of enhancing the exception here and
-            #  having source information in the instance we could preprocess
-            #  the ast and offset the line numbers there.
             if not err.filename:
                 err.filename = filename
-            if not isinstance(err, DialectSyntaxError):
-                err.lineno = (err.lineno or 1) + starting_line - 1
+            if not err.text:
+                err.text = source
```
egparedes, line 89: Another suggestion: I don't know if it is clear when or why `err.filename` and `err.text` are expected to be empty, but if it is indeed clear, I would add a comment here to explain it.

egparedes (commented):
The changes look good but I think the rewriter transformer can be implemented in a more simple way using utilities from the `ast` module

src/functional/ffront/ast_passes/rewrite_lineno.py
tehrengruber: As discussed in person `ast.fix_missing_locations` is not sufficient as not all nodes have the source attributes. I had missed `ast.increment_lineno` and switched to using that now.

tehrengruber:
@egparedes Yes. I had to rewrite most of this PR, but it is now ready for another round of review.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -86,7 +86,7 @@ class FieldOperatorTypeDeduction(NodeTranslator):
     >>> sdef = SourceDefinition.from_function(example)
     >>> cref = ClosureRefs.from_function(example)
```
egparedes, line 87:
```suggestion
    >>> src_def = SourceDefinition.from_function(example)
    >>> c_refs = ClosureRefs.from_function(example)
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -86,7 +86,7 @@ class FieldOperatorTypeDeduction(NodeTranslator):
...
     >>> untyped_fieldop = FieldOperatorParser(
-    ...     source=sdef.source, filename=sdef.filename, starting_line=sdef.starting_line, closure_refs=cref, externals_defs={}
+    ...     source_definition=sdef, closure_refs=cref, externals_defs={}
```
egparedes, line 89:
```suggestion
    ...     source_definition=src_def, closure_refs=c_refs, externals_defs={}
```

src/functional/ffront/dialect_parser.py
```diff
@@ -168,6 +154,7 @@ def from_AST(
             lineno=node.lineno,
             offset=node.col_offset,
             filename=filename,
-            end_lineno=node.end_lineno,
-            end_offset=node.end_col_offset,
+            end_lineno=node.end_lineno if hasattr(node, "end_lineno") else None,
+            end_offset=node.end_col_offset if hasattr(node, "end_col_offset") else None,
```
egparedes, line 158:
```suggestion
            end_lineno=getattr(node, "end_lineno", None),
            end_offset=getattr(node, "end_col_offset", None),
```

egparedes (commented):
Looks good. A couple of missing type hints and some style suggestions.

egparedes (approved):
Ready to go.

PR #685, opened 2022-03-04
Functional: Field View Frontend Design ADR update

tehrengruber:
Enhances Field View Frontend Design ADR by PAST dialect design decisions

### DIFF
--- a/docs/functional/architecture/0001-Field_View_Frontend_Design.md
+++ b/docs/functional/architecture/0001-Field_View_Frontend_Design.md
@@ -23,7 +23,7 @@ Overall we tried to avoid:
 And we tried to achieve:
 
 4. Make all error messages that relate to user code maximally useful from the start
-5. Separate parsing (Python -> FOAST) from lowering (FOAST -> Iterator IR)
+5. Separate parsing (Python -> PAST/FOAST) from lowering (PAST/FOAST -> Iterator IR)
 
 In order to achieve that we had to compromise in some aspects like, for example, when it comes to AST / IR nodes having or not having optional attributes.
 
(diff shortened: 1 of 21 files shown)


docs/functional/architecture/0001-Field_View_Frontend_Design.md
```diff
@@ -87,7 +84,38 @@ This step makes use of the type information on expressions, for example to decid
...
+Parses from Python AST into PAST.
+
+#### PAST Lowering
+Lowers from PAST to Iterator IR.
+
+__Field operator out argument slicing__
+The lowering contains some complex logic to emulate slicing of fields in order to restrict the output domain of a field operator call. This contradics guiding principle (3) and should be removed in the future. The concept of specifying the domain for Field operators was not investigated during the frontend design as we concentrated on the beautified iterator dialect where the output domain is explicitly given when lifting a local operator to a field operator. The alternative to also explicitly specify the domain for calls to field operators was rejected as it:
+- would require introducing significantly more syntax in order to specify domains, which was not only infeasible to implement in time, but there also didn't exist an accepted syntax to do so.
+- no intuitive syntax was found to do this
+  - `field_operator(in_field, out=out_field, domain=output_domain)`: Possible, but
```
DropD, line 109: I think after `, but` there was supposed to be something more.

docs/functional/architecture/0001-Field_View_Frontend_Design.md
```diff
@@ -87,7 +84,38 @@ This step makes use of the type information on expressions, for example to decid
...
+
+The Program decorator is not aware of the concept of a FieldOperator but instead uses a generic mechanism (`GTCallable`) to inject functions into the resulting ITIR. This allows to keep the the various decorators independent from each other and avoids coupling the typing system (which is conceptually on a lower level), to them (e.g. `make_symbol_type_from_value` uses the `__gt_type__` method).
+
 #### What could be changed
-##### Ad-hoc pass management
-Since this lowering goes from a statement based AST to a functional IR, all the statements in the AST have to be inlined. This happens in one of the visitor methods currently which calls a FOAST pass, which is perhaps not ideal.
 
+The GTCallable interface it no stable and was born as an ad-hoc concept to fullfil the requirements noted above. In particular the fact that it mixes types, i.e. a concept only found in the frontend, with the untyped iterator IR is not ideal. This leads to the strange situation, where the result of executing the ITIR, i.e. a stencil call, of a Field operator is a value, while the type suggests a Field return value and a call to the decorated function returns nothing.
```
DropD, line 121: typos: `it no stable` -> `is not stable`, `fullfil` -> `fulfill`. Probably should run a spellchecker.

DropD (changes requested):

docs/functional/architecture/0001-Field_View_Frontend_Design.md
```diff
@@ -35,26 +35,32 @@ We tried to avoid (2) partially to avoid merge conflicts, by keeping the logic i
...
+##### Correspondence Python AST <-> dialect AST
+In some cases there are Python constructs which would directly map to a function call in the iterator model (such as binary operators). It was chosen to still represent them as binary operators to keep the analogy to Python AST and clearly separate parsing and lowering. The correspondence between Python and dialect AST also means it would be relatively easy to "unparse" the dialect AST into valid Python code for debugging.
 
-##### Symbol types are not FOAST nodes
-The symbol types in FOAST are implemented as simple dataclasses to help reusability between different IRs, since they are just _leaf_ nodes (not recursed into during tree traverslas) describing which kind of symbol is defined in an IR node.
+##### Symbol types are not dialect nodes
+The symbol types in all dialects are implemented as simple dataclasses to ensure compatibility and reusability between different IRs. They are just _leaf_ nodes (not recursed into during tree traverslas) describing which kind of symbol is defined in an IR node.
```
egparedes, line 56:
```suggestion
The symbol types in all dialects are implemented as simple dataclasses to ensure compatibility and reusability between different IRs. They are just _leaf_ nodes (not recursed into during tree traversals) describing which kind of symbol is defined in an IR node.
```

docs/functional/architecture/0001-Field_View_Frontend_Design.md
```diff
@@ -87,7 +84,38 @@ This step makes use of the type information on expressions, for example to decid
 ##### Assume FOAST is correct
 As per guiding principle (5), the lowering should not worry if the FOAST is incorrect or invalid. It is the responsibility of the previous parser and passes to sanitize the user code.
 
+### Program AST (PAST)
+
+The Program AST is a stateful dialect whose main purpose is to apply operators to fields and storing the result on another field. In the future it could be extended to also support control flow (loops, conditionals).
+
+#### What could be changed
+
+The syntax to call field operators feels unnatural as their signature suggests they return a Field, but instead after decorating they return nothing and have an additional argument `out`.
```
egparedes, line 93: Just a question: do they really return nothing? In NumPy, a reference to the output value is return even if the `out` parameter is specified.

docs/functional/architecture/0001-Field_View_Frontend_Design.md
```diff
@@ -87,7 +84,38 @@ This step makes use of the type information on expressions, for example to decid
...
+- would require introducing significantly more syntax in order to specify domains, which was not only infeasible to implement in time, but there also didn't exist an accepted syntax to do so.
+- no intuitive syntax was found to do this
+  - `field_operator(in_field, out=out_field, domain=output_domain)`: Possible, but 
+  - `field_operator[output_domain](in_field, out=out_field)` This syntax is already used to create a field operator from a local operator.
+
+### Decorators (@program, @field_operator)
+
+#### What to keep
+The decorators are designed to be constructed from the root node of their respective dialect contrary to a python function object. This decision has been made to allow programmatically generating instances thereof.
```
egparedes, line 115:
```suggestion
The decorators are designed to be constructed from the root node of their respective dialect contrary to a Python function object. This decision has been made to allow programmatically generating instances thereof.
```

docs/functional/architecture/0001-Field_View_Frontend_Design.md
```diff
@@ -87,7 +84,38 @@ This step makes use of the type information on expressions, for example to decid
...
+The PAST is an independent dialect and (code-wise) only shares symbol types with the FOAST. This was a deliberate choice as there is currently no good way to share nodes and we wanted to allow changing each of the dialects independently until they become stable. Nonetheless nodes that exist in both dialects should stay in sync so that they could be generalized in the future.
+
+#### PAST parser
+Parses from Python AST into PAST.
+
+#### PAST Lowering
+Lowers from PAST to Iterator IR.
+
+__Field operator out argument slicing__
+The lowering contains some complex logic to emulate slicing of fields in order to restrict the output domain of a field operator call. This contradics guiding principle (3) and should be removed in the future. The concept of specifying the domain for Field operators was not investigated during the frontend design as we concentrated on the beautified iterator dialect where the output domain is explicitly given when lifting a local operator to a field operator. The alternative to also explicitly specify the domain for calls to field operators was rejected as it:
```
egparedes, line 106: I might be wrong but I think this is not completely accurate: a field operator could return multiple fields with different domains, so the main problem to specify the output domain in field view is not the syntax, but the fact that each output field could be defined in a different domain. 

Besides that, I agree that the syntax could be further discussed and improved. For example, an alternative strategy just thought about right now could be to use a special marker to allow in place updates of the fields with assignments syntax (a bit like `...` is used in NumPy for a full replace of the array content)

```python
out_a[IN_PLACE], out_b[IN_PLACE] = field_op(a, b)
```

egparedes (commented):
Just minor typos and one clarification.

docs/functional/architecture/0001-Field_View_Frontend_Design.md
tehrengruber: Yes, the reasoning was that `Program`s also return nothing. We could change that easily though. What do you think?

docs/functional/architecture/0001-Field_View_Frontend_Design.md
tehrengruber: With the current implementation of the iterator IR a stencil closure has a single domain and as such a field operator when called at the program level can only return fields with a single output domain. This is one of my critic points of the model.

With respect to the syntax, what speaks against using `...`?

egparedes: > With the current implementation of the iterator IR a stencil closure has a single domain and as such a field operator when called at the program level can only return fields with a single output domain. This is one of my critic points of the model.

You're right about the current implementation of the lowering from field view, although theoretically I guess it could be done differently to support different output domains per field in a (hypothetical) future.

> With respect to the syntax, what speaks against using `...`? 

I would argue that the accepted meaning of `...` is overwrite the **full** content of the field and we want something to indicate that only the matching domain should be update. Anyway, this is not important for now.

docs/functional/architecture/0001-Field_View_Frontend_Design.md
egparedes: If it's easy to change and doesn't conflict with other design decisions, I would return the reference to the output fields since it feels more coherent and matches NumPy behavior.

egparedes (approved):
It looks good.

docs/functional/architecture/0001-Field_View_Frontend_Design.md
tehrengruber: Ah, I see what you mean with the `IN_PLACE` now. I would argue that the domain as specified using the slice and the domain as given by the field returned from the field operator need to agree and that there is no real need to inherit the domain from the returned field (the same way as this is not needed in numpy). Having `...` overwrite the full content of the field is what I intended it to mean and I believe this is the common case for unstructured and also for cartesian if you handle the boundary conditions inside a field operator.

docs/functional/architecture/0001-Field_View_Frontend_Design.md
tehrengruber: Should be easy, I'll implement it later.

PR #687, opened 2022-03-10
functional: Regression test for Cartesian Laplacian field view

havogt:
Adds a Laplacian example in field view.

Fixes:
- allow offset with negative index
- workaround (to be discussed): if type of rhs or lhs of binop is not deduced (`None`), propagate a `None`.

Additional:
- moved filename and line sensitive test to separate file

### DIFF
--- a/src/functional/ffront/foast_passes/type_deduction.py
+++ b/src/functional/ffront/foast_passes/type_deduction.py
@@ -215,6 +215,11 @@ class FieldOperatorTypeDeduction(NodeTranslator):
         **kwargs,
     ) -> ct.SymbolType:
         left, right = TypeInfo(left_type), TypeInfo(right_type)
+
+        # TODO decide on if we support non-strict type checking
+        if left.type is None or right.type is None:
+            return None
+
         if (
             left.is_arithmetic_compatible
             and right.is_arithmetic_compatible
--- a/src/functional/ffront/foast_passes/type_deduction.py
+++ b/src/functional/ffront/foast_passes/type_deduction.py
@@ -216,7 +216,7 @@ class FieldOperatorTypeDeduction(NodeTranslator):
     ) -> ct.SymbolType:
         left, right = TypeInfo(left_type), TypeInfo(right_type)
 
-        # TODO decide on if we support non-strict type checking
+        # if one type is `None` (not deduced, generic), we propagate `None`
         if left.type is None or right.type is None:
             return None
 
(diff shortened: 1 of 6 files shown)


src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -215,6 +215,11 @@ def _deduce_arithmetic_binop_type(
         **kwargs,
     ) -> ct.SymbolType:
         left, right = TypeInfo(left_type), TypeInfo(right_type)
+
+        # TODO decide on if we support non-strict type checking
```
havogt, line 219:
```suggestion
        # if one type is `None` (not deduced, generic), we propagate `None`
```

tests/functional_tests/ffront_tests/test_error_line_number.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from functional.common import Field
+from functional.ffront.func_to_foast import FieldOperatorParser, FieldOperatorSyntaxError
+
+
+# NOTE: This test tests includes the current filename and the line number of the marked statement
```
havogt, line 7:
```suggestion
# NOTE: This test is sensitive to filename and the line number of the marked statement
```

tests/functional_tests/ffront_tests/test_error_line_number.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from functional.common import Field
+from functional.ffront.func_to_foast import FieldOperatorParser, FieldOperatorSyntaxError
+
+
+# NOTE: This test is sensitive to filename and the line number of the marked statement
```
DropD, line 7: Good solution.

tests/functional_tests/ffront_tests/test_interface.py
```diff
@@ -216,6 +200,15 @@ def bool_or(a: Field[..., "bool"], b: Field[..., "bool"]):
...
 
 
+def test_shift():
+    Offset = 0  # for parsing it doesn't matter what this object is
+
+    def shifts(a: Field[..., "bool"]):
+        return a(Offset[1]) + a(Offset[-1])  # TODO we should test these cases separate
+
+    _ = FieldOperatorParser.apply_to_function(shifts)
+
```
DropD, line 210: This test would pass even if the parser were to simply ignore and drop offsets. Ideally we would also add a negative offset case to `test_lowering.test_shift`. But if we have that, we don't really need this one.

DropD (changes requested):

tests/functional_tests/ffront_tests/test_interface.py
havogt: removed in favor of a lowering test

DropD (approved):

PR #689, opened 2022-03-15
Remove FillFlushToLocalKCaches

jdahm:
## Description

Although this pass is disabled by default in the gtcpp backends, the pass causes issues when used with it. Therefore, it should be disabled from the default pipeline and added explicitly to gtc:cuda.

### DIFF
--- a/src/gtc/passes/oir_pipeline.py
+++ b/src/gtc/passes/oir_pipeline.py
@@ -20,7 +20,6 @@ from typing import Callable, Optional, Protocol, Sequence, Type, Union
 from eve.visitors import NodeVisitor
 from gtc import oir
 from gtc.passes.oir_optimizations.caches import (
-    FillFlushToLocalKCaches,
     IJCacheDetection,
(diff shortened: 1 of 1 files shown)


gronerl:
bors try

src/gt4py/backend/gtc_backend/cuda/backend.py
```diff
@@ -53,6 +54,7 @@ def __call__(self, definition_ir) -> Dict[str, Dict[str, str]]:
             "oir_pipeline", DefaultPipeline(skip=[NoFieldAccessPruning])
         )
         oir = oir_pipeline.run(base_oir)
+        oir = FillFlushToLocalKCaches().visit(oir)
```
gronerl, line 57: Thats the non-overengineered way of doing it :) And I was talking about changing the pipeline infrastructure :man_facepalming:

gronerl (approved):
Looking good!

PR #693, opened 2022-03-18
Use `bool` instead of `np.bool`

jdahm:
## Description

Uses `bool` instead of `np.bool`. Resolves the following warning:
```
DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    BooleanTypes = (bool, np.bool,)
```


havogt (approved):
Looks good. I assume your implmentation trys to work also for potential similar changes for other datatypes, that's why you didn't get explicit on bool?

PR #695, opened 2022-03-21
Allow slices on gt4py storages

jdahm:
## Description

Follow-on to #576. Shoves the getitem index into a storage attribute so that the mask can be computed when the reduced storage is finalized.


tehrengruber (approved):
I still stand to my opinion that this approach towards storages is inherently broken given how numpy works. What you are doing here looks very nice on the syntactic level, but treating storages like numpy arrays has been a major pain in the past and I believe it is not worth it (see the initial PR for proposed solutions). Nonetheless this is perfectly in line with the design of the storages, so I won't stand in the way of this PR.

jdahm:
bors try

src/gt4py/storage/storage.py
```diff
@@ -212,13 +214,17 @@ def __array_finalize__(self, obj):
...
-                    )
                 self.__dict__ = {**obj.__dict__, **self.__dict__}
                 self.is_stencil_view = False
+                if hasattr(obj, "_new_index"):
+                    index_iter = itertools.chain(
+                        obj._new_index, [slice(None, None)] * (len(obj.mask) - len(obj._new_index))
+                    )
+                    sliced_indices = (isinstance(x, slice) for x in index_iter)
+                    self._mask = obj.mask and gt_utils.interpolate_mask(
+                        sliced_indices, obj.mask, False
```
egparedes, line 225: One question: why is `and` used here? This is **not** doing an element-wise `and` operation between the components of the current mask and the interpolated mask

jdahm: oh, I experimented with `and` and I thought it did do an element-wise `and`. I'll double check the behavior today!

jdahm:
bors try-
bors try

PR #696, opened 2022-03-21
Fix errors caught with updated mypy

jdahm:
## Description

An update to pre-commit hooks caught these issues.


havogt:
addressed in #694

PR #698, opened 2022-03-23
Remove legacy backends

jdahm:
## Description

Removes legacy backends and test files. Part of #566.

Blocked by #673.


jdahm:
No, but I haven't tested this branch with GPUs on Daint yet. I'll do that today.

egparedes:
bors try

egparedes (approved):
It looks good to me, but just a question: are the files in `backend/templates` still needed or could they be also removed in this PR?

jdahm:
I think `stencil_module.py.in` is still used, but the others I believe can be deleted. Good catch. I'll try that today.

jdahm:
@egparedes Can you re-review? I:

- Moved `location_to_source_location`
- Removed `USE_LEGACY_BACKEND`
- Removes `gt4py.ir` and moves source into `gt4py.frontend`

egparedes (commented):
Thanks for the extra changes. I suggest the following changes:

- rename the new `utils` in `frontend` to `node_utils` or something like that, which I think is more representative of the contents of the module
- delete `computation.src.in` from `backend/templates` which I think it's not used at all
- Update `frontend.nodes` with more suitable docstrings and remove Implementation IR nodes, which are not longer used.

jdahm:
bors try

PR #701, opened 2022-03-25
Use centralized extent analysis in dace

gronerl:
This cleans up the extent analysis in the DaCe backend and relies on existing/recently merged utilities.  

This helps avoid a large part of the diff to the gtc:dace backend for the horizontal regions.


jdahm (approved):
Looks good to me! The only suggestion I have is perhaps in the future any time we touch a function in the dace backend to add type hints.

gronerl:
bors try

PR #702, opened 2022-03-27
Field operator composition

### LINKED ISSUE #690: functional: field operator composition
The following doesn't work, because the `lap` function gets lost in this function https://github.com/GridTools/gt4py/blob/dcd927cd46d9c4cb097f09a0c4582336073c6da1/src/functional/ffront/decorator.py#L123

Hacking with something like
```python
        lowered_funcs = [vars_[name].__gt_itir__() for name in func_names]
        for v in vars_.values():
            for f in v.closure_refs.globals.values():
                lowered_funcs.append(f.__gt_itir__())
```
fixes the issue.

Example code
```python
@field_operator
def lap(
    in_field: Field[[IDim, JDim], "float"], a4: Field[[IDim, JDim], "float"]
) -> Field[[IDim, JDim], "float"]:
    return (
        a4 * in_field
        + in_field(Ioff[1])
        + in_field(Joff[1])
        + in_field(Ioff[-1])
        + in_field(Joff[-1])
    )


@field_operator
def laplap(
    in_field: Field[[IDim, JDim], "float"], a4: Field[[IDim, JDim], "float"]
) -> Field[[IDim, JDim], "float"]:
    return lap(lap(in_field, a4), a4)
```

Should be included in shaping (if it's not trivial to fix).

tehrengruber:
Fixes #690


src/functional/ffront/decorator.py
```diff
@@ -49,6 +50,9 @@ class GTCallable(Protocol):
     from ``ffront`` programs or operators.
     """
 
+    def __gt_closure_refs__(self) -> Optional[ClosureRefs]:
```
egparedes, line 53: Missing docstring

src/functional/ffront/decorator.py
```diff
@@ -119,6 +123,31 @@ def from_function(
             definition=definition,
         )
 
+    def _lowered_funcs_from_closureref(
+        self, closure_refs: ClosureRefs
+    ) -> list[itir.FunctionDefinition]:
+        lowered_funcs = []
+
+        vars_ = collections.ChainMap(closure_refs.globals, closure_refs.nonlocals)
+        for name, val in vars_.items():
```
egparedes, line 132: Just a suggestion, I think using `refs` instead of `vars_` and `value` instead of `val` would make it easier to read.

egparedes (commented):
It looks good. Just a missing docstring for a new element of a public interface.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -298,8 +300,17 @@ def visit_Call(self, node: foast.Call, **kwargs) -> foast.Call:
...
             return foast.Call(func=new_func, args=new_args, location=node.location, type=new_type)
-        return foast.Call(
-            func=new_func, args=self.visit(node.args, **kwargs), location=node.location
+        elif isinstance(new_func.type, ct.FunctionType):
+            return foast.Call(
+                func=new_func,
+                args=self.visit(node.args, **kwargs),
+                type=new_func.type.returns,
+                location=node.location,
+            )
```
egparedes, line 309: The type deduction part might be out of sync with the recently merged PR adding reductions and builtin functions. Please, take a look before merging: https://github.com/GridTools/gt4py/pull/640/files#diff-5a2d8f3beae7064a7845cfddc917538e55d53d9d3694a978d2e8b2c5a72ab3ae

tehrengruber: I checked and this still works. Nonetheless the type deduction of FOAST is incomplete in the sense that not all types are deduced (for sure after the reductions PR). We should fix this at some point, but this requires either a function to compute the resulting type of a builtin (not so hard) or a generic algorithm and type parameters (nice, but more work).

src/functional/ffront/decorator.py
tehrengruber: I have left `vars_` unchanged which is in line with the rename in https://github.com/GridTools/gt4py/pull/705, but changed `val` to `value`.

src/functional/ffront/decorator.py
```diff
@@ -119,6 +129,33 @@ def from_function(
             definition=definition,
         )
 
+    def _lowered_funcs_from_closureref(
```
egparedes, line 132: `_lowered_funcs_from_closureref` looks like the wrong name.

src/functional/ffront/decorator.py
```diff
@@ -119,6 +129,33 @@ def from_function(
...
+        self, captured_vars: CapturedVars
+    ) -> list[itir.FunctionDefinition]:
+        lowered_funcs = []
+
+        vars_ = collections.ChainMap(captured_vars.globals, captured_vars.nonlocals)
+        for name, value in vars_.items():
+            # With respect to the frontend offsets are singleton types, i.e.
+            #  they do not store any runtime information, but only type
+            #  information. As such we do not need their value.
+            if isinstance(value, Offset):
```
egparedes, line 142: Is `functional.iterator.runtime.Offset` the right type here? Shouldn't it be `fbuiltins.FieldOffset`?

(Check: https://github.com/GridTools/gt4py/pull/640/files#diff-5a2d8f3beae7064a7845cfddc917538e55d53d9d3694a978d2e8b2c5a72ab3ae)

egparedes (commented):
There are a couple of minor points to fix.

egparedes (approved):
It looks good.

PR #703, opened 2022-03-28
Fix invalid use of identity comparison

tehrengruber:
I am not completely certain about the exact circumstances on why this only fails in the CI and only when pytest is run from the root directory of the GT4Py folder, but apparently parameterized types from `typing` are no singletons (or some other library is messing around in python internals). As a result some comparison using `is` instead of `==` is failing, which is fixed by this PR.

### DIFF
--- a/src/gtc/common.py
+++ b/src/gtc/common.py
@@ -658,8 +658,11 @@ class _LvalueDimsValidator(NodeVisitor):
             raise ValueError(
                 f"Vertical loop type {vertical_loop_type} has no `loop_order` attribute"
             )
-        if not decl_type.__annotations__.get("dimensions") is Tuple[bool, bool, bool]:
-            raise ValueError(f"Field decl type {decl_type} has no `dimensions` attribute")
+        if not decl_type.__annotations__.get("dimensions") == Tuple[bool, bool, bool]:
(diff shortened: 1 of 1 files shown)


gronerl (approved):
I think I have seen this before :thinking: Certainly `is` is not the right thing here.

egparedes:
> parameterized types from `typing` are no singletons 

At run time, a "parameterized type hint" is not a singleton, but just a `typing._GenericAlias` instance with references to the generic type and the type arguments used in the instantiation. The `typing` module has an internal cache (`_tp_cache`, just a `wrapper` around `functools.lru_cache`) to try to reuse the `_GenericAlias` instance when the same generic type is instantiated with the same arguments, but this is just an optimization and it's not guaranteed to succeed. 

To sum up, @gronerl is right and `is` shouldn't be used in general to compare type parametrized type hints.

tehrengruber:
bors try

jdahm:
Thank you for tracking this down! That was going to be on my list to tackle today had you not already found it :).

jdahm:
bors try

tehrengruber:
Not sure what the CI is up to again. Merging anyway.

PR #704, opened 2022-03-28
Bump black version

jdahm:
CI is broken due to a click update. This works around the issue. Related to https://github.com/psf/black/issues/2964#issuecomment-1080974737=.


havogt:
This is now obsolete as black released a fix. We should bump black to 22.3.0 instead.

PR #705, opened 2022-03-29
Functional: Rename `ClosureRefs` to `CapturedVars`

tehrengruber:
This PR renames the `ClosureRefs` datastructure in the functional frontend to `CapturedVars`. The name `ClosureRefs` is not optimal (in my opinion) as the noun reference has a different meaning in computer science that is not meant here. The datastructure holds essentially the same information as `inspect.getclosurevars` which returns a named tuple `ClosureVars`. To avoid a name collision, the name `CapturedVars` was choosen (expressing that it holds the variables captured inside a function). Additionally the `closure` attribute of `past.Program` and `foast.FieldOperator` has been renamed, both because it does not actually store a closure and to reflect the change.

### DIFF
--- a/src/functional/ffront/foast_passes/type_deduction.py
+++ b/src/functional/ffront/foast_passes/type_deduction.py
@@ -110,7 +110,7 @@ class FieldOperatorTypeDeduction(NodeTranslator):
             id=node.id,
             params=self.visit(node.params, **kwargs),
             body=self.visit(node.body, **kwargs),
-            closure=self.visit(node.closure, **kwargs),
+            captured_vars=self.visit(node.captured_vars, **kwargs),
             location=node.location,
         )
 
--- a/src/functional/ffront/foast_passes/type_deduction.py
+++ b/src/functional/ffront/foast_passes/type_deduction.py
@@ -83,11 +83,15 @@ class FieldOperatorTypeDeduction(NodeTranslator):
     >>> def example(a: "Field[..., float]", b: "Field[..., float]"):
     ...     return a + b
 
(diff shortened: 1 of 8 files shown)


src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -78,15 +78,15 @@ class FieldOperatorTypeDeduction(NodeTranslator):
...
     >>> from functional.common import Field
-    >>> from functional.ffront.source_utils import SourceDefinition, ClosureRefs
+    >>> from functional.ffront.source_utils import SourceDefinition, CapturedVars
     >>> from functional.ffront.func_to_foast import FieldOperatorParser
     >>> def example(a: "Field[..., float]", b: "Field[..., float]"):
     ...     return a + b
 
     >>> sdef = SourceDefinition.from_function(example)
-    >>> cref = ClosureRefs.from_function(example)
+    >>> cref = CapturedVars.from_function(example)
```
egparedes, line 87: Make sure to handle a potential conflict with the `cref` name, since it has been also renamed in other PR.
```suggestion
    >>> c_vars = CapturedVars.from_function(example)
```

egparedes (commented):
Please add a description to the PR and fix a missing variable renaming in the docstring examples.

tehrengruber:
I tried to keep the overhead low and hence didn't write a description. I have added one now so that everyone can follow the motive. Only change since last review is that I renamed the `closure` attribute in both `past.Program` and `foast.FieldOperator`.

egparedes (approved):
It looks good to me.

PR #707, opened 2022-03-31
Functional: Type Inference & Checking

fthaler:
## Description

Constraint-based type inference (and type checking) for the functional IR. Currently supports all builtins and IR features apart from `is_none` and optional/maybe types (which need more discussion). Adds a type-checking backend ("type_check") which is enabled in all tests.

## Requirements


DropD (approved):
I won't pretend to have looked very closely, but from what I can tell from the tests and the talk about this topic it looks like it is doing what it should.

fthaler:
> I won't pretend to have looked very closely, but from what I can tell from the tests and the talk about this topic it looks like it is doing what it should.

I am going to port this to the new datamodels (just use dataclasses so far to avoid the Pydantic quirks). So that’s probably one of the PRs we should merge after moving the repo.

fthaler:
Superseded by #811

PR #708, opened 2022-04-01
Functional: Fix Unsupported Python Version in CI

fthaler:
## Description

Python 3.10.2 is not anymore available.

## Requirements

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -14,9 +14,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v3
       with:
-        python-version: 3.10
+        python-version: "3.10"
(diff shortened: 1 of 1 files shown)


havogt:
@egparedes was there a reason to pin to a patch version?

egparedes:
I think it was pinned to 3.10.0 in the first commit of the CI actions (maybe by @DropD ??, I'm not sure) for the `functional` branch and later updated to 3.10.1 and 3.10.2 whenever some error from that version was breaking CI or pre-commit. If it now works fine, I think it makes sense to avoid pinning the patch version.

fthaler:
Can we merge this as soon as possible? This issue breaks the whole functional CI, so would be nice if it would not stay open for too long… I don’t see any reason why the version is fixed to a specific release. This is only done for the pre-commit action, but not for the Tox action.

egparedes (approved):
It looks good.

fthaler:
Thanks!

PR #713, opened 2022-04-05
Add `DaCeFrozenStencil` and `DaCeStencilObject` with support for  `__sdfg__` interface.

gronerl:
This PR enables gt4py stencils to be called from `@dace.program` decorated functions.


src/gt4py/backend/gtc_backend/dace/backend.py
```diff
@@ -348,6 +350,36 @@ def apply(cls, gtir: gtir.Stencil, sdfg: dace.SDFG, module_name: str, *, backend
...
+_sdfg = None
+
+def __new__(cls, *args, **kwargs):
+    res = super().__new__(cls, *args, **kwargs)
+    cls._sdfg = dace.SDFG.from_file('{filepath}')
+    return res
+
+@property
+def sdfg(self) -> dace.SDFG:
+    return copy.deepcopy(self._sdfg)
```
jdahm, line 375: You're a better judge of this than I am not having written the code, but from the outside it seems non-performant to `deepcopy` a potentially huge nested structure every time a property is accessed.

src/gt4py/stencil_object.py
```diff
@@ -47,7 +58,7 @@ def _compute_cache_key(field_args, parameter_args, domain, origin) -> int:
 
 
 @dataclass(frozen=True)
-class FrozenStencil:
+class FrozenStencil(SDFGConvertible):
```
jdahm, line 61: Instead of adding all these attributes directly to `gtscript.StencilObject` and `gtscript.FrozenStencil`, I have two options:

1. I wonder if we could add a wrapper class on top of this that needs to be called if one wants to orchestrate. Something like 
```
obj = gtscript.stencil(backend="gtc:numpy", definition=test_def, **backend_kwargs) # if not using dace orchestration
...
obj = dace_util.wrapper(gtscript.stencil(...)) # if using dace orchestration
```

2. Add these in the module itself generated from the `gtc:dace` backend.

src/gt4py/stencil_object.py
```diff
@@ -488,19 +699,13 @@ def _call_run(
         if exec_info is not None:
             exec_info["call_run_start_time"] = time.perf_counter()
 
-        cache_key = _compute_cache_key(field_args, parameter_args, domain, origin)
-        if cache_key not in self._domain_origin_cache:
-            origin = self._normalize_origins(field_args, origin)
+        origin = self._normalize_origins(field_args, origin)
```
jdahm, line 702: Looks like there might have been a merge conflict here.

src/gt4py/stencil_object.py
```diff
@@ -404,10 +614,10 @@ def _validate_args(  # noqa: C901  # Function is too complex
 
         # assert compatibility of parameters with stencil
         for name, parameter_info in self.parameter_info.items():
-            if parameter_info.access != AccessKind.NONE:
+            if parameter_info is not None:
                 if name not in param_args:
                     raise ValueError(f"Missing value for '{name}' parameter.")
-                elif type(parameter := param_args[name]) != parameter_info.dtype:
+                if type(parameter := param_args[name]) != parameter_info.dtype:
```
jdahm, line 620: Double check this.

src/gt4py/stencil_object.py
gronerl: 3. Made an intermediate level of inheritance that adds the DaCe support.

src/gt4py/backend/gtc_backend/dace/backend.py
gronerl: Yes, they are a pipeline performance issue. However, it's currently unsafe to use the same graph instance in wrapping. Tal is looking into improving this.

src/gt4py/stencil_object.py
gronerl: How do you feel about still having the `__sdfg__` in the `StencilObject` just for raising errors? @tbennun  Do you see another way of printing a meaningful error? Currently, if we don't put it there, DaCe will try to parse the StencilObject.

tests/test_integration/test_dace_parsing.py
```diff
@@ -0,0 +1,248 @@
...
+        frozen_stencil(inp=inp, outp=outp)
+
+    call_frozen_stencil()
+
+    assert np.allclose(inp, 7.0)
+    assert np.allclose(
+        outp[
+            2:5,
+            2:5,
+            :,
```
jdahm, line 139: Remove `,` to fold these lines together.

jdahm, line 139:
```suggestion
            :
```

tests/test_integration/test_dace_parsing.py
```diff
@@ -0,0 +1,248 @@
...
+        frozen_stencil(inp=inp, unused_field=unused_field, outp=outp, unused_par=7.0)
+
+    call_frozen_stencil()
+
+    assert np.allclose(inp, 7.0)
+    assert np.allclose(
+        outp[
+            2:5,
+            2:5,
+            :,
```
jdahm, line 176:
```suggestion
            :
```

tests/test_integration/test_dace_parsing.py
```diff
@@ -0,0 +1,248 @@
...
+
+    @dace.function
+    def call_frozen_stencil():
+        defn(outp, par=inp)
+
+    call_frozen_stencil()
+
+    assert np.allclose(
+        outp,
+        7.0,
```
jdahm, line 50:
```suggestion
        7.0
```

tests/test_integration/test_dace_parsing.py
```diff
@@ -0,0 +1,248 @@
...
+
+    outp = np.zeros(dtype=np.float64, shape=(10, 10, 10))
+
+    inp = 7.0
+
+    @dace.function
+    def call_frozen_stencil():
+        defn(outp, par=inp)
+
+    call_frozen_stencil()
```
jdahm, line 46: This doesn't seem to be a `FrozenStencil` 😄

src/gt4py/backend/module_generator.py
```diff
@@ -140,6 +140,7 @@ def __call__(
             imports=self.generate_imports(),
             module_members=self.generate_module_members(),
             class_name=self.generate_class_name(),
+            base_class=self.generate_base_class_name(),
```
jdahm, line 143: 👍

tests/test_integration/test_dace_parsing.py
```diff
@@ -0,0 +1,248 @@
...
+    )
+
+
+def test_basic():
+    @gtscript.stencil(backend="gtc:dace")
+    def defn(outp: gtscript.Field[np.float64], par: np.float64):
+        with computation(PARALLEL), interval(...):
+            outp = par  # noqa F841: local variable 'outp' is assigned to but never used
+
+    outp = np.zeros(dtype=np.float64, shape=(10, 10, 10))
```
jdahm, line 38: Should these tests use both `numpy` storages as well as `gt4py.storage.Storage`?

src/gt4py/stencil_object.py
tbennun: so what exactly is happening? is it trying to get `StencilObject.__sdfg__` instead of the inherited class?

gronerl: No it all works fine. Just in the case of non-dace backends, we want to raise a meaningful error, currently we add `__sdfg__` etc to the StencilObject and override them in the dace backends. I was wondering if there was a way to keep the StencilObject free of all things dace. This works fine though, we can keep it.

tests/test_integration/test_dace_parsing.py
gronerl: I added one test that uses storages instead. DaCe doesn't like that storages are a view of the underlying ndarray, so we have to override `TypeError: Passing a numpy view (e.g., sub-array or "A.T") to DaCe programs is not allowed in order to retain analyzability. Please make a copy with "numpy.copy(...)". If you know what you are doing, you can override this error in the configuration by setting compiler.allow_view_arguments to True.`

src/gt4py/stencil_object.py
gronerl: @jdahm you deleted some message I think, do you think they should be NotImplementedError ? To me those usually mean they are not implemented _yet_ but we have no intention of supporting it in the future.

src/gt4py/testing/input_strategies.py
```diff
@@ -150,6 +150,7 @@ def scalar_value_st(dtype, min_value, max_value, allow_nan=False):
             max_value,
             allow_infinity=allow_infinity,
             allow_nan=allow_nan,
+            allow_subnormal=False,
```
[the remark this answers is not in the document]
gronerl: I'm not sure I put it due to an apparently unrelated error in the test suites. I reverted it now to see if it was just a hypothesis fluke.

gronerl:
Acutally the error message was right!
`E           FloatingPointError: Got allow_subnormal=True, but we can't represent subnormal floats right now, in violation of the IEEE-754 floating-point specification.  This is usually because something was compiled with -ffast-math or a similar option, which sets global processor state.  See https://simonbyrne.github.io/notes/fastmath/ for a more detailed writeup - and good luck!`
Gotta love the cynical "and good luck" part.

src/gt4py/testing/input_strategies.py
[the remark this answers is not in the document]
gronerl: See comment in main conversation thread. I reverted.

src/gt4py/stencil_object.py
jdahm: Sorry, I deleted it after I thought about it some more. The `TypeError` stubs you have are great!

gronerl:
bors try

gronerl:
bors try

tests/test_integration/test_dace_parsing.py
tbennun: You can resolve that by overriding the configuration. We will improve the condition to only occur on truly unsafe view arguments (i.e., mismatching strides, as opposed to the offsetted arrays which we are seeing here)

jdahm: > `TypeError: Passing a numpy view (e.g., sub-array or "A.T") to DaCe programs is not allowed in order to retain analyzability. Please make a copy with "numpy.copy(...)". If you know what you are doing, you can override this error in the configuration by setting compiler.allow_view_arguments to True.`

Ah, I understand, but it concerns me and points to a code smell/issue that needs to be addressed.

`numpy.ndarray` are not fully supported as arguments to stencils (since they lack metadata like `default_origin`, `mask`, `layout`, etc.), so stating "use numpy arrays with dace" essentially amounts to throwing away the carefully-designed launch API of gt4py and relying solely on frozen stencils (which are derived from the `StencilObject`s). In my opinion, frozen stencils should be employed as an optimization over the regular stencil interface and hidden in application utility layers, not used as a top-level interface.

I have one idea: Can this exchange from `Storage` to `ndarray` (probably just taking the `.data` attribute) be done in the `DaceStencilObject`?

cc: @havogt @tbennun

jdahm: @tbennun it seems like our comments stepped on each other. Glad to hear it there's a way around it.

gronerl: What Tal suggests is exactly what line 253 in test_dace_parsing.py does. I will change all the tests to work with storages, there's no point testing it with numpy at all.

jdahm (approved):
Thanks for hanging in there with the code changes. This looks great now!

gronerl:
bors try

Issue #714, opened 2022-04-05
Load path issue on Piz Daint with Pace


jdahm:
In the Pace test suite on daint we are consistently getting the error below, even after setting `GT_CACHE_LOAD_RETRIES=100`, `GT_CACHE_LOAD_RETRY_DELAY=1000`. This makes me think the problem is an issue with appending, loading, and quickly removing from `sys.path`. I'm looking into alternatives, including using `importlib` lower-level code here as well.

<details><summary>E   ModuleNotFoundError: No module named 'm_computation__gtcnumpy_23f1248fae'</summary>

```
../external/gt4py/src/gt4py/gtscript.py:288: in stencil
    return _decorator(definition)
../external/gt4py/src/gt4py/gtscript.py:276: in _decorator
    out = gt_loader.gtscript_loader(
../external/gt4py/src/gt4py/loader.py:72: in gtscript_loader
    stencil_class = load_stencil("gtscript", backend, definition_func, externals, build_options)
../external/gt4py/src/gt4py/loader.py:58: in load_stencil
    return builder.build()
../external/gt4py/src/gt4py/stencil_builder.py:86: in build
    stencil_class = None if self.options.rebuild else self.backend.load()
../external/gt4py/src/gt4py/backend/base.py:256: in load
    stencil_class = self._load()
../external/gt4py/src/gt4py/backend/base.py:270: in _load
    stencil_module = gt_utils.make_module_from_file(stencil_class_name, file_name)
../external/gt4py/src/gt4py/utils/base.py:325: in make_module_from_file
    return load()
../external/gt4py/src/gt4py/utils/base.py:299: in load
    spec.loader.exec_module(module)
<frozen importlib._bootstrap_external>:783: in exec_module
    ???
<frozen importlib._bootstrap>:219: in _call_with_frames_removed
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    import pathlib
    import time
    
    import numpy as np
    from numpy import dtype
    import sys
    import pathlib
    import numpy
    path_backup = sys.path.copy()
    sys.path.append(str(pathlib.Path(__file__).parent))
>   import m_computation__gtcnumpy_23f1248fae as computation
E   ModuleNotFoundError: No module named 'm_computation__gtcnumpy_23f1248fae'
```

Peek a boo!</details>

[...]

Issue #540, opened 2021-11-01
Importing `gtc:numpy` stencils import issues on Piz Daint


gronerl:
Sometimes, when running `gtc:numpy`stencils on Piz Daint (SCRATCH), the follwoing type of error appears:

```
self = <gt4py.backend.gtc_backend.gtcnumpy.backend.GTCNumpyBackend object at 0x2aab22d95ac0>

    def _load(self) -> Type["StencilObject"]:
        stencil_class_name = self.builder.class_name
        file_name = str(self.builder.module_path)
        stencil_module = gt_utils.make_module_from_file(stencil_class_name, file_name)
>       stencil_class = getattr(stencil_module, stencil_class_name)
E       AttributeError: 'NoneType' object has no attribute 'two_optional_fields____gtcnumpy_412215ca76'

../src/gt4py/backend/base.py:268: AttributeError

```
however the member `two_optional_fields____gtcnumpy_412215ca76` is there. It seems to work after deleting the corresponding file from the cache and re-generating.

havogt:
>however the member two_optional_fields____gtcnumpy_412215ca76 is there

Doesn't the error indicate that `stencil_module` is None and therefore something went wrong in `gt_utils.make_module_from_file(stencil_class_name, file_name)`?

havogt:
Shouldn't we raise an exception here https://github.com/GridTools/gt4py/blob/eab01fa4adfeead8bcb2024283f371f937cf081e/src/gt4py/utils/base.py#L313?

gronerl:
You're right. 

Yeah this is actually kind of annoying. It's been there since almost the beginning. @egparedes  Do you remember why we ignore exceptions here?

egparedes:
Yeah, there were a reason for that `try-except` in `make_module_from_file`, but I don't remember...

In any case, returning `None` from `make_module_from_file` has the same meaning of raising an exception and thus `_load` should have checked that the value is not `None` and later decide what to do. It should also be possible to check what was wrong in `make_module_from_file` because the silenced exception should have been printed.

For a future-proof solution, either we always raise exceptions in `make_module_from_file` or we could add a `raise_exceptions` keyword argument in `make_module_from_file` to let the caller set the behavior regarding exceptions (default value should be `False` for backward compatibility or `True` if we want to be strict).

gronerl:
[...]

PR #715, opened 2022-04-05
Use make_module_from_file in gtc:numpy

jdahm:
## Description

Use a load-retry loop with importlib instead of editing `sys.path` in the `gtc:numpy` backend.

Resolves https://github.com/GridTools/gt4py/issues/714.

### DIFF
--- a/src/gt4py/backend/gtc_backend/numpy/backend.py
+++ b/src/gt4py/backend/gtc_backend/numpy/backend.py
@@ -43,13 +43,15 @@ if TYPE_CHECKING:
 
 class GTCModuleGenerator(BaseModuleGenerator):
     def generate_imports(self) -> str:
-        comp_file = f"{self.builder.caching.module_prefix}computation{self.builder.caching.module_postfix}.py"
+        comp_pkg = (
+            self.builder.caching.module_prefix + "computation" + self.builder.caching.module_postfix
+        )
         return "\n".join(
             [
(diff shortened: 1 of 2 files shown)


egparedes (approved):
It looks good.

egparedes:
bors try

gronerl:
Huh I just arrived at the same solution :D Can someone explain what is different in this case, and why it seems to solve #540 and #714 ?

gronerl:
This fix also works on a branch that does not have support for `GT_CACHE_LOAD_RETRY_DELAY`, and reading the m_computation file directly after writing in my reliable reproducer hints that it can immediately be read without problems and that a delay wouldn't help. Do we fully understand the issue?

jdahm:
Not fully, but the code that was there originally was quite poor :)

Here is one concern of mine with the original code:
```
                "path_backup = sys.path.copy()",
                "sys.path.append(str(pathlib.Path(__file__).parent))",
                f"import {comp_pkg} as computation",
                "sys.path = path_backup",
                "del path_backup",
```
What happens (I think, haven't checked) is that `sys.path = path_backup` is not actually doing a deep copy, so the `del path_backup` sets the path in an odd state. Even if that's fine, this is not thread safe.

egparedes:
> Not fully, but the code that was there originally was quite poor :)
> 
I agree, the original code was very error-prone (to say the least) and the code in this PR is simply better.

PR #718, opened 2022-04-07
More aggressive `NoFieldAccessPruning`

gronerl:
HorizontalExecutions can not only be removed if there is no access at all, but also if there are read accesses but no write accesses. This PR adds those cases.


jdahm (approved):
Nice work. This makes sense - only keep Horizontal Executions that have outward edges to actual fields, eliminating those that only write to LocalScalars.

PR #722, opened 2022-04-08
Do not interpolate mask when indexing with an ndarray

jdahm:
## Description

This fixes a bug when indexing CPU storages using the numpy-style boolean ndarray.

Also makes a small optimization that skips mask propagation when all the axes are present.

### DIFF
--- a/src/gt4py/storage/storage.py
+++ b/src/gt4py/storage/storage.py
@@ -217,13 +217,17 @@ class Storage(np.ndarray):
                 self.__dict__ = {**obj.__dict__, **self.__dict__}
                 self.is_stencil_view = False
                 if hasattr(obj, "_new_index"):
-                    index_iter = itertools.chain(
-                        obj._new_index, [slice(None, None)] * (len(obj.mask) - len(obj._new_index))
-                    )
-                    interpolated_mask = gt_utils.interpolate_mask(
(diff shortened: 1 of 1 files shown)


gronerl:
It doesn't seem right to me to support boolean ndarrays as slice items. The result would be a flat array with none of the special properties of the storages. Is this an important use case? Would it work in a `setitem` situation?

jdahm:
We're definitely not using it and I wouldn't tell someone to, but there's a test in `pace` that compares storages with `numpy.testing.assert_array_equal`, and that method internally indexes with arrays.

Since we are subclassing `numpy.ndarray` by the Liskov Substitution Principle we should not throw an error with any indexing that the base class supports.

gronerl:
Yes, subclassing `np.ndarray` was a mistake and it will go away. Your fix here will just provide another way to break storages without a user noticing. (The resulting array view will have wrong meta information that if further used may lead to all kinds of funny stuff, including segfaults, trying to access non-existing GPU pointers etc.)

What is recommended to do in this case is what we also do in the gt4py tests and what is recommended to do in all interactions with `numpy` API that may trigger creation of `ndarrays`: Use  `numpy.testing.assert_array_equal(np.asarray(storage))`.

jdahm:
Ah ok I agree, that's the best path.

PR #723, opened 2022-04-11
Functional: get tests running on Windows

petiaccja:
I'll create the CI workflow in a separate PR

### DIFF
--- a/src/functional/ffront/common_types.py
+++ b/src/functional/ffront/common_types.py
@@ -12,7 +12,7 @@ class ScalarKind(IntEnum):
     BOOL = 1
     INT32 = 32
     INT64 = 64
-    INT = INT64 if np.int_ == np.int64 else INT32
+    INT = INT32 if np.int_ == np.int32 else INT64
     FLOAT32 = 1032
     FLOAT64 = 1064
     DIMENSION = 2001
--- a/src/functional/ffront/common_types.py
+++ b/src/functional/ffront/common_types.py
@@ -12,6 +12,8 @@ class ScalarKind(IntEnum):
     BOOL = 1
     INT32 = 32
     INT64 = 64
+    # Python's "int" type in the Python AST should be mapped to ScalarKind.INT in our ASTs. The size, as
+    # determined by numpy, varies by platform. (Size is the same as C's "long" type.)
     INT = INT32 if np.int_ == np.int32 else INT64
     FLOAT32 = 1032
     FLOAT64 = 1064
--- a/src/functional/ffront/common_types.py
+++ b/src/functional/ffront/common_types.py
@@ -2,6 +2,8 @@ import typing
 from dataclasses import dataclass
 from typing import Literal, Optional, Union
 
+import numpy as np
+
 from eve.type_definitions import IntEnum, StrEnum
 from functional import common as func_common
 
@@ -10,6 +12,7 @@ class ScalarKind(IntEnum):
     BOOL = 1
     INT32 = 32
     INT64 = 64
+    INT = INT64 if np.int_ == np.int64 else INT32
     FLOAT32 = 1032
     FLOAT64 = 1064
     DIMENSION = 2001
--- a/src/functional/ffront/past_to_itir.py
+++ b/src/functional/ffront/past_to_itir.py
@@ -121,9 +121,11 @@ class ProgramLowering(NodeTranslator):
         if slice_bound is None:
             lowered_bound = default_value
         elif isinstance(slice_bound, past.Constant):
-            assert isinstance(slice_bound.type, common_types.ScalarType) and (
-                slice_bound.type.kind == common_types.ScalarKind.INT64
-                or slice_bound.type.kind == common_types.ScalarKind.INT32
+            assert isinstance(
+                slice_bound.type, common_types.ScalarType
+            ) and slice_bound.type.kind in (
+                common_types.ScalarKind.INT32,
+                common_types.ScalarKind.INT64,
             )
             if slice_bound.value < 0:
                 lowered_bound = itir.FunCall(
--- a/src/functional/ffront/past_to_itir.py
+++ b/src/functional/ffront/past_to_itir.py
@@ -123,7 +123,7 @@ class ProgramLowering(NodeTranslator):
         elif isinstance(slice_bound, past.Constant):
             assert (
                 isinstance(slice_bound.type, common_types.ScalarType)
-                and slice_bound.type.kind == common_types.ScalarKind.INT64
+                and (slice_bound.type.kind == common_types.ScalarKind.INT64 or slice_bound.type.kind == common_types.ScalarKind.INT32)
             )
             if slice_bound.value < 0:
                 lowered_bound = itir.FunCall(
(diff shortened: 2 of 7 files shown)


src/functional/ffront/source_utils.py
```diff
@@ -32,7 +33,7 @@
 
 def make_source_definition_from_function(func: Callable) -> SourceDefinition:
     try:
-        filename = inspect.getabsfile(func) or MISSING_FILENAME
+        filename = os.path.realpath(inspect.getabsfile(func)) or MISSING_FILENAME
```
egparedes, line 36: `Pathlib.resolve()` should be equivalent and it is nowadays recommended to use `Pathlib` instead of `os.path` when possible.

src/functional/ffront/past_to_itir.py
```diff
@@ -112,9 +112,9 @@ def _visit_slice_bound(
         if slice_bound is None:
             lowered_bound = default_value
         elif isinstance(slice_bound, past.Constant):
-            assert (
-                isinstance(slice_bound.type, common_types.ScalarType)
-                and slice_bound.type.kind == common_types.ScalarKind.INT64
+            assert isinstance(slice_bound.type, common_types.ScalarType) and (
+                slice_bound.type.kind == common_types.ScalarKind.INT64
+                or slice_bound.type.kind == common_types.ScalarKind.INT32
```
egparedes, line 117: If `slice_bound.type.kind` has always the same value as `common_types.ScalarType.INT`, if should be just checked against `common_types.ScalarType.INT`. Otherwise I think using the `in` operator could be more readable, like:

```python
slice_bound.type.kind in (common_types.ScalarKind.INT32, common_types.ScalarKind.INT64)
```

tests/functional_tests/ffront_tests/test_program.py
```diff
@@ -145,7 +146,7 @@ def test_copy_parsing(copy_program_def):
                 kwargs={"out": P(past.Name, id="out_field")},
             )
         ],
-        location=P(past.SourceLocation, line=58, source=__file__),
+        location=P(past.SourceLocation, line=59, source=os.path.realpath(__file__)),
```
egparedes, line 149: Same comment as above about using `Pathlib`

egparedes (changes requested):
Thanks for the fixes. It looks good in general, just added some comments to slightly improve the code style.

src/functional/ffront/source_utils.py
petiaccja: I changed it to use 'pathlib', but converted it back to 'str'. We may be better off leaving at as Path, what do you think?

src/functional/ffront/past_to_itir.py
petiaccja: Yeah, that's a good question. Slices should probably be mapped to `ptrdiff_t` (or `intptr_t`)  in the generated C++ or CUDA code regardless of what numpy says. Should we add a `common_types.ScalarType.INTPTR`, or the existing `common_types.ScalarType.DIMENSION` is already for this?

src/functional/ffront/source_utils.py
egparedes: Yeah, I hadn't thought about it but it might be more reasonable to leave the data as an actual `Path` instance, although you would need then to change the type of the attribute in the `SourceDefinition` class

petiaccja: It seems like it's not necessarily a path: https://github.com/GridTools/gt4py/blob/3f261dde9f03590ffdfc7f7b8bf92ce64e4fedc1/tests/eve_tests/unit_tests/test_type_definitions.py#L77. It's probably better to just leave it as is.

src/functional/ffront/past_to_itir.py
petiaccja: I think I'm gonna just simplify it with a tuple and `in` as you suggested, so that it won't fail if somebody uses `numpy.int32` or `int64` to force the type.

src/functional/iterator/backends/roundtrip.py
```diff
@@ -125,11 +126,15 @@ def register_tmp(tmp, domain):
...
-        spec.loader.exec_module(foo)  # type: ignore
+        tmp.close()
+
+        try:
+            spec = importlib.util.spec_from_file_location("module.name", tmp.name)
+            foo = importlib.util.module_from_spec(spec)  # type: ignore
+            spec.loader.exec_module(foo)  # type: ignore
+        finally:
+            if not debug:
+                os.remove(tmp.name)
```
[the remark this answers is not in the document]
petiaccja: Yeah, that sounds good, I replaced it.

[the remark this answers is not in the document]
petiaccja: Refactored this section a bit, but probably the whole function could do with some refactoring as it's a bit too long.

src/functional/ffront/past_to_itir.py
egparedes: > so that it won't fail if somebody uses `numpy.int32` or `int64` to force the type.

I think that case can't happen because the values in the PAST are generated from values in a Python AST, which is created directly by the Python parser and it only uses `int`s for constants. It should be safe (and right) to be strict here and allow only `ScalarType.INT`

src/functional/ffront/common_types.py
```diff
@@ -10,6 +12,7 @@ class ScalarKind(IntEnum):
     BOOL = 1
     INT32 = 32
     INT64 = 64
+    INT = INT64 if np.int_ == np.int64 else INT32
```
egparedes, line 15: I realized it makes more sense to set the default to INT64 and only define INT as INT32 in case numpy does the same (windows platform).

src/functional/iterator/backends/roundtrip.py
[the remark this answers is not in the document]
egparedes: The whole `iterator` package is a prototype and it will be replaced or refactored in the future.

src/functional/ffront/common_types.py
petiaccja: You mean you just "flip" the if-else clauses? Yeah, maybe slightly better, but it should be only two options, right?

src/functional/ffront/past_to_itir.py
petiaccja: Could you not do something like:
```python
#program({"CONST": np.int32(42)})
def program(...):
  ... 
  op(data[CONST:-CONST])
```
This may not be a literal in the AST, but it's still a constant, right?

src/functional/ffront/common_types.py
egparedes: Yes, exactly.

src/functional/ffront/past_to_itir.py
petiaccja: Alright, let's have it with INT for now.

src/functional/ffront/common_types.py
```diff
@@ -10,6 +12,7 @@ class ScalarKind(IntEnum):
     BOOL = 1
     INT32 = 32
     INT64 = 64
+    INT = INT32 if np.int_ == np.int32 else INT64
```
egparedes, line 15: It could be worth to add a very short comment here to explain that this is needed because python `int`s are transformed to C `long` type by numpy, which may have different sizes on different platforms

egparedes (approved):
It looks good. I suggest to add a comment to explain the introduction of the INT scalar type and it is good to go.

PR #726, opened 2022-04-12
DaCe backend refactor

gronerl:
This PR replaces most of the current DaCe backend's way of generating SDFGs.

ToDo:
- [ ] validators for DaCeIR
- [ ] reintroduce oir_to_dace tests
- [ ] unit tests for expansion (to DaCeIR, to SDFG)
- [x] add GPU tests for test_dace_parsing (split off to #736)
- [x] add debuginfo
- [x] review pipeline performance (subset_str, sympy, debuginfo)
- [x] review file structure
- [ ] check for unused utilities
- [x] ~Generalize DaceStrMaker to work with both oir_to_dace and expansion~ instead integrated functionality in OirSDFGBuilder.SDFGContext
- [ ] review tiling with redundancy
- [x] ~remove expansion configuration beyond a simple working default~ it's used too much in the expansion to actually be removed. It was simplified somewhat though.
- [ ] add_map based on symbolic expressions once dace supports it in a released version


gronerl:
More detailed features  such as transformations on the DaceIR level and special treatment of regions will be added in later PR's. See #727

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try-
bors try

gronerl:
bors try

gronerl:
bors try

src/gtc/daceir.py
```diff
@@ -0,0 +1,892 @@
...
+class DomainLoop(IterationNode, ComputationNode):
+    axis: Axis
+    index_range: Range
+    loop_states: List[Union[ComputationState, "DomainLoop"]]
+
+
+class NestedSDFG(ComputationNode):
+    label: SymbolRef
+    field_decls: Dict[SymbolRef, FieldDecl]
+    symbol_decls: Dict[SymbolRef, SymbolDecl]
```
jdahm, line 886: The other IRs use a style where these are lists, and there is a symbol table (`symtable_` attribute) is implicitly defined. The visitors that need decls add the visitor context by adding

```
contexts = (SymbolTableTrait.symtable_merger,)
```

to the class. Is there a reason to not use this system for the daceir?

Side note: soon the method for adding the visitor context will have to change across gt4py, but for now this is the consistent approach.

src/gt4py/backend/dace_backend.py
```diff
@@ -50,6 +49,13 @@
     from gt4py.stencil_object import StencilObject
 
 
+def _serialize_sdfg(sdfg: dace.SDFG):
+    for sd in sdfg.all_sdfgs_recursive():
+        sd.transformation_hist = []
+        sd.orig_sdfg = None
+    return dumps(sdfg)
```
jdahm, line 56: This is a bit unintuitive: `_serialize_sdfg` actually mutates the input.

src/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -0,0 +1,401 @@
...
+                    edge.data,
+                )
+            self.sdfg.add_edge(old_state, state, dace.InterstateEdge())
+            self.state = state
+            return self
+
+        def add_loop(self, index_range: dcir.Range):
+
+            loop_state = self.sdfg.add_state()
+            after_state = self.sdfg.add_state()
```
jdahm, line 64: I'm not sure this is doing what you think it is. Both `loop_state` and `after_state` should be the same thing here, since it mutates the referenced `self`.

src/gtc/daceir.py
```diff
@@ -0,0 +1,893 @@
...
+        return set(ml.connector for ml in self.write_memlets)
+
+
+class IterationNode(Node):
+    grid_subset: GridSubset
+
+
+class Tasklet(ComputationNode, IterationNode, eve.SymbolTableTrait):
+    stmts: List[Union[LocalScalarDecl, Stmt]]
+    grid_subset: GridSubset = GridSubset.single_gridpoint()
```
jdahm, line 865: The unofficial rule we've been following is that IRs other than `gtir` do not store any Decls in statements. Could this be a list under the Tasklet outside the `stmts`?

src/gtc/dace/expansion/expansion.py
```diff
@@ -0,0 +1,145 @@
...
+                    inner_dim = inner_dim.subs(repldict)
+
+                equations.append(inner_dim - outer_dim)
+        if len(symbols) == 0:
+            return {}
+
+        # Solve for all at once
+        results = sympy.solve(equations, *symbols, dict=True)
+        result = results[0]
+        result = {str(k)[len("__SOLVE_") :]: v for k, v in result.items()}
```
jdahm, line 66: I'm a bit concerned by adding a sympy solve here but I get why it's needed.

src/gt4py/backend/dace_backend.py
gronerl: Done.

src/gtc/daceir.py
gronerl: Done.

src/gt4py/backend/dace_backend.py
```diff
@@ -143,7 +153,10 @@ def __call__(self, stencil_ir: gtir.Stencil) -> Dict[str, Dict[str, str]]:
         sdfg = OirSDFGBuilder().visit(oir_node)
 
         _to_device(sdfg, self.backend.storage_info["device"])
-        sdfg = _expand_and_finalize_sdfg(stencil_ir, sdfg, self.backend.storage_info["layout_map"])
+        sdfg = _pre_expand_trafos(stencil_ir, sdfg, self.backend.storage_info["layout_map"])
+        unexpanded_json = _serialize_sdfg(sdfg)
+        sdfg.expand_library_nodes()
+        _post_expand_trafos(sdfg)
```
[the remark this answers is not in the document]
gronerl: Is what you're proposing is something like:
```python
def sdfg_trafos(sdfg, *args):
    _pre_expand_trafos(sdfg, *args)
    yield
    _post_expand_trafos(sdfg)
    
with sdfg_trafos(sdfg):
    unexpanded_json = _serialize_sdfg(sdfg)
    sdfg.expand_library_nodes()
```
In that case, I don't think I agree, usually context managers denote some region where a context is built and torn down again, i.e. the status before the context is re-established upon exit. In this case, however, `_pre_expand_trafos` and `_post_expand_trafos` are not inverse functions. Rather, they are consecutive steps transforming the sdfg in a unidirectional manner. So I don't think using the `with` syntax will convey the right thing to a first-time reader of the code.

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try-
bors try

jdahm (approved):
Great commitment to getting this working! Modulo the remaining todos in https://github.com/GridTools/gt4py/issues/727 and ironing out some issues in ai2cm/pace, this first version should be ready to go.

gronerl:
bors try

gronerl:
bors try

Previous failure was only on GPU's py38 with a seemingly unrelated error. I'll try with a fix that has helped with similar issues in the past, but that could only be a temporary fix since the json roundtrip is rather expensive.

PR #731, opened 2022-04-14
Add gtc:dace:gpu backend

gronerl:
This PR adds the `gtc:dace:gpu` backend and renames `gtc:dace` to `gtc:dace:cpu`. 

This is done based on the soon-to-be-outdated expansion that is currently in `master`.  The intention is to already adapt the infrastructure for a dace GPU backend such that the backend refactor can be meaningfully be broken down to more digestable chunks. 

ToDo:
- [x] test with fv3core

### DIFF
--- a/src/gt4py/backend/gtc_backend/common.py
+++ b/src/gt4py/backend/gtc_backend/common.py
@@ -57,21 +57,24 @@ def pybuffer_to_sid(
     domain_dim_flags: Tuple[bool, bool, bool],
     data_ndim: int,
     stride_kind_index: int,
+    check_layout: bool = True,
     backend,
 ):
     domain_ndim = domain_dim_flags.count(True)
     sid_ndim = domain_ndim + data_ndim
 
-    as_sid = "as_cuda_sid" if backend.GT_BACKEND_T == "gpu" else "as_sid"
+    as_sid = "as_cuda_sid" if backend.storage_info["device"] == "gpu" else "as_sid"
 
     sid_def = """gt::{as_sid}<{ctype}, {sid_ndim},
-        gt::integral_constant<int, {unique_index}>, {unit_stride_dim}>({name})""".format(
+        gt::integral_constant<int, {unique_index}>{unit_stride_dim}>({name})""".format(
         name=name,
         ctype=ctype,
         unique_index=stride_kind_index,
         sid_ndim=sid_ndim,
         as_sid=as_sid,
-        unit_stride_dim=_get_unit_stride_dim(backend, domain_dim_flags, data_ndim),
+        unit_stride_dim=f", {_get_unit_stride_dim(backend, domain_dim_flags, data_ndim)}"
+        if check_layout
+        else "",
     )
     sid_def = "gt::sid::shift_sid_origin({sid_def}, {name}_origin)".format(
         sid_def=sid_def,
--- a/src/gt4py/backend/gtc_backend/common.py
+++ b/src/gt4py/backend/gtc_backend/common.py
@@ -29,6 +29,7 @@ from eve.codegen import MakoTemplate as as_mako
 from gt4py import backend as gt_backend
 from gt4py import ir as gt_ir
 from gt4py import utils as gt_utils
+from gt4py.backend import Backend
 from gt4py.backend.module_generator import BaseModuleGenerator, ModuleData
 from gt4py.definitions import AccessKind
 from gtc import gtir
@@ -57,8 +58,8 @@ def pybuffer_to_sid(
     domain_dim_flags: Tuple[bool, bool, bool],
     data_ndim: int,
     stride_kind_index: int,
+    backend: Backend,
     check_layout: bool = True,
-    backend,
 ):
     domain_ndim = domain_dim_flags.count(True)
     sid_ndim = domain_ndim + data_ndim
--- a/src/gt4py/backend/gtc_backend/dace/backend.py
+++ b/src/gt4py/backend/gtc_backend/dace/backend.py
@@ -14,6 +14,7 @@
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 import os
+import re
 from typing import TYPE_CHECKING, Dict, Optional, Tuple, Type
 
 import dace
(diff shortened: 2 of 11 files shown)


gronerl:
bors try

gronerl:
bors try-
bors try

gronerl:
bors try

gronerl:
For some reason, when testing fv3core in my usual setup on daint, cache building progresses very slowly (like 4 stencils/h build even in a sequential case.) CI tests on daint seem to progress as usual though. I'm opening this for review as-is but will try to find a way to verify validity on fv3core. (Or maybe someone from AI2 can verify? @jdahm)

src/gt4py/backend/gtc_backend/common.py
```diff
@@ -57,21 +57,24 @@ def pybuffer_to_sid(
     domain_dim_flags: Tuple[bool, bool, bool],
     data_ndim: int,
     stride_kind_index: int,
+    check_layout: bool = True,
     backend,
 ):
```
jdahm, line 62: It's a little nit of mine, but I prefer arguments (even inside the kwargs list) with defaults to come after those without. So, switch `backend` and `check_layout`. It's a bit orthogonal to the PR, but could you add a type hint to `backend` here as well as `_get_unit_stride_dim` above? Thanks!

src/gtc/dace/utils.py
```diff
@@ -490,3 +490,23 @@ def is_sdfg_equal(sdfg1: dace.SDFG, sdfg2: dace.SDFG):
...
+                ranks.extend(3 + c for c in range(len(mask) - 3))
+            else:
+                ranks.extend(-c for c in range(len(mask) - 3))
+
+        res_layout = [0] * len(ranks)
+        for i, idx in enumerate(np.argsort(ranks)):
+            res_layout[idx] = i
+        return tuple(res_layout)
+
+    return layout_maker
```
jdahm, line 512: I realize none of the other layout functions has type hints, but it'd be good to start doing that to make these consumable. I have had trouble parsing them in the past. Could you add some to this? Thanks!

tests/test_unittest/test_cli.py
```diff
@@ -94,7 +94,8 @@ def nocli_backend(scope="module"):
 
 BACKEND_ROW_PATTERN_BY_NAME = {
     "gtc:cuda": r"^\s*gtc:cuda\s*cuda\s*python\s*Yes",
-    "gtc:dace": r"^\s*gtc:dace\s*c\+\+\s*python\s*Yes",
+    "gtc:dace:cpu": r"^\s*gtc:dace:cpu\s*c\+\+\s*python\s*Yes",
+    "gtc:dace:gpu": r"^\s*gtc:dace:gpu\s*cuda\s*python\s*Yes",
```
jdahm, line 98: 👍

src/gt4py/backend/pyext_builder.py
```diff
@@ -122,6 +122,9 @@ def get_gt_pyext_build_opts(
         extra_compile_args["cxx"].append(
             "-isystem{}".format(os.path.join(dace_path, "runtime/include"))
         )
+        extra_compile_args["nvcc"].append(
```
jdahm, line 125: This could be combined with the append above to make the intent cleaner and the type check a little cleaner. Not critical though.
```python
# I assume this returns an Optional[str]? Would be better to check `is not None to avoid issues with empty string.
if dace_path := get_dace_module_path():
    comp_arg = "-isystem{}".format(os.path.join(dace_path, "runtime/include"))
    extra_compile_args["cxx"].append(comp_arg)
    extra_compile_args["nvcc"].append(comp_arg)
```

src/gtc/dace/expansion.py
```diff
@@ -926,6 +926,268 @@ def add_nodes_and_edges(self):
                 tasklet.in_connectors[edge.dst_conn] = dace.pointer(array.dtype)
 
 
+class BlockVerticalLoopExpander(NaiveVerticalLoopExpander):
+
+    default_tile_sizes = (64, 8)
```
jdahm, line 931: Probably better capitalized to indicate this is a constant. Could also be a [`ClassVar`](https://docs.python.org/3/library/typing.html#typing.ClassVar).

src/gtc/dace/expansion.py
```diff
@@ -926,6 +926,268 @@ def add_nodes_and_edges(self):
...
+
+    def get_tiled_subset_strs(self, nsdfg, extent):
```
jdahm, line 933: It's still quite difficult for me to parse this code without type hints. Is this going to be replaced with a new version that has those? If not, your IDE and the next reader through here would appreciate them!

gronerl:
bors try

gronerl:
bors try

src/gt4py/backend/pyext_builder.py
gronerl: They are different though: `-isystem{}` for `cxx` but `-isystem={}` for nvcc (I.e. without and with `=`)

gronerl:
I made changes according to your comments.

gronerl:
I wonder if I should investigate those bors timeouts. (The Gitlab run actually completed successfully.) Is this possibly the effect of the gtc:dace:gpu running slowly?

src/gt4py/backend/pyext_builder.py
jdahm: oh odd! I didn't realize that. Is that a property of the compiler?

jdahm (approved):
Looks great!

src/gt4py/backend/pyext_builder.py
gronerl: I think so. At least the options above follow the same pattern and nvcc seems not to support the version without `=`.

src/gt4py/backend/gtc_backend/dace/backend.py
```diff
@@ -66,11 +69,42 @@ def _specialize_transient_strides(sdfg: dace.SDFG, layout_map):
             sdfg.remove_symbol(k)
 
 
+def _to_device(sdfg: dace.SDFG, device):
```
jdahm, line 72:
```suggestion
def _to_device(sdfg: dace.SDFG, device: str) -> None:
    """Updates sdfg in place."""
```

jdahm:
As far as I can tell it's well within job timeout limits. We do periodically get timeouts from `bors`, which I have usually attributed to the background workflow manager launching k8 jobs.

I'd vote we try launching them one more time to see if it keeps happening, then just merge.

jdahm:
bors try

PR #735, opened 2022-04-19
Use the pretty printer in ir.Node.__str__

fthaler:
## Description

Overrides the `functional.ir.Node.__str__` method to use the IR pretty printer.

## Requirements

### DIFF
--- a/src/functional/iterator/ir.py
+++ b/src/functional/iterator/ir.py
@@ -1,9 +1,16 @@
 from typing import List, Union
 
-from eve import Node
+import eve
 from eve.traits import SymbolName, SymbolTableTrait
 from eve.type_definitions import SymbolRef
 from functional.iterator.util.sym_validation import validate_symbol_refs
+
+
+class Node(eve.Node):
+    def __str__(self):
+        from functional.iterator.pretty_printer import pformat
+
+        return pformat(self)
 
 
 class Sym(Node):  # helper
--- a/src/functional/iterator/ir.py
+++ b/src/functional/iterator/ir.py
@@ -19,6 +19,7 @@ class Sym(Node):  # helper
     id: SymbolName  # noqa: A003
 
 
+@noninstantiable
 class Expr(Node):
     ...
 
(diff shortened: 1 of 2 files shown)


src/functional/iterator/ir.py
```diff
@@ -1,11 +1,18 @@
 from typing import List, Union
 
-from eve import Node
+import eve
 from eve.traits import SymbolName, SymbolTableTrait
 from eve.type_definitions import SymbolRef
 from functional.iterator.util.sym_validation import validate_symbol_refs
 
 
+class Node(eve.Node):
```
egparedes, line 9: Would it make sense to make this class non-instantiable (`eve.utils.non_instantiable`)?

src/functional/iterator/ir.py
```diff
@@ -1,11 +1,18 @@
...
+    def __str__(self):
```
egparedes, line 10: Can you add a small test for this with a couple of different nodes ?

egparedes, line 10: Can you add a small test for this with a couple of different nodes ?

egparedes (commented):
It looks good, but I think we should add at least a small test to make sure we detect possible breaking changes int he future.

havogt:
My proposal was to implement a parser for the pretty printed code if it's cheap enough to maintain.

src/functional/iterator/ir.py
fthaler: Done.

src/functional/iterator/ir.py
fthaler: I added a minimal test to check the behavior of the `__str__` method (specific formatting is tested in the pretty printer).

fthaler:
> My proposal was to implement a parser for the pretty printed code if it's cheap enough to maintain.

Out of scope for this PR, but will try to get one with minimal effort.

havogt:
>Out of scope for this PR

sure...

egparedes (approved):
It looks good.

PR #741, opened 2022-04-22
Sync black versions

jdahm:
## Description

This ensures the same version of `black` is used in local development as in `pre-commit`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ include_package_data = True
 python_requires = >= 3.8
 install_requires =
     attrs>=20.3
-    black==22.3.0
+    black>=22.3.0
     cached-property>=1.5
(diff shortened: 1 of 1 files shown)


egparedes (commented):
I understand the motivation, but I'd rather use a less strict requirement here (>=) since this involves that any other package in the environment needs to work with exactly this black version, which could not be always the case.

egparedes (approved):
It looks good.

jdahm:
bors try

jdahm:
It timed out.

jdahm:
bors try

jdahm:
Argh, something is going on with piz daint today. Both here and on ai2cm/pace we're having trouble with tests timing out...

PR #744, opened 2022-04-26
Add support for PEP 563

stubbiali:
[PEP 563](https://peps.python.org/pep-0563/) delays the evaluation of type hints after the definition time. Type hints are converted into string literals and stored in `__annotations__`. This PR proposes to evaluate argument annotations of the definition function which are strings unless they are present in the `dtypes` dictionary (so to still accommodate for parametrized data types).

### DIFF
--- a/src/gt4py/gtscript.py
+++ b/src/gt4py/gtscript.py
@@ -112,6 +112,12 @@ _VALID_DATA_TYPES = (
 
 def _set_arg_dtypes(definition: Callable[..., None], dtypes: Dict[Type, Type]):
     def _parse_annotation(arg, annotation):
+        """
+        Evaluate the type hint ``annotation`` for the stencil argument ``arg``.
+
+        Note that ``typing.get_type_hints()`` cannot be used here since field
+        arguments are annotated using instances (and not subclasses) of ``Field``.
+        """
         if isinstance(annotation, _FieldDescriptor) and isinstance(annotation.dtype, str):
             if annotation.dtype in dtypes:
                 return _FieldDescriptor(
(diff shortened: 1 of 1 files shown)


src/gt4py/gtscript.py
```diff
@@ -111,23 +111,28 @@
...
-            if value in dtypes:
-                annotations[arg] = dtypes[value]
+                raise ValueError(f"Missing '{annotation.dtype}' dtype definition for arg '{arg}'")
+        elif isinstance(annotation, str):
+            if annotation in dtypes:
+                return dtypes[annotation]
             else:
-                raise ValueError(f"Missing '{value}' dtype definition for arg '{arg}'")
+                def_globals = getattr(definition, "__globals__", {})
+                return _parse_annotation(arg, eval(annotation, def_globals))
```
egparedes, line 127: I might be wrong, but it looks to me like this part can be fully replaced by a call to `typing.get_type_hints()` (maybe with customized `globalns` dict). Have you already tried?

egparedes (commented):
Thanks for the contribution. It looks good in general, but I wonder if the custom evaluation of the annotations couldn't be replace by the standard call to `get_type_hints()` which should obtain the same result without messing with explicit evaluation of annotation strings in our codebase.

src/gt4py/gtscript.py
stubbiali: I've just tried, but it doesn't seem to work since `typing.get_hints()` seeks to convert string annotations into types, but fields are annotated using *instances* of `Field`.

egparedes: Ok, makes sense. Could you add a small comment mentioning this issue in the sources?

egparedes:
bors try

egparedes (approved):
It looks good.

PR #746, opened 2022-04-27
allow simple case field broadcasting.

DropD:
## Description

Allow Field broadcasting in field view in the simple case that both fields have the same dimensions, except one has more.

Example:

`Field[[X, Y], dtype]` can be broadcast to `Field[[X, Y, Z], dtype]` but not to `Field[[Y, X], dtype]` or `Field[[Z, X, Y], dtype]`

This should be safe and is absolutely required for icon integration.

## Requirements

- [ ] All the authors are covered by a valid contributor assignment agreement,
- [ ] The names of all the new contributors have been added to an updated

### DIFF
--- a/src/functional/ffront/foast_passes/type_deduction.py
+++ b/src/functional/ffront/foast_passes/type_deduction.py
@@ -38,7 +38,17 @@ def are_broadcast_compatible(left: TypeInfo, right: TypeInfo) -> bool:
     True
 
     """
-    if all([left.dims, right.dims]) and left.dims != right.dims:
+    both_dims_given = bool(left.dims and right.dims)
+    both_dims_given &= left.dims is not Ellipsis
+    both_dims_given &= right.dims is not Ellipsis
+    if both_dims_given and any(
+        [
+            ldim != rdim
+            for ldim, rdim in zip(
+                left.dims, right.dims  # type: ignore[arg-type]  # we know they are lists here
+            )
+        ]
+    ):
         return False
     return left.dtype == right.dtype
 
@@ -60,6 +70,8 @@ def broadcast_typeinfos(left: TypeInfo, right: TypeInfo) -> Optional[TypeInfo]:
     if not are_broadcast_compatible(left, right):
         return None
     if left.is_scalar and right.is_field_type:
+        return right
+    elif left.dims and right.dims and len(right.dims) > len(left.dims):
         return right
     return left
 
--- a/src/functional/ffront/type_info.py
+++ b/src/functional/ffront/type_info.py
@@ -135,7 +135,10 @@ class TypeInfo:
 
     @property
     def dims(self) -> Optional[list]:
-        return getattr(self.type, "dims", None)
+        result = getattr(self.type, "dims", None)
+        if result is Ellipsis:
+            return []
+        return result
 
     @property
     def dtype(self) -> Optional[ScalarType]:


src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -38,7 +38,17 @@ def are_broadcast_compatible(left: TypeInfo, right: TypeInfo) -> bool:
...
     """
-    if all([left.dims, right.dims]) and left.dims != right.dims:
+    both_dims_given = bool(left.dims and right.dims)
+    both_dims_given &= left.dims is not Ellipsis
+    both_dims_given &= right.dims is not Ellipsis
+    if both_dims_given and any(
+        [
+            ldim != rdim
+            for ldim, rdim in zip(
+                left.dims, right.dims  # type: ignore[arg-type]  # we know they are lists here
```
egparedes, line 48: Check my comment above, maybe mypy is right here.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -61,6 +71,8 @@ def broadcast_typeinfos(left: TypeInfo, right: TypeInfo) -> Optional[TypeInfo]:
         return None
     if left.is_scalar and right.is_field_type:
         return right
+    elif left.dims and right.dims and len(right.dims) > len(left.dims):
```
egparedes, line 74: Just a question: is it guaranteed that the longer list is always at the right?

egparedes (commented):
I have some questions about the PR. It would also be nice to have some minimal test.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -38,7 +38,17 @@ def are_broadcast_compatible(left: TypeInfo, right: TypeInfo) -> bool:
     True
 
     """
-    if all([left.dims, right.dims]) and left.dims != right.dims:
+    both_dims_given = bool(left.dims and right.dims)
+    both_dims_given &= left.dims is not Ellipsis
+    both_dims_given &= right.dims is not Ellipsis
```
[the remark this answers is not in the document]
DropD: The `TypeInfo.dims` returns `None` if the type it holds does not have a `dims` attribute. I would agree that this is not great but I have an improved version of the type checking tools drafted somewhere else to work on during the cooldown.

src/functional/ffront/foast_passes/type_deduction.py
DropD: No. If it isn't, then the left typeinfo is returned. So the one with more or same amount of dims.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -38,7 +38,17 @@ def are_broadcast_compatible(left: TypeInfo, right: TypeInfo) -> bool:
...
+    if both_dims_given and any(
```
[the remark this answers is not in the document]
DropD: fixed

src/functional/ffront/foast_passes/type_deduction.py
egparedes: Ok, I see. I misread the code because of the unnecessary `elif` here which I think could be just a simple `if`

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -38,7 +38,15 @@ def are_broadcast_compatible(left: TypeInfo, right: TypeInfo) -> bool:
     True
 
     """
-    if all([left.dims, right.dims]) and left.dims != right.dims:
+    both_dims_given = bool(left.dims and right.dims)
+    both_dims_given &= left.dims != GenericDimensions()
```
egparedes, line 42: If `GenericDimensions` is not a dataclass or defines a meaningful eq operator, most likely `left.dims != GenericDimensions()` would always be `True` because all instances of the class will compare as not equal.

src/functional/ffront/type_info.py
```diff
@@ -15,6 +15,13 @@
 )
 
 
+class GenericDimensions:
```
egparedes, line 18: I think this class should be defined as an empty frozen dataclass to make sure it behaves as expected

egparedes (approved):
It looks good.

PR #750, opened 2022-04-28
Re-enable cuda, numpy and dace tests.

gronerl:
Many tests currently exclude cuda, numpy and dace:* backends due to the use of INTERNAL_BACKENDS which filters based on the backend name starting with "gt". 

This PR removes this filter and instead adds a requires_dace marker that can be used to exclude dace tests as necessary.


gronerl:
bors try

egparedes (approved):
It looks good.

gronerl:
bors try

PR #757, opened 2022-05-03
functional: More type validation tools for eve

egparedes:
Decouple type validation functionality from `datamodels` in `eve` and other enhacements.

Main changes:
- Added a new `type_validation` module for run-time type validation with a generic interface supporting different implementations
- Additions to `extended_typing`, including new functionality and new typing defintions
- Additions and improvements to `type_definitions`, including `frozenlist` (alias of tuple) and `frozendict` (3rd party) collections
- Minor improvements to typing annotations of `utils` subpackage

### DIFF
--- a/src/eve/type_definitions.py
+++ b/src/eve/type_definitions.py
@@ -23,11 +23,12 @@ import ast
 import enum
 import functools
 import re
+import sys
 
 import pydantic
 import xxhash
 from boltons.typeutils import classproperty as classproperty  # noqa: F401
-from frozendict import frozendict as frozendict  # noqa: F401
+from frozendict import frozendict as _frozendict  # noqa: F401
 from pydantic import validator  # noqa
 from pydantic import (  # noqa: F401
     NegativeFloat,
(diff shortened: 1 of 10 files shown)


src/eve/type_definitions.py
```diff
@@ -41,12 +42,37 @@
...
+    final,
+)
+
+
+frozenlist: Final = tuple
+frozendict: Final = _frozendict if sys.version_info >= (3, 9) else xtyping.FrozenDict
+
+
+@final
+class NothingType(type):
```
DropD, line 65: General note:  While you are working on this, why not add docstrings at least to the new classes (and maybe one or two old ones per PR)? Currently the learning curve for anyone new in `eve` would be quite steep. And it makes it more likely for the reviews to be meaningful.

DropD (changes requested):
Very minor typos and / or things I did not understand.

src/eve/utils.py
```diff
@@ -35,37 +35,41 @@
 import warnings
 
 import xxhash
-from boltons.iterutils import flatten, flatten_iter, is_collection  # noqa: F401
+from boltons.iterutils import (  # noqa: F401
+    flatten as flatten,
+    flatten_iter as flatten_iter,
+    is_collection as is_collection,
+)
```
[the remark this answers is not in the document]
DropD: I don't mind it, but agree it should be documented because it is the kind of thing I would accidentally remove if I didn't know.

src/eve/type_definitions.py
DropD: Actually, I noticed after writing that that almost everything else had a doctoring. And if it becomes simpler then all the better.

DropD (approved):

PR #760, opened 2022-05-09
Getting started guide

petiaccja:
The jupyter-notebook-style short guide for the field view frontend.

Note: Blocked by #761. Use that branch if you want to run the examples.


fthaler (changes requested):
This is not really a full review, but as I did a course about technical and scientific writing, I thought it might make sense to briefly go through the documentation.
In general, I like the guide. It provides simple examples that are easy to follow and introduce all relevant concepts one by one.
So my remarks are mostly of stylistic nature and follow the Google and Microsoft developer documentation guidelines (https://developers.google.com/style, https://docs.microsoft.com/en-us/style-guide/welcome/).
The most important points are probably the following:
- Use ‘you’ instead of ‘we’ (https://developers.google.com/style/person)
- Use the active voice (https://developers.google.com/style/voice)
- Use the present tense (https://developers.google.com/style/tense)

docs/functional/QuickstartGuide.md
```diff
@@ -0,0 +1,375 @@
...
+b = np_as_located_field(CellDim, KDim)(np.full(shape=grid_shape, fill_value=b_value, dtype=np.float32))
+```
+
+*Note: The interface to construct fields is provisional only and will change soon.*
+
++++
+
+#### Field operators
+
+Field operators perform operations on a set of fields, for example, elementwise addition or reduction along a dimension. You can write field operators as Python functions by using the `@field_operator` decorator. Field operators cannot have side effects, therefore you cannot modify its arguments within their body. Only a subset of the Python syntax is allowed inside field operators--the library checks for correctness.
```
fthaler, line 82:
```suggestion
Field operators perform operations on a set of fields, for example, elementwise addition or reduction along a dimension. You can write field operators as Python functions by using the `@field_operator` decorator. Field operators cannot have side effects, therefore you cannot modify its arguments within their body. Only a subset of the Python syntax is allowed inside field operators—the library checks for correctness.
```

fthaler (approved):

PR #761, opened 2022-05-10
Functional: Add support for reductions over neighbor/sparse fields.

tehrengruber:
This PR adds support for reductions over neighbor/sparse dimensions, e.g. the following works now:

```python
@field_operator
def reduction(nb_field: Field[[rs.Vertex, rs.V2EDim], "float64"]) -> Field[[rs.Vertex], "float64"]:  # type: ignore
    return neighbor_sum(nb_field, axis=V2EDim)
```

In order for neighbor fields to work the iterator prototype needs to distinguish between regular dimensions and local dimensions such that it can apply a partial shift when creating a field from such an iterator (this way of handling sparse fields is not ideal and should be changed at some point). As we choose to not use offsets as dimensions in the frontend this PR adds a `local` attribute to `Dimension` to allow the former.


src/functional/common.py
```diff
@@ -28,6 +29,14 @@
...
 class Dimension:
     value: str
+    local: bool = dataclasses.field(default=False)
+
+    def __gt_type__(self):
+        # TODO(tehrengruber): Revisit. This coupling to the frontend is not
+        #  optimal.
+        import functional.ffront.common_types as ct
+
+        return ct.DimensionType(self)
```
DropD, line 39: In fact we do have a pattern already in use, which avoids this coupling. Check out the `fbuiltins.FieldOffset`. It is not optimal, but this kind of wrong way around coupling is worth almost any price to avoid.

src/functional/ffront/fbuiltins.py
```diff
@@ -64,6 +64,7 @@ def __gt_type__(self):
 # TODO(ricoh): This should be reunified with ``iterator.runtime.Offset``
 # potentially in ``functional.common``, which requires lifting of
 # ``ffront.common_types`` into ``functional``.
+# TODO(tehrengruber): remove inheritance since this is conceptually different to itir offset
```
DropD, line 67: The two TODOs are contradicting each other.

If the latter applies, then it would require explanation why `Dimension` in field view is not also conceptually different from `Dimension` in iterator view.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -191,17 +191,21 @@ def visit_Symbol(
...
+                if not target2.local:
+                    raise FieldOperatorTypeDeductionError.from_foast_node(
+                        new_value, msg="Second dimension in offset must be a local dimension."
+                    )
+                new_type = ct.OffsetType(source=source, target=(target1,))
+            case ct.OffsetType(source=_, target=(_,)):
+                # for cartesian axes (e.g. I, J) the index of the subscript only
+                #  signifies the displacement in the respective dimension,
+                #  but does not change the target type.
+                new_type = new_value.type
```
DropD, line 208: It seems there might be a loop hole here, where `FieldOffset(source=X, target=[Y])` would silently be treated as `FieldOffset(source=X, target=[X])`.

It looks like it should either be disallowed or treated differently. If it is disallowed and caught at an earlier time, then this is fine here.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -369,11 +374,32 @@ def visit_Call(self, node: foast.Call, **kwargs) -> foast.Call:
...
+                    raise FieldOperatorTypeDeductionError.from_foast_node(
+                        node,
+                        msg=f"Incompatible field argument in {node.func.id}. "
+                        f"Expected a field with dimension {reduction_dim}, "
+                        f"but got {field_dims_str}.",
+                    )
+                return_type = ct.FieldType(
+                    dims=[dim for dim in field_type.dims if dim != reduction_dim],
+                    dtype=field_type.dtype,
+                )
```
DropD, line 394: Might I suggest the pattern from `FieldOperatorLowering`:

```python
        if node.func.id in FUN_BUILTIN_NAMES:
            visitor = getattr(self, f"_visit_{node.func.id}")
            return visitor(node, **kwargs)
        ...
        
    def _visit_neighbor_sum(self, node: foast.Call, **kwargs):
        <function specific stuff here>
```

src/functional/ffront/decorator.py
```diff
@@ -45,6 +45,40 @@
...
+                    if (
+                        all_captured_vars[potential_collision]
+                        != all_val_captured_vars[potential_collision]
+                    ):
+                        collisions.append(potential_collision)
+                if collisions:
+                    raise NotImplementedError(
+                        "Using closure vars with same name, but different value "
+                        "across functions is not implemented yet."
+                    )
```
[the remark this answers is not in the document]
tehrengruber: A list of collisions contrary to failing on the first collision is somewhat convenient, even if only accessible when debugging. I added the collisions to the error message.

src/functional/ffront/decorator.py
```diff
@@ -45,6 +45,40 @@
...
+                val_captured_vars = _collect_capture_vars(val_captured_vars)
+
+                all_val_captured_vars = collections.ChainMap(
+                    val_captured_vars.globals, val_captured_vars.nonlocals
+                )
+                collisions = []
+                for potential_collision in set(all_captured_vars) & set(all_val_captured_vars):
+                    if (
+                        all_captured_vars[potential_collision]
+                        != all_val_captured_vars[potential_collision]
```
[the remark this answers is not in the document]
tehrengruber: I agree, I changed to:
`new_captured_vars` + `flat_captured_vars`
`vars_of_val` + `flat_vars_of_val`

src/functional/ffront/decorator.py
```diff
@@ -45,6 +45,40 @@
 DEFAULT_BACKEND = "roundtrip"
 
 
+def _collect_capture_vars(captured_vars: CapturedVars) -> CapturedVars:
```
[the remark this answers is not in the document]
tehrengruber: Nope I renamed to `_collect_captured_vars`. I'd like to keep the underscore as the function should be kept private API later to be removed.

src/functional/ffront/foast_passes/type_deduction.py
DropD: 1) aliasing would not be supported. It's not exactly on purpose but it is a bearable limitation for now.

2) function calls on expressions other than a `Name` lead to a syntax error before type deduction.

src/functional/ffront/fbuiltins.py
```diff
@@ -61,10 +61,10 @@ def __gt_type__(self):
 BUILTINS = {name: globals()[name] for name in __all__ + ["bool", "int", "float"]}
 
 
-# TODO(ricoh): This should be reunified with ``iterator.runtime.Offset``
-# potentially in ``functional.common``, which requires lifting of
-# ``ffront.common_types`` into ``functional``.
-# TODO(tehrengruber): remove inheritance since this is conceptually different to itir offset
+# TODO(tehrengruber): FieldOffset and runtime.Offset are not an exact conceptual
+#  match. Revisit if we want to continue subclassing here. If we split
+#  them also check whether Dimension should continue to subclass or define
```
DropD, line 66: Tiny optional improvement before merging: `Dimension` is not subclassing anymore.

src/functional/ffront/foast_passes/type_deduction.py
tehrengruber: I sketched how to support this with Enrique a while ago. By lifting the identity of built-in functions to the type level we could dispatch without relying on what is stored inside `Name`. If this is needed we can implement it rather quickly, but until typing on the iterator level and its relation to the frontend is resolved I am more comfortable with keeping the type system lean. Note that the aliasing limitation also applies to data types, which prevents patterns like these:
```
dtype = float64
@field_operator
def foo(...):
  return dtype("1.")
```

PR #762, opened 2022-05-10
Update flake8 and workaround flakehell issue

jdahm:
## Description

Flake8 version 4.x is incompatible with flakehell, so deletes flakehell and updates flake8.


egparedes (approved):
It looks good.

PR #763, opened 2022-05-10
FieldView type inspection improvements

DropD:
## Description

While it is good that functions to extract specific information from a `ffront.common_types.SymbolType` are collected in the `ffront.type_info`, the design of that module itself requires an overhaul.

- The wrapper class `TypeInfo` which wraps a `SymbolType` and exposes some inspection methods / properties was envisioned to simplify interactions with multiple types (just wrap them all and let them interact). The use case for this never really materialised. Instead creating the wrapper class everywhere has become unnecessary boiler plate.

- Originally everything had to be very permissive, as not everything could be typed. This is no longer the case.

- Some questions about types are now asked a bit differently. It has become interesting whether a type represents something that would be lowered to a "Value" or an "Iterator" or neither. On the other hand scalars and fields are often interchangeable.

## Requirements


src/functional/ffront/common_types.py
```diff
@@ -69,7 +69,7 @@ class OffsetType(SymbolType):
     target: Optional[tuple[func_common.Dimension, ...]] = None
 
     def __str__(self):
-        return f"Offset[{self.id}]"
```
DropD, line 72: This must have been some copy pasta bug or a leftover (did this class ever have an `id` attribute?).

src/functional/ffront/field_operator_ast.py
```diff
@@ -54,7 +54,7 @@ class Symbol(eve.GenericNode, LocatedNode, Generic[SymbolT]):
 
 
 class Expr(LocatedNode):
-    type: Optional[common_types.SymbolType] = None  # noqa A003
+    type: common_types.SymbolType = common_types.DeferredSymbolType(constraint=None)  # noqa A003
```
DropD, line 57: These have always been treated as semantically equivalent. Deferred is just a little easier to work with.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -11,102 +11,41 @@
 # distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
```
DropD, line 13: Mainly cleanup, refactoring to use new `type_info` functions instead of `TypeInfo` and increased code reuse.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -367,11 +271,30 @@ def visit_Call(self, node: foast.Call, **kwargs) -> foast.Call:
...
-            return foast.Call(func=new_func, args=new_args, location=node.location, type=new_type)
+            return foast.Call(
+                func=new_func, args=new_args, kwargs={}, location=node.location, type=new_type
+            )
         elif isinstance(new_func.type, ct.FunctionType):
+            new_args = self.visit(node.args, **kwargs)
+            new_kwargs = {
+                keyword: self.visit(arg, **kwargs) for keyword, arg in node.kwargs.items()
+            }
+            try:
```
DropD, line 282: It does look like a bit of an oversight that signatures are checked only now, but keep in mind that there didn't use to be any builtin functions not too long ago...

src/functional/ffront/foast_to_itir.py
```diff
@@ -22,18 +22,14 @@
     fbuiltins,
     field_operator_ast as foast,
     itir_makers as im,
+    type_info,
```
DropD, line 25: Just refactoring for `type_info` changes.

src/functional/ffront/func_to_foast.py
```diff
@@ -423,13 +426,10 @@ def visit_Call(self, node: ast.Call, **kwargs) -> foast.Call:
...
 
-        args = node.args.copy()
-        for keyword in node.keywords:
-            args.append(keyword.value)
-
         return foast.Call(
             func=self.visit(node.func, **kwargs),
-            args=[self.visit(arg, **kwargs) for arg in args],
+            args=[self.visit(arg, **kwargs) for arg in node.args],
+            kwargs={keyword.arg: self.visit(keyword.value, **kwargs) for keyword in node.keywords},
```
DropD, line 432: Yes, kwargs were not previously allowed and `nbh_sum` had a workaround for that. Other than this it's just refactoring.

src/functional/ffront/type_info.py
```diff
@@ -1,48 +1,235 @@
...
 
-def is_complete_scalar_type(sym_type: Optional[SymbolType]) -> TypeGuard[ScalarType]:
-    if isinstance(sym_type, ScalarType):
+class TypeKind(enum.Enum):
+    FIELD = 0
+    SCALAR = 1
+    UNKNOWN = 2
+
+
+def type_kind(symbol_type: ct.SymbolType) -> TypeKind:
```
tehrengruber, line 24: Deducing if something is a field or a scalar is a common thing to ask, but can you make a case for the behavior of this function that goes beyond what can be done with a `type_class(symbol_type) == ct.FieldType | ct.ScalarKind`? I would argue that the additional behavior creates more harm then good. A tuple of fields or a function returning a field behaves different to a field not only with respect to the entity itself, but also with respect to the information the `SymbolType` will provide to you. Arguably they should also be treated differently. Consider for example https://github.com/GridTools/gt4py/blob/3b0425deee343ed890377cd9f396d40cdd746733/src/functional/ffront/type_info.py#L225-L227 which does not make sense for a function returning fields, but is still relying on this function. If you want to avoid writing out the `type_class` comparison each and every time I would propose to introduce two functions `is_field` and `is_scalar`.

src/functional/ffront/type_info.py
```diff
@@ -1,48 +1,235 @@
...
+    Examples:
+    ---------
+    >>> print(extract_dtype(ct.ScalarType(kind=ct.ScalarKind.FLOAT64)))
+    float64
+
+    >>> print(extract_dtype(ct.FieldType(dims=[], dtype=ct.ScalarType(kind=ct.ScalarKind.BOOL))))
+    bool
+
+    >>> print(extract_dtype(ct.FunctionType(args=[], kwargs={}, returns=ct.ScalarType(kind=ct.ScalarKind.INT32, shape=[2, 2]))))
+    int32[2, 2]
```
tehrengruber, line 100: I do not agree with the dtype of a function being the dtype of its return value, I believe that is a different question to ask.

src/functional/ffront/type_info.py
```diff
@@ -65,154 +252,51 @@ def function_signature_incompatibilities(
...
-    False
-    >>> typeinfo_b = TypeInfo(None)
-    >>> typeinfo_b.is_any_type
-    True
-    >>> typeinfo_b.is_arithmetic_compatible
-    False
-    >>> typeinfo_b.can_be_refined_to(typeinfo_a)
-    True
-
+def can_call(
```
tehrengruber, line 261: I am generally not a fan of the term `can` when naming functions and it is in my experience more common to use `is` + an adjectival suffix (-able, -like). I see the case for making this look sentence like, but I do not endorse this style as it inevitably breaks and thus can not be done consistently anyway.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -214,116 +154,91 @@ def visit_Subscript(self, node: foast.Subscript, **kwargs) -> foast.Subscript:
...
+        for one_type, other_type in permutations([left.type, right.type]):
+            if type_info.can_promote_dims(other_type, one_type):
+                return copy.copy(one_type)
+
+        # the case of left_type == right_type is already handled above
+        # so here they must be incompatible
         raise FieldOperatorTypeDeductionError.from_foast_node(
-            parent,
-            msg=f"Incompatible type(s) for operator '{op}': {left.type}, {right.type}!",
+            node, msg=f"Incompatible dimensions in operation {left.type} '{node.op}' {right.type}!"
```
tehrengruber, line 223: When I started programming I was often confused by this style of error messages especially if other errors give me the actual syntax I wrote.

tehrengruber (changes requested):
I like the new approach.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -367,11 +271,30 @@ def visit_Call(self, node: foast.Call, **kwargs) -> foast.Call:
...
             new_type = ct.FieldType(dims=new_dims, dtype=new_func.type.dtype)
-            return foast.Call(func=new_func, args=new_args, location=node.location, type=new_type)
+            return foast.Call(
+                func=new_func, args=new_args, kwargs={}, location=node.location, type=new_type
+            )
         elif isinstance(new_func.type, ct.FunctionType):
+            new_args = self.visit(node.args, **kwargs)
+            new_kwargs = {
+                keyword: self.visit(arg, **kwargs) for keyword, arg in node.kwargs.items()
+            }
```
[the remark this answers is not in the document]
DropD: Ah, I wanted to check on that but forgot...

src/functional/ffront/type_info.py
```diff
@@ -1,48 +1,235 @@
...
+
+    Examples:
+    ---------
+    >>> type_kind(ct.DeferredSymbolType(constraint=None)).name
+    'UNKNOWN'
+
+    >>> type_kind(ct.ScalarType(kind=ct.ScalarKind.BOOL)).name
+    'SCALAR'
+
+    >>> type_kind(ct.FunctionType(args=[], kwargs={}, returns=ct.FieldType(dims=[], dtype=ct.ScalarType(kind=ct.ScalarKind.INT32)))).name
```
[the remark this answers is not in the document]
DropD: It does in execution but for the type deduction and the lowering the question we want to ask is: "What type kind is  this part of the expression after execution?" 

If we want to ask the question "Is it a scalar, a field or a function?" then there already is one obvious way to do it: `isinstance`. We never need ask that question though.

Therefore I think the semantics are what they should be. How can we communicate them better?

Maybe it should be renamed to `resulting_type_kind`? Or possibly the phrase "resulting type kind" should appear in the docstring?

src/functional/ffront/type_info.py
DropD: See above for the general reasoning. The example here is definitely a place where I should have used `type_class` instead of `type_kind`.

Consider the following usage though:
 https://github.com/GridTools/gt4py/blob/3b0425deee343ed890377cd9f396d40cdd746733/src/functional/ffront/func_to_foast.py#L222

If you change it to `elif type_class(...) is ct.ScalarType` it will fail because the question we should be asking there is not "Is it a scalar?" but "Is the left hand side of the assignment going to be a scalar?", which is equivalent to "Is the resulting type kind scalar?".

Another motivating example is of course the lowering, where we want to know: "Should we treat this as a 'Value' or as an 'Iterator' (or do the concepts not apply)?". Which turns out to be equivalent to "Is the resulting type kind 'Scalar', 'Field' or 'Unknown'"?

DropD: To make it even less confusing, I have replaced all usage except in the lowering by `type_info.is_concrete(...) and type_info.type_kind(...) is ScalarType | FieldType`. I am not sure that is the right question to ask, but all the tests pass, so it seems a good enough question to ask for now.

src/functional/ffront/type_info.py
DropD: What other dtype would you extract from a function type? The dtype of a random argument?

src/functional/ffront/foast_passes/type_deduction.py
DropD: Do you think the original `'{node.op}': {left.type}, {right.type}` is less confusing to a beginner?

src/functional/ffront/type_info.py
DropD: It comes down to "Best effort" vs "Defensive". If the function should provide a best effort result, it is unambiguously expected to return the return value dtype. If the function should be defensive it should fail in as many cases as can reasonably argued for. I prefer best effort and will document the behaviour in the docstring.

DropD: > I do not agree with the dtype of a function being the dtype of its return value, I believe that is a different question to ask.

Nothing here says that the dtype of a function is the dtype of it's return value. It sais if you *extract* a dtype from a function, what you are really asking must be to *extract* the dtype from it's return type. Anything else would not make sense.

src/functional/ffront/type_info.py
DropD: I honestly do not care, `is_<adjective>` style is more consistent with what we have elsewhere. That's probably more important than the longer names it leads to.

src/functional/ffront/foast_passes/type_deduction.py
tehrengruber: I would say so. Your latter sentence actually surprises me, if easy to do can you give an example? Here is what gcc does:
```
main.cpp:7:6: error: no match for ‘operator+’ (operand types are ‘std::array’ and ‘int’)
```
Which is more akin to the previous error message.

tehrengruber: And Python:
```
TypeError: unsupported operand type(s) for +: 'Bla' and 'int'
```

DropD: That's good enough to change back I would say.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -284,27 +285,36 @@ def visit_Call(self, node: foast.Call, **kwargs) -> foast.Call:
...
+                    type_info.is_callable(
+                        new_func.type,
+                        with_args=[arg.type for arg in new_args],
+                        with_kwargs={keyword: arg.type for keyword, arg in new_kwargs.items()},
+                        raise_exception=True,
+                    )
+                except GTTypeError as err:
+                    raise FieldOperatorTypeDeductionError.from_foast_node(
+                        node, msg=f"Invalid argument types in call to '{node.func.id}'!"
+                    ) from err
```
tehrengruber, line 309: I think it makes sense to run this also for the `node` returned in the if branch.

tehrengruber (approved):
Tiny request, looks good beside that.

DropD:
Looks like the tests that are failing are failing on the `functional` branch also. Waiting for upstream fix to merge.

PR #764, opened 2022-05-11
Fix bug in mask_overlap_with_extent

### LINKED ISSUE #755: Unbounded horizontal regions prevent extent stacking in `StencilExtentComputer`
The issue can be observed when adding the test case
```
        (
                common.HorizontalMask(
                    i=common.HorizontalInterval.full(),
                    j=common.HorizontalInterval.full(),
                ),
                -1,
                ((-1, 0), (0, 0)),
        ),
```
to `test_stencil_extents_region`. 
The Cause for this issue is that the representation of the unbounded region extents with `(-10000, 10000)` in `_overlap_along_axis` and `mask_overlap_with_extent` will cause the result of the subtraction in `GeneralAccess.to_extent` be (10000, -10000)" which will then be snapped back to zero extents. 

This workaround works for my use case
```
diff --git a/src/gtc/passes/horizontal_masks.py b/src/gtc/passes/horizontal_masks.py
index cc600f10..f61dc9df 100644
--- a/src/gtc/passes/horizontal_masks.py
+++ b/src/gtc/passes/horizontal_masks.py
@@ -24,21 +24,25 @@ def _overlap_along_axis(
     extent: Tuple[int, int], interval: common.HorizontalInterval
 ) -> Optional[Tuple[int, int]]:
     """Return a tuple of the distances to the edge of the compute domain, if overlapping."""
-    if interval.start is not None and interval.start.level == common.LevelMarker.START:
-        start_diff = extent[0] - interval.start.offset
+    clamped_start = interval.start or common.AxisBound(
+        level=common.LevelMarker.START, offset=extent[0]
+    )
+    clamped_end = interval.end or common.AxisBound(level=common.LevelMarker.END, offset=extent[1])
+    if clamped_start.level == common.LevelMarker.START:
+        start_diff = extent[0] - clamped_start.offset
     else:
         start_diff = None
 
-    if interval.end is not None and interval.end.level == common.LevelMarker.END:
-        end_diff = extent[1] - interval.end.offset
+    if clamped_end.level == common.LevelMarker.END:
+        end_diff = extent[1] - clamped_end.offset
     else:
         end_diff = None
 
     if start_diff is not None and start_diff > 0 and end_diff is None:
-        if interval.end.offset <= extent[0]:
+        if clamped_end.offset <= extent[0]:
             return None
     elif end_diff is not None and end_diff < 0 and start_diff is None:
-        if interval.start.offset > extent[1]:
+        if clamped_start.offset > extent[1]:
             return None
 
     start_diff = min(start_diff, 0) if start_diff is not None else -10000
```
but seems unsafe since the 10000/-10000 is still a possible return value. A proper solution should probably deal with some representation of unbounded intervals. Using  `None` is already supported in Extent, yet doesn't seem to behave in a way under arithmetic operations that the current code remains correct for other cases.

jdahm:
Thanks to @gronerl for the change. Resolves #755.

### DIFF
--- a/src/gtc/passes/horizontal_masks.py
+++ b/src/gtc/passes/horizontal_masks.py
@@ -24,21 +24,26 @@ def _overlap_along_axis(
     extent: Tuple[int, int], interval: common.HorizontalInterval
 ) -> Optional[Tuple[int, int]]:
     """Return a tuple of the distances to the edge of the compute domain, if overlapping."""
-    if interval.start is not None and interval.start.level == common.LevelMarker.START:
-        start_diff = extent[0] - interval.start.offset
+    clamped_start = interval.start or common.AxisBound(
+        level=common.LevelMarker.START, offset=extent[0]
+    )
+    clamped_end = interval.end or common.AxisBound(level=common.LevelMarker.END, offset=extent[1])
+
+    if clamped_start.level == common.LevelMarker.START:
+        start_diff = extent[0] - clamped_start.offset
     else:
         start_diff = None
 
-    if interval.end is not None and interval.end.level == common.LevelMarker.END:
-        end_diff = extent[1] - interval.end.offset
+    if clamped_end.level == common.LevelMarker.END:
+        end_diff = extent[1] - clamped_end.offset
     else:
         end_diff = None
 
     if start_diff is not None and start_diff > 0 and end_diff is None:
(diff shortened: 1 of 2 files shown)


gronerl (approved):

jdahm:
bors try

PR #775, opened 2022-05-13
Functional: refactor of node traits and extension to visitors instead of contexts

egparedes:
## Description

Simplify existing node traits and add visitor traits to eve.

Main changes:
- Simplify and unify in the same module all the traits dealing with creation or validation of symbol tables.
- Add a similar simple traits-based mechanism to node visitors instead of the current `contexts` mechanism, which is more cumbersome.


src/eve/traits.py
```diff
@@ -20,73 +20,124 @@
...
+
+# --- Visitor Traits ---
+class VisitorWithSymbolTableTrait(visitors.NodeVisitor):
+    """Visitor trait to update or add the symtable to kwargs in the visitor calls.
+
+    Visitors inhering from this trait will automatically pass the active
+    symbol table to visitor methods as the 'symtable' keyword argument.
+    """
+
+    def visit(self, node: concepts.TreeNode, **kwargs: Any) -> Any:
```
havogt, line 133: Is this safe? What happens if we inherit from 2 traits, both defining a `visit`?

havogt (commented):
Just one question...

PR #777, opened 2022-05-16
allow broadcasting of [K] and [Edge, K] fields

DropD:
## Description

Relax implicit field broadcasting to: the smaller list of dimensions must be fully contained within the bigger one.

Example:

```
Field[[Edge, K], ...] + Field[[Edge], ...] # worked before, keeps working
Field[[K], ...] + Field[[Edge, K], ...] # works too now
```

The tradeoff is that less potential run time errors are caught, but this can only be solved by a less simplistic approach.

## Requirements

### DIFF
--- a/src/functional/ffront/foast_passes/type_deduction.py
+++ b/src/functional/ffront/foast_passes/type_deduction.py
@@ -24,8 +24,10 @@ def are_broadcast_compatible(left: TypeInfo, right: TypeInfo) -> bool:
     """
     Check if ``left`` and ``right`` types are compatible after optional broadcasting.
 
-    A binary field operation between two arguments can proceed and the result is a field.
-    on top of the dimensions, also the dtypes must match.
+    If both are fields and do not have the same list of dimensions, then the smaller
+    dimension list must be fully contained in the bigger one (ordered subset).
+
+    Dtypes must also match in any case.
 
     Examples:
     ---------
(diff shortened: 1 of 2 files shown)


samkellerhals (approved):
Looks good to me.

samkellerhals:
@DropD It seems that the `tests/functional_tests/ffront_tests/test_type_deduction.py::test_binop_nonmatching_dims` test still needs to be fixed.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -37,17 +39,37 @@ def are_broadcast_compatible(left: TypeInfo, right: TypeInfo) -> bool:
...
+        both_dims_given and isinstance(left.dims, list) and isinstance(right.dims, list)
+    ):  # isinstance calls are for mypy
+        smaller_dims, bigger_dims = (
+            (left.dims, right.dims)
+            if len(left.dims) <= len(right.dims)
+            else (right.dims, left.dims)
         )
-    ):
-        return False
+        if smaller_dims[0] in bigger_dims and smaller_dims[-1] in bigger_dims:
```
tehrengruber, line 68:
```suggestion
        if set(left.dims).issubset(right.dims) or set(right.dims).issubset(left.dims):
          return True
```
May I'm missing the point, but can't we just do something like this?

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -37,17 +39,37 @@ def are_broadcast_compatible(left: TypeInfo, right: TypeInfo) -> bool:
...
+            start_index = bigger_dims.index(smaller_dims[0])
+            end_index = bigger_dims.index(smaller_dims[-1])
+            if smaller_dims != bigger_dims[start_index : end_index + 1]:
+                return True
```
tehrengruber, line 72:
```suggestion
                return False
```
If the two don't match it should return `False` right?

src/functional/ffront/foast_passes/type_deduction.py
DropD: If you can prove that the order is not important, I would love to. For example we'd be saying `[K, Edge]` can be implicitly broadcast to `[Edge, K]`.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -37,17 +39,39 @@ def are_broadcast_compatible(left: TypeInfo, right: TypeInfo) -> bool:
...
     both_dims_given = bool(left.dims and right.dims)
     both_dims_given &= not isinstance(left.dims, GenericDimensions)
     both_dims_given &= not isinstance(right.dims, GenericDimensions)
-    if both_dims_given and any(
-        ldim != rdim
-        for ldim, rdim in zip(
-            left.dims, right.dims  # type: ignore[arg-type]  # we know they are lists here
+    if (
+        both_dims_given and isinstance(left.dims, list) and isinstance(right.dims, list)
+    ):  # isinstance calls are for mypy
```
havogt, line 62: Wouldn't the mypy hints be better expressed as asserts?

DropD: I would prefer that, will have to check if mypy accepts it.

DropD: 👍 ✅

havogt (approved):

src/functional/ffront/foast_passes/type_deduction.py
tehrengruber: With respect to iterator IR this is no problem. I tested this:
```
@field_operator
def add(inp1: Field[[IDim, JDim], float64], inp2: Field[[JDim, IDim], float64]) -> Field[[IDim, JDim], float64]:
    return inp1+inp2
```
and it does work as expected as long you give the right output type. The problem is that there is no good/easy way to decide what the output type should be. Is it `Field[[IDim, JDim], float64]` or `Field[[IDim, JDim], float64]`. Right now `inp1+inp2` will give you `Field[[IDim, JDim], float64]`, but `inp1(Ioff[1])+inp2(Joff[1])` gives `Field[[JDim, IDim], float64]`. I think it is not worth making the frontend capable of this, so let's leave it as is.

PR #782, opened 2022-05-17
Fix small bugs in functional mostly related to loose type checking

egparedes:
## Description

Fixes for type validation issues discovered during tentative migration to datamodels.


havogt (commented):
iterator/** looks good

Issue #785, opened 2022-05-19
Fix failing horizontal execution merging time test


jdahm:
After the latest commit [e2d713e](https://github.com/GridTools/gt4py/commit/e2d713e2242dafbed2fece47208540c80ee2ac18) the following test is failing

```
tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py::test_horiz_exec_merging_complexity
```

The failure seems unrelated to the change.

PR #605, opened 2022-01-07
Fix HorizontalExecutionMerging Complexity

fthaler:
## Description

Fixes #599.

## Requirements

Before submitting this PR, please make sure:

- [x] You have run the code checks, tests and documentation build successfully
- [x] All fixes and all new functionality are tested and documentation is up to date
- [x] You looked at the [review checklist](https://github.com/GridTools/gt4py/blob/master/REVIEW_CHECKLIST.txt)

Additionally, if this PR contains code authored by new contributors:

- [x] All the authors are covered by a valid contributor assignment agreement,
signed by the employer if needed, provided to ETH Zurich
- [x] The names of all the new contributors have been added to an updated
version of the AUTHORS.rst file included in the PR

PR #787, opened 2022-05-19
Make process time check depend on initial benchmark

jdahm:
## Description

A test in gt4py currently checks raw process time. This makes it check a non-dimensional measure compared to an initial benchmark, in order to eliminate failures based on processor clock rate.

Resolves https://github.com/GridTools/gt4py/issues/785.

### DIFF
--- a/tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py
+++ b/tests/test_unittest/test_gtc/test_passes/test_oir_optimizations/test_horizontal_execution_merging.py
@@ -97,6 +97,22 @@ def test_horiz_exec_merging_map_scalar():
 
 
 def test_horiz_exec_merging_complexity():
+    start_time = time.process_time()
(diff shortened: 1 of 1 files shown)


gronerl:
bors try

gronerl:
Thanks for the PR, I like the idea of normalizing the runtime. I think the condition should be n times the single merge case, not `n`/2. (There is 1 merging in the benchmark vs `n` mergings in the actual performance test.) I also added a 1.5 factor to give the run some slack for e.g. variance. I verified that this test is suitable for detecting the issue fixed in #605.

gronerl:
bors try

gronerl (approved):

jdahm:
@gronerl I initially had it as `n * ` as well, but realized that _techincally_ the `single_process_time` variant also includes building up the IR from the factories, and could include validators, etc. That said, I think what you did is great and should give it even a little more slack so we don't get random failures :)

PR #789, opened 2022-05-20
Functional: Temporary workaround pydantic update

havogt:
The real fix needs to go in `gtfn_backend.extract_fundefs_from_closures()` where `id(stencil)` is not stable with pydantic 1.9.1. (A new node is allocated (probably in `add_fundefs`)).


egparedes:
@havogt, Is this PR still relevant after migrating to datamodels?

PR #794, opened 2022-05-24
Use default cuda stream in dace orchestration (Possibly fix #793)

### LINKED ISSUE #793: Non-deterministic test failure on test_dace_parsing
During a CI run for #790 one parametrization of test_dace_parsing failed (with all values of the result array not being updated.) 
I was unable to reproduce on `daint`, with a `pip list` differeing from the one in the CI as follows (No failure in ~15 trials. ):
```
Repro env            vs CI env
------------------------------
cupy-cuda112==10.4.0 vs cupy-cuda101==9.2.0
distlib==0.3.4       vs distlib==0.3.2
fastrlock==0.8       vs fastrlock==0.6
filelock==3.7.0      vs filelock==3.0.12
numpy==1.22.4        vs numpy==1.21.0
packaging==21.3      vs packaging==21.0
pip==22.1.1          vs pip==21.1.2
pluggy==1.0.0        vs pluggy==0.13.1 ###??? same version as dace?
py==1.11.0           vs py==1.10.0
pyparsing==3.0.9     vs pyparsing==2.4.7
setuptools==62.3.2   vs setuptools==57.0.0
tox==3.25.0          vs tox==3.23.1
virtualenv==20.14.1  vs virtualenv==20.4.7
wheel==0.37.1        vs wheel==0.36.2

```
The issue seems to appear only non-deterministically, since it did not appear in a subsequent CI run anymore.

gronerl:
The recently observed CI failure (see issue #793) may be caused by cupy not causing a device synchronization when copying host-to-device. With this PR this possible cause should be excluded. Either way, in our expiericence with using dace orchestration it is good practice to set those options.

### DIFF
--- a/tests/test_integration/test_dace_parsing.py
+++ b/tests/test_integration/test_dace_parsing.py
@@ -39,6 +39,8 @@ def dace_env():
         / "dacecache"
     )
     with dace.config.temporary_config():
+        dace.config.Config.set("compiler", "cuda", "max_concurrent_streams", value=-1)
+        dace.config.Config.set("compiler", "cpu", "openmp_sections", value=False)
         dace.config.Config.set("compiler", "cpu", "args", value="")
         dace.config.Config.set("compiler", "allow_view_arguments", value=True)
         dace.config.Config.set("default_build_folder", value=str(gt_cache_path))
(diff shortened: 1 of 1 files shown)


jdahm (approved):
Thanks for the quick fix!

PR #802, opened 2022-05-26
CI test PR

gronerl:
Trying to manually blame a commit from where CI starts failing. 
Trying 50503087eae255be4bb2cc5ede0da27f3696e424

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -86,7 +86,7 @@ cuda101 =
 cuda102 =
     cupy-cuda102
 dace =
-    dace~=0.13
+    dace==0.13
 format =
     clang-format>=9.0
 testing =
--- a/tests/test_integration/test_dace_parsing.py
+++ b/tests/test_integration/test_dace_parsing.py
@@ -39,9 +39,6 @@ def dace_env():
         / "dacecache"
     )
     with dace.config.temporary_config():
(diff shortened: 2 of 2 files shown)


jdahm:
I'm not positive that one exists with dace in a relatively recent state. I think it's something to do with a system upgrade on the GitHub Actions runners, since it's even happening on `master` after you reverted the change.

gronerl:
Which is worrying, since I have no idea how to debug this in that case.

gronerl:
bors try

jdahm (approved):

tehrengruber (changes requested):
@gronerl Can you add a small comment that the tests should be enabled again as soon as the refactored dace backend lands. If I understood you correctly those are genuine failures, we just don't care for now.

jdahm:
Could you also open an issue so we make sure to get to the bottom of the errors? These are important features for the dace backend to support, especially for Pace.

gronerl:
@tehrengruber I think that the issues should be fixed is implied by the tests being marked `xfail`, which is exactly what that is used for. I added a `reason` string to the pytest mark and reference a newly opened issue (@jdahm). I'll also directly remove the xfail tag again in the #726 today right after merging this into master, so I won't forget.

gronerl:
bors try

PR #803, opened 2022-05-30
Add math built-ins

BenWeber42:
## Description

Describe the content of the PR and links to related issues, bugs of features.

## Requirements


BenWeber42:
Current status:

Contains tests for iterator IR. Ffront tests are still in development.

While adding the tests, various questions came up regarding the current testing strategy and the current stage of the project (prototyping vs _production ready_). Hannes mentioned that the testing strategy should probably be reviewed as the functional branch grows. However, I don't want to scope creep this PR too much. I'd prefer to split additional work into follow up tasks.

Note: I will apply formatting & linting later.

BenWeber42:
I took the math built-ins from: https://github.com/GridTools/gt4py/blob/bc50f396b85fef1978628d56e245395252668771/src/gt4py/frontend/nodes.py#L387-L419

src/functional/iterator/builtins.py
```diff
@@ -136,3 +163,141 @@ def make_tuple(*args):
 @builtin_dispatch
 def tuple_get(*args):
     raise BackendNotSelectedError()
+
+
+# FIXME(ben): various of the following built-ins will shadow python built-ins!
+# We should find a good way around this...
```
BenWeber42, line 169: This is particularly annoying if you want to create a jupyter notebook which executes stencils, but also uses built-in functions in normal Python code. E.g., `min` and `max` are built-ins that won't work anymore if one just `from ....builtins import *`.

tests/functional_tests/iterator_tests/test_builtins.py
```diff
@@ -108,6 +110,168 @@ def test_arithmetic_and_logical_builtins(backend, builtin, inputs, expected):
         assert np.allclose(np.asarray(out), expected)
 
 
+@pytest.mark.parametrize(
+    "math_fun, ref_impl, inputs",
+    [
+        # FIXME(ben): what about pow?
```
BenWeber42, line 116: I think `pow` should be there as well.

tests/functional_tests/iterator_tests/test_builtins.py
```diff
@@ -108,6 +110,168 @@ def test_arithmetic_and_logical_builtins(backend, builtin, inputs, expected):
...
+        # FIXME(ben): dataset is missing invalid ranges (mostly nan outputs)
+        # FIXME(ben): we're not properly testing different datatypes
```
BenWeber42, line 118: I think this should be split into a follow up task. Other tests don't test with different datatypes either. Could probably go into a task as part of a bigger testing improvements effort.

tests/functional_tests/iterator_tests/test_builtins.py
```diff
@@ -108,6 +110,168 @@ def test_arithmetic_and_logical_builtins(backend, builtin, inputs, expected):
...
+        (
+            abs,
+            np.abs,
+            ([-1, 1, -1.0, 1.0, 0, -0, 0.0, -0.0],),
+        ),
+        (
+            min,
+            # FIXME(ben): what's the signature & semantics of `min` & `max`?
+            # `np.min` and python's built-in `min` have very different signatures...
```
BenWeber42, line 127: Not sure about this one. Python has `min(<iterable>)` and numpy: `np.min(<array>, axis=...)`. So in numpy, you can't do `np.min(a, b)`.

tehrengruber: In addition to `np.min` there is `np.minimum(a, b)` (https://numpy.org/doc/stable/reference/generated/numpy.minimum.html), which is what we want. So to stay consistent with numpy I would rename to the latter.

src/functional/iterator/builtins.py
```diff
@@ -136,3 +163,141 @@ def make_tuple(*args):
...
+
+
+@builtin_dispatch
+def ceil(*args):
+    raise BackendNotSelectedError()
+
+
+@builtin_dispatch
+def trunc(*args):
+    raise BackendNotSelectedError()
```
tehrengruber, line 303: I would suggest to have a seperate list (of strings) for the builtins that can be appended to `__all__` and in addition be used to automatically create all of these functions (e.g. using `globals()[builtin_name] = builtin_dispatch(stub)`. This has the disadvantage that it might break some IDEs introspection, but at least we avoid the code bloat. Your call.

tests/functional_tests/iterator_tests/test_builtins.py
```diff
@@ -108,6 +110,168 @@ def test_arithmetic_and_logical_builtins(backend, builtin, inputs, expected):
         assert np.allclose(np.asarray(out), expected)
 
 
+@pytest.mark.parametrize(
+    "math_fun, ref_impl, inputs",
+    [
```
tehrengruber, line 115: I think the testing data should refactored into a seperate fixture.

tehrengruber:
@havogt Can you review the Iterator part of this PR?

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -319,6 +323,65 @@ def _ensure_signature_valid(self, node: foast.Call, **kwargs) -> None:
...
+        # validate arguments
+        error_msg_preamble = f"Incompatible argument in call to `{func_name}`."
+        if func_name in fbuiltins.UNARY_MATH_NUMBER_BUILTIN_NAMES:
+            arg_validator = type_info.is_arithmetic
+        elif func_name in fbuiltins.UNARY_MATH_FP_BUILTIN_NAMES:
+            arg_validator = type_info.is_floating_point
+        elif func_name in fbuiltins.UNARY_MATH_FP_PREDICATE_BUILTIN_NAMES:
+            arg_validator = type_info.is_floating_point
+        elif func_name in fbuiltins.BINARY_MATH_NUMBER_BUILTIN_NAMES:
+            arg_validator = type_info.is_arithmetic
```
BenWeber42, line 338: This introduces overloads for `power` and `mod` that also support integer types. This is technically fine. However, C++ also has these overloads, but will cast them to `double` in this case and return a double value (see below). Whereas, numpy seems to preserve integer types (see `numpy.power` vs `numpy.float_power`). I'm just unsure how easy/hard it would be to have a CUDA `power` function that preserves integer types. It might be easier to only support floating point values for now.

From: https://en.cppreference.com/w/cpp/numeric/math/pow
> If any argument has [integral type](https://en.cppreference.com/w/cpp/types/is_integral), it is cast to double

src/functional/iterator/tracing.py
```diff
@@ -211,6 +212,15 @@ def less(*args):
     return _f("less", *args)
 
 
+for math_builtin_name in builtins.MATH_BUILTINS:
+    decorator = getattr(iterator.builtins, math_builtin_name).register(TRACING)
+    exec(
+        f"""@iterator.builtins.{math_builtin_name}.register(TRACING)\n"""
+        f"""def {math_builtin_name}(*args):\n"""
+        f"""  return _f("{math_builtin_name}", *args)\n"""
+    )
```
BenWeber42, line 221: Is the `exec` really necessary here? Is this so that the function will have the correct value for the `__name__` attribute?

tests/functional_tests/ffront_tests/test_math_builtin_execution.py
```diff
@@ -0,0 +1,118 @@
...
+    elif builtin_name in fbuiltins.UNARY_MATH_FP_PREDICATE_BUILTIN_NAMES:
+        annotations = {"inp": Field[[IDim], float64], "return": Field[[IDim], bool]}
+    elif builtin_name in fbuiltins.BINARY_MATH_NUMBER_BUILTIN_NAMES:
+        annotations = {
+            "inp1": Field[[IDim], float64],
+            "inp2": Field[[IDim], float64],
+            "return": Field[[IDim], float64],
+        }
+    else:
+        raise AssertionError(f"Unknown builtin `{builtin_name}`")
```
BenWeber42, line 50: nitpick: assert?

tests/functional_tests/ffront_tests/test_math_builtin_execution.py
```diff
@@ -0,0 +1,118 @@
...
+
+    captured_vars = CapturedVars(
+        nonlocals={"IDim": IDim},
+        globals={builtin_name: getattr(fbuiltins, builtin_name)},
+        annotations=annotations,
+        builtins=set(),
+        unbound=set(),
+    )
+
+    loc = foast.SourceLocation(line=1, column=1, source="none")
```
BenWeber42, line 60: Question: There is no easier way to fix source locations? Similar to `ast.fix_missing_locations`?

tests/functional_tests/iterator_tests/math_builtin_test_data.py
```diff
@@ -0,0 +1,353 @@
...
+                ],
+            ),
+        ),
+        # (
+        #    math_fun_field_op_gamma,
+        #    np.frompyfunc(math.gamma, nin=1, nout=1),
+        #    # FIXME(ben): math.gamma throws when it overflows, maybe should instead yield `np.inf`?
+        #    # overflows very quickly, already at `173`
+        #    ([-1002.3, -103.7, -1.2, -0.7, -0.1, 0.1, 0.7, 1.0, 1, 1.2, 100, 103.7, 170.5],),
+        # ),
```
BenWeber42, line 302: Is gamma tested now anywhere?

BenWeber42 (commented):
I couldn't review the `decorator.py` changes that well, since I'm not that familiar how everything is wired up there.

I noticed that we now have roughly three places where all math built-ins are introduced and each of those has the same categorizations (unary fp, unary fp predicate, binary, etc.). Early on, I was considering implementing it only in one place and then add functionality/information so that it can be used for the ffront ir, iterator ir, etc. Might be interesting to consider this trade-off more.

Otherwise, looks good overall.

src/functional/ffront/decorator.py
```diff
@@ -441,9 +444,18 @@ def from_function(
...
+        return_type = self.foast_node.body[-1].value.type
+        if not isinstance(return_type, ct.FieldType):
+            raise GTTypeError(
+                f"Return type of a FieldOperator must be a Field, but got `{return_type}`"
+            )
+        type_ = ct.FunctionType(
+            args=[param.type for param in self.foast_node.params], kwargs={}, returns=return_type
+        )
+        assert type_info.is_concrete(type_)
+        return cast(ct.FunctionType, type_)
```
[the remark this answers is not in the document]
tehrengruber: Yes this is somewhat weird, but the `type_info.is_concrete(type_)` sadly makes mypy lose that information.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -319,6 +323,65 @@ def _ensure_signature_valid(self, node: foast.Call, **kwargs) -> None:
...
+        else:
+            raise AssertionError(f"Unknown math builtin `{func_name}`.")
```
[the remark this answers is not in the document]
tehrengruber: I can't find the link right now, but if I remember correctly mypy doesn't like `assert False`. Disabling the check is not meaningful or gives any performance anyway. The cost of the assert, or any assert in general, is the evaluation of its condition. If the condition is `False` there is nothing to optimize.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -319,6 +323,65 @@ def _ensure_signature_valid(self, node: foast.Call, **kwargs) -> None:
...
+            try:
+                return_type = type_info.promote(
+                    *((cast(ct.FieldType | ct.ScalarType, arg.type)) for arg in node.args)
+                )
+            except GTTypeError as ex:
+                raise FieldOperatorTypeDeductionError.from_foast_node(
+                    node, msg=error_msg_preamble
+                ) from ex
+        else:
+            raise AssertionError(f"Unknown math builtin `{func_name}`.")
```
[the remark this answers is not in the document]
tehrengruber: Same as above.

src/functional/iterator/tracing.py
tehrengruber: Yes, this is the motivation. I am not particularly happy about this, but I didn't find a cleaner solution. If we refactor the iterator part we should certainly get rid of such constructs.

tests/functional_tests/ffront_tests/test_math_builtin_execution.py
tehrengruber: No, this is really annoying if one generates foast expressions programmatically. Locations are mandatory to avoid accidentally not adding them in the Python -> FOAST lowering (a problem that made errors in the cartesian toolchain aweful), but this is the downside. Some way to disable this would be nice, e.g. a wrapper that behaves like the `foast` module, but automatically adds the locations whenever a node is created.

tests/functional_tests/iterator_tests/math_builtin_test_data.py
tehrengruber: Ups. Enabled now.

src/functional/ffront/foast_passes/type_deduction.py
tehrengruber: A power implementation for integer types in cuda should be rather easy or am I missing something? After all we will need to wrap the the built-ins anyway to make sure we adhere to numpy semantics. I would like to keep the functionality as otherwise users start to write code like `some_int**2.`.

DropD (commented):
This PR adds some `__init__.py` files to test folders, making them packages. Current convention seems to be that this should only be done if there is a reason for that folder to be a package (fixtures or utils need to be importable, that have a reason not to be in a `conftest.py`). Often the culprit is that some IDEs default to adding an `__init__.py` to every folder in which they find a python module.

tests/functional_tests/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+# All rights reserved.
+#
+# This file is part of the GT4Py project and the GridTools framework.
+# GT4Py is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or any later
+# version. See the LICENSE.txt file at the top-level directory of this
+# distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
```
DropD, line 13: Making `tests/functional_tests` a package seems unneccessary.

tests/functional_tests/ffront_tests/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+# All rights reserved.
+#
+# This file is part of the GT4Py project and the GridTools framework.
+# GT4Py is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or any later
+# version. See the LICENSE.txt file at the top-level directory of this
+# distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
```
DropD, line 13: Is there a reason for making `ffront_tests` a package?

DropD (changes requested):

src/functional/iterator/embedded.py
```diff
@@ -317,6 +318,16 @@ def less(first, second):
...
 
 
+for math_builtin_name in builtins.MATH_BUILTINS:
+    decorator = getattr(builtins, math_builtin_name).register(EMBEDDED)
+    if math_builtin_name == "gamma":
+        # numpy has no gamma function
+        impl = np.vectorize(math.gamma)
+    else:
+        impl = getattr(np, math_builtin_name)
+    globals()[math_builtin_name] = decorator(impl)
```
havogt, line 328:
```suggestion
    decorator(impl)
```
no need to add the function to the module

src/functional/iterator/tracing.py
havogt: I don't see the value, I propose to do the same as in embedded.

havogt: btw, this style could be extended to almost all other builtins here

havogt (changes requested):

tests/functional_tests/__init__.py
tehrengruber: The frontend tests and the iterator tests use the same test data (https://github.com/GridTools/gt4py/blob/cb84312eb0c6ef557405856be3f27592b7247818/tests/functional_tests/ffront_tests/test_math_builtin_execution.py#L20). After discussion with @egparedes I opted for relative imports (consequently requiring the init files). Using conftest is also a possibility, but placing the data in `tests/conftest.py` is not appealing.

src/functional/iterator/embedded.py
tehrengruber: All other builtins are accessible from the module so deviating from that for the math builtins doesn't feel right to me. If we don't expose them here, I would argue the decorator of the other builtins should not expose them either. I personally prefer exposing all of them, but making it a convention to not use them directly.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -115,6 +115,9 @@ def arithmetic(
 
 
 def test_power(fieldview_backend):
```
havogt, line 117: Should we delete these tests?

tehrengruber: No, but I can move them to `test_math_builtin_execution.py`. The tests here test the `base ** exp` syntax, while the other tests only test the direct call to `power`.

tests/functional_tests/ffront_tests/__init__.py
tehrengruber: Same as above.

havogt (approved):

PR #818, opened 2022-06-13
Functional: allow explicit domain in field view

havogt:
Adds the possibility to pass domain as argument in field view program, instead of relative to the output field. If all domains are explicit, the itir fencil will not get the extra size arguments. This needs cleanup and checks...

```python
@program
def copy(
    dom: Domain[IDim, JDim, KDim],
    inp: Field[[IDim, JDim, KDim], float64],
    out: Field[[IDim, JDim, KDim], float64],
):
    copy_stencil(inp, out=out[dom])
```

### DIFF
--- a/src/functional/common.py
+++ b/src/functional/common.py
@@ -20,9 +20,12 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import Any, Generic, TypeVar
 
+from eve.extended_typing import TypeVarTuple, Unpack
+
 
 DimT = TypeVar("DimT", bound="Dimension")
 DimsT = TypeVar("DimsT", bound=Sequence["Dimension"])
+DimsTTup = TypeVarTuple("DimsTTup")
 DT = TypeVar("DT", bound="DType")
 
 
@@ -55,6 +58,10 @@ class DType:
 
 
 class Field(Generic[DimsT, DT]):
+    ...
+
+
+class Domain(Generic[Unpack[DimsTTup]]):  # type: ignore[misc] # mypy doesn't like Unpack
     ...
 
 
--- a/tests/functional_tests/ffront_tests/test_execution.py
+++ b/tests/functional_tests/ffront_tests/test_execution.py
@@ -20,6 +20,7 @@ from typing import TypeVar
 import numpy as np
 import pytest
 
+from functional.common import Domain
 from functional.ffront.decorator import field_operator, program
 from functional.ffront.fbuiltins import (
     Dimension,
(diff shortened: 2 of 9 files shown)


tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -452,3 +453,24 @@ def simple_broadcast(inp: Field[[IDim], float64]) -> Field[[IDim, JDim], float64
...
+
+    @field_operator
+    def copy(inp: Field[[IDim], float64]) -> Field[[IDim], float64]:
+        return inp
+
+    @program(backend="roundtrip")
+    def copy_program(dom: Domain[IDim], inp: Field[[IDim], float64], out: Field[[IDim], float64]):
+        copy(inp, out=out[dom])
+
+    domain = {IDim: range(1, 9)}
```
havogt, line 471: This is what I already used in embedded execution to represent a domain. Open to change that (now or later).

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -452,3 +453,24 @@ def simple_broadcast(inp: Field[[IDim], float64]) -> Field[[IDim, JDim], float64
...
+    size = 10
+    a = np_as_located_field(IDim)(np.ones((size)))
+    b = np_as_located_field(IDim)(np.zeros((size)))
+
+    @field_operator
+    def copy(inp: Field[[IDim], float64]) -> Field[[IDim], float64]:
+        return inp
+
+    @program(backend="roundtrip")
+    def copy_program(dom: Domain[IDim], inp: Field[[IDim], float64], out: Field[[IDim], float64]):
```
havogt, line 468: domain type is not checked. Should that go in type deduction?

src/functional/common.py
```diff
@@ -20,9 +20,12 @@
 from dataclasses import dataclass
 from typing import Any, Generic, TypeVar
 
+from eve.extended_typing import TypeVarTuple, Unpack
+
 
 DimT = TypeVar("DimT", bound="Dimension")
 DimsT = TypeVar("DimsT", bound=Sequence["Dimension"])
+DimsTTup = TypeVarTuple("DimsTTup")
```
havogt, line 28: Does this make sense?

havogt:
This PR does actually 2 changes: allow domain as program argument, allow domain in __getitem__ of out field. I am not sure if the former is a good idea...

tests/functional_tests/ffront_tests/test_execution.py
DropD: Seems fine for now. Only reason to change that I see would be to simplify type deduction?

havogt:
Passing domain as program argument is not needed for now, creating an explicit domain is implemented in #940

PR #821, opened 2022-06-15
Functional: Illustrate slice sparse

havogt:
Just for discussion. This enables slicing a sparse field in the neighbor dimension (which is required if we want to use a sparse field outside of reductions).

The syntax here is
```python
@field_operator
def sparse_op(sparse: Field[[Edge, E2VDim], "float64"]) -> Field[[Edge], "float64"]:
    return sparse[0]
```
which slices the last dimension (if it is a local dimension).

We should discuss what we need and if the dimension needs to be named, e.g. 
```python
    return sparse[E2VDim[0]]
```


havogt:
See discussion in #823

PR #822, opened 2022-06-15
Allow "shift" on sparse field

havogt:
Alternative to #821, less intrusive with 1 to 1 mapping to itir

Syntax
```python
@field_operator
def sparse_op(sparse: Field[[Edge, E2VDim], "float64"]) -> Field[[Edge], "float64"]:
    return sparse(0)
```


tehrengruber:
See discussion in https://github.com/GridTools/gt4py/issues/823

Issue #352, opened 2021-02-17
Round-trip to json fails due to enums


gronerl:
serializing gtc nodes to json and deserializing does not work
```Python
json = oir.json()
Stencil.parse_raw(json)
```
with erros such as
```
E   vertical_loops -> 4 -> loop_order
E     Invalid value type [expected: <enum 'LoopOrder'>, received: <class 'str'>] (type=type_error)
```
This seems to be the solution:
https://stackoverflow.com/questions/65209934/pydantic-enum-field-does-not-get-converted-to-string/65211727#65211727

this currently prevents saving and loading sdfgs.

tehrengruber:
Fixed in #612

PR #825, opened 2022-06-17
LazyStencil SDFG Interface (and Related Changes)

gronerl:
This PR is a collection of changes related to FV3core DaCe-orchestration
* add `__sdfg__` functionality to Lazy Stencils. 
  It does this in a way that avoids compilation of the LazyStencil if the LazyStencil is used in dace orchestration.
  Also includes some changes that improve the pipeline performance of this 
* change expansion priority to the one used in the SC paper (avoids some cuda 701 errors)
* change to/from_sdfg of StencilComputation nodes so that only its properties are pickled, not the entire node. (Needed for stability in DaCe transformations.)


gronerl:
bors try

gronerl:
bors try-
bors try

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try-
bors try

src/gtc/passes/gtir_pipeline.py
```diff
@@ -31,9 +32,15 @@ class GtirPipeline:
     May only call existing passes and may not contain any pass logic itself.
     """
 
-    def __init__(self, node: gtir.Stencil):
+    _cache: Dict[Tuple[StencilID, Tuple[PASS_T, ...]], gtir.Stencil] = {}
```
jdahm, line 35: Can you help me understand this change? It looks to me like a unique `GtirPipeline` object is created for each `StencilID` anyway, since one is made for each stencil.

src/gt4py/backend/module_generator.py
```diff
@@ -53,12 +54,17 @@ def parameter_names(self) -> Set[str]:
...
+
+
 def make_args_data_from_gtir(pipeline: GtirPipeline) -> ModuleData:
     """
     Compute module data containing information about stencil arguments from gtir.
 
     This is no longer compatible with the legacy backends.
     """
+    if pipeline.stencil_id in _args_data_cache:
+        return _args_data_cache[pipeline.stencil_id]
```
jdahm, line 67: Oh, is this the reason?

src/gtc/passes/gtir_pipeline.py
gronerl: Indeed, with the usual `gtscript.stencil` decorator, there is never more than just one gtir pipeline, since subsequent calls to the decorator load and return the previously generated module from file cache. However, with the `lazy_stencil` this is not the case: There is no cache on disk, yet e.g. in check_syntax or when accessing field_info, this triggers a pipeline each time. This would cache those cases so that for the same stencil (by hash, not by instance) the pipeline is only executed once.

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try-
bors try

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try

gronerl:
bors tr

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try-
bors try

src/gt4py/backend/dace_stencil_object.py
```diff
@@ -22,14 +22,32 @@
...
 from dace.frontend.python.common import SDFGClosure, SDFGConvertible
-from dace.sdfg.utils import inline_sdfgs
 
-import gt4py.backend as gt_backend
 from gt4py.definitions import AccessKind
 from gt4py.stencil_object import FrozenStencil, StencilObject
 from gt4py.utils import shash
 
 
+def add_optional_fields(sdfg, field_info, parameter_info, **kwargs):
```
jdahm, line 31:
```suggestion
def add_optional_fields(sdfg: dace.SDFG, field_info: Dict[str, Any], parameter_info: Dict[str, Any], **kwargs):
```

src/gt4py/caching.py
```diff
@@ -154,6 +156,10 @@ def class_name(self) -> str:
         """Calculate the name for the stencil class, default is to read from build options."""
         return self.builder.options.name
 
+    def capture_externals(self) -> Dict[str, Any]:
+        """Extract externals from the annotated stencil definition for fingerprinting. Freezes the references."""
```
jdahm, line 160: I would replace "Freezes the references" with something like "Returns a snapshot of the externals when called".

src/gt4py/caching.py
```diff
@@ -154,6 +156,10 @@ def class_name(self) -> str:
...
+        pass
```
jdahm, line 161: 👍

src/gt4py/dace_lazy_stencil.py
```diff
@@ -0,0 +1,63 @@
+from typing import TYPE_CHECKING, Any, Dict, Optional, Sequence, Set, Tuple
```
jdahm, line 1: I like that the `DaceStencilObject` is stored in the `backend` module. Can this be moved there too?

src/gt4py/gtscript.py
```diff
@@ -300,6 +305,7 @@ def lazy_stencil(
     build_info=None,
     dtypes=None,
     externals=None,
+    format_source=True,
```
jdahm, line 308: Missing doc below for this option.

src/gt4py/lazy_stencil.py
```diff
@@ -77,3 +78,18 @@ def __call__(self, *args: Any, **kwargs: Any) -> None:
     def run(self, *args: Any, **kwargs: Any) -> None:
         """Pass through to the implementation.run."""
         self.implementation.run(*args, **kwargs)
+
+    def __sdfg__(self, **kwargs):
```
jdahm, line 82: Does this need a `__deepcopy__`?

src/gt4py/stencil_builder.py
```diff
@@ -215,6 +219,7 @@ def with_externals(self: "StencilBuilder", externals: Dict[str, Any]) -> "Stenci
         """
         self._build_data = {}
         self._externals = externals
+        self.with_caching(self.caching.name)
```
jdahm, line 222: The `__init__` already set the caching strategy. Is this needed because of a quirk?

src/gt4py/lazy_stencil.py
```diff
@@ -39,6 +39,7 @@ class LazyStencil:
 
     def __init__(self, builder: "StencilBuilder"):
         self.builder = builder
+        self.builder.capture_externals()
```
jdahm, line 42: Could this be a call to `self.builder.caching.capture_externals` instead? That would avoid having to define it as a method in the builder as well.

src/gt4py/stencil_builder.py
```diff
@@ -201,6 +201,10 @@ def definition(self) -> AnnotatedStencilFunc:
             self.frontend.prepare_stencil_definition(self._definition, self.externals),
         )
 
+    def capture_externals(self) -> None:
+        """Extract externals from the annotated stencil definition for fingerprinting. Freezes the references."""
+        self.caching.capture_externals()
```
jdahm, line 206: See comment in lazy_stencil.py.

src/gt4py/backend/dace_stencil_object.py
```diff
@@ -249,8 +108,15 @@ def freeze(
         except FileNotFoundError:
             # otherwise, wrap and save sdfg from scratch
             inner_sdfg = self.sdfg()
-
-            frozen_sdfg = self._sdfg_freeze_domain_and_origin(inner_sdfg, domain, origin)
+            from gt4py.backend.dace_backend import freeze_origin_domain_sdfg
```
jdahm, line 111: Just curious: why is this here and not at the top of the file?

src/gt4py/stencil_object.py
```diff
@@ -245,9 +245,11 @@ def _make_origin_dict(origin: Any) -> Dict[str, Index]:
 
         raise ValueError("Invalid 'origin' value ({})".format(origin))
 
+    @staticmethod
     def _get_max_domain(
```
jdahm, line 249: Is my understanding correct that the reason for making this static is that it's used in `normalize_args` which is used in `__sdfg__`, and that dunder call needs to be static?

jdahm: If that's correct, I guess a follow-up question is: does this need to change if we only need them static in the `DaceStencilObject`, since they're not part of the public interface of the object?

src/gt4py/backend/dace_backend.py
```diff
@@ -11,13 +11,16 @@
 # distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
+import copy
```
jdahm, line 14:
```suggestion
# SPDX-License-Identifier: GPL-3.0-or-later

import copy
```

src/gtc/dace/expansion/daceir_builder.py
```diff
@@ -246,9 +248,18 @@ def visit_BinaryOp(self, node: oir.BinaryOp, **kwargs: Any) -> dcir.BinaryOp:
...
-        self, node: oir.HorizontalRestriction, **kwargs: Any
+        self, node: oir.HorizontalRestriction, *, symbol_collector, **kwargs: Any
     ) -> dcir.HorizontalRestriction:
-        return dcir.HorizontalRestriction(mask=node.mask, body=self.visit(node.body, **kwargs))
+        for axis, interval in zip(dcir.Axis.dims_horizontal(), node.mask.intervals):
+            for bound in (interval.start, interval.end):
+                if bound is not None:
+                    symbol_collector.add_symbol(axis.iteration_symbol())
+                    if bound.level == common.LevelMarker.END:
+                        symbol_collector.add_symbol(axis.domain_symbol())
```
jdahm, line 258:
```suggestion
            for bound in filter(None, (interval.start, interval.end)):
                symbol_collector.add_symbol(axis.iteration_symbol())
                if bound.level == common.LevelMarker.END:
                    symbol_collector.add_symbol(axis.domain_symbol())
```

src/gtc/dace/expansion/daceir_builder.py
```diff
@@ -246,9 +248,18 @@ def visit_BinaryOp(self, node: oir.BinaryOp, **kwargs: Any) -> dcir.BinaryOp:
         )
 
     def visit_HorizontalRestriction(
-        self, node: oir.HorizontalRestriction, **kwargs: Any
+        self, node: oir.HorizontalRestriction, *, symbol_collector, **kwargs: Any
```
jdahm, line 251: Add type info here.

tests/test_unittest/test_gtc/test_definitive_assignment_analysis.py
```diff
@@ -36,6 +36,16 @@ def _wrapper(definition):
...
 
 
+@pytest.fixture
+def clear_gtir_caches():
+    from gt4py.backend import module_generator
+    from gtc.passes.gtir_pipeline import GtirPipeline
+
+    GtirPipeline._cache.clear()
+    module_generator._args_data_cache.clear()
+    yield
```
jdahm, line 46: Is this here just to be careful?

src/gtc/dace/nodes.py
```diff
@@ -40,32 +40,58 @@ def _set_expansion_order(
...
+        protocol = pickle.DEFAULT_PROTOCOL
+        pbytes = pickle.dumps(obj, protocol=protocol)
+        jsonobj = dict(pickle=base64.b64encode(pbytes).decode("utf-8"))
+        return jsonobj
+
+    @classmethod
+    def from_json(cls, d, sdfg=None):
+        b64string = d["pickle"]
+        byte_repr = base64.b64decode(b64string)
+        return pickle.loads(byte_repr)
```
jdahm, line 54: Probaly for manual testing, but in case I wanted to call out that from a quick search in gt4py, these methods _seem_ to be neither used nor tested.

jdahm, line 54: Not something that directly affects this PR, but just a thought: is pickling the right approach for serializing SDFGs? AFAIK It's not safe between versions, and is a direct memory representation. Perhaps something like JSON directly is better?

tests/test_integration/test_dace_parsing.py
```diff
@@ -204,28 +229,22 @@ def stencil(
...
-@pytest.mark.parametrize(
-    "backend",
-    ["dace:cpu", pytest.param("dace:gpu", marks=[pytest.mark.requires_gpu])],
-)
-def test_optional_arg_provide(backend):
-    @gtscript.stencil(backend=backend)
+def test_optional_arg_provide(decorator):
+    backend = "dace:cpu"
+
+    @decorator(backend=backend)
```
jdahm, line 247: No more GPU support?

src/gtc/dace/expansion/expansion.py
```diff
@@ -133,7 +133,14 @@ def expansion(
...
 
-        nsdfg = StencilComputationSDFGBuilder().visit(daceir)
+        nsdfg: dace.nodes.NestedSDFG = StencilComputationSDFGBuilder().visit(daceir)
+
+        delkeys = set()
+        for sym in node.symbol_mapping.keys():
+            if str(sym) not in nsdfg.free_symbols:
+                delkeys.add(str(sym))
+        for key in delkeys:
+            del node.symbol_mapping[key]
```
jdahm, line 143: The rest of this function is high level function calls. Would be clean if this is in a separate funtion with a descriptive name. :)

jdahm (commented):
This is looking good! I left a few questions and comments.

src/gt4py/stencil_object.py
gronerl: `__sdfg__` doesn't have to be static, but the `__sdfg__` method of DaCeLazyStencil can't access the `self.implementation` in `__sdfg__`, since that would trigger a build. So we _must_  access `normalize_args` through the type instead of the instance.

src/gtc/dace/nodes.py
gronerl: Yes, this is a workaround for a limitation similar to https://github.com/GridTools/gt4py/issues/352 (AFAIK still not possible) even if the particular example in the issue may be fixed.

tests/test_unittest/test_gtc/test_definitive_assignment_analysis.py
gronerl: Since the GtirPipeline is now cached per StencilID, not per GtirPipeline instance, only the first occurance of the stencil would print all the warnings / raise exceptions. This fixture resets the cache in between so that the conditions for the tests below are met.

src/gt4py/backend/dace_stencil_object.py
gronerl: I am sometimes lazy when writing code. Came back like a boomerang.

src/gt4py/stencil_builder.py
gronerl: After changing externals, we need to avoid using a previous cache built around previous externals, so we need to reset it.

src/gt4py/lazy_stencil.py
gronerl: You mean the class needs a custom deep copy functionality? Why so?

src/gtc/dace/expansion/expansion.py
gronerl: I moved it to the above `_fix_context` which I think is a suitable place, and added a comment.

gronerl:
This is ready for another review pass. Thanks!

src/gt4py/lazy_stencil.py
jdahm: Ah, because `StencilObject` has one.

jdahm: Nevermind, I understand now. All good!

jdahm:
I wonder if we could make the `stencil_id` generation a method on the `stencil_ir` in order to make it consistent, since `gtir` should be one-to-one with the stencil definitions in `gtscript`. That should not block this PR because it's orthogonal. I made an issue to track it: https://github.com/GridTools/gt4py/issues/904.

tests/test_integration/test_dace_parsing.py
```diff
@@ -358,13 +365,47 @@ def numpy_stencil(inp: gtscript.Field[np.float64], outp: gtscript.Field[np.float
...
             "Only dace backends are supported in DaCe-orchestrated programs." ' (found "numpy")'
         ),
     ):
-        call_frozen_stencil()
+        call_stencil()
+
+
+from gt4py.backend.dace_lazy_stencil import (  # noqa: E402 (needs to be guarded by above importorskip)
+    DaCeLazyStencil,
+)
```
jdahm, line 384: Is this here instead of at the top of the file for clarity, or is there another reason?

jdahm (approved):
Looks good! Thanks for taking the time to address the review. I haven't run this through Pace but I assume someone else has and this is compatible. @FlorianDeconinck?

gronerl:
bors try

tests/test_integration/test_dace_parsing.py
gronerl: missed that, I think.

gronerl:
bors try-
bors try

PR #830, opened 2022-06-21
Backend cleanup

DropD:
## Description

Make a separation between "Backend"s that execute a fencil and those that create / print a string.

It is undisputed that it is useful to be able to have a unified interface for testing. However for actual use it seems more useful that calling a fencil carries the semantics of executing the fencil and nothing else.

This PR separates into two interfaces, `execute_fencil` and `format_fencil`. For convenience the return type of the `@fencil` decorator calls the former in it's `__call__` method and the latter in it's `string_format` method.

A unified method for calling the appropriate one is introduced inside the tests.

### Naming
The concepts are currently named as follows (better ideas welcome):

- `fencil_processor`: either a
  - `fencil_executor` (executes a fencil)
  - `fencil_formatter` (generates text from a fencil)

### Rejected names
The name "backend" was avoided because what is commonly known as a backend may actually be comprised of a formatter and executor, or only an executor, or in the future (compiled backends integration project) something that feeds into a whole bindings > compile > load pipeline.

"fencil_util" / "fencil_tool" was considered instead of "fencil_processor" but didn't seem a significant improvement.

"fencil_generator" or similar was considered instead of "fencil_formatter" but seemed misleading when considering pretty printing.

## Caveats

The ADR & changes to src/functional/iterator/ARCHITECTURE.md are still missing.

## Requirements


src/functional/iterator/processor_interface.py
```diff
@@ -0,0 +1,62 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+import enum
+import typing
+
+from functional.iterator import ir as itir
+
+
+class ProcessorKind(enum.Enum):
+    EXECUTOR = 0
+    FORMATTER = 1
```
egparedes, line 22: I understand the decision to skip the definition a hierarchy of `Processor`s for now until it is clear that is needed, but the current solution doesn't look much simpler to me. Tagging every processor with a  `ProcessorKind` value is basically the same as defining a Protocol but the definition of their interfaces is not to easy to find (for example, inside the `run_processor()` function in tests). Why do you think is better to define the enum instead of single-method `Protocol`s for each one of the kinds?

egparedes (commented):
The PR looks good, although it took me some effort to review since it touches so many files (which is, of course, unavoidable for the goal of the PR).

I have some minor suggestions and a couple of questions.

src/functional/ffront/decorator.py
```diff
@@ -86,7 +94,7 @@ def _collect_capture_vars(captured_vars: CapturedVars) -> CapturedVars:
 
 
 @typing.runtime_checkable
-class GTCallable(Protocol):
+class GTCallable(typing.Protocol):
```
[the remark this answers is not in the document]
DropD: Do you have any Idea why `mypy` complained about `Protocol` from `extended_typing` but not about `Protocol` from `typing`? It should be the same class, right?

src/functional/iterator/processor_interface.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class Executor(typing.Protocol):
```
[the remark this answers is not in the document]
DropD: Don't module namespaces solve that problem already?

src/functional/ffront/dialect_parser.py
```diff
@@ -61,15 +61,15 @@ class DialectParser(ast.NodeVisitor, Generic[DialectRootT]):
     captured_vars: CapturedVars
     externals_defs: dict[str, Any]
 
-    syntax_error_cls: ClassVar[Type[DialectSyntaxError]]
+    syntax_error_cls: ClassVar[Type[DialectSyntaxError]] = "DialectSyntaxError"
```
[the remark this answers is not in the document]
DropD: I only put them because mypy wasn't letting me commit otherwise. Any idea why?

src/functional/iterator/runtime.py
```diff
@@ -27,6 +36,48 @@ def offset(value):
...
+            debug(fencil_definition)
+        return fencil_definition
+
+    def __call__(self, *args, backend: Optional[Executor] = None, **kwargs):
+        args, kwargs = self._rewrite_args(args, kwargs)
+
+        if backend is not None:
+            ensure_executor(backend)
+            if fendef_codegen is None:
+                raise RuntimeError("Backend execution is not registered")
```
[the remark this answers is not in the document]
DropD: I would say yes, but I see how this and the equivalent block below are confusing. This relates to the `iterator` internal registration of `iterator.tracing` for turning a function into IR and `iterator.embedded` for embedded execution.

It is correct, but we might want to think about if it should be an `InternalError` instead and whether the message is clear enough.

src/functional/iterator/processor_interface.py
DropD: I have tried a protocol for each. The result is that they are indistinguishable at runtime and you have to add something to distinguish them at runtime anyway. To clarify:

```python
@runtime_checkable
Foo(Protocol):
    def __call__(fencil: FencilDefinition, *args, **kwargs) -> None:
        ...


@runtime_checkable
Bar(Protocol):
    def __call__(fencil: FencilDefinition, *args, **kwargs) -> str:
        ...


# runtime checking does not check any type annotations only existence of the attributes and methods.
isinstance(roundtrip.executor, Foo) == True
isinstance(roundtrip.executor, Bar) == True
```

As you can see `ProcessorKind` is necessary anyway. If the additional protocol seems worth it to you we can add it.

src/functional/iterator/ARCHITECTURE.md
```diff
@@ -15,7 +15,7 @@ The builtins dispatcher is implemented in `dispatcher.py`. Implementations are r
 
 `fundef` returns a wrapper around the function, which dispatches `__call__` to a hook if a predicate is met (used for *tracing*). By default the original function is called (used in *embedded* mode).
```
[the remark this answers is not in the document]
DropD: It was there since the dawn of the prototype... Somebody who actually understands the code should transform it into an ADR.

src/functional/iterator/runtime.py
```diff
@@ -27,6 +36,48 @@ def offset(value):
...
+            backend(self.itir(*args, **kwargs), *args, **kwargs)
+        else:
+            if fendef_embedded is None:
+                raise RuntimeError("Embedded execution is not registered")
```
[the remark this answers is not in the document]
DropD: basically this is a type guard for the `fendef_embedded()` call below. That is set to `None` initially but `iterator.embedded` assigns a callable to it ("registering" the embedded execution).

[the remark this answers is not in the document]
DropD: This means it is the "correct" error message in the sense that it is the same Error message under the same conditions as before this PR.

src/functional/iterator/processor_interface.py
DropD: A passable alternative is to create wrapper types. I have made that change now and it looks cleaner to me, what do you think?

src/functional/iterator/processor_interface.py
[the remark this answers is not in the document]
DropD: Changed to `FencilExecutor`

src/functional/iterator/processor_interface.py
```diff
@@ -0,0 +1,62 @@
...
+    processor_kind: ProcessorKind
+
+    def __call__(fencil: itir.FencilDefinition, *args, **kwargs) -> None:
+        ...
+
+
+class Formatter(typing.Protocol):
```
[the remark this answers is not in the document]
DropD: Changed to `FencilFormatter`.

src/functional/iterator/processor_interface.py
egparedes: Ok, I see your point. Somehow I hadn't consider this case where all the protocols would use `__call__` for implementation. I had in mind that each protocol defined a different entry point, which would simplify thinks. For example:

```python
@runtime_checkable
class Executor(Protocol):
    def __call__(fencil: FencilDefinition, *args, **kwargs) -> None:
    # or: def execute(fencil: FencilDefinition, *args, **kwargs) -> None:
        ...


@runtime_checkable
class Formatter(Protocol):
    def format_ir(fencil: FencilDefinition, *args, **kwargs) -> str:
        ...
...

```
What do you think about that?

src/functional/ffront/dialect_parser.py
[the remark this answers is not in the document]
DropD: Actually, doesn't that simply assign a string? Why is mypy okay with that?

src/functional/iterator/processor_interface.py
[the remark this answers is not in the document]
egparedes: Yes, this is just an optional comment. Feel free to ignore if you think is irrelevant.

src/functional/ffront/decorator.py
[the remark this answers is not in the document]
egparedes: Hmmm, I also thought it should be the same class but recently read that might be a difference if it gets imported from `typing_extensions` (which I don't understand why should happen in python3.10). I will try to investigate it. In the meantime, use the one that makes your life simpler.

src/functional/iterator/runtime.py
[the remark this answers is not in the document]
egparedes: Ok, fine for me if you also think it's fine.

src/functional/iterator/processor_interface.py
DropD: In that case I would still want to write the decorators to allow writing a simple function and decorating it. Amounts to almost the same in terms of usability as the current improved state, except for the usage when you start from a `FencilDefinition`. I never minded that you can call `formatter(fencil_ir, *args, **kwargs)`, only that you can call `fencil(*args, backend=formatter, **kwargs)`. The latter violates sane fencil call semantics, the former seems harmless. Your proposal would replace it with `formatter.format_ir(fencil_ir, *args, **kwargs)`.

src/functional/ffront/dialect_parser.py
```diff
@@ -61,7 +61,7 @@ class DialectParser(ast.NodeVisitor, Generic[DialectRootT]):
     captured_vars: CapturedVars
     externals_defs: dict[str, Any]
 
-    syntax_error_cls: ClassVar[Type[DialectSyntaxError]] = "DialectSyntaxError"
+    syntax_error_cls: ClassVar[Type[DialectSyntaxError]] = field(default=lambda: DialectSyntaxError)
```
egparedes, line 64: I don't think this works as expected, since class members which are `ClassVar`s are not processed by dataclasses. I don't know if I understand the intended purpose here, but if `syntax_error_cls` can be a different class in different subclasses, I think there should be another `ErrorT` TypeVar (`ErrorT = TypeVar(ErrorT, bound=DialectSyntaxError)`

and then:
```
class DialectParser(ast.NodeVisitor, Generic[RootT, ErrorT]):
    ...
    syntax_error_cls: ClassVar[Type[ErrorT]] = DialectSyntaxError
```

Does any of this make sense to you?

src/functional/ffront/decorator.py
```diff
@@ -231,6 +239,45 @@ def _validate_args(self, *args, **kwargs) -> None:
...
+        rewritten_args, size_args, kwargs = self._process_args(args, kwargs)
+
+        backend = self.backend
+        if not self.backend:
+            warnings.warn(
+                UserWarning(
+                    f"Field View Program '{self.itir.id}': Using default ({DEFAULT_BACKEND}) backend."
+                )
+            )
+            backend = DEFAULT_BACKEND
```
egparedes, line 251: It looks cleaner to me this way, but it's just a suggestion. Feel free to ignore:
```suggestion
        if not self.backend:
            warnings.warn(
                UserWarning(
                    f"Field View Program '{self.itir.id}': Using default ({DEFAULT_BACKEND}) backend."
                )
            )
            backend = DEFAULT_BACKEND
        else:
            backend = self.backend
```

egparedes (commented):
It looks ready to go. Just one open question about typing the DialectParser.

src/functional/ffront/dialect_parser.py
DropD: It does make sense but the intention is that error types in subclasses have to be subclasses of `DialectSyntaxError`. Isn't that already expressed here?

The reason for assigning the base exception class here is that otherwise mypy complains that the dataclass constructor isn't passed the mandatory keyword arg `syntax_error_cls` somewhere down the line. I could also just silence that mypy error 🤷 .

src/functional/ffront/decorator.py
DropD: the suggestion seems incomplete to me, it would leave the following state:

```python
backend = self.backend
if not self.backend:
    ...
    backend = DEFAULT_BACKEND
else:
    backend = self.backend  # <-- seems pointless
```

Did you mean the following?

```python
# backend undefined at this point
if self.backend:
    backend = self.backend
else:
    ...
    backend = DEFAULT_BACKEND
```

I didn't do it like that purely out of personal preference.

src/functional/ffront/dialect_parser.py
egparedes: Yes, the typing annotation here expresses exactly what you say, but I'm confused with the `= field(...)` assignment, since a `ClassVar` should be ignored by `dataclasses` and it will be never requested at initialization.

Shouldn't be enough to move the definition of `DialectSyntaxError` in front of `DialectParser` so this works as it should:
```
syntax_error_cls: ClassVar[Type[DialectSyntaxError]] = DialectSyntaxError
```

DropD: Yes, that is the other solution. Let's go with that.

src/functional/ffront/decorator.py
```diff
@@ -231,6 +239,44 @@ def _validate_args(self, *args, **kwargs) -> None:
...
+
+        backend(
+            self.itir,
+            *rewritten_args,
+            *size_args,
+            **kwargs,
+            offset_provider=offset_provider,
+        )
+
+    def format_itir(
```
tehrengruber, line 264: I think this does not belong into the decorator, but in a dedicated utility function. It's sole tasks is (and believe should be) to create a callable object. The FieldOperator class is already too heavy right now.

DropD: De facto the signature of the callable object it creates and owns is an implementation detail of `Program`, so it's responsibility is also to be a facade to that callable.

If we want to free it from this facade responsibility, then it's callable object has to own the `_process_args` functionality.

DropD: ... After due consideration, I think it is not a refactoring that fits into this scope. So I added a TODO.

PR #832, opened 2022-06-22
Functional: embedded execution position/offset based on str not obj

havogt:
Iterator position in embedded execution is now a `dict[str, int]`. Offsets are translated directly from `Offset` object to `str`.
This avoids that embedded execution accidentally uses extra information from `Offset` or `Dimension`. This information needs to be provided via `offset_provider`.

Additionally:
- adds type hints

### DIFF
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -31,9 +31,8 @@ EMBEDDED = "embedded"
 
 
 Position = dict[str, Union[tuple[Optional[int], ...], Optional[int]]]
-MaybePosition = Optional[
-    Position
-]  # if a neighbor-table lookup results in not-a-neighbor, we set the position to None
+#: A ``None`` position flags invalid not-a-neighbor results in neighbor-table lookups
+MaybePosition: TypeAlias = Optional[Position]  
 AnyOffset = str | int
 OffsetProvider = dict[str, Any]
 
@@ -266,8 +265,7 @@ def execute_shift(
 ) -> MaybePosition:
     assert pos is not None
     if tag in pos and pos[tag] is None:  # sparse field with offset as neighbor dimension
-        new_pos = pos.copy()
-        new_pos[tag] = index
+        new_pos = pos | {tag: index}    
         return new_pos
     assert tag in offset_provider
     offset_implementation = offset_provider[tag]
@@ -281,7 +279,7 @@ def execute_shift(
     elif isinstance(offset_implementation, NeighborTableOffsetProvider):
         assert offset_implementation.origin_axis.value in pos
         new_pos = pos.copy()
-        del new_pos[offset_implementation.origin_axis.value]
+        new_pos.pop(offset_implementation.origin_axis.value)
         if offset_implementation.tbl[pos[offset_implementation.origin_axis.value], index] is None:
             return None
         else:
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -8,7 +8,6 @@ from typing import (
     Iterable,
     Optional,
     Protocol,
-    Tuple,
     TypeVar,
     Union,
     runtime_checkable,
@@ -27,7 +26,7 @@ from functional.iterator.utils import tupelize
 EMBEDDED = "embedded"
 
 
-PosT = dict[str, Union[Tuple[Optional[int], ...], Optional[int]]]
+PosT = dict[str, Union[tuple[Optional[int], ...], Optional[int]]]
 MaybePosT = Optional[PosT]
 OffsetT = Union[str, int]
 OffsetProviderT = dict[str, Any]
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -10,15 +10,15 @@ from typing import (
     Optional,
     Protocol,
     Sequence,
+    TypeAlias,
     TypeGuard,
     TypeVar,
     Union,
-    cast,
     runtime_checkable,
 )
 
 import numpy as np
-import numpy.typing
+import numpy.typing as npt
 
 from functional import iterator
 from functional.common import Dimension
@@ -30,15 +30,21 @@ from functional.iterator.utils import tupelize
 EMBEDDED = "embedded"
 
 
-Position = dict[str, Union[tuple[Optional[int], ...], Optional[int]]]
+Position: TypeAlias = dict[str, Union[tuple[Optional[int], ...], Optional[int]]]
 #: A ``None`` position flags invalid not-a-neighbor results in neighbor-table lookups
-MaybePosition: TypeAlias = Optional[Position]  
-AnyOffset = str | int
-OffsetProvider = dict[str, Any]
+MaybePosition: TypeAlias = Optional[Position]
+AnyOffset: TypeAlias = str | int
+OffsetProvider: TypeAlias = dict[str, Any]
 
 
 @runtime_checkable
 class ItIterator(Protocol):
+    """
+    Prototype for the Iterator concept of Iterator IR.
+
+    `ItIterator` to avoid name clashes with `Iterator` from `typing` and `collections.abc`.
+    """
+
     def shift(self, *offsets: AnyOffset) -> ItIterator:
         ...
 
@@ -74,7 +80,7 @@ class AssignableLocatedField(LocatedField):
 class NeighborTableOffsetProvider:
     def __init__(
         self,
-        tbl: np.ndarray,  # TODO(havogt): define neighbor table concept
+        tbl: npt.NDArray,  # TODO(havogt): define neighbor table concept
         origin_axis: Dimension,
         neighbor_axis: Dimension,
         max_neighbors: int,
@@ -250,22 +256,19 @@ def named_range_(axis: str, range_: Iterable[int]) -> Iterable[tuple[str, int]]:
     return ((axis, i) for i in range_)
 
 
-def domain_iterator(domain: dict[str, range]) -> Sequence[Position]:
-    return cast(
-        Sequence[Position],
-        (
-            dict(elem)
-            for elem in itertools.product(*(named_range_(tup[0], tup[1]) for tup in domain.items()))
-        ),
+def domain_iterator(domain: dict[str, range]) -> Iterable[Position]:
+    return (
+        dict(elem)
+        for elem in itertools.product(*(named_range_(axis, rang) for axis, rang in domain.items()))
     )
 
 
 def execute_shift(
-    pos: MaybePosition, tag: str, index: int, *, offset_provider: OffsetProvider
+    pos: Position, tag: str, index: int, *, offset_provider: OffsetProvider
 ) -> MaybePosition:
     assert pos is not None
     if tag in pos and pos[tag] is None:  # sparse field with offset as neighbor dimension
-        new_pos = pos | {tag: index}    
+        new_pos = pos | {tag: index}
         return new_pos
     assert tag in offset_provider
     offset_implementation = offset_provider[tag]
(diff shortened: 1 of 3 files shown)


src/functional/iterator/embedded.py
```diff
@@ -245,18 +304,20 @@ def group_offsets(*offsets):
...
 
 
-def shift_position(pos, *complete_offsets, offset_provider):
+def shift_position(
+    pos: MaybePosT, *complete_offsets: tuple[str, int], offset_provider: OffsetProviderT
+) -> MaybePosT:
     if pos is None:
         return None
-    new_pos = pos.copy()
+    new_pos: MaybePosT = pos.copy()
```
egparedes, line 312:
```suggestion
    new_pos: Pos = pos.copy()
```

src/functional/iterator/embedded.py
```diff
@@ -311,15 +372,21 @@ def _setup_math_operations(cls):
...
+        column_axis=None,
     ) -> None:
         self.field = field
         self.pos = pos
         self.incomplete_offsets = incomplete_offsets or []
         self.offset_provider = offset_provider
         self.column_axis = column_axis
 
-    def shift(self, *offsets):
+    def shift(self, *offsets: OffsetT):
```
egparedes, line 389: Missing return type hint

src/functional/iterator/embedded.py
```diff
@@ -332,9 +399,9 @@ def shift(self, *offsets):
     def max_neighbors(self):
```
egparedes, line 399: Missing type hint. Also, would it make more sense to use a `@property` here?

src/functional/iterator/embedded.py
```diff
@@ -332,9 +399,9 @@ def shift(self, *offsets):
...
         assert self.incomplete_offsets
         assert isinstance(
-            self.offset_provider[self.incomplete_offsets[0].value], NeighborTableOffsetProvider
+            self.offset_provider[self.incomplete_offsets[0]], NeighborTableOffsetProvider
         )
-        return self.offset_provider[self.incomplete_offsets[0].value].max_neighbors
+        return self.offset_provider[self.incomplete_offsets[0]].max_neighbors
 
     def can_deref(self):
```
egparedes, line 406: Missing type hint. Also, would it make more sense to use a `@property` here?

src/functional/iterator/embedded.py
```diff
@@ -369,16 +436,23 @@ def deref(self):
             return _UNDEFINED
 
 
-def make_in_iterator(inp, pos, offset_provider, *, column_axis):
+def make_in_iterator(
```
egparedes, line 439: Missing type hints: inp, column_axis

src/functional/iterator/embedded.py
```diff
@@ -399,18 +473,22 @@ def make_in_iterator(inp, pos, offset_provider, *, column_axis):
...
-    """A Field with named dimensions/axises.
+class LocatedFieldImpl:
+    """A Field with named dimensions/axes.
 
     Axis keys can be any objects that are hashable.
     """
 
-    def __init__(self, getter, axises, dtype, *, setter=None, array=None):
+    @property
+    def axes(self):
```
egparedes, line 486: Missing type hint

src/functional/iterator/embedded.py
```diff
@@ -468,15 +547,15 @@ def combine_slice(s, t):
     return tuple(combine_slice(*i) for i in zip(a, b))
 
 
-def np_as_located_field(*axises, origin=None):
+def np_as_located_field(*axes: Dimension, origin=None):
```
egparedes, line 550: Missing origin and return type hints

src/functional/iterator/embedded.py
```diff
@@ -485,35 +564,35 @@ def setter(indices, value):
...
+        return it.shift(*list(o.value if isinstance(o, Offset) else o for o in offsets))
 
     return impl
 
 
 @dataclass
 class Column:
-    axis: CartesianAxis
+    axis: str
     range: range  # noqa: A003
```
egparedes, line 591: Can `range` type hint be changed to `Iterable[int]` here?

src/functional/iterator/embedded.py
```diff
@@ -1,23 +1,75 @@
 import itertools
```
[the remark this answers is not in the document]
havogt: Why is this needed?

src/functional/iterator/embedded.py
```diff
@@ -1,23 +1,75 @@
...
 from functional.iterator.utils import tupelize
 
 
 EMBEDDED = "embedded"
 
 
+PosT = dict[str, Union[tuple[Optional[int], ...], Optional[int]]]
+MaybePosT = Optional[PosT]
+OffsetT = Union[str, int]
+OffsetProviderT = dict[str, Any]
```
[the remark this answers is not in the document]
havogt: I kept MaybePos for now and used the place of definition to add a comment: # if a neighbor-table lookup results in not-a-neighbor, we set the position to None

src/functional/iterator/embedded.py
```diff
@@ -1,23 +1,75 @@
...
+
+
+@runtime_checkable
+class ItIterator(Protocol):
```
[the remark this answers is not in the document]
havogt: Do you have a suggestion? I don't see what information I would add that is not already in the definition. Or is that needed to generate documentation?

src/functional/iterator/embedded.py
```diff
@@ -1,23 +1,75 @@
...
+class AssignableLocatedField(LocatedField):
+    def __setitem__(self, indices, value) -> None:
+        ...
+
+
 class NeighborTableOffsetProvider:
     def __init__(
-        self, tbl, origin_axis, neighbor_axis, max_neighbors, has_skip_values=True
+        self,
+        tbl,
```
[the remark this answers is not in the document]
havogt: How to do ndarray with 2 dimensions and type int? I find it surprisingly hard to google that information.

src/functional/iterator/embedded.py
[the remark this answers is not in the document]
havogt: Oh, saw the comment later...

src/functional/iterator/embedded.py
```diff
@@ -183,38 +237,43 @@ def less(first, second):
...
-        assert offset_implementation in pos
+    assert tag in offset_provider
+    offset_implementation = offset_provider[tag]
+    if isinstance(offset_implementation, Dimension):
+        assert offset_implementation.value in pos
         new_pos = pos.copy()
-        new_pos[offset_implementation] += index
+        val = new_pos[offset_implementation.value]
+        assert isinstance(val, int)  # make mypy happy
+        new_pos[offset_implementation.value] = val + index
```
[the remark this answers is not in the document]
havogt: That was my initial solution, but mypy didn't understand it. Do you think it should work?

src/functional/iterator/embedded.py
havogt: Yes, I wanted to fix that, but just like this mypy doesn't like it. Maybe we can have a look together.

src/functional/iterator/embedded.py
havogt: As far as I can see `@property` with `Protocol` has the disadvantage that you cannot do `issubclass` checks (though, I currently have no use-case other than testing an iterator implementation).

src/functional/iterator/embedded.py
havogt: I don't think it would be correct. It should be a contiguous sequence of ints.

src/functional/iterator/embedded.py
```diff
@@ -669,16 +748,24 @@ def fendef_embedded(fun, *args, **kwargs):  # noqa: 536
         raise RuntimeError("offset_provider not provided")
 
     @iterator.runtime.closure.register(EMBEDDED)
-    def closure(domain, sten, out, ins):  # domain is Dict[axis, range]
+    def closure(
+        domain_: dict[Union[str, Dimension], range],
+        sten,
```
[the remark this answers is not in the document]
havogt: `Callable[..., Any]` or can we make it more precise?

src/functional/iterator/embedded.py
```diff
@@ -245,18 +317,20 @@ def group_offsets(*offsets):
...
+) -> MaybePosition:
     if pos is None:
         return None
-    new_pos = pos.copy()
+    new_pos: MaybePosition = pos.copy()  # becomes MaybePos later
     for tag, index in complete_offsets:
         new_pos = execute_shift(new_pos, tag, index, offset_provider=offset_provider)
         if new_pos is None:
             return None
     return new_pos
```
egparedes, line 330: I'm not sure, byt maybe `mypy` accepts the following alternative implementation:
```suggestion
    new_pos = pos.copy()
    for tag, index in complete_offsets:
        if (shifted_pos := execute_shift(new_pos, tag, index, offset_provider=offset_provider)) is not None:
            new_pos = shifted_pos
        else:
            return None
    return new_pos
```

src/functional/iterator/embedded.py
```diff
@@ -697,13 +821,15 @@ def closure(domain, sten, out, ins):  # domain is Dict[axis, range]
             res = sten(*ins_iters)
 
             if column is None:
-                ordered_indices = get_ordered_indices(out.axises, pos)
+                assert _is_position_fully_defined(pos)
+                ordered_indices = get_ordered_indices(out.axes, pos)
                 out[ordered_indices] = res
             else:
                 colpos = pos.copy()
```
egparedes, line 828: Maybe?
```suggestion
                col_pos = pos.copy()
```

src/functional/iterator/embedded.py
```diff
@@ -183,39 +247,47 @@ def less(first, second):
...
+        assert isinstance(_extracted_val_for_mypy_check, int)
+        new_pos[offset_implementation.value] = _extracted_val_for_mypy_check + index
         return new_pos
     elif isinstance(offset_implementation, NeighborTableOffsetProvider):
-        assert offset_implementation.origin_axis in pos
+        assert offset_implementation.origin_axis.value in pos
         new_pos = pos.copy()
-        del new_pos[offset_implementation.origin_axis]
-        if offset_implementation.tbl[pos[offset_implementation.origin_axis], index] is None:
+        del new_pos[offset_implementation.origin_axis.value]
```
egparedes, line 284: For readability I prefer to use `dict.pop()`, but it's mostly a question of taste, so feel free to ignore.
```suggestion
        new_pos.pop(offset_implementation.origin_axis.value)
```

egparedes (commented):
It looks better now. I added some extra comments and suggestions.

_Alternatively_, I propose to refactor the code using type annotations extensively to show the high-level concepts from ITIR, instead of just describing the types used in the implementation. 

I understand that this change might be outside the scope of this PR, so it's not a requirement to merge this one, but I think it will greatly improve the readability and quality of the code. For example (names and definitions are just guesses):

```python
# Atoms
Tag: TypeAlias = str
IntIndex: TypeAlias = int

# Offsets
OffsetPart: TypeAlias = Tag | IntIndex  #  PartialOffset ?
ConcreteOffset: TypeAlias = tuple[Tag, IntIndex]  # CompleteOffset ?
OffsetProviderImpl: TypeAlias = Dimension | NeighborTableOffsetProvider
OffsetProvider: TypeAlias = dict[Tag, OffsetProviderImpl]

# Positions
ConcretePosition: TypeAlias = dict[Tag, IntIndex]
PartialPosition: TypeAlias = dict[Tag, Optional[IntIndex]]
# WTFPosition: TypeAlias = dict[Tag, tuple[Optional[IntIndex], ...]] ??
# BrokenPosition: TypeAlias = dict[Tag, Optional[tuple[Optional[IntIndex], ...]]] ??

Position: TypeAlias = ConcretePosition | PartialPosition | ...
MaybePosition: TypeAlias = Optional[Position]

```

Having separate aliases for each one of the position kinds favors readability since the code will emphasize the concepts behind the Python implementation details. Using custom `TypeGuards` for these concepts will also simplify mypy's analysis since we they can inform mypy that certain `Position` variables only hold `int`s as values (i.e. is a `ConcretePosition`) and therefore most of the mypy hacks like `_extracted_val_for_mypy_check` (inside `execute_shift()`) could be avoided.

src/functional/iterator/embedded.py
```diff
@@ -72,6 +139,8 @@ def lift(stencil):
     def impl(*args):
         class wrap_iterator:
             def __init__(self, *, offsets=None, elem=None) -> None:
+                if offsets:
+                    assert all(isinstance(o, (int, str)) for o in offsets)
```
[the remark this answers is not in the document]
egparedes: Just a minor suggestion.

egparedes (approved):
It looks good to me. The major refactoring of the annotations can be postponed to a different PR.

PR #828, opened 2022-06-21
Functional: Improvements and fixes around sparse fields

havogt:
Embedded (added tests and fixed implementation):
- slice a shifted sparse field
- shift a sliced sparse field
- multiple sparse dimensions

gtfn:
- wip: pass to correctly lower the above examples

PR #836, opened 2022-06-24
Functional: extra dimensions as tuple and sparse field improvements

havogt:
- Interprets axes that are marked as `None` as tuple axes.
- Support more than one sparse dimension (extracted embedded changes from #828)

Other cleanups


havogt:
Split into several PRs, the first is #838

PR #840, opened 2022-06-27
Functional: Sparse fields: support multiple sparse dimensions and more tests

havogt:


tests/functional_tests/iterator_tests/test_toy_connectivity.py
```diff
@@ -273,7 +372,7 @@ def shift_shift_stencil2(inp):
 
 @fundef
 def shift_sparse_stencil2(inp):
-    return deref(shift(1, 3)(shift(V2E)(inp)))
+    return deref(shift(3, 1)(shift(V2E)(inp)))
```
havogt, line 375: Please check this change carefully! Is it correct now? The change is here https://github.com/GridTools/gt4py/pull/840/files#diff-cb3ad20072297c872c4fcb6245750c91cc9aa6080da28e19b1495f78e3d9b67cR374

src/functional/iterator/embedded.py
```diff
@@ -318,16 +324,25 @@ def execute_shift(
...
+
+        cur_pos = new_pos[tag]
+        assert isinstance(cur_pos, list)
+        for i, p in reversed(
+            list(enumerate(cur_pos))
+        ):  # first shift applies to the last sparse dimensions of that axis type
+            if p is None:
+                cur_pos[i] = index
+                break
         return new_pos
```
egparedes, line 338:
```suggestion
        assert isinstance(pos[tag], list)
        new_entry = list(pos[tag])
        # assert None in new_entry  # Fail instead of ignore too many partial shifts
        for i, p in reversed(enumerate(new_entry)):
            # first shift applies to the last sparse dimensions of that axis type
            if p is None:
                new_entry[i] = index
                break
        return pos | {tag: new_entry}
```

src/functional/iterator/embedded.py
```diff
@@ -440,8 +455,13 @@ def _setup_math_operations(cls):
 _UNDEFINED = Undefined()
 
 
-def _is_position_fully_defined(pos: Position) -> TypeGuard[ConcretePosition]:
-    return all(isinstance(v, int) for v in pos.values())
+def _is_position_fully_defined(
```
egparedes, line 458: Wouldn't it be simpler to call the function `_is_concrete_position()` ?

src/functional/iterator/embedded.py
```diff
@@ -528,17 +548,17 @@ def make_in_iterator(
             # we just use the name of the axis to match the offset literal for now
             sparse_dimensions.append(axis.value)
 
-    assert len(sparse_dimensions) <= 1  # TODO multiple is not a current use case
     new_pos: Position = pos.copy()
-    for sparse_dim in sparse_dimensions:
-        new_pos[sparse_dim] = None  # type: ignore[assignment] # looks like mypy is confused
+    for sparse_dim in set(sparse_dimensions):
+        init = [None] * sparse_dimensions.count(sparse_dim)
+        new_pos[sparse_dim] = init  # type: ignore[assignment] # looks like mypy is confused
```
egparedes, line 554: Maybe annotating `init` with `IncompleteSparsePositionEntry` would help mypy here.

src/functional/iterator/embedded.py
```diff
@@ -586,11 +606,28 @@ def shape(self):
         return self.array().shape
 
 
+def _is_tuple_axis(axis: Axis):
+    return axis is None
+
+
 def get_ordered_indices(
-    axes: Iterable[Axis], pos: Mapping[str, FieldIndex]
+    axes: Iterable[Axis], pos: Mapping[str, FieldIndex | SparsePositionEntry]
```
egparedes, line 614: Just a question: what does `pos: Mapping[str, FieldIndex | SparsePositionEntry]` represent here?

src/functional/iterator/embedded.py
```diff
@@ -586,11 +606,28 @@ def shape(self):
...
 ) -> tuple[FieldIndex, ...]:
-    assert all(axis.value in [*pos.keys()] for axis in axes)
-    return tuple(pos[axis.value] for axis in axes)
+    res = list[FieldIndex]()
```
egparedes, line 616:
```suggestion
    res: list[FieldIndex] = []
```

src/functional/iterator/embedded.py
```diff
@@ -586,11 +606,28 @@ def shape(self):
...
+    pos = deepcopy(pos)
+    for axis in axes:
+        if _is_tuple_axis(axis):
+            res.append(slice(None))
+        else:
+            assert axis is not None
```
egparedes, line 622: redundant? `_is_tuple_axis` is checking exactly the same

src/functional/iterator/embedded.py
```diff
@@ -586,11 +606,28 @@ def shape(self):
...
+            assert axis.value in pos
+            elem = pos[axis.value]
+            if isinstance(elem, list):
+                res.append(elem.pop(0))
```
egparedes, line 626: Do you mean ```res.append(elem[0])```?

tests/functional_tests/iterator_tests/test_toy_connectivity.py
egparedes: It looks correct to me but the link seems broken and it just points to this PR

egparedes (commented):
It looks good. Some questions and suggestions.

src/functional/iterator/embedded.py
```diff
@@ -33,6 +34,10 @@
 EMBEDDED = "embedded"
 
 
+class SparseOffset(str):
```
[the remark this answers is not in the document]
havogt: makes sense

src/functional/iterator/embedded.py
```diff
@@ -66,8 +71,9 @@ def __init__(
 OffsetProvider: TypeAlias = dict[Tag, OffsetProviderElem]
```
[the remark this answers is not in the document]
havogt: I think you proposed it earlier, but I ignored, because in iterator IR language int and tag are both offsets, not part of an offset, but I agree that for new people it's easier to thank in the way you propose. Will change.

src/functional/iterator/embedded.py
havogt: no

src/functional/iterator/embedded.py
havogt: With `if _is_tuple_axis(axis):` i wanted to highlight the tuple axis check (instead of using `if isinstance(axis, None)`), however now in the else mypy doesn't understand that I excluded None already.

src/functional/iterator/embedded.py
havogt: No, it's correct, but super ugly. What I do by popping is consuming the next element in the list of sparse dimensions (therefore the deepcopy in the beginning).

tests/functional_tests/iterator_tests/test_toy_connectivity.py
havogt: Search for the change `tag_stack.pop(0)` to `tag_stack.pop()`

src/functional/iterator/embedded.py
havogt: took it with smaller changes, but couldn't make mypy happy...

src/functional/iterator/embedded.py
havogt: changed to `Mapping[Tag, FieldIndex | SparsePositionEntry]`. `Tag` here is dimension tag.

src/functional/iterator/embedded.py
egparedes: I think the `assert` will finally become unnecessary after PR #841 gets merged, since the `_is_tuple_axis` was converted to a TypeGuard, so I recommend to merge that one first and then update this PR.

tests/functional_tests/iterator_tests/test_toy_connectivity.py
egparedes: Yep, still looks correct to me.

egparedes (commented):
Almost ready to go. Just a couple of minor things.

src/functional/iterator/embedded.py
```diff
@@ -600,11 +618,28 @@ def shape(self):
...
+        else:
+            assert axis is not None
+            assert axis.value in pos
+            elem = pos[axis.value]
+            if isinstance(elem, list):
+                res.append(elem.pop(0))
+            else:
+                assert isinstance(elem, (int, slice))
+                res.append(elem)
+    return tuple(res)
```
egparedes, line 642: Why is it needed to return a tuple here? It looks like it would be ok returning the list and annotating the return type as `Sequence` since `Sequence` is immutable and thus the static type should guarantee that the returned list is never modified.

src/functional/iterator/embedded.py
egparedes: Maybe adding this explanation as a comment would help.

src/functional/iterator/embedded.py
havogt: This is used to in `[]` of ndarray, which should be a tuple as far as I know.

src/functional/iterator/embedded.py
havogt: added a comment

src/functional/iterator/embedded.py
egparedes: Ok, it makes sense.

PR #841, opened 2022-06-27
Functional: Tuple field improvements

havogt:
Supports:
- nested tuples of fields
- extra dimensions marked by `None` as tuple dimensions

### DIFF
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -438,10 +438,7 @@ def _make_tuple(field_or_tuple: Union[LocatedField, tuple], indices) -> tuple:
 def _is_position_fully_defined(
     pos: Position,
 ) -> TypeGuard[ConcretePosition]:
-    return all(
-        isinstance(v, int) or (isinstance(v, list) and all(isinstance(e, int) for e in v))
-        for v in pos.values()
-    )
+    return all(isinstance(v, int) for v in pos.values())
 
 
 class MDIterator:
@@ -587,11 +584,15 @@ class LocatedFieldImpl:
         return self.array().shape
 
 
+def _is_tuple_axis(axis: Axis):
+    return axis is None
+
+
 def get_ordered_indices(
     axes: Iterable[Axis], pos: Mapping[str, FieldIndex]
 ) -> tuple[FieldIndex, ...]:
-    assert all(axis.value in [*pos.keys()] for axis in axes)
-    return tuple(pos[axis.value] for axis in axes)
+    assert all(axis.value in [*pos.keys()] for axis in axes if axis is not None)
+    return tuple(pos[axis.value] if not _is_tuple_axis(axis) else slice(None) for axis in axes)
 
 
 def _tupsum(a, b):
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -435,9 +435,7 @@ def _make_tuple(field_or_tuple: Union[LocatedField, tuple], indices) -> tuple:
         return field_or_tuple[indices]
 
 
-def _is_position_fully_defined(
-    pos: Position,
-) -> TypeGuard[ConcretePosition]:
+def _is_position_fully_defined(pos: Position) -> TypeGuard[ConcretePosition]:
     return all(isinstance(v, int) for v in pos.values())
 
 
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -7,12 +7,12 @@ from typing import (
     Any,
     Callable,
     Iterable,
+    Mapping,
     Optional,
     Protocol,
     Sequence,
     TypeAlias,
     TypeGuard,
-    TypeVar,
     Union,
     runtime_checkable,
 )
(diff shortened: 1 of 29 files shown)


src/functional/iterator/embedded.py
```diff
@@ -586,27 +594,31 @@ def shape(self):
         return self.array().shape
 
 
+def _is_tuple_axis(axis: Axis):
```
egparedes, line 597:
```suggestion
def _is_tuple_axis(axis: Axis) -> bool:
```
Additionally, this typing doesn't make sense. Either `Axis` is an `Union` including `None` or the type of the axis parameter is different here.  In case `Axis` should be an `Union`, this function could be implemented differently as a TypeGuard:

```
FieldAxis: TypeAlias = Dimension
Axis: TypeAlias = FieldAxis: | None

...

def _is_field_axis(obj) -> TypeGuard[FieldAxis]:
    return isinstance(obj, Axis)
```

src/functional/iterator/embedded.py
```diff
@@ -586,27 +594,31 @@ def shape(self):
...
+    return axis is None
+
+
 def get_ordered_indices(
     axes: Iterable[Axis], pos: Mapping[str, FieldIndex]
 ) -> tuple[FieldIndex, ...]:
-    assert all(axis.value in [*pos.keys()] for axis in axes)
-    return tuple(pos[axis.value] for axis in axes)
+    assert all(axis.value in [*pos.keys()] for axis in axes if axis is not None)
```
egparedes, line 604: Just an alternative: `Mapping`s' iterators are on keys by default:
```suggestion
    assert all(axis.value in [*pos] for axis in axes if axis is not None)
```

src/functional/iterator/embedded.py
```diff
@@ -586,27 +594,31 @@ def shape(self):
...
+    return tuple(pos[axis.value] if not _is_tuple_axis(axis) else slice(None) for axis in axes)
```
egparedes, line 605:
```suggestion
    return slice(None) if _is_tuple_axis(axis) else tuple(pos[axis.value] for axis in axes)
```

tests/functional_tests/iterator_tests/test_tuple.py
```diff
@@ -243,6 +271,34 @@ def test_field_of_tuple_input(fencil_processor):
...
+    rng = np.random.default_rng()
+
+    inp1 = rng.normal(size=(shape[0], shape[1], shape[2]))
+    inp2 = rng.normal(size=(shape[0], shape[1], shape[2]))
+    inp = np.zeros(shape + [2])
+    for i in range(shape[0]):
+        for j in range(shape[1]):
+            for k in range(shape[2]):
+                inp[i, j, k, 0] = inp1[i, j, k]
+                inp[i, j, k, 1] = inp2[i, j, k]
```
egparedes, line 287:
```suggestion
    inp = np.stack(inp1, inp2, axis=-1)
```

egparedes (commented):
It looks fine, just some questions and suggestions.

src/functional/iterator/embedded.py
```diff
@@ -440,6 +440,19 @@ def _setup_math_operations(cls):
 _UNDEFINED = Undefined()
 
 
+def _get_axes(field_or_tuple: Union[LocatedField, tuple]) -> list[Dimension]:
```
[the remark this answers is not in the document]
havogt: - It's arbitrary nesting of tuples...
- removed the `list()` and made return type `Sequence[Dimension]`

src/functional/iterator/embedded.py
havogt: I implemented your suggestion

src/functional/iterator/embedded.py
havogt: Makes sense, for my taste Python has the wrong default here, that's why I unconsciously use the explicit form.

src/functional/iterator/embedded.py
havogt: went with your other proposal for `_is_field_axis`

tests/functional_tests/iterator_tests/test_tuple.py
havogt: thanks, I was to lazy to try to figure out to do that, but I assumed it exists. 🙃

src/functional/iterator/embedded.py
```diff
@@ -586,27 +601,33 @@ def shape(self):
         return self.array().shape
 
 
+def _is_field_axis(axis: Axis) -> TypeGuard[FieldAxis]:
+    return isinstance(
+        axis, Dimension
+    )  # TODO should be isinstance(axis, FieldAxis) but mypy doesn't like it
```
havogt, line 607: for `isinstance(axis, FieldAxis)` mypy complains with

> error: Cannot use isinstance() with Any type  [misc]

Any ideas?

src/functional/iterator/embedded.py
```diff
@@ -586,27 +603,31 @@ def shape(self):
         return self.array().shape
 
 
+def _is_field_axis(axis: Axis) -> TypeGuard[FieldAxis]:
+    return isinstance(axis, (Dimension, Offset))  # TODO should be `isinstance(axis, FieldAxis)`
```
havogt, line 607: What's the correct pattern here, I want to use the alias FieldAxis in isinstance check, but mypy complains.

havogt: https://github.com/python/mypy/issues/11673

egparedes (approved):
It looks good.

PR #842, opened 2022-06-27
Exception -> warning for unsupported backend opt

### LINKED ISSUE #834: Verbose option not supported by numpy backend
For teaching purposes, we often use the verbose=True option when compiling a stencil. This option is not supported by the numpy backend. It is very annoying when switching back and forth between different backends in exemplary gt4py code. I would suggest to implement a verbose option into the numpy backend, even if it does absolutely nothing.

havogt:
Fixes #834

### DIFF
--- a/tests/test_integration/test_stencil_object.py
+++ b/tests/test_integration/test_stencil_object.py
@@ -68,9 +68,10 @@ def test_stencil_object_cache(backend: str):
 
 
 @pytest.mark.parametrize("backend", ALL_BACKENDS)
-@pytest.mark.parametrize("backend_opts", [{"verbose": True}])
-def test_common_stencil_options(backend, backend_opts):
-    @gtscript.stencil(backend=backend, **backend_opts)
-    def foo(f: Field[float]):
-        with computation(PARALLEL), interval(...):  # type: ignore
-            f = 42.0  # noqa F841
+def test_warning_for_unsupported_backend_option(backend):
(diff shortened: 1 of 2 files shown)


havogt:
bors try

havogt:
bors try

tests/test_integration/test_stencil_object.py
```diff
@@ -64,3 +65,12 @@ def runit(*args, **kwargs) -> float:
...
     assert cleaned_cache_time > fast_time
+
+
+@pytest.mark.parametrize("backend", ALL_BACKENDS)
+@pytest.mark.parametrize("backend_opts", [{"verbose": True}])
+def test_common_stencil_options(backend, backend_opts):
+    @gtscript.stencil(backend=backend, **backend_opts)
+    def foo(f: Field[float]):
+        with computation(PARALLEL), interval(...):  # type: ignore
+            f = 42.0  # noqa F841
```
jdahm, line 76: Is this testing that the `verbose` option is allowed in all backends, even if it is not defined as an option in some? Can you capture the warning and assert that it happens?

havogt: I changed the test to actually test to use a random option that none of the backends supports and check for the warning. I think it documents better the purpose of this change.

havogt:
bors try

jdahm (approved):
Looks great!

havogt:
Thanks for reviewing.

PR #843, opened 2022-06-27
Loop over node.params instead of api_signature for internal bindings

jdahm:
## Description

Params was not updated after pruning arguments, which left fields that were only referenced in eliminated horizontal regions still in the signature in the stencil module, but not in the pybind11 generated run method and caused a mismatch. This adds a test and fixes the issue.

It turns out that the easy fix is to not prune parameters at all, but to loop over the node.params instead of the api_signature when creating the internal bindings.


jdahm:
bors try

jdahm:
Ah yes, the CUDA backend is still failing, since I didn't touch it.

jdahm:
bors try

gronerl (approved):
Thanks!

gronerl:
bors try

PR #845, opened 2022-06-28
Unroll reduce with shift on any arg

### LINKED ISSUE #844: set ordering for parameters in reduction operation
UnrollReduce always assumes that the first argument to the lifted reduce is a shifted call, e.g.

```rot_vec = neighbor_sum(vec_e(V2E) * geofac_rot, axis=V2EDim)```works but

```rot_vec = neighbor_sum(geofac_rot * vec_e(V2E) , axis=V2EDim)``` doesn't

Hence ```assert isinstance(node, ir.FunCall)```in the function does not pass

### PR DESCRIPTION
**Pull request by havogt**
Fixes #844

### CODE REVIEW

File: src/functional/iterator/transforms/unroll_reduce.py [hunk 1]
```diff
@@ -1,14 +1,35 @@
+from itertools import groupby
+from typing import Iterable
```
fthaler, line 2: This is deprecated. Use `collections.abc.Iterable` instead.

src/functional/iterator/transforms/unroll_reduce.py
```diff
@@ -1,14 +1,35 @@
...
+
 from eve import NodeTranslator
 from eve.utils import UIDs
 from functional.iterator import ir
 
 
+def all_equal(it: Iterable):
+    g = groupby(it)
+    return next(g, True) and not next(g, False)
```
fthaler, line 11: I would comment that this code comes from a trustworthy source (itertools recipe) as it is not really well readable on its own.

src/functional/iterator/transforms/unroll_reduce.py
```diff
@@ -1,14 +1,35 @@
...
+                and isinstance(arg.fun, ir.FunCall)
+                and arg.fun.fun == ir.SymRef(id="shift")
+            ):
+                assert isinstance(arg.fun.args[0], ir.OffsetLiteral)
+                connectivities.append(offset_provider[arg.fun.args[0].value])
+
+        if not connectivities:
+            raise RuntimeError("Couldn't detect partial shift in any arguments of reduce.")
+
+        if not all_equal((c.max_neighbors, c.has_skip_values) for c in connectivities):
```
fthaler, line 30: …or maybe even just use the following which probably has the same readability for people used to Python:
```suggestion
        if len({(c.max_neighbors, c.has_skip_values) for c in connectivities}) != 1:
```

src/functional/iterator/transforms/unroll_reduce.py
```diff
@@ -1,14 +1,35 @@
+from itertools import groupby
+from typing import Iterable
+
 from eve import NodeTranslator
 from eve.utils import UIDs
 from functional.iterator import ir
 
 
+def all_equal(it: Iterable):
```
fthaler, line 9:
```suggestion
def _all_equal(it: Iterable):
```
and probably put it as staticmethod into `UnrollReduce` as done with all other helper functions in this module.

src/functional/iterator/transforms/unroll_reduce.py
```diff
@@ -1,14 +1,35 @@
...
-        return node.fun.args[0]
+    def _find_connectivity(reduce_args: Iterable[ir.Expr], offset_provider):
+        connectivities = []
+        for arg in reduce_args:
+            if (
+                isinstance(arg, ir.FunCall)
+                and isinstance(arg.fun, ir.FunCall)
+                and arg.fun.fun == ir.SymRef(id="shift")
+            ):
+                assert isinstance(arg.fun.args[0], ir.OffsetLiteral)
```
fthaler, line 24: Is this good enough? I guess there might be cases where we have `shift(dpart)(shift(dfull, 0)(x))` (maybe after some inlining) which after normalization becomes `shift(dfull, 0, dpart)`. So we might actually have to check if the last element is a non-integer offset literal and choose that one?

fthaler (changes requested):

fthaler (approved):

PR #846, opened 2022-06-29
Force use of default CUDA stream in dace orchestration tests

gronerl:
## Description
Try to fix current GPU CI failures.

### DIFF
--- a/tests/test_integration/test_dace_parsing.py
+++ b/tests/test_integration/test_dace_parsing.py
@@ -41,6 +41,7 @@ def dace_env():
     with dace.config.temporary_config():
         # Setting max_concurrent_streams to -1 configures dace to only use the default stream.
         dace.config.Config.set("compiler", "cuda", "max_concurrent_streams", value=-1)
+        dace.config.Config.set("compiler", "cuda", "args", value="--default-stream legacy")
         dace.config.Config.set("compiler", "cpu", "openmp_sections", value=False)
         dace.config.Config.set("compiler", "cpu", "args", value="")
         dace.config.Config.set("compiler", "allow_view_arguments", value=True)
--- a/tests/test_integration/test_dace_parsing.py
+++ b/tests/test_integration/test_dace_parsing.py
@@ -41,7 +41,9 @@ def dace_env():
     with dace.config.temporary_config():
         # Setting max_concurrent_streams to -1 configures dace to only use the default stream.
         dace.config.Config.set("compiler", "cuda", "max_concurrent_streams", value=-1)
-        dace.config.Config.set("compiler", "cuda", "args", value="-Xcompiler -fPIC --default-stream legacy")
+        dace.config.Config.set(
+            "compiler", "cuda", "args", value="-Xcompiler -fPIC --default-stream legacy"
+        )
         dace.config.Config.set("compiler", "cpu", "openmp_sections", value=False)
         dace.config.Config.set("compiler", "cpu", "args", value="")
         dace.config.Config.set("compiler", "allow_view_arguments", value=True)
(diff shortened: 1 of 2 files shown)


gronerl:
bors try

gronerl:
bors try-
bors try

gronerl:
bors try-
bors try

tests/test_integration/test_dace_parsing.py
```diff
@@ -41,6 +41,10 @@ def dace_env():
     with dace.config.temporary_config():
         # Setting max_concurrent_streams to -1 configures dace to only use the default stream.
         dace.config.Config.set("compiler", "cuda", "max_concurrent_streams", value=-1)
+        dace.config.Config.set(
+            "compiler", "cuda", "args", value="-Xcompiler -fPIC --default-stream legacy"
+        )
```
jdahm, line 46: Not sure how much this matters, but `-fPIC` is not the flag name on all host compilers (e.g. Microsoft and IBM compilers use different names). Is that needed, or only the `--default-stream` option?

gronerl:
I put back the -fPIC since the error message before asked for it :sweat_smile:

gronerl:
All those attempts were shots in the dark btw, so if you have any idea what could cause the issue, please let me know :worried:

gronerl:
bors try

tests/test_integration/test_dace_parsing.py
gronerl: I put back the -fPIC since the error message before asked for it :sweat_smile: 
But the original default flags also had the -fPIC so this should be harmless.

gronerl:
bors try

tests/test_integration/test_dace_parsing.py
jdahm: Ah, good to know!

gronerl:
I can't say I know why the copy didn't trigger the synchronization (can @fthaler enlighten me?), but this works, even though it's using the crow bar.

tests/test_integration/test_dace_parsing.py
```diff
@@ -94,6 +98,10 @@ def call_stencil_object(locoutp, locinp):
         defn(locoutp, par=locinp)
 
     call_stencil_object(locoutp=outp, locinp=inp)
+
+    if cp is not None:
+        cp.cuda.Device(0).synchronize()
+
     outp.device_to_host(force=True)
```
jdahm, line 105: Is this device synchronization required for other backends as well, or is it because `dace.program` doesn't synchronize as bare stencils do?

jdahm (approved):
Great work in tracking this down! I really appreciate the quick fixes to these CI issues involving DaCe.

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try

tests/test_integration/test_dace_parsing.py
gronerl: In general, we rely on the device_to_host copy to be synchronous with previous CUDA calls. This is the case if everything is running on the same CUDA stream. Since that seemed to not work, this was an attempt to solve the issue by explicitly putting a device-wide barrier. before copying to host memory.

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try-
bors try

gronerl:
bors try

gronerl:
bors try

PR #851, opened 2022-07-04
Small fix to eve.infer_type

tehrengruber:
Change `eve.infer_type((3, 4))` to return `tuple[int, int]` instead of `tuple[int, ...]`


havogt (approved):
Looks good to me.

PR #853, opened 2022-07-05
Improve UnpackedAssignPass pass to avoid expression duplication

tehrengruber:


src/functional/ffront/ast_passes/simple_assign.py
```diff
@@ -127,12 +130,19 @@ def visit_Assign(self, node: ast.Assign) -> Iterator[ast.Assign]:
     def _unpack_assignment(
         self, node: ast.Assign, *, targets: list[ast.expr]  # targets passed here for typing
     ) -> Iterator[ast.Assign]:
+        tuple_name = ast.Name(id=f"__tuple_tmp_{self.counter}", ctx=ast.Store())
```
petiaccja, line 133: Do you think we could extract unique temporary name generation to a utility function outside this class? I guess it may be useful at multiple places, and it would avoid having state inside all the node transformers. `counter` is also not very meaningful when reading the code.

tehrengruber: Indeed there are some places were temporary name generation is used and it would be great to have the same format for all of them. This is however outside of the scope of this PR. Having state inside the node transformers contrary to a global counter is however the right approach I would say. The counter only needs and consequently should be unique per function / scope only. Having a global counter makes testing unnecessarily harder, the foast represention of a field operator would depend on previous operators and generated could would be even harder to read.

PR #854, opened 2022-07-05
Functional: Allow returning multiple fields from a field operator

tehrengruber:
This PR fixes some problems and improves support for field operators returning tuples of fields.

```python
@field_operator
def foo(
    a: Field[[IDim], float64], b: Field[[IDim], float64]
) -> tuple[Field[[IDim], float64], Field[[IDim], float64]]:
    return a, b

@field_operator
def combine(a: Field[[IDim], float64], b: Field[[IDim], float64]) -> Field[[IDim], float64]:
    something = foo(a, b)
    return something[0] + something[1]

combine(a, b, out=out, offset_provider={})
```

Note that this does not include programs returning multiple fields.

### DIFF
--- a/src/functional/ffront/foast_to_itir.py
+++ b/src/functional/ffront/foast_to_itir.py
@@ -227,8 +227,8 @@ class FieldOperatorLowering(NodeTranslator):
         if iterator_type_kind(node.value.type) is ITIRTypeKind.ITERATOR:
             return self._lift_lambda(node)(im.tuple_get_(node.index, im.deref_(value)))
         elif iterator_type_kind(node.value.type) in (
-                ITIRTypeKind.VALUE,
-                ITIRTypeKind.ENCAPSULATED_ITERATOR,
+            ITIRTypeKind.VALUE,
+            ITIRTypeKind.ENCAPSULATED_ITERATOR,
         ):
             return im.tuple_get_(node.index, value)
         raise AssertionError("Unexpected `IteratorTypeKind`.")
@@ -242,8 +242,8 @@ class FieldOperatorLowering(NodeTranslator):
             elts = tuple(to_value(el)(self.visit(el, **kwargs)) for el in node.elts)
             return self._lift_lambda(node)(im.make_tuple_(*elts))
         elif iterator_type_kind(node.type) in (
-                ITIRTypeKind.VALUE,
-                ITIRTypeKind.ENCAPSULATED_ITERATOR,
+            ITIRTypeKind.VALUE,
+            ITIRTypeKind.ENCAPSULATED_ITERATOR,
         ):
             elts = tuple(self.visit(el, **kwargs) for el in node.elts)
             return im.make_tuple_(*elts)
--- a/src/functional/ffront/foast_to_itir.py
+++ b/src/functional/ffront/foast_to_itir.py
@@ -210,7 +210,7 @@ class FieldOperatorLowering(NodeTranslator):
             .filter(is_expr_with_iterator_type_kind(IteratorTypeKind.ENCAPSULATED_ITERATOR))
         ):
             raise NotImplementedError(
-                "Using a composite type containing a local field is not supported."
+                "Using composite types (e.g. tuples) containing local fields not supported."
             )
         param_names = (
             node.pre_walk_values()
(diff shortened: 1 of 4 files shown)


samkellerhals:
In your example then, is `out` a tuple of fields?

nfarabullini:
I have added an additional test, which fails because one of the inputs is a float

tehrengruber:
@samkellerhals 

> Can out also be a tuple of fields? The most common pattern I've seen in ICON is needing to return more than one field from a field_operator/program.

No that is what I mean with "Note that this does not include programs returning multiple fields.". This PR is a prerequisite for that though.

tehrengruber:
Reconsidering the approach here. Back to draft state for now.

src/functional/ffront/foast_to_itir.py
```diff
@@ -31,61 +32,80 @@
...
-    return TypeKind.UNKNOWN
-
-
-def can_be_value_or_iterator(symbol_type: ct.SymbolType):
-    return resulting_type_kind(symbol_type) is not TypeKind.UNKNOWN
-
-
-def is_field(expr: foast.Expr) -> bool:
-    return type_info.type_class(expr.type) is ct.FieldType
+    Return the kind of type on iterator level an foast expr of given type corresponds to.
```
DropD, line 52: I stumbled over this sentence and had to read it twice, so I tried to rephrase it:
```suggestion
    Return the corresponding type kind (on iterator level) to a FOAST expression of the given SymbolType.
```

src/functional/ffront/foast_to_itir.py
```diff
@@ -31,61 +32,80 @@
...
+        that contains at least one iterator.
+
+    +------------------------------------+------------------------+
+    | Expr                               | Iterator Type Kind     |
+    +====================================+========================+
+    | 1                                  | VALUE                  |
+    | regular_field                      | ITERATOR               |
+    | local_field                        | ITERATOR               |
+    | (1, 1)                             | VALUE                  |
+    | (1, regular_field)                 | ITERATOR               |
```
DropD, line 70: Looking at the definition of `ENCAPSULATED_ITERATOR` above, this seems to meet it. A tuple that contains at least one iterator. Am I reading the definition wrong? Or does it also meet the definition of `ITERATOR` (since it is treated as an iterator of tuples),  and that takes precedence?

src/functional/ffront/foast_to_itir.py
```diff
@@ -31,61 +32,80 @@
...
-    value or iterator expression during lowering.
-
-    Examples:
-    ---------
-    >>> resulting_type_kind(ct.DeferredSymbolType(constraint=None)).name
-    'UNKNOWN'
+class IteratorTypeKind(enum.Enum):
+    VALUE = 0
+    ITERATOR = 1
+    ENCAPSULATED_ITERATOR = 2
```
DropD, line 45: If you look at it through the eyes of someone unfamiliar with the codebase and the concepts behind it, do you see the double meaning of `iterator` here as problematic?

In `IteratorTypeKind` it means the type kind seen from the iterator level context, in `.ITERATOR` it means the type kind named iterator. That is, if I have understood it correctly.

DropD (changes requested):
Looks good, there are only some minor things which should be easy to resolve

src/functional/ffront/foast_to_itir.py
tehrengruber: The description of `VALUE, ITERATOR, ENCAPSULATED_ITERATOR` is on iterator level, which was misleading as it used _expression_ which was then also used as the column heading for FOAST expressions in the table. I changed the descriptions to _lowered expression_ and the heading to _FOAST Expr_. With respect to the example: `(1, regular_field)` will be lowered to an iterator of tuples so it's type kind is `ITERATOR`.

src/functional/ffront/foast_to_itir.py
tehrengruber: Your understanding is correct. Not sure it is problematic, but it is certainly not optimal so I am open for alternative names.

DropD: What do you think of `IterViewTypeKind` or something that references "IteratorView" or "ITIR" or "IteratorLevel" as opposed to "Iterator"?

src/functional/ffront/foast_to_itir.py
DropD: Much clearer now, thanks

src/functional/ffront/foast_to_itir.py
tehrengruber: I like ITIRTypeKind. Renamed it.

DropD (approved):

PR #860, opened 2022-07-13
Prototype for unstructured in compiled backends

havogt:
This is just a hacky demonstration. I open against `compiled-backend-integration` but should not be merged there, but come as a separate PR to `functional`.

### DIFF
--- a/src/functional/fencil_processors/callables/cpp/build.py
+++ b/src/functional/fencil_processors/callables/cpp/build.py
@@ -93,8 +93,8 @@ class CMakeListsGenerator(TemplatedGenerator):
                 return textwrap.dedent(
                     """\
                     FetchContent_Declare(GridTools
-                        GIT_REPOSITORY https://github.com/GridTools/gridtools.git
-                        GIT_TAG        master
+                        GIT_REPOSITORY https://github.com/havogt/gridtools.git
+                        GIT_TAG        python_neighbor_table
                     )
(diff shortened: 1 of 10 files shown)


src/functional/fencil_processors/runners/gtfn_cpu.py
```diff
@@ -46,11 +61,21 @@ def convert_arg(arg):
     return [convert_arg(arg) for arg in args]
 
 
+def neighbortable_args(offset_provider):
+    return [c.tbl for c in offset_provider.values() if isinstance(c, Connectivity)]
```
havogt, line 65: We need to define the Connectivity Protocol properly. `tbl` shouldn't be the API.

src/functional/fencil_processors/callables/cpp/build.py
```diff
@@ -93,8 +93,8 @@ def visit_FindDependency(self, dep: FindDependency):
                 return textwrap.dedent(
                     """\
                     FetchContent_Declare(GridTools
-                        GIT_REPOSITORY https://github.com/GridTools/gridtools.git
-                        GIT_TAG        master
+                        GIT_REPOSITORY https://github.com/havogt/gridtools.git
```
havogt, line 96: I added support for pybuffer as neighbortable in GridTools/gridtools, but not sure if we should keep it there or add it to GT4Py.

PR #868, opened 2022-07-25
Cartesian: "No Storage" Implementation

gronerl:
This PR removes the Storage classes and implements the new "No Storage" concept.

It is currently fully in the cartesian branch but the same shall be applied to `functional`.


gronerl:
bors try

gronerl:
bors try-

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try-
bors try
😐

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try-
bors try

docs/gt4py/quickstart.rst
```diff
@@ -183,29 +185,43 @@ set of routines used in `NumPy` for allocation: ``from_array``, ``ones``, ``zero
...
+        backend=backend, shape=(10, 10, 10), dtype=np.float64, aligned_index=(0, 0, 0)
     )
 
     stencil_example(field_a, field_b, field_c, result, alpha=0.5)
 
 
-The ``default_origin`` parameter plays two roles:
+The ``aligned_index`` specifies that the array is to be allocated such that memory address of the point specified in
+``aligned_index`` is `aligned` to a backend-dependent value. For optimal performance, you set the ``algined_index`` to
+a point which is the lower-left corner of the iteration domain most frequently used for this field.
```
jdahm, line 205: Still diving in here so apologies if this is clear later: does specifying `aligned_index` instead of `default_origin` mean that we always need to specify non-zero origins when calling stencils, or is this used as the default origin too?

gronerl: Kind of. If you just rely on the allocation routines, you get an unmodified NumPy ndarray. That is, there is no `default_origin` attribute like there used to be on the storages. However, if a user passes a buffer object that (This could e.g. be a `Quantity` in pace) They can opt to implenet an dunder attribute `__gt_dims__` that has then that functionality.

src/gt4py/storage/storage.py
```diff
@@ -39,659 +36,105 @@ def _error_on_invalid_backend(backend):
         raise RuntimeError(f"Backend '{backend}' is not registered.")
 
 
-def empty(backend, default_origin, shape, dtype, mask=None, *, managed_memory=False):
+def empty(backend, aligned_index, shape, dtype, dimensions=None):
```
jdahm, line 39: Can `aligned_index` be made optional and default to `(0, 0, 0)`?

jdahm:
It seems somehow inconsistent to allow (and currently require) aligning memory on an index, but then not defaulting to that as the origin. I understand the reason behind it is that we want "no storages", so we can't rely on attributes being present. In order to have maximum performance then the user has to specify an `aligned_index` in one part of the code, and then use _exactly_ that value everywhere else it's called. Seems like duplication, but right now I don't see a great way around it.

docs/gt4py/quickstart.rst
```diff
@@ -183,29 +185,43 @@ set of routines used in `NumPy` for allocation: ``from_array``, ``ones``, ``zero
...
-#. The data is allocated such that memory address of the point specified in ``default_origin`` is `aligned` to a
-   backend-dependent value. For optimal performance, you set the ``default_origin`` to a point which is the
-   lower-left corner of the iteration domain most frequently used for this storage.
+----------------
+Array Interfaces
+----------------
 
-#. If when calling the stencil, no other `origin` is specified, this value is where the `iteration domain` begins, i.e.
-   the grid point with the lowest index where a value is written.
+When passing buffers to stencils, they can be in any form that is compatible with ``np.asarray`` or ``cp.asarray``,
```
jdahm, line 211: Is the idea that it's really anything that implements the array interface protocol (buffer protocol) https://numpy.org/doc/stable/reference/arrays.interface.html?

src/gt4py/backend/dace_backend.py
```diff
@@ -366,13 +365,17 @@ def unique_index(self) -> int:
...
             if name in sdfg.arrays:
                 data = sdfg.arrays[name]
                 assert isinstance(data, dace.data.Array)
-                res[name] = "py::buffer {name}, std::array<gt::int_t,{ndim}> {name}_origin".format(
+                res[
+                    name
+                ] = "py::{pybind_type} {name}, std::array<gt::int_t,{ndim}> {name}_origin".format(
+                    pybind_type="object"
+                    if self.backend.storage_info["device"] == "gpu"
+                    else "buffer",
```
jdahm, line 378: Just curious: why do these need to be different? I'm sure there's a good reason. I'm not too familiar with the details of pybind11.

jdahm (commented):
Looks good. I wrote down a few thoughts to consider.

gronerl:
The failing tests should pass after merging/squashing/rebasing master. As overriding gitlab is currently disable, can @jdahm  please do so and then `bors try` again?

docs/gt4py/arrays.rst
```diff
@@ -0,0 +1,175 @@
...
+
+
+
+Allocation
+----------
+
+For the performance-optimal allocation and initialization of arrays to be used in GT4Py, we provide the following set of
+functions which closely resemble their NumPy counterparts (meaning of the common parameters is explained below).
+
+The return type is either a ``numpy.ndarray`` or a `cupy.ndarray`, for CPU and GPU backends, respectively.
```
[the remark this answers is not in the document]
gronerl: No, I think this did the roundtrip to markdown and back once too many, I'll make a pass through it to add the `:code:` consistently.

docs/gt4py/quickstart.rst
gronerl: In principle, yes. e.g. `numpy` backends and `dace:*` backends should be fine with that. The option to reject a buffer based on wrong layout, (including no contiguous dimension) is still there!

src/gt4py/backend/dace_backend.py
gronerl: They should have not been `py::buffer` even now. It works now, since the GPU storages are both a `py::buffer` (due to the dual CPU buffer)  but also have the `__cuda_array_interface__` attribute that `as_cuda_sid` accesses. However, now we pass `CuPy` arrays from the python side, so the old interface breaks down.

src/gt4py/storage/utils.py
```diff
@@ -12,72 +12,84 @@
...
+        dimensions = (
+            list("IJK"[: len(shape)])
+            if len(shape) <= 3
+            else list("IJK") + [str(d) for d in range(len(shape) - 3)]
+        )
 
+    if not all(
+        isinstance(d, (str, Axis)) and (str(d).isdigit() or str(d) in "IJK") for d in dimensions
+    ):
+        raise ValueError(f"Invalid dimensions definition: '{dimensions}'")
```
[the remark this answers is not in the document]
gronerl: I think this should be a user-facing error message, not a debug-stile assertion / internal consistency assumption. Why do you think an assertion is more suitable?

tests/test_integration/test_stencil_object.py
```diff
@@ -34,11 +36,13 @@ def stencil(in_field: Field[float], out_field: Field[float], *, offset: float):
             )
 
     shape = (4, 4, 4)
-    in_storage = gt_storage.ones(
-        backend=backend, default_origin=(0, 0, 0), shape=shape, dtype=float
+    in_storage = OriginWrapper(
+        array=gt_storage.ones(backend=backend, aligned_index=(0, 0, 0), shape=shape, dtype=float),
+        origin=(0, 0, 0),
```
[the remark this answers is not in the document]
gronerl: To be clear, `OriginWrapper` etc are just utility wrappers for testing. There won't be anything user-faceing like that. I think it isn't worth to have additional API just to avoid that small repetition.

gronerl:
Switching this to draft while I'm working on addressing review comments.

src/gt4py/storage/utils.py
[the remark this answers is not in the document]
jdahm: I believe I was thinking about performance at some point, but this isn't called in the `StencilObject` after all, only in the allocation methods. In that case it's correct to have this be a `ValueError`.

gronerl:
bors try

gronerl:
bors try-
bors try

src/gt4py/stencil_object.py
```diff
@@ -17,34 +17,76 @@
...
+def _compute_cache_key(array_infos, parameter_args, domain, origin) -> int:
     field_data = tuple(
-        (name, field.shape, getattr(field, "default_origin", (0, 0, 0)))
-        for name, field in field_args.items()
-        if field is not None
+        (name, info.array.shape, info.origin or (0, 0, 0))
+        for name, info in array_infos.items()
+        if info is not None
     )
     return hash((field_data, *parameter_args.keys(), dumps(domain), dumps(origin)))
```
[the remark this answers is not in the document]
gronerl: I changed it so we need `asarray(...).shape` and `getattr(arg, "__gt_dims__", (0, 0, 0))` I.e. there is an additional asarray call compared to before. I would assume that's still efficient enough.

docs/gt4py/arrays.rst
```diff
@@ -0,0 +1,175 @@
...
+
+:code:`empty(shape: Sequence[int], dtype: dtype_like = np.float64, **kwargs) -> ndarray`
+    Allocate an array with uninitialized (undefined) values.
```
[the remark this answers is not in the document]
gronerl: Fixed.

tests/test_integration/test_dace_parsing.py
```diff
@@ -81,20 +83,25 @@ def defn(outp: gtscript.Field[np.float64], par: np.float64):
         with computation(PARALLEL), interval(...):
             outp = par  # noqa F841: local variable 'outp' is assigned to but never used
 
-    outp = gt_storage.zeros(
-        dtype=np.float64, shape=(10, 10, 10), default_origin=(0, 0, 0), backend=backend
+    # DaCe in the currently tagged version of this repo (~=0.13 doesn't support non-numpy arrays)
+    wrapper_type = OriginWrapper if "gpu" in backend else NdarraySubclassOriginWrapper
```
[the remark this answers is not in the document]
gronerl: dace released 0.14 by now where this is no longer needed. I updated the setup.cfg to use it.

tests/test_unittest/test_storage.py
```diff
@@ -445,619 +388,59 @@ def test_cpu_constructor(alloc_fun, backend):
...
 
 
 @pytest.mark.requires_gpu
-@hyp.given(param_dict=mask_strategy())
+@hyp.given(param_dict=dimensions_strategy())
 def test_masked_storage_gpu(param_dict):
-    mask = param_dict["mask"]
-    default_origin = param_dict["default_origin"]
+
+    import cupy as cp
```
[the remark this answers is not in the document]
gronerl: Probably a mistake.

src/gt4py/stencil_object.py
```diff
@@ -17,34 +17,76 @@
...
+@dataclass
+class _ArgsInfo:
+    device: str
+    array: FieldType
+    original_object: Any
+    origin: Optional[Tuple[int]] = None
+    dimensions: Optional[Tuple[str]] = None
+
+
+def _extract_array_infos(field_args, device) -> Dict[str, Optional[_ArgsInfo]]:
```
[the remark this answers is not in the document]
gronerl: Done.

src/gt4py/storage/storage.py
gronerl: Docstrings done.

tests/test_unittest/test_storage.py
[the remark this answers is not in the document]
gronerl: fixed.

gronerl:
I addressed all comments where I agree and marked as resolved where I deemed it appropriate. We need to discuss making `aligned_index` optional. Please mark other conversations as resolved if your issue is addressed. Or leave further comments.

gronerl:
bors try

src/gt4py/stencil_object.py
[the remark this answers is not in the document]
jdahm: Thanks. That's great!

src/gt4py/storage/storage.py
jdahm: Sorry to bother, but I still don't see docstrings on `empty`, `ones`, and `zeros` here. Can you add those before merging? Thanks!

docs/gt4py/quickstart.rst
jdahm: Great!

docs/gt4py/quickstart.rst
```diff
@@ -183,29 +185,43 @@ set of routines used in `NumPy` for allocation: ``from_array``, ``ones``, ``zero
...
+        backend=backend, shape=(10, 10, 10), dtype=np.float64, aligned_index=(0, 0, 0)
     )
     field_c = gt_storage.zeros(
-        backend=backend, shape=(10, 10, 10), dtype=np.float64, default_origin=(0, 0, 0)
+        backend=backend, shape=(10, 10, 10), dtype=np.float64, aligned_index=(0, 0, 0)
     )
     result = gt_storage.empty(
-        backend=backend, shape=(10, 10, 10), dtype=np.float64, default_origin=(0, 0, 0)
+        backend=backend, shape=(10, 10, 10), dtype=np.float64, aligned_index=(0, 0, 0)
     )
```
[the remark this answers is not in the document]
jdahm: Or perhaps reformat these to
```python
result = gt_storage.empty(
    backend, (0, 0, 0), shape=(10, 10, 10), dtype=np.float64
)
```

as the function signature suggests.

src/gt4py/storage/storage.py
gronerl: They're there! I think if you click the file name above it takes you to the old version.

tests/test_integration/test_code_generation.py
```diff
@@ -527,26 +544,23 @@ def k_to_ijk(outp: Field[np.float64], inp: Field[gtscript.K, np.float64]):
     data = np.arange(10, dtype=np.float64)
     inp = gt_storage.from_array(
         data=data,
-        shape=(10,),
-        default_origin=(0,),
+        aligned_index=(0,),
         dtype=np.float64,
-        mask=[False, False, True],
+        dimensions="K",
```
jdahm, line 549: Should this be `dimensions=["K"],`?

jdahm (approved):
Great work on this!

tests/test_integration/test_code_generation.py
gronerl: A string is also a sequence of strings, which means either is fine, (unless there are >10 data dimensions).

docs/gt4py/quickstart.rst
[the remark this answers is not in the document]
gronerl: I reformatted according to the signature.

docs/gt4py/arrays.rst
```diff
@@ -0,0 +1,175 @@
...
+
+GT4Py does not provide its own data container class, but supports established python standards for exposing
+N-dimensional buffers. There is a minimalistic interface allowing to specify the correspondence of buffer dimensions
+to the semantic dimensions assumed in the stencil. This correspondence does not necessarily need to be specified since
+the stencils specify a default ordering.
+
+GT4Py provides utilities to allocate buffers that have optimal layout and alignment for a given backend.
+
+In this document, we describe the interfaces for
+* supported buffer interfaces
```
egparedes, line 13:
```suggestion
In this document, we describe the interfaces for

* supported buffer interfaces
```

docs/gt4py/arrays.rst
```diff
@@ -0,0 +1,175 @@
...
+buffers. GT4Py developers are advised to always use those utilities as to guarantee support across gt4py as the
+supported interfaces are extended.
+
+Dimension Mapping
+^^^^^^^^^^^^^^^^^
+
+The user can optionally implement a :code:`__gt_dims__` attribute in the object implementing any of the supported buffer
+interfaces. As a fallback if neither is specified the dimensions given in the annotations (by means of
+:code:`gtscript.Field`) are assumed. The returned object should be a tuple of strings labeling the dimensions in index
+order.
```
egparedes, line 47:
```suggestion
The user can optionally implement a :code:`__gt_dims__` attribute in the object implementing any of the supported buffer
interfaces. The returned object should be a tuple of strings labeling the dimensions in index order.
As a fallback if the attribute is not implemented, it is assumed that the buffer contains the dimensions given in the annotations
(by means of :code:`gtscript.Field`) exactly in the same order.
```

docs/gt4py/arrays.rst
```diff
@@ -0,0 +1,175 @@
...
+        + :code:`fill_value: Number`. The number to which the storage is initialized.
+
+        + :code:`dtype: dtype_like`
+          The dtype of the storage (NumPy dtype or accepted by :code:`np.dtype()`). It defaults to
+          :code:`np.float64`.
+
+    For common keyword-only arguments, please see below.
+
+:code:`from_array(data: array_like, device_data: array_like = None, *, dtype: dtype_like = np.float64, **kwargs) -> ndarray`
+    Used to allocate an array with values initialized to those of a given array.
```
egparedes, line 134:
```suggestion
    Used to allocate an array with values initialized from the content of a given array.
```

docs/gt4py/arrays.rst
```diff
@@ -0,0 +1,175 @@
...
+
+    Parameters:
+        + :code:`data: array_like`. The original array from which the storage is initialized.
+
+        + :code:`device_data: array_like`. The original array in case copying to a gpu buffer is
+          desired. The same buffer could also be passed through :code:`data` in that case, however this
+          parameter is here to provide the same interface like the :code:`as_storage` function.
+
+        + :code:`sync_state: gt4py.storage.SyncState`. If `managed="gt4py"` indicates which of the
+          provided buffers, :code:`data` or :code:`device_data`, is up to date at the time of initialization.
```
egparedes, line 144:
```suggestion
```

docs/gt4py/arrays.rst
```diff
@@ -0,0 +1,175 @@
...
+
+:code:`from_array(data: array_like, device_data: array_like = None, *, dtype: dtype_like = np.float64, **kwargs) -> ndarray`
+    Used to allocate an array with values initialized to those of a given array.
+
+    Parameters:
+        + :code:`data: array_like`. The original array from which the storage is initialized.
+
+        + :code:`device_data: array_like`. The original array in case copying to a gpu buffer is
+          desired. The same buffer could also be passed through :code:`data` in that case, however this
+          parameter is here to provide the same interface like the :code:`as_storage` function.
```
egparedes, line 141: Is `device_data` still needed/relevant? Cannot be removed?

egparedes (commented):
Just a partial review for now (only the docs part).

src/gt4py/storage/storage.py
```diff
@@ -39,659 +44,372 @@ def _error_on_invalid_backend(backend):
...
+            If illegal or inconsistent arguments are specified.
+    """
+    storage = empty(
+        shape=shape,
+        dtype=dtype,
+        backend=backend,
+        aligned_index=aligned_index,
+        dimensions=dimensions,
+    )
+    storage[...] = fill_value
```
egparedes, line 214: Same suggestion as above about doing a explicit type casting of the fill value

src/gt4py/storage/storage.py
```diff
@@ -39,659 +44,372 @@ def _error_on_invalid_backend(backend):
...
         shape=shape,
         dtype=dtype,
         backend=backend,
-        default_origin=default_origin,
-        mask=mask,
-        managed_memory=managed_memory,
+        aligned_index=aligned_index,
+        dimensions=dimensions,
     )
     storage[...] = 0
```
egparedes, line 267: Same comment as as above

src/gt4py/storage/storage.py
```diff
@@ -39,659 +44,372 @@ def _error_on_invalid_backend(backend):
...
     is_cupy_array = cp is not None and isinstance(data, cp.ndarray)
-    xp = cp if is_cupy_array else np
-    if shape is None:
-        shape = xp.asarray(data).shape
+    asarray = gt4py.storage.utils.as_cupy if is_cupy_array else gt4py.storage.utils.as_numpy
+    shape = asarray(data).shape
     if dtype is None:
-        dtype = xp.asarray(data).dtype
+        dtype = asarray(data).dtype
+    dtype = np.dtype(dtype)
```
egparedes, line 321: If `cupy.ndarray.dtype` returns `np.dtype` instances, then I guess this is redundant... although most likely a bit safer
```suggestion
```

src/gt4py/storage/utils.py
```diff
@@ -12,72 +12,85 @@
...
-    pass
+    cp = None
+
+
+class ArrayInterfaceType(Protocol):
+    __array_interface__: Dict[str, Any]
+
+
+class CudaArrayInterfaceType(Protocol):
+    __array_interface__: Dict[str, Any]
```
egparedes, line 36:
```suggestion
    __cuda_array_interface__: Dict[str, Any]
```

src/gt4py/storage/utils.py
```diff
@@ -217,44 +215,70 @@ def allocate_gpu_unmanaged(default_origin, shape, layout_map, dtype, alignment_b
...
 
-def allocate_cpu(default_origin, shape, layout_map, dtype, alignment_bytes):
-    def allocate_f(size, dtype):
-        raw_buffer = np.empty(size, dtype)
-        return raw_buffer, raw_buffer
+def allocate_cpu(aligned_index, shape, layout_map, dtype, alignment_bytes):
+    dtype = np.dtype(dtype)
+    assert (
+        alignment_bytes % dtype.itemsize
+    ) == 0, "Alignment must be a multiple of byte-width of dtype."
```
egparedes, line 225: Unless this check uses `assert` on purpose for performance reasons, I think it'd be better to use `if` and raise an exception here to avoid it can be disabled in performance mode.

egparedes (commented):
Second part of my review.

docs/gt4py/quickstart.rst
```diff
@@ -180,32 +182,46 @@ set of routines used in `NumPy` for allocation: ``from_array``, ``ones``, ``zero
...
     field_a = gt_storage.from_array(
-        data=np.random.randn(10, 10, 10),
+        np.random.randn(10, 10, 10),
+        np.float64,
         backend=backend,
-        dtype=np.float64,
-        default_origin=(0, 0, 0),
+        aligned_index=(0, 0, 0),
     )
     field_b = gt_storage.ones(
```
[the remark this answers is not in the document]
gronerl: Do we need one? It's a quick start, not a comprehensive documentation. Should be straight-forward.

src/gt4py/storage/storage.py
```diff
@@ -28,8 +30,11 @@
 except ImportError:
     dace = None
 
-from gt4py import backend as gt_backend
-from gtc import utils as gtc_utils
+if np.lib.NumpyVersion(np.__version__) >= "1.20.0":
+    from numpy.typing import ArrayLike, DTypeLike
+else:
+    ArrayLike = Any
+    DTypeLike = Any
```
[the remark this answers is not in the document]
gronerl: If we're going to approximate, I'd rather approximate "from above" i.e. include more in the type then is allowed. (e.g. your diff does not yet include `__array_interface__`, `__cuda_array_interface__` and others. I think `Any` is just good enough given that it's only a fallback for old NumPy versions. 

It's a lot of clutter that leads to type checkers falsly flagging correct code.

src/gt4py/storage/storage.py
gronerl: dtype is only derived from `data` if it is `None` in the explicit argument. Otherwise, it can be e.g. `str` instead and needs to be marshalled through `np.dtype`

gronerl:
I fixed as suggested or left a comment on all your comments. @egparedes

docs/gt4py/quickstart.rst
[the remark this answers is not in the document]
egparedes: Fair enough.

src/gt4py/storage/storage.py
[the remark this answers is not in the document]
egparedes: Fair enough.

src/gt4py/storage/storage.py
egparedes: Sure, you're fully right here. My mistake.

egparedes (approved):
It looks good.

gronerl:
bors try

egparedes:
bors try

gronerl:
bors try

egparedes:
bors try

PR #870, opened 2022-07-26
Functional: Grid deduction cleanup

tehrengruber:
The grid type used to determine whether `unstructured_domain` or `cartesian_domain` is emitted in PAST lowering is currently deduced in the `Program` decorator. This deduction should be the responsibility of the lowering and is hence moved to be located in the PAST -> ITIR lowering.


tehrengruber:
Closing... PAST is sadly not sufficient yet to deduce the grid type as the offsets of all captured field operators need to be collected recursively.

PR #871, opened 2022-07-26
Enable `run_gtfn` by default in iterator_tests

havogt:
Disables tests that are not supported yet:
- scan
- using `origin` attribute in field
- tuples in interface
- ...

### DIFF
--- a/tests/functional_tests/iterator_tests/test_toy_connectivity.py
+++ b/tests/functional_tests/iterator_tests/test_toy_connectivity.py
@@ -2,7 +2,7 @@ import numpy as np
 import pytest
 
 from functional.common import Dimension
-from functional.fencil_processor_no_gtfn_execs.formatters import gtfn
+from functional.fencil_processors.formatters import gtfn
 from functional.iterator.builtins import deref, lift, reduce, shift
 from functional.iterator.embedded import (
     NeighborTableOffsetProvider,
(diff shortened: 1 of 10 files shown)


tests/functional_tests/iterator_tests/test_toy_connectivity.py
```diff
@@ -110,15 +110,15 @@ def sum_edges_to_vertices(in_edges):
     )
 
 
-def test_sum_edges_to_vertices(fencil_processor):
-    fencil_processor, validate = fencil_processor
+def test_sum_edges_to_vertices(fencil_processor_no_gtfn_exec):
+    fencil_processor_no_gtfn_exec, validate = fencil_processor_no_gtfn_exec
```
tehrengruber, line 114:
```suggestion
    fencil_processor, validate = fencil_processor_no_gtfn_exec
```
As mentioned in #869 I didn't see the point of doing this. Activating the tests is much easier and the rest of the test remains more readable.

tehrengruber (changes requested):
Same comment as in #869.

tests/functional_tests/iterator_tests/test_toy_connectivity.py
havogt: The only reason was that my find/replace was simpler. I can change it...

tehrengruber (approved):

PR #874, opened 2022-07-27
Embedded backend performance improvements

tehrengruber:
- Ignore expensive _tupsum calls if no offsets are given to LocatedField
- Use tuple instead of list for sparse shifts
- Avoid unnecessary copy of position in execute_shift

### DIFF
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -20,7 +20,6 @@ from typing import (
     TypeAlias,
     TypeGuard,
     Union,
-    cast,
     runtime_checkable,
 )
 
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -668,7 +668,7 @@ def get_ordered_indices(
             if isinstance(elem, list):
                 pos = copy.copy(pos)  # copy such that we can modify
                 # we consume a sparse entry, this smells...
-                elem, *pos[axis.value] = cast(SparsePositionEntry, elem)  # type: ignore[index]
+                elem = pos[axis.value].pop(0)
             assert isinstance(elem, (int, slice))
             res.append(elem)
     return tuple(res)
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -2,11 +2,11 @@
 
 from __future__ import annotations
 
+import copy
 import itertools
 import math
 import numbers
 from abc import abstractmethod
-from copy import deepcopy
 from dataclasses import dataclass
 from types import NoneType
 from typing import (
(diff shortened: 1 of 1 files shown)


src/functional/iterator/embedded.py
```diff
@@ -370,8 +371,7 @@ def execute_shift(
     assert tag in offset_provider
     offset_implementation = offset_provider[tag]
     if isinstance(offset_implementation, Dimension):
-        assert offset_implementation.value in pos
-        new_pos = pos.copy()
+        new_pos = copy.copy(pos)
```
havogt, line 374: Is there any difference between `copy.copy(dict`) and `dict.copy()`?

src/functional/iterator/embedded.py
```diff
@@ -667,10 +666,11 @@ def get_ordered_indices(
             assert axis.value in pos
             elem = pos[axis.value]
             if isinstance(elem, list):
-                res.append(elem.pop(0))  # we consume a sparse entry, this smells...
-            else:
-                assert isinstance(elem, (int, slice))
-                res.append(elem)
+                pos = copy.copy(pos)  # copy such that we can modify
+                # we consume a sparse entry, this smells...
+                elem, *pos[axis.value] = cast(SparsePositionEntry, elem)  # type: ignore[index]
```
havogt, line 671: What about
```suggestion
                elem = pos[axis.value].pop(0)
```

havogt (changes requested):

src/functional/iterator/embedded.py
tehrengruber: Only with respect to typing. `Mapping` does not guarantee `mapping.copy` to exist.

havogt (approved):
There is just the unused import cast.

src/functional/iterator/embedded.py
```diff
@@ -20,6 +20,7 @@
     TypeAlias,
     TypeGuard,
     Union,
+    cast,
```
havogt, line 23:
```suggestion
```

src/functional/iterator/embedded.py
```diff
@@ -654,20 +653,28 @@ def _is_field_axis(axis: Axis) -> TypeGuard[FieldAxis]:
...
         else:
             assert _is_field_axis(axis)
             assert axis.value in pos
             elem = pos[axis.value]
-            if isinstance(elem, list):
-                res.append(elem.pop(0))  # we consume a sparse entry, this smells...
+            if _is_sparse_position_entry(elem):
+                sparse_position_tracker.setdefault(axis.value, 0)
+                sparse_position_tracker[axis.value] += 1
+                res.append(elem[sparse_position_tracker[axis.value]])
```
havogt, line 677:
```suggestion
                res.append(elem[sparse_position_tracker[axis.value]])
                sparse_position_tracker[axis.value] += 1
```

PR #879, opened 2022-07-28
Remove duplicated GTCallable

havogt:
We accidentally duplicated GTCallable. I think it makes sense to keep it in a separate file and moved it one level up to be reachable from the iterator module in the future.


tehrengruber (changes requested):
Looks good except the rename. `GTCallable` are a frontend construct currently, as there is no consistent typing with Iterator IR (e.g. `__gt_type__`).

havogt:
You mean the move? I disagree, it's frontend, but why should it be limited to fieldview frontend?

tehrengruber:
Yes. I would say the imports of a file should be a good indicator where something belongs. `GTCallable` imports from `ffront` and `iterator`. If we place it in functional we have a cycle `functional` -> `ffront` & `iterator` -> `functional`. Consequently `GTCallable` should be placed in a submodule of `functional`. If it would not use frontend concepts, e.g. the type, it could also be placed inside the `iterator` module. If we introduce a `__call__` method, e.g. when implementing embedded, it might make sense to move it elsewhere, but until then I believe it belongs into the frontend.

egparedes:
@havogt @tehrengruber is this PR is still relevant? If so, it should be merged soon.

I think the discussion arises from deciding if `functional.ffront` should be a subpackage only for the field view frontend or for any frontend. I tend to think of `ffront` as a package for all the frontend-related stuff, so I think `GTCallable` fits there, even if in the future we decide to add more subpackages for differente frontends (e.g. `functional.ffront.field_view`, `functional.ffront.local_view`, ....)

tehrengruber (approved):
We decided to keep `GTCallable` inside `ffront` and discuss that separately.

Issue #887, opened 2022-08-03
Race condition definitions differ between gt4py and gridtools


jdahm:
#885 so far fixes some but not quite all race conditions we face. It makes sure we don't add read-with-offset and write in a single vertical loop, but the problem is while we're writing them as multiple vertical loops (so gt4py sees no race conditions) gridtools still sees the race conditions.

havogt:
1) I agree. It is covered here https://github.com/GridTools/concepts/blob/master/GTScript-Parallel-model.md in point 4.
2) GridTools only fuses multistages if there is no extent, see https://github.com/GridTools/gridtools/blob/f174f81ba4fddca9c61ca4659be2cfaf8620b28e/include/gridtools/stencil/gpu/entry_point.hpp#L114. If you think it does otherwise, can you post an example?

jdahm:
>  2. GridTools only fuses multistages if there is no extent, see https://github.com/GridTools/gridtools/blob/f174f81ba4fddca9c61ca4659be2cfaf8620b28e/include/gridtools/stencil/gpu/entry_point.hpp#L114. If you think it does otherwise, can you post an example?


The code is a little difficult for me to understand at a glance. Can you tell if it would fuse this (including the deleted lines)? Yes it's pretty silly code, but it was literally ported over from Fortran. This was causing a race condition in our application until we removed that code.

![Screen Shot 2022-08-04 at 2 07 46 PM](https://user-images.githubusercontent.com/68032/182952875-ed48ea2f-4346-46b6-b62d-76e347c275a2.png)

jdahm:
Some more details about this for those who eventually stumble across this and want more info:

The user thinks they're writing valid code because `u_dt` and `v_dt` are read on center, then written to on center. However, the inlining pass creates an invalid state because the computation of `v3_1`, etc. are inlined wit horizontal offsets in the computation of `ue_1`, etc.. Therefore when they are later written to, it creates a write after read-with-offset (WAR) race.

PR #888, opened 2022-08-03
Add validation step to gtir_to_oir

jdahm:
Add a deeper validation of stencils on the OIR level using extent analysis.

Addresses #887.

### DIFF
--- a/src/gtc/gtir_to_oir.py
+++ b/src/gtc/gtir_to_oir.py
@@ -22,6 +22,15 @@ from gtc.passes.oir_optimizations.utils import compute_fields_extents
 
 
 def validate_stencil_memory_accesses(node: oir.Stencil) -> oir.Stencil:
+    """Check that no memory races occur in GridTools backends.
+
+    Since this is required for GridTools backends, it's imposed on all backends
+    at the OIR level. This is similar to the check at the gtir level for read-with-offset
+    and writes, but more complete because it involves extent analysis, so it catches
+    indirect read-with-offset through temporaries.
+
+    """
+
     def _writes(node: oir.Stencil) -> Set[str]:
         result = set()
         for left in node.iter_tree().if_isinstance(oir.AssignStmt).getattr("left"):
@@ -36,10 +45,11 @@ def validate_stencil_memory_accesses(node: oir.Stencil) -> oir.Stencil:
     names: Set[str] = set()
     for name in write_fields:
         if not field_extents[name].is_zero:
+            print(name, field_extents[name])
             names.add(name)
 
     if names:
-        raise ValueError(f"Found non-zero read extent on a written fields: {names}")
+        raise ValueError(f"Found non-zero read extent on written fields: {', '.join(names)}")
 
     return node
 
--- a/src/gtc/gtir_to_oir.py
+++ b/src/gtc/gtir_to_oir.py
@@ -45,7 +45,6 @@ def validate_stencil_memory_accesses(node: oir.Stencil) -> oir.Stencil:
     names: Set[str] = set()
     for name in write_fields:
         if not field_extents[name].is_zero:
-            print(name, field_extents[name])
             names.add(name)
 
     if names:
--- a/src/gtc/gtir_to_oir.py
+++ b/src/gtc/gtir_to_oir.py
@@ -13,11 +13,35 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from dataclasses import dataclass, field
-from typing import Any, List
+from typing import Any, List, Set
 
 from eve import NodeTranslator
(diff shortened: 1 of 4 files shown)


src/gtc/gtir_to_oir.py
```diff
@@ -13,11 +13,45 @@
...
+
+    field_names = {decl.name for decl in node.params if isinstance(decl, oir.FieldDecl)}
+    write_fields = _writes(node) & field_names
+
+    field_extents = compute_fields_extents(node)
+
+    names: Set[str] = set()
+    for name in write_fields:
+        if not field_extents[name].is_zero:
+            print(name, field_extents[name])
```
havogt, line 48:
```suggestion
```

src/gtc/gtir_to_oir.py
```diff
@@ -230,10 +264,12 @@ def visit_VerticalLoop(self, node: gtir.VerticalLoop, *, ctx: Context) -> oir.Ve
     def visit_Stencil(self, node: gtir.Stencil) -> oir.Stencil:
         ctx = self.Context()
         vertical_loops = self.visit(node.vertical_loops, ctx=ctx)
-        return oir.Stencil(
-            name=node.name,
-            params=self.visit(node.params),
-            vertical_loops=vertical_loops,
-            declarations=ctx.temp_fields,
-            loc=node.loc,
+        return validate_stencil_memory_accesses(
```
havogt, line 267: The place looks a bit weird. Maybe we can run this after the lowering, e.g. place it in the pipeline after the lowering. Or temporarily as a root_validator in the Stencil (temporarily because when we switch to the updated eve, we probably can not).

havogt (changes requested):
I didn't review deeply, if @gronerl has time, would be good if you could have a closer look.

src/gtc/gtir_to_oir.py
jdahm: Thanks for catching! Threw this together quickly...

src/gtc/gtir_to_oir.py
jdahm: I originally had it in `oir_pipeline.DefaultPipeline` as the first pass, but on second thought realized that passes should be reserved for optimizations, not correctness checks. What do you think?

jdahm:
I should probably add a test that trips this, as a gtir_to_oir unit test, even though I had to correct a few tests after I added it.

jdahm:
Questions that I thought about when creating this:
- Do we even want the equivalent check in gtir? This check catches all those cases.
- Should this always be enforced globally on the stencil level? Or, are there cases when read-with-offset and write are safe to do in different vertical loops?
- Where should we put correctness checks like this if the lowering isn't the right place, nor is the optimization pipeline?

src/gtc/gtir_to_oir.py
```diff
@@ -13,11 +13,44 @@
...
+
+    """
+
+    def _writes(node: oir.Stencil) -> Set[str]:
+        result = set()
+        for left in node.iter_tree().if_isinstance(oir.AssignStmt).getattr("left"):
+            result |= left.iter_tree().if_isinstance(oir.FieldAccess).getattr("name").to_set()
+        return result
+
+    field_names = {decl.name for decl in node.params if isinstance(decl, oir.FieldDecl)}
```
gronerl, line 40: Would we not want to include temporaries here?

gronerl (commented):
Why are the changes to `stencil_definitions.py` and to `test_code_generation.py` needed? Have they been invalid definitions all along? They were working before, can we not change the pipeline so that those cases are supported?

src/gtc/gtir_to_oir.py
jdahm: Because temporaries are allowed to be written to and read with a horizontal offset in the same stencil, since they're block-local. This check is supposed to catch data arrays where there are race conditions. This only happens in input/output fields.

jdahm:
> Why are the changes to stencil_definitions.py and to test_code_generation.py needed? Have they been invalid definitions all along?

Yes! they've been race conditions all along. There were a few that we had problems with in the past and I changed the definitions without really understanding the problem, but I think we have the root of the problem solved now, with this check.

> They were working before, can we not change the pipeline so that those cases are supported?

They were "kinda" working before, as I mentioned above. We had to play with the definitions at times to get them to work right, so it might have just been luck (and the small domain size they are run on) that they worked!

jdahm:
bors try

jdahm:
I don't quite understand why that didn't record the gitlab CI, but it passed!

gronerl:
bors try

jdahm:
@havogt is this ready to merge?

havogt (approved):
Looks good, sorry for the late approval.

PR #889, opened 2022-08-04
Add a new type of FencilProcessor for generating SourceModules

DropD:
## Description

Create a FencilGenerator interface (protocol, typing wrapper, decorator) for the step ITIR.FencilDefinition -> SourceModule. Make the current GTFN source module generator a FencilGenerator.

### Minor refactorings
* move `processor_interface` module to `fencil_processors` (by moving `iterator.trace` imports to where they are actually required)
* refactor local import-renaming of `source_modules` -> `defs`

## Requirements


tests/functional_tests/fencil_processor_tests/codegen_tests/test_gtfn.py
```diff
@@ -57,15 +60,15 @@ def fencil_example():
         ],
     )
     params = [
-        source_modules.BufferParameter("buf", ["I"], numpy.dtype(numpy.float32)),
-        source_modules.ScalarParameter("sc", numpy.dtype(numpy.float32)),
+        np_as_located_field(Dimension("I"))(numpy.empty((1,), dtype=numpy.float32)),
```
petiaccja, line 63: I think it would be better to first declare `IDim = Dimension("I")` to stick to the conventions.

I don't know the semantics of `Dimension("D") == Dimension("D")`. Is the identity tied to the object or to the string?

petiaccja (changes requested):
1. **Tracing**: Why are all the backends importing tracing just to register it? Is there not a better way to handle this?
2. **Pipeline & PRs**: I like the general direction of formalizing the two stages of ITIR->SourceModule->Callable conversion, but currently only the first stage is formalized, and this formalization is never used by a higher level structure that organizes the stages. Why not limit this PR to the refactorings and move the formalization to a separate PR?

tests/functional_tests/fencil_processor_tests/codegen_tests/test_gtfn.py
DropD: `Dimension` is a frozen dataclass with `.value` and a `.kind` attributes. By default, dataclasses generate an `__eq__` method which compares the values of the attributes. In the above example the identity is tied to the string, not the object's address in memory.

From the [dataclass docs](https://docs.python.org/3/library/dataclasses.html?highlight=dataclass#dataclasses.dataclass):
> eq: If true (the default), an __eq__() method will be generated. This method compares the class as if it were a tuple of its fields, in order. Both instances in the comparison must be of the identical type.

DropD:
> 2. **Pipeline & PRs**: I like the general direction of formalizing the two stages of ITIR->SourceModule->Callable conversion, but currently only the first stage is formalized, and this formalization is never used by a higher level structure that organizes the stages. Why not limit this PR to the refactorings and move the formalization to a separate PR?

I am planning to get away without the need for a higher level structure. My ideal pipeline design has approximately the following usage:
```
make_build_project((sm := make_source_module(fencil)), make_bindings(sm)).get_fencil_function()(*fencil_args)
```
for IR -> SourceModule -> build system of choice -> callable.

This current PR is a necessary step on the way and it is self contained. Therefore it should not have to wait for the subsequent steps to be reviewed and merged. Furthermore the refactorings came out of reducing friction while working on this, so they could be separated after the fact but what would be the benefit of that?

src/functional/fencil_processors/processor_interface.py
```diff
@@ -0,0 +1,155 @@
...
+PROCESSOR_RETURN_T = TypeVar("PROCESSOR_RETURN_T", covariant=True)
+PROCESSOR_KIND_T = TypeVar("PROCESSOR_KIND_T", bound="FencilProcessorProtocol", covariant=True)
+
+
+class FencilProcessorFunction(Protocol[PROCESSOR_RETURN_T]):
+    def __call__(self, fencil: FencilDefinition, *args, **kwargs) -> PROCESSOR_RETURN_T:
+        ...
+
+
+@runtime_checkable
```
egparedes, line 42: I don't think it makes a lot of sense to make this protocol runtime_checkable since it will only check if the object is callable and if it has a `kind` member, but the important check here is the actual value in `kind`. I propose to delete the annotation and instead create typeguards functions for each of the kinds which can be used at runtime and they also provide information to the type checker. For example, if `kind` is converted to a property then it can be as simple as:

```python
def is_fencil_formatter(obj: object, ) -> TypeGuard[FencilFormatter]:
    return callable(obj) and getattr(obj, "kind", None) is FencilFormatter

```

egparedes (commented):
It looks good to me. Added some suggestions and discussion points to try to simplify even further the processor_interface module.

src/functional/fencil_processors/processor_interface.py
DropD: I actually thought about changing it to a property. Seems to be the way to go.

src/functional/fencil_processors/processor_interface.py
```diff
@@ -0,0 +1,155 @@
...
+class FencilProcessorProtocol(
+    FencilProcessorFunction[PROCESSOR_RETURN_T], Protocol[PROCESSOR_RETURN_T, PROCESSOR_KIND_T]
+):
+    __name__: str = ""  # dummy to convince mypy that a __name__ is present for error messages
```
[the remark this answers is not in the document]
DropD: I believe that it should, thanks!

src/functional/fencil_processors/processor_interface.py
```diff
@@ -0,0 +1,155 @@
...
+    >>> ensure_processor_kind(format_foo, FencilFormatter)
+    """
+
+    class _FormatterClass(FencilFormatter):
+        def __call__(self, fencil: FencilDefinition, *args, **kwargs) -> str:
+            return func(fencil, *args, **kwargs)
+
+    formatter_instance = _FormatterClass()
+    update_wrapper(formatter_instance, func)
+    return formatter_instance
```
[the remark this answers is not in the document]
DropD: I will try that. Funnily this is almost exactly what my first draft looked like weeks ago.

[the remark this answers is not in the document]
DropD: And the reason why I abandoned it is that mypy righly points out that `func` does not have an attribute `kind`. I see multiple ways to circumvent that problem but none of them seem very good.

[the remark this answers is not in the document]
DropD: It doesn't work without some type ignores, however, because it is not necessary anymore to wrap the function into a class, the reason for using `__name__` in the error message goes away, and therefore the need to type declare `__name__` on the protocol class etc.

src/functional/fencil_processors/processor_interface.py
```diff
@@ -0,0 +1,139 @@
...
+
+class FencilFormatter(FencilProcessorProtocol[str, "FencilFormatter"], Protocol):
+    @property
+    def kind(self) -> type[FencilFormatter]:
+        return FencilFormatter
+
+
+def fencil_formatter(
+    func: FencilProcessorFunction[str],
+) -> FencilProcessorProtocol[str, FencilFormatter]:
```
egparedes, line 58: Doesn't mypy accept the concrete `FencilFormatter` type as return type?

```suggestion
def fencil_formatter(func: FencilProcessorFunction[str]) -> FencilFormatter:
```

src/functional/fencil_processors/processor_interface.py
```diff
@@ -0,0 +1,139 @@
...
+    FencilProcessorProtocol[SourceModule, "FencilSourceModuleGenerator"], Protocol
+):
+    @property
+    def kind(self) -> type[FencilSourceModuleGenerator]:
+        return FencilSourceModuleGenerator
+
+
+def fencil_source_module_generator(
+    func: FencilProcessorFunction[SourceModule],
+) -> FencilProcessorProtocol[SourceModule, FencilSourceModuleGenerator]:
```
egparedes, line 86: Same comment as above about using the concrete type in the return annotation

src/functional/fencil_processors/processor_interface.py
```diff
@@ -0,0 +1,139 @@
...
+
+class FencilExecutor(FencilProcessorProtocol[None, "FencilExecutor"], Protocol):
+    @property
+    def kind(self) -> type[FencilExecutor]:
+        return FencilExecutor
+
+
+def fencil_executor(
+    func: FencilProcessorFunction[None],
+) -> FencilProcessorProtocol[None, FencilExecutor]:
```
egparedes, line 113: Same comment as above about using the concrete type in the return annotation

src/functional/fencil_processors/processor_interface.py
```diff
@@ -0,0 +1,139 @@
...
+
+    >>> ensure_processor_kind(badly_execute, FencilExecutor)
+    """
+    # this operation effectively changes the type of func and that is the intention here
+    func.kind = FencilExecutor  # type: ignore[attr-defined]
+    return cast(FencilProcessorProtocol[None, FencilExecutor], func)
+
+
+def is_processor_kind(
+    obj: object, kind: type[ProcessorKindT]
```
egparedes, line 132: Nice trick to define a parametric typeguard! I'm wondering though, if it shouldn't contain also the `OutputT` also as type var of `FencilProcessorProtocol` to be able to narrow it down properly

```suggestion
    obj: FencilProcessorProtocol[OutputT, Any?? ProcessorKindT?? ...??], kind: type[ProcessorKindT]
```

src/functional/fencil_processors/processor_interface.py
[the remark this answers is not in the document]
egparedes: :+1:

egparedes (commented):
Well done, it's just ready to go. Just two additional questions about typing issues...

src/functional/fencil_processors/processor_interface.py
DropD: It does not consider `FencilFormatter` a concrete type. That makes sense, because it is deriving from `Protocol`.

src/functional/fencil_processors/processor_interface.py
DropD: If you want to do that, then we have the `__name__` problem again. Because (quote mypy):
>FencilProcessorProtocol[Any, Any] does not have a __name__ attribute

DropD: Also, this is basically only useful at runtime, where the output type would be completely ignored.

src/functional/fencil_processors/processor_interface.py
DropD: Basically the two are exchangeable here in every regard.

src/functional/fencil_processors/processor_interface.py
DropD: Ah, wait, we don't need the `__name__` anymore.

src/functional/fencil_processors/processor_interface.py
egparedes: Great. Could you also remove the unnecessary comma after the `func` parameter so black will be allowed to put the signature in one single line?

src/functional/fencil_processors/processor_interface.py
DropD: I added it but mind you, it is purely cosmetic since the `OutputT` only occurs that one time and doesn't have to match up with anything. Basically it is completely generic on `OutputT` with no constraints whatsoever.

egparedes: But I think it appears twice  in the `is_processor_kind` typeguard: once in the typing of the first argument and the other time in the `TypeGuard[]` result type, right?

egparedes (approved):
It looks good.

PR #883, opened 2022-08-02
Vertical for unstructured

havogt:
In unstructured: translates dimension corresponding to vertical to unstructured::dim::vertical.

PR #890, opened 2022-08-05
Small cleanup after #883

havogt:


DropD (approved):

src/functional/fencil_processors/codegens/gtfn/gtfn_ir.py
```diff
@@ -136,7 +136,7 @@ class StencilExecution(Node):
 
 class TagDefinition(Node):
     name: Sym
-    alias: Union[str, SymRef] = ""  # TODO eve has problem with Optional[str]
+    alias: Optional[Union[str, SymRef]] = None
```
havogt, line 139: hmm. unsure why it didn't work when I tried (at least 2 times).

PR #892, opened 2022-08-10
Cache settings

tehrengruber:
This PR adds the ability to pass the cache directory (both `root_path` and `dir_name`) to the `stencil` decorator. In FVM we compile multiple stencils in parallel and by default use a common cache path project-wide regardless of where the driver resides. Setting this directory in our stencil wrapper lead to a race-condition which we can avoid with the changes here.


tehrengruber:
bors try

src/gt4py/gtscript.py
```diff
@@ -235,6 +243,8 @@ def stencil(
         raise ValueError(f"Invalid 'name' string ('{name}')")
     if not isinstance(rebuild, bool):
         raise ValueError(f"Invalid 'rebuild' bool value ('{rebuild}')")
+    if cache_settings is not None and not isinstance(cache_settings, dict):
+        raise ValueError(f"Invalid 'cache_settings' dictionary ('{rebuild}')")
```
DropD, line 247: copy pasta? 
```suggestion
        raise ValueError(f"Invalid 'cache_settings' dictionary ('{cache_settings}')")
```

src/gt4py/gtscript.py
```diff
@@ -260,13 +270,17 @@ def stencil(
...
+        cache_settings = {}
+
     build_options = gt_definitions.BuildOptions(
         name=name,
         module=module,
         format_source=format_source,
         rebuild=rebuild,
         backend_opts=kwargs,
         build_info=build_info,
+        cache_settings=cache_settings,
```
DropD, line 283: Have you considered inlining the if block above?
```suggestion
        cache_settings=cache_settings or {},
```

DropD (changes requested):
👍 There is really just this one typo...

src/gt4py/gtscript.py
tehrengruber: wups

DropD (approved):

PR #893, opened 2022-08-10
More eve typing enhancements

egparedes:
Enhancements to type definitions and typing utils in eve.

Changes:
- add custom decorator to make Protocols checkable at runtime without performance degradation
- improve typing utils to cover hashable types
- improve other typing definitions from typeshed
- add frozenlist type

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,6 +49,7 @@ install_requires =
     devtools>=0.5
     deepdiff>=5.8
     frozendict>=2.3
+    frozenlist>=1.3.1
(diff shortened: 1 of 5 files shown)


fthaler (changes requested):
Performance is indeed significantly better on `pytest -k type_check` it gives at least 15% lower run times.
But I would like to drop the `FrozenList` as it more or less duplicates the builtin tuple without any reasonable advantage.

fthaler (approved):

PR #896, opened 2022-08-11
Remove global column_range in embedded

havogt:
TODO:
- [x] enable docstring tests in scan operator (once merged)
- [ ] architecture.md -> ADR


havogt:
Not possible like this if we want to support `scan` that has no input args, i.e. no iterators to attach this the column_range.

egparedes:
Proof of concept of alternative implementation which works for scans without args using `context_vars`: 
https://github.com/egparedes/gt4py/commit/67a3ba2b3f20e9e135521fb28a1f388589c5c781

havogt:
See #1120

PR #898, opened 2022-08-15
Functional: Add versioning config

egparedes:
Add infrastructure to support gt4py versioning in future commits.

Changes:
- Add version configuration in .bumpversion.cfg
- Add `version.py` submodules with `__version__` and `__version_info__` attributes to all components
- Change setup.cfg and config files in docs to use the proper version value
- Add `CHANGELOG.md` template
- Add basic unit tests


.bumpversion.cfg
```diff
@@ -0,0 +1,10 @@
+[bumpversion]
+current_version = 0.1.1
```
DropD, line 2: If we decided to decouple the `gt4py` and `eve` version numbers, would that be supported without changing the project structure? It seems to me there is only one single source of truth per repository supported by bumpversion.

Is it possible to make a `.bumpversion.gt4py.cfg` and a `.bumpversion.eve.cfg` and tell the CLI which to use? Or would we have one `.bumpversion.cfg` in `src/eve` and another one in `src/gt4py`?

egparedes: Good point. Decoupling eve and gt4py would only happen after changing the structure of the repo in such a way that eve, gt4py and other projects are folders at the root level of the repo. In this case, every project could have its own `.bumpversion.cfg` file. Anyway, the CLI tool is quite flexible and accepts different config files using the option `--config-file`

DropD: 👍

DropD (approved):

DropD (approved):

PR #899, opened 2022-08-16
Functional: pass scalar as constant field

tehrengruber:
On iterator level the arguments to all stencils / functions used inside a fencil closure need to be iterators (due to the compiled backend using a single SID composite to pass the arguments). #837 broke this which is fixed by this PR.


havogt (approved):
Looks good, except tests are not passing.

Maybe you could add a comment with a marker, e.g. "scalar_to_iterator" to link the places that you touched, in case we want to role back and also link to the note in the ADR. Since we don't use such a pattern yet, I am not sure if it makes sense to introduce it here. Maybe you have another idea.

Some general comments (maybe relevant for people reviewing this change at some point):
1. I would say currently a valid iterator program requires that the thing that is called in the closure is a stencil. A stencil is a function where all arguments are iterators. That's why we implement this change now. We could change that requirement, but it would require to change something between iterator ir and the cpp backend.
2. If we want to remove the requirement mentioned above, we could run a pass in iterator ir before lowering to gtfn_ir that promotes all values at the closure level to iterators and introduces the `deref`s where needed.
3. Alternative to 2, we could keep the requirement of a stencil in the closure, but lower from fieldview to something that is not a valid iterator ir program (because it violates closure calls a stencil) and then run the pass that makes it valid iterator ir.

tehrengruber:
I didn't use a marker, but improved the comments to make the connection clear.

I like approach 2 as the frontend is not bothered and with ITIR type inference such a pass is easy to write and comprehend.

PR #900, opened 2022-08-16
Functional: Check runtime argument in program call 

tehrengruber:
Currently when calling a program the arguments are not checked, leading to terrible error messages, sometimes deep in iterator ir. This PR checks deduces the types of all runtime arguments and checks they are compatible with the program.


src/functional/ffront/decorator.py
```diff
@@ -313,14 +313,24 @@ def format_itir(
...
+
+        try:
+            type_info.accepts_args(
+                self.past_node.type,
+                with_args=arg_types,
+                with_kwargs=kwarg_types,
+                raise_exception=True,
+            )
+        except GTTypeError as err:
+            raise ProgramTypeError.from_past_node(
```
tehrengruber, line 330: Maybe we should move `ProgramTypeError` somewhere else than the TypeDeduction. Not sure where though.

tests/functional_tests/ffront_tests/test_program.py
```diff
@@ -207,3 +206,25 @@ def prog(
...
+        copy_program(inp, out, offset_provider={})
+
+    msgs = [
+        "- Expected 0-th argument to be of type Field\[\[IDim], dtype=float64\],"
+        " but got Field\[\[JDim\], dtype=float64\].",
+        "- Expected 1-th argument to be of type Field\[\[IDim], dtype=float64\],"
+        " but got Field\[\[JDim\], dtype=float64\].",
+    ]
+    for msg in msgs:
+        assert re.search(msg, exc_info.value.__context__.args[0]) is not None
```
havogt, line 230: more style comment: if you need a complicated test condition, it's an indication that you are testing too much in this test. I'd just make one argument wrong and use standard pytest.raises(..., match=r"...") mechanism.

havogt (commented):
Looks good to me. I don't know if someone with better frontend knowledge wants to have a second look.

tests/functional_tests/ffront_tests/test_program.py
tehrengruber: That's true. I had chosen to use the wrong dimensions for both args as the `out` argument is special. But only on field operator not program level so this is not needed here. I only test for one argument now.

The purpose of this test is to test that the user becomes a meaningful error message signaling what is wrong. The outer exception that `pytest.raises(..., match=r"...")` matches on, i.e. `invalid call to program ...`, does not contain that information. Hence I am testing the inner exception, which I inconveniently have to access using `exc_info.value.__context__`. The `raise outer_exception from inner_exception` pattern we use is anyway not suitable for giving users good error messages and we should change that. Then this will disappear and we just use `pytest.raises(..., match=r"...")`.

havogt: ah, I missed that...

havogt (approved):

tests/functional_tests/ffront_tests/test_program.py
egparedes: Just a comment here for future reference/discussions: when using the `raise from` pattern, the inner exception is easily reachable through the `__cause__`  attribute (see [PEP-3134](https://peps.python.org/pep-3134/)). Also note that using `raise from` pattern does not prevent to give good error messages since the message of the outer exception can always be customized on creation when catching the inner exception.

PR #901, opened 2022-08-17
Add a New Offset Provider for the New Sparse Field Formulation

mroethlin:
## Description

This PR introduces a new offset provider, required to implement the [new sparse field formulation](https://hackmd.io/@havogt/rysAP5yPc)

## Requirements

### DIFF
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -71,7 +71,7 @@ class NeighborTableOffsetProvider:
         self.max_neighbors = max_neighbors
         self.has_skip_values = has_skip_values
 
-class NewSparseOffsetProvider:
+class StridedNeighborOffsetProvider:
     def __init__(
         self,
         shape,
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -420,7 +420,7 @@ def execute_shift(
             raise AssertionError()
         return new_pos
     else:
-        assert isinstance(offset_implementation, Connectivity):
+        assert isinstance(offset_implementation, Connectivity)
         assert offset_implementation.origin_axis.value in pos
         new_pos = pos.copy()
         new_pos.pop(offset_implementation.origin_axis.value)
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -419,7 +419,8 @@ def execute_shift(
         else:
             raise AssertionError()
         return new_pos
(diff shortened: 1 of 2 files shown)


src/functional/iterator/embedded.py
```diff
@@ -72,6 +71,32 @@ def __init__(
         self.max_neighbors = max_neighbors
         self.has_skip_values = has_skip_values
 
+class NewSparseOffsetProvider:
```
havogt, line 74:
```suggestion
class StridedNeighborOffsetProvider:
```

havogt (changes requested):
I'd propose a different name for the offset provider. I started to introduce the Connectivity protocol but it is not used in all places consistently. In a next round of cleanup we should better define the Connectivity concept to avoid the weird `tbl` usage.

src/functional/iterator/embedded.py
```diff
@@ -391,7 +416,7 @@ def execute_shift(
         else:
             raise AssertionError()
         return new_pos
-    elif isinstance(offset_implementation, NeighborTableOffsetProvider):
+    elif isinstance(offset_implementation, NeighborTableOffsetProvider) or isinstance(offset_implementation, NewSparseOffsetProvider):
```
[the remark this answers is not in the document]
tehrengruber: The `Connecitivity` is not complete and sufficient for what is used below. Additionally `isinstance` checks with protocols are expensive and I would avoid them in such a performance critical part. I would propose either `isinstance(offset_implementation, (NeighborTableOffsetProvider, NewSparseOffsetProvider))` for now or an extension of `Connectivity` but instead of an `elif` branch just an assert that can be disabled in non-debug mode.

src/functional/iterator/embedded.py
```diff
@@ -391,7 +419,7 @@ def execute_shift(
         else:
             raise AssertionError()
         return new_pos
-    elif isinstance(offset_implementation, NeighborTableOffsetProvider):
+    elif isinstance(offset_implementation, Connectivity):
```
havogt, line 422: Ok, let's do 
```suggestion
    else:
        assert isinstance(offset_implementation, Connectivity):
```
as Till suggested. 
Also then remove the AssertionError in line 434.

@tehrengruber I don't get why it is not sufficient here. Connectivity is not well defined, e.g. the `tbl` should not be part of the public API, but as far as I can see it should work here, no?

havogt (approved):
Approving my own changes. 🙃

PR #903, opened 2022-08-17
Make build systems interchangeable (prepare for new alternative build system)

DropD:
## Description

Tweak the compiled backends code to allow every step of the process to be executed from the output of the previous step. Add Protocols for the steps as well as the step outputs to ensure that steps are as interchangeable as possible.

This will be the basis for two future directions of work:
* an additional build system, which uses CMake only once per session (and per language / dependency sets) to get a compile command and then skips CMake configure calls. (Use case: testing, interactive experimentation).
* caching and versioning of compilation results (by versioning I mean avoiding that different compiled versions of the same fencil clash).

## For reviewers

I would recommend starting with `gtfn_cpu`, as that change is what everything else builds up to.

## Requirements


src/functional/fencil_processors/builders/cpp/__init__.py
```diff
@@ -11,9 +11,3 @@
 # distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
-
```
DropD, line 14: Not needed anymore, since the `.callable` module is gone.

src/functional/fencil_processors/builders/cpp/bindings.py
```diff
@@ -171,22 +171,23 @@ def make_argument(
 
 
 def create_bindings(
-    target: source_modules.Function, target_header: str
+    source_module: source_modules.SourceModule, language: CppLanguage = CPP_DEFAULT
```
DropD, line 174: While it is true that this particular bindings creator does not require the full `source_module`, when caching and name mangling for versioning purposes come into play that might not be true anymore. In addition, we want to call each step directly with the output of the previous step.

This means, it might be better to combind this `language` with the `SourceModule.language` tag to further streamline the interface.

src/functional/fencil_processors/builders/cpp/build.py
```diff
@@ -135,61 +136,82 @@ def _get_python_module_suffix():
...
-        self._extension = _get_python_module_suffix()
-        self._cmakelists = _render_cmakelists(name, dependencies, list(sources.keys()))
-        self._sources = sources
-
-    @classmethod
-    def get_binary_path(
-        cls, root_folder: pathlib.Path, name: str, extension: str = _get_python_module_suffix()
-    ) -> pathlib.Path:
-        return root_folder / _BUILD_SUBDIR / "bin" / (name + "." + extension)
+@dataclass(frozen=True)
```
DropD, line 139: The thought behind making this a frozen dataclass is to make it hashable which could be beneficial for caching. It might be considered premature and comes at the cost that some of the properties (see below) have to be dynamically recomputed multiple times (although they are cheap).

There are two ways of solving this: either CMakeProject itself is again made stateful and the properties can be computed at construction (hashability is gone then) or the class is split into a dataclass which is hashable and a non-hashable class which can be created from it (this second class could be free of any hashing concerns and know only how to deal with files and the `cmake` command).

I am not sure it is worth deciding either way before we start implementing both the caching and the second build system. Both might influence the decision.

src/functional/fencil_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -30,7 +30,9 @@ def get_param_description(
 ) -> source_modules.ScalarParameter | source_modules.BufferParameter:
     view = numpy.asarray(obj)
     if view.ndim > 0:
-        return source_modules.BufferParameter(name, [dim.value for dim in obj.axes], view.dtype)
+        return source_modules.BufferParameter(
```
DropD, line 33: Have a look at `pipeline.py` for an explanation for the change to tuples.

tests/functional_tests/fencil_processor_tests/builder_tests/cpp_tests/test_cpp_pipeline.py
```diff
@@ -21,7 +21,10 @@
 import pytest
```
DropD, line 21: This replaces both the `test_build` as well as `test_callable` tests.

src/functional/fencil_processors/source_modules/source_modules.py
```diff
@@ -29,14 +55,14 @@ class ScalarParameter:
 @dataclass(frozen=True)
 class BufferParameter:
     name: str
-    dimensions: Sequence[str]
+    dimensions: tuple[str, ...]
```
DropD, line 58: Changing to tuple here and elsewhere to ensure they stay immutable (hashable), which is essential for caching to work well in the future.

src/functional/fencil_processors/runners/gtfn_cpu.py
```diff
@@ -13,36 +13,56 @@
...
+        """
+        return CMakeProject(
+            source_module=(
+                source_module := gtfn_module.GTFNSourceModuleGenerator(self.language_settings)(
+                    fencil, *args, **kwargs
+                )
+            ),
+            bindings_module=bindings.create_bindings(source_module),
+            cache_strategy=cache.Strategy.SESSION,
+        ).get_implementation()(*[convert_arg(arg) for arg in args])
```
DropD, line 61: This is the pipeline in action. Every step feeds directly into the next. Most of the other changes in this PR are aimed at achieving this.

src/functional/fencil_processors/runners/gtfn_cpu.py
```diff
@@ -13,36 +13,56 @@
...
+
+    @property
+    def __name__(self):
+        return self.name or repr(self)
```
DropD, line 65: This one is not strictly necessary but allows for easier to read test fixture ids.

src/functional/fencil_processors/runners/gtfn_cpu.py
```diff
@@ -13,36 +13,56 @@
...
         return arg
 
 
-# TODO(ricoh): change style to declarative pipeline
-@fencil_executor
-def run_gtfn(itir: ir.FencilDefinition, *args, **kwargs):
-    """
-    Execute the iterator IR fencil with the provided arguments.
+@dataclass(frozen=True)
+class GTFNExecutor(FencilExecutor):
```
DropD, line 39: Why turn this into a class? This comes from a discussion with @petiaccja a while back, where we thought of customizing fencil processors by generating them with customization, rather than passing configuration through the keyword args.

In this case I chose a class, rather than a function generating a function because I suspect it is more readable to the uninitiated. Also it allows naming the configured processor, which helps with test fixture ids.

src/functional/fencil_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -30,43 +34,53 @@ def get_param_description(
...
-    *args,
-    **kwargs,
-) -> source_modules.SourceModule:
-    """Generate GTFN C++ code from the ITIR definition."""
-    parameters = [
-        get_param_description(itir_param.id, obj) for obj, itir_param in zip(args, itir.params)
-    ]
-    function = source_modules.Function(itir.id, parameters)
+@dataclass(frozen=True)
+class GTFNSourceModuleGenerator(FencilSourceModuleGenerator):
```
DropD, line 45: This has become a class for the same reason as `gtfn_cpu.GTFNExecutor`. More detail there.

src/functional/fencil_processors/source_modules/cpp_gen.py
```diff
@@ -14,15 +14,35 @@
...
+from eve.codegen import format_source
+from functional.fencil_processors.source_modules import source_modules
 
 
-LANGUAGE_ID: Final = "cpp"
+@dataclass(frozen=True)
+class CppLanguage:
+    """Implements source_modules.IncludeImplementationLanguageProtocol for C++ language settings."""
+
+    name: str
```
petiaccja, line 31: If we identify languages by the class type, `name` here is redundant. If we still want a string name for messages to the user, it might be nicer to use pretty names like `"C++"` and `"Python"`.

src/functional/fencil_processors/source_modules/cpp_gen.py
```diff
@@ -14,15 +14,35 @@
...
+    implementation_extension: str
+    include_extension: str
+    formatting_style: str
```
petiaccja, line 34: Maybe a comment on where to find the available options for formatting styles? Perhaps an enum in eve?

src/functional/fencil_processors/source_modules/source_modules.py
```diff
@@ -15,11 +15,37 @@
...
+
+    @property
+    def name(self) -> str:
+        ...
+
+    def format_source(self, source_code: str) -> str:
+        ...
+
+
+class IncludeImplementationLanguageProtocol(SupportedLanguageProtocol, Protocol):
```
petiaccja, line 34: How about `LanguageFileExtensionsProtocol` or `SourceFileExtensionsProtocol` or something similar? I interpreted the name as "include the ImplementationLanguageProtocol" and it was pretty confusing.

petiaccja (changes requested):

src/functional/fencil_processors/source_modules/cpp_gen.py
```diff
@@ -14,15 +14,35 @@
...
+
+    def format_source(self, source_code: str) -> str:
+        return format_source(self.name, source_code, style=self.formatting_style)
+
+
+CPP_DEFAULT: Final = CppLanguage(
+    name="cpp",
+    implementation_extension=".cpp",
+    include_extension=".cpp.inc",
```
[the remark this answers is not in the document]
DropD: I see it like this: if someone writes a backend that separates headers cleanly, then they are free to provide a different default `CppLanguage` instance, which sets `implementation_extension` to "hpp".

Only once someone really writes a backend that needs three estensions at the same time should there be a new language protocol and a default implementation for it. Seems pointless otherwise.

src/functional/fencil_processors/source_modules/cpp_gen.py
```diff
@@ -14,15 +14,35 @@
...
+    include_extension: str
+    formatting_style: str
+
+    def format_source(self, source_code: str) -> str:
+        return format_source(self.name, source_code, style=self.formatting_style)
+
+
+CPP_DEFAULT: Final = CppLanguage(
+    name="cpp",
+    implementation_extension=".cpp",
```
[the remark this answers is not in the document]
DropD: Good point.

src/functional/fencil_processors/source_modules/cpp_gen.py
DropD: I will leave this up to eve...

src/functional/fencil_processors/source_modules/source_modules.py
DropD: * `LanguageFileExtensionsProtocol`: looks like `Python` might want to implement that. Except python does not split "include" and "implementation" files. -> Too general
* `SourceFileExtensionsProtocol`: `Python` also has source file extensions...
* `LanguagesWhichSplitImplementationAndIncludeFilesProtocol`: too long
* `HeaderAndImplementationFilesLanguageProtocol`: a bit long.

Considering we don't really need the `Protocol`, maybe `LanguageWithHeaders`?

src/functional/fencil_processors/source_modules/source_modules.py
```diff
@@ -46,14 +72,14 @@ class LibraryDependency:
...
     entry_point: Function
     source_code: str
-    library_deps: Sequence[LibraryDependency]
-    language: str
+    library_deps: tuple[LibraryDependency, ...]
+    language: LanguageT
 
 
 @dataclass(frozen=True)
 class BindingModule:
```
egparedes, line 83: If `SourceModule` is generic with respect to a `LanguageT` typevar, shouldn't `BindingModule` be generic with respect to the target language, and maybe even also with respect to the `SourceModule` language (contravariantly)?

src/functional/fencil_processors/pipeline.py
```diff
@@ -0,0 +1,40 @@
...
+
+from .builders.cache import Strategy as CacheStrategy
+from .source_modules.source_modules import BindingModule, SourceModule, SupportedLanguage
+
+
+LanguageT_contra = TypeVar("LanguageT_contra", bound=SupportedLanguage, contravariant=True)
+
+
+class BindingsGenerator(Protocol[LanguageT_contra]):
+    def __call__(self, source_module: SourceModule[LanguageT_contra]) -> BindingModule:
```
egparedes, line 24: Just a question: why is the type not covariantly parametrized with respect to the output binding language?

src/functional/fencil_processors/builders/cpp/build.py
```diff
@@ -18,14 +18,16 @@
...
 from eve import Node
 from eve.codegen import JinjaTemplate as as_jinja, TemplatedGenerator
 from functional.fencil_processors import source_modules
-
-
-_BUILD_SUBDIR: Final = "build"
+from functional.fencil_processors.builders import cache
+from functional.fencil_processors.builders.importer import import_from_path
+from functional.fencil_processors.pipeline import BuildProject
+from functional.fencil_processors.source_modules.source_modules import LanguageWithHeaders
```
egparedes, line 30: Too many direct imports: https://google.github.io/styleguide/pyguide.html#22-imports

egparedes (commented):
I have several comments but it looks good. The main issue in my opinion is the design of the Language Protocol as a protocol for instances and not for classes used as singleton/enums. Check my comments for further details and discussion.

src/functional/fencil_processors/pipeline.py
DropD: This took me a while to get (mypy complained about using the covariant type var), because I hadn't understood co- and contravariant all the way.

A BindingsGenerator written for the most general language protocol works with specific ones but not vice-versa. So you can substitute *super*-classes for *sub*-classes (contravariant) and not *sub*-classes for *super*-classes (covariant).

src/functional/fencil_processors/source_modules/source_modules.py
DropD: Probably should be 👍

src/functional/fencil_processors/pipeline.py
DropD: On the other hand, now mypy complains the other way around, so I apparently don't understand anything.

src/functional/fencil_processors/runners/gtfn_cpu.py
```diff
@@ -13,36 +13,56 @@
...
 from functional.iterator import ir
 
 
 def convert_arg(arg: Any) -> Any:
     view = numpy.asarray(arg)
     if view.ndim > 0:
-        return memoryview(view)
+        return memoryview(view)  # type: ignore[arg-type] # mypy seems unaware that ndarray is compatible with buffer protocol
     else:
         return arg
```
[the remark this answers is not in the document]
DropD: It is this function which is missing something, namely support for arguments other than scalars and located fields (for example constant fields are not supported). That support is explicitly out of scope, however if the tests pass without wrapping arrays in `memoryview`, then I don't mind removing that.

src/functional/fencil_processors/pipeline.py
egparedes: In my experience, understanding covariance and contravariance is much simpler than understanding what the hell is mypy understanding for covariance and contravariance. I almost gave up completely on using covariant and contravariant typevars in Python annotations...

src/functional/fencil_processors/runners/gtfn_cpu.py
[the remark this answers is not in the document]
DropD: Funny story: I commented out the `memoryview` line and suddenly all the builtins tests passed locally for the first time for me. Could have been something different but the change didn't make anything fail, so I am keeping it.

src/functional/fencil_processors/runners/gtfn_cpu.py
```diff
@@ -13,36 +13,56 @@
...
+            bindings_module=bindings.create_bindings(source_module),
+            cache_strategy=cache.Strategy.SESSION,
+        ).get_implementation()(*[convert_arg(arg) for arg in args])
+
+    @property
+    def __name__(self):
+        return self.name or repr(self)
+
+
+run_gtfn: FencilProcessorProtocol[None, FencilExecutor] = GTFNExecutor(name="run_gtfn")
```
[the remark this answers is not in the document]
DropD: My intention was actually not to support mypy in deducing the type. I am expressing that this should always implement the protocol. Is there a better way to warn future developers immediately in case of accidental incompatible changes?

src/functional/fencil_processors/source_modules/source_modules.py
```diff
@@ -15,11 +15,37 @@
...
 
 import numpy
 
 
+class SupportedLanguage(Protocol):
+    """Ensures consistent code formatting along the pipeline."""
+
+    @property
+    def name(self) -> str:
+        ...
```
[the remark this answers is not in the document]
DropD: It is intended to label different sets of language preferences (instances). If it was a class var it would be redundant (The class already has a readable name) and you would only be able to tell one set of `C++` preferences from another by looking at each of the values.

src/functional/fencil_processors/source_modules/source_modules.py
```diff
@@ -15,11 +15,37 @@
...
+class LanguageWithHeaders(SupportedLanguage, Protocol):
+    """Ensures consistent file naming for languages that split code into include (header) and implementation files."""
+
+    @property
+    def implementation_extension(self) -> str:
+        ...
+
+    @property
+    def include_extension(self) -> str:
+        ...
```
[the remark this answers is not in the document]
DropD: Same reply as above. If someone needs all the headers to be called `.my.custom.hpp` for their integration purposes, they would need to implement a new (otherwise redundant) class.

src/functional/fencil_processors/source_modules/cpp_gen.py
```diff
@@ -14,15 +14,35 @@
...
 import numpy
 
-from functional.fencil_processors import source_modules
+from eve.codegen import format_source
+from functional.fencil_processors.source_modules import source_modules
 
 
-LANGUAGE_ID: Final = "cpp"
+@dataclass(frozen=True)
+class CppLanguage:
```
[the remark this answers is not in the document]
DropD: Protocols are in a way necessarily singletons (only the class exists, no instances). However, the implementations are explicitly designed to not be singletons. There is a usable default value, which is not instanciated everywhere separately, which may look like a singleton?

petiaccja (changes requested):

src/functional/fencil_processors/source_modules/source_modules.py
petiaccja: A bit awkward still, but I can live with that.

src/functional/fencil_processors/runners/gtfn_cpu.py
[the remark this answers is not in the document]
petiaccja: Yeah, this is a mess. 

One problem is that numpy will convert literally anything to an array with `asarray`, even scalars. You must check the dimensions afterwards.

As far as I know, the buffer protocol is supported if the object can be converted to a memory view. The other problem is that our `Field` class does not support the buffer protocol, but it does support conversion to numpy arrays, and numpy arrays do support the buffer protocol.

The real fix to this would be to make `Field` support the buffer protocol and do an attempt to cast the argument into a memory view to detect support.

src/functional/fencil_processors/builders/cpp/bindings.py
```diff
@@ -153,7 +154,7 @@ def make_parameter(
     name = parameter.name
     ndim = 0 if isinstance(parameter, source_modules.ScalarParameter) else len(parameter.dimensions)
     scalar_type = parameter.scalar_type
-    return FunctionParameter(name=name, ndim=ndim, dtype=scalar_type)
+    return FunctionParameter(name=name, ndim=ndim, dtype=numpy.dtype(scalar_type))
```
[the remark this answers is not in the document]
DropD: This is outdated indeed. It was there to isolate the mypy error to the `memoryview` line which was removed later.

src/functional/fencil_processors/builders/cpp/bindings.py
```diff
@@ -165,28 +166,30 @@ def make_argument(
...
-    wrapper_name = target.name + "_wrapper"
+    if not isinstance(source_module.language, cpp.CppLanguage):
+        raise NotImplementedError("Can only create bindings for C++ source modules.")
+    wrapper_name = source_module.entry_point.name + "_wrapper"
 
     file_binding = BindingFile(
-        callee_header_file=target_header,
+        callee_header_file=source_module.entry_point.name
+        + "."
+        + source_module.language.include_extension,
```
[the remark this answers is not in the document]
DropD: 👍

src/functional/fencil_processors/pipeline.py
```diff
@@ -0,0 +1,40 @@
...
+        ...
+
+
+class BuildProject(Protocol):
```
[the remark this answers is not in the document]
DropD: Definitely an improvement!

src/functional/fencil_processors/runners/gtfn_cpu.py
```diff
@@ -13,36 +13,56 @@
...
-    source_module = gtfn_codegen.create_source_module(itir, *args, **kwargs)
-    wrapper = cpp_callable.create_callable(source_module)
-    wrapper(*[convert_arg(arg) for arg in args])
+        The fencil is compiled to machine code with C++ as an intermediate step,
+        so the first execution is expected to have a significant overhead, while subsequent
+        calls are very fast. Only scalar and buffer arguments are supported currently.
+
+        See ``FencilExecutorFunction`` for details.
+        """
+        return build.CMakeProject(
```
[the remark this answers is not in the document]
DropD: I would like to avoid that. It is an invitation to put custom logic between the steps (rather than improve the steps). If this is not readable enough, we indeed need a pipeline runner of some kind. Something along the lines of

```python
return <pipeline_runner_name>(
    source_module_gen=gtfn_module.GTFNSourceModuleGenerator(self.language_settings),
    binding_module_gen=bindings.create_bindings,
    buildable_project_gen=build.CMakeProject,
    cache_strategy=cache.Strategy.SESSION,
)(fencil).get_implementation()
```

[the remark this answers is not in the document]
DropD: I assume you meant something like

```python
# Never ever put any logic in between these lines!
# Instead, if something is missing it has to go into one of the functions we are calling here.
source_module = gtfn_module.GTFNSourceModuleGenerator(self.language_settings)(fencil, *args, **kwargs)
bindings_module = bindings.create_bindings(source_module)
cmake_project =  build.CMakeProject(
    source_module=source_module,
    bindings_module=bindings_module,
    cache_strategy=cache.Strategy.SESSION,
)
cpp_python_extension_function = cmake_project.get_implementation()
cpp_python_extension_function(*[convert_arg(arg) for arg in args])
```

src/functional/fencil_processors/source_modules/cpp_gen.py
DropD: I put that into the docstring of `source_modules.LanguageSettings`

src/functional/fencil_processors/builders/cpp/build.py
DropD: I would like to do that in the following PR, where I am touching the `CMakeProject` anyway. Also to avoid conflicts.

petiaccja: If you think this form is better for this PR and you remove it in the next, it's fine for me.

src/functional/fencil_processors/source_modules/cpp_gen.py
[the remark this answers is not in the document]
petiaccja: What use case have you had in mind for this extensions feature? I don't really understand how it's connected to unique names.

Include guards don't help here, they prevent multiple inclusions into the same translation unit, but they allow inclusion into many translation units. The problem with `.cpp.inc` is that it should be included into no more than one translation unit.

src/functional/fencil_processors/pipeline.py
DropD: side note: what mypy expects depends on the presence (or not) of the infamous `src/__init__.py`. So this might need to be revisited...

egparedes (commented):
It looks good to me. Just minor comments and questions.

src/functional/fencil_processors/runners/gtfn_cpu.py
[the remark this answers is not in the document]
petiaccja: Yeah, let's leave it like this with a TODO

src/functional/fencil_processors/builders/cpp/build.py
petiaccja: Rico, can you put a TODO on this before merging?

src/functional/fencil_processors/runners/gtfn_cpu.py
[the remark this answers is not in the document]
petiaccja: How about leaving the mess and waiting for the no-storage implementation? That'll make the problem go away, right?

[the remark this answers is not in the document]
egparedes: Not at all. The no-storage implementation in functional will only add the feature of allocating buffers with specific memory layouts.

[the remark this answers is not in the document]
egparedes: Hmmm, it's not exactly correct what I said before, let me rephrase it: the no-storage implementation will not solve this problem in particular but it will require to deal with field buffers in a different and more coherent way.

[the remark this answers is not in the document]
egparedes: Anyway, the important point of my suggestion is that we should restrict the conversion to ndarray only when the argument defines an `__array__` method, not always.

src/functional/fencil_processors/source_modules/cpp_gen.py
```diff
@@ -14,39 +14,59 @@
...
+    """Implements source_modules.IncludeImplementationLanguageProtocol for C++ language settings."""
+
+    name: str
+    implementation_extension: str
+    include_extension: str
+    formatting_style: str
```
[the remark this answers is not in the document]
DropD: Nope, dead code.

src/functional/fencil_processors/source_modules/cpp_gen.py
```diff
@@ -14,39 +14,59 @@
...
+
+    def format_source(self, source_code: str) -> str:
+        return format_source("cpp", source_code, style=self.formatting_style)
```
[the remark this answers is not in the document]
DropD: It's dead, thanks.

src/functional/fencil_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -13,60 +13,80 @@
...
-        f"""
-        #include <gridtools/fn/backend/naive.hpp>
-        {stencil_src}
-        {decl_src}
-        """.strip(),
-        style="LLVM",
+@dataclasses.dataclass(frozen=True)
+class GTFNSourceModuleGenerator(fpi.FencilSourceModuleGenerator):
+    language_settings: source_modules.LanguageWithHeaderFilesSettings = dataclasses.field(
+        default=cpp_gen.CPP_DEFAULT
```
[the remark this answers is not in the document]
DropD: I changed it in both places, thanks for noticing.

src/functional/fencil_processors/runners/gtfn_cpu.py
```diff
@@ -13,36 +13,61 @@
...
 
+# TODO(ricoh): Add support for the whole range of arguments that can be passed to a fencil.
 def convert_arg(arg: Any) -> Any:
-    view = numpy.asarray(arg)
+    view = np.asarray(arg)
     if view.ndim > 0:
-        return memoryview(view)
+        return view
     else:
         return arg
```
egparedes, line 35: Not critical for this PR, but still suggest to add a ToDo comment here or change it to something more restrictive like:
```python
if hasattr(arg, "__array__"):
    return np.asarray(arg)
else:
    return arg

```

egparedes (approved):
It looks good.

src/functional/fencil_processors/builders/cpp/bindings.py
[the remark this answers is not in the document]
DropD: The problem was actually in `test_bindings.py`. `ScalarParameter.dtype` is not validated, as opposed to `FunctionParameter.dtype`, so the test there was able to sneak in a `_ctypes.PyCSimpleType` instead.

src/functional/fencil_processors/runners/gtfn_cpu.py
DropD: I just went ahead and changed it, if it breaks tests I will revert and add a TODO.

petiaccja (approved):

PR #905, opened 2022-08-18
Implement better column handling

havogt:
Introduce Column class.

- Fixes problem with column range when start is != 0
- Fixes problem in tuple_get for Columns when tuple is constructed from numpy.array with `None` dimensions.

### DIFF
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -142,7 +142,11 @@ class MutableLocatedField(LocatedField, Protocol):
 
 
 class Column(np.lib.mixins.NDArrayOperatorsMixin):
-    """Represents a column when executed in column mode (`column_axis != None`)."""
+    """Represents a column when executed in column mode (`column_axis != None`).
+
+    Implements `__array_ufunc__` and `__array_function__` to isolate
+    and simplify dispatching in iterator ir builtins.
+    """
 
     def __init__(self, kstart: int, data: np.ndarray) -> None:
         self.kstart = kstart
@@ -165,10 +169,14 @@ class Column(np.lib.mixins.NDArrayOperatorsMixin):
 
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs) -> Column:
         assert method == "__call__"
+        assert all(inp.kstart == self.kstart for inp in inputs)
+        assert all(inp.data.shape == self.data.shape for inp in inputs)
         return self.__class__(self.kstart, ufunc(*(inp.data for inp in inputs), **kwargs))
 
     def __array_function__(self, func, types, args, kwargs) -> Column:
         assert all(issubclass(t, self.__class__) for t in types)
+        assert all(arg.kstart == self.kstart for arg in args)
+        assert all(arg.data.shape == self.data.shape for arg in args)
         return self.__class__(self.kstart, func(*(arg.data for arg in args), **kwargs))
 
 
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -553,14 +553,20 @@ def _get_axes(
 
 
 def _make_tuple(
-    field_or_tuple: LocatedField | tuple, indices: int | slice | tuple[int | slice, ...]
-) -> tuple:  # arbitrary nesting of tuples of LocatedField
+    field_or_tuple: LocatedField | tuple,  # arbitrary nesting of tuples of LocatedField
+    indices: int | slice | tuple[int | slice, ...],
+    *,
+    as_column: bool = False,
+) -> tuple:  # arbitrary nesting of tuples of field values or `Column`s
     if isinstance(field_or_tuple, tuple):
         return tuple(_make_tuple(f, indices) for f in field_or_tuple)
     else:
         data = field_or_tuple[indices]
-        if isinstance(data, np.ndarray):
-            return Column(0, data)
+        if as_column:
+            # wraps a vertical slice of an input field into a `Column`
+            return Column(
+                0, data
+            )  # TODO(havogt) when we support LocatedFields with origin (i.e. non-zero origin), kstart needs to be adapated here
         else:
             return data
 
@@ -623,7 +629,7 @@ class MDIterator:
             {**shifted_pos, **slice_column},
         )
         try:
-            return _make_tuple(self.field, ordered_indices)
+            return _make_tuple(self.field, ordered_indices, as_column=self.column_axis is not None)
         except IndexError:
             return _UNDEFINED
 
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -159,7 +159,7 @@ class Column(numpy.lib.mixins.NDArrayOperatorsMixin):
     def __setitem__(self, i: int, v: Any) -> None:
         self.data[i - self.kstart] = v
 
-    def __array__(self, dtype: Optional[DTypeLike]=None) -> np.ndarray:
+    def __array__(self, dtype: Optional[npt.DTypeLike] = None) -> np.ndarray:
         return self.data if dtype or dtype == self.data.dtype else self.data.astype(dtype)
 
     def __array_function__(self, func, types, args, kwargs):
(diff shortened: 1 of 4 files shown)


tests/functional_tests/iterator_tests/test_tuple.py
```diff
@@ -494,9 +488,9 @@ def test_tuple_of_tuple_of_field_input(fencil_processor_no_gtfn_exec):
         )
 
 
-@pytest.mark.xfail(
-    reason="tuple_get is not compatible with this way of expressing tuples.",
-)
+# @pytest.mark.xfail(
+#     reason="tuple_get is not compatible with this way of expressing tuples.",
+# )
```
havogt, line 493:
```suggestion
```

src/functional/iterator/embedded.py
```diff
@@ -143,8 +142,33 @@ def __array__(self) -> np.ndarray:
...
+        if self.data.dtype.names is not None:
+            return Column(self.kstart, self.data[f"f{i}"])
+        else:
+            return Column(self.kstart, self.data[i, ...])
+
+    def __setitem__(self, i: int, v: Any):
+        self.data[i - self.kstart] = v
+
+    def __array__(self, dtype=None) -> np.ndarray:
+        return self.data if dtype is None or dtype == self.data.dtype else self.data.astype(dtype)
```
tehrengruber, line 163: You can use `DTypeLike` from `numpy.typing` to annotate the argument.

```suggestion
    def __array__(self, dtype: Optional[DTypeLike]=None) -> np.ndarray:
        return self.data if dtype or dtype == self.data.dtype else self.data.astype(dtype)
```

src/functional/iterator/embedded.py
```diff
@@ -143,8 +142,33 @@ def __array__(self) -> np.ndarray:
...
+    def __getitem__(self, i: int):
+        return self.data[i - self.kstart]
+
+    def tuple_get(self, i: int):
+        if self.data.dtype.names is not None:
+            return Column(self.kstart, self.data[f"f{i}"])
+        else:
+            return Column(self.kstart, self.data[i, ...])
+
+    def __setitem__(self, i: int, v: Any):
```
tehrengruber, line 159:
```suggestion
    def __setitem__(self, i: int, v: Any) -> None:
```

src/functional/iterator/embedded.py
```diff
@@ -143,8 +142,33 @@ def __array__(self) -> np.ndarray:
...
+    def __init__(self, kstart: int, data: np.ndarray):
+        self.kstart = kstart
+        self.data = data
+
+    def __getitem__(self, i: int):
+        return self.data[i - self.kstart]
+
+    def tuple_get(self, i: int):
+        if self.data.dtype.names is not None:
+            return Column(self.kstart, self.data[f"f{i}"])
```
tehrengruber, line 155:
```suggestion
        if self.data.dtype.names:
            return Column(self.kstart, self.data[self.data.dtype.names[i]])
```
Seen this cleaner version in one of Felix's PRs.

src/functional/iterator/embedded.py
```diff
@@ -143,8 +142,33 @@ def __array__(self) -> np.ndarray:
...
+    def __array__(self, dtype=None) -> np.ndarray:
+        return self.data if dtype is None or dtype == self.data.dtype else self.data.astype(dtype)
+
+    def __array_function__(self, func, types, args, kwargs):
+        assert all(issubclass(t, self.__class__) for t in types)
+        return self.__class__(self.kstart, func(*(arg.data for arg in args), **kwargs))
+
+
+def _make_column(column_range: range, dtype=np.dtype):
+    return Column(min(column_range), np.zeros(len(column_range), dtype=dtype))
```
tehrengruber, line 171:
```suggestion
    return Column(column_range.start, np.zeros(len(column_range), dtype=dtype))
```
Have you checked that min does not iterate though the entire column? In that case this is just a suggestion. You could also use `column_range[0]`

src/functional/iterator/embedded.py
```diff
@@ -534,7 +558,11 @@ def _make_tuple(
     if isinstance(field_or_tuple, tuple):
         return tuple(_make_tuple(f, indices) for f in field_or_tuple)
     else:
-        return field_or_tuple[indices]
+        data = field_or_tuple[indices]
+        if isinstance(data, np.ndarray):
+            return Column(0, data)
```
tehrengruber, line 563: This part is not clear to me. Can you give an example when this happens? I think a comment would be beneficial.

tehrengruber (changes requested):
Looks good. Only minor things.

src/functional/iterator/embedded.py
havogt: Thanks for pointing that out. The implementation is not very nice, I wanted to think of a cleaner solution.

This is the case where no scan is involved but the fencil is executed in columns. Here we are packing the column slice of an input field into a `Column`. It should not be hidden in the tuple creation. Additionally, honestly I am not 100% sure if the 0 is correct here. I'll have to add a test.

src/functional/iterator/embedded.py
```diff
@@ -529,12 +553,22 @@ def _get_axes(
...
     if isinstance(field_or_tuple, tuple):
         return tuple(_make_tuple(f, indices) for f in field_or_tuple)
     else:
-        return field_or_tuple[indices]
+        data = field_or_tuple[indices]
+        if as_column:
+            # wraps a vertical slice of an input field into a `Column`
+            return Column(
+                0, data
+            )  # TODO(havogt) when we support LocatedFields with origin (i.e. non-zero origin), kstart needs to be adapated here
```
havogt, line 569: I tried to improve readability of this part by adding the `as_column` parameter and added comments. Also added 2 TODOs, but I wouldn't like to address them here.

src/functional/iterator/embedded.py
havogt: see comment below

egparedes (commented):
The new column handling implementation looks fine but I have some comments to improve interoperability of `Column` instances with numpy.

src/functional/iterator/embedded.py
```diff
@@ -143,8 +142,33 @@ def __array__(self) -> np.ndarray:
...
+        if self.data.dtype.names:
+            return Column(self.kstart, self.data[self.data.dtype.names[i]])
+        else:
+            return Column(self.kstart, self.data[i, ...])
+
+    def __setitem__(self, i: int, v: Any) -> None:
+        self.data[i - self.kstart] = v
+
+    def __array__(self, dtype: Optional[npt.DTypeLike] = None) -> np.ndarray:
+        return self.data if dtype or dtype == self.data.dtype else self.data.astype(dtype)
```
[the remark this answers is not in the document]
havogt: Yes, `not` was missing. I think for performance we want the view. We need the `dtype` argument because probably the mixed-in operators call it with dtype.

src/functional/iterator/embedded.py
```diff
@@ -143,8 +142,33 @@ def __array__(self) -> np.ndarray:
...
+
+    def __array_function__(self, func, types, args, kwargs):
```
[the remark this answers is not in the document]
havogt: This is for np.where etc.

egparedes (approved):
It looks good to me. Optionally, you could also add a simple test for `__array_ufunc__` and another `__array_function__` just to make sure they don't break in the future.

src/functional/iterator/embedded.py
```diff
@@ -143,8 +141,47 @@ def __array__(self) -> np.ndarray:
...
+        if self.data.dtype.names:
+            return Column(self.kstart, self.data[self.data.dtype.names[i]])
+        else:
+            return Column(self.kstart, self.data[i, ...])
+
+    def __setitem__(self, i: int, v: Any) -> None:
+        self.data[i - self.kstart] = v
+
+    def __array__(self, dtype: Optional[npt.DTypeLike] = None) -> np.ndarray:
+        return self.data if not dtype or dtype == self.data.dtype else self.data.astype(dtype)
```
havogt, line 168:
```suggestion
        return self.data.astype(dtype, copy=False)
```

PR #906, opened 2022-08-18
Functional: Fix exception chaining tests

tehrengruber:
We frequently use exception chaining in the frontend to enhance exceptions with additional information only available in "outer" stack frames. In the tests we used `__context__` to access the inner most exceptions, but since we explicitly chain we should actually use `__cause__`. This PR fixes that.


egparedes (approved):
It looks good. Maybe you could mention PEP-3134 in the commit description for reference.

PR #907, opened 2022-08-18
Ternary operator

nfarabullini:
Ternary operator implementation, e.g.

```python
@field_operator
def ternary_field_op(a: Field[[Edge], float], b: Field[[Edge], float]) -> Field[[Edge], float]:
        c = a if 2 < 3 else b
        return c
```
 
Note: true and false expression must be the same type


src/functional/ffront/field_operator_ast.py
```diff
@@ -143,6 +143,12 @@ class Compare(Expr):
     right: Expr
 
 
+class IfExp(Expr):
+    test: Expr
+    body: Expr
+    orelse: Expr
```
tehrengruber, line 149:
```suggestion
class TernaryExpr(Expr):
    condition: Expr
    true_expr: Expr
    false_expr: Expr
```
As long as the naming in Python AST makes sense, we inherit the naming. The ternary operator is a good example where I believe this is not the case. I would propose the following naming.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,59 @@ def solve_tridiag(
...
+    Edge = rs.Edge
+
+    num_edges = rs.num_edges
+    a = np_as_located_field(Edge)(np.ones((num_edges,)))
+    b = np_as_located_field(Edge)(2 * np.ones((num_edges,)))
+    out = np_as_located_field(Edge)(np.zeros((num_edges,)))
+
+    @field_operator
+    def ternary_field_op(a: Field[[Edge], float], b: Field[[Edge], float]) -> Field[[Edge], float]:
+        c = a if 2 < 3 else b
```
tehrengruber, line 826: I would suggest to switch to a condition using two scalar arguments.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,59 @@ def solve_tridiag(
...
+        return c
+
+    @program
```
tehrengruber, line 829:
```suggestion
    # TODO(tehrengruber): directly call field operator when the generated programs support `out` being a tuple
    @program
```

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,59 @@ def solve_tridiag(
...
+    ternary_field(a, b, out, offset_provider={})
+    e = np.asarray(a) if 2 < 3 else np.asarray(b)
+    np.allclose(e, out)
+
+
+def test_ternary_operator_tuple(reduction_setup):
+    rs = reduction_setup
+    Edge = rs.Edge
+
+    num_edges = rs.num_edges
```
tehrengruber, line 842:
```suggestion
    Edge = reduction_setup.Edge
    num_edges = reduction_setup.num_edges
```
You are mixing different styles here. Either use `rs = reduction_setup` and then `rs.what_you_need` for extract what you need, but skip the `rs` aliasing.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,59 @@ def solve_tridiag(
...
+    a = np_as_located_field(Edge)(np.ones((num_edges,)))
+    b = np_as_located_field(Edge)(2 * np.ones((num_edges,)))
+    out_1 = np_as_located_field(Edge)(np.zeros((num_edges,)))
+    out_2 = np_as_located_field(Edge)(np.zeros((num_edges,)))
+
+    @field_operator
+    def ternary_field_op(
+        a: Field[[Edge], float], b: Field[[Edge], float]
+    ) -> tuple[Field[[Edge], float], Field[[Edge], float]]:
+        c, d = (a, b) if 2 < 3 else (b, a)
```
tehrengruber, line 852: Same as above.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -249,6 +249,45 @@ def visit_BinOp(self, node: foast.BinOp, **kwargs) -> foast.BinOp:
             op=node.op, left=new_left, right=new_right, location=node.location, type=new_type
```
tehrengruber, line 249: As mentioned on slack, the changes in this file need some reworking.

tehrengruber (changes requested):
Except for the type deduction is looks good.

tehrengruber (changes requested):
Partial review. As discussed I also don't understand the error in the CI. Maybe I find it tomorrow.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -249,6 +249,44 @@ def visit_BinOp(self, node: foast.BinOp, **kwargs) -> foast.BinOp:
...
+
+    def _deduce_ternaryexpr_type(
+        self,
+        node: foast.TernaryExpr,
+        *,
+        condition: foast.Expr,
+        true_expr: foast.Expr,
+        false_expr: foast.Expr,
+    ) -> Optional[ct.SymbolType]:
+
```
tehrengruber, line 275:
```suggestion
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -249,6 +249,44 @@ def visit_BinOp(self, node: foast.BinOp, **kwargs) -> foast.BinOp:
...
+            raise FieldOperatorTypeDeductionError.from_foast_node(
+                condition,
+                msg=f"Condition operator is of type `{condition.type}` "
+                f"but should be of type bool.",
+            )
+
+        if true_expr.type != false_expr.type:
+            raise FieldOperatorTypeDeductionError.from_foast_node(
+                node,
+                msg=f"Left and right types are not the same: `{true_expr.type}` and {false_expr.type}",
```
tehrengruber, line 286:
```suggestion
                msg=f"Left and right types are not the same: `{true_expr.type}` and `{false_expr.type}`",
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -249,6 +249,44 @@ def visit_BinOp(self, node: foast.BinOp, **kwargs) -> foast.BinOp:
...
+        *,
+        condition: foast.Expr,
+        true_expr: foast.Expr,
+        false_expr: foast.Expr,
+    ) -> Optional[ct.SymbolType]:
+
+        if condition.type != ct.ScalarType(kind=ct.ScalarKind.BOOL):
+            raise FieldOperatorTypeDeductionError.from_foast_node(
+                condition,
+                msg=f"Condition operator is of type `{condition.type}` "
```
tehrengruber, line 279:
```suggestion
                msg=f"Condition is of type `{condition.type}` "
```

src/functional/ffront/foast_to_itir.py
```diff
@@ -320,6 +320,22 @@ def visit_BinOp(self, node: foast.BinOp, **kwargs) -> itir.FunCall:
             )
         )
 
+    def visit_TernaryExpr(self, node: foast.TernaryExpr, **kwargs) -> itir.FunCall:
+        true_value = to_value(node.true_expr)(self.visit(node.true_expr, **kwargs))
+        false_value = to_value(node.false_expr)(self.visit(node.false_expr, **kwargs))
+        cond_value = to_value(node.condition)(self.visit(node.condition, **kwargs))
+
+        if isinstance(node.true_expr.type, ct.ScalarType):
+            return im.call_("if_")(cond_value, true_value, false_value)
```
tehrengruber, line 329:
```suggestion
```
Should be handled by `_lift_if_field` or am I missing something?

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,121 @@ def solve_tridiag(
...
+
+    left = 2.0
+    right = 3.0
+
+    @field_operator
+    def ternary_field_op(
+        a: Field[[Edge], float], b: Field[[Edge], float], left: float, right: float
+    ) -> Field[[Edge], float]:
+        c = a + b if left < right else b
+        return c
```
tehrengruber, line 831:
```suggestion
        return a if left < right else b
```
or
```suggestion
        return a if left < right else a+1
```

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,121 @@ def solve_tridiag(
     solve_tridiag(a, b, c, d, out=out, offset_provider={})
 
     np.allclose(expected, out)
+
+
+def test_ternary_operator(reduction_setup):
+    Edge = reduction_setup.Edge
+    num_edges = reduction_setup.num_edges
```
tehrengruber, line 817: No need to use the much heavier `reduction_setup` fixture. Just initialize these two directly here. Same below for the non-reduction test cases.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,121 @@ def solve_tridiag(
...
+
+    left = 2.0
+    right = 3.0
+
+    @field_operator
+    def ternary_field_op(
+        a: Field[[Edge], float], b: Field[[Edge], float], left: float, right: float
+    ) -> tuple[Field[[Edge], float], Field[[Edge], float]]:
+        c, d = (a, b) if left < right else (b, a)
+        return c, d
```
tehrengruber, line 862:
```suggestion
        return (a, b) if left < right else (b, a)
```

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,121 @@ def solve_tridiag(
...
+    expected = (
+        np.sum(np.asarray(b)[rs.v2e_table], axis=1)
+        if 2 < 3
+        else np.sum(np.asarray(a)[rs.v2e_table], axis=1)
+    )
+
+    assert np.allclose(expected, out)
+
+
+@pytest.mark.parametrize("forward", [True, True])
```
tehrengruber, line 915:
```suggestion
```
Does not test anything that is not already covered by existing tests.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,121 @@ def solve_tridiag(
...
+    KDim = Dimension("K", kind=DimensionKind.VERTICAL)
+    size = 10
+    init = 1.0
+    out = np_as_located_field(KDim)(np.zeros((size,)))
+    expected = np.arange(init + 1.0, init + 1.0 + size, 1)
+
+    @scan_operator(axis=KDim, forward=forward, init=init)
+    def simple_scan_operator(carry: float) -> float:
+        c = carry if 2 > 3 else carry + 1.0
+        return c
```
tehrengruber, line 926:
```suggestion
    def simple_scan_operator(carry: float, a: float) -> float:
        return carry if carry > a else carry + 1.0
```
I suggest to change to this more involved test case.

src/functional/ffront/foast_to_itir.py
```diff
@@ -486,6 +502,13 @@ def visit_Compare(self, node: foast.Compare, **kwargs) -> itir.FunCall:
             self.visit(node.left, **kwargs), self.visit(node.right, **kwargs)
         )
 
+    def visit_IfExp(self, node: foast.TernaryExpr, **kwargs) -> itir.FunCall:
```
tehrengruber, line 505:
```suggestion
    def visit_TernaryExp(self, node: foast.TernaryExpr, **kwargs) -> itir.FunCall:
```

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,121 @@ def solve_tridiag(
...
+    num_edges = rs.num_edges
+
+    a = np_as_located_field(Edge)(np.ones((num_edges,)))
+    b = np_as_located_field(Edge)(2 * np.ones((num_edges,)))
+    out = np_as_located_field(Vertex)(np.zeros((num_vertices,)))
+
+    @field_operator
+    def ternary_reduce(a: Field[[Edge], float], b: Field[[Edge], float]) -> Field[[Vertex], float]:
+        # line below does not work
+        # f = neighbor_sum(b(V2E) if 2 < 3 else a(V2E), axis=V2EDim)
```
tehrengruber, line 900: Maybe this fails because of the `IfExpr` typo in the lowering. In any case we should fix this before this PR is merged. As a general note: If a test fails you can use `pytest.xskip` together with a comment on the reason for the failure and potential solutions such that when someone revisits the test he can pick up where you left.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -810,3 +810,121 @@ def solve_tridiag(
...
+        out = neighbor_sum(b(V2E), axis=V2EDim) if 2 < 3 else neighbor_sum(a(V2E), axis=V2EDim)
```
tehrengruber, line 901: I would skip this test as it adds little value ontop of the existing tests.

tehrengruber (dismissed):

src/functional/ffront/foast_to_itir.py
nfarabullini: not for this part, it works without (you can run the scan test to see it)

tests/functional_tests/ffront_tests/test_execution.py
nfarabullini: I added some lines in the type_deduction for this part. The issue there was with the local dimensions

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -848,6 +812,118 @@ def solve_tridiag(
...
+    right = 3.0
+
+    @field_operator
+    def ternary_field_op(
+        a: Field[[IDim], float], b: Field[[IDim], float], left: float, right: float
+    ) -> Field[[IDim], float]:
+        return a if left < right else b
+
+    ternary_field_op(a, b, left, right, out=out, offset_provider={})
+    e = np.asarray(a) + np.asarray(b) if left < right else np.asarray(b)
```
egparedes, line 832: Why is there a sum of `a` and `b` here and why is the test passing? Shouldn't it be:
```suggestion
    e = np.asarray(a) if left < right else np.asarray(b)
```
?

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -848,6 +812,118 @@ def solve_tridiag(
...
+    np.allclose(e, out)
+
+    @field_operator
+    def ternary_field_op_scalars(left: float, right: float) -> Field[[IDim], float]:
+        return broadcast(3.0, (IDim,)) if left < right else broadcast(4.0, (IDim,))
```
egparedes, line 837: Maybe try out the other branch in this test (`left > right`)...

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -848,6 +812,118 @@ def solve_tridiag(
...
+
+    ternary_field_op_scalars(left, right, out=out, offset_provider={})
+    e = np.full(e.shape, 3.0) if left < right else e
+    np.allclose(e, out)
+
+
+def test_ternary_operator_tuple():
+
```
egparedes, line 845:
```suggestion
```

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -848,6 +812,118 @@ def solve_tridiag(
...
+        left: float,
+        right: float,
+        out_1: Field[[IDim], float],
+        out_2: Field[[IDim], float],
+    ):
+        ternary_field_op(a, b, left, right, out=(out_1, out_2))
+
+    ternary_field(a, b, left, right, out_1, out_2, offset_provider={})
+
+    e, f = (np.asarray(a), np.asarray(b)) if 2 < 3 else (np.asarray(b), np.asarray(a))
```
egparedes, line 875: Why not `left < right`?
```suggestion
    e, f = (np.asarray(a), np.asarray(b)) if left < right else (np.asarray(b), np.asarray(a))
```

egparedes (commented):
The code is good to go, just some suggestions in the tests.

src/functional/ffront/foast_to_itir.py
```diff
@@ -320,6 +320,23 @@ def visit_BinOp(self, node: foast.BinOp, **kwargs) -> itir.FunCall:
...
+        lowered_true_expr = self.visit(node.true_expr, **kwargs)
+        lowered_false_expr = self.visit(node.false_expr, **kwargs)
+
+        # Needed in case of sparse or shifted fields
+        if isinstance(node.type, ct.FieldType) and is_local_kind(node.type):
+            return im.call_("if_")(lowered_node_cond, lowered_true_expr, lowered_false_expr)
+
+        return self._lift_if_field(node)(
+            im.call_("if_")(
+                to_value(node.condition)(lowered_node_cond),
```
tehrengruber, line 334:
```suggestion
                lowered_node_cond,
```
Will be a value in any case.

src/functional/ffront/foast_to_itir.py
```diff
@@ -320,6 +320,23 @@ def visit_BinOp(self, node: foast.BinOp, **kwargs) -> itir.FunCall:
...
         )
 
+    def visit_TernaryExpr(self, node: foast.TernaryExpr, **kwargs) -> itir.FunCall:
+        lowered_node_cond = self.visit(node.condition, **kwargs)
+        lowered_true_expr = self.visit(node.true_expr, **kwargs)
+        lowered_false_expr = self.visit(node.false_expr, **kwargs)
+
+        # Needed in case of sparse or shifted fields
+        if isinstance(node.type, ct.FieldType) and is_local_kind(node.type):
+            return im.call_("if_")(lowered_node_cond, lowered_true_expr, lowered_false_expr)
```
tehrengruber, line 330: This doesn't look quite right even if it might work just fine. According to the spec (https://github.com/GridTools/concepts/blob/master/Iterator-View.md) `if_` only accepts values for its second and third argument, but you pass iterators. Please add a todo for me.

tests/functional_tests/ffront_tests/test_execution.py
tehrengruber: Because left and right are static such that the right branch is always returned. We could parameterize the test.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -848,6 +812,118 @@ def solve_tridiag(
...
+    )
+
+    assert np.allclose(expected, out)
+
+
+def test_ternary_scan():
+    KDim = Dimension("K", kind=DimensionKind.VERTICAL)
+    size = 10
+    init = 1.0
+    a = np_as_located_field(KDim)(np.ones((size,)))
```
tehrengruber, line 914:
```suggestion
    init = 0.0
    a = np_as_located_field(KDim)(4*np.ones((size,)))
```
What pattern is chosen here doesn't matter that much, but the idea why I asked for this test to be adopted was to have a test where the condition changes depending on the k-level on which the scan pass, i.e. the scalar function, is executed. Result should be `0, 1, 2, 3, 4, 5, 5, 5, 5, 5`.

tehrengruber (commented):
Looks good to me so far. Just adding my two cents here, the two of you sort out the rest :-)

src/functional/ffront/foast_to_itir.py
egparedes: @tehrengruber, you're right, it might be something wrong here. Looking at the generated IR of the test with the neighbor sum, it's easy to see that everything is fine in that case because the ternary is applied **inside** the function passed to `reduce`, which will only receive values.

To me, the lowering part is actually correct, but I'm not sure about the condition in the `if`. I think the condition should check if the ternary appears inside a reduction (which effectively ensures that the field is a partially shifted field) to use the reduced lowering path (without the `to_value()` call), and if it's not inside the reduction then use the normal lowering, which will fail for partially shifted fields and therefore likely means that the user code is incorrect...

Does this make any sense?

havogt: What's the status here?

egparedes: I finally found and fixed the bug here. The problem indeed was in the lowering of the ternary expression inside a reduction. 

The weird `if` case here was only needed because the lowering should be done different inside the reduction but, since this was already taken into account in the `InsideReductionLowering` visitor, why was the weird `if` needed? It turns out there was a typo in the name of the `visit_TernaryExpr` method of `InsideReductionLowering` which meant that it was never called and thus the different lowering path inside reduction never used.

I fixed the typo, remove the unnecessary `if` and now it's good to go.

egparedes (approved):
Good to go!

PR #913, opened 2022-08-24
Fix flake8 issue

havogt:
caused by a new flake8 version.

### DIFF
--- a/tests/eve_tests/unit_tests/test_utils.py
+++ b/tests/eve_tests/unit_tests/test_utils.py
@@ -72,7 +72,7 @@ def test_register_subclasses():
         pass
 
     @eve.utils.register_subclasses(MyVirtualSubclassA, MyVirtualSubclassB)
-    class MyBaseClass(abc.ABC):
+    class MyBaseClass(abc.ABC):  # noqa: B024
         pass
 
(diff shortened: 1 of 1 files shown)


egparedes (approved):
It looks good.

tests/eve_tests/unit_tests/test_utils.py
```diff
@@ -72,7 +72,7 @@ class MyVirtualSubclassB:
         pass
 
     @eve.utils.register_subclasses(MyVirtualSubclassA, MyVirtualSubclassB)
-    class MyBaseClass(abc.ABC):
+    class MyBaseClass(abc.ABC):  # noqa: B024
```
fthaler, line 75: Wouldn’t it make more sense to either add an abstract method or remove the `abc.ABC`? That would make the test more realistic _and_ remove the requirement for `noqa`.

egparedes:
bors try

tests/eve_tests/unit_tests/test_utils.py
egparedes: We can't remove the `abc.ABC` inheritance because that is exactly what is being tested here. Adding an `abstractmethod` is ok and makes sense (although note that there are also some valid use cases for ABCs without abstractmethods).

jdahm:
bors try

PR #914, opened 2022-08-25
Allow slicing multiple fields

havogt:


tests/eve_tests/unit_tests/test_concepts.py
```diff
@@ -144,3 +144,28 @@ def test_children(self, sample_node):
...
+        node2.loc = make_source_location()
+        assert node != node2
+        assert eve.concepts.eq_nonlocated(node, node2)
+
+    def test_source_location_group(self):
+        node = make_simple_node_with_loc()
+        node2 = copy.copy(node)
+        node2.loc = make_source_location_group()
+        assert node != node2
+        assert eve.concepts.eq_nonlocated(node, node2)
```
havogt, line 171: what's the test pattern that I should use here?

src/eve/concepts.py
```diff
@@ -269,3 +269,7 @@ class VType(datamodels.FrozenModel):
 
     # Unique name
     name: str
+
+
+def eq_nonlocated(a: Node, b: Node) -> bool:
+    return utils.ddiff(a, b, ignore_type_in_groups=[SourceLocation, SourceLocationGroup])
```
havogt, line 275: probably in the wrong place. suggestions?

src/eve/concepts.py
```diff
@@ -269,3 +269,7 @@ class VType(datamodels.FrozenModel):
 
     # Unique name
     name: str
+
+
```
havogt, line 273: TODO: docstring

tests/eve_tests/unit_tests/test_concepts.py
```diff
@@ -144,3 +144,28 @@ def test_children(self, sample_node):
...
+)
+
+
+class TestEqNonlocated:
+    def test_source_location(self):
+        node = make_simple_node_with_loc()
+        node2 = copy.copy(node)
+        node2.loc = make_source_location()
+        assert node != node2
+        assert eve.concepts.eq_nonlocated(node, node2)
```
havogt, line 164: TODO test that it compares != if values are different (currently broken).

src/eve/concepts.py
egparedes: Since I guess this is only used in tests, why don't you add it as an utility there? Maybe even define it as a `nonlocated_node_comparator` fixture which returns the comparator function and can be easily used by any test function. By the way, don't you need to explicitly build the `bool` return value to make type checker happy?
```suggestion
def eq_nonlocated(a: Node, b: Node) -> bool:
    return bool(utils.ddiff(a, b, ignore_type_in_groups=[SourceLocation, SourceLocationGroup]))
```

havogt: No, not necessarily only for tests.

src/eve/trees.py
```diff
@@ -55,7 +55,7 @@
     TreeLike = Any
 else:
 
-    class TreeLike(abc.ABC):
+    class TreeLike(abc.ABC):  # noqa: B024
```
DropD, line 58: *looking at the surrounding code*
Hang on, what is happening here? Looks like there must be an interesting story behind all this.

Doesn't concern this PR though

src/functional/ffront/past_to_itir.py
```diff
@@ -153,20 +153,21 @@ def _visit_slice_bound(
...
-                raise NotImplementedError("Slicing for tuple `out` argument not supported.")
             return itir.FunCall(
                 fun=itir.SymRef(id="make_tuple"),
                 args=[self._construct_itir_out_arg(el) for el in node.elts],
             )
         else:
             raise AssertionError()(
-                "Unexpected `out` argument. Must be a `past.Name` or nesting of "
-                "`past.TupleExpr` and `past.Name`."
+                "Unexpected `out` argument. Must be a `past.Name` or `past.Subscript` or nesting in a `past.TupleExpr`."
```
DropD, line 170: Not quite sure I understand this error message. What does it mean for the `out` argument to be a "nesting in a past.TupleExpr"?

tests/eve_tests/definitions.py
```diff
@@ -295,6 +303,13 @@ def make_source_location(*, fixed: bool = False) -> SourceLocation:
     return SourceLocation(line=line, column=column, source=source)
 
 
+def make_source_location_group(*, fixed: bool = False) -> SourceLocation:
+    loc1 = make_source_location(fixed=fixed)
+    loc2 = make_source_location(fixed=fixed)
+
+    return SourceLocationGroup(loc1, loc2, context=make_str_value(fixed=fixed))
```
DropD, line 310: Is it possible you meant the following?

```suggestion
def make_source_location_group(*, fixed: bool = False) -> SourceLocationGroup:
    loc1 = make_source_location(fixed=fixed)
    loc2 = make_source_location(fixed=fixed)

    return SourceLocationGroup(loc1, loc2, context=make_str_value(fixed=fixed))
```

src/functional/ffront/past_to_itir.py
```diff
@@ -240,9 +245,20 @@ def _visit_stencil_call_out_arg(
...
+            if isinstance(first_field, past.Subscript):
+                assert all(
+                    isinstance(field, past.Subscript) for field in flattened
+                ), "Incompatible field in tuple: either all fields or no field must be sliced."
+                assert all(
+                    concepts.eq_nonlocated(first_field.slice_, field.slice_) for field in flattened
+                ), "Incompatible field in tuple: all fields must be sliced in the same way."
+                field_slice = self._compute_field_slice(first_field)
+                first_field = first_field.value
+
```
DropD, line 258: Is there a frontend check for the rule that multiple output args have to all be sliced the same way with a more domain-scientist oriented error message (with source location) already? If not, would it be too much work to add one?

src/eve/trees.py
havogt: Maybe @egparedes wants to tell the story.

src/functional/ffront/past_to_itir.py
havogt: I'll make it more verbose. It means it can be a TupleExpr which contains either Name, Subscript or another TupleExpr which contains either Name, Subscript or another TupleExpr, ... recurse

src/functional/ffront/past_to_itir.py
havogt: In #872 Till proposed to not do it, because longer term we want to remove the restriction, see https://github.com/GridTools/gt4py/pull/872#discussion_r930978998.

src/eve/trees.py
egparedes: The think the added `#noqa` is related to an update of flake8 in CI which now triggers this error. This class itself is going to be changed soon, so adding the `noqa` here it's ok for now.

src/functional/ffront/past_to_itir.py
DropD: I guess it's a choice if we want to be user friendly or eventually user friendly. I'd personally go with the former but I see temptation to go for the latter.

DropD (approved):

PR #916, opened 2022-08-25
Embedded: Allow operations between scalars and columns

havogt:
In #905 the checks were too strict in cases where the user combines scalars with iterators.

### DIFF
--- a/src/eve/trees.py
+++ b/src/eve/trees.py
@@ -55,7 +55,7 @@ if TYPE_CHECKING:
     TreeLike = Any
 else:
 
-    class TreeLike(abc.ABC):
+    class TreeLike(abc.ABC):  # noqa: B024
         ...
 
 
--- a/src/functional/iterator/embedded.py
+++ b/src/functional/iterator/embedded.py
@@ -194,16 +194,25 @@ class Column(np.lib.mixins.NDArrayOperatorsMixin):
         return self.data.astype(dtype, copy=False)
 
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs) -> Column:
+        # note:
+        # - we allow scalars to silently pass through and be handled correctly by numpy
+        # - we let numpy do the checking of compatible shapes
         assert method == "__call__"
-        assert all(inp.kstart == self.kstart for inp in inputs)
-        assert all(inp.data.shape == self.data.shape for inp in inputs)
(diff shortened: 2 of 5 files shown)


tests/eve_tests/unit_tests/test_utils.py
```diff
@@ -73,7 +73,7 @@ class MyVirtualSubclassB:
         pass
 
     @eve.utils.register_subclasses(MyVirtualSubclassA, MyVirtualSubclassB)
-    class MyBaseClass(abc.ABC):
+    class MyBaseClass(abc.ABC):  # noqa: B024
```
havogt, line 76: unrelated

src/eve/trees.py
```diff
@@ -55,7 +55,7 @@
     TreeLike = Any
 else:
 
-    class TreeLike(abc.ABC):
+    class TreeLike(abc.ABC):  # noqa: B024
```
havogt, line 58: unrelated

src/functional/iterator/embedded.py
```diff
@@ -194,16 +194,25 @@ def __array__(self, dtype: Optional[npt.DTypeLike] = None) -> np.ndarray:
...
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs) -> Column:
+        # note:
+        # - we allow scalars to silently pass through and be handled correctly by numpy
+        # - we let numpy do the checking of compatible shapes
         assert method == "__call__"
-        assert all(inp.kstart == self.kstart for inp in inputs)
-        assert all(inp.data.shape == self.data.shape for inp in inputs)
-        return self.__class__(self.kstart, ufunc(*(inp.data for inp in inputs), **kwargs))
+        if not all(inp.kstart == self.kstart for inp in inputs if isinstance(inp, Column)):
+            raise ValueError("Incompatible Column.kstart")
```
egparedes, line 202: Not even sure if it helps, but in case you want to report more information in the exception you could do something like:
```suggestion
        if (wrong_kstarts := (set(inp.kstart for inp in inputs if isinstance(inp, Column)) - {self.kstart})):
            raise ValueError("Incompatible Column.kstart: it should be '{self.kstart}' but found other values: {wrong_kstarts}")
```

egparedes (approved):
It looks good.

PR #921, opened 2022-08-30
Functional: Modification to `where`

abishekg7:
## Description

Presently, the `where` clause in the example below yields a `float64` for `one_if_positive`, instead of a `Field[[EdgeDim, KDim], float]`, like we would expect in NumPy
```python
@field_operator
def _example_stencil(
    a: Field[[EdgeDim, KDim], float],
) -> Field[[EdgeDim, KDim], float]:
    one_if_positive = where(p_vn > 0.0, 1.0, 0.0)
```
The workaround, so far, has been to broadcast one of the floats into (EdgeDim, KDim) as shown below
```python
one_if_positive = where(p_vn > 0.0, broadcast(1.0, (EdgeDim, KDim)), 0.0)
```
This PR attempts to modify the type promotion behavior of `where` such that `where(a > 0.0, b, c)`, with `a` being a FieldType and `b` & `c` are ScalarType, yields a FieldType with the same dimensions as `a`.


src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -509,6 +509,16 @@ def _visit_where(self, node: foast.Call, **kwargs) -> foast.Call:
...
+
+            if isinstance(mask_type, ct.FieldType):
+                if isinstance(return_type, ct.ScalarType):
+                    tmp_type = return_type
+                elif isinstance(return_type, ct.FieldType):
+                    tmp_type = return_type.dtype
+                return_type = type_info.promote(
+                    return_type, ct.FieldType(dims=mask_type.dims, dtype=tmp_type)
+                )
+
```
nfarabullini, line 521: ` if isinstance(return_type, ct.ScalarType) and isinstance(mask_type, ct.FieldType):
                
return_type = type_info.promote(
                    return_type, ct.FieldType(dims=mask_type.dims, dtype=return_type)
                )`

nfarabullini (changes requested):

src/functional/ffront/foast_passes/type_deduction.py
abishekg7: done. thanks!

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -509,6 +509,12 @@ def _visit_where(self, node: foast.Call, **kwargs) -> foast.Call:
 
         try:
             return_type = type_info.promote(left_type, right_type)
+
+            if isinstance(mask_type, ct.FieldType) and isinstance(return_type, ct.ScalarType):
+                return_type = type_info.promote(
+                    return_type, ct.FieldType(dims=mask_type.dims, dtype=return_type)
+                )
+
```
havogt, line 517: just
```suggestion
            return_type = type_info.promote(mask_type, left_type, right_type)
```
should probably work

abishekg7: That change results in this exception, probably as the dtype in `mask_type` is bool, and the others are float. 
```bash
   raise GTTypeError("Could not promote scalars of different dtype (not implemented).")
E   functional.common.GTTypeError: Could not promote scalars of different dtype (not implemented).
```

havogt: oh that makes sense

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -609,6 +609,23 @@ def conditional_promotion(
...
+    def conditional_promotion(
+        mask: Field[[IDim], bool],
+    ) -> Field[[IDim], float64]:
+        return where(mask, 5.0, 10.0)
+
+    conditional_promotion(mask, out=out, offset_provider={})
+
+    assert np.allclose(np.where(mask, 5.0, 10), out)
+
+
```
havogt, line 628: Lets add a test for the case with (IDim, JDIm) in the mask and (IDim) in the true/false branches. This probably does not work with the current implementation

abishekg7: sure

abishekg7: I added that test, and it failed. So I'm reverting back to my original logic for the type promotion. It seems to pass now.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -509,6 +509,16 @@ def _visit_where(self, node: foast.Call, **kwargs) -> foast.Call:
 
         try:
             return_type = type_info.promote(left_type, right_type)
+
+            if isinstance(mask_type, ct.FieldType):
+                if isinstance(return_type, ct.ScalarType):
+                    tmp_type = return_type
```
havogt, line 515:
```suggestion
                    return_dtype = return_type
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -509,6 +509,16 @@ def _visit_where(self, node: foast.Call, **kwargs) -> foast.Call:
...
+                elif isinstance(return_type, ct.FieldType):
+                    tmp_type = return_type.dtype
```
havogt, line 517:
```suggestion
                    return_dtype = return_type.dtype
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -509,6 +509,16 @@ def _visit_where(self, node: foast.Call, **kwargs) -> foast.Call:
...
+                return_type = type_info.promote(
+                    return_type, ct.FieldType(dims=mask_type.dims, dtype=tmp_type)
```
havogt, line 519:
```suggestion
                    return_type, ct.FieldType(dims=mask_type.dims, dtype=return_dtype)
```

havogt (changes requested):
Let's improve readability a bit.

I am also not sure if we should introduce something like (pseudocode) `promoted_dims(field_or_scalar)->dims` as a helper.

abishekg7:
@havogt Not sure re. helper function for `promoted_dims(field_or_scalar)->dims`. Would there be other situations requiring this you think?

src/functional/ffront/foast_passes/type_deduction.py
abishekg7: @nfarabullini had to revert your suggested change, due to an additional test failing. See Hannes' comment below.

havogt (approved):
Looks good to me

nfarabullini (approved):

PR #922, opened 2022-08-30
Embedded Column handling: fixes for stencil with vertical offset

havogt:
- [ ] Needs tests for the cases that fail prior to this PR: can be taken from mo_solve_nohydro_stencil_52 in icon4py (currently PR).
- [ ] document the magic `_tupsum` and add tests
- [ ] maybe fix test_hdiff (triggered by removing a weird try except block), maybe move to separate PR


src/functional/iterator/embedded.py
```diff
@@ -806,21 +806,27 @@ def combine_slice(s, t):
...
             assert s.step is None
-            assert s.stop is None
         else:
             assert isinstance(s, numbers.Integral)
             first = s
+            first_stop = None
         if isinstance(t, slice):
             is_slice = True
             second = 0 if t.start is None else t.start
+            second_stop = t.stop
```
havogt, line 818: TODO: probably only one of the to stops can be != None for meaningful semantics

src/functional/iterator/embedded.py
```diff
@@ -843,29 +843,23 @@ def get_ordered_indices(
...
-            assert t.step is None
-            assert t.stop is None
-        else:
-            assert isinstance(t, numbers.Integral)
-            second = t
-        start = first + second
-        return slice(start, None) if is_slice else start
+def _shift_slice(
+    slice_or_index: slice | numbers.Integral, offset: numbers.Integral
+) -> slice | numbers.Integral():
```
havogt, line 848:
```suggestion
) -> slice | numbers.Integral:
```

tests/functional_tests/iterator_tests/test_column_stencil.py
```diff
@@ -254,4 +254,82 @@ def test_kdoublesum_scan(program_processor, lift_mode, kstart, reference):
...
 
+@fundef
+def sum_shifted(inp0, inp1):
+    return deref(inp0) + deref(shift(K, 1)(inp1))
+
+
+@fendef(column_axis=KDim)
+def sum_shifted_fencil(out, inp0, inp1, k_size):
+    closure(
+        cartesian_domain(named_range(KDim, 0, k_size)),
```
tehrengruber, line 265:
```suggestion
        cartesian_domain(named_range(KDim, 1, k_size)),
```
Just to ensure the start index is taken into account correctly.

src/functional/iterator/embedded.py
```diff
@@ -230,7 +230,7 @@ def __array_function__(self, func, types, args, kwargs) -> Column:
         )
 
 
-def _make_column(column_range: range, dtype=np.dtype):
+def _make_column(column_range: range, dtype: np.dtype):
     return Column(column_range.start, np.zeros(len(column_range), dtype=dtype))
```
tehrengruber, line 234: Unrelated, but I would just remove the function and directly construct the column.

src/functional/iterator/embedded.py
```diff
@@ -720,18 +719,18 @@ def deref(self) -> Any:
                 raise IndexError("Iterator position doesn't point to valid location for its field.")
         slice_column = dict[Tag, FieldIndex]()
         if self.column_axis is not None:
-            slice_column[self.column_axis] = slice(shifted_pos[self.column_axis], None)
+            assert _column_range is not None
+            k_pos = shifted_pos[self.column_axis]
+            assert isinstance(k_pos, int)
+            slice_column[self.column_axis] = slice(k_pos, k_pos + len(_column_range))
             shifted_pos.pop(self.column_axis)
```
tehrengruber, line 726:
```suggestion
            k_pos = shifted_pos.pop(self.column_axis)
            assert isinstance(k_pos, int)
            slice_column[self.column_axis] = slice(k_pos, k_pos + len(_column_range))
```
Feel free to ignore.

src/functional/iterator/embedded.py
```diff
@@ -845,29 +845,26 @@ def get_ordered_indices(
...
-        return slice(start, None) if is_slice else start
+def _shift_slice(
+    slice_or_index: slice | numbers.Integral, offset: numbers.Integral, size: numbers.Integral
+) -> slice | numbers.Integral:
+    if isinstance(slice_or_index, slice):
+        assert slice_or_index.step is None
+        return slice(
+            0 if slice_or_index.start is None else slice_or_index.start + offset,
+            None if slice_or_index.stop is None else min(slice_or_index.stop + offset, size),
+        )
```
tehrengruber, line 856:
```suggestion
        start = slice_or_index.start or 0
        stop = slice_or_index.stop or 0
        # avoid unintentional wrap around
        assert sign(start+offset) == sign(start) and sign(stop+offset) == sign(stop)
        return slice(start + offset, stop+offset)
```
Why is the offset not respected if the start or stop are None? Additionally the `min` appears to be intended for protecting against taking a too large slice. Shouldn't this actually throw an error?

src/functional/iterator/embedded.py
```diff
@@ -845,29 +845,26 @@ def get_ordered_indices(
...
+        assert isinstance(slice_or_index, numbers.Integral)
+        return slice_or_index + offset
+
 
-    return tuple(combine_slice(*i) for i in zip(a, b))
+def _shift_slices(
+    slices_or_indices: tuple[slice | numbers.Integral],
+    offsets: tuple[numbers.Integral],
+    shape: tuple[numbers.Integral],
+) -> tuple[slice | numbers.Integral]:
```
tehrengruber, line 866:
```suggestion
    slices_or_indices: tuple[slice | numbers.Integral, ...],
    offsets: tuple[numbers.Integral, ...],
    shape: tuple[numbers.Integral, ...],
) -> tuple[slice | numbers.Integral, ...]:
```

src/functional/iterator/embedded.py
```diff
@@ -1049,7 +1052,7 @@ def __getitem__(self, indices):
 
     def __setitem__(self, indices, value):
         if not isinstance(value, tuple):
-            raise RuntimeError("Value needs to be tuple.")
+            raise RuntimeError(f"Value needs to be tuple, got {value}.")
```
tehrengruber, line 1055:
```suggestion
            raise RuntimeError(f"Value needs to be tuple, got `{value}`.")
```

tehrengruber (changes requested):

tests/functional_tests/iterator_tests/test_column_stencil.py
havogt: done

src/functional/iterator/embedded.py
havogt: Thanks for this comment. I realized how messed-up the current implementation was: the slice that is passed here has the semantics of a range relative to the origin of a field, i.e. negative indices don't mean relative to the end.
I changed the behavior of  `__getitem__`/ `__setitem__` to operate on the full buffer, i.e. not relative to the origin and introduced 2 new functions that are used internally with the semantics relative to the origin. I'll test this now in icon4py.

src/functional/iterator/embedded.py
```diff
@@ -845,29 +845,25 @@ def get_ordered_indices(
...
-            second = 0 if t.start is None else t.start
-            assert t.step is None
-            assert t.stop is None
-        else:
-            assert isinstance(t, numbers.Integral)
-            second = t
-        start = first + second
-        return slice(start, None) if is_slice else start
+def _shift_range(
+    slice_or_index: slice | numbers.Integral, offset: numbers.Integral
```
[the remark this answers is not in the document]
havogt: There is a downside to that: in the case where _shift_range is not needed (most cases), you added a short-cut of returning the original slice. If we change to range, we have to transform to slice even if there is no offset. I'd propose to keep and cleanup later when the field is implemented properly (in embedded field view).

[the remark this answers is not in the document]
tehrengruber: I just found out it is not even needed to use a slice in any of the cases:
```python
>>> np.array([1, 2, 3])[range(1, 3)]
array([2, 3])
```

havogt:
I don't see why tests are now failing and I cannot reproduce.

src/functional/iterator/embedded.py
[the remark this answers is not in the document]
havogt: done

havogt:
Ah, only partially merged the functional branch...

src/functional/iterator/embedded.py
```diff
@@ -651,25 +643,55 @@ def _get_axes(
...
+        if column_axis_idx is not None:
+            # construct a Column of tuples
+            first = tuple(
+                _make_tuple(f, _single_vertical_idx(indices, column_axis_idx, _column_range.start))
+                for f in field_or_tuple
+            )
+            col = Column(
+                _column_range.start, np.zeros(len(_column_range), dtype=_column_dtype(first))
+            )
+            col[0] = first
```
havogt, line 669: Should  getitem/setitem of Column be a tuple_get?

havogt:
There are some problems that need a closer look with Col of tuple vs tuple of Col that need a closer look . 
- Scan constructs Column of tuple, which seems reasonable.
- in some other places tuple of Column feels more natural, however can be transformed to col of tuple
- in a partially beautified iterator there are some problems: we don't have hooks into the tuple constructor, therefore intermediate results (that are not created via `make_tuple`) might be tuple of col. They can be transformed if they are in the interface of a fundef, however within a single function, they remain tuple of column, which might result in weird behavior.

src/functional/iterator/embedded.py
```diff
@@ -701,25 +699,57 @@ def _get_axes(
...
-    as_column: bool = False,
-) -> tuple | Column:  # arbitrary nesting of tuples of field values or `Column`s
+    column_axis: Optional[Tag] = None,
+) -> npt.DTypeLike | Column | tuple:  # arbitrary nesting of tuples of field values or `Column`s
     if isinstance(field_or_tuple, tuple):
-        return tuple(_make_tuple(f, indices, as_column=as_column) for f in field_or_tuple)
+        if column_axis is not None:
+            assert _column_range
+            # construct a Column of tuples
+            column_axis_idx = _axis_idx(_get_axes(field_or_tuple), column_axis)
```
havogt, line 719: TODO: add test case for passing a tuple of contant_fields

src/functional/iterator/embedded.py
```diff
@@ -701,25 +699,57 @@ def _get_axes(
...
 
 
+def _single_vertical_idx(indices, column_axis_idx, column_index):
+    transformed = tuple(
+        index if i != column_axis_idx else column_index for i, index in enumerate(indices)
+    )
+    return transformed
+
+
 def _make_tuple(
```
egparedes, line 709: This function returns different types depending on the input values in a predictable way. If it doesn't become too complex to make it work with mypy, I would try to define overloads to express this behavior:

```python
@overload
def _make_tuple(
    field_or_tuple: LocatedField,
    indices: FieldIndexOrIndices
    *,
    column_axis: Literal[None],
) -> npt.DTypeLike:
    ...

@overload
def _make_tuple(
    field_or_tuple: LocatedField,
    indices: FieldIndexOrIndices,
    *,
    column_axis: Tag,
) -> Column:
    ...

@overload
def _make_tuple(
    field_or_tuple: tuple[LocatedField | Column, ...],
    indices: FieldIndexOrIndices,
    *,
    column_axis: Literal[None],
) -> tuple[LocatedField | Column, ...]:
    ...

@overload
def _make_tuple(
    field_or_tuple: tuple[LocatedField | Column, ...],
    indices: FieldIndexOrIndices,
    *,
    column_axis: Tag,
) -> Column:
    ...


```

src/functional/iterator/embedded.py
```diff
@@ -701,25 +699,57 @@ def _get_axes(
...
     field_or_tuple: LocatedField | tuple,  # arbitrary nesting of tuples of LocatedField
```
egparedes, line 710: Why not
```suggestion
    field_or_tuple: LocatedField | tuple[LocatedField | Column, ...],  # arbitrary nesting of tuples of LocatedField
```
?

src/functional/iterator/embedded.py
```diff
@@ -701,25 +699,57 @@ def _get_axes(
...
+            return Column(_column_range.start, data)
         else:
             return data
 
 
+def _axis_idx(axes: Sequence[common.Dimension], axis: Tag) -> int:
+    for i, a in enumerate(axes):
+        if a.value == axis:
+            return i
+    raise AssertionError()
```
egparedes, line 750: Even if it's an internal error that should never happen, I would add a descriptive message just for developers reading  the code.

src/functional/iterator/embedded.py
```diff
@@ -882,34 +924,38 @@ def get_ordered_indices(
...
-            second = 0 if t.start is None else t.start
-            assert t.step is None
-            assert t.stop is None
-        else:
-            assert isinstance(t, numbers.Integral)
-            second = t
-        start = first + second
-        return slice(start, None) if is_slice else start
+def _shift_range(
+    range_or_index: range | numbers.Integral, offset: numbers.Integral
```
egparedes, line 933: Does this really typechecks now? It should not work. The whole pyramid of numeric types in the standard library is older than `Protocols` and thus it is based on ABCs, which can only be typechecked dynamically (at run-type) but not statically. For example, this piece of code breaks in mypy:
```
a: numbers.Integral = int(3)
```
There are some complicated partial workarounds, but I think the only simple and stable solution for now is to just use `int` in the type annotations.

src/functional/iterator/embedded.py
```diff
@@ -882,34 +924,38 @@ def get_ordered_indices(
...
-            is_slice = True
-            second = 0 if t.start is None else t.start
-            assert t.step is None
-            assert t.stop is None
-        else:
-            assert isinstance(t, numbers.Integral)
-            second = t
-        start = first + second
-        return slice(start, None) if is_slice else start
+def _shift_range(
```
egparedes, line 932: Again, here we could definie typing `overload`s to get finer grain type checking, since the output types depend only on the input types in a simple way.

src/functional/iterator/embedded.py
```diff
@@ -882,34 +924,38 @@ def get_ordered_indices(
...
+    if isinstance(range_or_index, range):
+        # range_or_index describes a range in the field
+        assert range_or_index.step == 1
+        return slice(range_or_index.start + offset, range_or_index.stop + offset)
+    else:
+        assert isinstance(range_or_index, numbers.Integral)
+        return range_or_index + offset
+
+
+def _range2slice(r: range | numbers.Integral):
```
egparedes, line 944: Missing type annotation. Maybe 
```suggestion
def _range2slice(r: range | int) -> slice | int:
```
?

src/functional/iterator/embedded.py
```diff
@@ -882,34 +924,38 @@ def get_ordered_indices(
...
+    if isinstance(r, range):
+        assert r.start >= 0 and r.stop >= r.start
+        return slice(r.start, r.stop)
+    return r
 
-    return tuple(combine_slice(*i) for i in zip(a, b))
+
+def _shift_ranges(
```
egparedes, line 951: Asa mentioned above, I would convert `numbers.Integral` to `int`.

egparedes (commented):
I have some comments about coding style and typing errors. It looks good in general but I'm a bit lost in the specific details of the implementation changes... (maybe @tehrengruber can provide better feedback in that regard).

I also think that `field_setitem` and `field_getitem` behaviors should be set as default `__setitem__` and `__getitem__` in `Field`-like classes, but this change will most likely break many internal (and external?) tests relying on the current behavior, so I would just add it as a TODO for the future.

src/functional/iterator/embedded.py
```diff
@@ -465,7 +461,9 @@ def not_eq(first, second):
     return first != second
 
 
-CompositeOfScalarOrField: TypeAlias = Scalar | LocatedField | tuple["CompositeOfScalarOrField", ...]
+CompositeOfScalarOrField: TypeAlias = (
+    Scalar | LocatedField | tuple
```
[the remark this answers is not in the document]
havogt: A assume at some point there was a problem, but seems fine now.

src/functional/iterator/embedded.py
```diff
@@ -701,25 +699,57 @@ def _get_axes(
     )
 
 
+def _single_vertical_idx(indices, column_axis_idx, column_index):
```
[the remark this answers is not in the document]
havogt: done

src/functional/iterator/embedded.py
```diff
@@ -147,19 +151,15 @@ def axes(self) -> tuple[common.Dimension, ...]:
         ...
 
     @abc.abstractmethod
-    def __getitem__(self, indices: FieldIndexOrIndices) -> Any:
+    def field_getitem(self, indices: FieldIndexOrIndices) -> Any:
```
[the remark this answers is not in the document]
havogt: added a TODO

src/functional/iterator/embedded.py
havogt: because it's could recursively contain tuples

src/functional/iterator/embedded.py
havogt: done

src/functional/iterator/embedded.py
havogt: replaced by IntIndex

src/functional/iterator/embedded.py
havogt: done

src/functional/iterator/embedded.py
havogt: done

src/functional/iterator/embedded.py
havogt: done

src/functional/iterator/embedded.py
egparedes: Oops, yes, you're right, there was a bug in my suggestion. I meant:
```suggestion
    field_or_tuple: LocatedField | tuple[LocatedField | tuple, ...],  # arbitrary nesting of tuples of LocatedField
```
Of course it doesn't capture the full recursive definition, only one level, but maybe it's enough as documentation hint and for the most simple errors. Anyway, it's not something critical, feel free to ignore.

src/functional/iterator/embedded.py
```diff
@@ -887,34 +966,56 @@ def get_ordered_indices(
...
+def _range2slice(r: range) -> slice:
+    ...
+
+
+@overload
+def _range2slice(r: IntIndex) -> IntIndex:
+    ...
+
+
+def _range2slice(r):
```
egparedes, line 1004: Same comment as above about adding type hints to the actual implementation of a function with overloads.

src/functional/iterator/embedded.py
```diff
@@ -887,34 +966,56 @@ def get_ordered_indices(
...
+    if isinstance(r, range):
+        assert r.start >= 0 and r.stop >= r.start
+        return slice(r.start, r.stop)
+    return r
+
+
+def _shift_ranges(
```
egparedes, line 1011: Just a comment, `shift_range` and  `shift_ranges` are very similar in the name and signature, it could be a good idea to give them different names. Additionally, the name seems a bit misleading to me, since they accept not only ranges but also other types of arguments.

egparedes (commented):
It looks good to me. Just minor questions and comments I would like to quickly double check before approving.

src/functional/iterator/embedded.py
```diff
@@ -706,25 +706,91 @@ def _get_axes(
...
+) -> npt.DTypeLike:
+    ...
+
+
+def _make_tuple(
+    field_or_tuple,
+    indices,
+    *,
+    column_axis=None,
+):
```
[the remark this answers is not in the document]
havogt: done

src/functional/iterator/embedded.py
havogt: I renamed this one to `shift_field_indices` and kept `shift_range` but to be consistent I should rename it also to `shift_field_index` which would make the names again very similar. No opinion on what's better...

havogt:
The change in the last commit is to make ConstantFields work which don't have the column_axis. There was a test added recently that triggered this.

src/functional/iterator/embedded.py
egparedes: What about something like  `shift_index` and `shift_all_indices` to make the distinction more clear?

egparedes (approved):
It looks good to me. I added my suggestion for a possible renaming of the two `shift_indices` functions but feel free to ignore.

PR #924, opened 2022-08-31
Guidelines review

nfarabullini:
Some suggestions for the Coding Guidelines text


CODING_GUIDELINES.md
```diff
@@ -3,41 +3,41 @@
...
 
 ### Docstrings
 
-We generate the API documentation automatically from the docstrings using [Sphinx][sphinx] together with some extensions like [Sphinx-autodoc][sphinx-autodoc] and [Sphinx-napoleon][sphinx-napoleon], which follow the Google Python Style Guide docstring conventions to automatically format the generated documentation. Check out the [Example Google Style Python Docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google) for a complete overview.
+We generate the API documentation automatically from the docstrings using [Sphinx][sphinx] and some extensions such as [Sphinx-autodoc][sphinx-autodoc] and [Sphinx-napoleon][sphinx-napoleon]. These follow the Google Python Style Guide docstring conventions to automatically format the generated documentation. A complete overview can be found here: [Example Google Style Python Docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google).
```
egparedes, line 33: Just a question: what is exactly wrong with the `together with` part?

CODING_GUIDELINES.md
```diff
@@ -3,41 +3,41 @@
...
+Designing a clean architecture for a software product is a difficult task and developers can easily overlook it under pressure or when they are eager to deliver exciting new features. However, choosing the quick solution instead of a clean one will likely lead to a large amount of additional work in the future (_technical debt_).
 
-To keep technical debt at acceptable levels, follow best practices when designing and implementing new features:
+To keep technical debt at acceptable levels, design and implementation of new features should follow best practices:
 
-1. Make sure all your code is covered by automatic testing to ensure its correctness. Where unit tests are impractical, use integration tests.
+1. Make sure your code is covered by automatic testing to ensure its correctness. Where unit tests are impractical, use integration tests.
```
egparedes, line 10: Would it be possible to make this point stronger? Something like:
```suggestion
1. Make sure your code is always covered by automatic testing to ensure its correctness. Where unit tests are impractical, use integration tests.
```

CODING_GUIDELINES.md
```diff
@@ -3,41 +3,41 @@
...
-4. _You ain't gonna need it_ ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)): don't design solutions for problems that might come up in the future, chances are you will never need that code. Focus on the current problems and prepare for future requirements by writing clean code.
-5. _Do not repeat yourself_ ([DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)): if you are writing down the same code pattern in several places you should extract it into a function.
-6. Use meaningful names: the purpose of an object should be clear from its name. Class names are usually nouns and function names are verbs.
+3. Do not reinvent the wheel: if someone solved your problem within the project or in a third party library, consider using their solution or extending it before writing a new component for the same purpose.
+4. _You aren't gonna need it_ ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)): do not design solutions for problems that might come up in the future, as chances are that you will never need that code. Focus on current problems and prepare for future requirements by writing clean code.
```
egparedes, line 13: I think the original phrasing is still the most common way to refer to this idiom:
```suggestion
4. _You ain't gonna need it_ ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)): do not design solutions for problems that might come up in the future, as chances are that you will never need that code. Focus on current problems and prepare for future requirements by writing clean code.
```

CODING_GUIDELINES.md
```diff
@@ -3,41 +3,41 @@
...
 
 ## Code Style
 
-We follow the [Google Python Style Guide][google-style-guide] with very few minor changes (mentioned below). Since the best way to remember something is to understand the reasons behind it, make sure you go through the style guide at least once, paying special attention to the discussions in the _Pros_, _Cons_ and _Decision_ subsections.
+We follow the [Google Python Style Guide][google-style-guide] with a few minor changes (mentioned below). Since the best way to remember something is to understand the reasons behind it, make sure you go through the style guide at least once, paying special attention to the discussions in the _Pros_, _Cons_, and _Decision_ subsections.
 
 We deviate from the [Google Python Style Guide][google-style-guide] only in the following points:
 
-- [`pylint`][pylint] is not required. We use [`flake8`][flake8] with some plugins.
+- We use [`flake8`][flake8] with some plugins insetad of [`pylint`][pylint].
```
egparedes, line 26:
```suggestion
- We use [`flake8`][flake8] with some plugins instead of [`pylint`][pylint].
```

CODING_GUIDELINES.md
```diff
@@ -3,41 +3,41 @@
...
+- According to subsection [_2.19 Power Features_](https://google.github.io/styleguide/pyguide.html#219-power-features), direct use of _power features_ (e.g. custom metaclasses, import hacks, reflection) should be avoided, but standard library classes that internally use these power features are accepted. Following the same spirit, we allow the use of power features in infrastructure code with similar functionality and scope as the Python standard library.
+- According to subsection [_3.19.12 Imports For Typing_](https://google.github.io/styleguide/pyguide.html#31912-imports-for-typing), symbols from `typing` and `collections.abc` modules used in type annotations _"can be imported directly to keep common annotations concise and match standard typing practices"_. Following the same spirit, we allow symbols to be imported directly from third-party or internal modules.
```
egparedes, line 29: Without the last constraint, the point is useless since it would allow importing symbols directly from anywhere:
```suggestion
- According to subsection [_3.19.12 Imports For Typing_](https://google.github.io/styleguide/pyguide.html#31912-imports-for-typing), symbols from `typing` and `collections.abc` modules used in type annotations _"can be imported directly to keep common annotations concise and match standard typing practices"_. Following the same spirit, we allow symbols to be imported directly from third-party or internal modules when they only contain a collection of frequently used typying definitions.
```

CODING_GUIDELINES.md
nfarabullini: I don't have strong feelings about it so we can revert that bit, I just felt that an "and" was sufficient

egparedes (approved):
It looks good to go. Thanks for your work reviewing the guidelines.

PR #926, opened 2022-09-01
Build systems: add build system using preconfigured compile_commands

DropD:
## Description

Consolidate the "Compiled Backends" work in a new `otf` (On The Fly compilation) module.

### Additions

- `otf` module
  - `CompiledbProject` build process, which can reuse CMake configurations for separate GT4Py programs if appropriate, this reduces CI time noticeably
  - `otf.workflow` - descriptive composition of translation, binding, compilation and other steps for on-the-fly compiling GT4Py programs, decouples build steps and stages and allows improvements on each in isolation.
  - `otf.build_data` - standardized way (accross build systems) of keeping track of build status and results

### Changes

- move supporting code for shared OTF steps and stages to the `otf` module, change some of the names to reflect the new code structure
- use `otf.workflow` to compose steps inside `gtfn_run` program executor
- rename `fencil_processors` module to `program_processors` (only slightly expands the amount of touched files anyway) and replace the term `fencil` with `program` or `GT4Py program` in documentation / strings / variable names.


src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+            build_data.BuildData(
+                status=build_data.BuildStatus.CONFIGURED,
+                module=pathlib.Path(compile_db[-1]["output"]),
+                entry_point_name=self.fencil_name,
+            ),
+            self.root_path,
+        )
+
+
+def _cc_cache_name(
```
petiaccja, line 181:
```suggestion
def _cc_prototype_program_name(
```

src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+    deps: tuple[source.LibraryDependency, ...], build_type: str, flags: list[str]
+) -> str:
+    fencil_name = "compile_commands_cache"
+    deps_str = "_".join(f"{dep.name}_{dep.version}" for dep in deps)
+    flags_str = "_".join(flags)
+    return "_".join([fencil_name, deps_str, build_type, flags_str]).replace(".", "_")
+
+
+def _cc_cache_module(
```
petiaccja, line 190:
```suggestion
def _cc_prototype_program_source(
```

src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+                entry_point_name=self.fencil_name,
+            ),
+            self.root_path,
+        )
+
+
+def _cc_cache_name(
+    deps: tuple[source.LibraryDependency, ...], build_type: str, flags: list[str]
+) -> str:
+    fencil_name = "compile_commands_cache"
```
petiaccja, line 184:
```suggestion
    base_name = "compile_commands_cache"
```
or
```suggestion
    program_base_name = "compile_commands_cache"
```

src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+            formatter_key="",
+            formatter_style=None,
+            file_extension="",
+            header_extension="",
+        ),
+    )
+
+
+def _cc_get_compiledb(
+    source_module: stages.ProgramSource, cache_strategy: cache.Strategy
```
petiaccja, line 211:
```suggestion
    prototype_program_source: stages.ProgramSource, cache_strategy: cache.Strategy
```

src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+        language_settings=languages.LanguageWithHeaderFilesSettings(
+            formatter_key="",
+            formatter_style=None,
+            file_extension="",
+            header_extension="",
+        ),
+    )
+
+
+def _cc_get_compiledb(
```
petiaccja, line 210:
```suggestion
def _cc_find_compiledb(
```

src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+    cache_path = cache.get_cache_folder(
+        stages.CompilableSource(source_module, None), cache_strategy
+    )
+    compile_db_path = cache_path / "compile_commands.json"
+    if compile_db_path.exists():
+        return compile_db_path
+    return None
+
+
+def _cc_generate_compiledb(
```
petiaccja, line 222: IMHO *find/create* are more explicit about what they do than *get/generate*, but *find/generate* is also fine.
```suggestion
def _cc_create_compiledb(
```

src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+    source_module: stages.ProgramSource,
+    build_type: str,
+    cmake_flags: list[str],
+    cache_strategy: cache.Strategy,
+) -> pathlib.Path:
+    name = source_module.entry_point.name
+    cache_path = cache.get_cache_folder(
+        stages.CompilableSource(source_module, None), cache_strategy
+    )
```
petiaccja, line 231: I find this a bit of an abuse of the cache system. You create a dummy program source with barely any data just so that you can use the existing cache system to acquire a folder. The cache folder for the compile DB prototypes don't need to follow the same scheme, they merely have to be in the same session or persistent cache root folder. You could directly use the barely any data to form the subfolders instead of abusing the existing system. This would require duplicating or opening up parts of the existing cache system, which also has its drawbacks.

Overall, I think it's fine if we keep it like this, because the other method might not be cleaner. I added suggestions to make this behaviour explicit by renaming related function to advertise that they are creating only *prototype* programs.

src/functional/otf/compilation/build_systems/compiledb.py
petiaccja, line 181: The dependencies are already included in the cache system. The build type and the build flags aren't, but they should be. That would mean that this function does not need to exist, because it does not make the cache folder any more unique than it already is.

Could you perhaps add the build type and build flags into the existing cache system, and delete this function?

src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+
+    otf_builder = cmake.CMakeProject(
```
petiaccja, line 233:
```suggestion
    prototype_project = cmake.CMakeProject(
```

src/functional/otf/compilation/compiler.py
```diff
@@ -0,0 +1,84 @@
...
+from typing import Protocol, TypeVar
+
+from functional.otf import languages, stages, step_types, workflow
+from functional.otf.compilation import build_data, cache, importer
+from functional.otf.step_types import LS, SrcL, TgtL
+
+
+SL = TypeVar("SL", bound=languages.LanguageTag)
+ST = TypeVar("ST", bound=languages.LanguageSettings)
+NT = TypeVar("NT")
```
petiaccja, line 27: Can you give these some more meaningful names?

src/functional/otf/compilation/build_data.py
```diff
@@ -0,0 +1,79 @@
...
+class BuildData:
+    """
+    Tracks build status and results of GT4Py program compilations.
+
+    Build systems projects should keep this information up to date for each build,
+    stored to a json file in the build folder at all times.
+    """
+
+    status: BuildStatus
+    module: pathlib.Path
```
petiaccja, line 38:
```suggestion
    binary: pathlib.Path
```

src/functional/otf/compilation/compiler.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def is_compiled(data: build_data.BuildData) -> bool:
+    return data.status >= build_data.BuildStatus.COMPILED
+
+
+def module_exists(data: build_data.BuildData, src_dir: pathlib.Path) -> bool:
```
petiaccja, line 34:
```suggestion
def binary_exists(data: build_data.BuildData, src_dir: pathlib.Path) -> bool:
```

petiaccja (changes requested):
A few notes:
- Build data: can you not extract the information directly from the build folder? Why add an extra JSON?
- License headers missing for empty __init__.py files: I asked @egparedes if that's needed, will let you know
- We should probably turn the CMake build type into an enum instead of throwing around string like "Debug" and "Release"

Also added a bunch of comments, but there is more to come.

src/functional/ffront/decorator.py
```diff
@@ -58,6 +52,8 @@
 from functional.ffront.source_utils import SourceDefinition, get_closure_vars_from_function
 from functional.iterator import ir as itir
 from functional.iterator.embedded import constant_field
+from functional.program_processors import processor_interface as fpi
```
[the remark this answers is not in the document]
DropD: It is now standard practice in GT4Py to not directly import module level members (as per google style guide) with very few exceptions. `import as` syntax is allowed though for long names like `processor_interface`.

src/functional/otf/binding/pybind.py
```diff
@@ -203,36 +202,45 @@ def create_bindings(
...
+    return stages.BindingSource(
         src,
-        (source_modules.LibraryDependency("pybind11", "2.9.2"),),
+        (source.LibraryDependency("pybind11", "2.9.2"),),
     )
+
+
+@workflow.make_step
+def bind_source(
+    inp: stages.ProgramSource[languages.Cpp, languages.LanguageWithHeaderFilesSettings],
```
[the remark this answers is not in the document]
DropD: Then it would not implement the `otf.workflow.StepProtocol` anymore.

src/functional/otf/stages.py
```diff
@@ -0,0 +1,116 @@
...
+    This can also optionally be added to compile bindings for other languages than python
+    when using GT4Py as part of the build for a non-python driver project.
+    """
+
+    source_code: str
+    library_deps: tuple[source.LibraryDependency, ...]
+
+
+@dataclasses.dataclass(frozen=True)
+class CompilableSource(Generic[SrcL, SettingT, TgtL]):
```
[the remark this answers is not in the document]
DropD: I would put this on the list of later fine tunings. For now the only source code is from `GTFN` and is indeed not compilable (header only, nothing to compile).

src/functional/otf/compilation/build_data.py
```diff
@@ -0,0 +1,79 @@
...
+import dataclasses
+import enum
+import json
+import pathlib
+from typing import Final, Optional
+
+
+_DATAFILE_NAME: Final = "gt4py.json"
+
+BuildStatus = enum.IntEnum("BuildStatus", ["STARTED", "CONFIGURED", "COMPILED"], start=0)
```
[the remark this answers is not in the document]
DropD: That would be an improvement, yes.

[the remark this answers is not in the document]
DropD: Why not indeed?

src/functional/otf/compilation/build_data.py
```diff
@@ -0,0 +1,79 @@
...
+    if contains_data(path):
+        return BuildData.from_json(json.loads((path / _DATAFILE_NAME).read_text()))
+    return None
+
+
+def write_data(data: BuildData, path: pathlib.Path) -> None:
+    (path / _DATAFILE_NAME).write_text(json.dumps(data.to_json()))
+
+
+def update_status(new_status: BuildStatus, path: pathlib.Path) -> None:
```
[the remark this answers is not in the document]
DropD: Given that usage will always look like `build_data.<function_name>`, it feels redundant to put `build_data` into the function name. I would take it the other way and go `build_data.exists`, `build_data.read/load`, `build_data.write/dump` etc. What do you think?

src/functional/otf/compilation/build_data.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+
+def contains_data(path: pathlib.Path) -> bool:
+    return (path / _DATAFILE_NAME).exists()
+
+
+def read_data(path) -> Optional[BuildData]:
+    if contains_data(path):
+        return BuildData.from_json(json.loads((path / _DATAFILE_NAME).read_text()))
```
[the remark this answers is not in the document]
DropD: In which case an error will be raised, which is perfectly fine, given that something really, really weird just happened.

DropD:
> * Build data: can you not extract the information directly from the build folder? Why add an extra JSON?

How do you extract the build status from a folder in which an unknown build system has done it's thing? You don't even know whether to expect a `build` subfolder. Alternatively, each build system class would have to implement it's own heuristics. Same goes for the location of the compilation result, it is an implementation detail of the build system class.The fact that both currently implemented build systems tend to create a `build` folder and place the resulting module file in `build/bin` is purely accidental and should not be relied upon.

Going that route also means that when starting cold and reusing permanent cached builds, you still have to recreate the build system class instance (going through translation, binding etc) to access these heuristics. Meaning you actually have to keep track of what build system was used last time, or have yet another level of heuristics to guess. 

--> This should probably be in the OTF ADR.

> * We should probably turn the CMake build type into an enum instead of throwing around string like "Debug" and "Release"

I'm fine with that. Do you think it is pressing enough to include here?

src/functional/otf/compilation/compiler.py
DropD: Do you really want to read those meaningful names in the type hints?

src/functional/otf/compilation/build_systems/compiledb.py
DropD: That would include updating the existing cache system to work on `CompilableSource` level, because this function also takes the binding's dependencies into account. In addition it is unclear how to pass build system configuration to the cache system, because the cache system is invoked before the build system. Furthermore this would require changes to allow to build system agnostically find and return the cached program in the compilation step.

This is not to say this work shouldn't be done, only to ask should it be done here?

src/functional/otf/compilation/build_data.py
DropD: What is the rationale behind replacing the description "this is a python module and can be imported" with the less specific description "this is a file in binary format"?

src/functional/otf/compilation/build_data.py
```diff
@@ -0,0 +1,79 @@
...
+    Build systems projects should keep this information up to date for each build,
+    stored to a json file in the build folder at all times.
+    """
+
+    status: BuildStatus
+    module: pathlib.Path
+    entry_point_name: str
+
+    def to_json(self) -> dict[str, str]:
+        return dataclasses.asdict(self) | {"status": self.status.name, "module": str(self.module)}
```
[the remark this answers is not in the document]
DropD: The idea is to prevent having to write out all the current (and future) string members.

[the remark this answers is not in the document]
DropD: Basically short hand for

```python
    def to_json(self) -> dict[str, str]:
        json_dict: dict[str, str] = {}
        for key, value in dataclasses.asdict(self).items():
            if converter := getattr(self, f"_{key}_to_json", None):
                json_dict[key] = converter(value)
            else:
                try:
                    json_dict[key] = json.dumps(value)
                except TypeError:
                    json_dict[key] = str(value)
        return json_dict

    @classmethod
    def _status_to_json(cls, status: BuildStatus) -> str:
        return status.name
```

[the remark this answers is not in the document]
DropD: Which would be overkill IMO.

src/functional/otf/compilation/build_data.py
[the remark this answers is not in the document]
DropD: When it comes to `load / dump`, I am thinking that the JSON format is an implementation detail, which does not need to bleed into function names. That said `load / dump` is fairly common for serialization formats.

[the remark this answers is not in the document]
DropD: Instead of `BuildData`, maybe you like `BuildInfo` more?

src/functional/otf/compilation/build_data.py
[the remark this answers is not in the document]
DropD: Changed it to handle each attribute manually. It should be obvious later when it becomes time to handle this more automatically.

src/functional/otf/compilation/build_data.py
```diff
@@ -0,0 +1,79 @@
+# GT4Py Project - GridTools Framework
+#
+# Copyright (c) 2014-2021, ETH Zurich
```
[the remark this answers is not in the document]
DropD: Should this be changed globally?

src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+from typing import Optional
+
+from functional.otf import languages, stages
+from functional.otf.compilation import build_data, cache, compiler
+from functional.otf.compilation.build_systems import cmake, cmake_lists
+from functional.otf.source import source
+
+
+@dataclasses.dataclass
+class CompiledbFactory(
```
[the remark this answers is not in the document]
DropD: The question is: do we want to tread "compiledb" as one word, a "label" for the concept or as a short form of "compile data base". I argue it doesn't matter in the slightest and therefore chose the one that's easier to read and type.

src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+):
+    """
+    Create a CompiledbProject from a ``CompilableSource`` stage object with given CMake settings.
+
+    Use CMake to generate a compiledb with the required sequence of build commands.
+    Generate a compiledb only if there isn't one for the given combination of cmake configuration and library dependencies.
+    """
+
+    cmake_build_type: str = "Debug"
+    cmake_extra_flags: Optional[list[str]] = None
```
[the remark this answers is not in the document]
DropD: Yes, however I will implement a version without the mutable default values trap.

src/functional/otf/compilation/build_systems/compiledb.py
```diff
@@ -0,0 +1,293 @@
...
+            languages.Cpp,
+            languages.LanguageWithHeaderFilesSettings,
+            languages.Python,
+        ],
+        cache_strategy: cache.Strategy,
+    ) -> CompiledbProject:
+        if not source.binding_source:
+            raise compiler.CompilerError(
+                "CMake build system project requires separate bindings code file."
+            )
```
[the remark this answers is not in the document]
DropD: If that's so, I'm happy to remove it.

src/functional/otf/compilation/build_systems/cmake.py
```diff
@@ -0,0 +1,145 @@
...
+import subprocess
+from typing import Optional
+
+from functional.otf import languages, stages
+from functional.otf.compilation import build_data, cache, common, compiler
+from functional.otf.compilation.build_systems import cmake_lists
+
+
+@dataclasses.dataclass
+class CMakeFactory(
```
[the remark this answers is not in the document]
DropD: I had them as functions. It was incredibly ugly and much harder to read.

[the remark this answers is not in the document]
DropD: To clarify: I had them as functions, which returned functions. The compiling step would then call the inner function, so that it doesn't have to know about build system specific config. Hence it was an uglier and harder to read version of this and the config was still stored in the inner function's closure. Implementing this as one function would couple the compiling step to each and every build system.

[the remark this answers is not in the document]
DropD: Also conceptually you would then have a build system builder configurator (builder builder?, maker builder?), which makes for really ugly names.

src/functional/otf/compilation/build_systems/cmake.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+
+    root_path: pathlib.Path
+    source_files: dict[str, str]
+    fencil_name: str
+    generator_name: str = "Ninja"
+    build_type: str = "Debug"
+    extra_cmake_flags: list[str] = dataclasses.field(default_factory=list)
+
+    def build(self):
```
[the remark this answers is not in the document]
DropD: Only the method `.build()` is part of the `BuildSystemProject` protocol. The rest are implementation details. This should be extremely obvious from the protocol itself.

src/functional/otf/compilation/build_systems/compiledb.py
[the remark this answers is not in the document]
DropD: I had simply taken this over from the `CMakeFactory`, which does explicitly access the bindings and therefore needs this. So does this code. To avoid checking for existence of bindings and implementing code without a use case I will leave this as it is. Whenever there is a usecase it will be necessary to adapt this code and remove the check. At that time there will also be something to test it against.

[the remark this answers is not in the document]
DropD: What is true is that the reason for the check here is that the `Compiledb` build system requires it. For now. Therefore the message should be adapted.

src/functional/otf/compilation/build_systems/compiledb.py
DropD: As you noted, I was trying to stick to existing APIs where possible given the width of this PR. The caching API has only one public function.

src/functional/otf/compilation/compiler.py
```diff
@@ -0,0 +1,91 @@
...
+
+        if not data or not is_compiled(data) or self.force_recompile:
+            self.builder_factory(inp, self.cache_strategy).build()
+
+        new_data = build_data.read_data(src_dir)
+
+        if not new_data or not is_compiled(new_data) or not module_exists(new_data, src_dir):
+            raise CompilerError(
+                "On-the-fly compilation unsuccessful for {inp.source_module.entry_point.name}!"
+            )
```
petiaccja, line 74: Can you include the STDOUT & STDERR from the build system?

src/functional/otf/languages.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class LanguageSettings:
+    """
+    Basic settings for any language.
+
+    Formatting will happen through ``eve.codegen.format_source``.
+    For available formatting styles check, which formatter is used (depends on ``.formatter_key``)
+    and then check which styles are available for that (if any).
```
petiaccja, line 41: Great, I missed this last time!

src/functional/otf/languages.py
```diff
@@ -0,0 +1,63 @@
...
+
+class LanguageTag:
+    """
+    Represent a programming language.
+
+    ``.settings_level`` should be set to the ``LanguageSettings`` subclass
+    with the minimum amount of settings required for the language.
+    """
+
+    settings_level: ClassVar[type[LanguageSettings]]
```
petiaccja, line 30: What does *level* stand for? The name is not clear to me

src/functional/otf/source/source.py
```diff
@@ -0,0 +1,50 @@
+# GT4Py Project - GridTools Framework
+#
+# Copyright (c) 2014-2021, ETH Zurich
```
petiaccja, line 3:
```suggestion
# Copyright (c) 2014-2022, ETH Zurich
```

src/functional/otf/stages.py
```diff
@@ -0,0 +1,116 @@
...
+SrcL = TypeVar("SrcL", bound=languages.LanguageTag)
+TgtL = TypeVar("TgtL", bound=languages.LanguageTag)
+SettingT = TypeVar("SettingT", bound=languages.LanguageSettings)
+SrcL_co = TypeVar("SrcL_co", bound=languages.LanguageTag, covariant=True)
+TgtL_co = TypeVar("TgtL_co", bound=languages.LanguageTag, covariant=True)
+SettingT_co = TypeVar("SettingT_co", bound=languages.LanguageSettings, covariant=True)
+
+
+@dataclasses.dataclass(frozen=True)
+class ProgramCall:
```
petiaccja, line 33: I'm wondering if closure is not a more appropriate name. I just brought this up though, I'm happy with `ProgramCall` too.
```suggestion
class ProgramClosure:
```

src/functional/otf/stages.py
```diff
@@ -0,0 +1,116 @@
...
+
+    program: itir.FencilDefinition
+    args: tuple[Any, ...]
+    kwargs: dict[str, Any]
+
+
+@dataclasses.dataclass(frozen=True)
+class ProgramSource(Generic[SrcL, SettingT]):
+    """
+    Standalone sourcecode translated from an IR representation along with OTF relevant information.
```
petiaccja, line 44: Probably best to avoid nested multi-part adjectives like *(on-the-fly)-compilation-relevant*:
```suggestion
    Standalone source code translated from an IR along with information relevant for OTF compilation.
```

src/functional/otf/stages.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+
+    program_source: ProgramSource[SrcL, SettingT]
+    binding_source: Optional[BindingSource[SrcL, TgtL]]
+
+    @property
+    def library_deps(self) -> tuple[source.LibraryDependency, ...]:
+        if not self.binding_source:
+            return self.program_source.library_deps
+        return tuple((*self.program_source.library_deps, *self.binding_source.library_deps))
```
petiaccja, line 97: Nice that you added this property. Can you remove duplicate elements from the concatenated list?

src/functional/otf/workflow.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __call__(self, inp: StartT) -> EndT:
+        return self.step(inp)
+
+    def chain(self, step: StepProtocol[EndT, NewEndT]) -> Workflow[StartT, EndT, NewEndT]:
+        return Workflow(first=self.step, second=step)
+
+
+@dataclasses.dataclass(frozen=True)
+class Workflow(Generic[StartT, IntermediateT, EndT]):
```
petiaccja, line 92: As I mentioned in person, you could rename this to `CombinedStep` or something to emphasize the use of `Step.chain` like you recommended as opposed to manually using `Workflow`.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -0,0 +1,94 @@
...
+def get_param_description(name: str, obj: Any) -> source.ScalarParameter | source.BufferParameter:
+    view: np.ndarray = np.asarray(obj)
+    if view.ndim > 0:
+        return source.BufferParameter(name, tuple(dim.value for dim in obj.axes), view.dtype)
+    else:
+        return source.ScalarParameter(name, view.dtype)
+
+
+@dataclasses.dataclass(frozen=True)
+class GTFNSourceGenerator(
```
petiaccja, line 38: I would like to see *translation* and/or *step* in the name due to the decision to call the ITIR->src step *translation*:
```suggestion
class GTFNTranslationStep(
```

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -0,0 +1,94 @@
...
+    step_types.TranslationStep[languages.Cpp, languages.LanguageWithHeaderFilesSettings],
+):
+    language_settings: languages.LanguageWithHeaderFilesSettings = cpp_gen.CPP_DEFAULT
+
+    def __call__(
+        self,
+        inp: stages.ProgramCall,
```
petiaccja, line 45:
```suggestion
        program_call: stages.ProgramCall,
```
I see you consistently call these `inp`. Don't you think it's better to have a more descriptive name?

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -0,0 +1,94 @@
...
+    view: np.ndarray = np.asarray(obj)
+    if view.ndim > 0:
+        return source.BufferParameter(name, tuple(dim.value for dim in obj.axes), view.dtype)
+    else:
+        return source.ScalarParameter(name, view.dtype)
+
+
+@dataclasses.dataclass(frozen=True)
+class GTFNSourceGenerator(
+    step_types.TranslationStep[languages.Cpp, languages.LanguageWithHeaderFilesSettings],
```
petiaccja, line 39: Is it possible to derive these steps from the `Step` in `workflow.py`? It's a bit ugly that `chain` is reimplemented again and again.

petiaccja (changes requested):
Another round, more to come.

src/functional/otf/compilation/compiler.py
```diff
@@ -0,0 +1,91 @@
...
+        self, step: workflow.StepProtocol[stages.CompiledProgram, T]
+    ) -> workflow.Workflow[
+        stages.CompilableSource[SourceLanguageType, LanguageSettingsType, languages.Python],
+        stages.CompiledProgram,
+        T,
+    ]:
+        return workflow.Workflow(first=self, second=step)
+
+
+class CompilerError(Exception):
```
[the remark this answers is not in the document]
DropD: Assuming it does not match a builtin exception (there is only a handful, so I encourage you to look at them and check if you find a matching one, I didn't): You are basically saying it should be called `CompilationError` because that tells the user the compilation didn't work, which is more helpful than to know that something in the compiler component went wrong.

src/functional/otf/compilation/compiler.py
DropD: Same question applies, why replace the description of the role of the file with the description of it's format?

src/functional/otf/compilation/compiler.py
```diff
@@ -0,0 +1,91 @@
...
+@dataclasses.dataclass(frozen=True)
+class Compiler(
+    step_types.CompilationStep[SourceLanguageType, LanguageSettingsType, languages.Python]
+):
+    cache_strategy: cache.Strategy
+    builder_factory: BuildSystemProjectGenerator[
+        SourceLanguageType, LanguageSettingsType, languages.Python
+    ]
+    force_recompile: bool = False
+    """Use any build system (via configured factory) to compile a GT4Py program to a ``functional.otf.stages.CompiledProgram``."""
```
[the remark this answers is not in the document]
DropD: I agree. What's interesting is that flake8 doesn't seem to care.

src/functional/otf/compilation/compiler.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def __call__(
+        self,
+        inp: stages.CompilableSource[SourceLanguageType, LanguageSettingsType, languages.Python],
```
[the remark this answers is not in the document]
DropD: Would violate the `workflow.StepProtocol`

src/functional/otf/compilation/compiler.py
DropD: You mean trying to read the logs and handling all possible cases of logs not being written / corrupted in the error handling?

I would prefer a message on how to proceed: set the cache path manually, run again and then look at the logs.

src/functional/otf/languages.py
DropD: The mental model was level of detail. The LanguageSettings are currently implemented as a hierarchy with the most general one at the top and successively more detailed ones at the bottom. Each language has a minimum level of detail it needs.

As always, open for better suggestions if (better = supported by a more applicable mental model).

src/functional/otf/stages.py
DropD: I was wondering the same but was persuaded by https://en.wikipedia.org/wiki/Closure_(computer_programming) that closure is not the correct name. According to that a closure "Stores a function with an *environmnent*" and "environment" is defined as a mapping of values that were in scope at function definition to "free variables" inside the function. That explicitly excludes function parameters as far as I can tell.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
DropD: Would violate `workflow.StepProtocol`

src/functional/otf/workflow.py
DropD: Sure, but just to check first: In a few places I have placed `Final[Workflow[...]]` annotations to make it obvious that a variable is a Workflow going from this stage to that. Are you willing to make the readability sacrifice that comes with changing these annotations to `Final[CombinedStep[...]]`?

tests/functional_tests/otf_tests/test_workflow.py
```diff
@@ -0,0 +1,58 @@
+# GT4Py Project - GridTools Framework
+#
+# Copyright (c) 2014-2021, ETH Zurich
```
petiaccja, line 3:
```suggestion
# Copyright (c) 2014-2022, ETH Zurich
```

tests/functional_tests/program_processor_tests/test_processor_interface.py
```diff
@@ -0,0 +1,48 @@
+# GT4Py Project - GridTools Framework
+#
+# Copyright (c) 2014-2021, ETH Zurich
```
petiaccja, line 3:
```suggestion
# Copyright (c) 2014-2022, ETH Zurich
```

petiaccja (changes requested):
Few more comments.
As for the entire PR, can you please add license headers to all empty `__init__` files as well? Enrique said we'd better add it now in case the file gets populated later.

src/functional/ffront/decorator.py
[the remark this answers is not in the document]
petiaccja: Yeah, sounds good. Nevertheless, does "f" in "fpi" stand for "fencil" or for "functional"?

src/functional/otf/binding/pybind.py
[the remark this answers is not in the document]
petiaccja: Sorry, I meant:
```suggestion
    program_source: stages.ProgramSource[languages.Cpp, languages.LanguageWithHeaderFilesSettings],
```

src/functional/otf/compilation/build_data.py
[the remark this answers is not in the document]
petiaccja: Yeah, I think going the other way and using only `exists` and the like is a good idea.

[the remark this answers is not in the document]
petiaccja: I'd honestly call it `BuildStatus`, but that's already taken for the enum. `BuildInfo` may be slightly better than `BuildData`.

src/functional/otf/compilation/build_data.py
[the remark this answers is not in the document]
petiaccja: But it's still incorrect in that corner case, because it should return `None` instead of throwing an exception. It would be better like
```python
    try:
      BuildData.from_json(json.loads((path / _DATAFILE_NAME).read_text()))
    except FileNotFoundError:
      return None
```

src/functional/otf/compilation/build_data.py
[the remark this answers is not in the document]
petiaccja: > Changed it to handle each attribute manually. It should be obvious later when it becomes time to handle this more automatically.

Sounds good! Keep it simple, stupid, we'll change it when and if need comes.

src/functional/otf/compilation/build_systems/cmake.py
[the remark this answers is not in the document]
petiaccja: What I meant was:
```python
def cmake_factory(
    source: stages.CompilableSource[...],
    cache_strategy: cache.Strategy,
    cmake_generator_name: str = "Ninja"
    cmake_build_type: str = "Debug"
    cmake_extra_flags: Optional[list[str]] = None
):
  ...
```
That is, merge the parameter list of `__init__` and `__call__`, and have a single function. But now I understand that you need to separate the build-system-specific information from the program-specific information, in which case I think it's good the way it is.

src/functional/otf/compilation/build_systems/cmake.py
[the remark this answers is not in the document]
petiaccja: Fair enough. Would it be possible to make them private then?

src/functional/otf/compilation/build_systems/compiledb.py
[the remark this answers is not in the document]
petiaccja: Yeah, no big difference either way. Let's leave it as is.

src/functional/otf/compilation/build_systems/compiledb.py
petiaccja: The `Compiler` has access to the `CMakeFactory` at the moment it asks for the cache directory. Build system factories could have a method that returns the hash of the build configuration, and only the hash is passed to the cache system. What do you think? If you don't want to do it now, can you make an issue for this?

src/functional/program_processors/processor_interface.py
```diff
@@ -0,0 +1,108 @@
...
+    obj: Callable[..., OutputT], kind: type[ProcessorKindT]
+) -> TypeGuard[ProgramProcessorProtocol[OutputT, ProcessorKindT]]:
+    return callable(obj) and getattr(obj, "kind", None) is kind
+
+
+def ensure_processor_kind(
+    obj: ProgramProcessorProtocol[OutputT, ProcessorKindT], kind: type[ProcessorKindT]
+) -> None:
+    if not is_processor_kind(obj, kind):
+        raise RuntimeError(f"{obj} is not a {kind.__name__}!")
```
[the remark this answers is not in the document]
DropD: That would be equivalent to replacing `RuntimeError` with `AssertionError` in the function, except with a less helpful error message.

tests/functional_tests/otf_tests/conftest.py
```diff
@@ -53,24 +51,38 @@ def source_module_example():
...
+        binding_source=pybind.create_bindings(program_source_example),
+    )
+
+
+@pytest.fixture
+def clean_example_session_cache(compilable_source_example):
+    cache_dir = cache.get_cache_folder(compilable_source_example, cache.Strategy.SESSION)
+    if cache_dir.exists():
+        shutil.rmtree(cache_dir)
+    yield
```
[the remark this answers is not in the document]
DropD: I am not sure why this is not supposed to be compatible if used correctly. The point is to make the compiler / build system not just skip all the work when you want to test that it does the work correctly. If you want to change a program, recompile and see the changes, you first have to implement proper caching.

In other words, using this in production would not be compatible anywhere, it's a testing tool that achieves what it was written for.

petiaccja:
> > * Build data: can you not extract the information directly from the build folder? Why add an extra JSON?
> 
> How do you extract the build status from a folder in which an unknown build system has done it's thing? You don't even know whether to expect a `build` subfolder. Alternatively, each build system class would have to implement it's own heuristics. Same goes for the location of the compilation result, it is an implementation detail of the build system class.The fact that both currently implemented build systems tend to create a `build` folder and place the resulting module file in `build/bin` is purely accidental and should not be relied upon.
> 
> Going that route also means that when starting cold and reusing permanent cached builds, you still have to recreate the build system class instance (going through translation, binding etc) to access these heuristics. Meaning you actually have to keep track of what build system was used last time, or have yet another level of heuristics to guess.
> 
> --> This should probably be in the OTF ADR.

Yeah, put in the ADR, that's a good idea.

For some reason I was under the impression that a build folder is tied to the build system. I.e., you go and build the exact same program with the exact same bindings with two different build systems, they then get two different build folders. In that case you don't need heuristics to decide which build system was used, you know it. Then that build system can also figure things out from its own output, at least theoretically.

Now, you can also do it in a way that the build systems do share the build folder for the exact same input. In that case it would indeed be complicated to try to figure out which build system wrote the data previously.

Do we want build systems to share folders?

petiaccja:
> > * We should probably turn the CMake build type into an enum instead of throwing around string like "Debug" and "Release"
> 
> I'm fine with that. Do you think it is pressing enough to include here?

I don't think it's pressing, but at the same time it's also not much work, so you might as well. If you don't want to do it now, could you put a TODO perhaps?

DropD:
> > > * Build data: can you not extract the information directly from the build folder? Why add an extra JSON?
> > 
> > 
> > How do you extract the build status from a folder in which an unknown build system has done it's thing? You don't even know whether to expect a `build` subfolder. Alternatively, each build system class would have to implement it's own heuristics. Same goes for the location of the compilation result, it is an implementation detail of the build system class.The fact that both currently implemented build systems tend to create a `build` folder and place the resulting module file in `build/bin` is purely accidental and should not be relied upon.
> > Going that route also means that when starting cold and reusing permanent cached builds, you still have to recreate the build system class instance (going through translation, binding etc) to access these heuristics. Meaning you actually have to keep track of what build system was used last time, or have yet another level of heuristics to guess.
> > --> This should probably be in the OTF ADR.
> 
> Yeah, put in the ADR, that's a good idea.
> 
> For some reason I was under the impression that a build folder is tied to the build system. I.e., you go and build the exact same program with the exact same bindings with two different build systems, they then get two different build folders. In that case you don't need heuristics to decide which build system was used, you know it. Then that build system can also figure things out from its own output, at least theoretically.
> 
> Now, you can also do it in a way that the build systems do share the build folder for the exact same input. In that case it would indeed be complicated to try to figure out which build system wrote the data previously.
> 
> Do we want build systems to share folders?

More importantly, even if we don't, do we want to force implementation of the heuristics for each build system? I don't.

tests/functional_tests/otf_tests/test_pybind_build_integration.py
```diff
@@ -0,0 +1,52 @@
...
+
+from functional.otf import workflow
+from functional.otf.binding import pybind
+from functional.otf.compilation import cache, compiler
+from functional.otf.compilation.build_systems import cmake, compiledb
+
+
+def test_gtfn_cpp_with_cmake(program_source_with_name):
+    source_module_example = program_source_with_name("gtfn_cpp_with_cmake")
+    build_the_program = workflow.Workflow(
```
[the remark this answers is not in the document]
DropD: `build_step(example_program_source)` would then imply that we are building a step, which would be misleading. Even we would want to name this callable not for what it does but what it is, it should be `build_workflow` or `build_steps` because it isn't a step, it's a workflow composed of multiple steps.

tests/functional_tests/otf_tests/test_pybind_build_integration.py
```diff
@@ -0,0 +1,52 @@
...
+    otf_fencil = build_the_program(source_module_example)
+    buf = np.zeros(shape=(6, 5), dtype=np.float32)
+    sc = np.float32(3.1415926)
+    res = otf_fencil(buf, sc)
+    assert math.isclose(res, 6 * 5 * 3.1415926, rel_tol=1e-4)
+
+
+def test_gtfn_cpp_with_compiledb(program_source_with_name):
+    source_module_example = program_source_with_name("gtfn_cpp_with_compiledb")
+    build_the_program = workflow.Workflow(
```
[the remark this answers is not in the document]
DropD: `build_step(example_program_source)` would then imply that we are building a step, which would be misleading. Even we would want to name this callable not for what it does but what it is, it should be `build_workflow` or `build_steps` because it isn't a step, it's a workflow composed of multiple steps.

src/functional/otf/compilation/compiler.py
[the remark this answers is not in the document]
petiaccja: Shite, didn't know names are also part of it, but makes sense

src/functional/otf/compilation/compiler.py
petiaccja: No-no, I used to include the `cmake` subprocess' standard output and standard error here. You could do it similarly with the compile DB, only you'd include `gcc`'s std out.

src/functional/otf/languages.py
petiaccja: I see. It's okay for me then. What's your opinion about `settings_class` or `settings_type`, which makes it clear that you expect a type (though it's in the annotation) and also allows for having a non-linear structure for language settings?

DropD:
> > > * We should probably turn the CMake build type into an enum instead of throwing around string like "Debug" and "Release"
> > 
> > 
> > I'm fine with that. Do you think it is pressing enough to include here?
> 
> I don't think it's pressing, but at the same time it's also not much work, so you might as well. If you don't want to do it now, could you put a TODO perhaps?

One small concern:
https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html states
> Typical values include Debug, Release, RelWithDebInfo and MinSizeRel, but custom build types can also be defined.

By representing it with an enum, we would theoretically make this an extension bottleneck. Are we ok with that?

src/functional/otf/workflow.py
petiaccja: I'm not such a pro on these type annotations, but can you not use `StepProtocol` at those places so that it accepts any step, which `Workflow` is?

src/functional/program_processors/codegens/gtfn/gtfn_module.py
petiaccja: Understood :)

src/functional/ffront/decorator.py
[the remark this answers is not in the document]
DropD: Fencil and I have been meaning to change it.

src/functional/otf/binding/pybind.py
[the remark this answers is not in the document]
DropD: No, literally changing the name of the parameter violates the protocol. Try it if you don't believe me, change it and run mypy.

src/functional/otf/compilation/build_systems/compiledb.py
DropD: That should be broadly the way to go but I would prefer to do that when changing the cache system anyway, so I will make an issue

DropD: https://github.com/GridTools/gt4py/issues/985

src/functional/otf/compilation/build_data.py
DropD: I would leave it and add an `other_binaries` attribute to `BuildData` when there is a usecase. Currently the `.module` attribute pointing to an existing file is used to mean it can be imported and must contain a callable Python object with name `.entry_point_name`.

While in principle the build systems are general enough, supporting code would be required to make it useful. Such work would then include extending `BuildData` if appropriate.

Alternatively the current semantics could be changed but the value of that is questionable, since there is no usecase.

DropD: I clarified this in the `BuildData` docstring.

src/functional/otf/compilation/compiler.py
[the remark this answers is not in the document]
DropD: For compiler input validation it would make sense to derive from `ValueError`, as that is in line with the semantics of `ValueError`. However, not making it a custom subclas of `ValueError` is missing an opportunity to make debuggin easier. Making the error type for when the compiler fails a subclass of `ValueError` is not in line with it's semantics, that would be a `RuntimeError`. These changes are simple enough and deserve to be included.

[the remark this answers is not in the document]
DropD: There is now a `CompilerInputError`, derived from `ValueError` as well as a `CompilationError`, derived from `RuntimeError`.

src/functional/otf/compilation/compiler.py
DropD: The changes required would be:
1) change the build system interface for `.build()` to return CLI output.
2) change build system implementations to gather, log and return stdout and stderr from all commands.
3) research how to helpfully convey / truncate the stdout and stderr in the error message. This is necessary because these might well overflow the buffer in some terminals (depending on build system flags, compilers available etc).

The simple logging approach was chosen to avoid all of that, while achieving almost the same amount of debuggability, if not the same. If you run in a debugger the cache folder will anyway live until you terminate the session.

DropD: If you want I can put the above either in the ADR or in an issue.

src/functional/otf/languages.py
DropD: Fine with me.

petiaccja:
> > > > * Build data: can you not extract the information directly from the build folder? Why add an extra JSON?
> > > 
> > > 
> > > How do you extract the build status from a folder in which an unknown build system has done it's thing? You don't even know whether to expect a `build` subfolder. Alternatively, each build system class would have to implement it's own heuristics. Same goes for the location of the compilation result, it is an implementation detail of the build system class.The fact that both currently implemented build systems tend to create a `build` folder and place the resulting module file in `build/bin` is purely accidental and should not be relied upon.
> > > Going that route also means that when starting cold and reusing permanent cached builds, you still have to recreate the build system class instance (going through translation, binding etc) to access these heuristics. Meaning you actually have to keep track of what build system was used last time, or have yet another level of heuristics to guess.
> > > --> This should probably be in the OTF ADR.
> > 
> > 
> > Yeah, put in the ADR, that's a good idea.
> > For some reason I was under the impression that a build folder is tied to the build system. I.e., you go and build the exact same program with the exact same bindings with two different build systems, they then get two different build folders. In that case you don't need heuristics to decide which build system was used, you know it. Then that build system can also figure things out from its own output, at least theoretically.
> > Now, you can also do it in a way that the build systems do share the build folder for the exact same input. In that case it would indeed be complicated to try to figure out which build system wrote the data previously.
> > Do we want build systems to share folders?
> 
> More importantly, even if we don't, do we want to force implementation of the heuristics for each build system? I don't.

Well, it's not heuristics, you just need to parse the information from the build system. Anyways, I don't know how to do it with CMake right off the bat, but I wanted to mention it. I think it's perfectly fine for now to keep it like this in this PR.

src/functional/otf/workflow.py
DropD: Is the readability delta of `Workflow` vs `CombinedStep` really any different from the one of `Workflow` vs `StepProtocol`? As in, I am trying to convey that this is a *workflow*. Is telling the user that it is a StepProtocol, which is technically equivalent to a workflow going to achieve that goal?

petiaccja:
> > > > * We should probably turn the CMake build type into an enum instead of throwing around string like "Debug" and "Release"
> > > 
> > > 
> > > I'm fine with that. Do you think it is pressing enough to include here?
> > 
> > 
> > I don't think it's pressing, but at the same time it's also not much work, so you might as well. If you don't want to do it now, could you put a TODO perhaps?
> 
> One small concern: https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html states
> 
> > Typical values include Debug, Release, RelWithDebInfo and MinSizeRel, but custom build types can also be defined.
> 
> By representing it with an enum, we would theoretically make this an extension bottleneck. Are we ok with that?

Good point. In this case I don't think it makes sense to expose this to the user, and it will only be us adding a new build type, in which case we can extend the enum. I don't see us losing anything.

src/functional/otf/compilation/compiler.py
petiaccja: I find it quite handy to see the compiler messages in the unit test output. Sometimes it saves me going for the debugger and I can fix the issue right away. Can you perhaps make an issue for the general improvement of native code compilation error messages? It can be open-ended and we can discuss the approach to take outside of this PR.

src/functional/otf/compilation/build_data.py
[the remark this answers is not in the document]
DropD: It is not incorrect because a deleted build data file means there was an error in a build rather than no build has been attempted. 

In both cases below there is a point in time where the build data file does not exist, but they are not the same case.
- No build attempted -> should return None
- build started and build data got lost -> should raise Error

That said this corner case should be rare and was not a design consideration so I have nothing about the try-except version which neglects it.

src/functional/otf/compilation/build_systems/cmake.py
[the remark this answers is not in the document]
DropD: sure

src/functional/otf/compilation/build_systems/compiledb.py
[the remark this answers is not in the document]
DropD: Done

src/functional/otf/compilation/compiler.py
DropD: sure

DropD: Issue is here https://github.com/GridTools/gt4py/issues/987

src/functional/program_processors/processor_interface.py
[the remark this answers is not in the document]
petiaccja: Sorry, I meant deleting `ensure_processor_kind` altogether and using a simple assertion wherever you are using `ensure_processor_kind` right now.

tests/functional_tests/otf_tests/conftest.py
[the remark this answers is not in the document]
petiaccja: What's the use case for this cache clearing?

tests/functional_tests/otf_tests/test_pybind_build_integration.py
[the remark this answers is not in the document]
petiaccja: Yeah, it's just that the name is a bit lame. But if you don't have a better idea we can keep it, it doesn't matter much.

DropD:
> ### Sketch of the overall structure:
> * functional
>   
>   * jit/otf
>     
>     * codegens
>       
>       * gtfn
>       * pybind11_gtfn
>       * cpp_gen
>     * compilation
>     * source
>     * languages
>     * ...
>   * program_processors
>     
>     * codegens
>     * executors
>     * formatters
>   * ...

Perhaps one remark up front: It is a legitimate way to structure code by *type* of functionality rather than by keeping the components of subsystems together. However it is not currently the way GT4Py is structered by any means.

It would lead to a structure like this:
- GT4Py
  - trees
    - iterator
    - fieldview_operator_ir
    - fieldview_program_ir
    - fieldview_types
    - cmake_lists
    - cpp_snippets
    - gtfn_ir
    - ...
  - tree_visitors
    - ir_passes
      - ...
    - codegens
      - ...
    - ...
  - ...

And I believe this example is enough to motivate why it isn't that way.

petiaccja (approved):
We've discussed all concerns and comments, the appropriate changes have been made or delayed with a note in a TODO or in an issue.

PR #928, opened 2022-09-02
Functional: test indirection in iterator IR

havogt:
Tests indirection via a field with indices. This is outside of the original iterator IR specification as offsets are required to be literals, however it is supported in embedded (except, allowing different kinds of ints) and the C++ backend without modifications.


havogt:
TODO in https://github.com/GridTools/gt4py/pull/928/files#diff-cb3ad20072297c872c4fcb6245750c91cc9aa6080da28e19b1495f78e3d9b67cL664 assert that incomplete offsets is empty

PR #929, opened 2022-09-02
Where tuple return

nfarabullini:
Implementation of tuple return in where with either fields or scalars in returns, e.g.
```python
@field_operator
    def conditional_tuple_with_fields(
        mask: Field[[IDim], bool], a: Field[[IDim], float64], b: Field[[IDim], float64]
    ) -> tuple[Field[[IDim], float64], Field[[IDim], float64]]:
        return where(mask, (a, b), (b, a))
```
 
or

```python
@field_operator
    def conditional_tuple_with_fields_and_scalars(
        mask: Field[[IDim], bool], a: Field[[IDim], float64], b: Field[[IDim], float64]
    ) -> tuple[Field[[IDim], float64], Field[[IDim], float64]]:
        return where(mask, (a, b), (1.0, 2.0))
```

as well as nested tuples


src/functional/ffront/type_info.py
```diff
@@ -385,6 +385,37 @@ def promote_dims(
     return topologically_sorted_list
 
 
+def promote_to_mask_type(
```
DropD, line 388: Since the semantic meaning of this function and the terms used in it's code and docstring are quite limited to the only place it is being used, it should probably be moved to `type_deduction.py` until it turns out to be more widely useful.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -508,16 +508,31 @@ def _visit_where(self, node: foast.Call, **kwargs) -> foast.Call:
...
-
-            if isinstance(mask_type, ct.FieldType):
-                if isinstance(return_type, ct.ScalarType):
-                    return_dtype = return_type
-                elif isinstance(return_type, ct.FieldType):
-                    return_dtype = return_type.dtype
-                return_type = type_info.promote(
-                    return_type, ct.FieldType(dims=mask_type.dims, dtype=return_dtype)
-                )
+            if isinstance(left_type, ct.TupleType) and isinstance(right_type, ct.TupleType):
```
DropD, line 511: I know this touches existing code, but I would still suggest changing `left` and `right` to `true_branch` and `false_branch` or similar. It is quite a contained change but a big plus for readability. 

For operators it makes sense to use `left` and `right` because there is no semantic difference between the left and right operand but here we can call things by their meaning.

DropD (changes requested):
Currently this seems to be limited to flat tuples (no nesting). At the very least there is no test that demonstrates otherwise. Is that intentional? Similar things like tuple returns for field operators allow nested tuples.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -533,6 +541,46 @@ def _visit_where(self, node: foast.Call, **kwargs) -> foast.Call:
...
+            else:
+                self._construct_tuple_type(element.types, promoted_field_tuple, single_tuple)
+        return node_type_ls
+
+    def _promote_to_mask_type(
+        self, mask_type: ct.FieldType, input_type: ct.FieldType | ct.ScalarType
+    ) -> ct.FieldType:
+
+        return_dtype = input_type.dtype if isinstance(input_type, ct.FieldType) else input_type
+        return type_info.promote(input_type, ct.FieldType(dims=mask_type.dims, dtype=return_dtype))
```
DropD, line 582: Since `_extract_promoted_tuple_type`, `_construct_tuple_type`, `_promote_to_mask_type` deal with types and none of them makes use of the `self` instance, except to call themselves, I would suggest making them free functions. The precedent for this is the `boolified_type` function in the same file.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -533,6 +541,46 @@ def _visit_where(self, node: foast.Call, **kwargs) -> foast.Call:
...
 
+    def _extract_promoted_tuple_type(
+        self,
+        node_type_ls: list,
+        mask_type: ct.FieldType,
+        return_type: None | ct.FieldType | ct.ScalarType,
+    ) -> ct.FieldType:
+        for element in node_type_ls:
+            if isinstance(element, ct.TupleType):
+                return self._extract_promoted_tuple_type(element.types, mask_type, return_type)
```
DropD, line 552: It seems a bit unintuitive that this function should return on the first `TupleType` element, completely discarding all succeeding elements. Is that intentional?

With the current nested tuple test it doesn't seem to make a difference either way, that's why I ask.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -533,6 +541,46 @@ def _visit_where(self, node: foast.Call, **kwargs) -> foast.Call:
...
+            else:
+                if return_type is None:
+                    return_type = element
+                else:
+                    return_type = type_info.promote(return_type, element)
+                    if isinstance(return_type, ct.ScalarType) or not all(
+                        item in return_type.dims for item in mask_type.dims
+                    ):
```
DropD, line 560: If this condition could be built into `promote_to_mask_type` instead (or extracted into another function), then the whole `for` level `else` branch could be written

```python
            else:
                return_type = type_info.promote(return_type, element) if return_type else element
                return_type = self._promote_to_mask_type(mask_tye, return_type)  # or new function if preferred
```

Which significantly reduces the brain power required to follow the control flow (for me at least).

DropD (changes requested):

DropD (approved):
Looks good, thank you! I'll leave it up to you if you want to add docstrings to the type handling functions.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -498,8 +541,8 @@ def _visit_max_over(self, node: foast.Call, **kwargs) -> foast.Call:
 
     def _visit_where(self, node: foast.Call, **kwargs) -> foast.Call:
```
DropD, line 542: Actually, I just realized there is no test for any of this in `test_type_deduction`. 

The problem with this is that the types may be deduced completely wrongly here while all the tests still pass. Let's fix that before merging.

DropD (changes requested):

tests/functional_tests/ffront_tests/test_type_deduction.py
```diff
@@ -479,3 +480,69 @@ def badtype_broadcast(a: Field[[ADim], float64]):
...
+            ),
+            ct.TupleType(
+                types=[
+                    ct.FieldType(dims=[ADim], dtype=ct.ScalarType(kind=ct.ScalarKind.FLOAT64)),
+                    ct.FieldType(dims=[ADim], dtype=ct.ScalarType(kind=ct.ScalarKind.FLOAT64)),
+                ]
+            ),
+        ]
+    )
+
```
DropD, line 536: The following **crucial** test is missing:

```python
ADim = Dimension("ADim")
BDim = Dimension("BDim")

def tuple_where(a: Field[[ADim], bool], b: Field[[ADim], float64], c: Field[[ADim, BDim], float64]):
    return where(a, ((c, 9.0), (b, 6.0)), ((8.0, b), (5.0, 9.0)))
```

I am not 100 % sure what the return type should be but I would expect something like

```python
(
    (Field[[ADim, BDim], ...], Field[[ADim], ...]),
    (Field[[ADim], ...], Field[[ADim], ...]),
)
```

In any case looking at the code I don't think the question this test asks has been addressed: "What happens if I mix dimensions in a nested tuple (in a way that the two branches are compatible)?".

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -46,6 +46,88 @@ def boolified_type(symbol_type: ct.SymbolType) -> ct.ScalarType | ct.FieldType:
...
+def extract_promoted_tuple_type(
+    node_type_ls: list[ct.TupleType],
+    mask_type: ct.FieldType,
+    promoted_type: None | ct.FieldType | ct.ScalarType,
+) -> ct.FieldType:
+    """
+    Extract argument type in tuple to which all others need to be promoted to.
+
+    If arguments types in a tuple are different (e.g. some are Fields and others Scalars), they all need to be promoted to be the same.
+    This function extract which type the promotion has to be done with respect of.
```
DropD, line 58:
```suggestion
    Extract argument type in tuple to which all others need to be promoted.

    If arguments types in a tuple are different (e.g. some are Fields and others Scalars), they all need to be promoted to be the same.
    This function extracts the type with respect to which the promotion has to be done.
```

In addition to the phrasing I am questioning the described semantics. Do all tuple elements necessarily have to be the same? What happens if the tuples contain mixed dimension fields? See comment in the type deduction tests for a better description of the question.

tests/functional_tests/ffront_tests/test_type_deduction.py
nfarabullini: As the implementation is atm, the return dimensions should be [ADim, BDim] for all tuples items.

src/functional/ffront/foast_passes/type_deduction.py
nfarabullini: If you mix dimensions, all items will be promoted to a mutual type. Such type contains all dimensions from all inout fields

tests/functional_tests/ffront_tests/test_type_deduction.py
havogt: Yes, I think we should discuss this case together.

src/functional/iterator/embedded.py
```diff
@@ -843,7 +843,7 @@ def _maker(a: np.ndarray) -> LocatedFieldImpl:
             offsets = None
 
         def setter(indices, value):
-            indices = tupelize(indices)
+            indices = tupelize(indices) if len(a.shape) >= 2 or a.shape[0] <= 1 else (indices[0],)
```
havogt, line 846: I think the changes here should be rolled back, see my other comment.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -720,6 +719,127 @@ def conditional_program(
...
+    def conditional_tuple_4_p(
+        mask: Field[[IDim], bool],
+        a: Field[[IDim, JDim], float64],
+        b: Field[[IDim], float64],
+        c: Field[[IDim, JDim], float64],
+        d: Field[[IDim], float64],
+        e: Field[[IDim], float64],
+        f: Field[[IDim], float64],
+    ):
+        conditional_tuple_4_field_op(mask, a, b, out=((c, d), (e, f)))
```
havogt, line 838: The reason you had to touch embedded is not because of how where behave, but because of your test setup here. All out fields should have the same dimensions.

One option would be to take the result of the where and add all the tuple elements into a single out field if you want to keep that test. But I don't think this tests is actually testing something extra that is not already tested (maybe I am missing the point).

havogt (changes requested):
Just reviewed the changes in embedded part. Passing further review back to @DropD .

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -46,6 +46,68 @@ def boolified_type(symbol_type: ct.SymbolType) -> ct.ScalarType | ct.FieldType:
     raise GTTypeError(f"Can not boolify type {symbol_type}!")
 
 
+def construct_tuple_type(
+    element_types: list[ct.TupleType],
+    false_branch_types: list[ct.TupleType],
+    mask_type: ct.FieldType,
+) -> list:
+    """
+    Construct recursively the return types for the tuple return branch.
```
DropD, line 55:
```suggestion
    Recursively construct  the return types for the tuple return branch.
```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -46,6 +46,68 @@ def boolified_type(symbol_type: ct.SymbolType) -> ct.ScalarType | ct.FieldType:
     raise GTTypeError(f"Can not boolify type {symbol_type}!")
 
 
+def construct_tuple_type(
+    element_types: list[ct.TupleType],
```
DropD, line 50: Is there any case where `element_types` does not refer to the types in the "true-branch"? If not, would you agree that there would be some benefit in renaming?

nfarabullini: agree and edited

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -720,6 +719,92 @@ def conditional_program(
...
+    a = np_as_located_field(IDim)(np.ones((size,)))
+    b = np_as_located_field(IDim)(np.ones((size,)))
+    c = np_as_located_field(IDim)(np.zeros((size,)))
+    d = np_as_located_field(IDim)(np.zeros((size,)))
+
+    @field_operator
+    def conditional_tuple_2_field_op(
+        mask: Field[[IDim], bool], a: Field[[IDim], float64], b: Field[[IDim], float64]
+    ) -> tuple[tuple[Field[[IDim], float64], Field[[IDim], float64]], Field[[IDim], float64]]:
+        return where(mask, ((3.0, 6.0), a), ((2.0, 7.0), b))
```
DropD, line 765: shouldn't the result be checked here?

DropD (approved):
There is one test where the check of the result seems to be missing, however I consider it non-critical, since it seems to be more or less covered by other tests.

tests/functional_tests/ffront_tests/test_execution.py
havogt: Just for clarity: if the test is useful, but the result of execution is not important, the test should be a parsing test or lowering test, but not in `test_execution`.

src/functional/ffront/past_to_itir.py
```diff
@@ -242,9 +242,6 @@ def _visit_stencil_call_out_arg(
             flattened = _flatten_tuple_expr(node)
 
             first_field = flattened[0]
-            assert all(
```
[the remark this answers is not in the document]
havogt: this was a pending comment that I apparently never submitted

havogt (changes requested):

havogt (approved):

PR #931, opened 2022-09-13
Fix nested call with scalar arguments

tehrengruber:
Fixes a typo in the lowering from FOAST to ITIR where a nested call to a field operator with a scalar argument failed. The following works now:

```python
def scalar_arg_inner(scalar_inp: float64) -> Field[[Vertex], float64]:
    return broadcast(scalar_inp + 1.0, (Vertex,))

def scalar_arg(scalar_inp: float64) -> Field[[Vertex], float64]:
    return scalar_arg_inner(scalar_inp + 1.0)

scalar_arg(inp, out=out, offset_provider={})
```

### DIFF
--- a/tests/functional_tests/ffront_tests/test_execution.py
+++ b/tests/functional_tests/ffront_tests/test_execution.py
@@ -480,7 +480,6 @@ def test_scalar_arg(fieldview_backend):
 
 
 def test_nested_scalar_arg(fieldview_backend):
-    """Test scalar argument being turned into 0-dim field."""
     if fieldview_backend == gtfn_cpu.run_gtfn:
         pytest.skip("ConstantFields are not supported yet.")
     Vertex = Dimension("Vertex")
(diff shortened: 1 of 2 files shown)


tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -479,6 +479,28 @@ def scalar_arg(scalar_inp: float64) -> Field[[Vertex], float64]:
     assert np.allclose(ref, out.array())
 
 
+def test_nested_scalar_arg(fieldview_backend):
```
egparedes, line 482: Would it make sense to add another test for operators with multiple inputs mixing scalar and field arguments?

egparedes (commented):
It looks good. Just a question about adding more tests.

tests/functional_tests/ffront_tests/test_execution.py
tehrengruber: Given how the lowering works and balancing between number of tests and coverage I would say it's not necessary. Your call.

egparedes: Ok, makes sense.

egparedes (approved):
Good to go.

PR #933, opened 2022-09-13
Functional: Fix scan tuple output

tehrengruber:
Fixes writing the tuple result of a scan operator into multiple output fields.

```python
@scan_operator(axis=KDim, forward=True, init=init, backend=fieldview_backend)
def simple_scan_operator(carry: tuple[float, float], x: float) -> tuple[float, float]:
    return (x, carry[1] + 1.0)

simple_scan_operator(x, out=(out1, out2))
```

This should fix the problem surfacing here: https://github.com/C2SM/icon4py/commit/b279780504e4e9467af53b388518103cf9596cb9#diff-f24f39bd690c6fffc48b87978e947d4df6e45eb767814ca796aee287bb8cb070L116

FYI @nfarabullini @leuty @egparedes

### DIFF
--- a/src/functional/iterator/transforms/inline_lambdas.py
+++ b/src/functional/iterator/transforms/inline_lambdas.py
@@ -31,7 +31,7 @@ class CountSymbolRefs(NodeVisitor):
 
 @dataclasses.dataclass
 class InlineLambdas(NodeTranslator):
-    """Inline lambda calls by substituting every arguments by its value."""
+    """Inline lambda calls by substituting every argument by its value."""
 
     opcount_preserving: bool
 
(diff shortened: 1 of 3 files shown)


src/functional/iterator/transforms/inline_lambdas.py
```diff
@@ -31,7 +31,7 @@ def visit_Lambda(self, node: ir.Lambda, *, active_refs: set[str]):
 
 @dataclasses.dataclass
 class InlineLambdas(NodeTranslator):
-    """Inline lambda calls by substituting every arguments by its value."""
+    """Inline lambda calls by substituting every argument by its value."""
```
egparedes, line 34: I guess this is an unrelated change...

egparedes (approved):
It looks good. Thanks for the quick fix.

src/functional/iterator/transforms/inline_lambdas.py
tehrengruber: Yes, still had allowed it to sneak in :-)

PR #938, opened 2022-09-17
Add flock to gt_src_manager install

### LINKED ISSUE #937: CPU backend fails in parallel without pre-installed gridtools sources
When I run our parallel tests using the CPU backend (probably any gridtools backend) without gridtools sources already installed, every rank tries to clone gridtools at the same time, leading to errors when earlier ranks continue on to compile code while other ranks are still deleting and overwriting those sources.

This could be fixed by importing mpi4py in a try: except block and doing MPI-parallel logic if more than one rank exists, or there may be some filesystem-based solution. If you decide this is expected behavior, we can manually clone gridtools sources on the first rank only when compiling a stencil through our wrapper layer.

jdahm:
## Description

Adds a OS file lock to the gridtools repo cloning process in `gt_src_manger`. This means no more than one process will clone it, so the will not clobber files.

The downside is that a file lock is left in the filesystem after cloning the repo. I would like to remove that, but I'm not sure when we have a guarantee that no procs are waiting on others to clone it.

Resolves #937.

### DIFF
--- a/src/gt4py/gt_src_manager.py
+++ b/src/gt4py/gt_src_manager.py
@@ -28,61 +28,48 @@ _GRIDTOOLS_INCLUDE_PATHS = gt_config.build_settings["gt_include_path"]
 _GRIDTOOLS_REPO_DIRNAMES = gt_config.GT_REPO_DIRNAME
 
 
-def install_gt_sources() -> bool:
-    if has_gt_sources():
-        return True
+def install_gt_sources() -> None:
+    if not has_gt_sources():
+        GIT_BRANCH = _GRIDTOOLS_GIT_BRANCH
+        GIT_REPO = _GRIDTOOLS_GIT_REPO
 
-    GIT_BRANCH = _GRIDTOOLS_GIT_BRANCH
-    GIT_REPO = _GRIDTOOLS_GIT_REPO
+        install_path = os.path.dirname(__file__)
+        external_path = os.path.abspath(os.path.join(install_path, "_external_src"))
+        target_path = os.path.join(external_path, _GRIDTOOLS_REPO_DIRNAMES)
+        lock_file = os.path.join(external_path, ".gridtools.lock")
 
-    install_path = os.path.dirname(__file__)
-    external_path = os.path.abspath(os.path.join(install_path, "_external_src"))
-    target_path = os.path.join(external_path, _GRIDTOOLS_REPO_DIRNAMES)
-    lock_file = os.path.join(external_path, ".gridtools.lock")
+        with open(lock_file, mode="w") as f:
+            fcntl.flock(f.fileno(), fcntl.LOCK_EX)
+            if not has_gt_sources():
+                git_cmd = f"git clone --depth 1 -b {GIT_BRANCH} {GIT_REPO} {target_path}"
+                print(f"Getting GridTools C++ sources...\n$ {git_cmd}")
+                subprocess.check_call(git_cmd.split(), stderr=subprocess.STDOUT)
 
-    with open(lock_file, mode="w") as f:
-        fcntl.flock(f.fileno(), fcntl.LOCK_EX)
         if not has_gt_sources():
-            git_cmd = f"git clone --depth 1 -b {GIT_BRANCH} {GIT_REPO} {target_path}"
-            print(f"Getting GridTools C++ sources...\n$ {git_cmd}")
-            subprocess.check_call(git_cmd.split(), stderr=subprocess.STDOUT)
-
-    is_ok = has_gt_sources()
-    if is_ok:
-        print("Success!!")
-    else:
-        print(
-            f"\nOooops! GridTools sources have not been installed!\n"
-            f"Install them manually in '{install_path}/_external_src/'\n\n"
-            f"\tExample: git clone --depth 1 -b {GIT_BRANCH} {GIT_REPO} {target_path}\n"
-        )
-
-    return is_ok
+            raise RuntimeError(
+                f"\nOooops! GridTools sources have not been installed!\n"
+                f"Install them manually in '{install_path}/_external_src/'\n\n"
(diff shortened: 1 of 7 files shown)


src/gt4py/gt_src_manager.py
```diff
@@ -28,29 +29,33 @@
 
 
 def install_gt_sources() -> bool:
-    is_ok = has_gt_sources()
-    if not is_ok:
-        GIT_BRANCH = _GRIDTOOLS_GIT_BRANCH
-        GIT_REPO = _GRIDTOOLS_GIT_REPO
+    if has_gt_sources():
+        return True
```
jdahm, line 33: I made this an early return instead of indenting the entire function

mcgibbon: This is generally considered an antipattern, since it's easy to miss the fact that the function returns early when scanning code. If you want to avoid indenting a lot of code, one way would be to make this function 2/3 lines (`if not has_gt_sources(): _force_install_gt_sources()`) and put most of the logic in a helper function.

src/gt4py/gt_src_manager.py
```diff
@@ -28,29 +29,33 @@
...
-            )
+    is_ok = has_gt_sources()
+    if is_ok:
+        print("Success!!")
+    else:
+        print(
+            f"\nOooops! GridTools sources have not been installed!\n"
+            f"Install them manually in '{install_path}/_external_src/'\n\n"
+            f"\tExample: git clone --depth 1 -b {GIT_BRANCH} {GIT_REPO} {target_path}\n"
+        )
```
mcgibbon, line 58: I know this is the current behavior, but should this really fail silently (without raising an exception)? It was very hard to tell that this was the problem in the past because it's hidden in long log output, and this isn't even logged with ERROR, it's just printed.

Won't this failing always lead to an exception down the line when we try to use the sources?

src/gt4py/gt_src_manager.py
```diff
@@ -28,29 +29,33 @@
...
-        target_path = os.path.abspath(
-            os.path.join(install_path, "_external_src", _GRIDTOOLS_REPO_DIRNAMES)
-        )
-        if not os.path.exists(target_path):
+    GIT_BRANCH = _GRIDTOOLS_GIT_BRANCH
+    GIT_REPO = _GRIDTOOLS_GIT_REPO
+
+    install_path = os.path.dirname(__file__)
+    external_path = os.path.abspath(os.path.join(install_path, "_external_src"))
+    target_path = os.path.join(external_path, _GRIDTOOLS_REPO_DIRNAMES)
```
mcgibbon, line 40: There is no guarantee that `gt_config.GT_REPO_DIRNAME` is a parent directory of `gt_config.build_settings["gt_include_path"]`. In other words, this logic will give unintended behavior depending on how gt_config is configured. The sources could very well get cloned, but still print saying they're not installed - I'm not sure whether they would be accessible for compilation in that case.

src/gt4py/gt_src_manager.py
jdahm: Yes. Nice call out. It's a good opportunity to fix that behavior and raise the proper exception here.

src/gt4py/gt_src_manager.py
jdahm: You're right that we expose a lever here that when set gives some unintended behavior. The value of the `gt_include_path` key in the `build_settings` dict is:
```
os.environ.get("GT_INCLUDE_PATH", GT_INCLUDE_PATH)
```
so the intended usage is that the user sets the env var `GT_INCLUDE_PATH` instead of adding a new key to the global state dict.

Is that what you are getting at? I might be misunderstanding the point.

jdahm: > If I set GT_INCLUDE_PATH to anything other than its default, this function will still be called and will still clone sources and report on whether it was successful

Are you positive this always happens? I'm only able to reproduce when I set `GT_INCLUDE_PATH` to one that does not contain `gridtools/common/defs.hpp`.

src/gt4py/gt_src_manager.py
jdahm: Good to know. It's fairly short after refactoring, so I just inlined it.

src/gt4py/gt_src_manager.py
jdahm: Does https://github.com/GridTools/gt4py/pull/938/commits/52ea3517adffa095ca2588691c123006467320b7 look like what you meant?

mcgibbon: Yes something like this. I would be careful about merging this and broadcast it to the rest of the gt4py team, since this could be very breaking if someone relies on it (even though I don't think anyone should or does).

It also strikes me that because we're always checking whether gt sources exist before calling the install command that there's no need to check whether the sources exist within the install command. In other words, the install command could install them regardless of whether they already exist, and the behavior would be unchanged (as long as the check were also added to the main block to update `python3 -m gt4py.gt_src_manager install`).

A different interpretation could be that we don't need to check if the sources are already installed within the backend code.

gronerl (commented):
I like the changes to the warning behavior, but agree that we should make an announcement and give a grace/rebuttal period.

A remark: FLOCK will not work on Windows.  On the other hand, gridtools will most likely not work there anyways, so I'm not sure if it's a real concern.

Let's also have the discussion whether this is needed or if a pip package for gridtools should be pushed.

mcgibbon:
We have a work-around now on our end so this isn't urgent, though the work-around is kind of ugly and the way the code breaks right now is hard to debug when it happens.

IMO the gridtools sources should be a subtree of this repository, unless you want/use the functionality of swapping gridtools versions with a static gt4py version. Alternatively a pip package would be great. It's always been strange that you need a runtime step to finish "installing" gt4py.

jdahm:
I'm happy to close this for now. This was a fun experiment, but I agree it's heavyweight. Let's focus on the pip package or putting the GridTools headers a subtree, as @mcgibbon suggests.

PR #939, opened 2022-09-19
Functional: Compiled backends integration unstructured

tehrengruber:
Support for using unstructured backends directly from Python.


havogt:
Can you try enable the tests in test_toy_connectivity and potentially others

tehrengruber:
As discussed offline the iterator tests will be enabled in a seperate PR after #1130 is merged. Until then only one test was enabled.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -48,25 +52,87 @@ def __call__(
...
+            interface.ScalarParameter | interface.BufferParameter | interface.ConnectivityParameter
+        ] = [
             get_param_description(program_param.id, obj)
             for obj, program_param in zip(inp.args, program.params)
+        ]
+        for name, connectivity in inp.kwargs["offset_provider"].items():
+            if isinstance(connectivity, Connectivity):
+                parameters.append(
+                    interface.ConnectivityParameter(
+                        "__conn_" + name.lower(), connectivity.origin_axis.value, name
```
havogt, line 65:
```suggestion
                        "conn_" + name.lower(), connectivity.origin_axis.value, name
```
`__something` is reserved in c++

src/functional/otf/binding/pybind.py
```diff
@@ -149,19 +149,34 @@ def visit_SidConversion(self, sid: SidConversion):
 
 
 def make_parameter(
-    parameter: interface.ScalarParameter | interface.BufferParameter,
+    parameter: interface.ScalarParameter
+    | interface.BufferParameter
+    | interface.ConnectivityParameter,
 ) -> FunctionParameter:
+    if isinstance(parameter, interface.ConnectivityParameter):
+        return FunctionParameter(name=parameter.name, ndim=2, dtype=np.dtype("int32"))
```
havogt, line 157: Maybe let's extract `np.dtype("int32")` into a variable that can be reused in make_argument, to make sure we don't miss changing both, if we have to change them later.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -48,25 +52,87 @@ def __call__(
...
             self.language_settings,
             f"""
-            #include <gridtools/fn/backend/naive.hpp>
-            {stencil_src}
-            {decl_src}
-            """.strip(),
+                    #include <gridtools/fn/backend/naive.hpp>
+                    #include <gridtools/stencil/global_parameter.hpp>
+                    {stencil_src}
+                    using gridtools::fn::sid_neighbor_table::as_neighbor_table;
```
havogt, line 133:
```suggestion
```
and use the full name above or create a namespace alias here, if you think it's too long and ugly.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -48,25 +52,87 @@ def __call__(
     ) -> stages.ProgramSource[languages.Cpp, languages.LanguageWithHeaderFilesSettings]:
         """Generate GTFN C++ code from the ITIR definition."""
         program = inp.program
-        parameters = tuple(
+        parameters: list[
```
havogt, line 55: Please refactor this function into smaller pieces. There is a lot going on now and I can't easily follow.

src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -139,42 +139,56 @@ def _collect_offset_definitions(
...
-    offset_tags: Iterable[itir.OffsetLiteral] = (
+    used_offset_tags: set[itir.OffsetLiteral] = (
         node.walk_values()
         .if_isinstance(itir.OffsetLiteral)
         .filter(lambda offset_literal: isinstance(offset_literal.value, str))
-    )
+        .getattr("value")
+    ).to_set()
+    if not used_offset_tags.issubset(set(offset_provider.keys())):
+        raise ValueError()  # TODO(tehrengruber): give meaningful error message
```
havogt, line 149: Resolve TODO

src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -139,42 +139,56 @@ def _collect_offset_definitions(
...
+        elif isinstance(dim_or_connectivity, common.Connectivity):
             offset_definitions[offset_name] = TagDefinition(name=Sym(id=offset_name))
+
+            connectivity: common.Connectivity = dim_or_connectivity
+            for dim in [
+                connectivity.origin_axis,
+                connectivity.neighbor_axis,
+            ]:
+                if grid_type == common.GridType.CARTESIAN:
+                    offset_definitions[dim.value] = TagDefinition(name=Sym(id=dim.value))
```
havogt, line 180: Does this make sense? Maybe just assert?

havogt (changes requested):
A first round of feedback. Didn't look into the gtfn_module where I asked for splitting into smaller pieces.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
tehrengruber: Done. It's still not a beauty, but should be fine now.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -43,33 +47,118 @@ class GTFNTranslationStep(
...
+            if isinstance(connectivity, Connectivity):
+                if connectivity.index_type not in [np.int32, np.int64]:
+                    raise ValueError(
+                        "Neighbor table indices must be of type `np.int32` or `np.int64`."
+                    )
+
+                # parameter
+                parameters.append(
+                    interface.ConnectivityParameter(
+                        "__conn_" + name.lower(),
```
havogt, line 108: `__conn` is not a legal C++ name. `__something` and `_Capital_something` is reserved.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -43,33 +47,118 @@ class GTFNTranslationStep(
...
+                connectivity_args.append(
+                    f"gridtools::hymap::keys<generated::{name}_t>::make_values({nbtbl})"
+                )  # TODO(havogt): std::forward, type and max_neighbors
+            elif isinstance(connectivity, Dimension):
+                pass
+            else:
+                raise ValueError(
+                    f"Expected offset provider `{name}` to be a `Connectivity` or `Dimension`, "
+                    f"but got {type(connectivity).__name__}."
+                )
```
havogt, line 131:
```suggestion
```
Or do you want to raise for future extensions of offset_provider? In that case maybe `NotImplementedError`?

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -43,33 +47,118 @@ class GTFNTranslationStep(
...
+                # connectivity argument expression
+                nbtbl = (
+                    f"gridtools::fn::sid_neighbor_table::as_neighbor_table<"
+                    f"generated::{connectivity.origin_axis.value}_t, "
+                    f"generated::{name}_t, {connectivity.max_neighbors}"
+                    f">(__conn_{name.lower()})"
+                )
+                connectivity_args.append(
+                    f"gridtools::hymap::keys<generated::{name}_t>::make_values({nbtbl})"
+                )  # TODO(havogt): std::forward, type and max_neighbors
```
havogt, line 124: Did I add this TODO in the first prototype? I don't know what it refers to now.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -43,33 +47,118 @@ class GTFNTranslationStep(
...
+            # argument expression
+            if (
+                isinstance(parameter, interface.ScalarParameter)
+                and parameter.name in closure_scalar_parameters
+            ):
+                # convert into sid
+                parameter_args.append(f"gridtools::stencil::global_parameter({parameter.name})")
+            else:
+                # pass as is
+                parameter_args.append(parameter.name)
```
havogt, line 95: I propose to do one small step more and put this into a function. This is a big improvement in my opinion, because, I can clearly see that in the next 15 lines
- only `args` and `program.params` is used
- plus an additional weird thing that we should get rid off

src/functional/program_processors/codegens/gtfn/gtfn_module.py
havogt, line 131: Same here, but even larger codeblock. I don't want to read all these details if I want to get a high-level overview.

src/functional/program_processors/runners/gtfn_cpu.py
```diff
@@ -35,6 +37,28 @@ def convert_arg(arg: Any) -> Any:
...
+                    "Only `NeighborTable` connectivities implemented at this point."
+                )
+            args.append(conn.table)
+        elif isinstance(conn, common.Dimension):
+            pass
+        else:
+            raise ValueError(
+                f"Expected offset provider `{name}` to be a `Connectivity` or `Dimension`, "
+                f"but got {type(conn).__name__}."
+            )
```
havogt, line 58:
```suggestion
```
Same comment as above.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -43,33 +47,118 @@ class GTFNTranslationStep(
...
+        self,
+        program: itir.FencilDefinition,
+        args: tuple[Any, ...],
+        offset_provider: dict[str, Connectivity],
+    ):
+        """Given an ITIR program and arguments generate the interface parameters and arguments."""
+        parameters: list[
+            interface.ScalarParameter | interface.BufferParameter | interface.ConnectivityParameter
+        ] = []
+        parameter_args: list[str] = ["gridtools::fn::backend::naive{}"]
```
havogt, line 60: maybe `program_args`

src/functional/program_processors/codegens/gtfn/gtfn_module.py
havogt: I believe if you extract the 2 loops into functions it's as good as it can be with the problematic interface regarding scalar args.

havogt (changes requested):
Can we do one more iteration in cleaning up the gtfn_module?

src/functional/program_processors/codegens/gtfn/gtfn_module.py
tehrengruber: Not sure what to do here. On the one hand you are absolutely right and we should not use double underscores `__` in generated C++ code, on the other hand this is not the only place were we use it, so using something different is also confusing. To name some examples where `__` is used: auto generated size arguments and auto generated symbol names in the lift inliner. The or at least one somewhat clean solution would be to replace those with something not reserved, but this is a larger task. Would you be fine with a todo for now?

src/functional/program_processors/codegens/gtfn/gtfn_module.py
tehrengruber: Yes, it is from your original PR. Shall I removed it then?

havogt (approved):
Thanks! Looks good.

PR #940, opened 2022-09-19
Domain implementation

nfarabullini:
Domain implementation in field_operator such that domain can directly be implemented within gt4py, rather than outside, e.g
```python
    @field_operator
    def fieldop_domain(a: Field[[IDim, JDim], float64]) -> Field[[IDim, JDim], float64]:
        return a + a

    @program
    def program_domain(a: Field[[IDim, JDim], float64]) -> Field[[IDim, JDim], float64]:
        fieldop_domain(a, out=a, domain={"IDim": (1, 9), "JDim": (4, 6)})

    program_domain(a, offset_provider={})
```


tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -870,6 +870,43 @@ def solve_tridiag(
...
+        return a
+
+    @program
+    def test_docstring(a: Field[[IDim], float64]) -> Field[[IDim], float64]:
+        """My docstring."""
+        fieldop_with_docstring(a, out=a)
+
+    test_docstring(a, offset_provider={})
+
+
```
havogt, line 889: Why is this here? Let's put it in a separate PR, if you want to add this test. Or is a merge with the main branch missing?

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -100,6 +97,27 @@ def visit_Call(self, node: past.Call, **kwargs):
...
+            if not isinstance(domain_kwarg, past.Dict):
+                raise GTTypeError(
+                    f"Only calls Dictionary allowed in field_domain, but got `{type(domain_kwarg)}`."
+                )
+            for domain_values in domain_kwarg.values_:
+                if not isinstance(domain_values.type, ct.TupleType):
+                    raise GTTypeError(
+                        f"Only Tuples allowed in `field_domain` dictionary values, but got `{domain_values.type}`."
+                    )
+
```
havogt, line 120: Additionally check that the tuple length is 2.

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -100,6 +97,27 @@ def visit_Call(self, node: past.Call, **kwargs):
...
+        if "out" not in new_kwargs:
+            raise GTTypeError("Missing required keyword argument(s) `out`.")
+        elif "field_domain" in new_kwargs:
+            domain_kwarg = new_kwargs["field_domain"]
+            if not isinstance(domain_kwarg, past.Dict):
+                raise GTTypeError(
+                    f"Only calls Dictionary allowed in field_domain, but got `{type(domain_kwarg)}`."
+                )
+            for domain_values in domain_kwarg.values_:
```
havogt, line 115: Also check that keys are `Dimensions` (see also my request for changing `str` -> `Dimension` in the test).

havogt: Before adding more checks, I propose to implement some type deduction test that you expect to fail, e.g.

`domain={1:(2,3)}`, `domain={Dim:(2,3,4)}` etc...

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -870,6 +870,43 @@ def solve_tridiag(
...
+    size = 10
+    a = np_as_located_field(IDim, JDim)(np.ones((size, size)))
+
+    @field_operator(backend=fieldview_backend)
+    def fieldop_domain(a: Field[[IDim, JDim], float64]) -> Field[[IDim, JDim], float64]:
+        return a + a
+
+    @program
+    def program_domain(a: Field[[IDim, JDim], float64]) -> Field[[IDim, JDim], float64]:
+        fieldop_domain(a, out=a, field_domain={"IDim": (1, 9), "JDim": (4, 6)})
```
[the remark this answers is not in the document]
havogt: Was there a reason for `field_domain` instead of just `domain`?

havogt (changes requested):
I only browsed quickly through the code and added some high-level feedback. In general I would add more tests for cases that should not pass and think about corner cases.

E.g.
- What is the expected behavior if `domain` has a dimension that the `out` field doesn't have and vice versa.
- Is there something to consider in the case of tuple out (I didn't think about it, maybe there is nothing special).

tests/functional_tests/ffront_tests/test_execution.py
nfarabullini: there was some duplication when I updated my branch off of functional

tests/functional_tests/ffront_tests/test_execution.py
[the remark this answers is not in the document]
nfarabullini: "domain" is used in other parts of the code, e.g. when defining the size of the output field in the embedded. I used "field_domain" to avoid confusion but I can change it back

havogt:
Domain and out-field(s) disagree in dimensions:

```python
out_field: Field[[IDim], ...]
domain={IDim:(is,ie), KDim:(ks,ke)}
# =>
for i in range(is,ie):
    for k in range(ks,ke): #parallel for
        out_field(i) = ...
```
order is not well-defined, let's forbid it for now

```python
out_field: Field[[IDim, KDim], ...]
domain={IDim:(is,ie), #implicit: KDim:(0, out_field.shape[KDim])}
# =>
for i in range(is,ie):
    out_field(i,?) = ...
```
2 choices:
- fill missing dimensions from the out-field size in the respective dimension(s)
- forbid

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -870,6 +870,67 @@ def solve_tridiag(
...
+
+    @field_operator(backend=fieldview_backend)
+    def fieldop_domain_tuple(
+        a: Field[[IDim, JDim], float64]
+    ) -> tuple[Field[[IDim, JDim], float64], Field[[IDim, JDim], float64]]:
+        return (a + a, a)
+
+    @program
+    def program_domain_tuple(a: Field[[IDim, JDim], float64], b: Field[[IDim, JDim], float64]):
+        fieldop_domain_tuple(a, out=(b, a), field_domain={IDim: (1, 9), JDim: (4, 6)})
```
havogt, line 923: Let's add a test where start and end values are not literals but passed to the program.

tests/functional_tests/ffront_tests/test_execution.py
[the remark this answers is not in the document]
havogt: In this other case it refers to the same concept: it defines the domain where the computation is applied. I'd propose to make it `domain` here.

src/functional/ffront/past_to_itir.py
```diff
@@ -123,7 +123,11 @@ def visit_Program(
     def _visit_stencil_call(self, node: past.Call, **kwargs) -> itir.StencilClosure:
         assert type_info.is_field_type_or_tuple_of_field_type(node.kwargs["out"].type)
 
-        output, domain = self._visit_stencil_call_out_arg(node.kwargs["out"], **kwargs)
+        if "field_domain" not in node.kwargs:
+            node.kwargs["field_domain"] = {}
+        output, domain = self._visit_stencil_call_out_arg(
+            node.kwargs["out"], node.kwargs["field_domain"], **kwargs
+        )
```
havogt, line 130: Is this only for passing field_domain or an empty dict? Then it's much cleaner to do something like 
```suggestion
        field_domain = node.kwargs.get("field_domain", {})
        output, domain = self._visit_stencil_call_out_arg(
            node.kwargs["out"], field_domain, **kwargs
        )
```
instead of manipulating the node object.

havogt: Actually, let's not pass an empty dict, but `None` in case the domain is not given.
Which reminds me that we should add a test with an empty domain, i.e. `field_op(..., out=..., domain={})`, and think about what the expected result is.

src/functional/ffront/past_to_itir.py
```diff
@@ -226,20 +280,25 @@ def _compute_field_slice(node: past.Subscript):
         return out_field_slice_
 
     def _visit_stencil_call_out_arg(
-        self, node: past.Expr, **kwargs
+        self, node_out: past.Expr, node_field_domain: past.Dict, **kwargs
```
havogt, line 283: This names should be more expressive (calling the out_arg `node` was not nice in the original code) and prepending node doesn't convey extra information in my opinion.
```suggestion
        self, out_arg: past.Expr, domain_arg: past.Dict, **kwargs
```

src/functional/ffront/past_to_itir.py
```diff
@@ -226,20 +280,25 @@ def _compute_field_slice(node: past.Subscript):
...
     ) -> tuple[itir.SymRef, itir.FunCall]:
-        if isinstance(node, past.Subscript):
+        if isinstance(node_out, past.Subscript):
```
havogt, line 285: if `node_out` is a Subscript, we probably must not have the `field_domain` argument (otherwise it's overspecified (or we need to define what the slicing means relative to the given domain)). Let's go with either domain or slicing. Please add a test in `test_past_parsing` that checks for an error in this case, i.e. `field_op(..., out=out_field[...], domain=...`. (`...` meaning, please fill in what makes sense.)

havogt: Note that there is another Subscript below inside `elif isinstance(node_out, past.TupleExpr)`. Test both cases.

havogt (changes requested):
I added some more comments about corner cases.

Small remark about reviewing best practices: if for any comments there is an open discussion (e.g. you reply with a question) then don't mark it as resolved, otherwise the reviewer might miss it and leave it to the reviewer to mark as resolved.

The other open thing that we discussed is to not generate the `__size` parameters in case domain is present, but we can also briefly discuss this with Till once he is back.

src/functional/iterator/embedded.py
```diff
@@ -374,6 +374,10 @@ def unstructured_domain(*args: NamedRange) -> UnstructuredDomain:
 
 @builtins.named_range.register(EMBEDDED)
 def named_range(tag: Tag | Dimension, start: int, end: int) -> NamedRange:
+    if isinstance(start, ConstantField):
+        start = start.value
+    if isinstance(end, ConstantField):
+        end = end.value
```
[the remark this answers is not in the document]
egparedes: I forgot to mention yesterday but there is a simpler and cleaner way to avoid problems like this with `ConstantField`s: `ConstantField` should support conversion to scalars by implementing  `__int__` and `__float__`,  `__bool__` and  `__complex__` (throwing TypeErrors when its value does not match the required type). Then, in client code potentially accepting scalars or 0-d fields, it would be enough to do something like: 

```python
from typing import SupportsInt
...
assert isinstance(start, SupportsInt)
assert isinstance(end SupportsInt)
...
 range(int(start), int(end))
```

This is completely in line with the idea of 0d-fields being both fields and scalars, and even the typing can be nicely expressed by using `SupportsInt` (or `SupportsFloat`, ...) from `typing`.

src/functional/ffront/past_to_itir.py
```diff
@@ -205,6 +217,39 @@ def _construct_itir_domain_arg(
...
+        try:
+            if node_domain.keys_[dim_i].type.dim == dim:
+                domain_bounds = []
+                for domain_i, domain_input in enumerate(node_domain.values_[dim_i].elts):
+                    if isinstance(domain_input, past.Name):
+                        domain_bounds.append(itir.SymRef(id=domain_input.id))
+                    elif isinstance(domain_input, past.Constant):
+                        domain_bounds.append(
+                            itir.Literal(value=str(domain_input.value), type="int")
+                        )
```
havogt, line 235: Ideally we would write this as something like
```suggestion
                    domain_bounds.append(self.visit(domain_input))
```
This already works for the case of `past.Constant` (it even does a better job because it has some checks, see `visit_Constant`), but not for `past.Name` because it would create a `Sym` instead of a `SymRef`. One solution could be to add a `as_ref` parameter to `visit_Name` (with default to False maybe) and then to a `self.visit(domain_input, as_ref=True)` here. Maybe ask for @tehrengruber 's feedback here.

src/functional/ffront/past_to_itir.py
```diff
@@ -205,6 +217,39 @@ def _construct_itir_domain_arg(
...
+    def _construct_itir_initialized_domain_arg(
+        self,
+        dim_i: int,
+        dim: Dimension,
+        node_domain: past.Dict,
+    ) -> tuple[itir.Literal, itir.Literal]:
+        try:
+            if node_domain.keys_[dim_i].type.dim == dim:
+                domain_bounds = []
+                for domain_i, domain_input in enumerate(node_domain.values_[dim_i].elts):
```
havogt, line 229: I would remove the for loop as we already know it has only 2 values and do something like 
```
assert len(node_domain.values_[dim_i].elts) == 2 
start, end = node_domain.values_[dim_i].elts
```
(assert here is fine as this should never trigger because we did a check in the frontend already)

src/functional/ffront/past_to_itir.py
```diff
@@ -205,6 +217,39 @@ def _construct_itir_domain_arg(
...
+                raise GTTypeError(
+                    f"Dimensions in out field and field domain are not equivalent"
+                    f"Expected {dim}, but got {node_domain.keys_[dim_i].type.dim} "
+                )
+        except Exception:
+            raise GTTypeError(
+                f"Upper and lower bounds could not be determined for {dim_i + 1}th dimension"
+            )
+
+        return domain_bounds[0], domain_bounds[1]
```
havogt, line 251: implementing my comments above step by step, I would expect this function to become more or less a one-liner (probably with an extra assert)
Something like
```
assert len(node_domain.values_[dim_i].elts) == 2 
return (self.visit(bound, as_ref=True) for bound in node_domain.values_[dim_i].elts)

**Review by havogt — CHANGES_REQUESTED** [review]
Some more comments and suggestions for cleanup

File: src/functional/ffront/past_to_itir.py [hunk 13]
```diff
@@ -172,21 +173,32 @@ def _construct_itir_out_arg(
...
             # bounds
-            lower = self._visit_slice_bound(
-                slices[dim_i].lower if slices else None,
-                itir.Literal(value="0", type="int"),
-                dim_size,
-            )
-            upper = self._visit_slice_bound(
-                slices[dim_i].upper if slices else None, dim_size, dim_size
-            )
+            if bool(node_domain) and bool(slices):
```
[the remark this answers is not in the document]
nfarabullini: I am more in favor of the 3rd option. I can implement it as such

src/functional/ffront/past_to_itir.py
nfarabullini: by doing that I get the following error for one of the tests:

TypeError: 'FunCall.args' must be <class 'functional.iterator.ir.Expr'> (got 'lower_i' which is a <class 'functional.iterator.ir.Sym'>

nfarabullini: I implemented a new approach, you can lmk if you want me to change it. Additionally, I added a test in the test_program for the error message in this section

havogt: The current implementation looks much better, but I still prefer my original suggestion because we are doing extra steps that should be handled by the visitor. The error that you got suggests that you were lowering it as `Sym` instead of `SymRef`. Did you implement the `as_ref` pattern that I suggested?

havogt (changes requested):
As far as I can see, removing the `__size` parameters is also not implemented. As this is close to done, we can also make it a separate PR.

havogt (approved):
Thanks! Looks good to me!

PR #946, opened 2022-09-22
Create field from sparse field outside of reduce [only test, missing implementation]

havogt:
This should be supported, but currently fails in lowering.


tehrengruber:
Solved in #965 https://github.com/GridTools/gt4py/pull/965/files#diff-96628d32cf28469162232288e2e12defc26d5cb658ac283aa0dbbd1d176e815aR505

PR #949, opened 2022-09-23
Functional: fix mypy config

egparedes:
Remove spurious `__init__.py` in the `src/` root folder and fix current `mypy` and `pre-commit` configurations to avoid silent ignoring meaningful errors.

The new `mypy` configuration is less permissive:
- it requires to explicitly ignore missing imports or it will trigger an error
- it requires to explicitly annotate optionals
- it disallows incomplete typing of functions (either is fully annotated or not)

The new `pre-commit` configuration runs `mypy` in the local development environment instead of using an isolated one. This change is in line with a recent trend in Python projects to run automated linting and checking tools in the same environment used for development to make sure errors are aligned and avoid duplicated configurations in `setup.cfg`/`pyproject.toml` and `pre-commit-config.yaml`.

All typing errors in the `eve` package have been fixed, but there are still more than 100 mypy errors waiting to be fixed in future PRs. They have been temporarily ignored in this PR (see multiple `ignore_errors = True` lines in mypy config in `setup.cfg`) to avoid creating a huge PR.


.github/workflows/qa.yml
```diff
@@ -16,12 +16,14 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v3
       with:
-        python-version: "3.10"
+        python-version: "3.10.6"
```
havogt, line 19: Why this change?

setup.cfg
```diff
@@ -185,23 +185,95 @@ known_third_party = atlas4py,attr,black,boltons,click,dace,devtools,factory,hypo
...
+# Global options
+install_types = True
+non_interactive = True
+
+# File paths and patterns
+exclude = (?x)(
+          /_external_src/
+          | /build/
+          | /setup\.py$
+          | /tests/
```
havogt, line 197: Does it mean tests are generally excluded?

havogt (approved):
Looks good to me. I like the strategy to get rid of the remaining errors.

setup.cfg
egparedes: Yes, they are fully excluded, this setting was already in place, but maybe we should change it in the future...

.github/workflows/qa.yml
egparedes: CPython 3.10.7 breaks mypy. The fix has been already added to the changes for the next version, but since it hasn't been released yet, this is required for now.

egparedes: I added the 0.980 release branch for mypy in the requirements so this change is not longer needed. Removed

PR #950, opened 2022-09-26
Functional: Fix problems with caching due to unstable ids in IR transformation

tehrengruber:
Previously some passes generated globally unique ids such that when one of those passes was executed multiple times in multiples field operator invocations a different id was used. Consequently a new cache id was generated and caching failed. This PR fixes that by using unique ids per pass invocation.

### DIFF
--- a/src/functional/fencil_processors/codegens/gtfn/itir_to_gtfn_ir.py
+++ b/src/functional/fencil_processors/codegens/gtfn/itir_to_gtfn_ir.py
@@ -11,13 +11,12 @@
 # distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
-
-
-from typing import Any, Iterable, Optional, Type, Union
+import dataclasses
+from typing import Any, ClassVar, Iterable, Optional, Type, Union
 
 import eve
 from eve.concepts import SymbolName
-from eve.utils import UIDs
+from eve.utils import UIDGenerator
 from functional import common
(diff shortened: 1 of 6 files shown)


src/functional/fencil_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -172,9 +172,9 @@ def _collect_offset_definitions(
     return offset_definitions
 
 
+@dataclasses.dataclass(frozen=True)
 class GTFN_lowering(eve.NodeTranslator, eve.VisitorWithSymbolTableTrait):
```
tehrengruber, line 176: Except for the `uids` change, the changes here are not required, but a partial cleanup of the pass such that the pass is still a frozen dataclass.

egparedes (approved):
It looks good.

src/functional/iterator/transforms/unroll_reduce.py
```diff
@@ -1,11 +1,21 @@
...
 from eve import NodeTranslator
-from eve.utils import UIDs
+from eve.utils import UIDGenerator
 from functional.iterator import ir
 
 
+@dataclasses.dataclass(frozen=True)
 class UnrollReduce(NodeTranslator):
+    # we use one UID generator per instance such that the generated ids are
+    #  stable across multiple runs (required for caching to properly work)
```
tehrengruber, line 12:
```suggestion
    # stable across multiple runs (required for caching to properly work)
```

PR #956, opened 2022-10-03
Functional: Scalar vs 0d ADR + tests

havogt:
This still needs implementation of the expected behavior in some cases.


docs/functional/architecture/0011-Scalar_vs_0d_Fields.md
```diff
@@ -0,0 +1,122 @@
+---
```
halungge, line 1: - (+1) for keeping the scalar in the user interface

For the user's perspective it would be awkward having to write x:Field[[], float] if you want to just pass a float and I suspect they hardly ever really have a 0d Field. As a user, I would rather have the approach, I have a float, and if you internally treat that as a 0d Field, i don't care but don't bother me with it. 

- (+1) for keeping the type annotation the annotations of the function and not the annotated function

Anything else is somehow intransparent and confusing for the user. 

And I think it should be a major concern to keep a user friendly interface in the front end. So if it doesn not complicate that we do down the path, I think this is a good solution.

docs/functional/architecture/0011-Scalar_vs_0d_Fields.md
```diff
@@ -0,0 +1,122 @@
...
+```python=
+@field_operator
+def foo() -> Field[[], float]
+    return 1.0
+```
+
+would be correct, e.g. if we implement `Field[[], float]` as a protocol that matches `1.0`.
+
+If we implement our own types, we have full freedom, to really make them aliases and get the following semantic work
+
```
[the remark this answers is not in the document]
havogt: It looks weird, but wouldn't be a problem, no?

docs/functional/architecture/0011-Scalar_vs_0d_Fields.md
```diff
@@ -0,0 +1,122 @@
...
+```python=
+scalar: gt4py.float(4.0)
+assert isinstance(scalar, gt4py.float)
+assert isinstance(scalar, Field[[], gt4py.float])
+assert scalar.dtype == gt4py.float
+assert scalar.dtype == Field[[], gt4py.float]
+assert scalar.dtype.dtype.dtype.dtype == gt4py.float
+assert scalar.dtype.dtype.dtype.dtype == Field[[], gt4py.float]
+```
+
```
egparedes, line 116: I would add another alternative here (as an ideal solution for the future): fields are different than scalars, but 0-d fields implement the appropriate scalar conversion operator matching its data type  (e.g. `__bool__`, `__int__`, `__float__`). In this way, 0-d field would also behave as scalar values in Python code. This is the approach used in the Python array API standard (https://data-apis.org/array-api/latest/API_specification/array_object.html). Additionally, although I personally think we should define our own scalar types, this alternative doesn't require it.

```python

a_0d_field

assert isinstance(a_0d_field, Field)

if a_0d_field > 3.5:
    ...
else:
    ...

```

havogt: Can you do the edit? Probably you can formulate that better...

egparedes: Done

tests/functional_tests/ffront_tests/test_foast_parsing.py
```diff
@@ -266,6 +275,54 @@ def conditional_wrong_arg_type(
...
+
+
+def test_wrong_return_type_annotation():
+    """See ADR 11."""
+
+    def wrong_return_type_annotation() -> Field[[], float]:
+        return 1.0
+
+    with pytest.raises(FieldOperatorTypeDeductionError, match="Expected `bool`"):
+        _ = FieldOperatorParser.apply_to_function(wrong_return_type_annotation)
```
halungge, line 302: Does this work? isn't that a strange error message? why bool? shouldn't it rather be something 'like Expected float'?

I had to set something like 

`    with pytest.raises(GTTypeError, match="Expected `float64`"):
        _ = FieldOperatorParser.apply_to_function(wrong_return_type_annotation)`

docs/functional/architecture/0011-Scalar_vs_0d_Fields.md
```diff
@@ -0,0 +1,139 @@
...
+```python=
+@field_operator
+def foo() -> Field[[], float]
+    return broadcast(1.0, ())
+```
+
+#### Implicit broadcast to Field
+
+In the following call an implicit broadcast to Field is performed in the decorator (the annotation influences behavior).
+
```
halungge, line 46: to make this more explicit (and consistent with the rest) you could add the return type explicitly here
`@field_operator
def foo(scalar: Field[[], float])->Field[[], float]:
    return scalar
`

halungge (approved):
Just some small comments. One of the tests looks strange to me.

tests/functional_tests/ffront_tests/test_foast_parsing.py
havogt: Thanks, good catch. There is another test later in the file with the same name, therefore this is never executed. I fixed the test.

PR #958, opened 2022-10-04
allow comparison between numeric and Undefined types

abishekg7:
## Description

When using K-offsets in relational/comparison operations, the edge cases can result in a comparison between a float/int type and an Undefined, which presently throws an error (below). This only shows up in embedded execution, as the correct vertical bounds are specified in integration.  This PR allows these patterns to bypass these errors in embedded execution. 

`Examples`
```python
p_face > p_face(Koff[1])
```
```python
p_cc < minimum(p_face, p_face(Koff[1]))
```

`Error`
```
TypeError: '<=' not supported between instances of 'float' and 'Undefined'
```


havogt (approved):
:+1:

PR #963, opened 2022-10-06
Domain ignore size args

nfarabullini:
Ignore computation of size_args when domain is specified in function call


src/functional/ffront/decorator.py
```diff
@@ -289,6 +289,7 @@ def _validate_args(self, *args, **kwargs) -> None:
     def _process_args(self, args: tuple, kwargs: dict) -> tuple[tuple, tuple, dict[str, Any]]:
         self._validate_args(*args, **kwargs)
 
+        domain_init = True if "domain" in self.past_node.body[0].kwargs else False
```
havogt, line 292: check all field operator calls, not only the first (`body[0]`)

src/functional/ffront/past_to_itir.py
```diff
@@ -107,7 +107,10 @@ def visit_Program(
         #  a workaround we add additional arguments to the fencil definition
         #  containing the size of all fields. The caller of a program is (e.g.
         #  program decorator) is required to pass these arguments.
-        size_params = self._gen_size_params_from_program(node)
+
+        params = [itir.Sym(id=inp.id) for inp in node.params]
+        if "domain" not in node.body[0].kwargs:
```
havogt, line 112: see other comment

havogt (changes requested):

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -69,9 +69,14 @@ def _validate_call_params(new_func: past.Name, new_kwargs: dict):
...
-                kind=ct.ScalarKind.INT64
-            ) or domain_values.elts[1].type != ct.ScalarType(kind=ct.ScalarKind.INT64):
+            scalar_types_check = [
+                ct.ScalarType(kind=ct.ScalarKind.INT64),
+                ct.ScalarType(kind=ct.ScalarKind.INT32),
+            ]
+            if (
+                domain_values.elts[0].type not in scalar_types_check
+                or domain_values.elts[1].type not in scalar_types_check
+            ):
```
havogt, line 79: I think it's worth adding a small utility function in `type_info.py`. There is already `is_integral` but that would also be true for integral fields. Therefore I propose to add `is_integral_scalar`, then the code here would be much easier to read:
```
if not is_inegral_scalar(domain_values.elts[0].type) or not is_integral_scalar(domain_values.elts[1].type)
```
Also you would not forget to add `ScalarKind.INT` to the list.

src/functional/ffront/past_to_itir.py
```diff
@@ -107,7 +107,14 @@ def visit_Program(
...
         #  program decorator) is required to pass these arguments.
-        size_params = self._gen_size_params_from_program(node)
+
+        params = [itir.Sym(id=inp.id) for inp in node.params]
+        params_add = False
+        for _i, body_entry in enumerate(node.body):
+            if "domain" not in body_entry.kwargs:
+                params_add = True
+
+        params = params + self._gen_size_params_from_program(node) if params_add else params
```
havogt, line 117: I think the following is easier to read:
```suggestion
        if params_add:
            params = params + self._gen_size_params_from_program(node)
```

src/functional/ffront/past_to_itir.py
```diff
@@ -107,7 +107,14 @@ def visit_Program(
         #  a workaround we add additional arguments to the fencil definition
         #  containing the size of all fields. The caller of a program is (e.g.
         #  program decorator) is required to pass these arguments.
-        size_params = self._gen_size_params_from_program(node)
+
+        params = [itir.Sym(id=inp.id) for inp in node.params]
+        params_add = False
+        for _i, body_entry in enumerate(node.body):
+            if "domain" not in body_entry.kwargs:
+                params_add = True
```
havogt, line 115: This can be simplified similarly to my proposal in decorator.py

src/functional/ffront/past_to_itir.py
```diff
@@ -107,7 +107,14 @@ def visit_Program(
         #  a workaround we add additional arguments to the fencil definition
         #  containing the size of all fields. The caller of a program is (e.g.
         #  program decorator) is required to pass these arguments.
-        size_params = self._gen_size_params_from_program(node)
+
+        params = [itir.Sym(id=inp.id) for inp in node.params]
```
havogt, line 111: It's not your code, but can you try if 
```suggestion
        params = self.visit(node.params)
```
works? I expect that all node.params are past.Name, but not sure if I overlook something.

src/functional/ffront/decorator.py
```diff
@@ -283,6 +283,11 @@ def _validate_args(self, *args, **kwargs) -> None:
     def _process_args(self, args: tuple, kwargs: dict) -> tuple[tuple, tuple, dict[str, Any]]:
         self._validate_args(*args, **kwargs)
 
+        initialized_domain = True
+        for _i, body_entry in enumerate(self.past_node.body):
```
havogt, line 287: not relevant, as I propose to remove the for loop anyway, but here it should be just
```suggestion
        for body_entry in self.past_node.body:
```
No need to enumerate.

havogt (dismissed):
I have a few style comments.

Can you also add a past lowering test that documents the behavior we are implementing here. Test that the itir fencil only contains the parameters that the user defined, not additional ones. (And maybe also a test that checks that the size params are added in the case where no domain is there.) Let me know if you need help with this kind of test.

src/functional/ffront/past_to_itir.py
nfarabullini: @tehrengruber self.visit does not work here

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -31,6 +31,11 @@ def _ensure_no_sliced_field(entry: past.Expr):
             _ensure_no_sliced_field(param)
 
 
+def _is_integral_scalar(scalar_type: past.Expr) -> bool:
+    """Check that past type is an integral scalar."""
```
tehrengruber, line 35:
```suggestion
    """Check that expression is an integral scalar."""
```

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -31,6 +31,11 @@ def _ensure_no_sliced_field(entry: past.Expr):
...
+    return type(scalar_type.type) == ct.ScalarType and type_info.is_integral(scalar_type.type)
```
tehrengruber, line 36:
```suggestion
    return isinstance(scalar_type.type, ct.ScalarType) and type_info.is_integral(scalar_type.type)
```

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -31,6 +31,11 @@ def _ensure_no_sliced_field(entry: past.Expr):
             _ensure_no_sliced_field(param)
 
 
+def _is_integral_scalar(scalar_type: past.Expr) -> bool:
```
tehrengruber, line 34:
```suggestion
def _is_integral_scalar(expr: past.Expr) -> bool:
```
The argument is not necessarily a scalar and in no case a type.

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -69,9 +74,9 @@ def _validate_call_params(new_func: past.Name, new_kwargs: dict):
                 raise GTTypeError(
                     f"Only 2 values allowed in domain range, but got `{len(domain_values.elts)}`."
                 )
-            if domain_values.elts[0].type != ct.ScalarType(
-                kind=ct.ScalarKind.INT64
-            ) or domain_values.elts[1].type != ct.ScalarType(kind=ct.ScalarKind.INT64):
+            if not _is_integral_scalar(domain_values.elts[0]) or not _is_integral_scalar(
+                domain_values.elts[1]
+            ):
```
tehrengruber, line 79:
```suggestion
            if not all(isinstance(el.type, ct.ScalarType) and type_info.is_integral(el.type) for el in domain_values.elts)
):
```
Just a suggestion.

tehrengruber (changes requested):

tehrengruber (changes requested):

tehrengruber (approved):

PR #970, opened 2022-10-11
[Cartesian] dace:cpu performance: OpenMP and Expansion Tuning

gronerl:
This is a first of a sequence of PRs improving the performance of the dace:cpu backend. 
In this first iteration, 
* adding OpenMP parallelism
* the loop order, array layout and tile sizes are chosen equal to those of cpu_kfirst
* test region predicates on a single index  with equality instead of two inequalities

I was not able to fully verify this on pace:
* this is already includes "no storage" with which pace is not  yet compatible
* runing pace translate tests with the equivalent branch `gronerl/dace_cpu_performance_oldstorage` which is still based on the storage classes gives a hand full of failed grid points with very small errors on some translate tests. @jdahm  coul you or someone at ai2 please verify that this is acceptable? I'll make sure to update `dace_cpu_performance_oldstorage` alongside this one.


gronerl:
We get within a factor of 1.5 of `gt:cpu_kfirst` with this PR for a 6 rank c192 run.

jdahm (approved):
Looks right to me too!

gronerl:
bors try

PR #972, opened 2022-10-12
Functional: scalar if

BenWeber42:
```python
@field_operator
def simple_if(a: Field[[IDim, JDim], float64], b: Field[[IDim, JDim], float64], condition: bool):
    if condition:
        result = a
    else:
        result = b
    return result
```

Including extended SSA pass supporting `If`s.


tests/functional_tests/ffront_tests/test_single_static_assign.py
```diff
@@ -133,3 +133,513 @@ def test_empty_annotated_assign():
...
+    expected = textwrap.dedent(
+        """
+        def f(a, a__0):
+            a__0 = a
+            a__0__0 = a__0
+            return a__0__0
+        """
+    ).strip()
+
+    assert result == expected
```
BenWeber42, line 621: We could fix this with by using a tuple assign: https://github.com/BenWeber42/gt4py/commit/698f45b80c3301024fcb9aae19d19d1a7629f7e0

But I feel like the implementation gets quite complicated then...

src/functional/ffront/ast_passes/single_static_assign.py
```diff
@@ -88,6 +104,62 @@ def __init__(self, separator="__"):
...
+
+        # For practical purposes, this is sufficient, but really not general at all.
+        # However, the algorithm was never intended to be general.
+
+        pre_assigns: list[ast.stmt] = [
+            _make_assign(arg.arg, arg.arg, arg) for arg in node.args.args
+        ]
+        # This forces versioning for parameters which is required for if-stmt support
+        node.body = pre_assigns + node.body
+        node.body = [self.visit(stmt) for stmt in node.body]
```
tehrengruber, line 117:
```suggestion
```
I would propose to avoid this and instead use the original name of the variable in the `visit_Name` and `visit_If` in case its name_counter is zero.

src/functional/ffront/ast_passes/single_static_assign.py
```diff
@@ -17,13 +17,23 @@
...
+
+
 class SingleStaticAssignPass(ast.NodeTransformer):
     """
     Rename variables in assignments to avoid overwriting.
 
     Mutates the python AST, variable names will not be valid python names anymore.
     This pass must be run before any passes that linearize unpacking assignments.
 
+    Requires AST to have no missing location info (see ``FixMissingLocations``).
```
tehrengruber, line 36: Is this really true? If some location info is missing `ast.copy_location` will just do nothing.

src/functional/ffront/ast_passes/single_static_assign.py
```diff
@@ -48,11 +58,17 @@ def foo():
...
+    Note that each variable name is assigned only once (per branch) and never updated / overwritten.
 
     Note also that after parsing, running the pass and unparsing we get invalid but
     readable python code. This is ok because this pass is not intended for
     python-to-python translation.
+
+    WARNING: This pass is not intended as a general-purpose SSA transformation.
+    The pass does not support any general Python AST. Known limitations include:
+        * Nested functions aren't supported
+        * Returns inside if blocks aren't supported
```
tehrengruber, line 70: Does this pass care about the returns? For later parts of the pipeline this is true, but the SSA property should still hold even for such cases.

tehrengruber (changes requested):

tests/functional_tests/ffront_tests/test_single_static_assign.py
tehrengruber: I think we should just use a common scheme for generated symbol names and include a check that verifies the user doesn't use such names himself. Something like `__a_0`.

src/functional/ffront/ast_passes/single_static_assign.py
```diff
@@ -48,11 +58,17 @@ def foo():
         a__2 = 3 + a__1
         return a__2
 
-    Note that each variable name is assigned only once and never updated / overwritten.
+    Note that each variable name is assigned only once (per branch) and never updated / overwritten.
```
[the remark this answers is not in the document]
BenWeber42: I think it's needed, because it breaks the assumption, that there's only ever one assignment per variable.

```
if True:
   a__0 = 1
else:
   # Having a second assignment anywhere else to the same variable wasn't possible before
   a__0 = 2
```

I'm unsure if any of the following passes relied on this.

It's a bit unclear, because the original sentence never clarified whether a variable _is assigned only once_ during an execution or within the source code.

src/functional/ffront/ast_passes/single_static_assign.py
BenWeber42: Well for once, it mayb add dead code like:

```python
a = 0
if True:
   a = 1
else:
   return 2
```
->
```python
a__0 = 0
if True:
   a__1 = 1
else:
   return 2
   a__1 = a__0
```

and secondly, it will break definite assignment analysis:

```python
if True:
   a = 1
else:
   return x
# because the else branch returned, `a` is guaranteed to be defined here
b = a
```
Currently, this is not supported and will become:
```python
if True:
   a__0 = 1
else:
   return x
b__0 = a
```

I might add support for returns, but I first want to continue with the foast parser & lowering.

tests/functional_tests/ffront_tests/test_single_static_assign.py
BenWeber42: Clearly, there should be a way to generate _fresh_ variable names. But of course, we're still at python AST level for this pass, so it makes sense that this functionality doesn't exist at this stage. After symbol resolution, it would be easier to support this.

As for the scheme to create fresh variables, I usually prefer a proper scheme that doesn't restrict the user in any way. But there are easier schemes if you're ok restricting the user. A friend of mine just took the md5 sum of the current date/time and prepended it to fresh variable names with the reasoning that the probability of a collision is negligible :sweat_smile:

src/functional/ffront/ast_passes/single_static_assign.py
```diff
@@ -88,6 +104,62 @@ def __init__(self, separator="__"):
     def _rename(self, node):
         return self.RhsRenamer.apply(self.name_counter, self.separator, node)
 
+    def visit_FunctionDef(self, node: ast.FunctionDef):
+
```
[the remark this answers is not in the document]
BenWeber42: It's interesting, I like to generously add empty lines to logically group code. I find it much more readable to have many empty lines for grouping than if there are no empty lines. I used to often add two empty lines to group code for bigger functions, but sadly can't do that anymore in python. I also really like to add an empty line at the beginning of a function. Just feels weird to have code on the first line (unless it's just a very small function). I feel like the function definition is like a title of a chapter or section, and the following text shouldn't hang under it immediately.

src/functional/ffront/ast_passes/single_static_assign.py
BenWeber42: I have a faint memory that it was needed for something, but I think my mind is in the wrong here :sweat_smile: 

Will remove the comment.

tests/functional_tests/ffront_tests/test_single_static_assign.py
egparedes: Just repeating here the same suggestion I quickly discussed offline with @tehrengruber: we could use an easily distinguishable Unicode char from rare/unused alphabets to avoid collisions and, at the same time, keep the names short, intelligible and traceable (by _traceable_ I mean that it should be obvious to infer the original name of the symbols from their name mangled versions)

src/functional/ffront/ast_passes/single_static_assign.py
BenWeber42: Fixed this in the latest version

src/functional/ffront/ast_passes/single_static_assign.py
BenWeber42: Removed

src/functional/ffront/ast_passes/single_static_assign.py
BenWeber42: SSA pass supports returns now, changed doc accordingly. Removed the line about returns.

src/functional/ffront/foast_to_itir.py
```diff
@@ -219,23 +219,40 @@ def visit_ScanOperator(self, node: foast.ScanOperator, **kwargs) -> itir.Functio
...
+        return im.let(self.visit(node.target, **kwargs), self.visit(node.value, **kwargs))(
+            inner_expr
+        )
+
+    def visit_IfStmt(self, node: foast.IfStmt, inner_expr: itir.Expr, **kwargs) -> itir.Expr:
+        cond = self.visit(node.condition, **kwargs)
+
+        common_symbols: dict[str, foast.Symbol] = node.annex._common_symbols
+        # pack the common symbols into a tuple
+        common_symrefs = im.make_tuple_(*(im.ref(sym) for sym in common_symbols.keys()))
```
BenWeber42, line 244: Should we copy those for both branches to avoid duplicate nodes in the tree?

src/functional/ffront/ast_passes/single_static_assign.py
```diff
@@ -11,12 +11,105 @@
...
 
 from functional.ffront.fbuiltins import TYPE_BUILTIN_NAMES
 
 
+def _make_assign(target: str, source: str, location_node: ast.AST):
+    result = ast.Assign(
+        targets=[ast.Name(ctx=ast.Store(), id=target)], value=ast.Name(ctx=ast.Load(), id=source)
+    )
+    for node in ast.walk(result):
+        ast.copy_location(node, location_node)
```
BenWeber42, line 27: This doesn't work for `result.value.ctx`, see #1031

src/functional/ffront/itir_makers.py
```diff
@@ -213,6 +213,11 @@ def tuple_get_(tuple_expr, index):
     return call_("tuple_get")(tuple_expr, index)
 
 
+def if_(cond, true_val, false_val):
+    """Create a not_ FunCall, shorthand for ``call("if_")(expr)``."""
+    return call_("if_")(cond, true_val, false_val)
```
BenWeber42, line 218: @tehrengruber you originally introduced this as `if__` (double underscore) and I changed it to `if_` (single underscore). Was there a particular reason why you used double underscore, or was it just a typo?

tests/functional_tests/ffront_tests/test_single_static_assign.py
```diff
@@ -133,3 +133,681 @@ def test_empty_annotated_assign():
...
+            a: int # annotations always apply to the next assignment
+            a = a + 1
+            """
+        )
+    )
+
+    expected = textwrap.dedent(
+        """
+        a__0: int
+        a__0 = a + 1
```
BenWeber42, line 659: This used to yield:

```python
        a__0: int
        a__0 = a__-1 + 1
```

but is fixed now...

tests/functional_tests/ffront_tests/test_single_static_assign.py
BenWeber42: This particular problem doesn't exist anymore. This PR uses the same name mangling scheme as the previous code. I think changing that is outside the scope of this PR.

tehrengruber:
Closed in favor of #1079

PR #973, opened 2022-10-12
Make aligned_index an optional keyword arg of storage methods

jdahm:
## Description

Built-in numpy allocation methods can now be used to create storages, but the axes nearly always end up in the wrong order unless you do some numpy magic. Given that, the allocation methods from gt4py.storage are still nearly always used, so requiring `aligned_index` doesn't make sense. Users are finding it confusing that they have to pass `aligned_index` to these, given that they do not set a default origin for the `ndarray`.

What this does:
- makes that keyword argument optional
- adds a bit of type hinting internally to make the methods more understandable
- adds the storage files to pre-commit and fixes the few errors flake8 found

### DIFF
--- a/tests/test_unittest/test_storage.py
+++ b/tests/test_unittest/test_storage.py
@@ -27,7 +27,7 @@ except ImportError:
 import gt4py.backend as gt_backend
 import gt4py.storage
 from gt4py import gtscript
-from gt4py.storage.utils import normalize_storage_spec
+from gt4py.storage.utils import allocate_cpu, allocate_gpu, normalize_storage_spec
 
 from ..definitions import CPU_BACKENDS, GPU_BACKENDS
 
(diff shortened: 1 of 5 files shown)


havogt:
bors try

tests/test_unittest/test_storage.py
```diff
@@ -189,12 +187,11 @@ def test_allocate_gpu(param_dict):
...
     layout_map = param_dict["layout_order"]
-    device_raw_buffer, device_field = gt_storage_utils.allocate_gpu(
-        aligned_index, shape, layout_map, dtype, alignment_bytes
+    device_raw_buffer, device_field = allocate_gpu(
+        shape, layout_map, dtype, alignment_bytes, aligned_index
     )
 
-    # assert (device_field.base is device_raw_buffer) # as_strided actually returns an ndarray where base=None??
-    # instead check that the memory of field is contained in raws buffer:
+    # check that the memory of field is contained in raws buffer
```
jdahm, line 194:
```suggestion
    # Would like to check device_field.base against device_raw_buffer but
    # as_strided returns an ndarray where device_field.base is set to None.
    # Instead, check that the memory of field is contained in raws buffer
```

gronerl:
I agree that the allocation of a suitable array is difficult and is an issue. I would however propose to solve this with #981 instead.

gronerl (approved):
No opposition.

PR #978, opened 2022-10-14
Update dependencies to match those in 'functional'

jdahm:
## Description

... and fix the few errors that the new mypy version catches.

PR #979, opened 2022-10-15
Remove src/__init__.py and fix many errors

### LINKED ISSUE #1007: gt4py dependency analysis thinks parameters are written when used as data indices
This newly added test currently fails:
```python
    @pytest.mark.parametrize("backend", ALL_BACKENDS)
    def test_write_data_dim_indirect_addressing(backend):
        INT32_VEC2 = (np.int32, (2,))
    
        def stencil(
            input_field: gtscript.Field[gtscript.IJK, np.int32],
            output_field: gtscript.Field[gtscript.IJK, INT32_VEC2],
            index: int,
        ):
                output_field[0, 0, 0][index] = input_field
    
        aligned_index = (0, 0, 0)
        full_shape = (1, 1, 2)
        input_field = gt_storage.ones(
            full_shape, backend=backend, aligned_index=aligned_index, dtype=np.int32
        )
        output_field = gt_storage.zeros(
            full_shape, backend=backend, aligned_index=aligned_index, dtype=INT32_VEC2
        )
    
>       gtscript.stencil(definition=stencil, backend=backend)(input_field, output_field, index := 1)
```

jdahm:
## Description

Major changes:
- Adds type checking on a few more files.
- `Expr.dtype`'s type is no longer `Optional[DataType]`, because that caused type havoc. Therefore, the default is changed from `None` to `DataType.AUTO` when not set explicitly. The `gtir_dtype_resolver` and other visitors are updated to check for `DataType.AUTO` instead of `None`.
- Moves most of the code that was in `eve.iterators` to `eve.concepts` to work around a circular dependency after solving https://github.com/GridTools/gt4py/issues/565 using @egparedes's suggestion. This is a temporary workaround until we replace Eve with the new version.
- `gt4py.stencil_object.ArgsInfo` is made non-private and used in `gt4py.dace_stencil_object` to fix a type issue.
- Ignores typing in eve visitor contexts -- need to fix that in a follow-up PR.
- `npir.HorizontalMask` is created as a tuple of tuples, to avoid the `Optional[AxisBound]` in the `common.HorizontalInterval` type.
- Replaces `gt_backend` references with `gt4py.backend`.

Follows #978.

Resolves https://github.com/GridTools/gt4py/issues/565.
Resolves https://github.com/GridTools/gt4py/issues/992.
Resolves #1007.


gronerl:
Looks like there's a circular import problem now, how should we proceed with this?

jdahm:
> Looks like there's a circular import problem now

I committed the code and pushed after nearly a whole day of working out these issues, only to find out about it 😆 .

> how should we proceed with this?

The circular dependency should be solved, or rather avoided, by moving to the next version of `eve`. But in this case a bit of analysis and moving around code definitions within this version of `eve` could solve it.

src/gt4py/stencil_object.py
```diff
@@ -284,27 +283,24 @@ def __call__(self, *args, **kwargs) -> None:
         pass
 
     @staticmethod
-    def _make_origin_dict(origin: Any) -> Dict[str, Index]:
+    def _make_origin_dict(origin: Any) -> Dict[str, Tuple[int, ...]]:
```
jdahm, line 286: TODO: This should be a union of types, not `Any`.

src/gt4py/stencil_builder.py
```diff
@@ -70,9 +70,15 @@ def __init__(
         )
         backend = backend or "numpy"
         backend = gt4py.backend.from_name(backend) if isinstance(backend, str) else backend
+        assert backend is not None
```
jdahm, line 73: TODO: This should raise an assertion with a nicely formatted error message.

jdahm:
bors try

src/gt4py/backend/module_generator.py
```diff
@@ -102,7 +102,9 @@ def make_args_data_from_gtir(pipeline: GtirPipeline) -> ModuleData:
         )
 
     for decl in (param for param in all_params if isinstance(param, gtir.ScalarDecl)):
-        access = accesses[decl.name]
+        access = cast(Literal[AccessKind.NONE, AccessKind.READ], accesses[decl.name])
+        print(access)
```
jdahm, line 106:
```suggestion
```

src/gt4py/stencil_object.py
```diff
@@ -396,6 +395,8 @@ def _validate_args(  # noqa: C901  # Function is too complex
                 f"Compute domain too small. Sequential axis is {domain[2]}, but must be at least {self.domain_info.min_sequential_axis_size}."
             )
 
+        gt_backend = gt4py.backend.from_name(self.backend)
```
jdahm, line 398: In order not to confuse this with past code that uses `gt_backend` to refer to `gt4py.backend`, I will rename this to `stencil_backend`.

jdahm:
bors try

jdahm:
bors try

src/gt4py/cli.py
```diff
@@ -62,15 +62,16 @@ def convert(
...
     @staticmethod
     def enabled_backend_cls_from_name(backend_name: str) -> Optional[Type[CLIBackendMixin]]:
         """Check if a given backend is enabled for CLI."""
-        backend_cls = gt_backend.from_name(backend_name)
+        backend_cls = gt4py.backend.from_name(backend_name)
+        assert backend_cls is not None
         if not issubclass(backend_cls, CLIBackendMixin):
             return None
-        return backend_cls
+        return cast(Optional[Type[CLIBackendMixin]], backend_cls)
```
egparedes, line 74: Is the `Optional` typing needed here?
```suggestion
        return cast(Type[CLIBackendMixin], backend_cls)
```

src/gt4py/stencil_object.py
```diff
@@ -294,17 +296,17 @@ def _make_origin_dict(origin: Any) -> Dict[str, Index]:
             if origin is None:
                 return {}
             if isinstance(origin, collections.abc.Iterable):
-                return {"_all_": Index.from_value(origin)}
+                return {"_all_": cast(Tuple[int, ...], Index.from_value(origin))}
```
egparedes, line 299: This is just an idea not a requirement because it might involve a lot of work: it would be conceptually better to declare `NumericTuple` as a generic subclass of tuple (`Tuple[T, ...]`) and add the missing typing hints. Concrete subclasses could then be defined as concrete instances of `NumericTuple` (for example `class Index(NumericTuple[int]):`

src/gtc/dace/expansion/daceir_builder.py
```diff
@@ -263,9 +264,9 @@ def pop(self) -> "DaCeIRBuilder.IterationContext":
 
     @dataclass
     class SymbolCollector:
-        symbol_decls: Dict[SymbolRef, dcir.SymbolDecl] = dataclasses.field(default_factory=dict)
+        symbol_decls: Dict[str, dcir.SymbolDecl] = dataclasses.field(default_factory=dict)
```
egparedes, line 267: Is really `str` the correct typing? Why is not longer `SymbolRef`?

src/gtc/passes/oir_access_kinds.py
```diff
@@ -37,7 +37,8 @@ def visit_ScalarAccess(self, node: oir.ScalarAccess, **kwargs: Any) -> None:
         self._visit_Access(node.name, **kwargs)
 
     def visit_FieldAccess(self, node: oir.FieldAccess, **kwargs: Any) -> None:
-        self.generic_visit(node, **kwargs)
+        self.visit(node.offset, **{**kwargs, "kind": AccessKind.READ})
+        self.visit(node.data_index, **{**kwargs, "kind": AccessKind.READ})
```
egparedes, line 41: Is it possible/expected that `kwargs` has a `kind` item? If it is not possible, I would write this in a simpler way:
```suggestion
        self.visit(node.offset, kind=AccessKind.READ, **kwargs)
        self.visit(node.data_index, kind=AccessKind.READ, **kwargs)
```
Additionally, why have been these two functions added here?

egparedes (commented):
It's impressive the amount of work you've done in this PR to fix all typings! It looks good to me, just a couple of very minor questions and suggestions, but I don't see anything blocking the merge.

src/gt4py/stencil_object.py
jdahm: This sounds like a great idea. Let's try it in a separate PR.

src/gtc/dace/expansion/daceir_builder.py
jdahm: There were some issues using `SymbolRef`. I can try adding it back in a separate PR, but for now let's stick with `str` and merge as soon as possible.

src/gtc/passes/oir_access_kinds.py
jdahm: `kind` is sometimes in the dict when it comes directly from the `visit_AssignStmt`.

src/gtc/dace/expansion/daceir_builder.py
egparedes: Agree

src/gt4py/stencil_object.py
egparedes: :+1:

src/gt4py/cli.py
jdahm: Nice catch!

egparedes (approved):
It looks good to go.

jdahm:
bors try

PR #980, opened 2022-10-17
[Cartesian] Dace CPU Performance: Corner regions and improved tiling

gronerl:
## Description
Second batch of changes to the LibraryNode expansion of the `dace:cpu` backend, addressing performance. 

Here, the changes are
* improvements to region splitting and corner-only regions
* changes to the tiling scheme (there is now the same stride 8 in all tiling maps)
* tile temporaries thread-locally where possible (minor impact)
* codegen trick: NestedSDFG's on all inner maps to more eagerly offset pointers.

There is a `gronerl/dace_cpu_performance_oldstorage-2` branch that is equivalent, but with the old storage classes in place for compatibility with pace. With the that, the performance of the `dace:cpu` backend is equivalent to that of `gt:cpu_kfirst` for `pace`. More specifically, `dace:cpu` has an advantage on regions-heavy stencils working only on corners. (By not iterating the ij domain at all) On the other hand,  `gt:cpu_kfirst` is faster on stencils where expansion features are not yet implemented in the `dace:cpu` backends. Most prominently, those are stencils with multiple `VerticalLoop` nodes in OIR.
These will be adressed in the next iteration of this optimization work.


gronerl:
bors try

src/gtc/dace/transformations.py
```diff
@@ -0,0 +1,116 @@
...
+
+        toremove = InlineTransients._candidates(sdfg, graph, scope_subgraph.nodes()[0])
+        return len(toremove) > 0
+
+    def apply(self, graph, sdfg):
+        map_entry = self.map_entry
+
+        scope_subgraph = graph.scope_subgraph(map_entry, include_entry=False, include_exit=False)
+        nsdfg_node = scope_subgraph.nodes()[0]
+        toremove = InlineTransients._candidates(sdfg, graph, nsdfg_node)
```
tbennun, line 70: misleading return value name

src/gtc/dace/transformations.py
```diff
@@ -0,0 +1,116 @@
...
+        InlineTransients.apply_to(sdfg, nsdfg=nsdfg_node)
```
tbennun, line 71: use `save=False` to avoid duplicate history saving

src/gtc/dace/transformations.py
```diff
@@ -0,0 +1,116 @@
...
+            visited.add(node)
+        if map_entry.schedule == dace.ScheduleType.Sequential:
+            map_entrys = [map_entry]
+            for me in reversed(map_entrys):
+                subgraph = state.scope_subgraph(me, include_entry=False, include_exit=False)
+                nest_state_subgraph(sdfg, state, subgraph)
+
+    for state in sdfg.nodes():
+        for map_entry in filter(
+            lambda n: isinstance(n, dace.nodes.MapEntry) and n not in visited, state.nodes()
```
tbennun, line 113: this is not going to be necessarily topologically sorted (it may visit an internal map before an external, or the last one before the first). In most cases, states are ordered topologically but that is not always the case. Perhaps use `state.scope_children()[None]` to get the top-level maps.

src/gtc/dace/transformations.py
```diff
@@ -0,0 +1,116 @@
...
+        applied = False
+        for map_entry, state in sdfg.all_nodes_recursive():
+            if isinstance(map_entry, dace.nodes.MapEntry):
+                if map_entry.map.schedule in {
+                    dace.ScheduleType.Sequential,
+                    dace.ScheduleType.CPU_Multicore,
+                }:
+                    # exclude maps with empty edges as workaround for a bug in TrivialMapElimination
+                    if any(
+                        edge.data.data is None
```
tbennun, line 20:
```suggestion
                        edge.data.is_empty()
```

tbennun (changes requested):
I did a pass over the dace transformation part. It looks good, but there is some common "gotcha" pattern that I found (looping over nodes() assuming they are ordered).

src/gtc/dace/transformations.py
gronerl: Technically, I don't expect it to, hence all the `visited` tomfoolery. But the `state.scope_children()[None]` approach should make all of this less convoluted, I will implement it.

src/gtc/dace/transformations.py
gronerl: I put `candidates` now and check if it has actually been applied in `apply`.

tbennun (commented):
DaCe API use looks good now, thanks!

src/gt4py/backend/dace_backend.py
```diff
@@ -452,16 +467,25 @@ class DaCeComputationCodegen:
...
+            (
+                global_fmt if array.storage != dace.StorageType.CPU_ThreadLocal else threadlocal_fmt
+            ).format(
+                name=name,
+                sdfg_id=array_sdfg.sdfg_id,
                 dtype=array.dtype.ctype,
                 size=array.total_size,
             )
             for array_sdfg, name, array in sdfg.arrays_recursive()
             if array.transient and array.lifetime == dace.AllocationLifetime.Persistent
```
jdahm, line 486: I get what you're trying to do here, and it's neat, but it's also a really long statement. Consider splitting this up for readability?

src/gtc/dace/transformations.py
```diff
@@ -0,0 +1,112 @@
...
+                if map_entry.map.schedule in {
+                    dace.ScheduleType.Sequential,
+                    dace.ScheduleType.CPU_Multicore,
+                }:
+                    # exclude maps with empty edges as workaround for a bug in TrivialMapElimination
+                    if any(
+                        edge.data.is_empty()
+                        for edge in state.in_edges(map_entry)
+                        + state.out_edges(state.exit_node(map_entry))
+                    ):
```
jdahm, line 23: What does the + mean in this context? Is it adding a generator expression to another? :O

src/gtc/dace/transformations.py
```diff
@@ -0,0 +1,112 @@
...
+from dace.transformation.interstate import InlineTransients
+
+
+def eliminate_trivial_maps(sdfg: dace.SDFG):
+    """Remove maps and map ranges where the iteration is over a single index."""
+    applied = True
+    while applied:
+        applied = False
+        for map_entry, state in sdfg.all_nodes_recursive():
+            if isinstance(map_entry, dace.nodes.MapEntry):
```
jdahm, line 13: If this were made a separate Iterable, then the indentation from the check on line 13 could be removed (this has a lot of levels of indentation already :).
```suggestion
        map_entry_and_state = (x for x in sdfg.all_nodes_recursive() if isinstance(x[0], dace.nodes.MapEntry))
        for map_entry, state in map_entry_and_state:
```

src/gtc/dace/transformations.py
```diff
@@ -0,0 +1,112 @@
...
+from dace.transformation.dataflow import TrivialMapElimination
+from dace.transformation.helpers import nest_state_subgraph
+from dace.transformation.interstate import InlineTransients
+
+
+def eliminate_trivial_maps(sdfg: dace.SDFG):
+    """Remove maps and map ranges where the iteration is over a single index."""
+    applied = True
+    while applied:
+        applied = False
```
jdahm, line 11: When I first read this I thought it was a bug, but I see what you're doing here. Is there a way to write the termination condition in a way that it doesn't look like one?

src/gtc/dace/expansion/utils.py
```diff
@@ -107,30 +107,45 @@ def mask_includes_inner_domain(mask: common.HorizontalMask):
...
+        for stmt in he.body:
+            if isinstance(stmt, oir.HorizontalRestriction) and not mask_includes_inner_domain(
+                stmt.mask
+            ):
+                continue
+            elif isinstance(stmt, oir.LocalScalar):
+                continue
+            elif isinstance(stmt, oir.AssignStmt) and isinstance(stmt.left, oir.ScalarAccess):
+                continue
+            return False
```
jdahm, line 121: I don't understand why the check on line 117 is needed. `oir.HorizontalRestriction.body: List[oir.Stmt]` (is a list of OIR statements), which means that `stmt` can never be an `oir.LocalScalar`. Or am I misunderstanding something here?

jdahm: If you need a check that there are no local scalars defined on the horizontal execution, then something like `not he.declarations` should suffice.

jdahm (approved):

src/gtc/dace/transformations.py
gronerl: lists can be joined with the + operator, we do this everywhere :)
I would like to avoid making this longer, the do something for inputs... and outputs... is everywhere in the dace backend and I'd like to keep the overhead that causes minimal.

src/gtc/dace/transformations.py
gronerl: i moved all three levels of conditionals into a separate filter method. That way it's even 2 levels shallower and the `continue` statement is also avoided.

src/gtc/dace/transformations.py
gronerl: The problem is the double loop, so a `break` won't help us Would you like this better?
```
def _apply_once()
    for map_entry, state in filter(_filter_map_entries, sdfg.all_nodes_recursive()):
        try:
            TrivialMapElimination.apply_to(
                state.parent, map_entry=map_entry, verify=True, save=False
            )
            return True
        except ValueError:
          continue
    return False
    
while _apply_once()
    pass
```

gronerl: Never mind, I redid the entire thing using dace transformation API, not sure why I thought this wasn't possible before.

src/gtc/dace/transformations.py
gronerl: Never mind, I redid the entire thing using dace transformation API, not sure why I thought this wasn't possible before.

src/gtc/dace/expansion/utils.py
gronerl: Oh yeah, I wrongly remembered that declarations were inlined at the HorizontalExecution level. Fixed.

gronerl:
I updated both branches, please see if your comments were adequately addressed.

gronerl:
bors try

src/gtc/common.py
```diff
@@ -831,6 +831,27 @@ def is_single_index(self) -> bool:
 
         return abs(self.end.offset - self.start.offset) == 1
 
+    def overlap(self, other: "HorizontalInterval") -> bool:
```
jdahm, line 834: I would like to rename this `overlaps`, but since the GitLab CI has already run, we should do this as a future step next time this file is touched.

src/gtc/dace/transformations.py
jdahm: Ah ok, but it's not a list in this case, rather `edge.data.is_empty() for edge in state.in_edges(map_entry)` is a Generator expression. I wasn't aware that that `+` was overloaded for them the same way it is for Lists. It must be that there is something like an `itertools.chain` happening behind the scenes with a closure. Interesting...

src/gtc/dace/expansion/utils.py
```diff
@@ -107,30 +107,43 @@ def mask_includes_inner_domain(mask: common.HorizontalMask):
...
-                    (
-                        isinstance(last_stmts[0], oir.HorizontalRestriction)
-                        and not mask_includes_inner_domain(last_stmts[0].mask)
-                    )
-                )
-            ):
-                last_stmts = [stmt]
-                res_he_stmts.append(last_stmts)
+            if isinstance(stmt, oir.LocalScalar):
+                local_scalars.append(stmt)
```
jdahm, line 138: Forgot to delete this? :)

jdahm (approved):

src/gtc/dace/transformations.py
gronerl: Ah, I see what you are saying. I don't think the `+` is overloaded like that, but this example has different associtaivity though: The  `+` operates on the operands `state.in_edges(map_entry)` and `state.out_edges(state.exit_node(map_entry))`. I put parantheses now to make it clearer to read (and `black` lets me keep them :sweat_smile: )

src/gtc/dace/expansion/utils.py
gronerl: yes!

src/gtc/common.py
gronerl: udpated

gronerl:
bors try

PR #981, opened 2022-10-17
Allow any layout

gronerl:
## Description
This PR will allow ndarrays with any strides to be passed in to any backend. The performance backends will raise a Warning instead of an Exception. 

The PR also unifies some of the code used across the different backends to determine and check optimal layouts.

This PR is a counter proposal to  #973

### DIFF
--- a/src/gt4py/backend/base.py
+++ b/src/gt4py/backend/base.py
@@ -71,7 +71,7 @@ class Backend(abc.ABC):
     #:  - "alignment": in bytes
     #:  - "device": "cpu" | "gpu"
     #:  - "layout_map": callback converting a mask to a layout
-    #:  - "is_compatible_layout": callback checking if a storage has compatible layout
+    #:  - "is_optimal_layout": callback checking if a storage has compatible layout
     storage_info: ClassVar[Dict[str, Any]]
 
     #: Language support:
--- a/src/gt4py/backend/gtc_common.py
+++ b/src/gt4py/backend/gtc_common.py
@@ -36,7 +36,7 @@ from gtc.passes.oir_pipeline import OirPipeline
 if TYPE_CHECKING:
     from gt4py.stencil_builder import StencilBuilder
     from gt4py.stencil_object import StencilObject
-    from gt4py.storage.storage import Storage
+    from gt4py.storage.storage import ArrayLike
 
 
 def _get_unit_stride_dim(backend, domain_dim_flags, data_ndim):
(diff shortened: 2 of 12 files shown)


jdahm:
I think this is a great addition. I don't see this as a counter to #973, in fact I think both would be ideal. There's no reason that a user (especially those just starting out with gt4py) would always want to set `aligned_index` on memory, especially since it doesn't set an origin. However, there is a case to use the `gt4py.storage.{empty,ones,...} simply because they always yield the optimal layout.

So that is to say I think we should move forward with this, but let's also still discuss/move forward with making `aligned_index` optional. I don't see a reason why that would be a bad idea, but if you do, let me know.

gronerl:
It is true that layout is much more important than alignment, I think you've convinced me :thinking: I'll address this again with the team.

src/gt4py/backend/gtc_common.py
```diff
@@ -383,88 +370,24 @@ def _permute_layout_to_dimensions(
...
-    mask = _dimensions_to_mask(dimensions)
-    ctr = iter(range(sum(mask)))
-    if len(mask) < 3:
-        layout: List[Optional[int]] = [next(ctr) for m in mask if m]
-    else:
-        swapped_mask: List[Optional[int]] = [*mask[3:], *mask[:3]]
-        layout = [next(ctr) if m else None for m in swapped_mask]
-
-        layout = [*layout[-3:], *layout[:-3]]
+def make_gtcpu_kfirst_layout_map(dimensions: Tuple[str, ...]) -> Tuple[int, ...]:
```
jdahm, line 373: Nice work reworking these! This has been needed for a while now.

jdahm: They are so much easier to understand now, too.

src/gt4py/backend/base.py
```diff
@@ -71,7 +71,7 @@ class Backend(abc.ABC):
     #:  - "alignment": in bytes
     #:  - "device": "cpu" | "gpu"
     #:  - "layout_map": callback converting a mask to a layout
-    #:  - "is_compatible_layout": callback checking if a storage has compatible layout
+    #:  - "is_optimal_layout": callback checking if a storage has compatible layout
```
jdahm, line 74:
```suggestion
    #:  - "is_optimal_layout": callback checking if a storage has an optimal memory layout
```

jdahm (approved):
Looks good to me!

gronerl:
bors try

gronerl:
bors try

gronerl:
bors try

gronerl:
This PR will allow ndarrays with any strides to be passed in to any backend. The performance backends will raise a warning instead of an exception.

The PR also unifies some of the code used across the different backends to determine and check optimal layouts.

PR #983, opened 2022-10-17
Fix typing: iterator.embedded

DropD:
## Description

Best effort to fix typing annotations in `iterator.embedded`.

## Requirements


src/functional/iterator/embedded.py
```diff
@@ -100,16 +96,21 @@ def __getitem__(self, indices: tuple[int, int]) -> int:
 # Offsets
 OffsetPart: TypeAlias = Tag | IntIndex
 CompleteOffset: TypeAlias = tuple[Tag, IntIndex]
-OffsetProviderElem: TypeAlias = Dimension | Connectivity
+OffsetProviderElem: TypeAlias = (
+    common.Dimension
+    | common.Connectivity
+    | NeighborTableOffsetProvider
+    | StridedNeighborOffsetProvider
```
havogt, line 103: Is this a mypy issue? We should not have to mention every implementation of a Connectivity here.

src/functional/iterator/embedded.py
```diff
@@ -44,10 +41,9 @@
 FieldIndexOrIndices: TypeAlias = FieldIndex | tuple[FieldIndex, ...]
 
 FieldAxis: TypeAlias = (
-    Dimension | Offset
+    common.Dimension | runtime.Offset
 )  # TODO Offset should be removed, is sometimes used for sparse dimensions
-TupleAxis: TypeAlias = NoneType
```
havogt, line 49: It was here because specifying tuple dimensions as `None` might not be the right long term choice. Therefore, it was increased to improve readability. I'd prefer rolling back all the changes regarding `TupleAxis`, but I'd like to hear your thoughts about it.

src/functional/iterator/embedded.py
```diff
@@ -479,19 +484,23 @@ def execute_shift(
...
         new_pos = copy.copy(pos)
         if is_int_index(value := new_pos[offset_implementation.value]):
             new_pos[offset_implementation.value] = value + index
         else:
             raise AssertionError()
         return new_pos
     else:
-        assert isinstance(offset_implementation, Connectivity)
+        assert isinstance(
+            offset_implementation, (NeighborTableOffsetProvider, StridedNeighborOffsetProvider)
```
havogt, line 502: This introduces strange coupling, if it's a mypy issue, we need to move it to an `is_connectivity()` function to have this check in only one place.

src/functional/iterator/embedded.py
```diff
@@ -798,13 +794,14 @@ def get_ordered_indices(
     axes: Iterable[Axis], pos: Mapping[Tag, FieldIndex | SparsePositionEntry]
 ) -> tuple[FieldIndex, ...]:
     res: list[FieldIndex] = []
-    sparse_position_tracker: dict[Tag, int] = {}
+    sparse_position_tracker: dict[Tag | int, int] = {}
     for axis in axes:
-        if _is_tuple_axis(axis):
+        if axis is None:
```
havogt, line 799: see comment above

src/functional/iterator/embedded.py
```diff
@@ -798,13 +794,14 @@ def get_ordered_indices(
     axes: Iterable[Axis], pos: Mapping[Tag, FieldIndex | SparsePositionEntry]
 ) -> tuple[FieldIndex, ...]:
     res: list[FieldIndex] = []
-    sparse_position_tracker: dict[Tag, int] = {}
+    sparse_position_tracker: dict[Tag | int, int] = {}
```
havogt, line 797: This change looks wrong to me, but not sure if I see everything.

src/functional/iterator/embedded.py
```diff
@@ -798,13 +794,14 @@ def get_ordered_indices(
...
     for axis in axes:
-        if _is_tuple_axis(axis):
+        if axis is None:
             res.append(slice(None))
         else:
             assert _is_field_axis(axis)
             assert axis.value in pos
+            assert isinstance(axis.value, Tag)
```
havogt, line 804: maybe with this change, the line at my previous comment should be rolled back?

havogt: or maybe the comment in line 45 needs to be resolved, not sure?

havogt (changes requested):
Just saw this PR and had some spare time in the train. Here are some comments.

src/functional/iterator/embedded.py
DropD: The problem is that `Connectivity` does not have some of the attributes that are accesed later. So either these attributes are at the wrong level or you actually wanted specifically a subclass that has them. I tried not to change the meaning of the code, only reflect the usage in the type annotations.

src/functional/iterator/embedded.py
DropD: I did not delete this for fun. Mypy was specifically complaining about `NoneType` and saying to replace it with `None`. But since it is used as a type later, that caused problems because `None` is not a type.

If there is a reason for `TupleAxis` to be there, the solution is probably a type ignore comment.

src/functional/iterator/embedded.py
DropD: It is an issue of attributes being accessed on things that are typed as `Connectivity` which are only defined in these subclasses. If you think these attributes should be defined at the base class level then we should go for that. Other solutions might be to introduce another level in the hierarchy in between, or the use of a protocol.

I didn't have the time to get into evaluating design changes.

src/functional/iterator/embedded.py
DropD: The change is driven by how the function is called elsewhere, again I simply traced the usage and made it explicit. If this looks wrong then the usage of this function is probably wrong in some places.

src/functional/common.py
```diff
@@ -74,6 +76,12 @@ def generate_operator(self, ir):
 class Connectivity(Protocol):
     max_neighbors: int
     has_skip_values: bool
+    origin_axis: Dimension
+
+    def mapped_index(
+        self, cur_index: int | np.integer, neigh_index: int | np.integer
+    ) -> Optional[int | np.integer]:
+        """Return neighbor index."""
```
havogt, line 84: we should split this into static part encoded in the type and the runtime part, i.e. the function, but I'll leave that for later (parts are designed in the embedded field view prototype)

src/functional/iterator/embedded.py
```diff
@@ -537,36 +536,42 @@ def execute_shift(
...
         assert None in new_entry
+        assert isinstance(index, int)
         for i, p in reversed(list(enumerate(new_entry))):
             # first shift applies to the last sparse dimensions of that axis type
             if p is None:
                 new_entry[i] = index
                 break
-        return pos | {tag: new_entry}
+        # the assertions above confirm pos is incomplete
+        # casting here to avoid duplicating work in a type guard
```
tehrengruber, line 546: Nitpicking: line length is 100 characters.

src/functional/iterator/embedded.py
```diff
@@ -478,7 +472,9 @@ def promote_scalars(val: CompositeOfScalarOrField):
     """Given a scalar, field or composite thereof promote all (contained) scalars to fields."""
     if isinstance(val, tuple):
         return tuple(promote_scalars(el) for el in val)
-    val_type = xtyping.infer_type(val)
+    val_type = cast(
+        npt.DTypeLike, xtyping.infer_type(val)
+    )  # TODO(havogt) how can we assert that something is DTypeLike?
```
tehrengruber, line 477:
```suggestion
    val_type = xtyping.infer_type(val)
    assert issubclass(val_type, (np.generic, int, float))
```
This should be fine for all practical purposes (not great though). Hopefully the cast is also superfluous after the assert.

tehrengruber (changes requested):
Small things only.

src/functional/iterator/embedded.py
havogt: done as typeguard

src/functional/iterator/embedded.py
```diff
@@ -961,33 +961,31 @@ def axes(self) -> tuple[()]:
         return ()
 
 
-def constant_field(value: Any, dtype: npt.DTypeLike = None) -> LocatedField:
+def constant_field(value: Any, dtype: Optional[npt.DTypeLike] = None) -> LocatedField:
     if dtype is None:
-        dtype = xtyping.infer_type(value)
+        dtype = xtyping.infer_type(value)  # type: ignore[assignment]
```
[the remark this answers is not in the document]
havogt: fix with the type guard

src/functional/iterator/embedded.py
```diff
@@ -961,33 +961,31 @@ def axes(self) -> tuple[()]:
...
 
 
-@dataclass
+@dataclasses.dataclass
 class ColumnDescriptor:
     axis: str
     col_range: range  # TODO(havogt) introduce range type that doesn't have step
 
 
+@dataclasses.dataclass
```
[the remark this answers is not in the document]
havogt: done

tehrengruber (approved):

havogt (approved):

PR #986, opened 2022-10-18
Add binop to past

nfarabullini:
Implementation of past.BinOp to e.g. accept operation within domain bounds:

```python
    def fieldop_domain(a: Field[[IDim, JDim], float64]) -> Field[[IDim, JDim], float64]:
        return a + a

    @program
    def program_domain(
        a: Field[[IDim, JDim], float64],
        lower_i: int64,
        upper_i: int64,
        lower_j: int64,
        upper_j: int64,
    ):
        fieldop_domain(a, out=a, domain={IDim: (lower_i - 1, upper_i), JDim: (lower_j, upper_j)})

    program_domain(a, lower_i, upper_i, lower_j, upper_j, offset_provider={})
```


samkellerhals:
Duplicated by #989

tehrengruber:
@samkellerhals Please push your changes to this PR as I already did a review here.

samkellerhals:
> @samkellerhals Please push your changes to this PR as I already did a review here.

All my changes are here now.

src/functional/ffront/func_to_past.py
```diff
@@ -129,3 +159,6 @@ def visit_UnaryOp(self, node: ast.UnaryOp) -> past.Constant:
     def visit_Constant(self, node: ast.Constant) -> past.Constant:
         symbol_type = symbol_makers.make_symbol_type_from_value(node.value)
         return past.Constant(value=node.value, type=symbol_type, location=self._make_loc(node))
+
+    def visit_Sub(self, node: ast.Sub, **kwargs) -> past.BinaryOperator:
+        return past.BinaryOperator.SUB
```
nfarabullini, line 164: I would move visit_Sub where the others are for more readability

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -43,7 +43,7 @@
 )
 
 
-@pytest.fixture(params=[roundtrip.executor, gtfn_cpu.run_gtfn])
```
nfarabullini, line 46: this line needs to be put back

src/functional/ffront/func_to_past.py
tehrengruber, line 164:
```suggestion
```

tehrengruber (changes requested):

src/functional/ffront/func_to_past.py
nfarabullini: actually remove as Till suggested below since it's a duplicate

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -1164,7 +1164,7 @@ def program_domain(
         lower_j: int64,
         upper_j: int64,
     ):
-        fieldop_domain(a, out=a, domain={IDim: (lower_i, upper_i), JDim: (lower_j, upper_j)})
+        fieldop_domain(a, out=a, domain={IDim: (lower_i - 1, upper_i), JDim: (lower_j, upper_j)})
```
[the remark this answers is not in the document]
tehrengruber: This would require implementing the calls to them and will be problematic with the c++ backend, as they might return floating point types for integer arguments (which shouldn't be the case, but we never implemented this).

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -118,6 +120,37 @@ def visit_TupleExpr(self, node: past.TupleExpr, **kwargs):
...
+    def _deduce_binop_type(
+        self,
+        node: past.BinOp,
+        *,
+        left: past.Expr,
+        right: past.Expr,
+        **kwargs,
+    ) -> Optional[ct.SymbolType]:
+        left_type = cast(ct.FieldType | ct.ScalarType, left.type)
+        right_type = cast(ct.FieldType | ct.ScalarType, right.type)
```
tehrengruber, line 132:
```suggestion
        logical_ops = {foast.BinaryOperator.BIT_AND, foast.BinaryOperator.BIT_OR}
        is_compatible = type_info.is_logical if node.op in logical_ops else type_info.is_arithmetic

        # check both types compatible
        for arg in (left, right):
            if not isinstance(arg.type, ct.ScalarType) or not is_compatible(arg.type):
                raise FieldOperatorTypeDeductionError.from_foast_node(
                    arg, msg=f"Type {arg.type} can not be used in operator `{node.op}`!"
                )
```
Why did you remove the check on the argument types?

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -1164,7 +1164,11 @@ def program_domain(
         lower_j: int64,
         upper_j: int64,
     ):
-        fieldop_domain(a, out=a, domain={IDim: (lower_i, upper_i), JDim: (lower_j, upper_j)})
+        fieldop_domain(
```
tehrengruber, line 1167: The specification of the backend is not correct in this file. Since `@program` has no backend specified it will fallback to the embedded one, while it should use `fieldview_backend` (which is only specified on the field operator). We should add a check for this at some point. If you specify the backend this test will fail for gtfn. Please use `pytest.skip` for that case and add another test that doesn't use `floordiv` and `pow` which are not (fully) supported in gtfn.

src/functional/iterator/embedded.py
```diff
@@ -1103,16 +1111,26 @@ def closure(
...
+                if isinstance(inp, LocatedField) or isinstance(inp, tuple):
+                    promoted_ins.append(inp)
+                else:
+                    inp_type = xtyping.infer_type(inp)
+                    if np.issubdtype(inp_type, np.number):
+                        promoted_ins.append(constant_field(inp))
+                    else:
+                        raise ValueError(
+                            "Expected a `Field` or a number (`float`, `np.int64`, ...)."
+                        )
```
tehrengruber, line 1125:
```python
Scalar: TypeAlias = SupportsInt | SupportsFloat | np.int32 | np.int64 | np.float32 | np.float64
CompositeOfScalarOrField = Scalar | LocatedField | tuple["CompositeOfScalarOrField", ...]
def promote_scalars(val: CompositeOfScalarOrField):
    """Given a scalar, field or composite thereof promote all (contained) scalars to fields."""
    if isinstance(val, tuple):
        return tuple(promote_scalars(el) for el in val)
    val_type = xtyping.infer_type(val)
    if np.issubdtype(val_type, np.number):
        promoted_ins.append(constant_field(val))
    else:
        raise ValueError(
            f"Expected a `Field` or a number (`float`, `np.int64`, ...), but got `val_type`."
        )
```
I just recognized this fails for nested tuples. Please add this (recursive) function at the top of the file (type aliases to the other ones) then you can simply do this:
```suggestion
            promoted_ins = [promote_scalars(inp) for inp in ins]
```
Salami slicing my own proposals now...

tehrengruber (changes requested):

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -118,6 +120,47 @@ def visit_TupleExpr(self, node: past.TupleExpr, **kwargs):
...
+
+        # check both types compatible
+        for arg in (left, right):
+            if not isinstance(arg.type, ct.ScalarType) or not is_compatible(arg.type):
+                raise ProgramTypeError.from_past_node(
+                    arg, msg=f"Type {arg.type} can not be used in operator `{node.op}`!"
+                )
+
+        left_type = cast(ct.FieldType | ct.ScalarType, left.type)
+        right_type = cast(ct.FieldType | ct.ScalarType, right.type)
```
tehrengruber, line 142:
```suggestion
        left_type = cast(ct.ScalarType, left.type)
        right_type = cast(ct.ScalarType, right.type)
```
Please check that this is still necessary or if mypy is able to deduce the type of `left` and `right` from the constraints in the loop above.

tehrengruber (approved):
Looks good.

PR #1002, opened 2022-10-20
Scan nested tuple args and return values

tehrengruber:
This PR adds support for nested tuple arguments and return values in scan operator. For convenience I also fixed calling field operators directly with tuple out arguments here (to keep the tests short).

```python
@scan_operator(axis=KDim, forward=forward, init=init, backend=fieldview_backend)
    def simple_scan_operator(
        carry: tuple[float, tuple[float, float]]
    ) -> tuple[float, tuple[float, float]]:
        return (carry[0] + 1.0, (carry[1][0] + 1.0, carry[1][1] + 1.0))

simple_scan_operator(out=(out1, (out2, out3)), offset_provider={})
```

```python
@scan_operator(axis=KDim, forward=forward, init=init, backend=fieldview_backend)
    def simple_scan_operator(carry: float, a: tuple[float, float]) -> float:
        return carry + a[0] + a[1]

simple_scan_operator((inp1, inp2), out=out, offset_provider={})
```


src/functional/ffront/common_types.py
```diff
@@ -59,8 +59,8 @@ class DimensionType(SymbolType):
 
 @dataclass(frozen=True)
 class OffsetType(SymbolType):
-    source: Optional[func_common.Dimension] = None
-    target: Optional[tuple[func_common.Dimension, ...]] = None
+    source: func_common.Dimension
+    target: tuple[func_common.Dimension] | tuple[func_common.Dimension, func_common.Dimension]
```
nfarabullini, line 63: Maybe I am missing something, but why do you have a tuple with only one argument?

src/functional/ffront/decorator.py
```diff
@@ -128,6 +128,23 @@ def is_cartesian_offset(o: FieldOffset):
...
+            yield from _field_constituents_shape_and_dims(el, el_type)
+    elif isinstance(arg_type, ct.FieldType):
+        dims = type_info.extract_dims(arg_type)
+        if hasattr(arg, "shape"):
+            assert len(arg.shape) == len(dims)
+            yield (arg.shape, dims)
+        else:
+            yield (None, dims)
+    else:
+        raise ValueError("Expected a `FieldType` or `TupleType` thereof.")
```
nfarabullini, line 145:
```suggestion
        raise ValueError("Expected `FieldType` or `TupleType`.")
```
This is more consistent with other error messages

src/functional/ffront/decorator.py
```diff
@@ -128,6 +128,23 @@ def is_cartesian_offset(o: FieldOffset):
     return deduced_grid_type if requested_grid_type is None else requested_grid_type
 
 
+def _field_constituents_shape_and_dims(
+    arg, arg_type: ct.FieldType | ct.ScalarType
```
nfarabullini, line 132: you need to add ct.TupleType as an option for arg_type

src/functional/ffront/foast_to_itir.py
```diff
@@ -467,19 +467,18 @@ def _visit_type_constr(self, node: foast.Call, **kwargs) -> itir.Literal:
         raise FieldOperatorLoweringError(f"Encountered a type cast, which is not supported: {node}")
 
     def _make_literal(self, val: Any, type_: ct.ScalarType) -> itir.Literal:
```
nfarabullini, line 469: same here, type_ should also include ct.TupleType as an option

src/functional/ffront/foast_to_itir.py
```diff
@@ -467,19 +467,18 @@ def _visit_type_constr(self, node: foast.Call, **kwargs) -> itir.Literal:
...
-            assert all(isinstance(type_, ct.ScalarType) for type_ in node.type.types)
+        # TODO(tehrengruber): check constant of this type is supported in iterator ir
+        if isinstance(type_, ct.TupleType):
             return im.make_tuple_(
-                *(self._make_literal(val, type_) for val, type_ in zip(node.value, node.type.types))
+                *(self._make_literal(val, type_) for val, type_ in zip(val, type_.types))
             )
-        raise FieldOperatorLoweringError(f"Unsupported scalar type: {node.type}")
+        elif isinstance(type_, ct.ScalarType):
+            typename = type_.kind.name.lower()
```
nfarabullini, line 476: Question: is name always an available within kind?

src/functional/ffront/past_to_itir.py
```diff
@@ -185,7 +190,21 @@ def _construct_itir_domain_arg(
...
+        out_field_types = type_info.primitive_constituents(out_field.type).to_list()
+        if any(
+            not isinstance(out_field_type, common_types.FieldType)
+            or out_field_type.dims != out_field_types[0].dims
+            for out_field_type in out_field_types
+        ):
+            raise AssertionError(
+                f"Expected constituents of `{out_field.id}` argument to be"
+                f" fields defined on the same dimensions. This error should be "
+                f" catched in type deduction already."
```
nfarabullini, line 203:
```suggestion
                f" caught in type deduction already."
```

tests/functional_tests/ffront_tests/test_type_deduction.py
```diff
@@ -75,14 +75,18 @@ def type_info_cases() -> list[tuple[Optional[ct.SymbolType], dict]]:
...
         if not isinstance(symbol_type, ct.CallableType)
     ]
 
+    IDim = Dimension("I")
+    JDim = Dimension("J")
+    KDim = Dimension("K", kind=DimensionKind.VERTICAL)
+
     bool_type = ct.ScalarType(kind=ct.ScalarKind.BOOL)
     float_type = ct.ScalarType(kind=ct.ScalarKind.FLOAT64)
     int_type = ct.ScalarType(kind=ct.ScalarKind.INT64)
```
nfarabullini, line 92: int32 the one mostly used in ICON and some times we had to change some things to account for that. Could you test this case as well?

src/functional/ffront/past_to_itir.py
```diff
@@ -95,9 +95,14 @@ def _gen_size_params_from_program(self, node: past.Program):
         """Generate symbols for each field param and dimension."""
         size_params = []
         for param in node.params:
-            if isinstance(param.type, common_types.FieldType):
-                for dim_idx in range(0, len(param.type.dims)):
+            if type_info.is_field_type_or_tuple_of_field_type(param.type):
+                fields_dims: list[list[Dimension]] = (
+                    type_info.primitive_constituents(param.type).getattr("dims").to_list()
+                )
+                assert all(field_dims == fields_dims[0] for field_dims in fields_dims)
```
nfarabullini, line 102:
```suggestion
                assert all(field_dim == fields_dims[0] for field_dim in fields_dims)
```

src/functional/ffront/past_to_itir.py
```diff
@@ -95,9 +95,14 @@ def _gen_size_params_from_program(self, node: past.Program):
...
+                for dim_idx in range(0, len(fields_dims[0])):
```
nfarabullini, line 103:
```suggestion
                for dim_idx in range(len(fields_dims[0])):
```

src/functional/ffront/type_info.py
```diff
@@ -75,6 +76,32 @@ def constituents_yielder(symbol_type: ct.SymbolType):
...
+                apply_to_primitive_constituents(
+                    el, f, _path=(*_path, i), with_path_arg=with_path_arg
+                )
+                for i, el in enumerate(symbol_type.types)
+            ]
+        )
+    if with_path_arg:
+        return f(symbol_type, _path)
+    else:
+        return f(symbol_type)
```
nfarabullini, line 102: I don't like f here, can you rename it?

src/functional/ffront/type_info.py
```diff
@@ -75,6 +76,32 @@ def constituents_yielder(symbol_type: ct.SymbolType):
...
+
+
 def extract_dtype(symbol_type: ct.SymbolType) -> ct.ScalarType:
     """
     Extract the data type from ``symbol_type`` if it is one of FieldType or ScalarType.
```
nfarabullini, line 107:
```suggestion
    Extract the data type from ``symbol_type`` if it is either FieldType or ScalarType.
```

src/functional/ffront/foast_to_itir.py
tehrengruber: Good catch.

src/functional/ffront/foast_to_itir.py
tehrengruber: Yes, if you follow the class hierarchy `ScalarKind` is a regular Python `IntEnum` and here is the corresponding entry in the documentation: https://docs.python.org/3/library/enum.html#programmatic-access-to-enumeration-members-and-their-attributes

src/functional/ffront/past_to_itir.py
tehrengruber: `field_dims` is a list of dimensions (see the `list[list[Dimension]]` annotation above), hence the name. Contrary to that _field**s**\_dims_ is a list of dimensions of all the field**s**.

src/functional/ffront/past_to_itir.py
tehrengruber: Hehe, good catch again :-)

src/functional/ffront/type_info.py
tehrengruber: Are you happy with `fun`?

tests/functional_tests/ffront_tests/test_type_deduction.py
tehrengruber: The actual type here doesn't matter, I could have equally used just `FLOAT64` and `FLOAT32`. The argument checking is and should never be dependent on what `ScalarType` is used in the signature. What matters is that different types. If in the implementation I screw up the order of the arguments, but my test would be for
```
def f(a: float, b: float):
    ...

f(some_float, some_float)
```
then the test would pass. Using different argument types however helps:
```
def f(a: int, b: float):
    ...

f(some_float, some_int)
```

src/functional/ffront/type_info.py
nfarabullini: yes

src/functional/ffront/common_types.py
nfarabullini: gotcha

src/functional/ffront/foast_to_itir.py
nfarabullini: aah ok thanks

tests/functional_tests/ffront_tests/test_type_deduction.py
nfarabullini: :+1:

src/functional/ffront/decorator.py
```diff
@@ -123,6 +123,23 @@ def is_cartesian_offset(o: FieldOffset):
...
+            yield from _field_constituents_shape_and_dims(el, el_type)
+    elif isinstance(arg_type, ct.FieldType):
+        dims = type_info.extract_dims(arg_type)
+        if hasattr(arg, "shape"):
+            assert len(arg.shape) == len(dims)
+            yield (arg.shape, dims)
+        else:
+            yield (None, dims)
+    else:
+        raise ValueError("Expected `FieldType` or `TupleType` thereof.")
```
nfarabullini, line 140:
```suggestion
        raise ValueError("Expected either `FieldType` or `TupleType`.")
```

src/functional/ffront/past_to_itir.py
```diff
@@ -95,9 +95,14 @@ def _gen_size_params_from_program(self, node: past.Program):
         """Generate symbols for each field param and dimension."""
         size_params = []
         for param in node.params:
-            if isinstance(param.type, common_types.FieldType):
-                for dim_idx in range(0, len(param.type.dims)):
+            if type_info.is_field_type_or_tuple_of_field_type(param.type):
+                fields_dims: list[list[Dimension]] = (
+                    type_info.primitive_constituents(param.type).getattr("dims").to_list()
+                )
```
nfarabullini, line 101:
```suggestion
```
Can you remove this set of brackets?

src/functional/ffront/type_info.py
```diff
@@ -75,9 +76,35 @@ def constituents_yielder(symbol_type: ct.SymbolType):
     return xiter(constituents_yielder(symbol_type))
 
 
+def apply_to_primitive_constituents(
+    symbol_type: ct.SymbolType, fun: Callable[[ct.SymbolType], ct.SymbolType], with_path_arg=False, _path=()
+):
```
nfarabullini, line 84: add return type?

src/functional/ffront/type_info.py
```diff
@@ -514,11 +544,11 @@ def function_signature_incompatibilities_fieldop(
 def function_signature_incompatibilities_scanop(
     scanop_type: ct.ScanOperatorType, args: list[ct.SymbolType], kwargs: dict[str, ct.SymbolType]
 ) -> Iterator[str]:
-    if not all(isinstance(arg, ct.FieldType) for arg in args):
-        yield "Arguments to scan operator must be fields."
+    if not all(is_field_type_or_tuple_of_field_type(arg) for arg in args):
+        yield "Arguments to scan operator must be fields or tuples thereof."
```
nfarabullini, line 551:
```suggestion
        yield "Arguments to scan operator must be either fields or tuples."
```

src/functional/ffront/type_info.py
```diff
@@ -528,13 +558,30 @@ def function_signature_incompatibilities_scanop(
         yield f"Scan operator takes {len(scanop_type.definition.args)-1} arguments, but {len(args)} were given."
         return
 
+    promoted_args = []
+    for i, scan_pass_arg in enumerate(scanop_type.definition.args[1:]):
+        # Helper function that given a scalar type in the signature of the scan
+        # pass return a field type with that dtype and the dimensions of the
+        # corresponding field type in the requested `args` type. Defined here
+        # as we capture `i`.
+        def _as_field(dtype: ct.ScalarType, path: tuple[int, ...]):
```
nfarabullini, line 570: I don't understand why the function definition is inside the for loop

src/functional/ffront/past_to_itir.py
nfarabullini: oki doki

src/functional/ffront/type_info.py
tehrengruber: The *thereof* is important to signify that the tuple must contain fields. I don't think the *either* is necessary and makes it only more verbose. If you think it's easier to understand with the either I can add it.

src/functional/ffront/type_info.py
tehrengruber: I tried to explain that in the comment. The function uses the iteration variable `i` inside. I could write a more complicated function outside, but this felt easier to me. I am open to suggestions, as I don't particularly like it either.

src/functional/ffront/type_info.py
nfarabullini: it is really not a big deal :)

nfarabullini (approved):

PR #1006, opened 2022-10-24
Allow bools in scan init

havogt:


src/functional/ffront/type_info.py
```diff
@@ -97,27 +97,10 @@ def extract_dtype(symbol_type: ct.SymbolType) -> ct.ScalarType:
...
-    ---------
-    >>> is_arithmetic(ct.ScalarType(kind=ct.ScalarKind.FLOAT64))
-    True
-    >>> is_arithmetic(ct.ScalarType(kind=ct.ScalarKind.BOOL))
-    False
-    >>> is_arithmetic(ct.FieldType(dims=[], dtype=ct.ScalarType(kind=ct.ScalarKind.INT32)))
-    True
-    """
-    if extract_dtype(symbol_type).kind in [
-        ct.ScalarKind.INT32,
```
havogt, line 114: There was a bug here: ct.ScalarKind.INT was missing. Now that I look at my PR, I would just revert and implement is_arithmetic as
`return is_floating_point(symbol_type) or is_integral(symbol_type)` 
let me know what you prefer.

havogt: ok, maybe not a bug, but at least a inconsistency...

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -211,7 +211,7 @@ def visit_ScanOperator(self, node: foast.ScanOperator, **kwargs) -> foast.ScanOp
             )
         new_init = self.visit(node.init, **kwargs)
         if not all(
-            type_info.is_arithmetic(type_)
+            type_info.is_arithmetic(type_) or type_info.is_logical(type_)
```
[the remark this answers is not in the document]
tehrengruber: Apparently none: `True ** True == 1.0`, so I think it makes sense to add them to `is_arithmetic`.

src/functional/ffront/type_info.py
tehrengruber: `return is_floating_point(symbol_type) or is_integral(symbol_type)` looks good.

tehrengruber (approved):

havogt:
hmm, looks like I am a bit undecided. since there were tests that check that some operations are not allowed on bools and since using these is most likely a user mistake, I implemented no change in behavior, but introduced a new category is_number.

tehrengruber (approved):

Issue #1014, opened 2022-10-26
Functional: `FixMissingLocations` sometimes doesn't copy `end_lineno`


BenWeber42:
Test case from: https://github.com/BenWeber42/gt4py/commit/b85013a26eec495ac3592b4ad9f8a9c52aeeb91c

```python
def test_consistent_end_lineno_bug():
    """
    If a parent node has attributes `end_lineno` and `end_col_offset`, then
    its child nodes have those too.
    """
    astree = get_fixed_ast(
        """
        5 + 4
        """
    )

    assert hasattr(astree.body[0].value, "end_lineno")
    assert hasattr(astree.body[0].value, "end_col_offset")
    assert hasattr(astree.body[0].value.op, "end_lineno")
    assert hasattr(astree.body[0].value.op, "end_col_offset")
```

`FixMissingLocations` will only copy `end_lineno` and `end_col_offset` to a child node if it already has those attributes. But many nodes don't have any location information attributes. So then only `lineno` and `col_offset` are copied, even though `end_lineno` and `end_col_offset` could be copied as well.

See here: https://github.com/GridTools/gt4py/blob/c63c11eced04a60ef6584788a92b3dc2851a34af/src/functional/ffront/ast_passes/fix_missing_locations.py#L46-L49

Issue #1013, opened 2022-10-26
Functional: Weird behavior of `ast.parse` breaks FixMissingLocations


BenWeber42:
Test case from: https://github.com/BenWeber42/gt4py/commit/b85013a26eec495ac3592b4ad9f8a9c52aeeb91c

```python
def get_fixed_ast(code: str) -> ast.AST:
    python_ast = ast.parse(textwrap.dedent(code))
    python_ast.lineno = 1
    python_ast.end_lineno = 999
    python_ast.col_offset = 0
    python_ast.end_col_offset = 999
    return FixMissingLocations.apply(python_ast)


def test_reused_op_nodes():
    """
    Python's `ast.parse` will reuse the same nodes multiple times in the AST sometimes
    (e.g., for operators like `ast.Add`). In this case, all those shared nodes
    will share the same location information which can be plain wrong in some cases.
    """

    astree = get_fixed_ast(
        """
        a = (
            1
            /
            2
            /
            3
            /
            4
        )
        """
    )

    match astree:
        case ast.Module(
            body=[
                ast.Assign(
                    value=ast.BinOp(
                        left=ast.BinOp(
                            left=ast.BinOp(op=ast.Div() as first), op=ast.Div() as second
                        ),
                        op=ast.Div() as third,
                    )
                )
            ]
        ):
            pass
        case _:
            assert False

    assert first is not second
    assert second is not third
    assert third is not first
```

For some reason, Python's `ast.parse` will reuse the same `ast.Div` instances throughout the whole tree. This will break the `FixMissingLocations` pass (not really break it, but it will produce wrong results). It's not necessarily an issue in `FixMissingLocations`, but maybe a pre-condition should be that no nodes are reused within the same tree and a new pass should be added that makes sure this pre-condition is established before.

tehrengruber:
Fixed in #1020 by doing a copy (fine for leaf nodes).

PR #1020, opened 2022-10-26
Bugfixes for FixMissingLocations pass

### LINKED ISSUE #1015: Functional: `FixMissingLocations` sometimes copies location information from wrong nodes
Testcase from: https://github.com/BenWeber42/gt4py/commit/b85013a26eec495ac3592b4ad9f8a9c52aeeb91c

```python
def test_sibling_nesting_bug():
    """
    Test for a bug where the location information sometimes wasn't taken from the
    parent node, but from a nested sibling node.
    """
    astree = get_fixed_ast(
        """
        a = (
            1 * 2
            + 3
        )
        """
    )


    assert (
        isinstance(astree.body[0].value.left.op, ast.Mult)
        and astree.body[0].value.left.op.end_lineno == 3
    )
    assert isinstance(astree.body[0].value.op, ast.Add) and astree.body[0].value.op.end_lineno == 4
```
(This testcase will only trigger the bug, after #1014 has been fixed)

`FixMissingLocations` doesn't use a stack to track parent nodes, but a field instead. Because of this, nested children of sibling nodes will overwrite `_parent_node`. Then other children will copy the location information of nested children of their sibling nodes instead of their parent node.

https://github.com/GridTools/gt4py/blob/c63c11eced04a60ef6584788a92b3dc2851a34af/src/functional/ffront/ast_passes/fix_missing_locations.py#L18-L51

tehrengruber:
Fixes #1015, #1014, #1013. See the issues for examples.


src/functional/ffront/ast_passes/fix_missing_locations.py
```diff
@@ -29,23 +30,30 @@ class FixMissingLocations(ast.NodeTransformer):
...
+            if hasattr(self._parent_node[-1], "end_col_offset"):
+                node.end_col_offset = self._parent_node[-1].end_col_offset
 
-        return super().generic_visit(node)
+        result = super().generic_visit(node)
+
+        if has_location:
+            self._parent_node.pop()
+
+        return result
```
BenWeber42, line 59: I feel like it's more readable like this:

```python
        if not hasattr(node, "lineno"):
            node = copy.deepcopy(node)
            parent_node = self._parent_nodes[-1]

            node.lineno = parent_node.lineno
            node.col_offset = parent_node.col_offset

            # the end positions are optional according to
            #  https://docs.python.org/3/library/ast.html#ast.AST.end_col_offset
            if hasattr(self.parent_node, "end_lineno"):
                node.end_lineno = parent_node.end_lineno
            if hasattr(self.parent_node, "end_col_offset"):
                node.end_col_offset = parent_node.end_col_offset

            return super().generic_visit(node)

        self._parent_nodes.append(node)
        result = super().generic_visit(node)
        self._parent_nodes.pop()

        return result
```

src/functional/ffront/ast_passes/fix_missing_locations.py
```diff
@@ -29,23 +30,30 @@ class FixMissingLocations(ast.NodeTransformer):
...
     def generic_visit(self, node: ast.AST) -> ast.AST:
-        if getattr(node, "lineno", None):
-            self._parent_node = node
+        has_location = hasattr(node, "lineno")
+        if has_location:
+            self._parent_node.append(node)
         else:
-            node.lineno = self._parent_node.lineno
-            node.col_offset = self._parent_node.col_offset
+            node = copy.deepcopy(node)
```
BenWeber42, line 44: `deepcopy` will preserve duplicates, so there's no advantage using `deepcopy` over shallow `copy.copy` (`deepcopy` will also change worst case runtime from O(n) to O(n^2)).

I also think it would be better to have a separate `NodeDeduplication` pass that solves the problem in a general manner. This pass would run before `FixMissingLocations` and establish the pre-condition that there are no duplicate (by identity) nodes in the tree.

BenWeber42 (commented):
It works, but could be improved imho.

BenWeber42 (approved):

src/functional/ffront/ast_passes/fix_missing_locations.py
tehrengruber: I switched to a shallow copy.

I see the point of a `NodeDeduplication`, but since this is the only pass that requires it and since it is good practice to copy a node if it is modified based on a property of a parent node I would like to leave it as is.

PR #1021, opened 2022-10-26
Functional: Avoid duplicating expressions in SimpleAssignTargetPass

### LINKED ISSUE #1016: Functional: Avoid duplicating expressions in `SimpleAssignTargetPass`
The `SimpleAssignTargetPass` rewrites assignments such as `a = b = 0` into single assignments. However, in the process, the _right-most hand-side_ will get duplicated multiple times.

This:
```python
<<first assign target>> = <<second assign target>> = <<very big expression>>
```

Will become:

```python
<<first assign target>> = <<very big expression>>
<<second assign target>> = <<very big expression>>
```

To avoid duplicating  `<<very big expression>>`, we could rewrite it instead to:

```python
tmp = <<very big expression>>
<<first assign target>> = tmp
<<second assign target>> = tmp
```
(where `tmp` is a fresh variable name)

https://github.com/GridTools/gt4py/blob/8e9ac25300ac8d4eda920ee206b046ae9edf9d30/src/functional/ffront/ast_passes/simple_assign.py#L50-L80

tehrengruber:
Fixes #1016


src/functional/ffront/ast_passes/simple_assign.py
```diff
@@ -67,17 +67,35 @@ class SingleAssignTargetPass(NodeYielder):
...
+            synthetic_assign.targets = [synthetic_target]
+            yield synthetic_assign
+
+            for target in node.targets:
+                new_assign = copy.copy(node)
+                new_assign.targets = [target]
+                new_assign.value = ast.Name(id=synthetic_target.id, ctx=ast.Load())
+                yield new_assign
+        else:
+            yield node
```
BenWeber42, line 98:
```suggestion
        if len(node.targets) == 1:
            yield node
            return

        synthetic_target = self._unique_symbol_name()
        ast.copy_location(synthetic_target, node)
        synthetic_assign = copy.copy(node)
        synthetic_assign.targets = [synthetic_target]
        yield synthetic_assign

        for target in node.targets:
            new_assign = copy.copy(node)
            new_assign.targets = [target]
            new_assign.value = ast.Name(id=synthetic_target.id, ctx=ast.Load())
            yield new_assign
```

I guess this is more a matter of taste, but I prefer to first read about the special cases and have the general/more complicated case later.

src/functional/ffront/ast_passes/simple_assign.py
```diff
@@ -67,17 +67,35 @@ class SingleAssignTargetPass(NodeYielder):
...
         return (a, b)
 
     """
 
+    unique_name_id: int = 0
+
+    def _unique_symbol_name(self) -> ast.Name:
+        name = ast.Name(id=f"__sat_tmp{self.unique_name_id}", ctx=ast.Store())
+        self.unique_name_id += 1
+        return name
```
BenWeber42, line 82: I guess it works for now, but I feel like it might be worth considering having some sort of infrastructure to generate fresh variable names (that are guaranteed not to be used anywhere else) and maybe have some sort of way to distinguish between user-provided names and newly generated names.

BenWeber42 (approved):

PR #1029, opened 2022-11-03
Foast pretty printer

tehrengruber:
Pretty printer for foast nodes.
```python
>>> @field_operator
... def field_op(a: Field[..., int]):
...     return a+1
>>> print(pretty_format(field_op.foast_node))
@field_operator
def field_op(a):
  return a + 1
```


src/functional/ffront/foast_pretty_printer.py
```diff
@@ -0,0 +1,195 @@
...
+    def visit_Compare(self, node: foast.Compare, **kwargs) -> str:
+        left = self._parenthesize(node.left, node, Group.LEFT)
+        right = self._parenthesize(node.right, node, Group.RIGHT)
+        return f"{left} {node.op} {right}"
+
+    def visit_TernaryExpr(self, node: foast.TernaryExpr, **kwargs) -> str:
+        cond = self.visit(node.condition)
+        true_expr = self._parenthesize(node.true_expr, node, Group.LEFT)
+        false_expr = self._parenthesize(node.false_expr, node, Group.RIGHT)
+        return f"{true_expr} if {cond} else {false_expr}"
```
nfarabullini, line 116: can you move these function after all as_fmt calls, i.e. after the Return definition

src/functional/ffront/foast_pretty_printer.py
```diff
@@ -0,0 +1,195 @@
...
+            def ${id}(${', '.join(params_annotated)})${return_type}:
+              ${'\\n  '.join(body)}
+            """
+        ).strip()
+    )
+
+    def visit_FunctionDefinition(self, node: foast.FunctionDefinition, **kwargs):
+        params = self.visit(node.params)
+        types = [
+            str(param.type) if not isinstance(param.type, common_types.DeferredSymbolType) else None
```
nfarabullini, line 136: I like the condition to evaluate to True:
```suggestion
            None if isinstance(param.type, common_types.DeferredSymbolType) else str(param.type)
```
Also, are you sure that param will always have a type?

src/functional/ffront/foast_pretty_printer.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def _grouping(self, node: foast.LocatedNode) -> Group:
+        prop_id = _property_identifier(node)
+        return Group.RIGHT if prop_id in RIGHT_GROUPING else Group.LEFT
+
+    def _parenthesize(
+        self,
+        inner_node: foast.LocatedNode,
+        outer_node: foast.LocatedNode,
+        group: Group | None = None,
```
nfarabullini, line 169: None = None ??

tests/functional_tests/ffront_tests/test_foast_pretty_printer.py
```diff
@@ -0,0 +1,74 @@
...
+        def bar(inp: Field[[], int64]) -> Field[[], int64]:
+          return foo(inp)
+        """
+    ).strip()
+
+    assert pretty_format(bar.foast_node) == expected
+
+
+def test_scanop():
+    from functional.ffront.fbuiltins import int64
```
nfarabullini, line 58: can you place this line outside the functions?

src/functional/ffront/foast_pretty_printer.py
```diff
@@ -0,0 +1,196 @@
...
+# see https://docs.python.org/3/reference/expressions.html#operator-precedence
+PRECEDENCE: Final[dict[PropertyIdentifier, int]] = {
+    # Binding or parenthesized expression, list display, dictionary display, set display
+    foast.TupleExpr: 18,
+    # Subscription, slicing, call, attribute reference
+    foast.Subscript: 17,
+    foast.Call: 17,
+    foast.Attribute: 17,
+    # Await expression
+    #  not used
```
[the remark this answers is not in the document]
nfarabullini: i.e., the comments on line 23 and 24

src/functional/ffront/foast_pretty_printer.py
tehrengruber: This is a stylistic question, but I would argue that the reader of the code should be presented with the common case first for conditionals, which is to stringify the type here. For if statements, contrary to conditionals, the reader is first presented with the condition so one can easily follow along and I always suggest avoiding unnecessary negations. Not so much for conditionals where one is first presented with a value only after which the condition follows. I personally dislike the choice of the conditional syntax in Python for exactly that reason.

nfarabullini (approved):

nfarabullini (approved):

PR #1030, opened 2022-11-03
Scalar if

tehrengruber:
```python
@field_operator
def simple_if(a: Field[[IDim, JDim], float64], b: Field[[IDim, JDim], float64], condition: bool):
    if condition:
        result = a
    else:
        result = b
    return result
```

WIP just a horizontal slice right now.


tehrengruber:
Superseeded by #972

PR #1024, opened 2022-10-27
Update eve

jdahm:
## Description

Integrates the `attrs`-based version of `eve` used on the `functional` branch into `gt4py` and `gtc`.

PR #1032, opened 2022-11-04
Parallelize tox tests

jdahm:
## Description

Runs tox tests in parallel using `pytest-xdist`. To do that, this PR also removes the `test_cpp_regression.py` and build files that use the test headers from GridTools C++.

Blocked by #1024. It technically does not require those changes, but my editor is configured to use those dependencies at the moment, and I don't want to switch things over ;)


jdahm:
This worked on my local copy, hmm. Will retry after the eve transition.

jdahm:
bors try

.gitlab-ci.yml
```diff
@@ -72,6 +72,7 @@ test py38:
     SLURM_CONSTRAINT: gpu
     SLURM_JOB_NUM_NODES: 1
     SLURM_TIMELIMIT: 120
+    NUM_PROCESSES: 3
```
gronerl, line 75: What does this cause in the gitlab-ci? Does this PR have any impact on gitlab CI at all? It's all going to end up on the same GPU, though it could be shared I guess.

jdahm: Right. I believe the API calls will time-share the GPU, so in theory it should be fine. It looks like it passes in practice too :-)

jdahm:
🎉

- tox test time went from `70` minutes on the dual-core GitHub runner to `35`, a perfect speedup.

- GitLab CI time on Piz Daint went from `100` minutes to `35` minutes.

requirements-dev.txt
```diff
@@ -21,7 +21,7 @@ pytest~=6.1
 pytest-cache>=1.0
 pytest-cov>=2.8
 pytest-factoryboy>=2.0
-pytest-xdist[psutil]>=2.2
+pytest-xdist[psutil]>=2.4
```
jdahm, line 24: Version `2.4` was the first one that supported Python `3.10`, so I switched.

.gitlab-ci.yml
jdahm: We could use `auto` here, but I'm afraid of `~24` compilation processes potentially running over the memory limit.

Do you think we should go a bit higher?

gronerl:
Wow those are nice speedups!

gronerl:
I think `auto` will pick physical cores, so would be 12. But with tox tests still running 35 minutes, the usefulness of shorter run times becomes less, even if scaling stays perfect. If you have time to play with it a bit today, I'm happy to merge whatever you pick tomorrow.

.gitlab-ci.yml
gronerl: I think `auto` will pick physical cores, so would be 12. But with tox tests still running 35 minutes, the usefulness of shorter run times becomes less, even if scaling stays perfect. If you have time to play with it a bit today, I'm happy to merge whatever you pick tomorrow.

gronerl:
bors try

gronerl (approved):
Looks like scaling on daint-cpu is a lot worse with 4 processes already. Small gains still though.

PR #1033, opened 2022-11-07
Functional: implicit broadcast 0d field

nfarabullini:
Use scalar as input and perform implicit broadcasting to return field of empty dimensions
```python
@field_operator
def fieldop_implicit_broadcast(
    zero_dim_inp: Field[[], float], inp: Field[[IDim], float], scalar: float
) -> Field[[IDim], float]:
    return inp + zero_dim_inp * scalar

fieldop_implicit_broadcast(1.0, inp, 1.0)
```


src/functional/ffront/func_to_past.py
```diff
@@ -37,6 +37,8 @@ class ProgramParser(DialectParser[past.Program]):
     def _postprocess_dialect_ast(
         cls, output_node: past.Program, closure_vars: dict[str, Any], annotations: dict[str, Any]
     ) -> past.Program:
+        if "return" in annotations and not isinstance(None, annotations["return"]):
```
tehrengruber, line 40: This should be done in the PAST type deduction.

src/functional/ffront/type_info.py
```diff
@@ -171,6 +171,32 @@ def is_integral(symbol_type: ct.SymbolType) -> bool:
...
+    """
+    if not isinstance(symbol_type, ct.ScalarType):
+        return False
+    return symbol_type.kind in [
+        ct.ScalarKind.INT,
+        ct.ScalarKind.INT32,
+        ct.ScalarKind.INT64,
+        ct.ScalarKind.FLOAT32,
+        ct.ScalarKind.FLOAT64,
+    ]
```
tehrengruber, line 197:
```suggestion
    if not isinstance(symbol_type, ct.ScalarType):
        return False
    return is_arithmetic(symbol_type)
```
To avoid code duplication.

src/functional/ffront/type_info.py
```diff
@@ -286,6 +312,37 @@ def is_concretizable(symbol_type: ct.SymbolType, to_type: ct.SymbolType) -> bool
     return False
 
 
+def _is_empty_field(field: ct.SymbolType) -> bool:
```
tehrengruber, line 315:
```suggestion
def _is_zero_dimensional_field(field: ct.SymbolType) -> bool:
```
Same for the rest.

src/functional/ffront/type_info.py
```diff
@@ -286,6 +312,37 @@ def is_concretizable(symbol_type: ct.SymbolType, to_type: ct.SymbolType) -> bool
...
+    return isinstance(field, ct.FieldType) and len(field.dims) == 0
+
+
+def is_not_empty_field_compatible(a_arg: ct.SymbolType, b_arg: ct.SymbolType) -> bool:
+    """
+    Check if first argument is an empty field and whether second is either another emtpy field or a scalar.
```
tehrengruber, line 321: Have to think about this, but the description is very technical and doesn't enhance what is understood by simply reading the code.

src/functional/ffront/type_info.py
```diff
@@ -286,6 +312,37 @@ def is_concretizable(symbol_type: ct.SymbolType, to_type: ct.SymbolType) -> bool
...
+
+    In both cases, arguments dtypes have to be the same
+    Examples:
+    ---------
+    >>> is_not_empty_field_compatible(
```
tehrengruber, line 326: When all examples return `True` they give little insight into what the function does.

src/functional/ffront/type_info.py
```diff
@@ -501,14 +558,20 @@ def function_signature_incompatibilities(
...
+    args: list[ct.SymbolType],
+    kwargs: dict[str, ct.SymbolType],
 ) -> Iterator[str]:
 
     # check positional arguments
     if len(func_type.args) != len(args):
         yield f"Function takes {len(func_type.args)} argument(s), but {len(args)} were given."
     for i, (a_arg, b_arg) in enumerate(zip(func_type.args, args)):
-        if a_arg != b_arg and not is_concretizable(a_arg, to_type=b_arg):
+        if (
```
tehrengruber, line 570: The mechanism you implement in this PR is only valid for Field Operators, so it belongs into `function_signature_incompatibilities_fieldop`. Not sure how this will turn out, but I suggest to try an approach that resembles what one would do if one would implement this not on the code generation level, but in regular Python: If I get an argument that is a scalar and expect zero dimensional field, I convert the scalar into a zero dimensional field. Let me know if this is too cryptic and needs further explanation.

tehrengruber (commented):
Partial review only.

src/functional/ffront/type_info.py
nfarabullini: it is not exactly linear as `is_arithmetic`returns true for emtpy fields

src/functional/ffront/type_info.py
nfarabullini: this name is a bit too long IMO, how about `_is_zero_dim_field`?

src/functional/ffront/type_info.py
nfarabullini: ok, I can add a couple when it outputs False

src/functional/ffront/func_to_past.py
nfarabullini: it is not possible to do it in the type_deduction as the return type there is the deduced type. Additionally, checks to match the specified and deduced type for foast were done in func_to_foast

src/functional/ffront/type_info.py
nfarabullini: this is a fair point, I will add a TODO here and change it the way you suggested. I want to do these edits in a separate PR

src/functional/ffront/type_info.py
nfarabullini: the latter part of your suggestion is already done in `return_type_fieldop` as a general case for scalars

src/functional/ffront/type_info.py
nfarabullini: I made it slightly less technical

src/functional/ffront/type_info.py
```diff
@@ -286,6 +307,45 @@ def is_concretizable(symbol_type: ct.SymbolType, to_type: ct.SymbolType) -> bool
     return False
 
 
+def _is_zero_dim_field(field: ct.SymbolType) -> bool:
+    return isinstance(field, ct.FieldType) and field.dims != Ellipsis and len(field.dims) == 0
+
+
+def is_empty_field_compatible(a_arg: ct.SymbolType, b_arg: ct.SymbolType) -> bool:
```
tehrengruber, line 314:
```suggestion
def is_zero_dim_field_compatible(a_arg: ct.SymbolType, b_arg: ct.SymbolType) -> bool:
```
As discussed.

tests/functional_tests/ffront_tests/test_past_parsing.py
```diff
@@ -266,3 +267,12 @@ def domain_format_6_program(in_field: Field[[IDim], float64]):
     assert exc_info.match("Invalid call to `domain_format_6`")
 
     assert re.search("Empty domain not allowed.", exc_info.value.__cause__.args[0]) is not None
+
+
+def test_implicit_broadcast(empty_identity_def):
```
tehrengruber, line 272: Looks like this is a left-over from the previous state of the PR. There is no implicit broadcast for the argument here. I think this can be safely removed.

src/functional/ffront/type_info.py
```diff
@@ -584,7 +662,8 @@ def _as_field(dtype: ct.ScalarType, path: tuple[int, ...]) -> ct.FieldType:
 def function_signature_incompatibilities_program(
     program_type: ct.ProgramType, args: list[ct.SymbolType], kwargs: dict[str, ct.SymbolType]
 ) -> Iterator[str]:
-    yield from function_signature_incompatibilities_func(program_type.definition, args, kwargs)
+    args_new = check_zero_dims_fields(args, program_type)
```
tehrengruber, line 665:
```suggestion
    new_args = check_zero_dims_fields(args, program_type)
```

src/functional/ffront/type_info.py
```diff
@@ -528,7 +605,8 @@ def function_signature_incompatibilities_func(
 def function_signature_incompatibilities_fieldop(
     fieldop_type: ct.FieldOperatorType, args: list[ct.SymbolType], kwargs: dict[str, ct.SymbolType]
 ) -> Iterator[str]:
-    yield from function_signature_incompatibilities_func(fieldop_type.definition, args, kwargs)
+    args_new = check_zero_dims_fields(args, fieldop_type)
```
tehrengruber, line 608:
```suggestion
    new_args = check_zero_dims_fields(args, fieldop_type)
```

tehrengruber (changes requested):
Only cosmetics. The implementation looks solid. Please update the example in the PR description (e.g. use the core part of the `test_implicit_broadcast_mixed_dim` test).

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -1277,6 +1277,42 @@ def return_undefined():
...
+    def fieldop_implicit_broadcast_2(inp: Field[[IDim], float]) -> Field[[IDim], float]:
+        fi = fieldop_implicit_broadcast(1.0, inp, 1.0)
+        return fi
+
+    @program
+    def program_implicit_broadcast_2(inp: Field[[IDim], float], out: Field[[IDim], float]):
+        fieldop_implicit_broadcast_2(inp, out=out)
+
+    program_implicit_broadcast_2(input1, out, offset_provider={})
+    assert np.allclose(out, np.asarray(inp) * 2)
```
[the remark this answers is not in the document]
nfarabullini: the program is there to make sure that everything foes smoothly on that end as well

src/functional/ffront/type_info.py
```diff
@@ -286,6 +307,60 @@ def is_concretizable(symbol_type: ct.SymbolType, to_type: ct.SymbolType) -> bool
     return False
 
 
+def _is_zero_dim_field(field: ct.SymbolType) -> bool:
+    return isinstance(field, ct.FieldType) and field.dims != Ellipsis and len(field.dims) == 0
+
+
+def is_zero_dim_field_compatible(a_arg: ct.SymbolType, b_arg: ct.SymbolType) -> bool:
+    """
+    Check if first argument is an empty field and second is of a compatible type.
```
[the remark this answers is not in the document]
nfarabullini: I have made an edit now, it should be clear

src/functional/ffront/type_info.py
```diff
@@ -171,6 +171,27 @@ def is_integral(symbol_type: ct.SymbolType) -> bool:
...
+    >>> is_number(ct.ScalarType(kind=ct.ScalarKind.INT32))
+    True
+    >>> is_number(ct.ScalarType(kind=ct.ScalarKind.BOOL))
+    False
+    >>> is_number(ct.FieldType(dims=[], dtype=ct.ScalarType(kind=ct.ScalarKind.INT)))
+    False
+    """
+    if not isinstance(symbol_type, ct.ScalarType):
+        return False
+    # TODO: @nfarabullini re-factor is_arithmetic such it only checks for scalars and the emtpy field pass is in an another function
```
tehrengruber, line 191: Line length. You can configure PyCharm to do that for you (Ctrl + Alt + L on Linux). The length is configured in Settings > Editor > Code Style > Hard wrap at.

src/functional/ffront/type_info.py
```diff
@@ -286,6 +307,35 @@ def is_concretizable(symbol_type: ct.SymbolType, to_type: ct.SymbolType) -> bool
...
+        return True
+    elif is_number(b_arg):
+        raise GTTypeError(f"{b_arg} is not compatible with {a_arg}")
+    return False
+
+
+def promote_zero_dims(
+    args: list[ct.SymbolType], function_type: ct.FieldOperatorType | ct.ProgramType
+):
+    """Cast arg types to zero dimensional fields if compatible and required by function signature."""
```
tehrengruber, line 330: Naming as above.

src/functional/ffront/type_info.py
```diff
@@ -286,6 +307,35 @@ def is_concretizable(symbol_type: ct.SymbolType, to_type: ct.SymbolType) -> bool
...
+    new_args = args
```
tehrengruber, line 331: This is not sufficient to avoid modifying the argument of the caller. See this example:
```
>>> a = [1, 2, 3]
>>> b = a
>>> b[0] = 5
>>> a[0]
5  # -> a has only changed
```
You can either do a shallow copy or what I would suggest is this pattern:
```python
new_args = []
for arg in args:
    if ...:
        new_arg = ...
        new_args.append(new_arg)
    else:
        new_args.append(arg)

File: tests/functional_tests/ffront_tests/past_common_fixtures.py [hunk 16]
```diff
@@ -41,6 +41,14 @@ def identity(in_field: Field[[IDim], "float64"]) -> Field[[IDim], "float64"]:
...
 
 
+@pytest.fixture
+def empty_identity_def():
+    def empty_identity(empty_field: Field[[], "float64"]):
+        return empty_field
+
+    return empty_identity
+
+
```
tehrengruber, line 51:
```suggestion
```
Looks like dead-code by now.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -1277,6 +1277,38 @@ def return_undefined():
             return undefined_symbol
 
 
+def test_zero_dims_fields():
```
tehrengruber, line 1280: Please parameterize the backend, i.e. using the `fieldview_backend` fixture.

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -1277,6 +1277,38 @@ def return_undefined():
...
+    out = np_as_located_field(IDim)(np.ones((10,)))
+
+    @field_operator
+    def fieldop_implicit_broadcast(
+        zero_dim_inp: Field[[], float], inp: Field[[IDim], float], scalar: float
+    ) -> Field[[IDim], float]:
+        return inp + zero_dim_inp * scalar
+
+    @field_operator
+    def fieldop_implicit_broadcast_2(inp: Field[[IDim], float]) -> Field[[IDim], float]:
```
tehrengruber, line 1304: Please parameterize the backend.

tehrengruber (changes requested):
Some of the previous points are not resolved. I tried to be more elaborate in the comments, so there are more now.

tests/functional_tests/ffront_tests/past_common_fixtures.py
nfarabullini: I left this because it may be useful in the future

src/functional/ffront/type_info.py
```diff
@@ -286,6 +307,35 @@ def is_concretizable(symbol_type: ct.SymbolType, to_type: ct.SymbolType) -> bool
     return False
 
 
+def _is_zero_dim_field(field: ct.SymbolType) -> bool:
+    return isinstance(field, ct.FieldType) and field.dims != Ellipsis and len(field.dims) == 0
+
+
+def is_zero_dim_field_broadcastable(a_arg: ct.SymbolType, b_arg: ct.SymbolType) -> bool:
```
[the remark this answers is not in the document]
nfarabullini: I decided to remove the function, based on the way the code is now it does not make a lot of sense

src/functional/ffront/type_info.py
```diff
@@ -171,6 +171,28 @@ def is_integral(symbol_type: ct.SymbolType) -> bool:
...
+    >>> is_number(ct.ScalarType(kind=ct.ScalarKind.INT32))
+    True
+    >>> is_number(ct.ScalarType(kind=ct.ScalarKind.BOOL))
+    False
+    >>> is_number(ct.FieldType(dims=[], dtype=ct.ScalarType(kind=ct.ScalarKind.INT)))
+    False
+    """
+    if not isinstance(symbol_type, ct.ScalarType):
+        return False
+    # TODO(nfarabullini) re-factor is_arithmetic such that it only checks for scalars
```
tehrengruber, line 191:
```suggestion
    # TODO(nfarabullini): re-factor is_arithmetic such that it only checks for scalars
```

PR #1034, opened 2022-11-07
Setting cache directory programmatically

stubbiali:
This light-weight PR proposes not to reset the caching strategy within `StencilBuilder` when setting the external parameters. This would allow to set the cache directory (both `root_name` and `dir_name`) without resorting to environmental variables.

### DIFF
--- a/src/gt4py/stencil_builder.py
+++ b/src/gt4py/stencil_builder.py
@@ -229,7 +229,6 @@ class StencilBuilder:
         """
         self._build_data = {}
         self._externals = externals
-        self.with_caching(self.caching.name)
         return self
 
     @property


tehrengruber:
Thanks, I have overlooked this in #892. I pushed a small change on-top that avoids duplicating the forwarding of the `cache_settings` and gives `with_caching` a clearer semantic. Can you take a quick look?

tehrengruber:
bors try

src/gt4py/stencil_builder.py
```diff
@@ -125,8 +123,10 @@ def with_caching(
         -----
         Resets all cached build data.
         """
-        self._build_data = {}
-        self.caching = gt4py.caching.strategy_factory(caching_strategy_name, self, *args, **kwargs)
+        self._build_data: Dict[str, Any] = {}
+        self.caching = gt4py.caching.strategy_factory(
+            caching_strategy_name, self, *args, **self.options.cache_settings, **kwargs
+        )
```
stubbiali, line 129: To avoid duplicated arguments, I'd do e.g.
```suggestion
        kwargs = self.options.cache_settings.copy().update(kwargs)
        self.caching = gt4py.caching.strategy_factory(
            caching_strategy_name, self, *args, **kwargs
        )
```

tehrengruber: Good catch! I have opted for this version:
```
kwargs = {**self.options.cache_settings, **kwargs}
```

tehrengruber:
bors try

PR #1036, opened 2022-11-08
Implementation of Invert and XOR 

nfarabullini:
Implementation of ~ and ^ for booleans and fields of booleans:

```python
@field_operator
def tilde_fieldop(inp1: Field[[IDim], bool]) -> Field[[IDim], bool]:
    return ~inp1
```

```python
@field_operator
def tilde_xor(inp1: Field[[IDim], bool], inp2: Field[[IDim], bool]) -> Field[[IDim], bool]:
    return inp1 ^ inp2
```


havogt (changes requested):
Looks very good! Can you add a test case for the xor at iterator ir level here https://github.com/GridTools/gt4py/blob/06a3428e2632ed92570e5a850dd10059b19fe965/tests/functional_tests/iterator_tests/test_builtins.py#L95

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -133,6 +133,24 @@ def pow(inp1: Field[[IDim], float64]) -> Field[[IDim], float64]:
     assert np.allclose(a.array() ** 2, b)
 
 
+def test_xor(fieldview_backend):
+    if fieldview_backend == gtfn_cpu.run_gtfn:
+        pytest.xfail("gtfn does not yet support math builtins")
+
```
havogt, line 139:
```suggestion
```
it does support math builtins. does it break?

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -201,6 +219,23 @@ def uneg(inp: Field[[IDim], int32]) -> Field[[IDim], int32]:
     assert np.allclose(b, np.full((size), -1, dtype=int32))
 
 
+def test_unary_invert(fieldview_backend):
+    if fieldview_backend == gtfn_cpu.run_gtfn:
+        pytest.xfail("gtfn does not yet support math builtins")
+
```
havogt, line 225:
```suggestion
```

tests/functional_tests/ffront_tests/test_execution.py
havogt: Can you also try to remove it from the other math builtin tests? Could be in a separate PR if you don't want to pollute this one.

havogt (changes requested):
Sorry, I overlooked in the last review that gtfn is excluded. I think it should work.

havogt (approved):

PR #1039, opened 2022-11-09
Functional: Refactoring of math and unary builtins tests 

nfarabullini:
Almost all math built-ins have been integrated in the backend, hence check for fieldview_backend in tests is not longer necessary:
```python
if fieldview_backend == gtfn_cpu.run_gtfn:
    pytest.xfail("gtfn does not yet support math builtins")
```

Addition of trigonometric, log, exp, isinfinite, etc. builtins in front-end tests


src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -409,6 +409,16 @@ def _deduce_binop_type(
         if node.op == ct.BinaryOperator.POW:
             return left_type
 
+        if node.op == ct.BinaryOperator.MOD and (
+            isinstance(right_type, ct.ScalarType) and not type_info.is_integral(right_type)
```
havogt, line 413: I don't think the right hand side needs to be a scalar (if that is what you wanted to express, currently the conditions seem wrong).

src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -188,6 +188,8 @@ class GTFN_lowering(eve.NodeTranslator, eve.VisitorWithSymbolTableTrait):
         "and_": "&&",
         "or_": "||",
         "xor_": "^",
+        "mod": "%",
+        "floordiv": "//",
```
havogt, line 192:
```suggestion
```
C++ has only one kind of division operator. We have to implement the Python floordiv, by casting (or by introducing a helper function that behaves that way).

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -1194,9 +1185,6 @@ def program_domain(a: Field[[IDim, JDim], float64]):
 
 
 def test_domain_input_bounds(fieldview_backend):
-    if fieldview_backend == gtfn_cpu.run_gtfn:
-        pytest.skip("FloorDiv and Power not fully supported in gtfn.")
```
havogt, line 1198: To keep this PR simple you could keep "FloorDiv not fully supported in gtfn" if it fails, see the other comment. But most likely it works, because otherwise `//` should give an error in the generated code.

havogt (changes requested):

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -409,6 +409,12 @@ def _deduce_binop_type(
         if node.op == ct.BinaryOperator.POW:
             return left_type
 
+        if node.op == ct.BinaryOperator.MOD and not type_info.is_integral(right_type):
+            raise FieldOperatorTypeDeductionError.from_foast_node(
+                arg,
+                msg=f"Type {right_type} can not be used in operator `{node.op}`, it can only accept scalar ints",
```
havogt, line 415:
```suggestion
                msg=f"Type {right_type} can not be used in operator `{node.op}`, it can only accept ints",
```

src/functional/ffront/past_passes/type_deduction.py
```diff
@@ -143,6 +143,12 @@ def _deduce_binop_type(
         if node.op == ct.BinaryOperator.POW:
             return left_type
 
+        if node.op == ct.BinaryOperator.MOD and not type_info.is_integral(right_type):
+            raise ProgramTypeError.from_past_node(
+                arg,
+                msg=f"Type {right_type} can not be used in operator `{node.op}`, it can only accept scalar ints",
```
havogt, line 149:
```suggestion
                msg=f"Type {right_type} can not be used in operator `{node.op}`, it can only accept ints",
```

havogt (changes requested):
Can you also add a test for the error path, i.e. when the right hand side of mod is not scalar int or field of integer.

tests/functional_tests/ffront_tests/builtins_tests/test_gt4py_builtins.py
```diff
@@ -0,0 +1,427 @@
...
+
+size = 10
+mask = np_as_located_field(IDim)(np.zeros((size,), dtype=bool))
+mask.array()[size // 2] = True
+a_int = np_as_located_field(IDim)(np.random.randn(size).astype("int64"))
+b_int = np_as_located_field(JDim)(np.random.randn(size).astype("int64"))
+out_int = np_as_located_field(IDim, JDim)(np.zeros((size, size), dtype=int64))
+a_float = np_as_located_field(IDim)(np.random.randn(size).astype("float64"))
+b_float = np_as_located_field(IDim)(np.random.randn(size).astype("float64"))
+out_float = np_as_located_field(IDim)(np.zeros((size), dtype=float64))
```
havogt, line 73: There is as lot of code duplication now in other files, that should be moved to a separate file.

havogt (changes requested):

havogt (changes requested):
Only this small rename request, otherwise looks good!

havogt (approved):
Thanks!

PR #1042, opened 2022-11-09
Update AUTHORS.rst

ofuhrer:
## Description

Update to correct affiliation of authors from AI2.


jdahm:
Thanks! What do you think of just removing the affiliations altogether and maybe adding an email or github profile link?

ofuhrer:
I would not be in favor of that, since the code your write when under contract typically belongs to the institution you are hired by. In our case the ownership has been transferred from Vulcan to AI2.

PR #1043, opened 2022-11-10
Allow aligned_index to be None

jdahm:
## Description

There was a bug previously that didn't actually allow `aligned_index` to be unset in the storage creation calls. This fixes that.


gronerl (approved):

gronerl:
bors try

PR #1046, opened 2022-11-10
Fix warning dump from -Wdeprecated-declarations when compiling stencils

jdahm:
## Description

After upgrading the GridTools release used in gt4py, a new deprecation warning is printed. This corrects the usage so that no warning is printed by the C++ compiler.


jdahm:
bors try

jdahm:
Passed but the check timed out.

jdahm:
bors try

PR #1047, opened 2022-11-11
Resolve circular imports in gtc.daceir and gtc.dace

### LINKED ISSUE #1041: Importing daceir causes circular dependency if dace_backend is not already imported
When updating the eve version, I found that the dace backend has a circular dependency between `gtc.dace.utils` and `gtc.daceir`. There could be others, too.

One possible resolution we started working on was to move the functions that clash to a separate `gtc.dace.daceir_utils` module.

gronerl:
Resolves #1041


src/gtc/dace/symbol_utils.py
```diff
@@ -0,0 +1,59 @@
+from functools import lru_cache
```
jdahm, line 1: 👍

src/gtc/daceir.py
```diff
@@ -436,7 +439,7 @@ def set_interval(
         return GridSubset(intervals=intervals)
 
     @classmethod
-    def from_gt4py_extent(cls, extent: gtc.definitions.Extent):
+    def from_gt4py_extent(cls, extent: Extent):
```
jdahm, line 442: I actually prefer seeing the `gtc.definitions.Extent` here, but we'll take care of that in an upcoming PR :)

src/gtc/dace/expansion/daceir_builder.py
```diff
@@ -150,7 +153,7 @@ def all_regions_same(scope_nodes):
 class DaCeIRBuilder(NodeTranslator):
     @dataclass
     class GlobalContext:
-        library_node: StencilComputation
+        library_node: "StencilComputation"
```
jdahm, line 156: If you import `from __future__ import annotations` do you need the quotes? https://peps.python.org/pep-0563/

jdahm (approved):
Looks good!

PR #1053, opened 2022-11-17
Functional: astype builtin implementation

nfarabullini:
Implementation of cast builtin analogous to behavior in typing.cast:
```python
@field_operator
def cast_fieldop(a: Field[[IDim], float64]) -> Field[[IDim], int64]:
    return cast(int64, a)
```


src/functional/program_processors/codegens/gtfn/gtfn_ir.py
```diff
@@ -53,6 +53,11 @@ class TernaryExpr(Expr):
     false_expr: Expr
 
 
+class TypeExpr(Expr):
```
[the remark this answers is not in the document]
havogt: or forward as a FunCall and resolve on gtfn_ir level.

src/functional/program_processors/codegens/gtfn/codegen.py
```diff
@@ -88,6 +88,7 @@ def visit_Literal(self, node: gtfn_ir.Literal, **kwargs: Any) -> str:
     UnaryExpr = as_fmt("{op}({expr})")
     BinaryExpr = as_fmt("({lhs}{op}{rhs})")
     TernaryExpr = as_fmt("({cond}?{true_expr}:{false_expr})")
+    TypeExpr = as_fmt("{new_dtype}({obj_expr})")
```
havogt, line 91:
```suggestion
    TypeExpr = as_fmt("static_cast<{new_dtype}>({obj_expr})")
```

havogt (changes requested):
- [x] Frontend part looks good, however we need to understand how the represent the type of the first argument, which is a type of type.
- [ ] Iterator IR: we should also represent the type differently, e.g. as builtin. Have to double-check with someone.
- [ ] Iterator IR tests should be added, including comparison of results between backends (ensure that casting from one type to the other does exactly the same thing in python and in c++).

src/functional/iterator/builtins.py
```diff
@@ -59,6 +59,11 @@ def if_(*args):
     raise BackendNotSelectedError()
 
 
+@builtin_dispatch
+def cast_(*args):
```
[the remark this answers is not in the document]
nfarabullini: I lean towards keeping the `_` for that reason

src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -55,6 +56,7 @@ def pytype_to_cpptype(t: str):
             "float32": "float",
             "float64": "double",
             "int": "int",
+            "str": "str",
```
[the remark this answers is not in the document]
nfarabullini: Without it the following error is thrown:

`TypeError: Unsupported type 'str'`

src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -337,6 +339,12 @@ def _collect_offset_or_axis_node(
             )
         return result
 
+    def _check_shift_scan(self, node: itir.FunCall) -> None:
+        if node.fun.id == "shift":
+            raise ValueError("unapplied shift call not supported: {node}")
+        elif node.fun.id == "scan":
+            raise ValueError("scans are only supported at the top level of a stencil closure")
+
```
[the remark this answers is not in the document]
nfarabullini: because I added the "cast" case, pre-commit was outputting that `visit_FunCall` is too complex (hence I had to make it simpler).

I am ok changing the name of `_check_shift_scan` to `_error_on_illegal_function_calls`

src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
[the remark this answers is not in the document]
havogt: If other problems are fixed, this should not happen.

src/functional/ffront/foast_to_itir.py
```diff
@@ -417,6 +417,17 @@ def visit_Call(self, node: foast.Call, **kwargs) -> itir.FunCall:
...
+    def _visit_cast(self, node: foast.Call, **kwargs) -> itir.FunCall:
+        if type_info.is_arithmetic(node.args[0].type.returns):
+            if type_info.is_floating_point(node.args[0].type.returns):
+                dtype = itir.Literal(value="float", type="str")
+            else:
+                dtype = itir.Literal(value="int", type="str")
+        else:
+            dtype = itir.Literal(value="bool", type="str")
+        obj = (to_value(node.args[1]))(self.visit(node.args[1], **kwargs))
+        return self._lift_lambda(node)(im.call_("cast_")(dtype, obj))
```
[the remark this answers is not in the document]
nfarabullini: dtype being represented as a literal of type string was discussed with @DropD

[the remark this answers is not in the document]
havogt: Confirmed that types should be builtins.

src/functional/ffront/fbuiltins.py
```diff
@@ -75,6 +75,17 @@ def __gt_type__(self):
     )
 )
 
+cast = BuiltInFunction(
+    ct.FunctionType(
+        args=[
+            ct.DeferredSymbolType(constraint=ct.ScalarType),
```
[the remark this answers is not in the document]
havogt: Note that above in the file we already have `TYPE_BUILTINS`.
(Which is used to support type constructors `float64(42.0)` currently. We should check that for general expressions it either works or gives an error.)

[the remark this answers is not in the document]
tehrengruber: At some point we want to have a type of a type, but I would keep it simple here: `ct.DeferredSymbolType(constraint=None)`

egparedes:
I haven't read the PR but I would like to add my two cents here in case it helps: I think converting a scalar value from one type to another should use the type constructor syntax (e.g. `float(3)`, `int64(value)`) as is already implemented AFAIK. For converting fields to another dype, I would use the  syntax from the standard Array API (https://data-apis.org/array-api/latest/API_specification/data_type_functions.html) which defines the free functions `astype` and `broadcast_to` for creating new fields with different types and for broadcasting.

tehrengruber:
@egparedes This was also my first though when @havogt showed me the proposal for the syntax. However we concluded that `cast` is a good intermediate step because otherwise we would need to implement both `.astype` for fields and the constructors for scalars, e.g. `float64(my_int)`, which arguably needs more work. For reference the project description:
> TODO decide with Enrique field.astype(dtype) + dtype(scalar) or cast(field, dtype) (and the other as syntax sugar later)

@nfarabullini You are currently implementing `cast(dtype, val)`, but it should be `cast(val, dtype)` (when a function transforms a value these values appear first in the signature, followed by the parameters to the transformation).

nfarabullini:
@tehrengruber I was mirroring the semantics from typing.cast: https://docs.python.org/3/library/typing.html#typing.cast

tehrengruber:
I have to admit that it looks absolutely awful, but that makes sense. Thanks for the clarification. See https://github.com/python/typing/issues/63 for the discussion on the topic.

egparedes:
> However we concluded that `cast` is a good intermediate step because otherwise we would need to implement both `.astype` for fields and the constructors for scalars, e.g. `float64(my_int)`, which arguably needs more work. For reference the project description:

@tehrengruber In the Array API standard `astype(x, dtype)` is a free function, not a method of Field objects, so I'd say the effort is exactly the same and still think we should use the `astype()`. Regarding the second point of implementing `float64(my_var)`, AFAIK this is already implemented so there is nothing to do here, right?

src/functional/program_processors/codegens/gtfn/gtfn_ir.py
```diff
@@ -53,6 +53,11 @@ class TernaryExpr(Expr):
     false_expr: Expr
 
 
+class CastExpr(Expr):
```
tehrengruber, line 56: This is a valid way to the express a `cast` in an IR (@petiaccja actually asked me a while ago why we don't express more information in the types of the IR nodes the same way you did here). However it is not in line with the rest of the IR and for consistency I suggest the `cast` to be represented as a regular `FunCall` node to the `builtin` function `cast`.

nfarabullini: I am asking this for personal understanding, but I don't think that we should do:

```FunCall = as_fmt("static_cast<{new_dtype}>({obj_expr})")```

Perhaps it is not even necessary under these circumstances (?)

Comment (petiaccja) [reply src/functional/program_processors/codegens/gtfn/gtfn_ir.py:56]: I think the central problem is that `FunCall` has a `list[Expr]` as arguments, so you cannot pass a dtype there.

The options I see:
1. `FunCall(fun=SymRef(id="cast"), args=[TypeLiteral(dtype=float), <value>])`
2. `CastExpr(type=float, value=<value>)`
3. `FunCall(fun=SymRef(id="cast"), args=[SymRef(id="float"), <value>])`
4. `FunCall(fun=SymRef(id="float"), args=[<value>])`

For number 3) and 4), you have to add supported types as builtin symbols. IMHO we can drop 3), it's just a more complicated version of 4), and also makes the type of the new builtins questionable, because with 3) they are clearly functions.

If you want to represent types in the IR by other than strings, I believe you have to add a new node. It could be either a `TypeLiteral` or a `CastExpr`. If you use a `cast` builtin function, then there is one more special casing for the `FunCall` node and there is one more constraint (1st argument must be `TypeLiteral`) that is not verified by static analysis. `CastExpr` doesn't add new special cases or constraints. On the other hand, using a `cast` builtin is more in line with the rest of the IR.

Just my 2 cents.

File: src/functional/ffront/fbuiltins.py [hunk 9]
```diff
@@ -75,6 +75,17 @@ def __gt_type__(self):
     )
 )
 
+astype = BuiltInFunction(
+    ct.FunctionType(
+        args=[
+            ct.DeferredSymbolType(constraint=ct.FieldType),
+            ct.DeferredSymbolType(constraint=ct.FunctionType),
```
havogt, line 82:
```suggestion
            ct.DeferredSymbolType(constraint=None),
```
as far as I can see that was Till's proposal.

havogt: Should `astype` also work on scalars? @egparedes

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -604,6 +604,38 @@ def _visit_max_over(self, node: foast.Call, **kwargs) -> foast.Call:
...
+        dtype_obj_type = node.args[1].type
+        if not isinstance(dtype_obj_type, ct.FunctionType) or not (
+            type_info.is_arithmetic(dtype_obj_type.returns)
+            or type_info.is_logical(dtype_obj_type.returns)
+        ):
+            raise FieldOperatorTypeDeductionError.from_foast_node(
+                node,
+                msg=f"Incompatible argument in call to `{node.func.id}`'s new dtype. "
+                f"Expected an arithmetic or boolean dtype, but got {dtype_obj_type}.",
+            )
```
havogt, line 618: Please document why it makes sense that `dtype_obj_type` is a FunctionType. I guess it is because `float32` etc are all functions and we abuse it's return type.

Where do we actually deduce the type of these? @tehrengruber

src/functional/ffront/type_info.py
```diff
@@ -277,6 +277,8 @@ def is_concretizable(symbol_type: ct.SymbolType, to_type: ct.SymbolType) -> bool
     True
 
     """
+    if isinstance(symbol_type, ct.FunctionType):
+        raise NotImplementedError("`FunctionType` is not concretizable")
```
havogt, line 281: What's this change about? Shouldn't this case be handled in the `is_concrete()` case?

src/functional/iterator/ir.py
```diff
@@ -105,6 +105,9 @@ def _output_validator(self: datamodels.DataModelTP, attribute: datamodels.Attrib
 }
 UNARY_MATH_FP_PREDICATE_BUILTINS = {"isfinite", "isinf", "isnan"}
 BINARY_MATH_NUMBER_BUILTINS = {"minimum", "maximum", "fmod", "power"}
+# create a new list with all type builtins, e-g- float64
```
havogt, line 108:
```suggestion
```

src/functional/iterator/transforms/unroll_reduce.py
```diff
@@ -59,6 +59,13 @@ def _make_if(cond: ir.Expr, true_expr: ir.Expr, false_expr: ir.Expr):
             args=[cond, true_expr, false_expr],
         )
 
+    @staticmethod
+    def _make_astype(obj_expr: ir.Expr, new_dtype: ir.SymRef):
+        return ir.FunCall(
+            fun=ir.SymRef(id="astype_"),
+            args=[obj_expr, new_dtype],
+        )
+
```
havogt, line 68: seems unused

src/functional/program_processors/codegens/gtfn/codegen.py
```diff
@@ -64,6 +64,10 @@ class GTFNCodegen(codegen.TemplatedGenerator):
     def visit_SymRef(self, node: gtfn_ir.SymRef, **kwargs: Any) -> str:
         if node.id == "get":
             return "::gridtools::tuple_util::get"
+
+        if node.id in ["int64", "int32"]:
+            return node.id + "_t"
+
```
havogt, line 70: I think we should do the following:
- put the builtin type to c++ name mapping into the `_builtins_mapping`
- move the logic from `visit_FunCall` that dispatches on SymRef here (needs to be adapted)
  ```        
        if isinstance(node.fun, gtfn_ir.SymRef) and node.fun.id in self._builtins_mapping:
            return self.generic_visit(node, fun_name=self._builtins_mapping[node.fun.id])
        if isinstance(node.fun, gtfn_ir.SymRef) and node.fun.id in gtfn_ir.GTFN_BUILTINS:
            qualified_fun_name = f"gtfn::{node.fun.id}"
return self.generic_visit(node, fun_name=qualified_fun_name)```

havogt (changes requested):
Thanks, it's going in a good direction.
Main missing pieces are tests on iterator ir:
- basic tests, similar to the one you made in test_execution: if properly setup, this will test the full iterator ir part, including tracing, etc. Most likely there is still some work todo, let me know if you get stuck there.
- tests to should compare that selected corner cases have the same behavior in all backends (all = c++ and Python currently).

src/functional/ffront/fbuiltins.py
nfarabullini: We discussed this in a call with @egparedes and @tehrengruber and the conclusion was that it would not

src/functional/ffront/type_info.py
tehrengruber: In a prior version of this PR @nfarabullini used something along this lines
```suggestion
ct.FunctionType(args=[ct.DeferredSymbolType(constraint=...)], ...)
```
in `fbuiltins.py` to restrict the arguments of the number type constructors. Currently this will wrongly return `False` here. `is_concrete` also does not handle the case correctly and should consequently also throw a `NotImplementedError`. However I would argue that the check is still valid here as after adding support for these cases in `is_concrete` this function would again return `False`, which it shouldn't.

tests/functional_tests/iterator_tests/test_builtins.py
```diff
@@ -90,6 +93,21 @@ def fenimpl(size, arg0, arg1, arg2, out):
     return run_processor(fenimpl, processor, out.shape[0], *inps, out)
 
 
+def fencil_cast(builtin, out, *inps, processor, as_column=False):
```
nfarabullini, line 96: builtin not needed here, no parameterization either

tests/functional_tests/iterator_tests/test_builtins.py
```diff
@@ -134,11 +152,29 @@ def test_arithmetic_and_logical_builtins(program_processor, builtin, inputs, exp
         assert np.allclose(np.asarray(out), expected)
 
 
+@pytest.mark.parametrize("as_column", [False, True])
+@pytest.mark.parametrize("builtin, inputs, expected", [(cast_, [[1], [float64]], [1.0])])
```
nfarabullini, line 156: add other cases, e.g. from float64 to float32, booleans

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -720,6 +721,30 @@ def conditional(
...
 
+def test_astype(fieldview_backend):
+    size = 10
+    b = np_as_located_field(IDim)(np.ones(size))
+    c = np_as_located_field(IDim)(np.ones((size,), dtype=bool))
+    out_bool = np_as_located_field(IDim)(np.zeros((size,), dtype=bool))
+    out_int = np_as_located_field(IDim)(np.zeros((size,), dtype=int))
+
+    @field_operator(backend=fieldview_backend)
+    def astype_fieldop_int(b: Field[[IDim], float64]) -> Field[[IDim], int]:
```
nfarabullini, line 732: test all cases here too

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -658,6 +658,38 @@ def _visit_max_over(self, node: foast.Call, **kwargs) -> foast.Call:
     def _visit_min_over(self, node: foast.Call, **kwargs) -> foast.Call:
         return self._visit_reduction(node, **kwargs)
 
+    def _visit_astype(self, node: foast.Call, **kwargs) -> foast.Call:
+        casted_obj_type = node.args[0].type
+        dtype_obj_type = node.args[1].type
+        if not isinstance(dtype_obj_type, ct.FunctionType) or not (
```
nfarabullini, line 664: check here that obj is in TYPE_BUILTINS instead of being of FunctionType

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -658,6 +658,38 @@ def _visit_max_over(self, node: foast.Call, **kwargs) -> foast.Call:
     def _visit_min_over(self, node: foast.Call, **kwargs) -> foast.Call:
         return self._visit_reduction(node, **kwargs)
 
+    def _visit_astype(self, node: foast.Call, **kwargs) -> foast.Call:
+        casted_obj_type = node.args[0].type
+        dtype_obj_type = node.args[1].type
+        if not node.args[1].id.lower() in fbuiltins.TYPE_BUILTINS.__str__() or not (
+            type_info.is_arithmetic(dtype_obj_type.returns)
+            or type_info.is_logical(dtype_obj_type.returns)
+        ):
```
havogt, line 667:
```suggestion
        if not node.args[1].id in fbuiltins.TYPE_BUILTIN_NAMES:
```
or maybe a string conversion is needed like
```suggestion
        if not str(node.args[1].id) in fbuiltins.TYPE_BUILTIN_NAMES:
```

tests/functional_tests/iterator_tests/test_builtins.py
```diff
@@ -134,11 +156,29 @@ def test_arithmetic_and_logical_builtins(program_processor, builtin, inputs, exp
         assert np.allclose(np.asarray(out), expected)
```
havogt, line 158: Let's use something like this as test setup

```
@pytest.mark.parametrize(
    "input_value, dtype, expected_value",
    [
        (float64("0.1"), float32, float64(float32("0.1"))),
        (int64(42), bool, int64(1)),
        (int64(2147483648), int32, int64(int32(2147483648))),
    ],
)
@pytest.mark.parametrize("as_column", [False, True])
def test_cast(program_processor, as_column, input_value, dtype, expected_value):
    program_processor, validate = program_processor
    column_axis = IDim if as_column else None

    inp = asfield(*asarray(input_value))[0]
    out = asfield((np.zeros_like(*asarray(expected_value))))[0]

    @fundef
    def sten_cast(value):
        return cast_(deref(value), dtype)

    run_processor(
        sten_cast[{IDim: range(1)}],
        program_processor,
        inp,
        out=out,
        offset_provider={},
        column_axis=column_axis,
    )

    if validate:
        assert out[0] == expected_value
```

havogt: Let's chat about that in detail. (It works if `as_column == False`, but fails for `True`)

tests/functional_tests/ffront_tests/test_execution.py
```diff
@@ -223,6 +225,41 @@ def fencil(out: Field[[Edge], float64], inp: Field[[Edge], float64], factor: flo
...
+    astype_fieldop_bool(b_float_64, out=out_bool, offset_provider={})
+    assert np.allclose(c_bool, out_bool)
+
+    @field_operator(backend=fieldview_backend)
+    def astype_fieldop_float(b: Field[[IDim], int64]) -> Field[[IDim], int32]:
+        d = astype(b, int32)
+        return d
+
+    astype_fieldop_float(c_int64, out=out_int_32, offset_provider={})
+    assert np.allclose(c_int32.array(), out_int_32)
```
havogt, line 260: I think it should be separate test cases, as failure of one will not be reported nicely.

tests/functional_tests/ffront_tests/test_type_deduction.py
```diff
@@ -648,6 +649,33 @@ def tuple_where_mix_dims(
...
+def test_astype_wrong_input():
+    ADim = Dimension("ADim")
+    msg = "Could not deduce type"
+
+    def bad_input_astype(a: Field[[ADim], float64]):
+        return astype(a, "float64")
+
+    with pytest.raises(
+        FieldOperatorTypeDeductionError,
+        match=msg,
```
havogt, line 674: The message should either be more precise or is not needed ("Could not deduce type" is already implied by the type of the error.)

havogt (changes requested):
Mainly the iterator IR tests should be improved, we can have a chat about that.

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -664,6 +664,42 @@ def _visit_max_over(self, node: foast.Call, **kwargs) -> foast.Call:
...
+                f"Expected an arithmetic or boolean dtype, but got {dtype_obj_type}.",
+            )
+        if not isinstance(casted_obj_type, ct.FieldType) or not (
+            type_info.is_arithmetic(casted_obj_type) or type_info.is_logical(casted_obj_type)
+        ):
+            raise FieldOperatorTypeDeductionError.from_foast_node(
+                node,
+                msg=f"Incompatible argument in call to `{node.func.id}`'s object. "
+                f"Expected arithmetic or boolean FieldType, but got {casted_obj_type}.",
+            )
```
havogt, line 690: maybe should be assert as `as_type` already guarantees that the first argument is a FieldType

havogt (approved):
Looks good to go!

PR #1055, opened 2022-11-18
ITIR pass improvements

tehrengruber:
- Fixed bug in eve.NodeVisitor leading to outdated symbol tables (independently discovered by @jdahm and fixed the same way here -> no merge conflicts)
- Incomplete offsets for lifted stencil call in embedded backend (allows using them in reductions)
- Factored out `CountSymbolRefs` pass from `inline_lambdas` module (now also used in new `MergeLet` pass)
- Fixed several bugs in the common subexpression elimination (BFS instead of DFS to collect larger expressions + refactoring of `CollectSubexpressions` subpass)
- Factored out lambda inlining into visitor and actual inliner
- Fixes for corner cases in collision resolution of lambda inliner
- Allow force inling lifted stencil arguments to lambda functions in lambda inliner
- Reworked lift inliner
  - Allow inlining nested lifts without a deref
  - Stateless inlining heuristics predicate
  - Directly use lambda inliner for nested lifts to reduce ping-pong between inline lift and inline lambda pass
- New MergeLet pass to reduce tree depth and readability of ir


src/functional/iterator/transforms/inline_lambdas.py
```diff
@@ -1,32 +1,94 @@
...
+
+    if len(eligible_params) != 0 and not any(eligible_params):
+        return node
 
-        self.generic_visit(node, active_refs=active_refs)
+    refs = set().union(
+        *(
+            arg.pre_walk_values().if_isinstance(ir.SymRef).getattr("id").to_set()
+            for i, arg in enumerate(node.args)
+            if eligible_params[i]
```
fthaler, line 48:
```suggestion
            for arg, eligible in zip(node.args, eligible_params)
            if eligible
```

src/functional/iterator/transforms/inline_lambdas.py
```diff
@@ -1,32 +1,94 @@
...
+        for i, arg in enumerate(node.args):
+            if (
+                isinstance(arg, ir.FunCall)
+                and isinstance(arg.fun, ir.FunCall)
+                and isinstance(arg.fun.fun, ir.SymRef)
+                and arg.fun.fun.id == "lift"
+            ):
+                eligible_params[i] = True
+
+    if len(eligible_params) != 0 and not any(eligible_params):
```
fthaler, line 41: Maybe a bit more clear what’s checked:
```suggestion
    if node.fun.params and not any(eligible_params):
```

src/functional/iterator/transforms/inline_lambdas.py
```diff
@@ -1,32 +1,94 @@
...
+
+        for sym in clashes:
+            name_map[sym] = new_name(sym)
+
+        expr = RenameSymbols().visit(expr, name_map=name_map)
+
+    symbol_map = {
+        param.id: arg
+        for i, (param, arg) in enumerate(zip(node.fun.params, node.args))
+        if eligible_params[i]
```
fthaler, line 74:
```suggestion
        for param, arg, eligible in zip(node.fun.params, node.args, eligible_params)
        if eligible
```

src/functional/iterator/transforms/inline_lifts.py
```diff
@@ -1,99 +1,214 @@
...
+            assert sym not in extracted_args or extracted_args[sym] == arg
+            extracted_args[sym] = arg
+            new_args.append(arg)
+        else:
+            new_symbol = _generate_unique_symbol(
+                desired_name=(inner_stencil, i),
+                occupied_names=symtable.keys(),
+                occupied_symbols=extracted_args.keys(),
+            )
+            # assert new_symbol not in extracted_args
```
fthaler, line 86: Why commented out? (And if correctly commented out, please remove or add explanation)

src/functional/iterator/transforms/merge_let.py
```diff
@@ -0,0 +1,54 @@
...
+            inner_lambda_args = node.fun.expr.args
+            # skip if we have a collision
+            if set(outer_lambda.params) & set(inner_lambda.params):
+                return node
+            # check if the arguments to the outer lambda call use a symbol of the inner lambda
+            # e.g. (λ(a) → (λ(b) → b)(a))(b)
+            ref_counts_outer = CountSymbolRefs.apply(
+                outer_lambda_args, [param.id for param in inner_lambda.params]
+            )
+            if any(ref_count != 0 for ref_count in ref_counts_outer.values()):
```
fthaler, line 40: Same, but shorter:
```suggestion
            if any(ref_counts_outer.values()):
```

src/functional/iterator/transforms/merge_let.py
```diff
@@ -0,0 +1,54 @@
...
+                return node
+            # check if the argument to the inner lambda call depend on an argument to the outer lambda
+            ref_counts = CountSymbolRefs.apply(
+                inner_lambda_args, [param.id for param in outer_lambda.params]
+            )
+            if any(ref_count != 0 for ref_count in ref_counts.values()):
```
fthaler, line 46: Same as above…

src/functional/iterator/transforms/merge_let.py
fthaler: But actually, this should be faster I guess (early exit):
```python
if any(trees.walk_values(outer_lambda_args).if_isinstance(ir.SymbolRef).getattr("id").if_in({p.id for p in inner_lambda.params})):
```

src/functional/iterator/transforms/simple_inline_heuristic.py
```diff
@@ -1,44 +1,33 @@
...
-                return self.generic_visit(node, inline=inline, force_noinline=False)
-        elif (
-            isinstance(node.fun, ir.FunCall)
-            and node.fun.fun == ir.SymRef(id="lift")
-            and isinstance(node.fun.args[0], ir.FunCall)
-            and node.fun.args[0].fun == ir.SymRef(id="scan")
-        ):
-            return self.generic_visit(node, inline=inline, force_noinline=True)
-        self.generic_visit(node, inline=inline, force_noinline=force_noinline)
+def is_eligable_for_inlining(node: ir.FunCall, is_scan_pass_context: bool) -> bool:
```
fthaler, line 5:
```suggestion
def is_eligible_for_inlining(node: ir.FunCall, is_scan_pass_context: bool) -> bool:
```

fthaler (changes requested):

src/functional/iterator/transforms/merge_let.py
tehrengruber: I kept the more explicit version, which I find easier to digest. The `walk_values` approach is not able to handle symbol collisions if there are new symbol declarations in `outer_lambda_args`. We might want to consider using unique symbols at some point to use this faster approach.

PR #312, opened 2021-01-21
Fix installation of GT C++ sources required for regression tests

egparedes:

PR #1057, opened 2022-11-21
GridTools as pip package

### LINKED ISSUE #937: CPU backend fails in parallel without pre-installed gridtools sources

havogt:
Resolves #937.


tests/test_integration/test_cpp_regression.py
```diff
@@ -20,6 +20,9 @@
 import numpy as np
 import pytest
 
+
```
havogt, line 23: I vote for removing these tests to resolve #312. I don't see too much value in these tests. @jdahm

tests/test_unittest/test_gtc/test_cuir_compilation.py
```diff
@@ -17,10 +17,7 @@
 import pytest
 import setuptools
 
-from gt4py import (  # TODO(havogt) this is a dependency from gtc tests to gt4py, ok?
-    config,
-    gt_src_manager,
-)
+from gt4py import config  # TODO(havogt) this is a dependency from gtc tests to gt4py, ok?
```
jdahm, line 20: It looks like this is only used for `config.GT_INCLUDE_PATH` later in a test. Do we still need that with a pip package?

.gitignore
```diff
@@ -92,7 +92,7 @@ venv.bak/
 .idea/
 
 # VSCode project settings
-.vscode/
+/.vscode
```
jdahm, line 95: Is this only ignoring the top-level `.vscode` directory now?

.gitpod.Dockerfile
```diff
@@ -1,7 +1,8 @@
...
-
-COPY .gitpod/aliases.txt .
-RUN cat aliases.txt >> $HOME/.bashrc
+USER root
+RUN apt-get update \
+    && apt-get install -y libboost-dev \
+    && apt-get clean && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/* && rm -rf /tmp/*
+USER gitpod
+RUN pyenv install 3.9.7
+RUN pyenv global 3.9.7
```
jdahm, line 8: It's not a big deal, but some of your file edits are missing newlines at the end of the file. Can you change `Files: Insert Final Newline` in the settings UI (`files.insertFinalNewline` in JSON) to `true`.

requirements-dev.txt
```diff
@@ -28,3 +28,6 @@ sphinx~=4.4
 sphinx_rtd_theme~=1.0
 tox>=3.14
 devtools~=0.6
+# temporarily until we have a pypi package
+cmake
```
jdahm, line 32: Just wondering: My quick understanding was that `scikit-build` was used so that `cmake` was not required. Is that not the case?

tests/test_integration/test_cpp_regression.py
jdahm: I agree. We have other tests that check the specific output of stencil calls to validate the correctness.

.gitignore
havogt: Yes. There is the `.gitpod/.vscode` which I want to track with git (that's why putting the `/` to the front). Then, Gitpod will setup a symlink from `.vscode` to `.gitpod/.vscode`, therefore I have to remove the trailing `/` as otherwise the symlink is not ignored.

.gitpod.Dockerfile
havogt: This change was already merged in a PR that was included here. I'll still update it here.

tests/test_unittest/test_gtc/test_cuir_compilation.py
havogt: fixed.

tests/test_unittest/test_gtc/test_gtcpp_compilation.py
```diff
@@ -51,7 +45,7 @@ def build_gridtools_test(tmp_path: Path, code: str):
     ext_module = setuptools.Extension(
         "test",
         [str(tmp_src.absolute())],
-        include_dirs=[config.GT_INCLUDE_PATH, config.build_settings["boost_include_path"]],
+        include_dirs=[gridtools.get_include_dir(), config.build_settings["boost_include_path"]],
```
jdahm, line 48: Oh right, `config` is also needed here, so we can't totally eliminate this dependency. 👍

.github/workflows/gt4py-sphinx.yml
```diff
@@ -22,6 +22,8 @@ jobs:
       with:
         python-version: 3.8
     - name: Install dependencies
+      env:
+      PIP_EXTRA_INDEX_URL: https://test.pypi.org/simple
```
egparedes, line 26: We should remove PIP_EXTRA_INDEX_URL before merging to avoid security issues and using unstable versions by mistake (gridtools_cpp should have been already released in the official PyPI at that point).

DropD: For reference: the same env variable should be removed from `tox.ini` as soon as the official release is made.

havogt:
bors try

havogt:
bors try

src/gt4py/config.py
```diff
@@ -33,9 +35,7 @@
 
 
 GT_REPO_DIRNAME: str = "gridtools"
```
egparedes, line 37: This is not longer needed.

```suggestion
```

egparedes (approved):
It looks good. Just a minor unused variable which should be removed.

havogt:
bors try

PR #965, opened 2022-10-06
Functional: POC reduce complexity in FOAST lowering

tehrengruber:
Proof of concept implementation that drastically reduces the complexity of the lowering from FOAST to ITIR.

PR #1058, opened 2022-11-22
Fix can_deref in unroll_reduce

havogt:
can_deref should be applied on a shifted iterator, not a sparse field.

TODO:
- [x] add test


src/functional/iterator/transforms/unroll_reduce.py
```diff
@@ -16,70 +67,14 @@ class UnrollReduce(NodeTranslator):
...
-            ):
-                assert isinstance(arg.fun.args[-1], ir.OffsetLiteral), f"{arg.fun.args}"
-                connectivities.append(offset_provider[arg.fun.args[-1].value])
-            # if the argument is a lift call by itself look at its arguments recursively
-            elif (
-                isinstance(arg, ir.FunCall)
-                and isinstance(arg.fun, ir.FunCall)
-                and arg.fun.fun == ir.SymRef(id="lift")
-            ):
-                connectivities.extend(UnrollReduce._find_connectivities(arg.args, offset_provider))
```
havogt, line 36: This change was removed by merge conflict, have to think about if that actually did the right thing...

havogt: I think it didn't make sense.
```
def foo(it):
    return deref(shift(1)(it))

...
def bar(it):
    return reduce(plus, 0)(lift(foo)(shift(offset)(it)))
                                           ^^^^^^
```
it would detect marked offset as the partial shift for reduce which doesn't make sense. I don't see a case where the change introduced in #1055 makes sense...

havogt: Maybe I see now, let me think again...

havogt:
The connectivity detection part is not really nice. I am divided between rewriting it again or leaving it as we want to cleanup the partial shifts anyway...

src/functional/iterator/transforms/unroll_reduce.py
tehrengruber: This is meant for cases where `ìt` is not partially shifted, e.g. just a regular iterator as given to a stencil in a closure. I introduced it because the lift inliner now transforms `shift(...)(lift(...)(...))` (supported by `find_connectivity`) into `lift(...)(shift(...)(...))` (something that has been equivalent before already), which is not supported here.

```
reduce(plus, 0)(lift(deref)(shift("V2V")(it)))
```

tehrengruber:
I would not really invest time here as this is already broken conceptually. Retrieving an offset literal on a syntactic level from an arbitrary expression wasn't a good idea in the first place. Whether there are partial shifts or not `reduce(it, axis)` is a much more solid solution already implemented in #965.

src/functional/iterator/transforms/unroll_reduce.py
```diff
@@ -1,11 +1,98 @@
...
 
+def _is_shifted(arg: ir.Expr) -> TypeGuard[ir.FunCall]:
+    return (
+        isinstance(arg, ir.FunCall)
+        and isinstance(arg.fun, ir.FunCall)
+        and arg.fun.fun == ir.SymRef(id="shift")
+    )
+
+
+def _is_applied_lift(arg: ir.Expr) -> TypeGuard[ir.FunCall]:
```
tehrengruber, line 19: I like the name. We should rename / unify with the equivalents in the lift inliner.

PR #1059, opened 2022-11-25
Fix assigning function calls to subscripts

### LINKED ISSUE #520: Fix support for field offset arguments to functions
Offsets to functions seem to not be supported in gtscript, but would be immensely helpful in certain instances. Consider this reproducer:
```python
from gt4py import gtscript
import numpy as np


@gtscript.function
def func(a):
    return a


@gtscript.stencil(backend="gtc:gt:cpu_ifirst")
def stencil(
    input_field: gtscript.Field[gtscript.IJK, np.int32],
    output_field: gtscript.Field[gtscript.IJK, np.int32],
):
        output_field[0, 0, 0] = func(input_field[-1, 0, 0])
```

gronerl:
Fixes #520 

#520 was reproducible, but the cause was not the offset in the argument but rather in  the assignment. This fixes it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,7 +92,7 @@ venv.bak/
 .idea/
 
 # VSCode project settings
-.vscode/
+/.vscode
 
 # SublimeText project settings
 *.sublime-project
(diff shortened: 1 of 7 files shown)


egparedes (approved):
It looks good to me.

gronerl:
bors try

PR #1060, opened 2022-11-25
Allow function calls as function arguments

### LINKED ISSUE #248: Support function calls as arguments to functions
Return values from functions calls seem to not be supported as arguments to function calls themselves (#247 improves the error message). Should this be supported in the future? Presumably this would be a fairly minor change in the frontend only.

gronerl:
Would resolve #248 

Since calls are already allowed in expresions like `lhs = 1 + func()` it seems reasonable to allow them as call arguments as well (or not allow either) but I'm not 100% sure of the implications in cases of extents. A critical review is appreciated.

### DIFF
--- a/src/gt4py/frontend/gtscript_frontend.py
+++ b/src/gt4py/frontend/gtscript_frontend.py
@@ -483,7 +483,11 @@ class CallInliner(ast.NodeTransformer):
 
         # Replace returns by assignments in subroutine
         if target_node is None:
-            if len(assign_targets) > 1:
+            if any(
+                isinstance(nd.value, ast.Tuple)
+                for nd in ast.walk(call_ast)
+                if isinstance(nd, ast.Return)
+            ):
                 raise GTScriptSyntaxError("Function returning tuple used in expression.")
             target_node = ast.Name(
                 ctx=ast.Store(),
--- a/src/gt4py/frontend/gtscript_frontend.py
+++ b/src/gt4py/frontend/gtscript_frontend.py
@@ -488,7 +488,10 @@ class CallInliner(ast.NodeTransformer):
                 for nd in ast.walk(call_ast)
                 if isinstance(nd, ast.Return)
             ):
-                raise GTScriptSyntaxError("Function returning tuple used in expression.")
+                raise GTScriptSyntaxError(
+                    "Only functions with a single return value can be used in expressions, including as call arguments. "
+                    "Please assign the function results to symbols first."
+                )
             target_node = ast.Name(
                 ctx=ast.Store(),
                 lineno=node.lineno,
--- a/src/gt4py/frontend/gtscript_frontend.py
+++ b/src/gt4py/frontend/gtscript_frontend.py
@@ -416,18 +416,10 @@ class CallInliner(ast.NodeTransformer):
     ):
         call_name = gt_meta.get_qualified_name_from_node(node.func)
 
+        node.args = [self.visit(arg) for arg in node.args]
         if call_name in gtscript.IGNORE_WHEN_INLINING:
-            # Not a function to inline. Visit arguments and return as-is.
-            node.args = [self.visit(arg) for arg in node.args]
+            # Not a function to inline, return as-is.
             return node
-        elif any(
-            isinstance(arg, ast.Call) and arg.func.id not in gtscript.MATH_BUILTINS
-            for arg in node.args
-        ):
-            raise GTScriptSyntaxError(
-                "Function calls are not supported in arguments to function calls",
-                loc=nodes.Location.from_ast_node(node),
-            )
         elif call_name not in self.context or not hasattr(self.context[call_name], "_gtscript_"):
             raise GTScriptSyntaxError("Unknown call", loc=nodes.Location.from_ast_node(node))
 
--- a/src/gt4py/frontend/gtscript_frontend.py
+++ b/src/gt4py/frontend/gtscript_frontend.py
@@ -483,6 +483,8 @@ class CallInliner(ast.NodeTransformer):
 
         # Replace returns by assignments in subroutine
         if target_node is None:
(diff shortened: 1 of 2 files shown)


tests/test_unittest/test_gtscript_frontend.py
```diff
@@ -240,6 +240,59 @@ def definition_func(inout_field: gtscript.Field[float]):
                 definition_func, name=inspect.stack()[0][3], module=self.__class__.__name__
             )
 
+    def test_use_in_expr(self):
+        @gtscript.function
+        def func():
```
egparedes, line 245: Just a question outside of the scope of this PR: is it explicitly check that there are not recursive calls (or mutually recursive calls) inside function definitions?

egparedes (commented):
In principle it looks good to me, but I agree that there might be a lot of corners cases we're not yet aware of. I suggest to add more tests and maybe even add a warning in the inlining pass when we detect function calls in the arguments, explaining that this is a experimental feature that might fail (or  even be removed in the future)

gronerl:
bors try

tests/test_unittest/test_gtscript_frontend.py
```diff
@@ -240,6 +240,59 @@ def definition_func(inout_field: gtscript.Field[float]):
...
+        def definition_func(inout_field: gtscript.Field[float]):
+            from gt4py.__gtscript__ import PARALLEL, computation, interval
+
+            with computation(PARALLEL), interval(...):
+                inout_field = func_outer(func() + 1)
+
+        parse_definition(
+            definition_func, name=inspect.stack()[0][3], module=self.__class__.__name__
+        )
+
```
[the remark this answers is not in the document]
gronerl: I think both are already sufficiently covered in the tests.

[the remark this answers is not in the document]
gronerl: Basically, we have the former with function call, with expression, as well as function calls being used in expressions, just not the combination of all three.

The latter is not an issue since the higher dimensional fields don't behave any different in this case.

tests/test_unittest/test_gtscript_frontend.py
gronerl: I opened an issue for recursive calls. As we expected, it just ends up as an infinite recursion in the frontend.

egparedes (commented):
I agree with the implementation, just a comment to improve an error message.

src/gt4py/frontend/gtscript_frontend.py
```diff
@@ -491,6 +483,12 @@ def visit_Call(  # noqa: C901 # Cyclomatic complexity too high
 
         # Replace returns by assignments in subroutine
         if target_node is None:
+            if any(
+                isinstance(nd.value, ast.Tuple)
+                for nd in ast.walk(call_ast)
+                if isinstance(nd, ast.Return)
+            ):
+                raise GTScriptSyntaxError("Function returning tuple used in expression.")
```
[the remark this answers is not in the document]
gronerl: I think this is also orthogonal to the PR: The same questions would apply if it were the call is not in an expression/call argument. I'm not sure what you're suggesting regarding unpacking, can you make an example? Do you mean to tell the user to first assign the returns to a symbol before using said symbol in the expression?

tests/test_unittest/test_gtscript_frontend.py
```diff
@@ -293,6 +293,50 @@ def definition_func(inout_field: gtscript.Field[float]):
...
+            with computation(PARALLEL), interval(...):
+                inout_field = func_outer(func())
+
+        with pytest.raises(
+            gt_frontend.GTScriptSyntaxError, match="Function returning tuple used in expression."
+        ):
+            parse_definition(
+                definition_func, name=inspect.stack()[0][3], module=self.__class__.__name__
+            )
```
jdahm, line 338: This test sort of makes my brain hurt. Could we disallow multiple return values being used in a function call? I'd rather see the regular Python syntax here `inout_field = func_outer(*func())` if `func` has multiple return values.

What do you think?

tests/test_unittest/test_gtscript_frontend.py
```diff
@@ -240,6 +240,103 @@ def definition_func(inout_field: gtscript.Field[float]):
...
+
+        def definition_func(inout_field: gtscript.Field[float]):
+            from gt4py.__gtscript__ import PARALLEL, computation, interval
+
+            with computation(PARALLEL), interval(...):
+                inout_field = func_outer(func() + 1)
+
+        parse_definition(
+            definition_func, name=inspect.stack()[0][3], module=self.__class__.__name__
+        )
```
jdahm, line 294: Have you verified that this works when `func` has a few statements instead of just a return? I realize the tests below sort of do this, but they raise for other reasons so it might be a blind spot.

gronerl: I think these issues are not related to what this PR: We do not implement function calls here, rather that they are dealt with in the context of expressions. Which should e.g. not be affected by the complexity of the function. I'd propose opening a separate issue if there should be more test cases for functions in general.

jdahm: Sorry I probably wasn't clear. What I meant was that there is some tricky logic that unrolls functions and inserts them before the statement with the call. I hope that the `CallInliner` logic still works when the function call is embedded in a Call.

Your tests probably test that sufficiently, but it was just a thought as I went through.

jdahm (approved):

tests/test_unittest/test_gtscript_frontend.py
gronerl: Oh yes, the call inliner works recursively.

egparedes (approved):
It looks good to me too.

gronerl:
bors try

gronerl:
bors try

PR #1063, opened 2022-11-28
Allow for new syntax: field[J-1]

jdahm:
## Description

Allows for new-style syntax in field offsets in the frontend.

To do:
- [x] Add to documentation
- [x] Add more tests (see @egparedes comments below)
- [x] Remove dependence on `gtscript.Axis`
- [x] Allow `gtscript.Axis` as externals
- [x] `ASTEvaluator` should copy the context


src/gt4py/frontend/gtscript_frontend.py
```diff
@@ -1024,18 +1024,37 @@ def visit_Index(self, node: ast.Index):
...
+            for index_node in index_nodes:
+                for node in ast.walk(index_node):
+                    if isinstance(node, ast.Name) and node.id in axes:
+                        new_style_syntax = True
+                        break
+
+        if new_style_syntax:
+            index_dict = {axis: 0 for axis in axes}
+            for index_node in index_nodes:
+                axis_context = {axis: gtscript.Axis(axis) for axis in axes}
```
egparedes, line 1053: I would suggest to change the `ASTEvaluator` implementation to make a copy of the passed context to avoid modifying it, which would also imply the context can be created only once here:

```suggestion
            index_dict = {axis: 0 for axis in axes}
            axis_context = {axis: gtscript.Axis(axis) for axis in axes}
            for index_node in index_nodes:
```

src/gt4py/frontend/gtscript_frontend.py
```diff
@@ -1024,18 +1024,37 @@ def visit_Index(self, node: ast.Index):
...
+                value = gt_meta.ASTEvaluator.apply(index_node, axis_context)
+                assert isinstance(value, gtscript.Axis)
```
egparedes, line 1055: I guess `ASTEvaluator` may fail in unexpected ways depending on the user input, so I would a `try-except` block here to give meaningful error messages, since

src/gt4py/frontend/gtscript_frontend.py
```diff
@@ -1024,18 +1024,37 @@ def visit_Index(self, node: ast.Index):
         index = self.visit(node.value)
         return index
 
-    def _eval_index(self, node: ast.Subscript) -> Optional[List[int]]:
-        invalid_target = GTScriptSyntaxError(message="Invalid target in assignment.", loc=node)
-
+    def _eval_index(
+        self, node: ast.Subscript, axes: Optional[Set[str]] = None
```
egparedes, line 1028: The current implementation seems to only support passing "I", "J" or "K", so I would check that `axes` only contain valid axis names (at least in typing)
```suggestion
        self, node: ast.Subscript, axes: Optional[Set[Literal["I", "J", "K"]]] = None
```
One enhancement we could consider though, is supporting external symbols in the offset, which would work as an alias to one of the cartesian axes:

```python
@stencil(... externals={"AXIS": gtscript.I})
def stencil(field):
    ....
    return field[AXIS - 1] + field[AXIS + 1]
```

tests/test_unittest/test_gtscript_frontend.py
```diff
@@ -241,6 +229,37 @@ def definition_func(inout_field: gtscript.Field[float]):
             )
 
 
+class TestAxisSyntax:
+    def test_good_syntax(self):
+        def definition_func(in_field: gtscript.Field[float], out_field: gtscript.Field[float]):
+            with computation(PARALLEL), interval(...):
+                out_field = in_field[I + 1, J - 1]
```
egparedes, line 236: I would also test indices without offset like just `I`, 0-offsets like `J + 0` or `K - 0` and offsets where the indices are not in the same order of the field (which I guess it should be supported) like `field[J-1,K+1,I]`

tests/test_unittest/test_gtscript_frontend.py
```diff
@@ -241,6 +229,37 @@ def definition_func(inout_field: gtscript.Field[float]):
...
+
+        parse_definition(
+            definition_func, name=inspect.stack()[0][3], module=self.__class__.__name__
+        )
+
+    def test_bad_mul_syntax(self):
```
egparedes, line 242: I would also test using indices in fields without that axis and using repeated axis names like `field[I+1, I+3]`

src/gt4py/gtscript.py
```diff
@@ -523,12 +523,13 @@ def __len__(self):
 
 # GTScript builtins: domain axes
 class Axis:
```
egparedes, line 525: I think reusing the `Axis` class as a kind of `ShiftedAxis` is conceptually wrong. I think it would be better to have to different classes because they express actually different things, but maybe it is overkill. Can you explain your motivation to use this strategy here?

egparedes (commented):
Thanks for working on this. I have some comments but overall looks good.

src/gt4py/frontend/gtscript_frontend.py
jdahm: Thanks. I had this thought last night too. One of the key benefits of this is the ability to do that, as well as writing generic code that can work with 1d, 2d, or 3d fields, e.g. in functions.

src/gt4py/gtscript.py
jdahm: I used it for the simplicity, but sure, I can change it. Actually we only need it internally anyway so it doesn't need to be exposed in `gtscript.py`.

egparedes (commented):
Almost ready to go. I have just a question about the typing.

src/gt4py/frontend/gtscript_frontend.py
```diff
@@ -1027,33 +1029,97 @@ def visit_Index(self, node: ast.Index):
...
+            else:
+                axis_index = all_spatial_axes.index(value.name)
+                if axis_index < 0:
+                    raise GTScriptSyntaxError(
+                        message=f"Unrecognized axis: {value.name}", loc=index_node
+                    )
+                elif axis_index < last_index:
+                    raise GTScriptSyntaxError(
+                        message=f"Axis {value.name} is specified out of order", loc=index_node
+                    )
```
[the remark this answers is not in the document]
jdahm: Yeah, I'm for keeping this right now and we can remove later if it becomes a problem. I've virtually never seen a field offset on two axes.

src/gt4py/frontend/gtscript_frontend.py
```diff
@@ -1027,33 +1029,97 @@ def visit_Index(self, node: ast.Index):
...
     def visit_Subscript(self, node: ast.Subscript):
         assert isinstance(node.ctx, (ast.Load, ast.Store))
 
-        index = self._eval_index(node)
         result = self.visit(node.value)
+        if isinstance(result, nodes.FieldRef):
+            axes = set(result.offset.keys())
+        else:
+            axes = None
+        index = self._eval_index(node, axes)
```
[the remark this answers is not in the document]
jdahm: Ah, this is because the file is ignored by mypy. I'll add an issue to fix this later.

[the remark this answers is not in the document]
jdahm: Tracked in https://github.com/GridTools/gt4py/issues/1066.

jdahm:
bors try

jdahm:
@egparedes is this ready to go?

egparedes (approved):
It looks good.

PR #1072, opened 2022-12-02
`import attrs` requires attrs 21.3+

jdahm:
Found that the lower bound version specified in `setup.cfg` was too low to support `import attrs`.


jdahm:
bors try

egparedes (approved):
It looks good.

Issue #1012, opened 2022-10-25
Deprecate use of __INLINED


jdahm:
`__INLINED` looks like a compiler internal reserved word.

jdahm:
We don't have to remove it for the release, but we should consider removing it in the future. The steps should be:
- Make an optimization pass to automatically convert if conditions to compile-time removal
- Mark it as deprecated and remove it from the public documentation
- Remove it and see if users complain
- Remove it entirely.

jdahm:
I did some more thinking about this today, and had some realizations:

Deprecating `__INLINE` would mean that there would no longer be a way to specify “dead code” in a stencil, and therefore there would be no reasonable situation where passing `None` as a field value to a stencil makes sense.

Therefore we would also have to deprecate and remove the ability to give fields default values of `None` in the definition function type hint* and the ability to pass `None` as for instance [this](https://github.com/GridTools/gt4py/blob/master/tests/test_unittest/test_call_interface.py#L193) test does.

*: This behavior was “broken”/inconsistent anyway, since the type hint did not specify `Optional[Field[...]] = None`, rather it took the form `Field[...] = None`, which is wrong.

I believe this will break @stubbiali's workflow, but it will clean up this very messy business parameter matching after pruning, since there will be no pruning!

stubbiali:
Thanks @jdahm for making me aware of this development path. In principle, I don't have anything against dropping the `__INLINED` construct. Removing the ability to make fields default to None is a surely more problematic for Tasmania, but nothing which cannot be worked-around. At worst, we can always pin GT4Py to an older commit. No problem for FVM though, as we don't make use of either feature.

stubbiali:
However, according to your first bullet point above:

> - Make an optimization pass to automatically convert if conditions to compile-time removal

it sounds like you aim to remove the `__INLINED` syntax, but not the ability to have and detect dead code, should the if-condition be evaluable at compile-time. So do we really need to prevent fields from defaulting to None? I'm not disputing the usefulness of such a change, but rather the connection between the two features.

havogt:
Was deprecated in #1068

PR #1073, opened 2022-12-05
Remove __INLINED and ability to set fields to a default of None

jdahm:
## Description

Resolves https://github.com/GridTools/gt4py/issues/1012.

This is a work in progress.


havogt:
Closing. `__INLINED` is deprecated, but we don't plan to remove it for now.

PR #1074, opened 2022-12-06
Fix mypy foast/type_deduction

nfarabullini:
Un-ignore `foast_passes/type_deduction` and fix all the reported mypy errors


samkellerhals (approved):
I added a `todo` regarding the `Starred` node, other than that it looks good!

PR #1078, opened 2022-12-08
Scalar if

tehrengruber:
```python
def both_branches_return(inp: Field[[IDim, JDim], float64], condition1: bool, condition2: bool):
    if condition1:
        tmp1 = inp
        if condition2:
            return tmp1 + 1.0
        result = tmp1 + 2.0
    else:
        result = inp + 3.0
    return result
```
Including extended SSA pass supporting Ifs.

Superseeds #972


tehrengruber:
Closed. Picked the wrong source branch...

PR #1079, opened 2022-12-08
feature[next]: Scalar if

tehrengruber:
```python
def both_branches_return(inp: Field[[IDim, JDim], float64], condition1: bool, condition2: bool):
    if condition1:
        tmp1 = inp
        if condition2:
            return tmp1 + 1.0
        result = tmp1 + 2.0
    else:
        result = inp + 3.0
    return result
```
~Including extended SSA pass supporting Ifs.~

Superseeds #972


tehrengruber:
@egparedes This PR is ready for review (after #1250 is merged). The files `single_static_assign.py` and `test_single_static_assign.py` only appear here until we merge #1250.

src/gt4py/next/iterator/embedded.py
```diff
@@ -470,7 +470,9 @@ def promote_scalars(val: CompositeOfScalarOrField):
     elif isinstance(val, LocatedField):
         return val
     val_type = infer_dtype_like_type(val)
-    if np.issubdtype(val_type, np.number):
+    if np.issubdtype(val_type, np.number) or (
+        isinstance(val_type, type) and issubclass(val_type, bool)
+    ):
```
egparedes, line 475: Following https://numpy.org/doc/stable/reference/arrays.scalars.html, I would rewrite the condition in this way:
```suggestion
    if np.issubdtype(val_type, np.number) or (
        np.issubdtype(val_type, np.bool_)
    ):
```
Actually, it might be interesting to factor this out as its own function, and use it everywhere in the embedded implementation. I have the feeling that right now there are several places using slightly wrong variations of this same check.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_scalar_if.py
```diff
@@ -0,0 +1,405 @@
...
+    )
+
+
+@pytest.mark.parametrize("condition1, condition2", [[True, False], [True, False]])
+def test_nested_if_stmt_conditinal(cartesian_case, condition1, condition2):
+    if cartesian_case.backend in [gtfn_cpu.run_gtfn, gtfn_cpu.run_gtfn_imperative]:
+        pytest.skip("If-stmts are not supported yet.")
+
+    @field_operator
+    def both_branches_return(inp: cases.IField, condition1: bool, condition2: bool) -> cases.IField:
```
egparedes, line 252: Is this function's name correct?

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -132,6 +138,64 @@ def visit_BlockStmt(
         assert inner_expr
         return inner_expr
 
+    def visit_IfStmt(
+        self, node: foast.IfStmt, *, inner_expr: itir.Expr | None, **kwargs
```
egparedes, line 142: I think the exact meaning of `inner_expr` should be documented somewhere, because it is not so easy to understand completely by looking at the codebase for the first time.

egparedes (changes requested):
The main parts look good, just have some questions and minor comments about style. I omitted the `eve` modifications in this review, which I'll address later in another review round.

tehrengruber (commented):
Forgot to submit my comments...

src/gt4py/next/iterator/embedded.py
```diff
@@ -470,7 +472,7 @@ def promote_scalars(val: CompositeOfScalarOrField):
     elif isinstance(val, LocatedField):
         return val
     val_type = infer_dtype_like_type(val)
-    if np.issubdtype(val_type, np.number):
+    if isinstance(val, Scalar):  # type: ignore # mypy bug
```
egparedes, line 475: Please add a reference to the bug/issue description or a very short description of what's the problem here for mypy, otherwise the comment is not very useful.

src/gt4py/eve/traits.py
```diff
@@ -59,7 +59,13 @@ def visit_Node(self, node: concepts.Node, /) -> None:
                         )
                     self.collected_symbols[symbol_name] = node
 
-            if not isinstance(node, SymbolTableTrait):
+            if hasattr(node.annex, "propagated_symbols"):
```
egparedes, line 62: Please add a TODO here for me to review and generalize the mechanism to solve name collisions in the `SymbolTableTrait`

egparedes (commented):
A couple of comments and it's ready to go.

PR #1080, opened 2022-12-09
Fix mypy func to foast

nfarabullini:
Un-ignore `func_to_foast` and fix all the reported mypy errors


src/functional/ffront/func_to_foast.py
```diff
@@ -113,9 +113,9 @@ def _postprocess_dialect_ast(
             )
             # TODO(tehrengruber): use `type_info.return_type` when the type of the
             #  arguments becomes available here
-            if annotated_return_type != foast_node.type.returns:
+            if annotated_return_type != foast_node.type.returns:  # type: ignore
                 raise common.GTTypeError(
-                    f"Annotated return type does not match deduced return type. Expected `{foast_node.type.returns}`"
+                    f"Annotated return type does not match deduced return type. Expected `{foast_node.type.returns}`"  # type: ignore
```
egparedes, line 118: What's the actual problem? It looks like nothing should be ignored here.

According to the [coding guidelines](https://github.com/GridTools/gt4py/blob/functional/CODING_GUIDELINES.md#ignoring-qa-errors), ignoring QA messages should only be used as last resort and they should come accompanied with a small comment to explain what's happening. Additionally, recent mypy version provide a code error that should be used in the type ignore comment to avoid catching all errors indiscriminately (e.g. `type: ignore[attr-defined]  # foo.bla is always added dynamically `)

egparedes (commented):
It looks good. Just one question about the ignored errors.

src/functional/ffront/func_to_foast.py
nfarabullini: I put `type: ignore` here because of the TODO above, which will change the condition syntax, otherwise pre-commit outputs the following error: 
`error: Item "DeferredSymbolType" of "Union[FunctionType, DeferredSymbolType]" has no attribute "returns"`

I do agree on writing a more explicating message

egparedes: Ok, makes sense. Just make sure that the proper mypy error code is used in the ignore and that you briefly mention that is related to the above TODO and  `foast_node.type.returns

egparedes (approved):
It looks good.

PR #1089, opened 2022-12-15
Fix mypy past_to_itir typing

nfarabullini:
Un-ignore `past_to_itir` and fix all the reported mypy errors


src/functional/ffront/past_to_itir.py
```diff
@@ -178,37 +180,38 @@ def _construct_itir_out_arg(
...
         out_field: past.Name,
-        node_domain: Optional[past.Dict],
+        node_domain: Optional[past.Expr],
         slices: Optional[list[past.Slice]] = None,
-    ):
+    ) -> itir.FunCall:
         domain_args = []
 
         out_field_types = type_info.primitive_constituents(out_field.type).to_list()
+        out_field_types0_dims = cast(ts.FieldType, out_field_types[0]).dims
```
samkellerhals, line 197: Could this be named more clearly? Maybe something like `first_out_field_type_dims`  ?

samkellerhals (commented):
Overall looks good to me.

samkellerhals (approved):

src/functional/ffront/past_to_itir.py
nfarabullini: that is an even longer name...

PR #1093, opened 2022-12-19
Replace deprecated numpy types

petiaccja:


src/functional/iterator/embedded.py
```diff
@@ -483,6 +483,12 @@ def promote_scalars(val: CompositeOfScalarOrField):
     if math_builtin_name == "gamma":
         # numpy has no gamma function
         impl = np.vectorize(math.gamma)
+    elif math_builtin_name in ["int", "float", "bool", "str"]:
+        # TODO: Using python builtins instead of numpy fixed size types to keep
+        # behaviour the same. Maybe change in the future.
+        import builtins as python_builtins  # noqa: T100 functional.iterator.builtins shadows builtins
```
nfarabullini, line 489: you already import builtins on line 34

gronerl:
Haven't looked at this too closely, but make sure that this remains backwards compatible. (I.e. if a `np.bool` from some user code makes it to the changed points in your PR, will the behavior be the same?)

src/functional/iterator/embedded.py
petiaccja: That's precisely why this is needed. I have to access the Python `builtins`, not `functional.iterator.builtins`, which is being imported there.

petiaccja: But I guess it's also possible to import `functional.iterator.builtins` as an alias, or add 4 special cases instead of this blanket one.

src/functional/iterator/embedded.py
```diff
@@ -479,10 +479,13 @@ def promote_scalars(val: CompositeOfScalarOrField):
 
 
 for math_builtin_name in builtins.MATH_BUILTINS:
+    special_builtins = {"int": int, "float": float, "bool": bool, "str": str}
```
tehrengruber, line 482:
```suggestion
    python_builtins = {"int": int, "float": float, "bool": bool, "str": str}
```

tehrengruber (approved):
Looks good.

src/functional/iterator/embedded.py
tehrengruber: The solution you picked in the end is fine I think. In principle the `functional.iterator.builtins` should not have been called `builtins` to avoid the shadowing you stumbled upon here.

src/functional/iterator/embedded.py
petiaccja: Done

PR #1097, opened 2022-12-20
Update documentation

havogt:


docs/commandline.rst
```diff
@@ -90,11 +90,10 @@ each stencil are put in a subfolder named after the stencil:
 The line
 
 .. code-block:: python
-   :caption: my_stencils.gt.py
 
    # [GT] using-dsl: gtscript
 
-is equivalent to writing
+in ``my_stencils.gt.py`` is equivalent to writing
```
gronerl, line 96: Not really this PR, but do we want to propose the`.gt.py` ending?

docs/quickstart.rst
```diff
@@ -20,7 +20,7 @@ We strongly recommended to create a virtual environment for any new project:
     pip install --upgrade pip wheel setuptools
 
 
-Then clone the GT4Py repository and install the local copy or install it directly from GitHub with `pip`.
+Then clone the GT4Py repository and install the local copy or install it directly from pypi.org.
```
gronerl, line 23:
```suggestion
Then clone the GT4Py repository and install the local copy or install it directly from PyPI: `pip install gt4py`
```

gronerl: Ok you say it below, but the section mentions a bunch of things and only shows one. maybe at least make it clear that pip==PyPI. I don't think all our users know ;)

setup.cfg
```diff
@@ -106,7 +106,7 @@ testing =
 [options.entry_points]
 # CLI entry points to be installed
 console_scripts =
-    gtpyc = gt4py.cli:gtpyc
+    gtpyc = gt4py.cartesian.cli:gtpyc
```
gronerl, line 109: We now use gt4pyc as shorthand for `gt4py.cartesian` in imports. I don't know a better naming but the names are very close.

gronerl (commented):
Some points to discuss

setup.cfg
havogt: yeah, I agree, but wouldn't change now

docs/commandline.rst
havogt: I don't know. Probably not.

gronerl (approved):
I'll open a separate PR removing the `.gt.py` ending from examples in docs.

gronerl:
Re-ran seemingly spuriously failed tox test.

docs/commandline.rst
egparedes: I think it's actually required by the CLI. I vaguely remember that Rico used some `importlib` features using the extension to implement the functionality

egparedes:
bors try

docs/commandline.rst
havogt: yes, I also tested it with a file that had just .py suffix.

PR #1104, opened 2022-12-21
Fix TestExecInfo failures in daint-ci

gronerl:
This PR changes the build mechanism of TestExecInfo so that it builds the stencil in separate folders per worker.

### DIFF
--- a/tests/cartesian_tests/unit_tests/test_exec_info.py
+++ b/tests/cartesian_tests/unit_tests/test_exec_info.py
@@ -12,6 +12,7 @@
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+import inspect
 from typing import Any, Dict
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


gronerl:
bors try

gronerl:
bors try-
bors try

gronerl:
bors try-
bors try

havogt:
bors try-
bors try

tests/cartesian_tests/unit_tests/test_exec_info.py
```diff
@@ -69,9 +70,18 @@ def diffusion_def(
                 flux_x[0, 0, 0] - flux_x[-1, 0, 0] + flux_y[0, 0, 0] - flux_y[0, -1, 0]
             )
 
-    def compile_stencils(self, backend):
-        self.advection = gtscript.stencil(backend=backend, definition=self.advection_def)
-        self.diffusion = gtscript.stencil(backend=backend, definition=self.diffusion_def)
+    def compile_stencils(self, backend, worker_id):
+        modulename = inspect.currentframe().f_back.f_globals["__name__"]
```
egparedes, line 74: Just a question: is this really needed? I guess any name could be used as prefix, isnt'it?

tests/cartesian_tests/unit_tests/test_exec_info.py
```diff
@@ -69,9 +70,18 @@ def diffusion_def(
                 flux_x[0, 0, 0] - flux_x[-1, 0, 0] + flux_y[0, 0, 0] - flux_y[0, -1, 0]
             )
 
-    def compile_stencils(self, backend):
-        self.advection = gtscript.stencil(backend=backend, definition=self.advection_def)
-        self.diffusion = gtscript.stencil(backend=backend, definition=self.diffusion_def)
+    def compile_stencils(self, backend, worker_id):
```
egparedes, line 73: We could document here , that `worker_id` comes from pytest-xdist but it's not critical

egparedes (approved):
It looks good, I've only have   question and a comment, both are non-blocking.

tests/cartesian_tests/unit_tests/test_exec_info.py
gronerl: Sure. This gives some additional guarantee that it is unique. It is also exactly what the default is if you don't provide `name` (It's a copy paste from there.)

tests/cartesian_tests/unit_tests/test_exec_info.py
gronerl: Alternatively, there is also the option to not base `name` on `worker_id` but instead just use different names for test_backcompatibility and test_accumulate.

PR #1106, opened 2022-12-22
Remove ellipsis as (generic) field dimension 

petiaccja:
*Edit @tehrengruber*
Very early in the development of the frontend the ellipsis, i.e. `...`, was used to signify that a field defined on any dimensions is permitted as an input to a field operator or program. The concept was never fully supported and at least partially broken. While fixing typing errors this became a burden and it was decided to remove the concept until we introduce a more powerful type system that is capable of this in a proper way.
```python
def field_op(in_field: Field[..., float64]) -> Field[..., float64]:
  ...
```

### DIFF
--- a/src/functional/ffront/type_info.py
+++ b/src/functional/ffront/type_info.py
@@ -16,7 +16,7 @@ from functools import reduce
 from typing import Iterable, Iterator, cast
 
 import functional.ffront.type_specifications as ts
-from functional.common import GTTypeError, Dimension
+from functional.common import Dimension, GTTypeError
 from functional.type_system.type_info import (  # noqa: F401
     accepts_args as accepts_args,
     apply_to_primitive_constituents as apply_to_primitive_constituents,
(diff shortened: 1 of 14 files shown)


src/functional/ffront/type_info.py
```diff
@@ -117,7 +117,7 @@ def _as_field(dtype: ts.ScalarType, path: tuple[int, ...]) -> ts.FieldType:
                 # The structure of the scan passes argument and the requested
                 # argument type differ. As such we can not extract the dimensions
                 # and just return a generic field shown in the error later on.
-                return ts.FieldType(dims=..., dtype=dtype)
+                return ts.FieldType(dims=[], dtype=dtype)
```
tehrengruber, line 120:
```suggestion
                # TODO: we want some generic field type here, but our type system does not support it yet.
                return ts.FieldType(dims=Dimension("..."), dtype=dtype)
```
How about this one? Or pick another name that 1) signifies this can be any field type 2) does not collide with a user picked one easily.

tehrengruber (changes requested):
Looks good only minor thing.

tehrengruber (approved):

PR #1107, opened 2022-12-23
Cartesian: Partial expansion

gronerl:
This PR adds "partial expansion" functionality for the dace backends (currently applied only on CPU) in IJ dimensions.
This means that full gtscript stencils or larger scopes in an orchestrated context can be tiled with a single map.


gronerl:
bors try

gronerl:
bors try

gronerl:
bors try-
bors try

PR #1113, opened 2023-01-03
Fix mypy itir_to_gtfn_ir typing

nfarabullini:
Un-ignore `itir_to_gtfn_ir` and fix all the reported mypy errors


src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -419,7 +425,7 @@ def visit_FunctionDefinition(
     def _is_scan(node: itir.Node):
         return isinstance(node, itir.FunCall) and node.fun == itir.SymRef(id="scan")
 
-    def _visit_output_argument(self, node: itir.SymRef | itir.FunCall):
+    def _visit_output_argument(self, node: itir.Expr | itir.FunCall):
```
havogt, line 428:
```suggestion
    def _visit_output_argument(self, node: itir.Expr):
```
since itir.FunCall is an expression already.

src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -149,17 +153,18 @@ def _collect_offset_definitions(
             raise ValueError(f"Missing offset_provider entry for {o.value}")
 
         offset_name = o.value
-        if isinstance(offset_provider[o.value], common.Dimension):
-            dim = offset_provider[o.value]
+        if isinstance(offset_provider[offset_name], common.Dimension):
```
havogt, line 156:
```suggestion
        dim_or_conn = offset_provider[offset_name]
        if isinstance(dim_or_conn, common.Dimension):
``` 
then can undo the other changes and don't need ignore

src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -135,7 +139,7 @@ def _collect_dimensions_from_domain(
 def _collect_offset_definitions(
     node: itir.Node,
     grid_type: common.GridType,
-    offset_provider: dict[str, common.Dimension | common.Connectivity],
+    offset_provider: dict[str | int, common.Dimension | common.Connectivity],
```
havogt, line 142: Undo this and add `assert isinstance(o.value, str)` after `for o in offset_tags:`
It can only be str as there is the `.filter`, but mypy doesn't understand it.

src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -69,10 +69,10 @@ def pytype_to_cpptype(t: str):
 _horizontal_dimension = "gtfn::unstructured::dim::horizontal"
 
 
-def _get_domains(closures: Iterable[itir.StencilClosure]) -> Iterable[itir.FunCall]:
+def _get_domains(closures: Iterable[itir.StencilClosure]) -> list[itir.FunCall]:
```
havogt, line 72: replace by just
```
def _get_domains(closures: Iterable[itir.StencilClosure]) -> Iterable[itir.FunCall]:
    return (c.domain for c in closures)
```
and fix `itir.StencilClosure` `domain` to be `itir.FunCall` in `ir.py`. It seems we heavily rely on that...

havogt (changes requested):
Some improvements to get rid of the `#type: ignore`s.

havogt (approved):

PR #1117, opened 2023-01-05
Fix mypy gtfn_backend

nfarabullini:
Un-ignore `gtfn_backend` and fix all the reported mypy errors


havogt:
I would undo your last change and instead pass an empty offset_provider in copy_stencil.py.

havogt (approved):

PR #1119, opened 2023-01-09
GTFN codegen: unapplied operator as function object

havogt:


tests/functional_tests/iterator_tests/test_trivial.py
```diff
@@ -167,3 +167,30 @@ def test_vertical_shift_unstructured(program_processor):
 
     if validate:
         assert np.allclose(inp_s[:, 1:], np.asarray(out_s)[:, :-1])
+
+
+# TODO We should bypass all passes to ensure that we actually test this code as is
```
havogt, line 172: Currently, this test does not test an non-applied operator because we aggressively inline and we don't have easy access to disable passes.

src/functional/program_processors/codegens/gtfn/codegen.py
```diff
@@ -124,7 +124,12 @@ def visit_OffsetLiteral(self, node: gtfn_ir.OffsetLiteral, **kwargs: Any) -> str
     )
 
     def visit_FunCall(self, node: gtfn_ir.FunCall, **kwargs):
-        return self.generic_visit(node, fun_name=self.visit(node.fun))
+        if isinstance(node.fun, gtfn_ir.SymRef) and node.fun.id in self.user_defined_function_ids:
+            fun_name = f"{self.visit(node.fun)}{{}}()"
+        else:
+            fun_name = self.visit(node.fun)
+
+        return self.generic_visit(node, fun_name=fun_name)
```
havogt, line 132: These changes enable calling a user-defined function from a stencil, we didn't/don't have a test for this because we always inline.

src/functional/program_processors/codegens/gtfn/codegen.py
```diff
@@ -124,7 +124,12 @@ def visit_OffsetLiteral(self, node: gtfn_ir.OffsetLiteral, **kwargs: Any) -> str
     )
 
     def visit_FunCall(self, node: gtfn_ir.FunCall, **kwargs):
-        return self.generic_visit(node, fun_name=self.visit(node.fun))
+        if isinstance(node.fun, gtfn_ir.SymRef) and node.fun.id in self.user_defined_function_ids:
+            fun_name = f"{self.visit(node.fun)}{{}}()"
```
havogt, line 128: User-defined functions are structs with call operator returning the actual function, i.e. `{}` for creating the instance `()` for the call operator, see below.

src/functional/program_processors/codegens/gtfn/gtfn_ir.py
```diff
@@ -155,6 +167,23 @@ class TemporaryAllocation(Node):
...
+    "std::less<void>{}",
+    "std::less_equal<void>{}",
+    "std::greater<void>{}",
+    "std::greater_equal<void>{}",
+    "std::logical_and<void>{}",
+    "std::logical_or<void>{}",
+    "std::bit_xor<void>{}",
+    "std::modulos<void>{}",
+    "std::logical_not<void>{}",
+]
```
petiaccja, line 186: As we discussed, it might make sense to not use `SymRef`s and create a new node such as `StlFunctor` or something similar. That could reduce modifications and special cases to the existing code. `SymRef`s also don't benefit code generation much, whether a symbol is defined or not is verified by the C++ compiler anyway.

petiaccja (changes requested):

src/functional/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -197,6 +197,23 @@ class GTFN_lowering(eve.NodeTranslator, eve.VisitorWithSymbolTableTrait):
...
+        "not_eq": "std::not_equal_to<void>{}",
+        "less": "std::less<void>{}",
+        "less_equal": "std::less_equal<void>{}",
+        "greater": "std::greater<void>{}",
+        "greater_equal": "std::greater_equal<void>{}",
+        "and_": "std::logical_and<void>{}",
+        "or_": "std::logical_or<void>{}",
+        "xor_": "std::bit_xor<void>{}",
+        "mod": "std::modulos<void>{}",
+        "not": "std::logical_not<void>{}",
```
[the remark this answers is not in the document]
havogt: Good point. I missed that ctad works here and we can omit the `<>`!

tests/functional_tests/iterator_tests/test_trivial.py
```diff
@@ -167,3 +167,30 @@ def test_vertical_shift_unstructured(program_processor):
 
     if validate:
         assert np.allclose(inp_s[:, 1:], np.asarray(out_s)[:, :-1])
+
+
+def test_unapplied_op():
```
petiaccja, line 172: Can you parameterize this test for `[plus, minus, divides, ...]`? Otherwise we can't really know if you used the correct C++ function objects.

petiaccja (changes requested):

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -41,6 +41,7 @@ class GTFNTranslationStep(
     step_types.TranslationStep[languages.Cpp, languages.LanguageWithHeaderFilesSettings],
 ):
     language_settings: languages.LanguageWithHeaderFilesSettings = cpp_interface.CPP_DEFAULT
+    apply_transforms: bool = True
```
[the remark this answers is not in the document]
DropD: This is okay, customization option exactly in the step that uses it

src/functional/program_processors/runners/gtfn_cpu.py
```diff
@@ -41,6 +41,7 @@ class GTFNExecutor(ppi.ProgramExecutor):
     builder_factory: compiler.BuildSystemProjectGenerator = compiledb.CompiledbFactory()
 
     name: Optional[str] = None
+    apply_transforms: bool = True
```
DropD, line 44: This we need to talk about. If this is really a regular feature of the backend it should be here. If we only need it internally for testing, I would think about building a no-transforms version of this backend in the tests package.

If this is a customization feature that needs to be conveniently available, that should be reflected in the docstring. It should not appear as if any and all step customization options have to be added to every backend.

Pro separate backend:
- does not affect complexity of the public API (think about how the test matrix grows with every customization option)
- uses the modularity of the workflows to full effect.

Cons:
- if needed by many users often, leads to unneccessary boiler plate
- a certain amount of code duplication (potentially limited to the tests though)
- additional work (copy paste the `__call__` method, change one line, give it a name)

src/functional/program_processors/codegens/gtfn/gtfn_backend.py
```diff
@@ -22,19 +22,19 @@
 from functional.program_processors.codegens.gtfn.itir_to_gtfn_ir import GTFN_lowering
 
 
-def generate(program: itir.FencilDefinition, **kwargs: Any) -> str:
-    transformed = program
+def generate(program: itir.FencilDefinition, apply_transforms: bool = True, **kwargs: Any) -> str:
```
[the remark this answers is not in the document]
havogt: We should introduce a more flexible pass manager, in Cartesian for example we allow to pass a custom optimization pipeline. Maybe the simple option here would be to pass a callable here instead of the boolean which defaults to the current `apply_common_transforms`. But even that feels slightly out of scope for this PR.

src/functional/program_processors/runners/gtfn_cpu.py
havogt: The main problem here is that actually the passes should not be part of the backend at all, they should run on any backend. I am not sure what the best strategy for this PR is. To me everything beyond this strategy seems to be out of scope. Not so nice alternative, would be to merge the feature of this PR without test.

src/functional/program_processors/runners/gtfn_cpu.py
```diff
@@ -82,3 +83,6 @@ def __name__(self) -> str:
 
 
 run_gtfn: Final[ppi.ProgramProcessor[None, ppi.ProgramExecutor]] = GTFNExecutor(name="run_gtfn")
+run_gtfn_no_transforms: Final[ppi.ProgramProcessor[None, ppi.ProgramExecutor]] = GTFNExecutor(
```
havogt, line 86: We decided to move this part to the tests, and annotate the apply_transforms flag linking to an issue with the desired behavior as outlined above.

tests/functional_tests/iterator_tests/test_trivial.py
havogt: done. removed this test again and moved to test_builtins

src/functional/program_processors/codegens/gtfn/gtfn_backend.py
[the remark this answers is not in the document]
havogt: I'll leave it like that and opened an issue.

src/functional/program_processors/codegens/gtfn/codegen.py
```diff
@@ -64,6 +64,21 @@ class GTFNCodegen(codegen.TemplatedGenerator):
...
+        "not_eq": "std::not_equal_to{}",
+        "less": "std::less{}",
+        "less_equal": "std::less_equal{}",
+        "greater": "std::greater{}",
+        "greater_equal": "std::greater_equal{}",
+        "and_": "std::logical_and{}",
+        "or_": "std::logical_or{}",
+        "xor_": "std::bit_xor{}",
+        "mod": "std::modulos{}",
+        "not": "std::logical_not{}",
```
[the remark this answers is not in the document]
havogt: discovered another typo with the better test coverage

src/functional/program_processors/codegens/gtfn/gtfn_backend.py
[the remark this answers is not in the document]
petiaccja: Okay, that sounds good to me.

src/functional/program_processors/runners/gtfn_cpu.py
```diff
@@ -41,6 +41,7 @@ class GTFNExecutor(ppi.ProgramExecutor):
     builder_factory: compiler.BuildSystemProjectGenerator = compiledb.CompiledbFactory()
 
     name: Optional[str] = None
+    apply_transforms: bool = True  # TODO replace by more general mechanism, see https://github.com/GridTools/gt4py/issues/1135
```
petiaccja, line 44: I'd consider renaming this variable. `apply_transforms` is obvious if you've already seen the implementation, but if you come from the top it might not be clear what transforms are applied to what. Transforming the arguments to column-major memory layout would also be possible.
```suggestion
    apply_itir_transforms
    apply_itir_passes
    enable_itir_passes
    enable_itir_optimizations
    enable_optimizations
```

tests/functional_tests/iterator_tests/test_builtins.py
```diff
@@ -141,6 +166,26 @@ def test_arithmetic_and_logical_builtins(program_processor, builtin, inputs, exp
         assert np.allclose(np.asarray(out), expected)
 
 
+@pytest.mark.parametrize("builtin, inputs, expected", arithmetic_and_logical_test_data())
+def test_unapplied_arithmetic_and_logical_builtins_gtfn(builtin, inputs, expected):
```
petiaccja, line 170: Unapplied sounds a bit weird to me:
```suggestion
def test_arithmetic_and_logical_functors_gtfn(builtin, inputs, expected):
```

petiaccja (changes requested):

src/functional/program_processors/runners/gtfn_cpu.py
havogt: I chose `enable_itir_transforms`

tests/functional_tests/iterator_tests/test_builtins.py
havogt: I like version more, but not enough to start a fight. ;)

petiaccja (approved):

PR #1120, opened 2023-01-09
feat[next]: use context vars instead of global state in embedded iterator execution

havogt:
Changed:
- In the iterator embedded execution, both `column_range` and `offset_provider` global variables have been converted to context variables and fencil closures are now executed in isolated contexts with properly initialized values, to avoid using global state.

### DIFF
--- a/src/gt4py/next/iterator/embedded.py
+++ b/src/gt4py/next/iterator/embedded.py
@@ -186,9 +186,7 @@ class MutableLocatedField(LocatedField, Protocol):
 #: Column range used in column mode (`column_axis != None`) in the current closure execution context.
 column_range: cvars.ContextVar[range] = cvars.ContextVar("column_range")
 #: Offset provider dict in the current closure execution context.
-offset_provider: cvars.ContextVar[OffsetProvider] = cvars.ContextVar(
-    "offset_provider"
-)
+offset_provider: cvars.ContextVar[OffsetProvider] = cvars.ContextVar("offset_provider")
 
 
 class Column(np.lib.mixins.NDArrayOperatorsMixin):
--- a/src/gt4py/next/iterator/embedded.py
+++ b/src/gt4py/next/iterator/embedded.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import abc
+import contextvars as cvars
 import copy
 import dataclasses
 import itertools
@@ -182,9 +183,10 @@ class MutableLocatedField(LocatedField, Protocol):
         ...
 
 
-# TODO see https://github.com/GridTools/gt4py/pull/1120
-_column_range: Optional[range] = None
-_offset_provider: Optional[OffsetProvider] = None
+context_column_range: cvars.ContextVar[range] = cvars.ContextVar("context_column_range")
+context_offset_provider: cvars.ContextVar[OffsetProvider] = cvars.ContextVar(
+    "context_offset_provider"
+)
 
 
 class Column(np.lib.mixins.NDArrayOperatorsMixin):
@@ -197,7 +199,8 @@ class Column(np.lib.mixins.NDArrayOperatorsMixin):
     def __init__(self, kstart: int, data: np.ndarray | Scalar) -> None:
         self.kstart = kstart
         assert isinstance(data, (np.ndarray, Scalar))  # type: ignore # mypy bug
-        self.data = data if isinstance(data, np.ndarray) else np.full(len(_column_range), data)  # type: ignore[arg-type]
+        _column_range = context_column_range.get()
+        self.data = data if isinstance(data, np.ndarray) else np.full(len(_column_range), data)
 
     def __getitem__(self, i: int) -> Any:
         result = self.data[i - self.kstart]
@@ -720,6 +723,7 @@ def _make_tuple(
     *,
     column_axis: Optional[Tag] = None,
 ) -> Column | npt.DTypeLike | tuple[tuple | Column | npt.DTypeLike, ...]:
+    _column_range = context_column_range.get()
     if isinstance(field_or_tuple, tuple):
         if column_axis is not None:
             assert _column_range
@@ -768,6 +772,7 @@ class MDIterator:
 
     def shift(self, *offsets: OffsetPart) -> MDIterator:
         complete_offsets = group_offsets(*offsets)
+        _offset_provider = context_offset_provider.get()
         assert _offset_provider is not None
         return MDIterator(
             self.field,
@@ -792,6 +797,7 @@ class MDIterator:
             if not all(axis.value in shifted_pos.keys() for axis in axes if axis is not None):
                 raise IndexError("Iterator position doesn't point to valid location for its field.")
         slice_column = dict[Tag, range]()
+        _column_range = context_column_range.get()
         if self.column_axis is not None:
             assert _column_range is not None
             k_pos = shifted_pos.pop(self.column_axis)
@@ -834,6 +840,7 @@ def make_in_iterator(
         init = [None] * sparse_dimensions.count(sparse_dim)
         new_pos[sparse_dim] = init  # type: ignore[assignment] # looks like mypy is confused
     if column_axis is not None:
+        _column_range = context_column_range.get()
         # if we deal with column stencil the column position is just an offset by which the whole column needs to be shifted
         assert _column_range is not None
         new_pos[column_axis] = _column_range.start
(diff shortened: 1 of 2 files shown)


havogt:
Opening as a reminder, instead of #896

egparedes:
I can clean up the code and provide a final version of this prototype without too much effort. @tehrengruber  do you also agree with using this implementation strategy to fix the problem?

tehrengruber:
Yes, given the model itself this is much better than the global variable.

egparedes:
I've requested a formal review from @tehrengruber because apparently @havogt can't approve it since he's the creator of the PR, but I think both of you should take a look.

src/gt4py/next/iterator/embedded.py
```diff
@@ -182,9 +183,12 @@ def field_setitem(self, indices: FieldIndexOrIndices, value: Any) -> None:
         ...
 
 
-# TODO see https://github.com/GridTools/gt4py/pull/1120
-_column_range: Optional[range] = None
-_offset_provider: Optional[OffsetProvider] = None
+#: Column range used in column mode (`column_axis != None`) in the current closure execution context.
+context_column_range: cvars.ContextVar[range] = cvars.ContextVar("context_column_range")
```
tehrengruber, line 187: I‘m not convinced of the naming scheme, but your choice.

tehrengruber (approved):

src/gt4py/next/iterator/embedded.py
egparedes: Any suggestion?

egparedes: Changed

src/gt4py/next/iterator/embedded.py
```diff
@@ -197,7 +199,8 @@ class Column(np.lib.mixins.NDArrayOperatorsMixin):
     def __init__(self, kstart: int, data: np.ndarray | Scalar) -> None:
         self.kstart = kstart
         assert isinstance(data, (np.ndarray, Scalar))  # type: ignore # mypy bug
-        self.data = data if isinstance(data, np.ndarray) else np.full(len(_column_range), data)  # type: ignore[arg-type]
+        _column_range = column_range_cvar.get()
```
tehrengruber, line 202:
```suggestion
        column_range = column_range_cvar.get()
```
Since it is a local variable now the underscore is not needed anymore right?

src/gt4py/next/iterator/embedded.py
```diff
@@ -720,6 +723,7 @@ def _make_tuple(
     *,
     column_axis: Optional[Tag] = None,
 ) -> Column | npt.DTypeLike | tuple[tuple | Column | npt.DTypeLike, ...]:
+    _column_range = column_range_cvar.get()
```
tehrengruber, line 726:
```suggestion
    column_range = column_range_cvar.get()
```
Same as above.

src/gt4py/next/iterator/embedded.py
```diff
@@ -768,6 +772,7 @@ class MDIterator:
 
     def shift(self, *offsets: OffsetPart) -> MDIterator:
         complete_offsets = group_offsets(*offsets)
+        _offset_provider = offset_provider_cvar.get()
```
tehrengruber, line 775:
```suggestion
        offset_provider = offset_provider_cvar.get()
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -792,6 +797,7 @@ def deref(self) -> Any:
             if not all(axis.value in shifted_pos.keys() for axis in axes if axis is not None):
                 raise IndexError("Iterator position doesn't point to valid location for its field.")
         slice_column = dict[Tag, range]()
+        _column_range = column_range_cvar.get()
```
tehrengruber, line 800:
```suggestion
        column_range = column_range_cvar.get()
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -834,6 +840,7 @@ def make_in_iterator(
         init = [None] * sparse_dimensions.count(sparse_dim)
         new_pos[sparse_dim] = init  # type: ignore[assignment] # looks like mypy is confused
     if column_axis is not None:
+        _column_range = column_range_cvar.get()
```
tehrengruber, line 843:
```suggestion
        column_range = column_range_cvar.get()
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -1083,6 +1090,7 @@ def __getitem__(self, _):
 def neighbors(offset: runtime.Offset, it: ItIterator) -> _List:
     offset_str = offset.value if isinstance(offset, runtime.Offset) else offset
     assert isinstance(offset_str, str)
+    _offset_provider = offset_provider_cvar.get()
```
tehrengruber, line 1093:
```suggestion
    offset_provider = offset_provider_cvar.get()
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -1141,6 +1149,7 @@ class SparseListIterator:
     offsets: Sequence[OffsetPart] = dataclasses.field(default_factory=list, kw_only=True)
 
     def deref(self) -> Any:
+        _offset_provider = offset_provider_cvar.get()
```
tehrengruber, line 1152:
```suggestion
        offset_provider = offset_provider_cvar.get()
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -1290,13 +1299,14 @@ def _column_dtype(elem: Any) -> np.dtype:
 @builtins.scan.register(EMBEDDED)
 def scan(scan_pass, is_forward: bool, init):
     def impl(*iters: ItIterator):
+        _column_range = column_range_cvar.get()
```
tehrengruber, line 1302:
```suggestion
        column_range = column_range_cvar.get()
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -1336,7 +1345,7 @@ def closure(
         if not (is_located_field(out) or can_be_tuple_field(out)):
             raise TypeError("Out needs to be a located field.")
 
-        global _column_range
+        _column_range = None
```
tehrengruber, line 1348:
```suggestion
        column_range = None
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -1321,7 +1331,6 @@ def fendef_embedded(fun: Callable[..., None], *args: Any, **kwargs: Any):
     if "offset_provider" not in kwargs:
         raise RuntimeError("offset_provider not provided")
 
-    global _offset_provider
     _offset_provider = kwargs["offset_provider"]
```
tehrengruber, line 1334:
```suggestion
    offset_provider = kwargs["offset_provider"]
```

tests/next_tests/unit_tests/iterator_tests/test_embedded_internals.py
```diff
@@ -12,85 +12,165 @@
...
+        b = embedded.Column(1, np.asarray([2, 2]))
+
+        with pytest.raises(ValueError):
+            np.where(cond, wrong_shape, b)
+
+    _run_within_context(test_func)
+
+
+def test_column_multithread():
+    def test_func(i: int, output: list) -> None:
```
tehrengruber, line 153: Is this really a meaningful test? I have no clue how the thread scheduling works in Python, but this might very well never fail even if the column range is implemented incorrectly. If we want this to be meaningful how about a lock that ensure the column construction of the first thread happens after the context setup of a second thread? (If it's not worth the effort I would remove the test altogether to be honest.)

tests/next_tests/unit_tests/iterator_tests/test_embedded_internals.py
```diff
@@ -12,85 +12,165 @@
...
+    def test_func():
+        a = embedded.Column(1, np.asarray(range(0, 3)))
+        b = embedded.Column(1, np.asarray(range(3, 6)))
+        res = a + b
+
+        assert isinstance(res, embedded.Column)
+        assert np.array_equal(res.data, a.data + b.data)
+        assert res.kstart == 1
+
+    _run_within_context(test_func)
```
tehrengruber, line 49: This is fine, but is there a reason to not use a context handler for this? Sure the columns would leak scope, but doesn't really matter right?

tehrengruber (changes requested):

tests/next_tests/unit_tests/iterator_tests/test_embedded_internals.py
egparedes: Ok, I agree and removed the test.

tests/next_tests/unit_tests/iterator_tests/test_embedded_internals.py
egparedes: A context handler doesn't really work here because code running inside a context needs to be an actual function (line 36). Setting the values of the context vars before running some lines of code and then restoring their previous values does not test the context vars and it doesn't work in the general case (it will fail when running multiple threads, for example).

tehrengruber (approved):

PR #1121, opened 2023-01-11
Fix mypy decorator

nfarabullini:
Un-ignore `decorator.py` and fix all the reported mypy errors


src/functional/ffront/decorator.py
```diff
@@ -480,6 +482,7 @@ def from_function(
             **operator_attribute_nodes,
         )
         foast_node = FieldOperatorTypeDeduction.apply(untyped_foast_node)
+        assert isinstance(foast_node, (foast.ScanOperator, foast.FieldOperator))
```
havogt, line 485: related to the comment below this should be an instance of `OperatorNodeT`, not sure how mypy deals with that.

src/functional/ffront/decorator.py
```diff
@@ -504,7 +508,8 @@ def __gt_itir__(self) -> itir.FunctionDefinition:
         if hasattr(self, "__cached_itir"):
             return getattr(self, "__cached_itir")  # noqa: B009
 
-        itir_node: itir.FunctionDefinition = FieldOperatorLowering.apply(self.foast_node)
+        itir_node = FieldOperatorLowering.apply(self.foast_node)
+        assert isinstance(itir_node, itir.FunctionDefinition)
```
havogt, line 512: maybe we can fix the return type annotation of `FieldOperatorLowering.apply` to make this work without assert?

src/functional/ffront/decorator.py
```diff
@@ -488,6 +491,7 @@ def from_function(
         )
 
     def __gt_type__(self) -> ts.CallableType:
+        assert isinstance(self.foast_node, (foast.FieldOperator, past.Program, foast.ScanOperator))
```
havogt, line 494: see comment below

src/functional/ffront/decorator.py
```diff
@@ -154,8 +155,8 @@ def is_cartesian_offset(o: FieldOffset):
 
 
 def _field_constituents_shape_and_dims(
-    arg, arg_type: ts.FieldType | ts.ScalarType | ts.TupleType
-) -> Generator[tuple[tuple[int, ...], list[Dimension]]]:
+    arg, arg_type: ts.DataType
+) -> Iterator[tuple[Optional[tuple[Any]], list[Dimension]]]:
```
havogt, line 159: this seems to general

havogt (changes requested):

src/functional/ffront/decorator.py
```diff
@@ -542,11 +547,11 @@ def as_program(
...
         )
         out_ref = past.Name(id="out", location=loc)
-
+        assert isinstance(self.foast_node, (foast.FieldOperator, past.Program, foast.ScanOperator))
         if self.foast_node.id in self.closure_vars:
             raise RuntimeError("A closure variable has the same name as the field operator itself.")
-        closure_vars = {self.foast_node.id: self}
-        closure_symbols = [
+        closure_vars: dict[str, Any] = {self.foast_node.id: self}
+        closure_symbols: list = [
```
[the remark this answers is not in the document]
nfarabullini: if I remove it, I get: 
```
src/functional/ffront/decorator.py:554:27: error: Need type annotation for "closure_symbols"  [var-annotated]
```
I will make it a bit more specific, i.e.:
 ```closure_symbols: list[past.Symbol] = [....```

src/functional/ffront/foast_passes/type_deduction.py
```diff
@@ -162,7 +162,7 @@ class FieldOperatorTypeDeduction(traits.VisitorWithSymbolTableTrait, NodeTransla
     """
 
     @classmethod
-    def apply(cls, node: foast.FunctionDefinition) -> foast.FunctionDefinition:
+    def apply(cls, node: foast.FunctionDefinition | foast.LocatedNode) -> foast.FunctionDefinition:
```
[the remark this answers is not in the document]
nfarabullini: if you look at `test_copy.py` in test_execution, this `apply` gets accessed with both foast.FunctionDefinition and foast.FieldOperator

src/functional/ffront/decorator.py
nfarabullini: I am not sure that the return type annotation is wrong. `FieldOperatorLowering.apply` is also used in foast_to_itir.py/_visit_shift, where I believe that itir.Expr is correct

src/functional/ffront/decorator.py
DropD: also, `arg` is missing an annotation.

src/functional/ffront/decorator.py
```diff
@@ -275,7 +276,7 @@ def __call__(self, *args, offset_provider: dict[str, Dimension], **kwargs) -> No
             )
         backend = self.backend or DEFAULT_BACKEND
 
-        ppi.ensure_processor_kind(backend, ppi.ProgramExecutor)
+        ppi.ensure_processor_kind(backend, ppi.ProgramExecutor)  # type: ignore[arg-type, type-abstract] # backend type is inspected inside function
```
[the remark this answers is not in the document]
DropD: I think the problem is the out-of-date annotation of DEFAULT_BACKEND as `Callable`. It should be a `ProgramProcessor`.

src/functional/ffront/decorator.py
nfarabullini: I put ` Any`there, tbh the zip creates issues with types

src/functional/ffront/decorator.py
DropD: I believe it could be done.

By changing the type annotations of `FieldOperatorLowering.apply`  like so:

```python
    ...
    def apply(self, node: foast.FunctionDefinition) -> itir.FunctionDefinition:
        ...
```

and changing the occurrence in `InsideReductionLowering._visit_shift` to

```python
        self.lambda_params[uid] = FieldOperatorLowering().visit(node)
```

we should get the typing we want in both places.

src/functional/ffront/foast_passes/type_deduction.py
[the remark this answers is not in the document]
DropD: I would suggest to create a `FoastNodeT` type var bound on `foast.LocatedNode` and then typing this `... node: FoastNodeT) -> FoastNodeT:`. Because that is really what it does.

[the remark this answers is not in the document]
nfarabullini: Though as far as I understand, `apply `should return `foast.FunctionDefinition`

mypy also complains about it if changed to what you suggested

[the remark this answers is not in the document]
DropD: That's a misunderstanding. `apply` should return exactly the same type of FOAST node as you input. It just calls the visitor functions and all of them adhere to that, as far as I can see. I thought `FoastNodeT -> FoastNodeT` would be the way to tell mypy that.

For me locally it accepts it too.

src/functional/ffront/decorator.py
```diff
@@ -542,11 +547,11 @@ def as_program(
             location=loc,
         )
         out_ref = past.Name(id="out", location=loc)
-
+        assert isinstance(self.foast_node, (foast.FieldOperator, past.Program, foast.ScanOperator))
```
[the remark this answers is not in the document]
DropD: I think the correct thing to do is to *constrain* `OperatorNodeT` on `foast.FieldOperator` and `foast.ScanOperator` because not other type is meaningful there currently. This solves all the problems, the only downside is `FieldOperator.from_function` can not have a default argument for `operator_node_cls`, as that argument is required by mypy to match up with whatever type `OperatorNodeT` represents in the concrete (non-generic) `FieldOperator[OperatorNodeT]` in question.

In practice this just means passing this argument in one place.

PR #1126, opened 2023-01-16
Fix mypy transforms files

nfarabullini:
Un-ignore `popup_tmps`, `merge_let`, `inline_lambdas`, `inline_lifts` and fix all the reported mypy errors

### DIFF
--- a/src/functional/iterator/transforms/inline_lambdas.py
+++ b/src/functional/iterator/transforms/inline_lambdas.py
@@ -57,7 +57,7 @@ def inline_lambda(
         #  in `name_map` that don't collide with each other. E.g. this
         #  must still work:
         # (lambda arg, arg_: (lambda arg_: ...)(arg))(a, b)  # noqa: E800
-        name_map: dict = {}
+        name_map: dict[ir.SymRef, str] = {}
 
         def new_name(name):
             while name in refs or name in syms or name in name_map.values():
--- a/src/functional/iterator/transforms/inline_lambdas.py
+++ b/src/functional/iterator/transforms/inline_lambdas.py
@@ -57,7 +57,7 @@ def inline_lambda(
         #  in `name_map` that don't collide with each other. E.g. this
         #  must still work:
         # (lambda arg, arg_: (lambda arg_: ...)(arg))(a, b)  # noqa: E800
-        name_map = {}
+        name_map: dict = {}
 
         def new_name(name):
             while name in refs or name in syms or name in name_map.values():
--- a/src/functional/iterator/transforms/inline_lifts.py
+++ b/src/functional/iterator/transforms/inline_lifts.py
@@ -68,7 +68,7 @@ def _transform_and_extract_lift_args(
     being ``{sym1: sym1, sym2: expr1}``.
     """
     assert _is_lift(node)
-    assert isinstance(node.fun, ir.FunCall)  # just for mypy
+    assert isinstance(node.fun, ir.FunCall)
     inner_stencil = node.fun.args[0]
 
     new_args = []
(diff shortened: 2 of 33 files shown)


src/functional/iterator/transforms/popup_tmps.py
```diff
@@ -57,11 +57,14 @@ def _extract_lambda(
             is_scan = isinstance(fun, ir.FunCall) and fun.fun == ir.SymRef(id="scan")
             is_reduce = isinstance(fun, ir.FunCall) and fun.fun == ir.SymRef(id="reduce")
             if is_scan:
-                assert isinstance(fun, ir.FunCall)  # just for mypy
+                assert isinstance(fun, ir.FunCall)
```
tehrengruber, line 60: I would leave the comment here as-is, as this is indeed only for mypy, but otherwise an invariant already ensured above.

src/functional/iterator/transforms/popup_tmps.py
```diff
@@ -77,6 +80,9 @@ def wrap(fun: ir.Lambda, args: list[ir.Expr]) -> ir.FunCall:
                         fun=ir.SymRef(id="scan"), args=scan_args
                     )
                 elif is_reduce:
+                    assert isinstance(node.fun, ir.FunCall) and isinstance(
+                        node.fun.args[0], ir.FunCall
+                    )
```
tehrengruber, line 85: These checks should validate `node` has the proper structure, but this one appears to only include stuff that is needed to make mypy happy. For example the length of the arguments is not validated, but probably should be. I can not comment on the actual check because I am missing the knowledge of the actual pass here.

tehrengruber (commented):
I left some comments, but I am not convinced of the usefulness of this approach to fixing failing type checks. For example [here](https://github.com/GridTools/gt4py/blob/0703c5dad3504afdf70aa9ca9b18955dc3061a91/src/functional/iterator/transforms/popup_tmps.py#L86) a proper explanation of the format expected by the pass that goes beyond what type checking requires is missing. Just fixing the type errors in the end only hides this fact and I believe only with knowledge of the actual pass one can properly fix this. I would suggest to either put TODOs in such places and do a best-effort approach to ensuring the correct format or otherwise leave the passes as ignored for the time being.

src/functional/iterator/transforms/popup_tmps.py
nfarabullini: what do you mean with the last part of the sentence?

src/functional/iterator/transforms/inline_lambdas.py
```diff
@@ -57,7 +57,7 @@ def inline_lambda(
         #  in `name_map` that don't collide with each other. E.g. this
         #  must still work:
         # (lambda arg, arg_: (lambda arg_: ...)(arg))(a, b)  # noqa: E800
-        name_map = {}
+        name_map: dict = {}
```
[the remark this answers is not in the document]
nfarabullini: makes sense

src/functional/iterator/transforms/popup_tmps.py
```diff
@@ -140,7 +146,7 @@ def visit_FunCall(
             new_params = [ir.Sym(id=p.id) for p in nested_lifts.values()]
             fun = ir.Lambda(params=fun.params + new_params, expr=fun.expr)
             # for the arguments, we have to resolve possible cross-references of lifts
-            symbol_map = {v.id: k for k, v in nested_lifts.items()}
+            symbol_map = {v.id: k for k, v in nested_lifts.items()}  # type: ignore[misc] # key expression automatically filled
```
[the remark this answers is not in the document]
nfarabullini: it's for nested_lifts.keys() aka k

src/functional/iterator/transforms/popup_tmps.py
nfarabullini: it can be treated as other asserts, i.e. I can add the comment "# just for mypy" or analogous

src/functional/iterator/transforms/popup_tmps.py
nfarabullini: spoke to Hannes about this, I am ok leaving the comment

src/functional/iterator/transforms/popup_tmps.py
```diff
@@ -57,11 +57,14 @@ def _extract_lambda(
             is_scan = isinstance(fun, ir.FunCall) and fun.fun == ir.SymRef(id="scan")
             is_reduce = isinstance(fun, ir.FunCall) and fun.fun == ir.SymRef(id="reduce")
             if is_scan:
-                assert isinstance(fun, ir.FunCall)  # just for mypy
+                assert isinstance(fun, ir.FunCall)
                 fun = fun.args[0]
+                assert isinstance(fun, ir.Lambda)
```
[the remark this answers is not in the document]
nfarabullini: I believe that in this case, the assert in indeed needed outside of mypy. However for cases where assertions are needed only for mypy to pass, I will put a type: ignore

src/functional/iterator/transforms/inline_lifts.py
```diff
@@ -182,11 +185,14 @@ def visit_FunCall(
                 # TODO(tehrengruber): we currently only inlining opcount preserving, but what we
                 #  actually want is to inline whenever the argument is not shifted. This is
                 #  currently beyond the capabilities of the inliner and the shift tracer.
-                new_args = {}
+                new_args_dict: dict = {}
```
tehrengruber, line 188:
```suggestion
                new_arg_exprs: dict[<todo>] = {}
```

tehrengruber (approved):

PR #1128, opened 2023-01-17
Fix mypy global_tmps

nfarabullini:
Un-ignore `global_tmps` and fix all the reported mypy errors


src/functional/iterator/transforms/global_tmps.py
```diff
@@ -250,7 +252,7 @@ def _named_range_with_offsets(
 
 
 def _extend_cartesian_domain(
-    domain: ir.FunCall, offsets: Sequence[tuple], offset_provider: Mapping[str, CartesianAxis]
+    domain: ir.Expr, offsets: Sequence[tuple], offset_provider: Mapping[str, CartesianAxis]
```
havogt, line 255: Would it make sense to pass the proper thing on the call-side? Because on line 259, we assert that it's a FunCall.

src/functional/iterator/transforms/global_tmps.py
```diff
@@ -290,6 +292,7 @@ def update_cartesian_domains(
     domains = dict[str, ir.Expr]()
     for closure in reversed(node.fencil.closures):
         if closure.domain == AUTO_DOMAIN:
+            assert isinstance(closure.output, ir.SymRef)
```
havogt, line 295: We need to check how to fix that, because closure.output can be a FunCall. That looks like a bug in this pass.

src/functional/iterator/transforms/global_tmps.py
```diff
@@ -401,6 +407,7 @@ def update_unstructured_domains(node: FencilWithTemporaries, offset_provider: Ma
...
```
havogt, line 410: see other comment

src/functional/iterator/transforms/global_tmps.py
```diff
@@ -442,7 +450,7 @@ def collect_tmps_info(node: FencilWithTemporaries) -> FencilWithTemporaries:
     domains: dict[str, ir.Expr] = {
         closure.output.id: closure.domain
         for closure in node.fencil.closures
-        if closure.output.id in tmps
+        if isinstance(closure.output, ir.SymRef) and closure.output.id in tmps
```
havogt, line 453: see other comment

havogt (changes requested):
I think we need to modify the pass to work for cases where `closure.output` is not a SymRef. A possible case could be `make_tuple(out0,out1)`, but we have to double check if that should actually be legal.

PR #1129, opened 2023-01-18
ADR: Generic Tuples in Iterator IR

havogt:
Incomplete summary of our discussion on tuples in itir


docs/functional/architecture/0014-No_Generic_Tuples_in_Iterator_IR.md
```diff
@@ -0,0 +1,64 @@
+---
+tags: [iterator]
+---
+
+# [No Generic Tuples in Iterator IR]
+
+- **Status**: valid
+- **Authors**: Enrique G. Pareded (@egparedes), Felix Thaler (@fthaler), Hannes Vogt (@havogt), Till Ehrengruber (@tehrengruber)
```
havogt, line 8:
```suggestion
- **Authors**: Enrique G. Paredes (@egparedes), Felix Thaler (@fthaler), Hannes Vogt (@havogt), Till Ehrengruber (@tehrengruber)
```

docs/functional/architecture/0014-No_Generic_Tuples_in_Iterator_IR.md
```diff
@@ -0,0 +1,64 @@
...
+
+Tuples on same location can be expressed as Iterator of tuples
+
+```python
+lift(lambda a,b: return make_tuple(deref(a), deref(b)))
+```
+
+## Conclusion
+
+Iterator IR is focused around liftable stencils. Passing tuples of iterators pointing to different locations is not a useful feature in Iterator IR as in the stencil body you need to work with the decomposed tuple anyway to move it to the right locations, to be able to deref.
```
fthaler, line 57:
```suggestion
The composability of Iterator IR is focused around liftable stencils. Lifting requires that all input iterators current location types match. Passing tuples of iterators pointing to different locations is not a useful feature in Iterator IR, as this makes the resulting function non-liftable and thus non-composable.
```

havogt:
Just dropping 2 notes:
1. Till found a reasonable example where tuple of iterator could make sense. He also found a way to implement it in the type_inference: if the tuple of iterator is transposable than it **is** also a iterator of tuple. Otherwise it's just a tuple of iterator.
2. Unfortunately conditionals on iterators (or tuple of iterators) can be used in C++ only in restricted cases, e.g. the following breaks:
```c++
cond ? it_a : it_b
```
if `it_a` and `it_b` have different type. (Note if the dtype of the iterator is the same, the derefed lifted version of that code is fine.)

PR #1130, opened 2023-01-18
Functional: Index fields C++ backend support

petiaccja:
Adds index fields as a distinct argument type for the C++ binding generator. The binding generator generates the correct C++ code that passes `gridtools::stencil::positional` iterators to the GTFN C++ program code.

### DIFF
--- a/src/functional/otf/binding/cpp_interface.py
+++ b/src/functional/otf/binding/cpp_interface.py
@@ -19,7 +19,7 @@ from typing import Final, Sequence, Type
 
 import numpy as np
 
-import functional.type_system.type_specifications as ts
+import functional.otf.binding.type_specifications as ts
 from functional.otf import languages
 from functional.otf.binding import interface
 
@@ -98,7 +98,7 @@ def render_scalar_type(scalar_type: ts.ScalarType) -> str:
 def _render_function_param(param: interface.Parameter, index: int) -> str:
     if isinstance(param.type_, ts.ScalarType):
         return f"{render_scalar_type(param.type_)} {param.name}"
-    elif isinstance(param.type_, ts.FieldType):
+    elif isinstance(param.type_, (ts.FieldType, ts.IndexFieldType)):
         return f"BufferT{index}&& {param.name}"
     else:
         raise ValueError("unsupported type for parameters")
@@ -114,7 +114,7 @@ def render_function_declaration(function: interface.Function, body: str) -> str:
     template_params = [
         f"class BufferT{index}"
         for index, param in enumerate(function.parameters)
-        if isinstance(param.type_, ts.FieldType)
+        if isinstance(param.type_, (ts.FieldType, ts.IndexFieldType))
     ]
     if template_params:
         return f"""
--- a/src/functional/otf/binding/cpp_interface.py
+++ b/src/functional/otf/binding/cpp_interface.py
@@ -19,6 +19,7 @@ from typing import Final, Sequence, Type
 
 import numpy as np
 
+import functional.type_system.type_specifications as ts
 from functional.otf import languages
 from functional.otf.binding import interface
 
@@ -79,13 +80,28 @@ def render_python_type(python_type: Type) -> str:
     return _TYPE_MAPPING[python_type]
 
 
-def _render_function_param(
-    param: interface.ScalarParameter | interface.BufferParameter, index: int
-) -> str:
-    if isinstance(param, interface.ScalarParameter):
-        return f"{render_python_type(param.scalar_type.type)} {param.name}"
+def render_scalar_type(scalar_type: ts.ScalarType) -> str:
+    if scalar_type.kind == ts.ScalarKind.BOOL:
+        return "bool"
+    elif scalar_type.kind == ts.ScalarKind.INT32:
+        return "int32_t"
+    elif scalar_type.kind == ts.ScalarKind.INT64:
+        return "int64_t"
+    elif scalar_type.kind == ts.ScalarKind.FLOAT32:
+        return "float"
+    elif scalar_type.kind == ts.ScalarKind.FLOAT64:
+        return "double"
     else:
+        raise ValueError("unsupported scalar type")
+
+
+def _render_function_param(param: interface.Parameter, index: int) -> str:
+    if isinstance(param.type_, ts.ScalarType):
+        return f"{render_scalar_type(param.type_)} {param.name}"
+    elif isinstance(param.type_, ts.FieldType):
         return f"BufferT{index}&& {param.name}"
+    else:
+        raise ValueError("unsupported type for parameters")
 
 
 def render_function_declaration(function: interface.Function, body: str) -> str:
@@ -98,7 +114,7 @@ def render_function_declaration(function: interface.Function, body: str) -> str:
     template_params = [
         f"class BufferT{index}"
         for index, param in enumerate(function.parameters)
-        if isinstance(param, interface.BufferParameter)
+        if isinstance(param.type_, ts.FieldType)
     ]
     if template_params:
         return f"""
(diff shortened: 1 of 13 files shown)


src/functional/otf/binding/pybind.py
```diff
@@ -108,11 +110,15 @@ class BindingCodeGenerator(TemplatedGenerator):
     )
 
     def visit_FunctionParameter(self, param: FunctionParameter):
-        if param.ndim > 0:
+        if isinstance(param.type_, ts.FieldType):
             type_str = "pybind11::buffer"
+        elif isinstance(param.type_, ts.IndexFieldType):
+            type_str = "pybind11::none"
```
petiaccja, line 116: There is no object shared across C++ and Python that could represents index fields. Furthermore, index fields are currently fully described by their type, they contain have no data members.

src/functional/otf/binding/pybind.py
```diff
@@ -34,13 +33,17 @@ class DimensionType(Expr):
     name: str
 
 
-class SidConversion(Expr):
-    buffer_name: str
+class BufferIterator(Expr):
```
havogt, line 36: I'd remove `Iterator` here, because they are not iterators (in the iterator ir sense). The iterators used in the backend are wrapped around these objects.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -18,22 +18,37 @@
 
 import numpy as np
 
+import functional.otf.binding.type_specifications as ts
```
DropD, line 21: `ts` name should be reserved for `functional.type_system.type_specifications`

DropD (changes requested):
only minor problems with applying the new import style.

src/functional/otf/binding/cpp_interface.py
```diff
@@ -19,6 +19,7 @@
 
 import numpy as np
 
+import functional.otf.binding.type_specifications as ts
```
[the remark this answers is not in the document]
petiaccja: Well, the idea was that the derived modules are a strict superset of the original `type_system.*` modules, and in that case you can just simply import the extended module.
Sounds like it would be kinda nice to use, but to be honest it might be better as you suggest. Re-importing all the stuff is ugly and error-prone, and then not knowing which flavour of the module you are using at the moment is also a problem. Plus, it's really nasty if you want to use two extended modules at the same time.

Anyways, I think this is out of the scope of this PR, but we can change it in a follow-up PR.

src/functional/otf/binding/pybind.py
petiaccja: Renamed it to `BufferSID` and `PositionalSID`.

src/functional/otf/binding/cpp_interface.py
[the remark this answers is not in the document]
egparedes: I agree we should not use the same alias for different modules. I think it would be good enough for this PR to just change the alias to a similar but different one. For example:
```
import functional.otf.binding.type_specifications as bts
```

src/functional/otf/binding/type_specifications.py
```diff
@@ -0,0 +1,27 @@
...
+    DimensionType as DimensionType,
+    FieldType as FieldType,
+    FunctionType as FunctionType,
+    OffsetType as OffsetType,
+    ScalarKind as ScalarKind,
+    ScalarType as ScalarType,
+    TupleType as TupleType,
+    TypeSpec as TypeSpec,
+    VoidType as VoidType,
+)
```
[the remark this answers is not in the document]
egparedes: I also agree with @DropD here. I think it would be better to have multiple imports to bring new types in scope instead of trying to expand the original `type_specifications` module (which makes basically impossible to compose type extensions from different modules):
```
# This allows adding extensions from different modules very easily
import functional.type_system.type_specifications as ts
import functional.otf.binding.type_specifications as bts
```

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -16,24 +16,28 @@
...
-def get_param_description(
-    name: str, obj: Any
-) -> interface.ScalarParameter | interface.BufferParameter:
-    view: np.ndarray = np.asarray(obj)
-    if view.ndim > 0:
-        return interface.BufferParameter(name, tuple(dim.value for dim in obj.axes), view.dtype)
+def get_param_description(name: str, obj: Any) -> interface.Parameter:
+    # TODO: make type_translation extendable and specialize it for the
+    #   bindings to distinguish between buffer-backed fields and index
+    #   fields.
```
egparedes, line 33: I now understand that the comment about _making `type_translation` extendable_ is related to use a different mechanism for the pattern matching part with the actual type, and not with the preliminary handling of the messy implementation details of python typing annotations. It makes sense, but still don't see it as a high priority for now.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -16,24 +16,28 @@
...
+    if isinstance(obj, IndexField):
+        dtype = type_translation.from_type_hint(obj.dtype.type)
+        assert isinstance(dtype, ts.ScalarType)
+        return interface.Parameter(name, ts.IndexFieldType(axis=obj.axis, dtype=dtype))
```
egparedes, line 37: I think, for coherency, this piece of code should be moved into `type_translation.from_value()`, since the similar handling of `embedded.LocatedField` is also implemented there. Additionally, it's easier to refactor the type translation code in the future if it is not spread across different functions.

egparedes (changes requested):
Overall, it looks good to me. I left a few comments and suggestions to improve the PR.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
petiaccja: I moved it to a new function `from_value` under `functional.otf.binding.type_translation`. It shouldn't be moved into the main `type_translation` module because `IndexFieldType` is not present there.

src/functional/program_processors/codegens/gtfn/gtfn_module.py
petiaccja: I wrote a new `from_value` function that tries to match for `IndexFieldType` and if it fails it delegates it to the original `functional.type_system.from_value` function.

petiaccja: I'd like to use a registry that allows you to register a match functor for every type of the `type_system`, then you can extend without raising the cyclomatic complexity through all these if statements and you can also simply chain the list of matchers to combine modules.

petiaccja: That would also replace the `__gt_type__` solution which is an intrusive implementation because it forces objects to be aware of the type system.

src/functional/otf/binding/cpp_interface.py
```diff
@@ -79,13 +80,28 @@ def render_python_type(python_type: Type) -> str:
...
+    elif scalar_type.kind == ts.ScalarKind.INT32:
+        return "int32_t"
+    elif scalar_type.kind == ts.ScalarKind.INT64:
+        return "int64_t"
+    elif scalar_type.kind == ts.ScalarKind.FLOAT32:
+        return "float"
+    elif scalar_type.kind == ts.ScalarKind.FLOAT64:
+        return "double"
     else:
+        raise ValueError("unsupported scalar type")
```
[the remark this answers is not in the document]
petiaccja: I didn't invest too much into exceptions because I want to clean the whole thing up next cycle. Anyways, you have a point that even clean-up is easier if the messages are better to begin with, so I added some better messages.

tehrengruber (commented):
Just adding the partial review I already started.

PR #1137, opened 2023-01-23
Eve: document behavior of annex

havogt:


src/eve/visitors.py
```diff
@@ -165,6 +165,7 @@ def generic_visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
                     if (new_child := self.visit(child, **kwargs)) is not NOTHING
                 },
             )
+            object.__setattr__(new_node, "__node_annex__", copy.deepcopy(node.annex))
```
tehrengruber, line 168: The annex also includes the symbol table (and potentially other stuff) which might have changed in `new_node`. Reusing the old annex has lead to very hard to track bugs and was removed https://github.com/GridTools/gt4py/pull/1055 and also in master branch (i.e. cartesian) independently.

havogt: We were actually looking for this change. Not sure why we couldn’t find it. We should document somewhere what annex can do and what it should do. The current implementation (the one in main and functional) means annex can only be used in the same visitor, which probably limits it to traits like the one for symtables.

egparedes: Yes, I knew there was a reason why the annex was not copied but we couldn't find Till's PR (which is strange). For the time being, I agree we should document this behavior. Another option could be to keep the current behavior for the annex and add another annex-like member or a special key in the annex  (e.g. `stored`, `clonable`, `static`, ...) which is preserved across passes. In the long term, I think the annex was a bad idea and we should get rid of it.

egparedes (approved):
Good to go.

PR #1141, opened 2023-01-26
Embedded: Tuple column fixes

havogt:
incomplete, need to handle tuple of column and column of tuple consistently everywhere


tehrengruber:
For the record: As discussed in #1280 this also fixes the tests disabled here: https://github.com/GridTools/gt4py/pull/1280/commits/e91b4be3cefe62c3513e8eca85c3cb8cef4987b0.

PR #1144, opened 2023-01-27
Add scan projector

havogt:


havogt:
closing as today the solution would probably look different

PR #1142, opened 2023-01-26
Adapt CI to ci-ext

finkandreas:
Use ci-ext service and simplified .gitlab-ci.yml for it.

PR #1153, opened 2023-02-03
Adapt CI to ci-ext

DropD:
Superseeds #1142 (purely for convenience).

Use ci-ext service and simplified ci config for it.


ci/base.Dockerfile
```diff
@@ -24,6 +24,7 @@ RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
     python-openssl \
     libreadline-dev \
     git \
+    rustc \
```
egparedes, line 27: Out of curiosity: why is `rust` needed?

ci/cscs-ci.yml
```diff
@@ -0,0 +1,91 @@
...
+    <<: *py310
+
+test py38:
+  extends: .container-runner-daint-gpu
+  needs: ["build py38 image"]
+  stage: test
+  image: $CSCS_REGISTRY_PATH/gt4py/gt4py-ci:$PYVERSION
+  script:
+  - cd /gt4py.src
+  - python -c "import cupy"
```
egparedes, line 68: Another question: is there any value in testing the `cupy` import before running `tox`?

egparedes (approved):
I have a couple of questions but they are mostly for my own understanding, so I'll approve it.

ci/base.Dockerfile
DropD: Previously, the CI failed at https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4455690602105886/4525297225819146/-/jobs/3744176151

This is an excerpt of the output:

> error: subprocess-exited-with-error
>  
>  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [6 lines of output]
>      
>      Cargo, the Rust package manager, is not installed or is not on PATH.
>      This package requires Rust and Cargo to compile extensions. Install it through
>      the system's package manager or via https://rustup.rs/
>      
>      Checking for Rust toolchain....
>      [end of output]

As much as I could find out there is a dependency of a dependency that causes this problem whenever a version is requested for which there is no wheel available (yet). It might have something to do with the python package `orjson` which one or multiple of our dependencies depend on in turn.

ci/cscs-ci.yml
DropD: If memory serves well, then this is to detect the case where the version of cuda/cupy in the docker image does not match the version of cuda available on daint. I will add a comment to that effect.

DropD: On second thought I will put that in a separate PR, updating CI docs as well.

PR #1155, opened 2023-02-07
Extend ITIR type inference to collect the types of all subexpressions

tehrengruber:
Type inferrence on ITIR was previously only returning the type of a single expression. This PR introduces extends the inference to collect the type of all subexpressions and introduces a new function `_infer_all` returning a mapping from the (Python) id of a node to its type. As we start using this function we will make it public.


src/gt4py/next/iterator/type_inference.py
```diff
@@ -27,6 +29,9 @@
 
 T = typing.TypeVar("T", bound="Type")
 
+# list of nodes that have a type
+TYPED_IR_NODES = (ir.Expr, ir.FunctionDefinition, ir.StencilClosure, ir.FencilDefinition, ir.Sym)
```
egparedes, line 33:
```suggestion
TYPED_IR_NODES: Final = (ir.Expr, ir.FunctionDefinition, ir.StencilClosure, ir.FencilDefinition, ir.Sym)
```

src/gt4py/next/iterator/type_inference.py
```diff
@@ -380,20 +385,54 @@ def _infer_shift_location_types(shift_args, offset_provider, constraints):
...
+    def apply(
+        cls,
+        node: ir.Node,
+        offset_provider: Optional[dict[str, Connectivity | Dimension]] = None,
+        reindex: bool = True,
+    ) -> dict[int, Type]:
+        """
+        Infer the types of the child expressions of a given iterator IR expression.
+
+        The result is a dictionary mapping the id of child nodes to their type.
```
egparedes, line 406: Just a suggestion, it could be worth to make clear here that by `id` is meant the Python `id()` 
```suggestion
        The result is a dictionary mapping the Python id of child nodes to their type.
```

src/gt4py/next/iterator/type_inference.py
```diff
@@ -380,20 +385,54 @@ def _infer_shift_location_types(shift_args, offset_provider, constraints):
...
+
+        # Compute the most general type that satisfies all constraints
+        unified_types = unify(list(collected_types.values()), inferrer.constraints)
+
+        if reindex:
+            unified_types = reindex_vars(list(unified_types))
+
+        return {
+            id_: unified_type for id_, unified_type in zip(collected_types.keys(), unified_types)
+        }
```
egparedes, line 423: Just a question: it looks to me like this part doesn't really belong to the Visitor since it's just postprocessing the result of the visit using the type unifier utilities from a different module. Wouldn't it be better to move this postprocessing part to the public entry point, the `infer()` function (and maybe also rename the pass)?

src/gt4py/next/iterator/type_inference.py
```diff
@@ -380,20 +385,54 @@ def _infer_shift_location_types(shift_args, offset_provider, constraints):
...
+
+    def visit(self, node, **kwargs):
+        if isinstance(node, TYPED_IR_NODES):
+            self.collected_types[id(node)] = super().visit(node, **kwargs)
+            return self.collected_types[id(node)]
+
+        return super().visit(node, **kwargs)
```
egparedes, line 430: Am I missing something or this could be simplified to:
```suggestion
        result = super().visit(node, **kwargs)
        if isinstance(node, TYPED_IR_NODES):
            self.collected_types[id(node)] = result

        return result
```
?

src/gt4py/next/iterator/type_inference.py
```diff
@@ -380,20 +385,54 @@ def _infer_shift_location_types(shift_args, offset_provider, constraints):
...
+
+    def visit_Sym(self, node: ir.Sym, **kwargs):
```
egparedes, line 432: Missing return type hint

src/gt4py/next/iterator/type_inference.py
```diff
@@ -380,20 +385,54 @@ def _infer_shift_location_types(shift_args, offset_provider, constraints):
...
+        unified_types = unify(list(collected_types.values()), inferrer.constraints)
+
+        if reindex:
+            unified_types = reindex_vars(list(unified_types))
+
+        return {
+            id_: unified_type for id_, unified_type in zip(collected_types.keys(), unified_types)
+        }
+
+    def visit(self, node, **kwargs):
```
egparedes, line 425: Missing return type hint

src/gt4py/next/iterator/type_inference.py
```diff
@@ -506,53 +542,55 @@ def visit_FunCall(
...
-            constraints=constraints,
-            symtypes=symtypes,
+        **kwargs,
+    ) -> LetPolymorphic:
+        fun = ir.Lambda(params=node.params, expr=node.expr)
+
+        # Since functions defined in a function definition are let-polymorphic we don't want
+        # their parameters to inherit the constraints of the arguments in a call to them. A simple
+        # way to do this is to run the type inference on the function itself and reindex its type
+        # vars when referencing the function, i.e. in a `SymRef`.
```
egparedes, line 569: Thanks for adding this explanation here. I think it would help eve more to expand a bit the docstring of the `LetPolmorphic`...

src/gt4py/next/iterator/type_inference.py
```diff
@@ -380,20 +385,54 @@ def _infer_shift_location_types(shift_args, offset_provider, constraints):
...
-class _TypeInferrer(eve.NodeTranslator):
+@dataclasses.dataclass
+class _TypeInferrer(eve.traits.VisitorWithSymbolTableTrait, eve.NodeTranslator):
     """Visit the full iterator IR tree, convert nodes to respective types and generate constraints."""
 
-    def __init__(self, offset_provider):
-        self.offset_provider = offset_provider
+    offset_provider: Optional[dict[str, Connectivity | Dimension]]
+    collected_types: dict[int, Type] = dataclasses.field(default_factory=dict)
+    constraints: set[tuple[Type, Type]] = dataclasses.field(default_factory=set)
```
egparedes, line 394: For future reference, could you add a short description of how the information is encoded in these types?

egparedes (commented):
It looks good. Besides a couple of missing type hints, the rest of my comments are just questions and optional suggestions.

src/gt4py/next/iterator/type_inference.py
tehrengruber: I thought about this while adding the apply here, but decided against it as the visitor itself uses the apply method (for let polymorphic functions in `visit_FunctionDefinition`). My perspective on the `TypeInferrer` is that its task includes unification of constraints. Does that make sense?

src/gt4py/next/iterator/type_inference.py
tehrengruber: I did. Not super happy with the clarity of them though.

src/gt4py/next/iterator/type_inference.py
```diff
@@ -387,7 +394,14 @@ def _infer_shift_location_types(shift_args, offset_provider, constraints):
 
 @dataclasses.dataclass
 class _TypeInferrer(eve.traits.VisitorWithSymbolTableTrait, eve.NodeTranslator):
-    """Visit the full iterator IR tree, convert nodes to respective types and generate constraints."""
+    """
+    Visit the full iterator IR tree, convert nodes to respective types and generate constraints.
+
+    Attributes:
+        collected_types Mapping from the (Python) id of a node to its type.
+        constraints     Set of constraints, where a constraint is a pair of types that need to
```
egparedes, line 402: Please use the correct Google style format for the docstrings (as required by sphix-napoleon: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google)
```suggestion
        collected_types: Mapping from the (Python) id of a node to its type.
        constraints: Set of constraints, where a constraint is a pair of types that need to
```

src/gt4py/next/iterator/type_inference.py
egparedes: I also saw that the visitor calls its own `apply()` but still think it makes more sense to strictly separate the visitor logic from the postprocessing, and call `infer()` when needed from the visitor. I agree that the type inference process includes the unification of constraints, but the `_TypeInferrer` visitor (which is not even public interface of the module) is just the implementation of one of the required steps, so implementation-wise makes sense to separate cleanly the different steps.

src/gt4py/next/iterator/type_inference.py
egparedes: Ok, I understand. We can keep as it is in the current PR and think about improvements in following PRs.

egparedes (commented):
Just two comments from my side (and the failing tests) and it will be ready to go.

tehrengruber:
Blocked by #1182

src/gt4py/next/iterator/type_inference.py
tehrengruber: Just calling `infer` from the visitor is not enough as we need not only the type of the node, but also the mapping id => node type for all child nodes. I put it into  `_infer_all` (which is what I previously had). One thing I noticed is that now the postprocessing can not be customized in subclasses of `_TypeInferrer` anymore. In any case this will likely change soon anyway when we make this public interface.

src/gt4py/next/iterator/type_inference.py
tehrengruber: Fixed. This should be adopted in the https://github.com/GridTools/gt4py/blob/main/CODING_GUIDELINES.rst at some point.

egparedes: It was already there in functional (https://github.com/GridTools/gt4py/blob/functional/CODING_GUIDELINES.md#docstrings), but it got lost during the merge. I will update soon the documents in master with the versions from the other branch.

PR #1156, opened 2023-02-08
Imperative Backend Bugfix (Init Conditional Problem)

mroethlin:
## Description

There was an issue with initializing the temporary variable holding the result of a conditional: 

```
      using cond_8_t =
          typename std::common_type<decltype(tupl_7), decltypetupl_7)>::type;
      cond_8_t cond_8 = 0.;
      ^^^^^^^^^^^^^^^^^^^^^
      can not initialize tuple with double
```
 
Instead of initializing with `0.`, the behavior was changed to use uniformed initialization with {}. In the above example this would read

```
cond_2_t cond_2 = {};
```

### DIFF
--- a/src/gt4py/next/program_processors/codegens/gtfn/codegen.py
+++ b/src/gt4py/next/program_processors/codegens/gtfn/codegen.py
@@ -269,10 +269,12 @@ class GTFNIMCodegen(GTFNCodegen):
 
     InitStmt = as_fmt("{init_type} {lhs} {op} {rhs};")
 
+    DefaultInitialized = as_fmt("{init_type} {name} = {{}};")
+
     Conditional = as_mako(
         """
           using ${cond_type} = typename std::common_type<decltype(${if_rhs_}), decltype(${else_rhs_})>::type;
-          ${init_stmt}
+          ${initializer}
           if (${cond}) {
             ${if_stmt}
           } else {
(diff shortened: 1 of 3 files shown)


mroethlin:
I decided I like your version better

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -269,6 +269,8 @@ class GTFNIMCodegen(GTFNCodegen):
 
     InitStmt = as_fmt("{init_type} {lhs} {op} {rhs};")
 
+    EmptyListInitializer = as_fmt("{{}}")
```
havogt, line 272:
```suggestion
    EmptyListInitializer = as_mako("{}")
```
nit, just saving the escape characters

havogt (approved):
I think it would be good to add a testcase that triggers this codepath.

tests/next_tests/iterator_tests/test_conditional.py
```diff
@@ -0,0 +1,62 @@
...
+
+    shape = [5, 7, 9]
+
+    inp = np.random.randint(0, 2, shape)
+    inp = np_as_located_field(IDim, JDim, KDim)(inp)
+
+    out = (
+        np_as_located_field(IDim, JDim, KDim)(np.zeros(shape)),
+        np_as_located_field(IDim, JDim, KDim)(np.zeros(shape)),
+    )
```
havogt, line 46: we cannot write tuple outputs, therefore probably you excluded gtfn. I'll try to come up with a testcase for that.

havogt (changes requested):
I'll see if I can improve the test

havogt (approved):

PR #1160, opened 2023-02-09
CI: inverse workflows for branch protection

havogt:
This is the official technique, see https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks


havogt:
(Unrelated: probably all should run if workflows are touched, currently eve and storage don't run because of positive filters)

PR #1164, opened 2023-02-13
[experiment] CSCS-CI: experimental matrix trigger setup

DropD:
## Description

Trying out a potentially cleaner way to describe the matrix of python versions on which to run the various tests, which may make it easier to simply drop in the GPU tests for `next` once there are some.

## Requirements

- [X] You looked at the [review checklist](https://github.com/GridTools/gt4py/blob/main/REVIEW_CHECKLIST.txt)


DropD:
cscs-ci run default

ci/tests.yml
```diff
@@ -0,0 +1,22 @@
...
+run tests:
+  extends: .container-runner-daint-gpu
+  stage: test
+  image: $CSCS_REGISTRY_PATH/gt4py/gt4py-ci:$PYVERSION
+  rules:
+  - if: $CI_PIPELINE_SOURCE == "parent_pipeline"
+  script:
+  - cd /gt4py.src
+  - python -c "import cupy"
+  - tox run -e $SUBPACKAGE-$PYVERSION_PREFIX-$BACKEND-cuda112
```
havogt, line 16: Just discovered we are not running cpu at all. We should discuss if that is what we want.

DropD: Yes, we can reconsider now that the queue times etc are not as problematic.

DropD: I will create another experimental branch running all the tests to collect data on how long it would take with this new system.

DropD:
This was an interesting experiment but the only advantage it offers is that python versions can be described in the matrix while also allowing the pipelines for the different python versions to move ahead without waiting for each other. GtLab is working on a solution for that though: https://gitlab.com/gitlab-org/gitlab/-/issues/254821.

The downside of this current experiment is reduced readability of the pipeline view on GitLab.

PR #1168, opened 2023-02-15
rebuild base image when necessary

finkandreas:
This PR adds a base image rebuild policy.
- Whenever ci/base.Dockerfile changes in its contents, we would rebuild the baseimage.
- No other dependency has been defined for the base image, but can be easily added if necessary
- Also the containers with the source code should have unique names, otherwise we can have race conditions, i.e. you could get build failures / build success only because the container in the test stage is not really the container that was created in the build stage (especially when a test job is restarted this would happen easily)

Untested, but if I see build failures after `cscs-ci run`, I will adapt.


DropD:
cscs-ci run

DropD:
cscs-ci run

DropD:
cscs-ci run

DropD (approved):

PR #1171, opened 2023-02-17
Test Hohgant

havogt:


finkandreas:
This is an unsorted and not yet well thought list of things that I noticed so far:
- `CRAY_CUDA_MPS` does not work anymore on hohgant
- CUDA_VISIBLE_DEVICES should be exposed, otherwise no GPUs are forwared by the container runtime
-  it could be that we need to set `NUM_PROCESSES` to something different than `auto`, related to `CRAY_CUDA_MPS`

In general this is a good usecase to get the container runner ready on hohgant, slowly adding the features needed such that it can really serve as the drop-in replacement for  the container runner on daint.

finkandreas:
The `CRAY_CUDA_MPS` variable is now being used to transparently correctly use MPS on Hohgant (the runner is doing all the boilerplate work)

`CUDA_VISIBLE_DEVICES` is now exporting by default all 4 GPUS, but this can be overridden by defining this variable in the job

`NUM_PROCESSES` can be kept as is.

All in all it should now "just work" on hohgant, you can try by writing the comment `cscs-ci run` (or push another change).

At the moment only one GPU is getting work with your framework, maybe that would be another interesting work, to dispatch test jobs to different GPUs.

havogt:
Cool thanks!

DropD:
Closing for now, because daint is currently more stable.

PR #1176, opened 2023-02-21
Improve Gitpod pre-commit and remove obsolete files

havogt:
- pull request template contains a link to a file that doesn't exist, regular contributors don't read it anyway and we don't have many non-regular contributors as of now. I am open to discussion...
- pre-commit doesn't work if the vscode git ui is used in gitpod, patch it to load the correct venv


egparedes:
I propose to make this PR specific about gitpod and I will clean up the pull request template part in another PR (https://github.com/GridTools/gt4py/pull/1189 )

egparedes (approved):
It looks good.

PR #1178, opened 2023-02-21
Pull images before using them for CSCS CI pipeline

finkandreas:
Since there are many test jobs using the same image, and all of them are starting at the same time, there is a high chance that the same image will be pulled several times.
While this gives correct results it is suboptimal since we download the same layers and squash them multiple times.

It is better to pull the image before using it, therefore I added a pull stage, which does exactly that.


DropD:
Already fixed by locking the image pull in the runner.

PR #1179, opened 2023-02-22
Replace partial shifts by neighbor lists

### LINKED ISSUE #1212: Bug: Inlining a neighbor_sum into a Koff
This is a minimal example of the bug, which results in a compiler error:

```
@field_operator
def _neighbor_sum_into_koff_bug(
    kh_smag_ec: Field[[EdgeDim, KDim], float],
) -> Field[[CellDim, KDim], float]:
    kh_c = neighbor_sum(kh_smag_ec(C2E), axis=C2EDim)
    hdef_ic = kh_c(Koff[-1])
    return hdef_ic
```

See also this icon4py PR: https://github.com/C2SM/icon4py/pull/148

Or specifically this file for a minimal example which can be used for debugging:
https://github.com/C2SM/icon4py/blob/2a750dac2d321a5fa02317d268595ae979eefd0a/atm_dyn_iconam/src/icon4py/atm_dyn_iconam/neighbor_sum_into_koff_bug.py

havogt:
Replaces reductions on partial shifts by reductions on _neighbor lists_ in the whole toolchain.

Iterator IR:
Partial shifts (`shift(V2E)(it)`) are no longer allowed. All shifts are full shifts e.g. `shift(V2E, 0)`. For reductions over neighbors we introduce a new neighbor list concept that are used in the following new builtins:
- `neighbors(offset: OffsetTag,  iterator: Iterator) -> List[dtype, max_length, has_skip_values]`: Returns a list of neighbor values corresponding to the connectivity identified by `offset`.
- `map(op: Callable, *args: List[...])`: Creates a new `List` such that the i-th element is the result of applying op on the i-th elements of `*args`.
- `make_const_list(value: dtype) -> List[...]`: Creates a `List` such that `list_get(i, make_const_list(value))` returns `value` for all `i`.
- `list_get(i: int, lst: List[dtype, ...]) -> dtype`: Returns the `i`-th element of a `List`.
Additionally the interface of an applied reduce is changed to:
- `reduce(op, init)(*args: List[...])`: Takes `List`s of values instead of partially shifted iterators.

Field view:
Lowering is simplified (partially taken from #965) which enables nested reductions in field view (`neighbor_sum(neighbor_sum(field(V2E), axis=...)(E2V), axis=...)`.

GTFN:
The C++ backend code is not changed, new passes are added to be able to generate equivalent code for unrolled reductions.

Fixes #1212


src/gt4py/next/iterator/type_inference.py
```diff
@@ -157,8 +157,9 @@ def handle_constraint(
                 )
                 for dtype, defined_loc in zip(dtypes, defined_locs)
             ]
-            add_constraint(self.dtypes, Tuple.from_elems(*dtypes))
-            add_constraint(self.defined_locs, Tuple.from_elems(*defined_locs))
+            for dtype, defined_loc in zip(dtypes, defined_locs):
+                add_constraint(self.dtypes, dtype)
+                add_constraint(self.defined_locs, defined_loc)
```
havogt, line 162: TODO make separate PR
Questions:
- Why are these 2 constraints handled separately and not included in the constraint on expanded

src/gt4py/next/iterator/type_inference.py
```diff
@@ -475,8 +512,82 @@ def visit_FunCall(
...
+                # TODO remove code duplication with tuple_get
+                # Calls to list_get are handled as being part of the grammar,
+                # not as function calls
+                if len(node.args) != 2:
+                    raise TypeError("list_get requires exactly two arguments")
+                if not isinstance(node.args[0], ir.Literal) or node.args[0].type != "int":
+                    raise TypeError("The first argument to list_get must be a literal int")
+                idx = int(node.args[0].value)
+                lst = self.visit(node.args[1], constraints=constraints, symtypes=symtypes)
+                kind = TypeVar.fresh()
```
havogt, line 539:
```suggestion
                kind = Value()
```

src/gt4py/next/iterator/type_inference.py
```diff
@@ -499,6 +546,54 @@ def visit_FunCall(
...
+                it = self.visit(node.args[1])
+                self.constraints.add(
+                    (
+                        it,
+                        Val(
+                            kind=Iterator(),
+                            dtype=dtype_,
+                            size=size,
+                            current_loc=current_loc_in,
+                            defined_loc=current_loc_out,  # TODO is this correct? since we are derefing in neighbors it has to?
```
fthaler, line 587: Looks correct to me, as the input iterator has to be derefable on the neighbor locations.

src/gt4py/next/iterator/transforms/trace_shifts.py
```diff
@@ -126,11 +126,15 @@ def apply(*args):
...
 def _reduce(f, init):
     def apply(*args):
-        return _combine(*args).shift((ALL_NEIGHBORS,)).deref()
+        return _combine(*args)
 
     return apply
 
 
+def _neighbors(o, x):
+    return x.shift((o, ALL_NEIGHBORS)).deref()
```
fthaler, line 135: Looks good (just because you asked specifically about this).

src/gt4py/next/iterator/transforms/fuse_maps.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    Fuses nested `map_`s.
+
+    Preconditions:
+      - `FunctionDefinitions` are inlined
+
+    Example:
+        map(λ(x, y)->f(x, y))(a, map(λ(z, w)->g(z, w))(b, c))
+    to
+        TODOmap(λ(a, b, c) → f(a, g(b, c)))(a, b, c)
```
fthaler, line 51:
```suggestion
        map(λ(a, b, c) → f(a, g(b, c)))(a, b, c)
```

fthaler (commented):
Looks good so far, just added some comments and remarks…

src/gt4py/next/iterator/transforms/unroll_reduce.py
```diff
@@ -123,6 +126,16 @@ def _make_if(cond: itir.Expr, true_expr: itir.Expr, false_expr: itir.Expr):
     )
 
 
+def _to_full_shifted_and_derefed(expr: itir.Expr, offset) -> itir.Expr:
+    if _is_neighbors(expr):
+        return _make_deref(_make_shift([expr.args[0], offset], expr.args[1]))
```
havogt, line 131: not needed because of collapse_list

src/gt4py/next/iterator/type_inference.py
```diff
@@ -221,6 +308,20 @@ class Iterator(Type):
...
+
+
+class BoolType(Type):
+    value: bool
+
+
+class List(Type):
+    dtype: Type = eve.field(default_factory=TypeVar.fresh)
+    max_length: Type = eve.field(default_factory=TypeVar.fresh)
+    has_skip_values: Type = eve.field(default_factory=TypeVar.fresh)
```
havogt, line 322: offset?

src/gt4py/next/iterator/type_inference.py
```diff
@@ -484,7 +531,7 @@ def visit_FunCall(
                     raise TypeError("The first argument to tuple_get must be a literal int")
                 idx = int(node.args[0].value)
                 tup = self.visit(node.args[1], **kwargs)
-                kind = TypeVar.fresh()
+                kind = TypeVar.fresh()  # TODO can be Value() as we cannot have tuple of iterator?
```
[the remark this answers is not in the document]
havogt: add a comment

src/gt4py/next/iterator/type_inference.py
```diff
@@ -499,6 +546,54 @@ def visit_FunCall(
                 )
                 self.constraints.add((tup, val))
                 return Val(kind=kind, dtype=elem, size=size)
+            if node.fun.id == "make_list":
+                elems = list(self.visit(arg) for arg in node.args)
```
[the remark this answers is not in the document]
havogt: deleted make_list for now...

src/gt4py/next/iterator/type_inference.py
```diff
@@ -499,6 +546,54 @@ def visit_FunCall(
...
+                    connectivity = self.offset_provider[node.args[0].value]
+                    assert isinstance(connectivity, Connectivity)
+                    max_length = Length(length=connectivity.max_neighbors)
+                    has_skip_values = BoolType(value=connectivity.has_skip_values)
+                current_loc_in, current_loc_out = _infer_shift_location_types(
+                    [node.args[0]], self.offset_provider, self.constraints
+                )
+                dtype_ = TypeVar.fresh()
+                size = TypeVar.fresh()
+                it = self.visit(node.args[1])
```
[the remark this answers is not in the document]
havogt: Actually it propagates the symtable. maybe there is a problem with that pass because symtable should be propagated implicitly...
Leaving this open for investigation.

src/gt4py/next/iterator/transforms/collapse_list_get.py
```diff
@@ -0,0 +1,53 @@
+# GT4Py - GridTools Framework
```
havogt, line 1: TODO test

tests/next_tests/iterator_tests/test_with_toy_connectivity.py
```diff
@@ -80,20 +106,41 @@ def test_sum_edges_to_vertices(program_processor, lift_mode):
 
 
 @fundef
-def sum_edges_to_vertices_reduce(in_edges):
```
havogt, line 83: to help the reviewer: this test is now a parameter of the test above

tests/next_tests/iterator_tests/test_with_toy_connectivity.py
```diff
@@ -214,30 +261,30 @@ def slice_twice_sparse_stencil(sparse):
     return deref(shift(2)(shift(1)(sparse)))
 
 
-def test_slice_twice_sparse(program_processor_no_gtfn_exec, lift_mode):
-    program_processor, validate = program_processor_no_gtfn_exec
-    inp = np_as_located_field(Vertex, V2V, V2V)(v2v_arr[v2v_arr])
-    out = np_as_located_field(Vertex)(np.zeros([9]))
+# def test_slice_twice_sparse(program_processor_no_gtfn_exec, lift_mode):
```
havogt, line 264: TODO implement double shifted field in embedded.
Is it `It[...,List[List[dtype]]` or `It[..., List[It[List[dtype]]]]`?

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -153,32 +71,17 @@ class FieldOperatorLowering(NodeTranslator):
...
-        for i, param in enumerate(func_definition.params):
-            if iterator_type_kind(node.definition.params[i].type) == ITIRTypeKind.VALUE:
-                new_body = im.let(param.id, im.deref_(param.id))(new_body)
-
-        assert isinstance(node.type, ts_ffront.FieldOperatorType)
-        if iterator_type_kind(node.type.definition.returns) == ITIRTypeKind.ITERATOR:
-            new_body = im.deref_(new_body)
+        # The pass works by making every expression a lifted stencil (i.e. takes iterators, returns iterator),
+        # except FieldOperators are lowered to stencils (i.e. takes iterators, returns value) and lifted in calls.
+        # TODO for consistency we could lift here, then deref in the program call.
```
havogt, line 83: discuss with @tehrengruber

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -243,124 +153,50 @@ def visit_Assign(
...
         dtype = type_info.extract_dtype(node.type)
-        assert type_info.is_arithmetic(dtype)
-        return self._lift_if_field(node)(
-            im.call_(node.op.value)(
-                im.literal_("0", dtype.kind.name.lower()),
-                to_value(node.operand)(self.visit(node.operand, **kwargs)),
-            )
-        )
+        if node.op in [dialect_ast_enums.UnaryOperator.NOT, dialect_ast_enums.UnaryOperator.INVERT]:
+            # TODO: invert only for bool right now
```
havogt, line 178: ?

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -436,144 +225,107 @@ def visit_Call(self, node: foast.Call, **kwargs) -> itir.FunCall | itir.Literal:
...
         )
 
     def _visit_astype(self, node: foast.Call, **kwargs) -> itir.FunCall:
-        obj = (to_value(node.args[0]))(self.visit(node.args[0], **kwargs))
-        dtype = self.visit(node.args[1]).id
-        return self._lift_lambda(node)(im.call_("cast_")(obj, dtype))
+        assert len(node.args) == 2 and isinstance(node.args[1], foast.Name)
+        obj, dtype = node.args[0], node.args[1].id
+
+        # TODO check that we test astype that results in a itir.map_ operation
```
havogt, line 244: TODO

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -436,144 +225,107 @@ def visit_Call(self, node: foast.Call, **kwargs) -> itir.FunCall | itir.Literal:
...
+        return self._map(self.visit(node.func, **kwargs), *node.args)
+
+    def _make_reduction_expr(
+        self,
+        node: foast.Call,
+        op: Any,  # TODO fix
+        init_expr: int | itir.Literal,
+        **kwargs,
+    ):
+        # TODO(havogt): deal with nested reductions of the form neighbor_sum(neighbor_sum(field(off1)(off2)))
```
havogt, line 263: This is partially implemented in a follow up pr #1198

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -436,144 +225,107 @@ def visit_Call(self, node: foast.Call, **kwargs) -> itir.FunCall | itir.Literal:
...
 
     def _visit_type_constr(self, node: foast.Call, **kwargs) -> itir.Literal:
         if isinstance(node.args[0], foast.Constant):
             node_kind = self.visit(node.type).kind.name.lower()
             target_type = fbuiltins.BUILTINS[node_kind]
             source_type = {**fbuiltins.BUILTINS, "string": str}[node.args[0].type.__str__().lower()]
             if target_type is bool and source_type is not bool:
-                return im.literal_(str(bool(source_type(node.args[0].value))), node.func.id)
-            return im.literal_(str(node.args[0].value), node_kind)
+                return im.as_lifted_capture(  # TODO seems this codepath is not tested
```
havogt, line 291: test this

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -436,144 +225,107 @@ def visit_Call(self, node: foast.Call, **kwargs) -> itir.FunCall | itir.Literal:
...
+                    im.literal_(str(bool(source_type(node.args[0].value))), node.func.id)
+                )
+            return im.as_lifted_capture(im.literal_(str(node.args[0].value), node_kind))
         raise FieldOperatorLoweringError(f"Encountered a type cast, which is not supported: {node}")
 
     def _make_literal(self, val: Any, type_: ts.TypeSpec) -> itir.Literal:
         # TODO(tehrengruber): check constant of this type is supported in iterator ir
```
havogt, line 298: It is not supported in embedded execution, because of an implementation detail: the `offset_provider` is propagated via iterators. If we `lift(lambda: 1.0)()` we will try to extract the offset_provider from the first argument of the lifted lambda (which has no arguments).
However, after we apply all passes, these expressions are inlined...
We should add an executor which does embedded without optimizations to make the pipline cleaner.

src/gt4py/next/ffront/itir_makers.py
```diff
@@ -271,5 +271,29 @@ def shift_(offset, value=None):
     return call_(call_("shift")(*args))
 
 
+def neighbors_(offset, it):
+    offset = ensure_offset(offset)  # TODO this function should not exist?
```
havogt, line 275:
```suggestion
    offset = ensure_offset(offset)
```
I think it should exist here, because itir_makers provides convenience functions, i.e. should automatically wrap in `OffsetLiteral` when needed.

Maybe rename to `as_offset` or `to_offset`.

src/gt4py/next/ffront/itir_makers.py
```diff
@@ -271,5 +271,29 @@ def shift_(offset, value=None):
     return call_(call_("shift")(*args))
 
 
+def neighbors_(offset, it):
```
havogt, line 274: TODO: document the functions following the style of this file

src/gt4py/next/iterator/embedded.py
```diff
@@ -887,20 +869,26 @@ def make_in_iterator(
             sparse_dimensions.append(axis.value)
 
     new_pos: Position = pos.copy()
-    for sparse_dim in set(sparse_dimensions):
-        init = [None] * sparse_dimensions.count(sparse_dim)
+    sparse_dimensions = list(set(sparse_dimensions))
+    assert len(sparse_dimensions) <= 1, sparse_dimensions
+    for sparse_dim in sparse_dimensions:
+        init = [None]  # * sparse_dimensions.count(sparse_dim)
```
havogt, line 875: TODO add back support for more than one sparse dimension

tests/next_tests/iterator_tests/test_fuse_maps.py
```diff
@@ -0,0 +1,212 @@
...
+                ],
+            ),
+        ),
+        ir.SymRef(id="a"),
+        ir.SymRef(id="b"),
+        ir.SymRef(id="c"),
+    )
+
+    actual = FuseMaps().visit(testee)
+    # assert expected == actual #TODO pattern match the result
```
havogt, line 60: TODO

tests/next_tests/iterator_tests/test_fvm_nabla.py
```diff
@@ -68,9 +81,8 @@ def compute_zavgS_fencil(
 @fundef
 def compute_pnabla(pp, S_M, sign, vol):
     zavgS = lift(compute_zavgS)(pp, S_M)
-    # pnabla_M = reduce(lambda a, b, c: a + b * c, 0.0)(shift(V2E)(zavgS), sign)
-    # pnabla_M = library.sum(lambda a, b: a * b)(shift(V2E)(zavgS), sign)
-    pnabla_M = library.dot(shift(V2E)(zavgS), sign)
+    pnabla_M = library.dot(neighbors(V2E, zavgS), deref(sign))
+    # pnabla_M = reduce(lambda a, b, c: a + b * c, 0.0)(neighbors(V2E)(zavgS), deref(sign))
```
havogt, line 85:
```suggestion
```

src/gt4py/next/ffront/foast_to_itir.py
havogt: I added this as a task to the cleanup/refactoring project for next cycle. I could add a test now that specifically tests as_type in a mapped context, but I prefer to setup a test-case that allows testing other kinds of builtins in mapped context as well.

src/gt4py/next/ffront/foast_to_itir.py
havogt: actually tested in the lowering tests

src/gt4py/next/ffront/foast_to_itir.py
havogt: add check that we do `~field` only on bool fields

havogt: in type deduction, and assert here

src/gt4py/next/ffront/foast_to_itir.py
havogt, line 83:
```suggestion
        # including calls to FieldOperators (but FieldOperators itself are non-lifted stencils)
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -1114,6 +1102,108 @@ def impl(it: ItIterator) -> ItIterator:
...
+    def offset_provider(self):
+        return self.it.offset_provider
+
+    def deref(self) -> Any:
+        return _List(
+            builtins.deref(builtins.shift(*self.offsets, SparseTag(self.list_offset), i)(self.it))
+            if builtins.can_deref(
+                builtins.shift(*self.offsets, SparseTag(self.list_offset), i)(self.it)
+            )
+            else None
```
havogt, line 1191: remove None

src/gt4py/next/iterator/type_inference.py
```diff
@@ -179,6 +179,83 @@ def handle_constraint(
...
+            and self.size == other.size
+        )
+
+    def handle_constraint(
+        self, other: Type, add_constraint: abc.Callable[[Type, Type], None]
+    ) -> bool:
+        if isinstance(other, Tuple):
+            list_dtypes = [TypeVar.fresh() for _ in other]
+            max_length = TypeVar.fresh()
+            has_skip_values = TypeVar.fresh()
```
fthaler, line 239: Not needed.

src/gt4py/next/iterator/type_inference.py
```diff
@@ -179,6 +179,83 @@ def handle_constraint(
...
+            expanded = [
+                Val(
+                    kind=self.kind,
+                    dtype=List(dtype=dtype, max_length=max_length, has_skip_values=has_skip_values),
```
fthaler, line 243: You can directly use `self.max_length` and `self.has_skip_values` here.
```suggestion
                    dtype=List(dtype=dtype, max_length=self.max_length, has_skip_values=self.has_skip_values),
```

src/gt4py/next/iterator/type_inference.py
```diff
@@ -179,6 +179,83 @@ def handle_constraint(
...
+                    size=self.size,
+                )
+                for dtype in list_dtypes
+            ]
+            add_constraint(self.list_dtypes, Tuple.from_elems(*list_dtypes))
+            add_constraint(self.max_length, max_length)
+            add_constraint(self.has_skip_values, has_skip_values)
```
fthaler, line 250: Not needed. The additional `list_dtypes` is required because `self.list_dtypes` can be a tuple, but also a simple type var. In the latter case, we can not directly impose the per-element constraints, thus we need to create the expanded tuple and add this additional equality constraints. `self.max_length` and `self.has_skip_values` do not require per-element constraints and thus can be used directly.

src/gt4py/next/iterator/type_inference.py
```diff
@@ -439,7 +559,7 @@ def visit_SymRef(self, node: ir.SymRef, *, symtable, **kwargs) -> Type:
         if node.id in (ir.BUILTINS - ir.TYPEBUILTINS):
             raise NotImplementedError(f"Missing type definition for builtin '{node.id}'")
 
-        return TypeVar.fresh()
+        return TypeVar.fresh()  # TODO document when this case (LetPolymorphic?)
```
fthaler, line 562: The simplest case for this path is quite trivial: it is required for type inference of the basic ITIR expression consisting of just a single symbol (e.g., `itir.SymRef(id="x")`).

fthaler (changes requested):

src/gt4py/next/iterator/embedded.py
```diff
@@ -887,20 +869,26 @@ def make_in_iterator(
...
     if column_axis is not None:
         # if we deal with column stencil the column position is just an offset by which the whole column needs to be shifted
         assert _column_range is not None
         new_pos[column_axis] = _column_range.start
-    return MDIterator(
+    it = MDIterator(
         inp,
         new_pos,
-        incomplete_offsets=[SparseTag(x) for x in sparse_dimensions],
+        incomplete_offsets=[],
```
egparedes, line 884: Are `incomplete_offsets` still used in actual code? If they are useless and meant to be removed soon, I would add a TODO comment in the `MDIterator` class definition.

src/gt4py/next/iterator/embedded.py
```diff
@@ -1114,6 +1102,108 @@ def impl(it: ItIterator) -> ItIterator:
...
+
+@builtins.neighbors.register(EMBEDDED)
+def neighbors(offset: runtime.Offset, it: ItIterator) -> _List:
+    # TODO offset provider should be part of the iterator interface?
+    offset_str = offset.value if isinstance(offset, runtime.Offset) else offset
+    return _List(
+        builtins.deref(it.shift(offset_str, i))
+        if builtins.can_deref(it.shift(offset_str, i))
+        else None
+        for i in range(it.offset_provider[offset_str].max_neighbors)  # type: ignore
```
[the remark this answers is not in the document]
egparedes: Use an assignment expression to avoid repeating expressions:
```suggestion
        builtins.deref(shifted_it)
        for i in range(it.offset_provider[offset_str].max_neighbors)  # type: ignore
        if builtins.can_deref(shifted_it := (it.shift(offset_str, i)))
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -1114,6 +1102,108 @@ def impl(it: ItIterator) -> ItIterator:
...
+            if isinstance(cur, _List):
+                lst = cur
+                break
+        n = len(lst)
+        res = init
+        for i in range(n):
+            # we can check a single argument
+            # because all arguments share the same pattern
+            if lst[i] is None:
+                break
```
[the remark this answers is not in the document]
egparedes: I agree with @fthaler

src/gt4py/next/iterator/embedded.py
```diff
@@ -1114,6 +1102,108 @@ def impl(it: ItIterator) -> ItIterator:
...
+        return self.it.offset_provider
+
+    def deref(self) -> Any:
+        return _List(
+            builtins.deref(builtins.shift(*self.offsets, SparseTag(self.list_offset), i)(self.it))
+            if builtins.can_deref(
+                builtins.shift(*self.offsets, SparseTag(self.list_offset), i)(self.it)
+            )
+            else None
+            for i in range(self.it.offset_provider[self.list_offset].max_neighbors)  # type: ignore
```
egparedes, line 1192: Same rationale of previous comments by @fthaler of using lists of different lengths instead of using lists with optionals:
```suggestion
            builtins.deref(shifted_it)
            for i in range(self.it.offset_provider[self.list_offset].max_neighbors)  # type: ignore
            if builtins.can_deref(
                shifted_it := (builtins.shift(*self.offsets, SparseTag(self.list_offset), i)(self.it))
            )
```

egparedes (approved):

egparedes (changes requested):
Ooops, I made a mistake earlier. I didn't mean to approve

src/gt4py/next/iterator/embedded.py
```diff
@@ -1114,6 +1102,108 @@ def impl(it: ItIterator) -> ItIterator:
...
+class _ConstList(Generic[DT]):
+    value: DT
+
+    def __getitem__(self, _):
+        return self.value
+
+
+@builtins.neighbors.register(EMBEDDED)
+def neighbors(offset: runtime.Offset, it: ItIterator) -> _List:
+    # TODO offset provider should be part of the iterator interface?
```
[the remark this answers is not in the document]
havogt: implemented as described

src/gt4py/next/iterator/embedded.py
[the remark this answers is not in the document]
havogt: done

src/gt4py/next/iterator/embedded.py
```diff
@@ -1114,6 +1102,108 @@ def impl(it: ItIterator) -> ItIterator:
...
+
+@builtins.make_const_list.register(EMBEDDED)
+def make_const_list(value):
+    return _ConstList(value)
+
+
+@builtins.reduce.register(EMBEDDED)
+def reduce(fun, init):
+    def sten(*lists):
+        # TODO: assert check_that_all_iterators_are_compatible(*iters)
```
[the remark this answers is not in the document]
havogt: done

src/gt4py/next/iterator/transforms/collapse_list_get.py
```diff
@@ -0,0 +1,53 @@
...
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from dataclasses import dataclass
+
+from gt4py import eve
+from gt4py.next.iterator import ir
+
+
+@dataclass(frozen=True)
```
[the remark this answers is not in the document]
havogt: copy pasted :(

src/gt4py/next/iterator/embedded.py
havogt: done

src/gt4py/next/iterator/transforms/unroll_reduce.py
```diff
@@ -133,11 +140,7 @@ class UnrollReduce(NodeTranslator):
     def apply(cls, node: itir.Node, **kwargs):
         return cls().visit(node, **kwargs)
 
-    def visit_FunCall(self, node: itir.FunCall, **kwargs):
-        node = self.generic_visit(node, **kwargs)
-        if not _is_reduce(node):
-            return node
-
+    def _visit_reduce(self, node: itir.FunCall, **kwargs) -> itir.Expr:
```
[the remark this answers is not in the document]
havogt: The suggestion is correct, but what I care is that it is `itir.Expr`. Should I still maximally constrain? I want that users of this function assume it's an Expr not a specific Expr (e.g. we could refactor such that it return `init` if `max_neighbors == 0`).

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -35,104 +32,26 @@ def is_local_kind(symbol_type: ts.FieldType) -> bool:
...
-    """
-    if iterator_type_kind(node.type) is ITIRTypeKind.ITERATOR:
-        # just to ensure we don't accidentally deref a local field
-        assert not (isinstance(node.type, ts.FieldType) and is_local_kind(node.type))
-        return im.deref_
-    elif iterator_type_kind(node.type) is ITIRTypeKind.VALUE:
-        return lambda x: x
-
-    raise AssertionError(f"Type {node.type} can not be turned into a value.")
+def promote_to_list(node: foast.Symbol | foast.Expr):
```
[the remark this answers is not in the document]
havogt: done

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -35,104 +32,26 @@ def is_local_kind(symbol_type: ts.FieldType) -> bool:
     return any(dim.kind == DimensionKind.LOCAL for dim in symbol_type.dims)
 
 
-class ITIRTypeKind(enum.Enum):
-    VALUE = 0
-    ITERATOR = 1
-    ENCAPSULATED_ITERATOR = 2
+def is_local_type_kind(type_):
```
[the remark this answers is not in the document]
havogt: done

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -243,124 +150,52 @@ def visit_Assign(
...
         dtype = type_info.extract_dtype(node.type)
-        assert type_info.is_arithmetic(dtype)
-        return self._lift_if_field(node)(
-            im.call_(node.op.value)(
-                im.literal_("0", dtype.kind.name.lower()),
-                to_value(node.operand)(self.visit(node.operand, **kwargs)),
-            )
+        if node.op in [dialect_ast_enums.UnaryOperator.NOT, dialect_ast_enums.UnaryOperator.INVERT]:
+            if dtype.kind != ts.ScalarKind.BOOL:
+                raise AssertionError(f"{node.op} is only supported on `bool`s.")
```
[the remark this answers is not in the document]
havogt: went for NotImplementedError

src/gt4py/next/iterator/type_inference.py
havogt: done

src/gt4py/next/iterator/type_inference.py
havogt: done

src/gt4py/next/iterator/type_inference.py
havogt: done

src/gt4py/next/iterator/type_inference.py
havogt: I think I added this comment, because this default case (lookup in symtable failed) can be reached if you forget to pass a symtable (e.g. by not propagating in visits). I don't remember now the exact case. I guess type inference should also work for symbols that are not defined (in the current subtree), therefore probably it cannot be improved.

src/gt4py/next/iterator/transforms/unroll_reduce.py
```diff
@@ -166,3 +170,9 @@ def visit_FunCall(self, node: itir.FunCall, **kwargs):
         )
 
         return expr
+
+    def visit_FunCall(self, node: itir.FunCall, **kwargs):
```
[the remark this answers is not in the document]
havogt: done, but used itir.Expr, see comment above.

src/gt4py/next/iterator/transforms/fuse_maps.py
```diff
@@ -0,0 +1,128 @@
...
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import dataclasses
+from typing import TypeGuard
+
+from gt4py.eve import NodeTranslator, traits
+from gt4py.eve.utils import UIDGenerator
+from gt4py.next.iterator import ir
+from gt4py.next.iterator.transforms.inline_lambdas import inline_lambda
```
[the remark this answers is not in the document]
havogt: done

src/gt4py/next/iterator/type_inference.py
havogt: removed the comment...

src/gt4py/next/iterator/transforms/fuse_maps.py
```diff
@@ -0,0 +1,128 @@
...
+    def _as_lambda(self, fun: ir.SymRef | ir.Lambda, param_count: int) -> ir.Lambda:
+        if isinstance(fun, ir.Lambda):
+            return fun
+        params = [ir.Sym(id=self.uids.sequential_id(prefix="sym")) for _ in range(param_count)]
+        return ir.Lambda(
+            params=params,
+            expr=ir.FunCall(fun=fun, args=[ir.SymRef(id=p.id) for p in params]),
+        )
+
+    # TODO think about clashes of symbol names
```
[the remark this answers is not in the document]
havogt: I think we are missing utilities (or they are hidden in passes) to easily create symbols that don't clash. I added a task in the cleanup project for next cycle, because I thought it's better to approach that systematically. https://hackmd.io/46Oa6MspRoSTiFozHzun5A#discuss-itir-conventions
I am not sure if that is enough to merge this pr...

src/gt4py/next/iterator/embedded.py
havogt: added a comment

src/gt4py/next/iterator/embedded.py
```diff
@@ -589,28 +568,18 @@ def execute_shift(
...
-            tag_stack.append(offset)
+            tag = offset
         else:
-            assert tag_stack
-            tag = tag_stack.pop()
+            assert tag is not None
             complete_offsets.append((tag, offset))
-    return complete_offsets, tag_stack
+            tag = None
+    assert tag is None
```
egparedes, line 581: It works but it's a bit too verbose/ugly. Would you consider other alternatives? e.g.: 
```python
while offsets:
    tag, offset, *offsets = offsets
    assert isinstance(tag, Tag) and isinstance(offset, (int, np.integer))
    complete_offsets.append((tag, offset))
```

```python
assert len(offsets) % 2 == 0
complete_offsets = [(tag, offset) for tag, offset in zip(offsets[::2], offsets[1::2])]
assert all(
    isinstance(tag, Tag) and isinstance(offset, (int, np.integer))
    for tag, offset in complete_offsets
)
```        

```python
assert len(offsets) % 2 == 0
complete_offsets = [(offsets[2 * i], offsets[2 * i + 1]) for i in range(len(offsets) // 2)]
assert all(
    isinstance(tag, Tag) and isinstance(offset, (int, np.integer))
    for tag, offset in complete_offsets
)
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -851,28 +808,26 @@ def deref(self) -> Any:
         )
 
 
+def _get_sparse_dimensions(axes: Sequence[common.Dimension | runtime.Offset]) -> list[Tag]:
+    return [
+        axis.value  # type: ignore[misc] # axis.value is always `str`
```
egparedes, line 813:
```suggestion
        cast(Tag, axis.value)  # axis.value is always `str`

```

egparedes (approved):
I have a couple of suggestions but it looks good to me overall.

src/gt4py/next/iterator/transforms/unroll_reduce.py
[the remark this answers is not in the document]
havogt: My question would be: is it Python best practices to use always the most derived type or instead document what the user of my function should assume about the return type?
The refactoring argument would be, with `FunCall`, users of my function can assume more than with `Expr`, so I might have to change not only my function but would have to check on all call sites. Again, not super relevant for this case, but I would like to understand what the best practices are in general.

src/gt4py/next/iterator/embedded.py
havogt: I was trying to find the canonical way to write your second version, but also the pattern of the first version was not in my toolbox...

havogt: Which one would you prefer? I like (2) without asserts, but (1) with asserts.

havogt: took 1 for now

havogt: switched to 2 because I didn't see a nice way to fix the typing in 1 (offsets: list vs tuple)

havogt: I hate it... back to 1 with type ignore

src/gt4py/next/iterator/transforms/fuse_maps.py
```diff
@@ -0,0 +1,128 @@
...
+    to
+        map(λ(a, b, c) → f(a, g(b, c)))(a, b, c)
+    """
+
+    uids: UIDGenerator = dataclasses.field(init=False, repr=False, default_factory=UIDGenerator)
+
+    def _as_lambda(self, fun: ir.SymRef | ir.Lambda, param_count: int) -> ir.Lambda:
+        if isinstance(fun, ir.Lambda):
+            return fun
+        params = [ir.Sym(id=self.uids.sequential_id(prefix="sym")) for _ in range(param_count)]
```
tehrengruber, line 59:
```suggestion
        params = [ir.Sym(id=self.uids.sequential_id(prefix="_fuse_maps")) for _ in range(param_count)]
```
The current (fragile) approach is to use a unique prefix per pass and only construct the pass once. Just using `sym` looks dangerous the me as `fun` might capture a `sym` from outside. I would also add a precondition in the docstring above reflecting that.

src/gt4py/next/iterator/transforms/fuse_maps.py
```diff
@@ -0,0 +1,128 @@
...
+                    return ir.FunCall(
+                        fun=ir.FunCall(fun=ir.SymRef(id="map_"), args=[new_op]),
+                        args=new_args,
+                    )
+                elif _is_reduce(node):
+                    return ir.FunCall(
+                        fun=ir.FunCall(fun=ir.SymRef(id="reduce"), args=[new_op, node.fun.args[1]]),
+                        args=new_args,
+                    )
+                raise AssertionError("unreachable")
```
tehrengruber, line 127:
```suggestion
                raise AssertionError("Unreachable. Node must be a `map` or `reduce` call.")
```
Maybe.

src/gt4py/next/iterator/transforms/trace_shifts.py
```diff
@@ -126,11 +126,15 @@ def apply(*args):
 
 def _reduce(f, init):
```
tehrengruber, line 127:
```suggestion
def _reduce(f, init):
    return _combine
```

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -35,104 +32,28 @@ def is_local_kind(symbol_type: ts.FieldType) -> bool:
...
+    if not is_local_type_kind(node.type):
+        return lambda x: im.as_lifted_lambda("make_const_list", x)
+    return lambda x: x
 
 
 class FieldOperatorLowering(NodeTranslator):
     """
     Lower FieldOperator AST (FOAST) to Iterator IR (ITIR).
 
+    The strategy is to lower every expression to lifted stencils,
```
tehrengruber, line 54: This change in approach invalidates `docs/development/ADRs/0002-Field_View_Lowering.md` (already outdated before this PR). Might be something I should take care of. Let's discuss in person.

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -153,65 +74,53 @@ class FieldOperatorLowering(NodeTranslator):
...
         func_definition: itir.FunctionDefinition = self.visit(node.definition, **kwargs)
         new_body = func_definition.expr
-        for param in func_definition.params[1:]:
-            new_body = im.let(param.id, im.deref_(param.id))(new_body)
+
+        # promote carry to iterator
+        # (this is the only place in the lowering were a variable is captured in a lifted lambda)
+        new_body = im.let(
+            func_definition.params[0].id,
+            im.lift_(im.lambda__()(func_definition.params[0].id))(),
```
tehrengruber, line 119:
```suggestion
            im.as_lifted_capture(func_definition.params[0].id)
```

src/gt4py/next/iterator/type_inference.py
```diff
@@ -475,108 +590,163 @@ def visit_Lambda(
...
-                dtype = Tuple(front=elem, others=TypeVar.fresh())
-                for _ in range(idx):
-                    dtype = Tuple(front=TypeVar.fresh(), others=dtype)
-
-                val = Val(
-                    kind=kind,
-                    dtype=dtype,
+    def _visit_make_tuple(self, node: ir.FunCall, **kwargs) -> Type:
+        # Calls to make_tuple are handled as being part of the grammar,
+        # not as function calls
```
tehrengruber, line 595:
```suggestion
        # Calls to `make_tuple` are handled as being part of the grammar, not as function calls.
```
line length + backticks.

src/gt4py/next/iterator/type_inference.py
```diff
@@ -475,108 +590,163 @@ def visit_Lambda(
...
+        )  # `kind == Iterator()` means zipping iterators into an iterator of tuples
+        size = TypeVar.fresh()
+        dtype = Tuple.from_elems(*(TypeVar.fresh() for _ in argtypes))
+        for d, a in zip(dtype, argtypes):
+            self.constraints.add((Val(kind=kind, dtype=d, size=size), a))
+        return Val(kind=kind, dtype=dtype, size=size)
+
+    def _visit_tuple_get(self, node: ir.FunCall, **kwargs) -> Type:
+        # Calls to tuple_get are handled as being part of the grammar,
+        # not as function calls
```
tehrengruber, line 608:
```suggestion
        # Calls to `tuple_get` are handled as being part of the grammar, not as function calls.
```
line length + backticks.

src/gt4py/next/iterator/type_inference.py
```diff
@@ -475,108 +590,163 @@ def visit_Lambda(
...
+        if len(node.args) != 2:
+            raise TypeError("tuple_get requires exactly two arguments")
```
tehrengruber, line 610:
```suggestion
            raise TypeError("`tuple_get` requires exactly two arguments.")
```

src/gt4py/next/iterator/type_inference.py
```diff
@@ -475,108 +590,163 @@ def visit_Lambda(
...
+        if not isinstance(node.args[0], ir.Literal) or node.args[0].type != "int":
+            raise TypeError("The first argument to tuple_get must be a literal int")
```
tehrengruber, line 612:
```suggestion
            raise TypeError("The first argument to `tuple_get` must be a literal int.")
```

tehrengruber (commented):
First round. Haven't reviewed everything yet.

src/gt4py/next/iterator/transforms/trace_shifts.py
havogt: done

src/gt4py/next/ffront/itir_makers.py
```diff
@@ -272,4 +272,55 @@ def shift_(offset, value=None):
...
+
+
+def as_lifted_capture(expr: str | itir.Expr):
+    """
+    Create a lifted nullary lambda that captures `expr`.
+
+    Examples
+    --------
+    >>> as_lifted_capture("foo")
+    FunCall(fun=FunCall(fun=SymRef(id=SymbolRef('lift')), args=[Lambda(params=[], expr=SymRef(id=SymbolRef('foo')))]), args=[])
```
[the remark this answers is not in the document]
havogt: done

src/gt4py/next/ffront/itir_makers.py
```diff
@@ -272,4 +272,55 @@ def shift_(offset, value=None):
...
+    """
+    return lift_(lambda__()(expr))()
+
+
+def as_lifted_lambda(op: str | Callable, *its):
+    """
+    Wrap `op` in a lifted lambda and call it with `*its`.
+
+    Examples
```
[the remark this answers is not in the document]
havogt: done

src/gt4py/next/iterator/transforms/fuse_maps.py
```diff
@@ -0,0 +1,128 @@
...
+@dataclasses.dataclass(frozen=True)
+class FuseMaps(traits.VisitorWithSymbolTableTrait, NodeTranslator):
+    """
+    Fuses nested `map_`s.
+
+    Preconditions:
+      - `FunctionDefinitions` are inlined
+
+    Example:
+        map(λ(x, y)->f(x, y))(a, map(λ(z, w)->g(z, w))(b, c))
```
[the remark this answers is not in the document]
havogt: no, will add a test and fix

src/gt4py/next/iterator/embedded.py
egparedes: Actually suggestion 2 is more complicated than needed. What about this ?
```python
    complete_offsets = [*zip(offsets[:-1:2], offsets[1::2])]
    assert (len(offsets) % 2 == 0) and all(
        isinstance(tag, Tag) and isinstance(offset, (int, np.integer))
        for tag, offset in complete_offsets
    ), f"Invalid sequence of offset parts: {offsets}"
```

src/gt4py/next/iterator/transforms/unroll_reduce.py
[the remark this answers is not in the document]
egparedes: To get as safe as possible type checks, the best practice is to be as specific as possible in the annotations of the outputs and as permissive as possible in the inputs of a function. The current implementation always returns `itir.FunCall`s but you say that this is just an implementation detail and the (implicit) semantics are that the caller should treat the returned value as an `itir.Expr`, right? In this case I think you're right and the most appropriate annotation would be `Expr`, because we want the type checker to catch any use of the returned value where an `Expr` is not good enough.

src/gt4py/next/iterator/transforms/fuse_maps.py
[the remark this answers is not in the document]
havogt: done

src/gt4py/next/iterator/transforms/fuse_maps.py
```diff
@@ -0,0 +1,128 @@
...
+    to
+        map(λ(a, b, c) → f(a, g(b, c)))(a, b, c)
+    """
```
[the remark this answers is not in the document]
havogt: done

src/gt4py/next/iterator/transforms/fuse_maps.py
havogt: changed, but with just `else`

src/gt4py/next/ffront/foast_to_itir.py
havogt: I marked the ADR as deprecated.

src/gt4py/next/iterator/transforms/unroll_reduce.py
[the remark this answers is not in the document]
havogt: Thanks!

src/gt4py/next/iterator/embedded.py
havogt: I slightly prefer the original version 2 for clarity, I didn't take it because I couldn't convince mypy that the result has the correct type. But now I added a TypeGuard which works. Pushing in a second.

tests/next_tests/iterator_tests/test_fuse_maps.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def test_simple():
+    testee = _map(
+        ir.SymRef(id="plus"),
+        ir.SymRef(id="a"),
+        _map(ir.SymRef(id="multiplies"), ir.SymRef(id="b"), ir.SymRef(id="c")),
+    )
+
+    expected = _map_p(
```
havogt, line 56: maybe deepdiff

havogt: check python pattern matching

src/gt4py/next/iterator/transforms/fuse_maps.py
```diff
@@ -0,0 +1,131 @@
...
+                        new_args.extend(map_call.args)
+                    else:
+                        inlined_args.append(ir.SymRef(id=outer_op.params[i + first_param].id))
+                        new_params.append(outer_op.params[i + first_param])
+                        new_args.append(node.args[i])
+                new_body = ir.FunCall(
+                    fun=outer_op,
+                    args=inlined_args,
+                )
+                new_body = inline_lambdas.InlineLambdas.apply(new_body)
```
havogt, line 116: TODO comment why we need the recursive version

havogt: TODO test if we really need the inlining for unroll_reduce (if not we can rely on the final inline_lambda in pass_manager).

src/gt4py/next/iterator/transforms/fuse_maps.py
```diff
@@ -0,0 +1,131 @@
...
+
+                for i in range(len(node.args)):
+                    if _is_map(node.args[i]):
+                        map_call = node.args[i]
+                        assert isinstance(map_call, ir.FunCall)
+                        assert isinstance(map_call.fun, ir.FunCall)
+                        assert isinstance(map_call.fun.args[0], (ir.Lambda, ir.SymRef))
+                        inner_op = self._as_lambda(map_call.fun.args[0], len(map_call.args))
+                        inlined_args.append(
+                            inline_lambdas.inline_lambda(
```
havogt, line 99: remove this if the recursive inline runs below

havogt: maybe

tests/next_tests/iterator_tests/test_with_toy_connectivity.py
```diff
@@ -79,20 +105,41 @@ def test_sum_edges_to_vertices(program_processor, lift_mode):
...
-    program_processor, validate = program_processor
-    if program_processor == gtfn_cpu.run_gtfn_imperative:
-        pytest.skip("gtfn_imperative has a bug.")
-    inp = edge_index_field()
+def test_map_neighbors(program_processor_no_gtfn_exec, lift_mode):
+    program_processor, validate = program_processor_no_gtfn_exec
+    inp = index_field(Edge)
     out = np_as_located_field(Vertex)(np.zeros([9]))
-    ref = np.asarray(list(sum(row) for row in v2e_arr))
+    ref = np.asarray(list(sum(row) for row in v2e_arr)) * 2.0
```
tehrengruber, line 116:
```suggestion
    ref = 2*sum(v2e_arr.table, axis=1)
```

src/gt4py/next/iterator/transforms/fuse_maps.py
havogt: kept the non-recursive one here and below...

src/gt4py/next/iterator/transforms/fuse_maps.py
havogt: the inlining is not needed for unroll_reduce, therefore I reverted back to the non-recursive inliner here. Instead I added the recursive inliner in the tests to have simpler and more stable patterns.

tests/next_tests/iterator_tests/test_fuse_maps.py
havogt: Something like this?

```python
    match actual:
        case ir.FunCall(
            fun=ir.FunCall(
                fun=ir.SymRef(id="map_"),
                args=[
                    ir.Lambda(
                        params=[_p0, _p1, _p2],
                        expr=ir.FunCall(
                            fun=ir.SymRef(id="plus"),
                            args=[
                                _s0,
                                ir.FunCall(
                                    fun=ir.SymRef(id="multiplies"),
                                    args=[_s1, _s2],
                                ),
                            ],
                        ),
                    )
                ],
            ),
            args=[ir.SymRef(id="a"), ir.SymRef(id="b"), ir.SymRef(id="c")],
        ):
            ...
        case _:
            assert False
```

src/gt4py/next/iterator/type_inference.py
```diff
@@ -475,108 +590,161 @@ def visit_Lambda(
...
+        self.constraints.add((tup, val))
+        return Val(kind=kind, dtype=elem, size=size)
+
+    def _visit_neighbors(self, node: ir.FunCall, **kwargs) -> Type:
+        if len(node.args) != 2:
+            raise TypeError("neighbors requires exactly two arguments")
+        if not isinstance(node.args[0], ir.OffsetLiteral):
+            raise TypeError("The first argument to neighbors must be an OffsetLiteral.")
+        if not isinstance(node.args[0].value, str):
+            raise TypeError("The first argument to neighbors must be an OffsetLiteral tag.")
```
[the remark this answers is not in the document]
havogt: fused

PR #1182, opened 2023-02-22
Fix typing of `cast_` builtin in ITIR type inference

tehrengruber:
The typing inference rule for `cast_` was not correctly implemented when the builtin was introduced. This PR fixes that by making `cast_` part of the grammar (similar to how this is handled in the frontend). This restriction could be lifted in the future (e.g. by introducting a type of a type).

### DIFF
--- a/src/gt4py/next/iterator/type_inference.py
+++ b/src/gt4py/next/iterator/type_inference.py
@@ -481,8 +481,6 @@ class _TypeInferrer(eve.NodeTranslator):
                 if len(node.args) != 2:
                     raise TypeError("cast_ requires exactly two arguments.")
                 val_arg_type = self.visit(node.args[0], constraints=constraints, symtypes=symtypes)
-                if not isinstance(val_arg_type, Val):
-                    raise TypeError("The first argument to `cast_` must be a value.")
                 type_arg = node.args[1]
                 if not isinstance(type_arg, ir.SymRef) or type_arg.id not in ir.TYPEBUILTINS:
                     raise TypeError("The second argument to `cast_` must be a type literal.")
--- a/src/gt4py/next/iterator/type_inference.py
+++ b/src/gt4py/next/iterator/type_inference.py
@@ -293,7 +293,6 @@ BUILTIN_TYPES: typing.Final[dict[str, Type]] = {
         ret=Val_BOOL_T1,
     ),
     "if_": FunctionType(args=Tuple.from_elems(Val_BOOL_T1, Val_T0_T1, Val_T0_T1), ret=Val_T0_T1),
-    "cast_": FunctionType(args=Tuple.from_elems(Val_T0_T1, Val_T0_T1), ret=Val_T0_T1),
     "lift": FunctionType(
         args=Tuple.from_elems(
             FunctionType(
@@ -402,6 +401,7 @@ class _TypeInferrer(eve.NodeTranslator):
             "shift",
             "cartesian_domain",
             "unstructured_domain",
+            "cast_",
         ):
             raise TypeError(
                 f"Builtin '{node.id}' is only supported as applied/called function by the type checker"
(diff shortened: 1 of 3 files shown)


tehrengruber:
@havogt Please also take a quick look at the implication of making `cast_` part of the grammar (see the test for an example).

src/gt4py/next/iterator/type_inference.py
```diff
@@ -477,6 +477,39 @@ def visit_FunCall(
                 )
                 constraints.add((tup, val))
                 return Val(kind=kind, dtype=elem, size=size)
+            if node.fun.id == "cast_":
```
havogt, line 480: The alternative would be to bring type of type into the typesystem, right?

havogt (changes requested):

src/gt4py/next/iterator/type_inference.py
tehrengruber: Yes, I discussed the idea with @egparedes a while ago (in the context of the frontend). I believe the conclusion was to wait until we really need this.

tests/next_tests/iterator_tests/test_builtins.py
```diff
@@ -347,9 +348,20 @@ def test_cast(program_processor, as_column, input_value, dtype, expected_value):
...
+            f"""
+            @fundef
+            def sten_cast(value):
+                return cast_(deref(value), {dtype})
+            """
+        ).strip(),
+        globals(),
+        locals,
+    )
+    sten_cast = locals["sten_cast"]
```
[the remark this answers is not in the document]
tehrengruber: Mh, previously the type checking tests failed and I (apparently) misinterpreted the reason to be located here. I works now, so I removed this change.

src/gt4py/next/iterator/type_inference.py
```diff
@@ -477,6 +477,39 @@ def visit_FunCall(
...
+                kind = TypeVar.fresh()
+                size = TypeVar.fresh()
+                current_loc = TypeVar.fresh()
+                defined_loc = TypeVar.fresh()
+
+                constraints.add(
+                    (
+                        val_arg_type,
+                        Val(
+                            kind=kind,
```
havogt, line 497:
```suggestion
                            kind=Value(),
```
since cast only works on values?

src/gt4py/next/iterator/type_inference.py
```diff
@@ -477,6 +477,39 @@ def visit_FunCall(
...
+                            dtype=TypeVar.fresh(),
+                            size=size,
+                            current_loc=current_loc,
+                            defined_loc=defined_loc,
```
havogt, line 501: If you agree with `kind==Value()`, then I think that should be removed as well.
```suggestion
```

src/gt4py/next/iterator/type_inference.py
```diff
@@ -477,6 +477,39 @@ def visit_FunCall(
...
+                        ),
+                    )
+                )
+
+                return Val(
+                    kind=kind,
+                    dtype=Primitive(name=type_arg.id),
+                    size=size,
+                    current_loc=current_loc,
+                    defined_loc=defined_loc,
```
havogt, line 511:
```suggestion
```

tests/next_tests/iterator_tests/test_builtins.py
```diff
@@ -13,6 +13,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import math
+import textwrap
```
havogt, line 16:
```suggestion
```

havogt (changes requested):
A few more questions

src/gt4py/next/iterator/type_inference.py
tehrengruber: Good point.

havogt (approved):
Looks good!

PR #1183, opened 2023-02-23
Subclass Dimension instead of `kind` member

havogt:
and introduce bounds to LocalDimension.

Additional: Remove `CartesianAxis` which was a leftover from the past and was alias to `Dimension`.


tehrengruber (commented):
I don't have an opinion with respect to removing `Dimension.kind`, however introducing the bounds directly in the dimension seems like a violation of the idea behind the offset provider. Namely now the dimension stores a property that is depending on the actual mesh instance. We might know this when defining the dimension, but we can't take that for granted. For example for FVM this would be a problem as ATLAS gives different values for the number of neighbors depending on mesh size (or some other "runtime" property that often changes, not sure anymore which one it was).

PR #1191, opened 2023-02-27
compiledb name: strip non-alphanumeric chars

havogt:
cmake flags might contain characters which are problematic in cmake library names, e.g. =. Remove all non-alphanumeric characters.


havogt:
#1154 into main...

PR #1193, opened 2023-03-02
Lowering test proposal

havogt:
I found the tests hard to read because the signal to noise ratio is pretty low. Not sure if people like this pattern...

This is incomplete...

### DIFF
--- a/tests/next_tests/ffront_tests/test_foast_lowering.py
+++ b/tests/next_tests/ffront_tests/test_foast_lowering.py
@@ -41,139 +41,83 @@ Cell = Dimension("Cell")
 V2EDim = Dimension("V2E", DimensionKind.LOCAL)
 V2E = FieldOffset("V2E", source=Edge, target=(Vertex, V2EDim))
 TDim = Dimension("TDim")  # Meaningless dimension, used for tests.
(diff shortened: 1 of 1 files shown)


tests/next_tests/ffront_tests/test_foast_lowering.py
```diff
@@ -41,139 +41,83 @@
 V2EDim = Dimension("V2E", DimensionKind.LOCAL)
 V2E = FieldOffset("V2E", source=Edge, target=(Vertex, V2EDim))
 TDim = Dimension("TDim")  # Meaningless dimension, used for tests.
+TOff = FieldOffset("TOff", source=TDim, target=(TDim,))
```
havogt, line 45: maybe just a function is good enough

```python
def lower(fun):
    parsed = FieldOperatorParser.apply_to_function(fun)
    lowered = FieldOperatorLowering.apply(parsed)
    return lowered.expr
```

havogt:
we defined a better strategy in the test pattern discussion:
>Short term solution: substitute current ITIR maker utilities with string templates (which can be easily inspected) and then parse them to generate the expected ITIR.

PR #1196, opened 2023-03-06
Small fixes to FOAST -> ITIR lowering

tehrengruber:
Just some fixes for minor bugs I found while working on the temporaries.
 - Fix literal dtype used for unary expressions
 - Use `maximum` and `minimum` for `max_over`, `min_over`
 - User proper initial values for `max_over`, `min_over` on other dtypes than `float64`


havogt (approved):
Looks good. I'll double-check that these changes don't get lost in my PR...

PR #1197, opened 2023-03-06
Experimental GTFN Executor Caching

tehrengruber:
A small caching mechanism that makes repeated execution of stencils with the GTFN backend much faster. We should integrate this in a better way in the OTF pipeline, but this makes live much easier for now. ~I have left the caching strategy as `SESSION` for now, but for people that know what they are doing running with `cache.Strategy.PERSISTENT` can make executing all tests a breeze (time-wise :-)).~

The caching of the GTFN executor can now be configured:
```
use_caching: bool = False
caching_strategy = cache.Strategy.SESSION
```

This feature is **experimental**!

Extracted from #965


src/gt4py/next/program_processors/runners/gtfn_cpu.py
```diff
@@ -87,22 +101,25 @@ def decorated_program(*args):
...
+                .chain(convert_args)
             )
-            .chain(convert_args)
-        )
 
-        otf_closure = stages.ProgramCall(program, args, kwargs)
+            otf_closure = stages.ProgramCall(program, args, kwargs)
 
-        compiled_runner = otf_workflow(otf_closure)
+            compiled_runner = self._cache[cache_key] = otf_workflow(otf_closure)
```
DropD, line 120:
```suggestion
            compiled_runner = self._cache.setdefault(cache_key, otf_workflow(otf_closure))
```

The whole `if` clause could be avoided (since building the workflow costs almost nothing) by using this pattern. `setdefault(key, default)` returns the value associated with `key` if `key` is in the dict, otherwise it sets it to `default` *and* returns default, eliminating the need for `if cache_key in self._cache`.

DropD (changes requested):
Since there are no tests that allow us to say with any confidence that this caching has no problem with false positives, there should be at least a way to switch it off. This should probably then be the default case on CI, if not the default case overall.

Even if it looks obvious now that it should do the right thing, there is nothing in place that ensures we would notice if it gets broken down the line.

src/gt4py/next/program_processors/runners/gtfn_cpu.py
tehrengruber: Maybe I'm missing something, but I don't see how the if else clause could be avoided. It is true that building the workflow is cheap (even though cheap, might be expensive from the perspective of running it on every stencil call), but what is not cheap is the call to the workflow `otf_workflow(otf_closure)`, which I only want to do in case the cache is not populated.

DropD:
> I fully agree, do you have a proposal on how to disable the caching without to much effort? We could make the caching an option in the `GTFNTranslator`, similar to `use_imperative_backend`, and make it default to `False`. If we also make the `cache.Strategy` a configurable option, this would fulfill all requirements with respect to caching I have currently. In the long term we should streamline the mechanism of passing backend options from the frontend, but for the time being this would be sufficient I believe.

Let's go with the options in GTFNExecutor for now. However, we seem to be accumulating them already so we shouldn't wait too long on a workflow based solution.

src/gt4py/next/program_processors/runners/gtfn_cpu.py
DropD: Right, I overlooked that the workflow would be called anyway.

src/gt4py/next/otf/workflow.py
```diff
@@ -128,3 +128,41 @@ def __call__(self, inp: StartT) -> EndT:
 
     def chain(self, step: Workflow[EndT, NewEndT]) -> CombinedStep[StartT, EndT, NewEndT]:
         return CombinedStep(first=self, second=step)
+
+
+@dataclasses.dataclass(frozen=True)
+class CachedWorkflow(Generic[StartT, EndT]):
```
DropD, line 134: Why is this not a Step subclass? It lives on the same level, has the same interoperability etc...

Either case, the current naming is `Workflow` for the concept and `XyzStep` for all the concrete implementations (Any workflow can be a step in a super-workflow anyway). So naming this Concrete implementation CachedWorkflow makes it potentially confusing.

src/gt4py/next/otf/workflow.py
```diff
@@ -128,3 +128,41 @@ def __call__(self, inp: StartT) -> EndT:
...
+
+    >>> cached_workflow(1)
+    This might take a while...
+    1
+    """
+
+    workflow: Workflow[StartT, EndT]
+    hash_function: Callable[[StartT], Any] = dataclasses.field(default=hash)
+
+    _cache: dict[Any, EndT] = dataclasses.field(repr=False, init=False, default_factory=dict)
```
DropD, line 162:
```suggestion
    hash_function: Callable[[StartT], HashT] = dataclasses.field(default=hash)

    _cache: dict[HashT, EndT] = dataclasses.field(repr=False, init=False, default_factory=dict)
```

No reason to use `Any` here, these types have to match. Just requires another unbounded TypeVar above.

src/gt4py/next/program_processors/runners/gtfn_cpu.py
```diff
@@ -58,6 +73,22 @@ def extract_connectivity_args(
...
 
+# TODO(tehrengruber): We shouldn't do this. See `GTFNExecutor.use_caching` for a comment.
+def compilation_hash(otf_closure: stages.ProgramCall) -> int:
+    """Given closure compute a hash uniquely determining if we need to recompile."""
+    offset_provider = otf_closure.kwargs["offset_provider"]
+    return hash(
+        (
+            otf_closure.program,
+            # TODO(tehrengruber): as the frontend types contain lists they are
+            #  not hashable. As a workaround we just use content_hash here.
```
DropD, line 84: If this comment only explains why `content_hash` is used here, it does not need to be a TODO. Else it should contain something like "make frontent types hashable" or at least "verify this is the best solution".

src/gt4py/next/program_processors/runners/gtfn_cpu.py
```diff
@@ -67,44 +98,50 @@ class GTFNExecutor(ppi.ProgramExecutor):
...
-            def decorated_program(*args):
-                return inp(
-                    *[convert_arg(arg) for arg in args],
-                    *extract_connectivity_args(kwargs["offset_provider"]),
-                )
+    # TODO(tehrengruber): The otf workflow currently takes the entire closure, i.e. the
+    #  fencil and its arguments, instead of the parts it actually needs to compile the program.
+    #  The hash function for the caching currently only uses a subset of the closure and implicitly
+    #  relies on the workflow to only use that information. As this is dangerous the caching is
+    #  disabled by default and should be considered experimental.
```
DropD, line 105: Again, what is the todo item here? Is it to restrict the information the workflow gets? Is it to improve the hashing? Is it to determine which is easier and then implement that? Or is it to redesign on a higher level?

DropD (changes requested):

src/gt4py/next/otf/workflow.py
tehrengruber: Done. I had add a type ignore as mypy is (in my understanding of the typing rules in python) wrongly complaining about the default `hash` function.

DropD (approved):

DropD:
looks like the doctests have gone out of sync, but once they are fixed, it's ready to merge

src/gt4py/next/otf/workflow.py
```diff
@@ -128,3 +129,41 @@ def __call__(self, inp: StartT) -> EndT:
...
+    workflow: Workflow[StartT, EndT]
+    hash_function: Callable[[StartT], HashT] = dataclasses.field(default=hash)  # type: ignore[assignment]
+
+    _cache: dict[HashT, EndT] = dataclasses.field(repr=False, init=False, default_factory=dict)
+
+    def __call__(self, inp: StartT) -> EndT:
+        hash_ = self.hash_function(inp)
+        if hash_ in self._cache:
+            return self._cache[hash_]
+        return self._cache.setdefault(hash_, self.workflow(inp))
```
tehrengruber, line 169:
```suggestion
        try:
            result = self._cache[hash_]
        except KeyError:
            result = self._cache[hash_] = self.workflow(inp)
        return result
```

PR #1199, opened 2023-03-06
Propagate foast & past typing information to ITIR

tehrengruber:
Adds support for constraining the type of symbols in ITIR (only `dtype` and `kind` right now) and propagates the information from FOAST and PAST to ITIR.


src/gt4py/next/iterator/ir.py
```diff
@@ -38,6 +39,22 @@ def __hash__(self) -> int:
 
 class Sym(Node):  # helper
     id: Coerced[SymbolName]  # noqa: A003
+    # TODO(tehrengruber): Revisit. Using strings is a workaround to avoid coupling with the
+    #   type inference.
+    kind: Optional[typing.Literal["Iterator", "Value"]] = None
```
egparedes, line 44: FYI, `None` can also be used as literal value.
```suggestion
    kind: typing.Literal["Iterator", "Value", None] = None
```

src/gt4py/next/iterator/ir.py
```diff
@@ -38,6 +39,22 @@ def __hash__(self) -> int:
...
+    def _kind_validator(self: datamodels.DataModelTP, attribute: datamodels.Attribute, value: str):
+        if value and value not in ["Iterator", "Value"]:
+            raise ValueError(f"Invalid kind `{value}`, must be one of `Iterator`, `Value`.")
+
+    @datamodels.validator("dtype")
+    def _dtype_validator(self: datamodels.DataModelTP, attribute: datamodels.Attribute, value: str):
+        if value and value not in TYPEBUILTINS:
+            raise ValueError(
+                f"Invalid dtype `{value}`, must be one of `{'`, `'.join(TYPEBUILTINS)}`."
+            )
```
egparedes, line 57: Why do we need two validators for the same field?

tests/next_tests/iterator_tests/test_type_inference.py
```diff
@@ -93,6 +93,26 @@ def test_lambda():
...
+        current_loc=ti.TypeVar(idx=1),
+        defined_loc=ti.TypeVar(idx=2),
+    )
+    expected = ti.FunctionType(
+        args=ti.Tuple.from_elems(expected_val),
+        ret=expected_val,
+    )
+    inferred = ti.infer(testee)
+    assert inferred == expected
+    assert ti.pformat(inferred) == "(It[T₁, T₂, float⁰]) → It[T₁, T₂, float⁰]"
```
egparedes, line 113: What is being tested here that hasn't been in the first `assert`?

src/gt4py/next/iterator/type_inference.py
```diff
@@ -417,7 +417,25 @@ def visit(self, node, **kwargs) -> typing.Any:
         return result
 
     def visit_Sym(self, node: ir.Sym, **kwargs) -> Type:
-        return TypeVar.fresh()
+        result = TypeVar.fresh()
+        if node.kind:
+            kind = {"Iterator": Iterator(), "Value": Value()}[node.kind]
```
egparedes, line 422: It's not very important, but I think it would better to use explicit pattern matching:
```suggestion
        match node.kind:
            case "Iterator":
                kind = Iterator()
            case "Value":
                kind = Value()
            case _:
                raise AssertionError(...)
```

egparedes (commented):
The PR looks good. Just a couple some questions, no need to change anything.

src/gt4py/next/iterator/ir.py
tehrengruber: Typo, the former was meant to validate kind not dtype. Fixed.

tests/next_tests/iterator_tests/test_type_inference.py
tehrengruber: I think the idea of @fthaler was to test the type pretty printer like this. Additionally this gives a more readable representation of the inferred type.

src/gt4py/next/iterator/type_inference.py
tehrengruber: I see the point, but I like the less verbose version more.

egparedes: Honestly, for just two cases, I don't think the dict version brings any value. Even a simple `if-else` would be better.

egparedes:
@tehrengruber What is the current state of this PR? Is it blocked by something else?

gronerl:
@egparedes you never clicked approve on this ;) we're waiting for your review.

PR #1202, opened 2023-03-07
Cartesian vs Next example

havogt:
Add an example illustrating using gt4py.cartesian and gt4py.next computations next to each other using gt4py.next storages.

Refactor GTFieldInterface and cleanup GTDimsInterface for next.


egparedes:
I would wait to merge this after storages are merged so both versions can use similar allocation code.

src/gt4py/next/common.py
```diff
@@ -599,13 +584,23 @@ def __gt_domain__(self) -> Domain:
 
 
 @extended_runtime_checkable
-class Field(NextGTDimsInterface, core_defs.GTOriginInterface, Protocol[DimsT, core_defs.ScalarT]):
+class Field(
+    core_defs.GTDimsInterface, core_defs.GTOriginInterface, Protocol[DimsT, core_defs.ScalarT]
```
havogt, line 588: add GTFieldInterface

src/gt4py/next/common.py
```diff
@@ -574,22 +574,7 @@ def __call__(self, func: fbuiltins.BuiltInFunction[_R, _P], /) -> Callable[_P, _
...
-    see :class:`~gt4py._core.definitions.GTDimsInterface` .
-    """
-
-    @property
-    def __gt_dims__(self) -> tuple[Dimension, ...]:
-        ...
-
-
-# TODO(egparedes): add support for this new protocol in the cartesian module
+# TODO(havogt): we need to specify when we should use this interface vs the `Field` protocol.
```
havogt, line 577: I assume this Protocol should be used if users define their own buffers that should be treated as fields. `__gt_domain__` is not enough, we need to specify how to get the buffer.
- Additionally: do want models of the concept to be also directly usable as field? or should this interface only be used to extract a buffer (that we could then wrap (possibly with host<->device copies) into a proper Field for embedded). Probably the latter makes sense, as the former is the `Field` interface.
- Note: forcing to add a member to a user object that returns an instance of a gt4py class, doesn't feel nice. Maybe we should at least allow `-> DomainLike` for `__gt_domain__`. Additionally, we would need `Dimension` to be any hashable object (similar to xarray) to be compatible in that sense.
- We should implement the same utility function for `__gt_domain__` that we have for `__gt_dims__`: not access the property directly but via the function, then we can provide default implementation for known types or additionally provide a single dispatch point for extending the list of default implementations.

src/gt4py/next/common.py
```diff
@@ -599,13 +584,23 @@ def __gt_domain__(self) -> Domain:
...
         ...
 
+    @property
+    def __gt_domain__(self) -> Domain:
+        return self.domain
+
+    @property
+    def __gt_dims__(self) -> tuple[str, ...]:
+        return tuple(d.value for d in self.domain.dims)
+
```
havogt, line 603: this default implementation can live in `GTFieldInterface` as it can be derived from `__gt_domain__`

src/gt4py/next/common.py
```diff
@@ -599,13 +584,23 @@ def __gt_domain__(self) -> Domain:
...
     def domain(self) -> Domain:
         ...
 
+    @property
+    def __gt_domain__(self) -> Domain:
+        return self.domain
+
+    @property
+    def __gt_dims__(self) -> tuple[str, ...]:
+        return tuple(d.value for d in self.domain.dims)
```
havogt, line 602: this make `__gt_dims__` again the same return type as for cartesian which probably makes sense, gtx should use the `__gt_domain__` property.

src/gt4py/next/common.py
havogt, line 588:
```suggestion
    core_defs.GTFieldInterface, Protocol[DimsT, core_defs.ScalarT]
```

src/gt4py/storage/cartesian/utils.py
```diff
@@ -192,6 +192,8 @@ def cpu_copy(array: Union[np.ndarray, "cp.ndarray"]) -> np.ndarray:
 def asarray(
     array: FieldLike, *, device: Literal["cpu", "gpu", None] = None
 ) -> np.ndarray | cp.ndarray:
+    if hasattr(array, "ndarray"):
```
havogt, line 195: export `__dlpack__` from the underlying field buffer.

src/gt4py/next/__init__.py
```diff
@@ -39,6 +39,11 @@
     index_field,
     np_as_located_field,
 )
+from .program_processors.runners.gtfn import (
+    run_gtfn_cached as gtfn_cpu,
+    run_gtfn_gpu_cached as gtfn_gpu,
```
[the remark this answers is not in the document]
havogt: On a second thought, I think we should keep the very explicit names, because people (at this stage at least) should know what they are using and being able to communicate to us. If we go for a default and switch it later, it might add unnecessary confusion.

src/gt4py/storage/cartesian/utils.py
havogt: Not doing it now as it is part of our Field/Array/Buffer conversation.

src/gt4py/next/common.py
```diff
@@ -574,38 +574,39 @@ def __call__(self, func: fbuiltins.BuiltInFunction[_R, _P], /) -> Callable[_P, _
...
+class GTFieldInterface(core_defs.GTDimsInterface, core_defs.GTOriginInterface, Protocol):
     """
-    Protocol for objects providing the `__gt_dims__` property, naming :class:`Field` dimensions.
+    Protocol for object providing the `__gt_domain__` property, specifying the :class:`Domain` of a :class:`Field`.
 
-    The dimension names are objects of type :class:`Dimension`, in contrast to
-    :mod:`gt4py.cartesian`, where the labels are `str` s with implied semantics,
-    see :class:`~gt4py._core.definitions.GTDimsInterface` .
+    Note:
+    - A default impolementation of the `__gt_dims__` interface from `gt4py.cartesian` is provided.
```
havogt, line 583:
```suggestion
    - A default implementation of the `__gt_dims__` interface from `gt4py.cartesian` is provided.
```

src/gt4py/next/iterator/embedded.py
```diff
@@ -680,7 +680,18 @@ def _get_axes(
         assert all(first == _get_axes(f) for f in field_or_tuple)
         return first
     else:
-        return field_or_tuple.__gt_dims__
+        return field_or_tuple.__gt_domain__.dims
+
+
+def _get_domain(
```
egparedes, line 686: Just a question: if there is now a `_get_domain(foo)`, shouldn't then `_get_axes(foo)` be implemented as an alias to   `_get_domain(foo).ndims`?

egparedes (approved):
It looks good to me. I only have a couple of questions.

PR #1205, opened 2023-03-08
Small improvements to ITIR type inference

tehrengruber:
- Add `maximum` & `minimum` builtin
- Make error messages more meaningful if a builtin is defined on IR level, but not implemented by the inference

### DIFF
--- a/src/gt4py/next/iterator/type_inference.py
+++ b/src/gt4py/next/iterator/type_inference.py
@@ -430,10 +430,20 @@ class _TypeInferrer(eve.traits.VisitorWithSymbolTableTrait, eve.NodeTranslator):
                 "cartesian_domain",
                 "unstructured_domain",
                 "cast_",
-                ir.TYPEBUILTINS,
             ):
                 raise TypeError(
-                    f"Builtin '{node.id}' is only supported as applied/called function by the type checker"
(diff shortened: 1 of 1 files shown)


src/gt4py/next/iterator/type_inference.py
```diff
@@ -418,26 +420,30 @@ def visit_Sym(self, node: ir.Sym, **kwargs) -> Type:
...
+            if node.id in BUILTIN_TYPES:
+                return freshen(BUILTIN_TYPES[node.id])
+            elif node.id in (
+                "make_tuple",
+                "tuple_get",
+                "shift",
+                "cartesian_domain",
+                "unstructured_domain",
+                "cast_",
+                ir.TYPEBUILTINS,
```
havogt, line 433: Should `TYPEBUILTINS` be here?

```
def foo():
    return int64
cast_(expr, foo())
```

tehrengruber: Cast is currently part of the grammar as we otherwise need the type of type thing to propagate the information. Hence the above does not work. The first thing when implementing such functionality would be to remove `ir.TYPEBUILTINS` here and instead return the type of type.

havogt: Let's add a comment

havogt (approved):

PR #1206, opened 2023-03-09
Fix lambda inliner if parameter is not referenced

tehrengruber:
This PR fixes a small bug in the lambda inliner such that parameters which are not referenced are correctly eliminated, i.e. now
```
(λ(x, y) → x)(x, y)
```
transforms to
```
x
```
Additionally tests for the inliner were introduced (didn't exist before).

### DIFF
--- a/tests/next_tests/iterator_tests/test_inline_lambdas.py
+++ b/tests/next_tests/iterator_tests/test_inline_lambdas.py
@@ -32,10 +32,10 @@ test_data = [
     ),
     (
         "name_collision",
-        im.call_(im.lambda__("x")(im.plus_("x", "x")))(im.plus_("x", "x")),
+        im.call_(im.lambda__("x")(im.plus_("x", "x")))(im.plus_("x", "y")),
         {
-            True: im.call_(im.lambda__("x")(im.plus_("x", "x")))(im.plus_("x", "x")),  # stays as is
-            False: im.plus_(im.plus_("x", "x"), im.plus_("x", "x")),
+            True: im.call_(im.lambda__("x")(im.plus_("x", "x")))(im.plus_("x", "y")),  # stays as is
+            False: im.plus_(im.plus_("x", "y"), im.plus_("x", "y")),
         },
     ),
     (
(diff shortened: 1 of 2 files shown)


tests/next_tests/iterator_tests/test_inline_lambdas.py
```diff
@@ -0,0 +1,58 @@
...
+            False: im.plus_(im.plus_("y", "y"), im.plus_("y", "y")),
+        },
+    ),
+    (
+        "name_collision",
+        im.call_(im.lambda__("x")(im.plus_("x", "x")))(im.plus_("x", "x")),
+        {
+            True: im.call_(im.lambda__("x")(im.plus_("x", "x")))(im.plus_("x", "x")),  # stays as is
+            False: im.plus_(im.plus_("x", "x"), im.plus_("x", "x")),
+        },
```
havogt, line 39:
```suggestion
        im.call_(im.lambda__("x")(im.plus_("x", "x")))(im.plus_("x", "y")),
        {
            True: im.call_(im.lambda__("x")(im.plus_("x", "x")))(im.plus_("x", "y")),  # stays as is
            False: im.plus_(im.plus_("x", "y"), im.plus_("x", "y")),
        },
```

tests/next_tests/iterator_tests/test_inline_lambdas.py
```diff
@@ -0,0 +1,58 @@
...
+        )(im.plus_("x", "x")),
+        im.multiplies_(im.plus_(2, 1), im.plus_("x", "x")),
+    ),
+]
+
+
+@pytest.mark.parametrize("opcount_preserving", [True, False])
+@pytest.mark.parametrize("name,testee,expected", test_data)
+def test(name, opcount_preserving, testee, expected):
+    if isinstance(expected, dict):
```
havogt, line 54: I don't like that some tests are executed twice, but I don't see a nice way to improve on that. We could skip in the else branch, but it wouldn't be very readable, too...

tehrengruber: While the result might be the same the test still tests different code paths `opcount_preserving` so they are not really executed twice.

havogt (approved):

PR #1208, opened 2023-03-14
Add proper test condition to tests in test_execution.py

gronerl:
some tests in test_execution were effectifely not testing anything, since
* np.allclose was used without assertion
* np.asarray was used, effectively asserting identity with the same memory
* in the same cases as np.asarray was used, the intended copy was based on result values so no values could have been different,  even if they were copied as intended 

Further cosmetic changes were made: (size)->(size,) (alternatively, parentheses could be dropped.)

Enabling these tests expose some failing test cases.

Besides fixing the failing tests, a review should include a second pass if the remaining tests are fine.


tests/next_tests/ffront_tests/test_execution.py
```diff
@@ -801,16 +804,16 @@ def program_domain(
...
 
-    expected = np.asarray(a_IJ_float)
-    expected[1:9, 4:6] = 2 * 2
-
     assert np.allclose(expected, a_IJ_float)
 
 
 def test_domain_tuple(fieldview_backend):
     a_IJ_float = np_as_located_field(IDim, JDim)(np.ones((size, size), dtype=float64))
     b2d_float = a_IJ_float
```
havogt, line 812: this is still weird

gronerl: Based on the function name, not what is supposed to be tested here probably, but I couldn't judge if it should be forbidden either.

tehrengruber (approved):
I fixed the failing tests, did some minor cleanup and double checked for similar errors in `test_execution.py`. Should be fine. @gronerl Please merge when you agree with my changes.

PR #1210, opened 2023-03-15
Fix mypy global_tmps 2

nfarabullini:
unignore `global_tmps.py`


havogt:
#1271 is heavily changing this file, should be done there...

PR #1211, opened 2023-03-16
Use of storage allocators in `gt4py.next` tests.

gronerl:
Adapts the `__gt_dims__` / `__gt_origin__`  interface in program executors and allows use of `gt4py.next` backends in storage allocators. 

ToDo:
- [x] test/fix cartesian GPU tests when using located field (through `array_as_located_field`)
- [x] update `quickstart.rst`
- [x] migrate `test_field.py`

### DIFF
--- a/src/gt4py/next/otf/binding/cpp_interface.py
+++ b/src/gt4py/next/otf/binding/cpp_interface.py
@@ -12,12 +12,9 @@
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-import ctypes
-import types
-from typing import Final, Sequence, Type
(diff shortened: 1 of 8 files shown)


gronerl:
Some feedback I already got from @tehrengruber is that a `name` based identification of backends / layout presets may not be enough due to backends being parametric. Unless an entirely different design is chosen, some way of hashing backends for registering/looking up layouts needs to put in place then. ProgramExecutors are however sometimes just annotated functions. How should we go about hashing in this case?

havogt:
implemented in a series of other PRs

PR #1216, opened 2023-03-21
Simplify ITIR shift collector (TraceShifts)

tehrengruber:
The `TraceShifts` pass had some bugs in particular related to literals and tuples. This PR simplifies the pass, fixes the bugs and adds some more tests.


src/gt4py/next/iterator/transforms/trace_shifts.py
```diff
@@ -117,16 +67,34 @@ def apply(arg):
...
 
 
+@dataclass(frozen=True)
+class AppliedLift:
+    stencil: Callable
+    its: List[Union[InputTracer, "AppliedLift"]]
+
+    def __post_init__(self):
+        for it in self.its:
+            assert isinstance(it, (InputTracer, AppliedLift))
```
havogt, line 77: This looks very conservative to me. What about doing this check before construction if it's really needed? Or no check, and instead assume type_inference works as a pre-condition of this pass.

src/gt4py/next/iterator/transforms/trace_shifts.py
```diff
@@ -12,15 +12,19 @@
...
 from dataclasses import dataclass
-from typing import Any, Final
+from typing import Any, Final, List, Union
 
 from gt4py.eve import NodeTranslator
 from gt4py.next.iterator import ir
 from gt4py.next.iterator.transforms.collect_shifts import ALL_NEIGHBORS
 
 
+VALUE_TOKEN = types.new_class("VALUE_TOKEN")
```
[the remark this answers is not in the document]
tehrengruber: Yes, previously the value was just `None` which was not very helpful when something went wrong. With this I can for example check if all arguments given to a function on values, i.e. `combine_` are actual values.

src/gt4py/next/iterator/transforms/trace_shifts.py
tehrengruber: Doing it in the lift looks nice, I still kept the assert both for safety and clarity.

tests/next_tests/iterator_tests/test_trace_shifts.py
```diff
@@ -120,15 +121,47 @@ def test_neighbors():
...
+def test_shifted_literal():
+    "Test shifting an applied lift of a stencil returning a constant / literal works."
+    testee = ir.StencilClosure(
+        # λ(x) → ·⟪Iₒ, 1ₒ⟫((↑(λ() → 1))())
+        stencil=im.lambda__("x")(im.deref_(im.shift_("I", 1)(im.lift_(im.lambda__()(1))()))),
         inputs=[ir.SymRef(id="inp")],
         output=ir.SymRef(id="out"),
         domain=ir.FunCall(fun=ir.SymRef(id="cartesian_domain"), args=[]),
     )
-    expected = dict()
```
tehrengruber, line 131: Note to the reviewer: This was actually wrong before. The stencil does a deref at the center point.

src/gt4py/next/iterator/transforms/trace_shifts.py
egparedes, line 25: This is just nitpicking but why is it created as a class using `types.new_class` instead of just declaring a new object? I mean, if this value is always only checked by identity as a sentinel, it is common to use an object:
```suggestion
VALUE_TOKEN = object()
```

tehrengruber: Occasionally something breaks and it is very useful to have a the information of what type of object, e.g. a `ValueToken`, directly visible in the debugger. Does that make sense?

tehrengruber:
@havogt The CI was failing for unrelated reasons. Can you approve?

havogt (approved):
Looks good to me. Let's see what Enrique comments on the last open question.

src/gt4py/next/iterator/transforms/trace_shifts.py
egparedes: Ok, fine for me. 
BTW, in case you didn't know, there is a `NOTHING` sentinel class in `eve.type_definitions` with its own metaclass to evaluate as `False` bool.

PR #1218, opened 2023-03-23
refactor[next]: Typing for bindings

havogt:
Extracted from #1130 and married with Connectivities.


src/gt4py/next/otf/binding/cpp_interface.py
```diff
@@ -29,63 +26,33 @@
...
+    elif scalar_type.kind == ts.ScalarKind.FLOAT32:
+        return "float"
+    elif scalar_type.kind == ts.ScalarKind.FLOAT64:
+        return "double"
+    elif scalar_type.kind == ts.ScalarKind.STRING:
+        return "std::string"
+    elif scalar_type.kind == ts.ScalarKind.DIMENSION:
+        raise AssertionError(f"Deprecated type '{scalar_type}' is not supported.")
+    else:
+        raise AssertionError(f"Scalar kind '{scalar_type}' is not implemented when it should be.")
```
egparedes, line 46: Using a dict wasn't such a bad idea (even if the original one had too many cases), but otherwise, I think using pattern matching makes the code more readable here:

```suggestion
    match scalar_type.kind:
        case ts.ScalarKind.BOOL:
            return "bool"
        case ts.ScalarKind.INT32:
            return "std::int32_t"
        case ts.ScalarKind.INT64:
            return "std::int64_t"
        case ts.ScalarKind.FLOAT32:
            return "float"
        case ts.ScalarKind.FLOAT64:
            return "double"
        case ts.ScalarKind.STRING:
            return "std::string"
        case ts.ScalarKind.DIMENSION:
            raise AssertionError(f"Deprecated type '{scalar_type}' is not supported.")
        case _:
            raise AssertionError(f"Scalar kind '{scalar_type}' is not implemented when it should be.")
```

src/gt4py/next/otf/binding/pybind.py
```diff
@@ -132,59 +132,39 @@ def visit_FunctionCall(self, call: FunctionCall):
...
                                   gridtools::integral_constant<int, {{dim_config}}>,\
-                                  999'999'999>({{buffer_name}})
+                                  999'999'999>({{source_buffer}})
             )"""
     )
 
     DimensionType = as_jinja("""generated::{{name}}_t""")
 
 
 def make_parameter(
```
egparedes, line 152: Not important, but is this maker function really needed? If it is only used in this module, I would say it only hides the mapping from an `interface.Parameter` to a `FunctionParameter`

tests/next_tests/otf_tests/conftest.py
```diff
@@ -18,6 +18,8 @@
 import numpy as np
 import pytest
 
+import gt4py.next.type_system.type_specifications as ts
+from gt4py.next.common import Dimension
```
egparedes, line 22: Importing symbols directly goes against our code conventions

src/gt4py/next/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -150,9 +148,7 @@ def __call__(
         )
 
         # combine into a format that is aligned with what the backend expects
-        parameters: list[
-            interface.ScalarParameter | interface.BufferParameter | interface.ConnectivityParameter
-        ] = [*regular_parameters, *connectivity_parameters]
+        parameters: list[interface.Parameter] = [*regular_parameters, *connectivity_parameters]
```
egparedes, line 151: Both arguments seem to be lists (even if there are missing return type hints) so I would do this and fix the missing return type hints:
```suggestion
        parameters: list[interface.Parameter] = regular_parameters + connectivity_parameters
```

egparedes (commented):
It looks good, just minor suggestions

src/gt4py/next/otf/binding/pybind.py
havogt: inlined

src/gt4py/next/program_processors/codegens/gtfn/gtfn_module.py
havogt: done

havogt:
tests failing because #1222

egparedes (approved):
It looks good to me. It could be merged now if it is urgent, since the test error is unrelated to this PR and it will be fixed in #1222

havogt:
cscs-ci not available this week, merging...

PR #1227, opened 2023-04-04
feature[eve]: Add functionality to preserve parts of the annex in `NodeTranslator`s

tehrengruber:
Allows preserving parts of the annex using the `PRESERVED_ANNEX_ATTRS` class variable of a `NodeTranslator`.

@egparedes I was a little surprised to not find any tests for `NodeVisitor`s or `NodeTranslator`s. Am I missing something? Otherwise we should add some in the future.

### DIFF
--- a/src/gt4py/eve/visitors.py
+++ b/src/gt4py/eve/visitors.py
@@ -165,15 +165,14 @@ class NodeTranslator(NodeVisitor):
                     if (new_child := self.visit(child, **kwargs)) is not NOTHING
                 },
             )
-            if self.PRESERVED_ANNEX_ATTRS:
-                new_annex = copy.deepcopy(new_node.annex)
-                for k in self.PRESERVED_ANNEX_ATTRS:
-                    try:
-                        assert not hasattr(new_annex, k)
-                        setattr(new_annex, k, getattr(node.annex, k))
-                    except AttributeError:
-                        pass
-                object.__setattr__(new_node, "__node_annex__", new_annex)
+            if self.PRESERVED_ANNEX_ATTRS and (old_annex := getattr(node, "__node_annex__", None)):
+                # note: access to `new_node.annex` creates the annex in the property
+                new_annex_dict = new_node.annex.__dict__
+                for key in self.PRESERVED_ANNEX_ATTRS:
+                    if value := getattr(old_annex, key, None):
+                        assert key not in new_annex_dict
+                        new_annex_dict[key] = value
+
             return new_node
 
         if isinstance(node, (list, tuple, set, collections.abc.Set)) or (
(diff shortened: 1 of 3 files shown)


src/gt4py/eve/visitors.py
```diff
@@ -162,6 +165,15 @@ def generic_visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
...
             )
+            if self.PRESERVED_ANNEX_ATTRS:
+                new_annex = copy.deepcopy(new_node.annex)
+                for k in self.PRESERVED_ANNEX_ATTRS:
+                    try:
+                        assert not hasattr(new_annex, k)
+                        setattr(new_annex, k, getattr(node.annex, k))
+                    except AttributeError:
+                        pass
+                object.__setattr__(new_node, "__node_annex__", new_annex)
```
egparedes, line 176: I don't understand why is the new annex deep-copied before being filled. If I'm not missing something, the new node has been just initialized and thus it doesn't have data in the annex. 

Another issue I see here is that the try-except pattern used to copy the data is only fast in the case where most of the nodes contains all the annex attributes to be preserved, because the `except` branch is much slower than the `try` branch. I would suggest to use a different strategy to copy the annex data, something like: 
```suggestion
            if self.PRESERVED_ANNEX_ATTRS and (old_annex := getattr(node, "__node_annex__", None)):
                new_annex_dict = new_node.annex.__dict__
                for key in self.PRESERVED_ANNEX_ATTRS:
                    if value := getattr(old_annex, key, None):
                        new_annex_dict[key] = value
```

egparedes (changes requested):
It looks good in general but I have some questions about the implementation strategy to copy the annex attributes.

src/gt4py/eve/visitors.py
tehrengruber: Agreed. I was a little sloppy... I sticked to your suggestion, just added an assert that ensures a preserved attribute is not a computed one.

src/gt4py/eve/visitors.py
```diff
@@ -162,6 +165,14 @@ def generic_visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
                     if (new_child := self.visit(child, **kwargs)) is not NOTHING
                 },
             )
+            if self.PRESERVED_ANNEX_ATTRS and (old_annex := getattr(node, "__node_annex__", None)):
+                # note: access to `new_node.annex` creates the annex in the property
```
egparedes, line 169:
```suggestion
                # note: access to `new_node.annex` implicitly creates the `__node_annex__` attribute in the property getter
```

src/gt4py/eve/visitors.py
```diff
@@ -162,6 +165,14 @@ def generic_visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
                     if (new_child := self.visit(child, **kwargs)) is not NOTHING
                 },
             )
+            if self.PRESERVED_ANNEX_ATTRS and (old_annex := getattr(node, "__node_annex__", None)):
+                # note: access to `new_node.annex` implicitly creates the `__node_annex__` attribute in the property getter
```
egparedes, line 169:
```suggestion
                # access to `new_node.annex` implicitly creates the `__node_annex__` attribute in the property getter
```

PR #1231, opened 2023-04-12
bug[next]: Fix shift / remap lowering

tehrengruber:
While working on something else I recognized a rather sneaky bug in the lowering of shifts from FOAST to ITIR. Contrary to the rest of the lowering we were not wrapping the shifts inside a lifted stencil, which can lead to bugs in user code. This is particularly devastating when executing with the gtfn backend in Release Mode, as you might just run into a segfault or incorrect results. I am actually rather surprised this did not surface earlier somewhere as something like this
```python
def composed_shift_unstructured_intermediate_result(
    inp: Field[[Vertex], float64]
) -> Field[[Cell], float64]:
    tmp = inp(E2V[0])
    return tmp(C2E[0])
```
lowers to the (incorrect) ITIR:
```
λ(inp) → ·(λ(tmp__0) → ⟪C2Eₒ, 0ₒ⟫(tmp__0))(⟪E2Vₒ, 0ₒ⟫(inp))
```
(Note how the ordering of the shifts is wrong). I had found this bug much earlier while working on (https://github.com/GridTools/gt4py/pull/965), but sadly the change went lost while merging (was very easy to overlook so @havogt and me bost missed it). What I did not recognize back then was that this bug could actually be triggered. I assumed it only occurred for expression like `field(E2V[0])(C2E[0])` which were not allowed (for unrelated reasons). For completeness this PR also adds support for such expressions. In principle it is possible to remove this feature from the PR.

### DIFF
--- a/src/gt4py/next/ffront/foast_passes/type_deduction.py
+++ b/src/gt4py/next/ffront/foast_passes/type_deduction.py
@@ -546,6 +546,29 @@ class FieldOperatorTypeDeduction(traits.VisitorWithSymbolTableTrait, NodeTransla
         arg_types = [arg.type for arg in new_args]
         kwarg_types = {name: arg.type for name, arg in new_kwargs.items()}
 
+        func_str_repr: str  # just for error handling
+        if isinstance(
+            new_func.type,
+            (ts.FunctionType, ts_ffront.FieldOperatorType, ts_ffront.ScanOperatorType),
+        ):
+            # Since we use the `id` attribute in the latter part of the toolchain ensure we
+            # have the proper format here.
+            if not isinstance(
+                new_func,
+                (foast.FunctionDefinition, foast.FieldOperator, foast.ScanOperator, foast.Name),
+            ):
+                raise FieldOperatorTypeDeductionError.from_foast_node(
+                    node, msg="Functions can only be called directly!"
+                )
+            func_str_repr = new_func.id
+        elif isinstance(new_func.type, ts.FieldType):
+            func_str_repr = str(new_func)
+        else:
+            raise FieldOperatorTypeDeductionError.from_foast_node(
+                node,
+                msg=f"Expression of type `{new_func.type}` is not callable, must be a `Function`, `FieldOperator`, `ScanOperator` or `Field`.",
+            )
+
         # ensure signature is valid
         try:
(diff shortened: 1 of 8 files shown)


src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -206,17 +206,19 @@ def _visit_shift(self, node: foast.Call, **kwargs) -> itir.Expr:
                 )
             case _:
                 raise FieldOperatorLoweringError("Unexpected shift arguments!")
-        return shift_offset(self.visit(node.func, **kwargs))
+        return im.lift_(im.lambda__("it")(im.deref_(shift_offset("it"))))(
```
tehrengruber, line 209: For the reviewer this is the core of this PR.

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -546,6 +546,29 @@ def visit_Call(self, node: foast.Call, **kwargs) -> foast.Call:
...
+            func_str_repr = str(new_func)
+        else:
+            raise FieldOperatorTypeDeductionError.from_foast_node(
+                node,
+                msg=f"Expression of type `{new_func.type}` is not callable, must be a `Function`, `FieldOperator`, `ScanOperator` or `Field`.",
+            )
+
         # ensure signature is valid
         try:
             type_info.accepts_args(
```
havogt, line 574: This snippet, including the `type_info.accepts_args` is confusing. How does that work in case of `new_func.type == FieldType`? Let's talk about it.

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -556,7 +579,7 @@ def visit_Call(self, node: foast.Call, **kwargs) -> foast.Call:
             )
         except GTTypeError as err:
             raise FieldOperatorTypeDeductionError.from_foast_node(
-                node, msg=f"Invalid argument types in call to `{node.func.id}`!"
+                node, msg=f"Invalid argument types in call to `{func_str_repr}`!"
```
havogt, line 582: Once you have implemented the change that you wanted to make to avoid func_str_repr, it should be good to go.

PR #1239, opened 2023-04-18
feature[next]: Allow manually selecting the grid type of a field operator

tehrengruber:
This PR adds support for manually selection to grid type of a field operator which was previously only supported for programs.
```python
fo = field_operator.with_grid_type(GridType.Cartesian)
```


egparedes (approved):
It looks good.

PR #1240, opened 2023-04-19
fix[next]: iterator scalar type inference

havogt:


tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -229,11 +229,6 @@ def testee(a: cases.IJKField, b: int32) -> cases.IJKField:
 
 
 def test_scalar_in_domain_spec_and_fo_call(cartesian_case):  # noqa: F811 # fixtures
```
havogt, line 231: To make this work in the C++ backend, we still have to solve the problem of where the global_parameter is created.
Currently we accept scalar (read: non-SID) arguments that are used in the domain and all arguments that are used in the stencil have to be SID. If we pass a domain size which we need in the stencil, we have to pass it twice.

PR #1241, opened 2023-04-20
refactor[next]: itir_makers -> iterator.makers

DropD:
## Description

- Changed:
  - `next.ffront.itir_makers`
    - renamed to `next.iterator.makers`
    - allow members to have the same name as iterator builtins (previously a "_" would be added to the end), since name clashes can not happen any more with the current coding guidelines.

## Requirements


tehrengruber (approved):
Looks good. I would prefer something like `ir_makers.py` as the `iterator` module contains much there then just the IR.

havogt (approved):
Looks good to me. Probably Till's name proposal makes sense to make it more obvious what the module is about.

egparedes:
It looks like this PR is ready. @DropD can you merge `main` to solve the conflicts and then merge?

PR #1242, opened 2023-04-24
Allow asynchronous building of gtscript stencils. (`gt4py.cartesian`)

gronerl:
ToDo:
- [ ] stress-test e.g. by means of use in `pace`
- [ ] update docs

Note that since `StencilObject` can not be pickled, it can't be returned from multiprocessing / result of `concurrent.future` with a ProcessPoolExecutor. In the implementation at hand, instead the `load` step is run multiple times (once in the building child process and once in the main process). If this is ok is to be seen (e.g. from `pace).

### DIFF
--- a/src/gt4py/cartesian/config.py
+++ b/src/gt4py/cartesian/config.py
@@ -13,6 +13,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import multiprocessing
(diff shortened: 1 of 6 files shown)


gronerl:
Since lazy_stencil is undocumented outside of the docstring, I just updated the docstring, without adding additional files.

PR #1245, opened 2023-04-27
[next] DaCe backend

petiaccja:
Initial implementation for the DaCe backend:
- translates iterator IR to DaCe SDFG
- incomplete but reasonable coverage on IR features
- full vertical slice: can run frontend stencils as usual


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -0,0 +1,355 @@
...
+                    name,
+                    shape=array_table[name].shape,
+                    strides=array_table[name].strides,
+                    dtype=array_table[name].dtype,
+                )
+
+        # Get output domain of the closure
+        program_arg_syms: dict[str, ValueExpr | IteratorExpr] = {}
+        for name, type_ in self.storages.items():
+            if isinstance(type_, ts.ScalarType):
```
edopao, line 179: For inputs of scalar type we are assuming that the value will be read from a symbol and stored into a container. If this is right, shouldn't this assignment take place in an entry state, before the main closure state?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -0,0 +1,355 @@
...
+            self.offset_provider,
+            index_domain,
+            input_arrays,
+            conn_arrays,
+            self.node_types,
+        )
+
+        # Map SDFG tasklet arguments to parameters
+        input_memlets = [
+            create_memlet_full(name, closure_sdfg.arrays[name]) for name in input_names
```
edopao, line 217: Suppose that the same container is used as both input and output. In this case, the entry and exit memlet should use the same subset. So we cannot mix memlet_full and memlet_at for in/out data.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -0,0 +1,355 @@
...
+
+        if input_nodes is None:
+            input_nodes = {
+                memlet.data: state.add_access(memlet.data) for name, memlet in inputs.items()
+            }
+        if output_nodes is None:
+            output_nodes = {
+                memlet.data: state.add_access(memlet.data) for name, memlet in outputs.items()
+            }
+        for name, memlet in inputs.items():
```
edopao, line 295: If inputs is empty, we still need an edge to connect the map entry node to the nested sdfg. Same applies to empty outputs and map exit node.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
gronerl: They could be, but there is no need to - the dataflow will determine the dependencies. 

I would rather argue that we don't need to make them scalars in the first place, since they can indeed be const. Let's leave it at what it is for now though.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
gronerl: fixed

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
gronerl: In fact, the nested "tasklet SDFG" will refer to the same data with different names in that case which is undefined. (it seems simplify resolves this well, but we shouldn't rely on it.)

I propose to put a temporary output a transient in the output and later copy back to the original array. Can someone confirm that this is in line with itir semantics?    Namely this is cases where the same field is used as input and output to closures? (@petiaccja ?)

gronerl: I pushed a version with the copy.

gronerl:
cscs-ci run default

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
petiaccja: Summary based on our chat:
- In ITIR semantics, it's only allowed as long as there are no shifts in the stencil
- Temporaries in the SDFG are the most sensible (if not the only) way
- DaCe should be able to optimize out the temporaries
So I guess it's fine.

edopao (dismissed):

edopao (approved):
Thank you for addressing my comments. It looks good, great job!

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -96,17 +96,25 @@ def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs) -> None:
     dace_conn_stirdes = get_stride_args(sdfg.arrays, dace_conn_args)
 
     sdfg.build_folder = cache._session_cache_dir_path / ".dacecache"
-
+    sdfg.simplify()
+    sdfg.view()
```
edopao, line 100: Need to remove the call to `view()`

PR #1236, opened 2023-04-12
test[next]: simplify boilerplate for fieldview single feature tests

DropD:
## Description

Added:
- `next_tests.feature_tests.cases`
  - helpers to express repetitive test setup concisely

Changed:
- `next_tests.feature_tests.ffront_tests.test_execution`
  - several test methods use the "cases" helpers
 
## Requirements

- [X] All fixes and/or new features come with corresponding tests.
- [x] Important design decisions have been documented in the approriate ADR inside the [docs/development/ADRs/](docs/development/ADRs/Index.md) folder.

PR #1247, opened 2023-04-28
feature[next]: Allow passing grid_type to field operator

tehrengruber:
This PR allows specifying the grid type of a field operator, i.e.

```
@field_operator(..., grid_type=GridType.UNSTRUCTURED)
def foo(...):
   ....
```

### DIFF
--- a/src/gt4py/next/ffront/decorator.py
+++ b/src/gt4py/next/ffront/decorator.py
@@ -634,6 +634,7 @@ def field_operator(
     definition=None,
     *,
     backend=None,
+    grid_type=None
 ):
     """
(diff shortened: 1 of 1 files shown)


DropD (approved):
Looks good. For reference: `Program.with_grid_type()` is 
 not in this PR because it will already be introduced in https://github.com/GridTools/gt4py/pull/1236/files#diff-d650e0c3533f991a12d42a2cd2779bc986cf8b4547a8fd966fda41f6248c37a2, right?

tehrengruber:
Exactly. Thanks!

PR #1250, opened 2023-05-04
feature[next]: Extend Single Static Assignment (SSA) pass to support if statements

tehrengruber:
This PR extends the SSA pass (operating on the Python AST) to support if statements such that every variable is assigned only once _per branch_, e.g. this
```python
if True:
    a = 1
else:
    a = 2
    if True:
        a = 3
    a = 4
a = 5
```
is transformed into
```python
if True:
    a__0 = 1
    a__2 = a__0
else:
    a__0 = 2
    if True:
        a__1 = 3
    else:
        a__1 = a__0
    a__2 = 4
a__3 = 5
```
Most of the code was written by @BenWeber42 reviewed by me. See https://github.com/GridTools/gt4py/pull/972 for the discussion. The PR is a prerequisite for https://github.com/GridTools/gt4py/pull/1079 and was spun off from there to ease review.

### DIFF
--- a/src/gt4py/next/ffront/ast_passes/single_static_assign.py
+++ b/src/gt4py/next/ffront/ast_passes/single_static_assign.py
@@ -14,6 +14,7 @@
 from __future__ import annotations
 
 import ast
+import collections.abc
 import dataclasses
 import typing
 
@@ -22,13 +23,15 @@ from gt4py.next.ffront.fbuiltins import TYPE_BUILTIN_NAMES
 
 ASTNodeT = typing.TypeVar("ASTNodeT", bound=ast.AST)
 
-_UNIQUE_NAME_SEPERATOR = "ᐞ"
+_UNIQUE_NAME_SEPARATOR = "ᐞ"
 
 
-def unique_name(name: str, num_assignments: int):
+# TODO(tehrengruber): use a global mechanism to get uniqueness across multiple passes (not only
+#  the ssa pass).
+def unique_name(name: str, num_assignments: int) -> str:
     """Given the name of a variable return a unique name after the given amount of assignments."""
     if num_assignments >= 0:
-        return f"{name}{_UNIQUE_NAME_SEPERATOR}{num_assignments}"
+        return f"{name}{_UNIQUE_NAME_SEPARATOR}{num_assignments}"
     assert num_assignments == -1
     return name
 
@@ -56,29 +59,31 @@ def _is_guaranteed_to_return(node: ast.stmt | list[ast.stmt]) -> bool:
 class _AssignmentTracker:
     """Helper class to keep track of the number of assignments to a variable."""
 
+    #: mapping from a variable name to the number of times it is assigned to. `-1` signifies it to
+    #:  be defined only, but not assigned to.
     _counts: dict[str, int] = dataclasses.field(default_factory=dict)
 
     def define(self, name: str) -> None:
         if name in self.names():
             raise ValueError(f"Variable {name} is already defined.")
+        # -1 signifies a
         self._counts[name] = -1
 
     def assign(self, name: str) -> None:
         self._counts[name] = self.count(name) + 1
 
-    def count(self, name: str):
+    def count(self, name: str) -> int:
         return self._counts.get(name, -1)
 
-    def names(self):
+    def names(self) -> collections.abc.Set:
         return self._counts.keys()
 
-    def copy(self):
+    def copy(self) -> _AssignmentTracker:
         return _AssignmentTracker({**self._counts})
 
 
 def _merge_assignment_tracker(a: _AssignmentTracker, b: _AssignmentTracker) -> _AssignmentTracker:
-    common_names = set(a.names()) & set(b.names())
-    return _AssignmentTracker({k: max(a.count(k), b.count(k)) for k in common_names})
+    return _AssignmentTracker({k: max(a.count(k), b.count(k)) for k in (a.names() & b.names())})
 
 
 @dataclasses.dataclass
@@ -108,10 +113,10 @@ class SingleStaticAssignPass(ast.NodeTransformer):
     ...     )
     ... ))
     def foo():
-        a__0 = 1
-        a__1 = 2 + a__0
-        a__2 = 3 + a__1
-        return a__2
+        aᐞ0 = 1
+        aᐞ1 = 2 + aᐞ0
+        aᐞ2 = 3 + aᐞ1
+        return aᐞ2
 
     Note that each variable name is assigned only once (per branch) and never updated / overwritten.
 
@@ -143,27 +148,27 @@ class SingleStaticAssignPass(ast.NodeTransformer):
         # For practical purposes, this is sufficient, but really not general at all.
         # However, the algorithm was never intended to be general.
 
-        old_versioning = self.assignment_tracker.copy()
+        prev_assignment_tracker = self.assignment_tracker.copy()
 
         for arg in node.args.args:
             self.assignment_tracker.define(arg.arg)
 
         node.body = [self.visit(stmt) for stmt in node.body]
 
-        self.assignment_tracker = old_versioning
+        self.assignment_tracker = prev_assignment_tracker
         return node
 
     def visit_If(self, node: ast.If) -> ast.If:
-        old_versioning = self.assignment_tracker
+        prev_assignment_tracker = self.assignment_tracker
 
         node.test = self._rename(node.test)
 
-        self.assignment_tracker = old_versioning.copy()
+        self.assignment_tracker = prev_assignment_tracker.copy()
         node.body = [self.visit(el) for el in node.body]
         body_assignment_tracker = self.assignment_tracker
         body_returns = _is_guaranteed_to_return(node.body)
 
-        self.assignment_tracker = old_versioning.copy()
+        self.assignment_tracker = prev_assignment_tracker.copy()
         node.orelse = [self.visit(el) for el in node.orelse]
         orelse_assignment_tracker = self.assignment_tracker
         orelse_returns = _is_guaranteed_to_return(node.orelse)
@@ -225,13 +230,12 @@ class SingleStaticAssignPass(ast.NodeTransformer):
             node.value = self._rename(node.value)
             node.target = self.visit(node.target)
         elif isinstance(node.target, ast.Name):
-            # An empty annotation always applies to the next assignment.
-            # So we need to use the correct versioning, but also ensure
-            # we restore the old versioning afterwards, because no assignment
-            # actually happens.
-            old_versioning = self.assignment_tracker.copy()
+            # An annotated assignment without a value always applies to the next assignment. Ensure
+            # the two have the same name by keeping the assignment tracker as is (after this
+            # statement).
+            prev_assignment_tracker = self.assignment_tracker.copy()
             node.target = self.visit(node.target)
-            self.assignment_tracker = old_versioning
+            self.assignment_tracker = prev_assignment_tracker
         return node
 
     def visit_Name(self, node: ast.Name) -> ast.Name:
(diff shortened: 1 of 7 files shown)


src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -82,11 +179,78 @@ def apply(cls, node: ast.AST) -> ast.AST:
...
+        for arg in node.args.args:
+            self.versioning.define(arg.arg)
+
+        node.body = [self.visit(stmt) for stmt in node.body]
+
+        self.versioning = old_versioning
+        return node
+
+    def visit_If(self, node: ast.If) -> ast.If:
+        old_versioning = self.versioning
```
egparedes, line 203: If understand this correctly, the following `self._rename()` call cannot change the versioning dict, so why is this copy needed here?

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -11,12 +11,105 @@
...
+def _make_assign(target: str, source: str, location_node: ast.AST):
+    result = ast.Assign(
+        targets=[ast.Name(ctx=ast.Store(), id=target)], value=ast.Name(ctx=ast.Load(), id=source)
+    )
+    for node in ast.walk(result):
+        ast.copy_location(node, location_node)
+    return result
+
+
+def is_guaranteed_to_return(node: ast.stmt | list[ast.stmt]) -> bool:
```
egparedes, line 31: Missing docstring

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -82,11 +179,78 @@ def apply(cls, node: ast.AST) -> ast.AST:
...
         super().__init__()
-        self.name_counter: dict[str, int] = {}
-        self.separator: str = separator
+        self.versioning: Versioning = Versioning()
+        self.name_encoder: NameEncoder = NameEncoder(separator)
+
+    def _rename(self, node: ast.AST):
+        return self.RhsRenamer.apply(self.versioning, self.name_encoder, node)
+
+    def visit_FunctionDef(self, node: ast.FunctionDef):
```
egparedes, line 188: Missing return type annotation

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -82,11 +179,78 @@ def apply(cls, node: ast.AST) -> ast.AST:
 
     def __init__(self, separator="__"):
         super().__init__()
-        self.name_counter: dict[str, int] = {}
-        self.separator: str = separator
+        self.versioning: Versioning = Versioning()
+        self.name_encoder: NameEncoder = NameEncoder(separator)
+
+    def _rename(self, node: ast.AST):
```
egparedes, line 185: Missing return type annotation

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -82,11 +179,78 @@ def apply(cls, node: ast.AST) -> ast.AST:
 
     def __init__(self, separator="__"):
```
egparedes, line 180: Some type annotations are missing. Additionally, I don't like the default separator since it may easily collide with user written code. I would suggest to use a very uncommon unicode symbol as separator to avoid problems (see below for further information about accepted unicode chars in python)
```suggestion
    def __init__(self, separator: str="ᐞ") -> None:
```
This would produce: `nameᐞ2`
```
Unicode symbols per class: https://www.compart.com/en/unicode/category
Python unicode symbols for identifiers:
    - xid_start  =  all characters in id_start after NFKC normalization, that is general categories Lu, Ll, Lt, Lm, Lo, Nl, and the underscore
    - xid_continue  =  all characters in id_continue after NFKC normalization, that is all chars in id_start plus characters in the categories Mn, Mc, Nd, Pc 

https://www.compart.com/en/unicode/U+10307: 𐌇
https://www.compart.com/en/unicode/U+10318: 𐌘
https://www.compart.com/en/unicode/U+A788: ꞈ
https://www.compart.com/en/unicode/U+141E: ᐞ
https://www.compart.com/en/unicode/U+1421: ᐡ
https://www.compart.com/en/unicode/U+1506: ᔆ
https://www.compart.com/en/unicode/U+15B0: ᖰ
https://www.compart.com/en/unicode/U+15B1: ᖱ
https://www.compart.com/en/unicode/U+15B2: ᖲ
https://www.compart.com/en/unicode/U+15B3: ᖳ
https://www.compart.com/en/unicode/U+1601: ᘁ
https://www.compart.com/en/unicode/U+166E: ᙮
https://www.compart.com/en/unicode/U+167F: ᙿ
```

egparedes (changes requested):
The algorithm look good but the implementation looks too verbose to me and unnecessarily hard to read. I recommend to simplify the code before approving the Pr.

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -11,12 +11,105 @@
...
+
+class NameEncoder:
+    """Helper class to encode names of versioned variables."""
+
+    _separator: str
+
+    def __init__(self, separator: str):
+        self._separator = separator
+
+    def encode_name(self, name: str, versions: Versioning) -> str:
```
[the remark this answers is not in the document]
tehrengruber: I removed the class altogether.

src/gt4py/next/ffront/ast_passes/single_static_assign.py
tehrengruber: Private now.

src/gt4py/next/ffront/ast_passes/single_static_assign.py
tehrengruber: This is true, but you need `old_versioning` again for the _else_-branch after `self.versioning` has been changed for the _if_-branch.

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -104,18 +268,19 @@ def visit_AnnAssign(self, node: ast.AnnAssign) -> ast.AnnAssign:
             node.value = self._rename(node.value)
             node.target = self.visit(node.target)
         elif isinstance(node.target, ast.Name):
-            target_id = node.target.id
+            # An empty annotation always applies to the next assignment.
+            # So we need to use the correct versioning, but also ensure
+            # we restore the old versioning afterwards, because no assignment
+            # actually happens.
```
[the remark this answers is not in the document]
tehrengruber: I don't fully understand the details, but this is referring to this case (test_empty_annotated_assign):
```
a = 0
a: int
b = a
```
If you want I take a closer look.

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -11,12 +11,105 @@
...
+    def __getitem__(self, name: str) -> None | int:
+        return self._versions[name]
+
+    def __iter__(self) -> typing.Iterator[tuple[str, None | int]]:
+        return iter(self._versions.items())
+
+    def copy(self) -> Versioning:
+        copy = Versioning()
+        copy._versions = {**self._versions}
+        return copy
```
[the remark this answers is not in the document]
tehrengruber: I simplified the class, but kept it as it makes the code where it is used much easier to comprehend.

tehrengruber:
Ready for another round.

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -11,12 +11,77 @@
...
 import ast
+import dataclasses
+import typing
 
 from gt4py.next.ffront.fbuiltins import TYPE_BUILTIN_NAMES
 
 
+ASTNodeT = typing.TypeVar("ASTNodeT", bound=ast.AST)
+
+_UNIQUE_NAME_SEPERATOR = "ᐞ"
```
egparedes, line 25:
```suggestion
_UNIQUE_NAME_SEPARATOR = "ᐞ"
```

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -11,12 +11,77 @@
...
+    if isinstance(node, ast.Return):
+        return True
+    if isinstance(node, ast.If):
+        return _is_guaranteed_to_return(node.body) and _is_guaranteed_to_return(node.orelse)
+    return False
+
+
+@dataclasses.dataclass
+class _AssignmentTracker:
+    """Helper class to keep track of the number of assignments to a variable."""
```
egparedes, line 57: Please, document the `-1` case here.

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -11,12 +11,77 @@
...
 
 from gt4py.next.ffront.fbuiltins import TYPE_BUILTIN_NAMES
 
 
+ASTNodeT = typing.TypeVar("ASTNodeT", bound=ast.AST)
+
+_UNIQUE_NAME_SEPERATOR = "ᐞ"
+
+
+def unique_name(name: str, num_assignments: int):
```
egparedes, line 28: Missing return type annotation. Additionally, is this the right place to add this function? It is used in many different tests but I understand it might be out of the scope of this PR, so I suggest to add at least a TODO mentioning it.

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -11,12 +11,77 @@
...
+        self._counts[name] = self.count(name) + 1
+
+    def count(self, name: str):
+        return self._counts.get(name, -1)
+
+    def names(self):
+        return self._counts.keys()
+
+    def copy(self):
+        return _AssignmentTracker({**self._counts})
```
egparedes, line 76: Missing return type annotations.

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -11,12 +11,77 @@
...
+
+
+def _merge_assignment_tracker(a: _AssignmentTracker, b: _AssignmentTracker) -> _AssignmentTracker:
+    common_names = set(a.names()) & set(b.names())
```
egparedes, line 80: The conversion to `set` is not needed since `dict_keys` is already a `collection.abc.Set`:
```suggestion
    common_names = a.names() & b.names()
```

src/gt4py/next/ffront/ast_passes/single_static_assign.py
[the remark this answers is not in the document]
egparedes: Yes, I understood the same core idea but don't really get the details. I think we should understand what the comment really means and rephrase it in a more comprehensible way before merging the PR, otherwise is useless.

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -48,45 +113,101 @@ def foo():
...
-        def __init__(self, name_counter, separator):
-            super().__init__()
-            self.name_counter: dict[str, int] = name_counter
-            self.separator: str = separator
+        node.test = self._rename(node.test)
 
-        def visit_Name(self, node: ast.Name) -> ast.Name:
-            if node.id in self.name_counter:
-                node.id = f"{node.id}{self.separator}{self.name_counter[node.id]}"
+        self.assignment_tracker = old_versioning.copy()
```
egparedes, line 161: Wouldn't it be simpler to remove line 159 and here just do this?
```suggestion
        old_versioning = self.assignment_tracker.copy()
```

src/gt4py/next/ffront/ast_passes/single_static_assign.py
```diff
@@ -48,45 +113,101 @@ def foo():
...
+        node.body = [self.visit(stmt) for stmt in node.body]
+
+        self.assignment_tracker = old_versioning
+        return node
 
-        @classmethod
-        def apply(cls, name_counter, separator, node):
-            return cls(name_counter, separator).visit(node)
+    def visit_If(self, node: ast.If) -> ast.If:
+        old_versioning = self.assignment_tracker
```
egparedes, line 157: See my comment below
```suggestion
```

egparedes (changes requested):
It looks almost ready to go. Just some comments about missing type annotations and docstrings.

PR #1254, opened 2023-05-22
feature[next]: Cache direct field operator call (`as_program`)

tehrengruber:
Cache direct calls to field operators by storing the autogenerated programs in a cache.


src/gt4py/next/ffront/decorator.py
```diff
@@ -607,6 +608,13 @@ def as_program(
         #  of arg and kwarg types
         # TODO(tehrengruber): check foast operator has no out argument that clashes
         #  with the out argument of the program we generate here.
+        hash_ = content_hash(
+            (tuple(arg_types), tuple((name, arg) for name, arg in kwarg_types.items()))
+        )
+        try:
+            return self._program_cache[hash_]
+        except KeyError:
+            pass
```
DropD, line 617: What is the advantage of try-except here over checking for presence of the hash? 

```python
if hash_ in self._program_cache:
    return self._program_cache[hash_]
```

or if that is too repetitive

```python
if (cached_program := self._program_cache.get(hash_, None)) is not None:
    return cached_program
```

DropD (changes requested):

src/gt4py/next/ffront/decorator.py
tehrengruber: Performance. I'm also not particularly found of how this looks like and would prefer your first proposal, but the try except version is faster as it only requires a single look-up. I quickly benchmarked the second version, but it doesn't perform nearly as well.

```
def test1():
   try:
     return d[7582283]
   except KeyError:
     pass

def test2():
    if (entry := d.get(7582283, None)):
        return entry

>>> timeit.timeit("test1()", setup="import random; d = {7582283: 5, **{hash(random.randint(0, 10000000)): 1 for _ in range(1000)}}; from __main__ import test1", number=100000000)
5.147236090999286
>>> timeit.timeit("test2()", setup="import random; d = {7582283: 5, **{hash(random.randint(0, 10000000)): 1 for _ in range(1000)}}; from __main__ import test2", number=100000000)
7.2443089640000835
```

DropD: Interesting. May be worth a comment.

DropD (approved):

PR #1255, opened 2023-05-22
feature[next]: Remove int type from FOAST, PAST, ITIR

tehrengruber:
This PR removes all occurrences of the `int`, `float` type and replaces them by the fixed size types `int32`, `int64`, `float32`, `float64`. In particular `int` was problematic as an undetermined size lead to problems in type inference giving strange errors akin of `Can not satisfy constraint on primitive types: int ≡ int32`.

Blocked by #1258 (already merged in this PR).


tests/next_tests/integration_tests/feature_tests/cases.py
```diff
@@ -47,11 +47,10 @@
 
 
 # mypy does not accept [IDim, ...] as a type
-IField: TypeAlias = common.Field[[IDim], np.int64]  # type: ignore [valid-type]
-IJKField: TypeAlias = common.Field[[IDim, JDim, KDim], np.int64]  # type: ignore [valid-type]
-IJKFloatField: TypeAlias = common.Field[[IDim, JDim, KDim], np.float64]  # type: ignore [valid-type]
-VField: TypeAlias = common.Field[[Vertex], np.int64]  # type: ignore [valid-type]
-EField: TypeAlias = common.Field[[Edge], np.int64]  # type: ignore [valid-type]
+IField: TypeAlias = common.Field[[IDim], np.float64]  # type: ignore [valid-type]
```
havogt, line 50: I would keep an `int` type as indicated in the other PR, but not very strong opinion.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -771,10 +772,10 @@ def fieldop_domain(a: Field[[IDim, JDim], float64]) -> Field[[IDim, JDim], float
     def program_domain(
         inp: Field[[IDim, JDim], float64],
         out: Field[[IDim, JDim], float64],
-        lower_i: int64,
-        upper_i: int64,
-        lower_j: int64,
-        upper_j: int64,
+        lower_i: int32,
```
havogt, line 775: is this a strong requirement now? if yes, we should introduce an alias or maybe even a distinct type (at the frontend level).

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -865,7 +866,7 @@ def fieldop_where_k_offset(
         a: Field[[IDim, KDim], float64],
         k_index: Field[[KDim], int64],
     ) -> Field[[IDim, KDim], float64]:
-        return where(k_index > 0, a(Koff[-1]), 2.0)
+        return where(k_index > int64(0), a(Koff[-1]), 2.0)
```
havogt, line 869:
```suggestion
        return where(k_index > int32(0), a(Koff[-1]), 2.0)
```
this is an index type

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_gt4py_builtins.py
```diff
@@ -141,7 +141,7 @@ def test_reduction_expression(reduction_setup, fieldview_backend):
     def reduce_expr(edge_f: Field[[Edge], int64]) -> Field[[Vertex], int64]:
         tmp_nbh_tup = edge_f(V2E), edge_f(V2E)
         tmp_nbh = tmp_nbh_tup[0]
-        return 3 * neighbor_sum(-edge_f(V2E) * tmp_nbh * 2, axis=V2EDim)
+        return int64(3) * neighbor_sum(-edge_f(V2E) * tmp_nbh * int64(2), axis=V2EDim)
```
havogt, line 144: hmm, this is ugly. we should allow upcasting int32...

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_scalar_if.py
```diff
@@ -85,10 +85,10 @@ def simple_if(
     ) -> cases.IField:
         if condition1:
             result1 = a
-            result2 = a + 1
+            result2 = a + 1.0
```
havogt, line 88: see comment about default

havogt (changes requested):
Main concern is the duplication of value to literal translation and hiding it in ir_makers.

tehrengruber:
@havogt Mostly as discussed. I have introduced a new function `literal_from_value` in `ir_makers` and left `literal` as is. Since `literal_from_value` does not itself introduce any new logic it made sense to me to leave it in `ir_makers` which contains other functions with similar tasks (i.e. `ensure_expr`).

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -29,6 +30,12 @@
 TYPE_BUILTINS = [Field, Dimension, int32, int64, float32, float64] + PYTHON_TYPE_BUILTINS
 TYPE_BUILTIN_NAMES = [t.__name__ for t in TYPE_BUILTINS]
 
+# Be aware: Type aliases are not fully supported in the frontend yet, e.g. `IndexType(1)` will not
+# work.
+IndexType: TypeAlias = int32
```
havogt, line 35: What was the problem why we decided to add an index type? I only realize now that this is a potential problem for bindings, if the driver code has an opinion on the size of `IndexType`. But let's keep it for now...

havogt (approved):
just a question, because I don't remember.

src/gt4py/next/ffront/fbuiltins.py
tehrengruber: You suggested it as it is occurring in the tests multiple times (e.g. in IndexFields). I think the bindings will still accept `int64` for NeighborTableOffsetProvider.

havogt: I suggested the alias but why did we introduce this constraint https://github.com/GridTools/gt4py/pull/1255/files#diff-1c4cbeb95e161c4e1ca0327817608ba3a8b48569e756847b2093bc43606863d6R685

PR #1258, opened 2023-05-30
refactor[next]: Change default datatype for tests to float

tehrengruber:
Currently most tests use `int64` as the default dtype for fields:
```
IField: TypeAlias = common.Field[[IDim], np.int64]
```
This PR changes this to `float` for 2 reasons:
- Most of our use cases are with fields of floats.
- Using int literals, e.g. `1`, `2` will break after https://github.com/GridTools/gt4py/pull/1255 has been merged as they will be `int32` instead of `int64`.

### DIFF
--- a/tests/next_tests/integration_tests/feature_tests/cases.py
+++ b/tests/next_tests/integration_tests/feature_tests/cases.py
@@ -50,8 +50,8 @@ from next_tests.integration_tests.feature_tests.ffront_tests.ffront_test_utils i
 IField: TypeAlias = common.Field[[IDim], np.float64]  # type: ignore [valid-type]
 IJKField: TypeAlias = common.Field[[IDim, JDim, KDim], np.float64]  # type: ignore [valid-type]
 IJKIntField: TypeAlias = common.Field[[IDim, JDim, KDim], np.int64]  # type: ignore [valid-type]
-VField: TypeAlias = common.Field[[Vertex], np.int64]  # type: ignore [valid-type]
-EField: TypeAlias = common.Field[[Edge], np.int64]  # type: ignore [valid-type]
+VField: TypeAlias = common.Field[[Vertex], np.float64]  # type: ignore [valid-type]
+EField: TypeAlias = common.Field[[Edge], np.float64]  # type: ignore [valid-type]
 
 # TODO(ricoh): unify the following with the `ffront_test_utils.reduction_setup`
 #   fixture if `ffront_test_utils.reduction_setup` is not completely superseded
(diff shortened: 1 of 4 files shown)


tehrengruber:
@havogt Rico mentioned you suggested to use `int` when he refactored the tests. Please take a short look on the approach.

DropD (approved):

havogt:
My reason for int is that you can do equality comparison, therefore if dtype doesn't matter for the test, I would argue it's better.

DropD:
> My reason for int is that you can do equality comparison, therefore if dtype doesn't matter for the test, I would argue it's better.

In that case we should change the default comparison in `cases.verify` to `np.equal`. It is currently `np.allclose`.

havogt:
> > My reason for int is that you can do equality comparison, therefore if dtype doesn't matter for the test, I would argue it's better.
> 
> In that case we should change the default comparison in `cases.verify` to `np.equal`. It is currently `np.allclose`.

We could, but probably it doesn't matter, I'd expect np.allclose == np.equal for all ints (but I'd have to do the math). Then np.allclose would be better as it works for all types.

havogt:
btw. for the second problem mentioned in the description: no problem changing the default to int32

havogt:
>We could, but probably it doesn't matter, I'd expect np.allclose == np.equal for all ints (but I'd have to do the math). Then np.allclose would be better as it works for all types.

Ok, that doesn't make sense... We should use equal then...

egparedes:
> > We could, but probably it doesn't matter, I'd expect np.allclose == np.equal for all ints (but I'd have to do the math). Then np.allclose would be better as it works for all types.
> 
> Ok, that doesn't make sense... We should use equal then...

Why doesn't it make sense? When using `np.allclose()` with ints, it will convert the values to floats first and then do the comparison, so if two int values are equal before the conversion, it will always work , right?

egparedes:
> >>> a = np.asarray([1e10], dtype=np.int64)
> >>> b = np.asarray([1e10+1], dtype=np.int64)
> >>> np.allclose(a,b)
> True

I think I misunderstood the point. I agree it doesn't make sense to use `int` as default type, what I meant is that it should be ok to use `np.allclose()` for comparisons even with `int` arrays

havogt:
> > > > > a = np.asarray([1e10], dtype=np.int64)
> > > > > b = np.asarray([1e10+1], dtype=np.int64)
> > > > > np.allclose(a,b)
> > > > > True
> 
> I think I misunderstood the point. I agree it doesn't make sense to use `int` as default type, what I meant is that it should be ok to use `np.allclose()` for comparisons even with `int` arrays

Now I am confused, I still think it makes sense to use `int` as default type and then do equality comparison, because it's a bit stricter than float with allclose (in the default case where we explicitly don't care about the values and the dtype).

tehrengruber:
Using `int` to allow for equality comparison makes sense. I switched to using `int32` as the default type in the tests. The change only works properly together with https://github.com/GridTools/gt4py/pull/1255 so closing here. Changing the comparison function is outside of the scope of this PR. I have opend an issue here: https://github.com/GridTools/gt4py/issues/1269

PR #1259, opened 2023-05-30
feature[next]: Save type to annex in ITIR type inference

tehrengruber:
This PR allows saving the results of type inference in the ITIR nodes. This is particularly useful for passes that modify the tree, but need type information during the process (in my case this is for the temporary extraction pass).

```python
testee = im.lambda_("a")(im.plus("a", im.literal("1", "float32")))
ti.infer(testee, save_to_annex=True)
param_type = testee.params[0].annex.type
assert isinstance(param_type, ti.Val) and param_type.dtype.name == "float32"
```

PR #1260, opened 2023-05-30
feature[next]: Use global type information in CollapseTuple pass

tehrengruber:
Previously the `CollapseTuple` Pass ran type inference only on the local expression it was currently considering. This PR uses the global information that works for more cases.

Blocked by ~#1259~, ~#1255~


src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -80,7 +86,9 @@ def visit_FunCall(self, node: ir.FunCall, **kwargs) -> ir.Node:
                     # tuple argument differs, just continue with the rest of the tree
                     return self.generic_visit(node)
 
-            if self.ignore_tuple_size or _get_tuple_size(first_expr) == len(node.args):
+            if self.ignore_tuple_size or _get_tuple_size(self._node_types[id(first_expr)]) == len(
```
havogt, line 89: I don't think that I want `self._node_types[id(first_expr)]` to become a pattern. Any thoughts?

src/gt4py/next/iterator/type_inference.py
```diff
@@ -902,15 +896,34 @@ def visit_FencilDefinition(
...
+                    f"Expected a type to be inferred for node `{child_node}`, but none was found."
+                )
+
+
 def infer_all(
     node: ir.Node,
+    *,
     offset_provider: Optional[dict[str, Connectivity | Dimension]] = None,
     reindex: bool = True,
+    save_to_annex=False,
```
havogt, line 918: Why is this added? It's not used in this PR.

havogt (changes requested):
Some questions

src/gt4py/next/iterator/type_inference.py
havogt: Ok, it's in the other PR and reviewed separately.

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -56,8 +57,13 @@ def apply(
         If `ignore_tuple_size`, apply the transformation even if length of the inner tuple
         is greater than the length of the outer tuple.
         """
+        node_types = it_type_inference.infer_all(node)
```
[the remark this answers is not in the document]
tehrengruber: I don't know, but probably it is terrible :-(

src/gt4py/next/iterator/transforms/collapse_tuple.py
havogt: When you say annex in this context, you mean we guarantee that the type is always updated/preserved in all passes, right? Otherwise we would also have to run it before each pass.
Independent of the performance problem, is there any disadvantage in using a pass-local annex here?

tehrengruber: With respect to the annex being "dangerous" yes. If it's only pass local you can still screw things up, but it is easier to find the problem (i.e. you do a transformation of the IR and some of the newly created nodes don't have a type or the type changed) as it is local to the pass.

tehrengruber: See also this commit on how the code looks like if we use the annex: https://github.com/GridTools/gt4py/pull/1260/commits/a09f18f1d227c9a557a6930b7aa17fd897d08294

havogt (approved):
As discussed, we leave it as is for now.

PR #1268, opened 2023-06-12
Test refactoring 1

nfarabullini:
refactoring of `test_gt4py_builtins.py`, `test_math_unary_builtins.py`, and `test_external_local_field.py`

### DIFF
--- a/tests/next_tests/integration_tests/feature_tests/cases.py
+++ b/tests/next_tests/integration_tests/feature_tests/cases.py
@@ -174,11 +174,9 @@ class UniqueInitializer(DataInitializer):
         svals = tuple(sizes.values())
         n_data = int(np.prod(svals))
         self.start += n_data
-        if dtype == "bool":
-            f = np.random.choice(a=[False, True], size=(svals))
-        else:
-            f = np.arange(start, start + n_data, dtype=dtype).reshape(svals)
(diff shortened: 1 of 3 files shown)


tests/next_tests/integration_tests/feature_tests/cases.py
```diff
@@ -173,9 +174,11 @@ def field(
         svals = tuple(sizes.values())
         n_data = int(np.prod(svals))
         self.start += n_data
-        return gtx.np_as_located_field(*sizes.keys())(
-            np.arange(start, start + n_data, dtype=dtype).reshape(svals)
-        )
+        if dtype == "bool":
+            f = np.random.choice(a=[False, True], size=(svals))
+        else:
```
DropD, line 179: I am not sure if UniqueInitializer should work for bools, because there is no way to make each item unique. This should probably be in a separate initializer class (I think it's ok to have to pass it manually when initializing a bool field in tests). UniqueInitializer should probably just throw an error saying that it can't do bools.

DropD (changes requested):

DropD (approved):
A way to preserve compatibility with `verify_with_default_data` would be to have `allocate` pick a special default strategy for `bool` fields. But this works for now.

PR #1270, opened 2023-06-14
refactor[next]: Frontend test refactoring part 2

nfarabullini:
refactoring of `test_program.py`and `test_math_builtin_execution.py`


tests/next_tests/integration_tests/feature_tests/ffront_tests/test_program.py
```diff
@@ -36,205 +40,260 @@
...
+        prog,
+        a,
+        b,
+        c,
+        out_a,
+        out_b,
+        out_c,
+        inout=((out_a.array(), out_b.array()), out_c.array()),
+        ref=((a.array(), b.array()), c.array()),
+        comparison=lambda a, b: (np.all(a == out_a), np.all(b == out_b)),
```
DropD, line 286: The comparison must return a single boolean value. A non-empty container of bool values is itself always `True`-like in Python (this is also the reason why `numpy.all()` exists in the first place). Examples:

```python
In [3]: bool((False, False))
Out[3]: True
```

```python
[ins] In [2]: assert (False, False)
<>:1: SyntaxWarning: assertion is always true, perhaps remove parentheses?
<ipython-input-2-f321c071ff0c>:1: SyntaxWarning: assertion is always true, perhaps remove parentheses?
  assert (False, False)
```

DropD: Maybe to avoid accidentally always-true tests, `verify` should assert that the result of the comparison is a `bool` instance?

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_program.py
```diff
@@ -36,205 +40,260 @@
...
+    cases.verify(
+        cartesian_case,
+        prog,
+        a,
+        b,
+        out_a,
+        out_b,
+        inout=(out_a.array()[1:], out_b.array()[1:]),
+        ref=(a.array()[1:], b.array()[1:]),
+        comparison=lambda a, b: (
```
DropD, line 241: Same issue as below. Non-empty containers are always True in Python. Must return a bool value.

DropD: The conditions can simply be connected with `and` rather than commas.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_program.py
```diff
@@ -36,205 +40,260 @@
...
+    cases.verify(
+        cartesian_case,
+        prog,
+        a,
+        b,
+        out_a,
+        out_b,
+        inout=(out_a.array(), out_b.array()),
+        ref=(a.array(), b.array()),
+        comparison=lambda a, b: (np.all(a == out_a), np.all(b == out_b)),
```
DropD, line 207: `comparison` must return a `bool`, see below.

DropD (changes requested):

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_program.py
```diff
@@ -36,205 +40,260 @@
...
-def fieldview_backend(request):
-    yield request.param
-
-
-def test_identity_fo_execution(fieldview_backend, identity_def):
-    size = 10
-    in_field = np_as_located_field(IDim)(np.ones((size)))
-    out_field = np_as_located_field(IDim)(np.zeros((size)))
-    identity = field_operator(identity_def, backend=fieldview_backend)
+def test_identity_fo_execution(cartesian_case, identity_def):
```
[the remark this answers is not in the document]
DropD: I see, now it makes sense.

PR #1273, opened 2023-06-19
fix[next]: Fix CSE extraction for if statements

tehrengruber:
The criterion for extracting common subexpressions out of `if_` calls was wrong. The correct criterion as implemented in this PR is that all subexpressions that either occur in the condition or both branches are eligible for extraction.


egparedes (approved):
It looks good.

PR #1275, opened 2023-06-21
refactor[next] error handling

petiaccja:
**Contents:**
- Add `CompilerError` and subclasses for specific errors
  - Location info + error msg
- Removed old `GT*Error` classes, `DialetSyntaxError` classes
- Replaced most `raise` statements in frontend [needs further work]
  - Most error messages now have proper location information
  - Some of the errors have updated types and error messages
- Replaced `raise` statements in `type_system` [needs further work]
- Added `sys.excepthook` for printing uncaught `CompilerError`s
- Added `__str__` method for printing uncaught errors in PyCharm

**Usage:**
- Compilation errors: raise one of the classes from `next.errors` (alias of `next.errors.exceptions)` or create a new class if makes sense
- General errors: use Python's builtin exception classes. If necessary, new classes can be added somewhere in `next.errors`.
- Don't override `sys.excepthook`!

**Results:**

Errors are displayed similarly to Python syntax errors:
```
Source location (most recent call last):
  File "/home/petiaccja/Work/ETHZ/code/gt4py_petiaccja/tests/next_tests/exception_printing.py", line 23
    def foo(var):
            ^^^
gt4py.next.errors.exceptions.MissingParameterAnnotationError: parameter 'var' is missing type annotations
```


src/gt4py/next/errors/exceptions.py
```diff
@@ -0,0 +1,111 @@
...
+
+import textwrap
+from typing import Any, Optional
+
+from gt4py.eve import SourceLocation
+
+from . import formatting
+
+
+class CompilerError(Exception):
```
egparedes, line 23: I think this name is a bit ambiguous and it could be understood as an error in the compiler ("internal compiler error") instead of an error emitted by the compiler but caused by wrong input code from the user. I  propose to change it to something along the lines of `DSLError` or `GT4PySourceError`  or similar

src/gt4py/next/errors/exceptions.py
```diff
@@ -0,0 +1,111 @@
...
+
+class MissingAttributeError(CompilerError):
+    attr_name: str
+
+    def __init__(self, location: Optional[SourceLocation], attr_name: str):
+        super().__init__(location, f"object does not have attribute '{attr_name}'")
+        self.attr_name = attr_name
+
+
+class CompilerTypeError(CompilerError):
```
egparedes, line 69: Maybe `DSLTypeError`?  `GT4PySourceTypeError`? ...?

src/gt4py/next/errors/exceptions.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, location: Optional[SourceLocation], param_name: str, type_: Any):
+        super().__init__(
+            location, f"parameter '{param_name}' has invalid type annotation '{type_}'"
+        )
+        self.param_name = param_name
+        self.annotated_type = type_
+
+
+class ArgumentCountError(CompilerTypeError):
+    num_excected: int
```
egparedes, line 95:
```suggestion
    expected_count: int
```

src/gt4py/next/errors/exceptions.py
```diff
@@ -0,0 +1,111 @@
...
+    num_provided: int
```
egparedes, line 96:
```suggestion
    provided_count: int
```

src/gt4py/next/errors/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+# This file is part of the GT4Py project and the GridTools framework.
+# GT4Py is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or any later
+# version. See the LICENSE.txt file at the top-level directory of this
+# distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from . import (  # noqa: module needs to be loaded for pretty printing of uncaught exceptions.
```
egparedes, line 15: I would add docstrings to all the submodules, or at  least a docstring here for the whole 'errors" subpackage.

src/gt4py/next/errors/formatting.py
```diff
@@ -0,0 +1,93 @@
...
+
+import pathlib
+import textwrap
+import traceback
+import types
+from typing import Optional
+
+from gt4py.eve import SourceLocation
+
+
```
egparedes, line 23: Missing return type annotations in multiple functions.

src/gt4py/next/errors/formatting.py
```diff
@@ -0,0 +1,93 @@
...
+def get_source_from_location(location: SourceLocation):
+    if not location.filename:
+        raise FileNotFoundError()
+    source_file = pathlib.Path(location.filename)
+    source_code = source_file.read_text()
+    source_lines = source_code.splitlines(False)
```
egparedes, line 29: Is there any reason to read the files from disk instead of getting the source code lines from `linecache`  (https://devdocs.io/python~3.11/library/linecache)?

src/gt4py/next/errors/formatting.py
```diff
@@ -0,0 +1,93 @@
...
+        caret_str = "".join([" "] * offset + ["^"] * width)
+    else:
+        caret_str = None
+
+    try:
+        snippet_str = get_source_from_location(loc)
+        if caret_str:
+            snippet_str = f"{snippet_str}\n{caret_str}"
+        return f"{loc_str}\n{textwrap.indent(snippet_str, '  ')}"
+    except Exception:
```
egparedes, line 53: Shouldn't it be
```suggestion
    except FileNotFoundError:
```
?

src/gt4py/next/errors/formatting.py
```diff
@@ -0,0 +1,93 @@
...
+    source_code = source_file.read_text()
+    source_lines = source_code.splitlines(False)
+    start_line = location.line
+    end_line = location.end_line + 1 if location.end_line else start_line + 1
+    relevant_lines = source_lines[(start_line - 1) : (end_line - 1)]
+    return "\n".join(relevant_lines)
+
+
+def format_location(loc: SourceLocation, caret: bool = False):
+    filename = loc.filename or "<unknown>"
```
egparedes, line 37: Optionally, you could add a docstring here to explain the meaning of the `caret` argument...

src/gt4py/next/errors/formatting.py
```diff
@@ -0,0 +1,93 @@
...
+        snippet_str = get_source_from_location(loc)
+        if caret_str:
+            snippet_str = f"{snippet_str}\n{caret_str}"
+        return f"{loc_str}\n{textwrap.indent(snippet_str, '  ')}"
+    except Exception:
+        return loc_str
+
+
+def _format_cause(cause: BaseException) -> list[str]:
+    """Print the cause of an exception plus the bridging message to STDERR."""
```
egparedes, line 58: The docstring seems wrong to me...

src/gt4py/next/errors/formatting.py
```diff
@@ -0,0 +1,93 @@
...
+    intro_message = "Traceback (most recent call last):"
+    traceback_strs = [
+        f"{intro_message}\n",
+        *traceback.format_tb(tb),
+    ]
+    return traceback_strs
+
+
+def format_compilation_error(
+    type_: type[Exception],
```
egparedes, line 75: Just a question, the type of `type_` should be `type[Exception]` or `type[CompilerError]` ?

tests/next_tests/unit_tests/ffront_tests/test_past_to_itir.py
```diff
@@ -19,7 +19,7 @@
 import gt4py.eve as eve
 import gt4py.next as gtx
 from gt4py.eve.pattern_matching import ObjectPattern as P
-from gt4py.next.common import GTTypeError
+from gt4py.next.errors import *
```
egparedes, line 22: Why???

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_type_deduction.py
```diff
@@ -31,10 +31,9 @@
     neighbor_sum,
     where,
 )
-from gt4py.next.common import GTTypeError
+from gt4py.next.errors import *
```
egparedes, line 34: Why?? this goes against the code conventions

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_scalar_if.py
```diff
@@ -19,7 +19,7 @@
 import pytest
 
 from gt4py.next import Field, field_operator, float64, index_field, np_as_located_field
-from gt4py.next.ffront.foast_passes.type_deduction import FieldOperatorTypeDeductionError
+from gt4py.next.errors import *
```
egparedes, line 22: Why??

src/gt4py/eve/concepts.py
```diff
@@ -58,58 +57,42 @@ class SymbolRef(ConstrainedStr, regex=_SYMBOL_NAME_RE):
...
-            )
-        if source is None:
-            source = f"<ast.{type(ast_node).__name__} at 0x{id(ast_node):x}>"
-        return cls(
-            ast_node.lineno,
-            ast_node.col_offset + 1,
-            source,
-            end_line=ast_node.end_lineno,
-            end_column=ast_node.end_col_offset + 1 if ast_node.end_col_offset is not None else None,
-        )
```
egparedes, line 87: Just a question, has this been moved to somewhere else or removed? Why?

tests/next_tests/exception_printing.py
egparedes, line 1: Just a question: what is this file and why is it here?

egparedes (changes requested):
The main changes in the PR look good to me. I've left many comments, questions and suggestions, but usually minor things.

src/gt4py/next/errors/excepthook.py
```diff
@@ -0,0 +1,94 @@
...
+
+import importlib_metadata
+
+from . import exceptions, formatting
+
+
+def _get_developer_mode_python() -> bool:
+    """Guess if the Python environment is used to develop gt4py."""
+    # Import gt4py and use its __name__ because hard-coding "gt4py" would fail
+    # silently if the module's name changes for whatever reason.
```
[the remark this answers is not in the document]
petiaccja: Yeah, what you're saying is probably right. The idea here is to avoid simply doing `s.startswith("gt4py")` because you would never realize it's wrong if the package name changed. If you `import gt4py`, that must be kept in sync with the real package name or else you'll crash with an import error, and the `__name__` is just to keep the string in sync with the import.

src/gt4py/next/errors/excepthook.py
```diff
@@ -0,0 +1,94 @@
...
+    # probably developing gt4py itself rather than something else using gt4py.
+    dists = importlib_metadata.distributions()
+    for dist in dists:
+        for req in dist.requires or []:
+            if req.startswith(package_name):
+                return False
+    return True
+
+
+def _get_developer_mode_os() -> Optional[bool]:
```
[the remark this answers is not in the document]
petiaccja: I've renamed them to `_get_developer_mode_envvar` and `_get_developer_mode_python_env` to make it clear.

src/gt4py/next/errors/exceptions.py
```diff
@@ -0,0 +1,111 @@
...
+# This file is part of the GT4Py project and the GridTools framework.
+# GT4Py is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or any later
+# version. See the LICENSE.txt file at the top-level directory of this
+# distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import textwrap
```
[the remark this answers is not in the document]
petiaccja: Done

src/gt4py/next/errors/exceptions.py
petiaccja: Agreed, that name is far from perfect, I'll put it up to a vote.

src/gt4py/next/errors/exceptions.py
petiaccja: Let's address this together with `CompilerError`.

src/gt4py/next/errors/formatting.py
petiaccja: Thanks, fixed the file.

src/gt4py/next/errors/formatting.py
petiaccja: I replaced it with `linecache`.

src/gt4py/next/errors/formatting.py
petiaccja: I added the list of exception handled explicitly.

src/gt4py/next/errors/formatting.py
petiaccja: Fixed.

src/gt4py/next/errors/formatting.py
petiaccja: Done, can you check if it makes things any better? Kinda difficult to explain this in one line. Any suggestions for a better variable name?

src/gt4py/next/errors/formatting.py
petiaccja: Maybe yes, but that would mean importing `exceptions` in `formatting`, but `exceptions` is already importing `formatting`.

tests/next_tests/exception_printing.py
petiaccja: That's for testing, I'll explain above.

petiaccja: It should not be merged.

tests/next_tests/unit_tests/errors_tests/test_exceptions.py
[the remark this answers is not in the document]
petiaccja: The fundamental issue is that testing these requires verifying conditions that depend on the state of the host system, not the state of the program, and forging the state of the host system for the test is complicated. How do you install the software in editable mode for a unit test?

That said, I'll do a refactoring and add tests because it's possible to better separate the logic from host system's state, and that can be tested.

src/gt4py/eve/concepts.py
petiaccja: I moved the functionality to the frontend: https://github.com/GridTools/gt4py/blob/89fc3a0a051cf4c1d85143c8a1351b32375b4c5a/src/gt4py/next/ffront/dialect_parser.py#L104

I wanted to break down the dependency between `SourceLocation` and the `ast` module to respect the Single Responsibility Principe so that `SourceLocation` is not taking up the responsibility of converting between different representations of location information.

I would be perfectly fine with a free function outside the frontend that converts between `ast` and `SourceLocation`, but I didn't see the need for it yet.

src/gt4py/next/errors/formatting.py
```diff
@@ -0,0 +1,99 @@
...
+    message: str,
+    location: Optional[SourceLocation],
+    tb: Optional[types.TracebackType] = None,
+    cause: Optional[BaseException] = None,
+) -> list[str]:
```
tehrengruber, line 86: Why is this a list? Sometimes you join using an empty string sometimes with "\n". Is there a reason I want to be able to choose? A small docstring with a doctest would be useful.

src/gt4py/next/ffront/dialect_parser.py
```diff
@@ -106,52 +62,20 @@ def apply(
...
-                    source_end_lineno = (
-                        (err.end_lineno - starting_line) if err.end_lineno else source_lineno
-                    )
-                    err.text = "\n".join(source.splitlines()[source_lineno : source_end_lineno + 1])
-                else:
-                    err.text = source
-
-            raise err
+        definition_ast: ast.AST
+        definition_ast = parse_source_definition(source_definition)
```
tehrengruber, line 66: Much cleaner :-)

src/gt4py/next/errors/formatting.py
petiaccja: It's to have an interface consistent with [traceback.format_exception](https://docs.python.org/3/library/traceback.html#traceback.format_exception). I have no idea why they chose a list of strings.

petiaccja: Nevertheless, it should always be joined without "\n", I'm not sure if I got all of them right.

src/gt4py/next/errors/excepthook.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def _get_developer_mode_envvar() -> Optional[bool]:
+    """Detect if the user set developer mode in environment variables."""
+    env_var_name = "GT4PY_DEVELOPER_MODE"
+    if env_var_name in os.environ:
+        try:
+            return bool(os.environ[env_var_name])
+        except TypeError:
+            return False
```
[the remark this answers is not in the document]
petiaccja: Right, returning `False` was not the best option. I changed it to return to `None` in that case, as this function is normally only called once at module load and the way to treat the exception is by using `None` in the subsequent functions.

src/gt4py/next/ffront/decorator.py
```diff
@@ -289,10 +289,8 @@ def _validate_args(self, *args, **kwargs) -> None:
                 with_kwargs=kwarg_types,
                 raise_exception=True,
             )
-        except GTTypeError as err:
-            raise ProgramTypeError.from_past_node(
-                self.past_node, msg=f"Invalid argument types in call to `{self.past_node.id}`!"
-            ) from err
+        except ValueError as err:
+            raise ValueError(f"Invalid argument types in call to `{self.past_node.id}`!") from err
```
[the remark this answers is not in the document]
petiaccja: Fixed.

tests/next_tests/unit_tests/ffront_tests/test_func_to_foast.py
```diff
@@ -195,8 +194,8 @@ def bool_and(a: gtx.Field[[TDim], "bool"], b: gtx.Field[[TDim], "bool"]):
         return a and b
 
     with pytest.raises(
-        FieldOperatorSyntaxError,
-        match=(r"`and`/`or` operator not allowed!"),
+        UnsupportedPythonFeatureError,
+        match=r".*and.*or.*",
```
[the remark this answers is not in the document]
petiaccja: I changed the names of the tests.

src/gt4py/next/errors/excepthook.py
```diff
@@ -0,0 +1,94 @@
...
+    package_name = gt4py.__name__
+
+    # Check if any package requires gt4py as a dependency. If not, we are
+    # probably developing gt4py itself rather than something else using gt4py.
+    dists = importlib_metadata.distributions()
+    for dist in dists:
+        for req in dist.requires or []:
+            if req.startswith(package_name):
+                return False
+    return True
```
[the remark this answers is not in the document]
petiaccja: I'll remove autodetection altogether.

src/gt4py/next/errors/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+from .excepthook import set_developer_mode
+from .exceptions import (
+    ArgumentCountError,
+    CompilerError,
+    InvalidParameterAnnotationError,
+    KeywordArgumentError,
+    MissingAttributeError,
+    MissingParameterAnnotationError,
+    UndefinedSymbolError,
+    UnsupportedPythonFeatureError,
```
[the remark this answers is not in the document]
petiaccja: I deleted the signature errors, and kept the rest as they are.

src/gt4py/next/errors/excepthook.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def _get_developer_mode_envvar() -> Optional[bool]:
+    """Detect if the user set developer mode in environment variables."""
+    env_var_name = "GT4PY_DEVELOPER_MODE"
```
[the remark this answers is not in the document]
petiaccja: Yeah, that's much better, renamed it.

src/gt4py/next/errors/excepthook.py
```diff
@@ -0,0 +1,94 @@
...
+    return _get_developer_mode_python_env()
+
+
+_developer_mode = _guess_developer_mode()
+
+
+def set_developer_mode(enabled: bool = False) -> None:
+    """In developer mode, information useful for gt4py developers is also shown."""
+    global _developer_mode
+    _developer_mode = enabled
```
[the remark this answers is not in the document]
petiaccja: Users might want programmatic control over this, you never know.

tests/next_tests/unit_tests/errors_tests/test_formatting.py
```diff
@@ -0,0 +1,57 @@
...
+frameinfo = inspect.getframeinfo(inspect.currentframe())
+loc = SourceLocation("/source/file.py", 5, 2, end_line=5, end_column=9)
+msg = "a message"
+
+module = CompilerError.__module__
+name = CompilerError.__name__
+try:
+    raise Exception()
+except Exception as ex:
+    tb = ex.__traceback__
```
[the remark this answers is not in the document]
petiaccja: No, there is just no other way to generate a traceback object.

tests/next_tests/unit_tests/errors_tests/test_exceptions.py
```diff
@@ -0,0 +1,56 @@
...
+from gt4py.eve import SourceLocation
+from gt4py.next.errors import CompilerError
+
+
+frameinfo = inspect.getframeinfo(inspect.currentframe())
+loc = SourceLocation("/source/file.py", 5, 2, end_line=5, end_column=9)
+loc_snippet = SourceLocation(
+    frameinfo.filename, frameinfo.lineno + 2, 15, end_line=frameinfo.lineno + 2, end_column=29
+)
+msg = "a message"
```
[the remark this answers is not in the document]
petiaccja: I changed them to fixtures.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_program.py
```diff
@@ -251,7 +249,50 @@ def empty_domain_program(a: cases.IJField, out_field: cases.IJField):
...
 
     with pytest.raises(
-        GTTypeError,
+        ValueError,
         match=(r"Dimensions in out field and field domain are not equivalent"),
     ):
         empty_domain_program(a, out_field, offset_provider={})
+
+
+def test_input_kwargs(fieldview_backend):
```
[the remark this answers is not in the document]
petiaccja: Can you elaborate?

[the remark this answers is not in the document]
tehrengruber: This test has been removed in another PR, but is reappearing here.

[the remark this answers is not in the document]
petiaccja: I see, I removed it.

src/gt4py/next/errors/exceptions.py
petiaccja: Renamed to `TypeError_`

src/gt4py/next/errors/exceptions.py
```diff
@@ -0,0 +1,115 @@
...
+            loc_str = formatting.format_location(self.location, show_caret=True)
+            return f"{self.message}\n{textwrap.indent(loc_str, '  ')}"
+        return self.message
+
+
+class UnsupportedPythonFeatureError(CompilerError):
+    feature: str
+
+    def __init__(self, location: Optional[SourceLocation], feature: str) -> None:
+        super().__init__(location, f"unsupported Python syntax: '{feature}'")
```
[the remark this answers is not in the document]
petiaccja: I'll leave things as they are now as the actual error messages are out of the scope of this PR.

src/gt4py/next/ffront/foast_passes/closure_var_folding.py
```diff
@@ -50,22 +51,12 @@ def visit_Name(
...
-            raise FieldOperatorSyntaxError.from_location(
-                msg="Constant does not have the attribute specified by the AST.",
-                location=node.location,
-            )
-        # TODO: use proper exception type (requires refactoring `FieldOperatorSyntaxError`)
-        raise FieldOperatorSyntaxError.from_location(
-            msg="Attribute can only be used on constants.", location=node.location
-        )
+            raise MissingAttributeError(node.location, node.attr)
+        raise CompilerError(node.location, "attribute access only applicable to constants")
```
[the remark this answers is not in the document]
petiaccja: I'll leave this as is, the actual error messages and their formatting should be addressed in another PR.

egparedes (commented):
All the coding and design issues have been addressed. I only have comments about coding style. I'll approve the PR as soon as they get addressed.

src/gt4py/next/errors/excepthook.py
```diff
@@ -0,0 +1,77 @@
...
+
+from . import exceptions, formatting
+
+
+def _get_verbose_exceptions_envvar() -> bool:
+    """Detect if the user enabled verbose exceptions in the environment variables."""
+    env_var_name = "GT4PY_VERBOSE_EXCEPTIONS"
+    if env_var_name in os.environ:
+        try:
+            return bool(os.environ[env_var_name])
```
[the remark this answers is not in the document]
petiaccja: No no, that's bad, what's the easiest way to convert a string to bool?

src/gt4py/next/errors/formatting.py
petiaccja: Done.

tests/next_tests/unit_tests/errors_tests/test_formatting.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_format(type_, qualname, message):
+    pattern = f"{qualname}: {message}"
+    s = "\n".join(format_compilation_error(type_, message, None, None, None))
+    assert re.match(pattern, s)
+
+
+def test_format_loc(type_, qualname, message, location):
+    pattern = "Source location.*\\n" '  File "/source.*".*\\n' f"{qualname}: {message}"
```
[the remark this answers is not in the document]
petiaccja: I think I wanted to split it into multiple lines so that it looks more similar to the actual text it represents, but the formatter was smart and put it all on one line.

I created separate variables for each line for now and joined them with `r"\n"`. The best alternative in my opinion would be to use a single-line multi-line raw f-string literal ( ̇r"""regex here""" ̇) so that quotes can be written easily.

src/gt4py/next/errors/excepthook.py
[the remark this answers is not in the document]
egparedes: The easiest way to convert a string to a bool is exactly what you did, but as str is a sequence, the default conversion is to True if it is not empty.

I think what you want to do here is to explicitly compare the envvar content against "True", "TRUE", "1" or "ON" to set the flag to `True`, against "False", "FALSE", "0" or "OFF" to set the flag to `False` and raise a warning otherwise. There is no builtin shortcut in Python for this.

tests/next_tests/unit_tests/errors_tests/test_excepthook.py
```diff
@@ -0,0 +1,34 @@
...
+from gt4py.next.errors import excepthook, exceptions
+
+
+def test_format_uncaught_error():
+    try:
+        loc = eve.SourceLocation("/src/file.py", 1, 1)
+        msg = "compile error msg"
+        raise exceptions.DSLError(loc, msg) from ValueError("value error msg")
+    except exceptions.DSLError as err:
+        str_devmode = "".join(excepthook._format_uncaught_error(err, True))
```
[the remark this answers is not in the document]
egparedes: Actually, I think I missed that the list of lines from `excepthook._format_uncaught_error` probably already contain a newline at then end, right? In that case, my comment is just wrong and you don't need to do anything here.

tests/next_tests/unit_tests/errors_tests/test_formatting.py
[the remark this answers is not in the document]
egparedes: > The best alternative in my opinion would be to use a single-line multi-line raw f-string literal ( ̇r"""regex here""" ̇)

I guess you mean `ḟr"""regex here"""` or `̇fr'''regex here'''`. Both of them, and also using separate variables look good to me. Pick the one you like the most.

src/gt4py/next/errors/excepthook.py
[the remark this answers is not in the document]
egparedes: Actually, I just remembered there is a super simple solution for this that I think we're using in some other places:
`bool(int(os.environ[env_var_name]))`
By converting the string to an int first, we guarantee that we follow meaningful semantics for "0" and "1" and any other value that is convertible to an int should be considered at least a warning.

tests/next_tests/unit_tests/errors_tests/test_excepthook.py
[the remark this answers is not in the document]
petiaccja: Yes, I tried to mimic `traceback` on this: 

> The return value is a list of strings, each ending in a newline and some containing internal newlines.

Given how you guys asked about this twice now it look like it's a pretty bad design. We can also deviate from the Python libraries if that's a better trade-off.

tests/next_tests/unit_tests/errors_tests/test_formatting.py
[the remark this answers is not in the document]
petiaccja: Yeah, that's what I meant. I'll keep them as separate variables for now, we can evolve it in the future in either direction as requirements change.

src/gt4py/next/errors/excepthook.py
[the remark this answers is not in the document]
petiaccja: Unfortunately, this doesn't work:
```
>>> int("False")
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: 'False'
```
I'll just use an explicit list.

src/gt4py/next/errors/excepthook.py
```diff
@@ -0,0 +1,94 @@
...
+
+def set_developer_mode(enabled: bool = False) -> None:
+    """In developer mode, information useful for gt4py developers is also shown."""
+    global _developer_mode
+    _developer_mode = enabled
+
+
+def compilation_error_hook(fallback: Callable, type_: type, value: BaseException, tb) -> None:
+    """
+    Format `CompilationError`s in a neat way.
```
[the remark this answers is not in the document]
petiaccja: Added information to the docstrings.

tests/next_tests/unit_tests/errors_tests/test_excepthook.py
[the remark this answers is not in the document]
egparedes: No, I think following the standard library makes sense here. Don't need to change anything.

PR #1276, opened 2023-06-22
ci-only: enable gridtools gpu code generation

DropD:
This draft PR purely exists to run the CI. CI is not expected to pass, neither on github nor on cscs-ci.  The interest lies in how they fail.


src/gt4py/next/otf/compilation/build_systems/cmake_lists.py
```diff
@@ -36,16 +44,20 @@ class CMakeListsFile(eve.Node):
...
     CMakeListsFile = as_jinja(
-        """
+        textwrap.dedent(
+            """
         project({{project_name}})
         cmake_minimum_required(VERSION 3.20.0)
 
         # Languages
         enable_language(CXX)
+        {{"\\n".join(extra_languages)}}
```
havogt, line 60: in a discussion with @egparedes we realized that probably we can just do 

```
include(CheckLanguage)
check_language(CUDA)
if(CMAKE_CUDA_COMPILER)
  enable_language(CUDA)
endif()
```
Ideally we would make the result of the check available to the Python side, because we can then error nicely, but as a first step we could just error by forwarding the CMake error which probably would be something like "`gtfn_gpu` is not a target".

havogt:
Implemented in #1325

PR #1279, opened 2023-07-03
feature[next]: Fix function signature validation for tuple arguments

tehrengruber:
Small fix to signature validation when calling a field-, scan-operator, program with a tuple parameter annotation, but a non-tuple argument in the call. E.g. something like:
```python
@field_operator
def foo_op(arg: tuple[...]):
  ...

foo_op(my_field)
```

__EDIT__:
While fixing this I stumbled upon another small bug that prevented scan operators to be called with boolean arguments. Since it is in the same function and only a couple of characters I put it into the same PR.

### DIFF
--- a/src/gt4py/next/ffront/type_info.py
+++ b/src/gt4py/next/ffront/type_info.py
@@ -36,13 +36,14 @@ def promote_zero_dims(
 
     def promote_arg(param: ts.TypeSpec, arg: ts.TypeSpec) -> ts.TypeSpec:
         def _as_field(arg_el: ts.TypeSpec, path: tuple[int, ...]) -> ts.TypeSpec:
-            param_el = reduce(lambda type_, idx: type_.types[idx] if isinstance(type_, ts.TupleType) else None, path, param)  # type: ignore[arg-type, return-value]
-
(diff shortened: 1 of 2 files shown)


src/gt4py/next/ffront/type_info.py
```diff
@@ -36,9 +36,17 @@ def promote_zero_dims(
 
     def promote_arg(param: ts.TypeSpec, arg: ts.TypeSpec) -> ts.TypeSpec:
         def _as_field(arg_el: ts.TypeSpec, path: tuple[int, ...]) -> ts.TypeSpec:
-            param_el = reduce(lambda type_, idx: type_.types[idx], path, param)  # type: ignore[attr-defined]
+            param_el = reduce(lambda type_, idx: type_.types[idx] if isinstance(type_, ts.TupleType) else None, path, param)  # type: ignore[arg-type, return-value]
```
egparedes, line 39: I'd rather decouple here the structure check (and error handling) from the actual logic.

egparedes (changes requested):

egparedes (approved):
ILGTM

PR #1280, opened 2023-07-03
bug[next]: itir.embedded: fix shift inside scan pass

tehrengruber:
Shifts were ignored in the `ScanArgIterator`, i.e. all shifts in the scan_pass were ignored. This partially fixes the `test_icon_like_scan`.

Additional:
- Run `InlineIntoScan` after temporary creation to inline remaining operations that block bubbling scans to the top.


src/gt4py/next/iterator/embedded.py
```diff
@@ -1196,7 +1195,7 @@ def can_deref(self) -> bool:
         return self.wrapped_iter.can_deref()
 
     def shift(self, *offsets: OffsetPart) -> ScanArgIterator:
-        return ScanArgIterator(self.wrapped_iter, self.k_pos, offsets=[*offsets, *self.offsets])
+        return ScanArgIterator(self.wrapped_iter.shift(*offsets), self.k_pos)
```
tehrengruber, line 1198: This is the important part.

src/gt4py/next/iterator/embedded.py
```diff
@@ -1392,7 +1391,12 @@ def _closure_runner():
                         col_pos[column.axis] = k
                         assert _is_concrete_position(col_pos)
                         ordered_indices = get_ordered_indices(out.axes, col_pos)
-                        out.field_setitem(ordered_indices, res[k])
+                        if isinstance(res, tuple):
+                            out.field_setitem(
+                                ordered_indices, tuple(res[i][k] for i in range(len(res)))
+                            )  # TODO(tehrengruber): only works for scalars
+                        else:
+                            out.field_setitem(ordered_indices, res[k])
```
havogt, line 1399: is this fixed in https://github.com/GridTools/gt4py/pull/1141/files? or is it another instance of that problem?

tests/next_tests/integration_tests/feature_tests/iterator_tests/test_scan.py
```diff
@@ -46,7 +46,7 @@ def shifted(inp):
 
     @fundef
     def wrapped(inp):
-        return scan(sum, True, 0.0)(inp, lift(shifted)(inp))
+        return scan(sum, True, 0.0)(inp, shift(Koff, 1)(inp))
```
havogt, line 49: This change breaks something else, if we undo we could merge.

havogt: I addressed this issue by adding another inline_into_scan pass after temporaries are created

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -286,6 +286,31 @@ def testee_op(
...
+    @gtx.field_operator
+    def testee(inp: gtx.Field[[KDim], float]) -> gtx.Field[[KDim], float]:
+        return testee_scan(inp(Koff[1]))
+
+    inp = cases.allocate(
+        cartesian_case,
+        testee,
+        "inp",
+        extend={KDim: (0, 1)},
+        strategy=cases.UniqueInitializer(start=1),
```
havogt, line 303: can you confirm that this should be `start=0`, with the `expected` below starting at `1`?

tests/next_tests/integration_tests/feature_tests/iterator_tests/test_scan.py
tehrengruber: We concluded that this is fine for now even though we are not sure whether the ordering of the passes makes sense anymore. This will be addressed as a whole in another PR.

havogt (approved):

tehrengruber:
The test failures seem to be a result of removing https://github.com/GridTools/gt4py/pull/1280/files/d4160bb32c7e091e94afe983fbe747c32c9f7514#r1250968455 again.

src/gt4py/next/iterator/embedded.py
tehrengruber: Not sure actually, might be two different things. This is not clean however as it only works for tuples not nested tuples.

PR #1283, opened 2023-07-05
test[next]: Add test for vertical boundary condition with scan

tehrengruber:
Draft, WIP

### DIFF
--- a/src/gt4py/next/iterator/embedded.py
+++ b/src/gt4py/next/iterator/embedded.py
@@ -817,7 +817,7 @@ class MDIterator:
             assert isinstance(k_pos, int)
             # the following range describes a range in the field
             # (negative values are relative to the origin, not relative to the size)
-            slice_column[self.column_axis] = range(k_pos, k_pos + len(column_range))
+            slice_column[self.column_axis] = range(k_pos, min(k_pos + len(column_range), column_range.stop))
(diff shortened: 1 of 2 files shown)


src/gt4py/next/iterator/embedded.py
```diff
@@ -817,7 +817,7 @@ def deref(self) -> Any:
             assert isinstance(k_pos, int)
             # the following range describes a range in the field
             # (negative values are relative to the origin, not relative to the size)
-            slice_column[self.column_axis] = range(k_pos, k_pos + len(column_range))
+            slice_column[self.column_axis] = range(k_pos, min(k_pos + len(column_range), column_range.stop))
```
tehrengruber, line 820: As discussed we want to implement a different solution where the size of the columns remains `column_range`, but is filled with `UNDEFINED` values for shifts that go out of bounds.

havogt: Filling with undefined is probably not good enough, in case we do `shift(K,-1)(shift(K,1)(inp))`, I believe we need the original values in the full column. I think, this requires the same shift and intersect logic that we discussed for embedded field view.
Maybe the undefined implementation is good enough for many use-cases, but when it breaks it will be hard to find the error.

PR #1293, opened 2023-07-21
fix[next]: DaCe scalar argument in stencil closure

edopao:
## Description

Add support for argument of stencil closure with `ScalarType`. Baseline only supported `FieldType` arguments.

## Requirements

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
@@ -293,9 +293,13 @@ class PythonTaskletCodegen(gt4py.eve.codegen.TemplatedGenerator):
         result: ValueExpr | SymbolExpr = self.visit(node.expr)[0]
         # Forwarding result through a tasklet needed because empty SDFG states don't properly forward connectors
         if isinstance(result, ValueExpr):
-            result = self.add_expr_tasklet([(result, "result")], "result", result.dtype, "forward")[
-                0
-            ]
+            result_name = unique_var_name()
+            self.context.body.add_scalar(result_name, result.dtype, transient=True)
+            result_access = self.context.state.add_access(result_name)
+            self.context.state.add_edge(
+                result.value, None, result_access, None, dace.Memlet(f"{result.value.data}[0]")
+            )
+            result = ValueExpr(value=result_access, dtype=result.dtype)
         else:
             result = self.add_expr_tasklet([], result.value, result.dtype, "forward")[0]
         self.context.body.arrays[result.value.data].transient = False
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
@@ -290,9 +290,14 @@ class PythonTaskletCodegen(gt4py.eve.codegen.TemplatedGenerator):
             context.body.add_array(name, shape=shape, strides=strides, dtype=dtype)
 
         # Translate the function's body
-        result: ValueExpr = self.visit(node.expr)[0]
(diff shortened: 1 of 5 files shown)


tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -917,7 +909,7 @@ def return_undefined():
 
 def test_zero_dims_fields(cartesian_case):
     if cartesian_case.backend == dace_iterator.run_dace_iterator:
-        pytest.xfail("Not supported in DaCe backend: broadcast")
+        pytest.xfail("Not supported in DaCe backend: zero-dimensional fields")
```
gronerl, line 912: Should we enable zero-dimensional fields as well? From running the tests quickly they seem to be treated differently throughout so could also be done separately.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -290,9 +290,14 @@ def visit_Lambda(
             context.body.add_array(name, shape=shape, strides=strides, dtype=dtype)
 
         # Translate the function's body
-        result: ValueExpr = self.visit(node.expr)[0]
+        result: ValueExpr | SymbolExpr = self.visit(node.expr)[0]
         # Forwarding result through a tasklet needed because empty SDFG states don't properly forward connectors
-        result = self.add_expr_tasklet([(result, "result")], "result", result.dtype, "forward")[0]
+        if isinstance(result, ValueExpr):
+            result = self.add_expr_tasklet([(result, "result")], "result", result.dtype, "forward")[
```
edopao, line 296: We can change the tasklet arg string to "inp" so we keep the assignment on one line.
```suggestion
            result = self.add_expr_tasklet([(result, "inp")], "inp", result.dtype, "forward")[0]
```

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
edopao: If they work, why not enabling them?

gronerl: The don't work!

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
gronerl: There, I made it even uglier :sweat_smile:  But this tasklet was really not necessary.

gronerl: Ok, I removed that part entirely. I don't think it is needed. @petiaccja can you think of a case where it is needed?

petiaccja: Yeah, we talked about this before, and we removed it because it was not functionally needed. But I had to add this "forwarding" thing back because it turned out that it *is* functionally needed. When you have an empty state with input and output connectors, and you want the input connectors to be forwarded to the output connectors, you eventually have to put something in the state because otherwise it doesn't work. I just put a dummy tasklet. If you have a simpler solution that's better, but before you do anything try to add a test and don't delete the comment.

gronerl: That's fair to leave it then, the version I put there now should be adequately transformed by simplify anyways.

petiaccja: Great! Unfortunately I don't remember the exact case that produced the empty state, and of course I didn't write a test for it either. We can of course use a less hacky method here as long as we don't run into problems with the source iterator IR. I'll resolve this for now.

petiaccja (approved):
The PR is pretty straightforward, we've already discussed the only questionable part, so it's approved.

PR #1295, opened 2023-07-24
Add minimal implementation of UnitRange and CartesianSet

samkellerhals:
Adds Cartesian set and UnitRange to gtx.common


havogt:
Part of #1296

PR #1297, opened 2023-07-25
fix[next]: DaCe scan connectivities

edopao:
## Description

Fix for scan operator in DaCe backend: add connectivities to lambda nested-SDFG in scan operator.

## Requirements

- [x] Tested in icon4py atmosphere dynamical core test

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
@@ -154,45 +154,48 @@ class ItirToSDFG(eve.NodeVisitor):
             # Create a new state for the closure.
             last_state = program_sdfg.add_state_after(last_state)
 
-            # Add access nodes for the program parameters to the closure's state in the program.
-            input_accesses = [last_state.add_access(name) for name in input_names]
-            connectivity_accesses = [last_state.add_access(name) for name in connectivity_names]
-            output_accesses = [last_state.add_access(name) for name in output_names]
+            # Create memlets to transfer the program parameters
+            input_memlets = [
+                create_memlet_full(name, program_sdfg.arrays[name]) for name in input_names
+            ]
+            connectivity_memlets = [
+                create_memlet_full(name, program_sdfg.arrays[name]) for name in connectivity_names
+            ]
+            output_memlets = [
+                create_memlet_full(name, program_sdfg.arrays[name]) for name in output_names
+            ]
 
-            # Map symbols by matching outer and inner strides, shapes, while defaulting to same symbol
-            symbol_mapping = {sym: sym for sym in closure_sdfg.free_symbols}
-            for inner_name, access_node in zip(
-                input_names + output_names, input_accesses + output_accesses
-            ):
-                outer_data = program_sdfg.arrays[access_node.data]
-                inner_data = closure_sdfg.arrays[inner_name]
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -432,24 +431,40 @@ def _visit_scan_stencil_closure(
...
+        symbol_mapping = {sym: sym for sym in lambda_context.body.free_symbols}
+        for inner_name, access_node in zip(connectivity_names, connectivity_access_nodes):
+            outer_data = scan_sdfg.arrays[access_node.data]
+            inner_data = lambda_context.body.arrays[inner_name]
+            for o_sh, i_sh in zip(outer_data.shape, inner_data.shape):
+                if str(i_sh) in lambda_context.body.free_symbols:
+                    symbol_mapping[str(i_sh)] = o_sh
+            for o_sh, i_sh in zip(outer_data.strides, inner_data.strides):
+                if str(i_sh) in lambda_context.body.free_symbols:
+                    symbol_mapping[str(i_sh)] = o_sh
```
gronerl, line 458: Why do we only have to do this for connectivities, but not all other arrays?

src/gt4py/next/program_processors/runners/dace_iterator/utility.py
```diff
@@ -59,8 +59,8 @@ def create_memlet_at(source_identifier: str, index: tuple[str, ...]):
 
 def map_nested_sdfg_symbols(
     parent_sdfg: dace.SDFG, nested_sdfg: dace.SDFG, array_mapping: dict[str, dace.Memlet]
-) -> dict[str, Any]:
-    symbol_mapping: dict[str, Any] = {}
+) -> dict[str, str]:
+    symbol_mapping: dict[str, str] = {}
```
gronerl, line 63: This function seems to do the same thing as mapping the symbols of connectivities in `_visit_scan_stencil_closure`. Can it not be used there to avoid code duplication?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
edopao: The output of a lambda seems to be a scalar, so not needed. The inputs of a lambda seem to be scalars and symbols. From the examples I have seen, the scan lambda behaves as a tasklet without side effects. Input data is passed as scalars (by value), while the memlet takes care of addressing the right element in the input array.
If this is the case, I guess the next question is why do we need to add connectivities to the scan lambda sdfg at all. The answer is that I was trying to address a runtime error in `visit_Lambda`, which is a generic function and should work for any kind of lambda function.
```
    def visit_Lambda(
        self, node: itir.Lambda, args: Sequence[ValueExpr | SymbolExpr]
    ):
        neighbor_tables = filter_neighbor_tables(self.offset_provider)
        param_names = [str(p.id) for p in node.params]
        conn_names = [connectivity_identifier(offset) for offset, _ in neighbor_tables]
```
Since we are passing the connectivity arrays, I can deduce that a lambda can access array elements directly (by reference) in the generic case.

src/gt4py/next/program_processors/runners/dace_iterator/utility.py
edopao: Right, fixed in next commit.

gronerl (approved):

PR #1298, opened 2023-07-25
Neighbor reductions

gronerl:
Implements neighbor reductions on a very restricted subset of itir:

Neighbor reductions need to be directly over field accesses, and use a mathematical binary builting as the reduction operation. 

If a general iterator is used, either a lift remains in the input of the reduction (which can not be inlined due to the parametric map-like nature of the reduction operation) or current itir pipeline alternatively seems to encode the operations defining the reduction input into a lambda that is used
(e.g. instead of `reduce(plus, a+b, 0.0)` , it will represent it as `reduce(lambda x1,x2,x3: x1+x2+x3), a,b,0.0)`
in which case we end up with the lambda case. 

Lambdas are not translateable to SDFGs in general, and it's own complexities arise with trying to support them in situations like that.


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -528,6 +627,49 @@ def _visit_indirect_addressing(self, node: itir.FunCall) -> IteratorExpr:
 
         return IteratorExpr(iterator.field, shifted_index, iterator.dtype, iterator.dimensions)
 
+    def _visit_reduce(self, node: itir.FunCall):
```
petiaccja, line 630: Why is this not a `builtin_reduce` like the other builtins?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -137,6 +137,105 @@ class Context:
...
+        mx,
+        result_access,
+        memlet=dace.Memlet(data=result_name, subset="neigh_idx"),
+        src_conn="__result",
+    )
+
+    return [ValueExpr(result_access, iterator.dtype)]
+
+
+def builtin_neighbors(
```
petiaccja, line 223: Why have this function separately from `_builtin_neighbors_indirect_addressing`?

edopao (commented):
I have checked the generated SDFG. No comments from my side, it looks good.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -137,6 +137,105 @@ class Context:
...
+        f"{offset_dim}_neighbors_map",
+        ndrange={"neigh_idx": f"0:{table.max_neighbors}"},
+    )
+    shift_tasklet = state.add_tasklet(
+        "shift",
+        code="__result = __table[__idx, neigh_idx]",
+        inputs={"__table", "__idx"},
+        outputs={"__result"},
+    )
+    data_access_tasklet = state.add_tasklet(
```
edopao, line 173: While running the icon4py diffusion tests, I encountered a problem on the ``data_access`` tasklet. The problem is that ``__field`` is supposed to have 1-d shape, but the memlet which is input to ``__field`` at line 202 is using the full shape. Validation does not catch this inconsistency, but the sdfg compilation fails.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -137,6 +137,105 @@ class Context:
...
+        dst_conn="__idx",
+    )
+    state.add_edge(
+        shift_tasklet,
+        "__result",
+        data_access_tasklet,
+        "__idx",
+        dace.Memlet(data=idx_name, subset="0"),
+    )
+    state.add_memlet_path(
```
edopao, line 202: Based on the assumption that `cell`/`edge` is always the 1st dim in shape, this code change in subset definition works in icon4py tests:
```suggestion
    field_subset = [f"0:{sdfg.arrays[iterator.field.data].shape[0]}"]
    if len(iterator.dimensions) > 1:
        field_subset += [f"i_{d}" for d in iterator.dimensions[1:]]
    state.add_memlet_path(
        iterator.field,
        me,
        data_access_tasklet,
        memlet=dace.Memlet(
            data=iterator.field.data,
            subset =",".join(field_subset),
        ),
        dst_conn="__field",
    )
```

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: It seems that the purpose of ``builtin_neighbors``is to check the preconditions to call ``_builtin_neighbors_indirect_addressing``
It leaves the possibility open to add other types of neighbor referencing in the future.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
petiaccja: Can't be because... `reduce(plus, 0)(neighbors(V2Eₒ, edge_f))`

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
petiaccja: I don't think it's a good idea to assume that cell/edge will always be the first dimension.

edopao: Right. I have addressed this comment in my latest commit.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: Fixed in [d04f60f](https://github.com/GridTools/gt4py/pull/1298/commits/d04f60ff954893a28edbc752bb2debd68e0b30e1)

edopao (approved):

PR #1305, opened 2023-07-27
fix: Suggestions and fixes for WIP embedded field implementation

egparedes:
Feel free to only take some parts

### DIFF
--- a/src/gt4py/next/common.py
+++ b/src/gt4py/next/common.py
@@ -228,6 +228,14 @@ class Field(Protocol[DimsT, gt4py_defs.ScalarT]):
 
 
 class FieldABC(Field[DimsT, gt4py_defs.ScalarT]):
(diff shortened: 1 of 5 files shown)


src/gt4py/next/common.py
```diff
@@ -203,6 +227,59 @@ def __pow__(self, other: Field | gt4py_defs.ScalarT) -> Field:
...
+        ...
+
+    @overload
+    @classmethod
+    def register_builtin_func(
+        cls, op: fbuiltins.BuiltInFunction[R, P], op_func: Callable[P, R]
+    ) -> Callable[P, R]:
+        ...
+
+    # TODO: discuss if this registration mechanism should be added as part of the
```
havogt, line 263: I think it shouldn't be part of the interface.

PR #1306, opened 2023-07-31
deps: add nanobind and (optional) jax for future features

havogt:
Adding in this separate PR to avoid (trivial but annoying) conflict resolution.


egparedes (approved):
It looks good to me.

PR #1307, opened 2023-08-02
refactor[next]: use __gt_dims__ in LocatedField

havogt:
Rename `LocatedField.axes` to `LocatedField.__gt_dims__` to make it compatible with the Cartesian and as preparation to switch to the new field implementation.

Note that gt4py.cartesian requires returning a tuple of `str`, for gt4py.next we require `tuple[Dimension]`. I propose to relax the cartesian requirement to convertible to `str`, which should already work as the documentation demands to go through `gt4py.storage.utils.get_dims()` which has string conversion.

### DIFF
--- a/src/gt4py/next/iterator/embedded.py
+++ b/src/gt4py/next/iterator/embedded.py
@@ -174,7 +174,7 @@ class LocatedField(Protocol):
 
     @property
     @abc.abstractmethod
-    def axes(self) -> tuple[common.Dimension, ...]:
+    def __gt_dims__(self) -> tuple[common.Dimension, ...]:
         ...
 
     # TODO(havogt): define generic Protocol to provide a concrete return type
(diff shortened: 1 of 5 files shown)


src/gt4py/next/iterator/embedded.py
```diff
@@ -1260,7 +1263,7 @@ def _get_axeses(field):
         return tuple(itertools.chain(*tuple(_get_axeses(f) for f in field)))
     else:
         assert is_located_field(field)
-        return (field.axes,)
+        return (_get_axes(field),)
```
tehrengruber, line 1266: Why not? The field can not be a tuple anymore anyway.
```suggestion
        return (field.__gt_dims__,)
```

tehrengruber (changes requested):

src/gt4py/next/iterator/embedded.py
havogt: Since this is the only place were I directly access __gt_dims_ (all other go through `_get_axes`) I thought it makes sense, but probably doesn't make sense. I'll revert.

tehrengruber (approved):

PR #1310, opened 2023-08-03
Introduce Fieldview Domain

samkellerhals:
## Description

Introduces intersectable `Domain` and `UnitRange` to `gt4py.next.common`

## Requirements

### DIFF
--- a/src/gt4py/next/common.py
+++ b/src/gt4py/next/common.py
@@ -130,11 +130,6 @@ class UnitRange(Sequence[int]):
     def __and__(self, other: UnitRange) -> UnitRange:
         start = max(self.start, other.start)
         stop = min(self.stop, other.stop)
-
-        # Handle the case where there is no overlap
-        if stop <= start:
-            return UnitRange(0, 0)
-
         return UnitRange(start, stop)
 
 
@@ -182,8 +177,8 @@ class Domain(Sequence):
 
     def __and__(self, other: "Domain") -> "Domain":
         broadcast_dims = promote_dims(self.dims, other.dims)
-        broadcasted_ranges_self = self.broadcast_ranges(broadcast_dims)
-        broadcasted_ranges_other = other.broadcast_ranges(broadcast_dims)
+        broadcasted_ranges_self = broadcast_ranges(broadcast_dims, self.dims, self.ranges)
+        broadcasted_ranges_other = broadcast_ranges(broadcast_dims, other.dims, other.ranges)
 
         intersected_ranges = []
         for range1, range2 in zip(broadcasted_ranges_self, broadcasted_ranges_other):
@@ -192,16 +187,19 @@ class Domain(Sequence):
 
         return Domain(broadcast_dims, intersected_ranges)
 
-    def broadcast_ranges(self, broadcast_dims: list[Dimension]) -> list[UnitRange]:
-        if len(self.dims) == len(broadcast_dims):
-            return self.ranges
 
-        # Broadcast dimensions with infinite sizes for missing ranges
-        broadcasted_ranges = list(self.ranges)
-        for i in range(len(broadcast_dims) - len(self.dims)):
-            broadcasted_ranges.append(UnitRange(Infinity.negative(), Infinity.positive()))
+def broadcast_ranges(
+    broadcast_dims: list[Dimension], dims: list[Dimension], ranges: list[UnitRange]
+) -> list[UnitRange]:
+    if len(dims) == len(broadcast_dims):
+        return ranges
 
-        return broadcasted_ranges
+    # Broadcast dimensions with infinite sizes for missing ranges
+    broadcasted_ranges = list(ranges)
+    for i in range(len(broadcast_dims) - len(dims)):
+        broadcasted_ranges.append(UnitRange(Infinity.negative(), Infinity.positive()))
+
+    return broadcasted_ranges
 
 
 if TYPE_CHECKING:
@@ -210,7 +208,6 @@ if TYPE_CHECKING:
     _Value: TypeAlias = "Field" | gt4py_defs.ScalarT
     _P = ParamSpec("_P")
     _R = TypeVar("_R", _Value, tuple[_Value, ...])
-
 
     class GTBuiltInFuncDispatcher(Protocol):
         def __call__(self, func: fbuiltins.BuiltInFunction[_R, _P], /) -> Callable[_P, _R]:
@@ -325,11 +322,11 @@ class FieldABC(Field[DimsT, gt4py_defs.ScalarT]):
 
 @functools.singledispatch
 def field(
-        definition: Any,
-        /,
-        *,
-        domain: Optional[Any] = None,  # TODO(havogt): provide domain_like to DomainT conversion
-        value_type: Optional[type] = None,
+    definition: Any,
+    /,
+    *,
+    domain: Optional[Any] = None,  # TODO(havogt): provide domain_like to DomainT conversion
+    value_type: Optional[type] = None,
 ) -> Field:
     raise NotImplementedError
 
@@ -359,7 +356,7 @@ class Connectivity(Protocol):
     index_type: type[int] | type[np.int32] | type[np.int64]
 
     def mapped_index(
-            self, cur_index: int | np.integer, neigh_index: int | np.integer
+        self, cur_index: int | np.integer, neigh_index: int | np.integer
     ) -> Optional[int | np.integer]:
         """Return neighbor index."""
 
--- a/src/gt4py/next/common.py
+++ b/src/gt4py/next/common.py
@@ -18,8 +18,9 @@ import abc
 import dataclasses
 import enum
 import functools
-from collections.abc import Sequence, Set, Collection
-from typing import overload
+import sys
+from collections.abc import Sequence, Set
+from typing import overload, AbstractSet
 
 import numpy as np
 import numpy.typing as npt
@@ -50,6 +51,16 @@ Scalar: TypeAlias = gt4py_defs.Scalar
 NDArrayObject = gt4py_defs.NDArrayObject
 
 
+class Infinity:
+    @staticmethod
+    def positive() -> int:
+        return sys.maxsize
+
+    @staticmethod
+    def negative() -> int:
+        return -sys.maxsize - 1
+
+
 @enum.unique
 class DimensionKind(StrEnum):
     HORIZONTAL = "horizontal"
@@ -73,7 +84,7 @@ DomainLike = Union[Sequence[Dimension], Dimension, str]
 
 
 @dataclasses.dataclass(frozen=True)
-class UnitRange(Sequence[int], Set[int]):
+class UnitRange(Sequence[int]):
     """Range from `start` to `stop` with step size one."""
 
     start: int
@@ -99,7 +110,7 @@ class UnitRange(Sequence[int], Set[int]):
     def __getitem__(self, index: slice) -> UnitRange:
         ...
 
-    def __getitem__(self, index: int | slice) -> int | UnitRange:
+    def __getitem__(self, index: int | slice) -> int | float | UnitRange:
         if isinstance(index, slice):
             start, stop, step = index.indices(len(self))
             if step != 1:
@@ -130,89 +141,10 @@ class UnitRange(Sequence[int], Set[int]):
 DomainT: TypeAlias = tuple[tuple[Dimension, UnitRange], ...]
 
 
-def promote_dims(*dims_list: list[Dimension]) -> list[Dimension]:
-    """
-    Find a unique ordering of multiple (individually ordered) lists of dimensions.
-
-    The resulting list of dimensions contains all dimensions of the arguments
-    in the order they originally appear. If no unique order exists or a
-    contradicting order is found an exception is raised.
-
-    A modified version (ensuring uniqueness of the order) of
-    `Kahn's algorithm <https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm>`_
-    is used to topologically sort the arguments.
-
-    >>> I, J, K = (Dimension(value=dim) for dim in ["I", "J", "K"])
-    >>> promote_dims([I, J], [I, J, K]) == [I, J, K]
-    True
-    >>> promote_dims([I, J], [K]) # doctest: +ELLIPSIS
-    Traceback (most recent call last):
-     ...
-    ValueError: Dimensions can not be promoted. Could not determine order of the following dimensions: J, K.
-    >>> promote_dims([I, J], [J, I]) # doctest: +ELLIPSIS
-    Traceback (most recent call last):
-     ...
-    ValueError: Dimensions can not be promoted. The following dimensions appear in contradicting order: I, J.
-    """
-    # build a graph with the vertices being dimensions and edges representing
-    #  the order between two dimensions. The graph is encoded as a dictionary
-    #  mapping dimensions to their predecessors, i.e. a dictionary containing
-    #  adjacency lists. Since graphlib.TopologicalSorter uses predecessors
-    #  (contrary to successors) we also use this directionality here.
-    graph: dict[Dimension, set[Dimension]] = {}
-    for dims in dims_list:
-        if len(dims) == 0:
-            continue
-        # create a vertex for each dimension
-        for dim in dims:
-            graph.setdefault(dim, set())
-        # add edges
-        predecessor = dims[0]
-        for dim in dims[1:]:
-            graph[dim].add(predecessor)
-            predecessor = dim
-
-    # modified version of Kahn's algorithm
-    topologically_sorted_list: list[Dimension] = []
-
-    # compute in-degree for each vertex
-    in_degree = {v: 0 for v in graph.keys()}
-    for v1 in graph:
-        for v2 in graph[v1]:
-            in_degree[v2] += 1
-
-    # process vertices with in-degree == 0
-    # TODO(tehrengruber): avoid recomputation of zero_in_degree_vertex_list
-    while zero_in_degree_vertex_list := [v for v, d in in_degree.items() if d == 0]:
-        if len(zero_in_degree_vertex_list) != 1:
-            raise ValueError(
-                f"Dimensions can not be promoted. Could not determine "
-                f"order of the following dimensions: "
-                f"{', '.join((dim.value for dim in zero_in_degree_vertex_list))}."
-            )
-        v = zero_in_degree_vertex_list[0]
-        del in_degree[v]
-        topologically_sorted_list.insert(0, v)
-        # update in-degree
-        for predecessor in graph[v]:
-            in_degree[predecessor] -= 1
-
-    if len(in_degree.items()) > 0:
-        raise ValueError(
-            f"Dimensions can not be promoted. The following dimensions "
-            f"appear in contradicting order: {', '.join((dim.value for dim in in_degree.keys()))}."
-        )
-
-    return topologically_sorted_list
-
-
-class Domain(Collection):
-    dims: tuple[Dimension]
-    ranges: tuple[UnitRange]
-
-    def __init__(self, dims: tuple[Dimension, ...], ranges: tuple[UnitRange, ...]):
-        self.dims = dims
-        self.ranges = ranges
+@dataclasses.dataclass(frozen=True)
+class Domain(Sequence):
+    dims: list[Dimension]
+    ranges: list[UnitRange]
 
     def __len__(self) -> int:
         return len(self.ranges)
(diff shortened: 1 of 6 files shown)


src/gt4py/next/common.py
```diff
@@ -117,10 +116,143 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+        if stop <= start:
+            return UnitRange(0, 0)
+
+        return UnitRange(start, stop)
+
 
 DomainT: TypeAlias = tuple[tuple[Dimension, UnitRange], ...]
 
 
+def promote_dims(*dims_list: list[Dimension]) -> list[Dimension]:
```
havogt, line 133: rename to something generic e.g. `topological_sort` and move to gt4py._core.utils (or something)
```suggestion
T = TypeVar("T", Hashable) #???
def topological_sort(*dims_list: list[T]) -> list[T]:
```

src/gt4py/next/common.py
```diff
@@ -117,10 +116,143 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+    if len(in_degree.items()) > 0:
+        raise ValueError(
+            f"Dimensions can not be promoted. The following dimensions "
+            f"appear in contradicting order: {', '.join((dim.value for dim in in_degree.keys()))}."
+        )
+
+    return topologically_sorted_list
+
+
+class Domain(Collection):
```
havogt, line 209:
```suggestion
class Domain(Sequence):
```

src/gt4py/next/common.py
```diff
@@ -117,10 +116,143 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+    def __init__(self, dims: tuple[Dimension, ...], ranges: tuple[UnitRange, ...]):
+        self.dims = dims
+        self.ranges = ranges
+
+    def __len__(self) -> int:
+        return len(self.ranges)
+
+    def __iter__(self):
+        return iter(zip(self.dims, self.ranges))
+
```
havogt, line 222:
```
@overload
def __getitem__(self, index: int) -> tuple[Dimension, UnitRange]
    ...
@overload
def __getitem__(self, index: slice) -> Domain:
    ...
@overload
def __getitem__(self, key: Dimension, index: int) -> UnitRange:
    ...
```
maybe

src/gt4py/next/common.py
```diff
@@ -117,10 +116,143 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+    def __and__(self, other: Domain) -> Domain:
+        broadcast_dims = tuple(promote_dims(self.dims, other.dims))
+        broadcasted_ranges_self = self.broadcast_ranges(broadcast_dims)
+        broadcasted_ranges_other = other.broadcast_ranges(broadcast_dims)
+
+        intersected_ranges = []
+        for range1, range2 in zip(broadcasted_ranges_self, broadcasted_ranges_other):
+            intersected_range = range1 & range2
+            intersected_ranges.append(intersected_range)
+
```
havogt, line 241: this could be in CartesianSet if tuple[UnitRange] is replaced

src/gt4py/next/common.py
```diff
@@ -117,10 +116,143 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+        return Domain(broadcast_dims, tuple(intersected_ranges))
+
+    def broadcast_ranges(self, broadcast_dims: tuple[Dimension, ...]) -> tuple[UnitRange, ...]:
```
havogt, line 244:
```suggestion
    def broadcasted_ranges(self, broadcast_dims: tuple[Dimension, ...]) -> tuple[UnitRange, ...]:
```
or make a free function for clarity (as it's not changing the current instance.

src/gt4py/next/common.py
```diff
@@ -117,10 +116,143 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+
+    if len(in_degree.items()) > 0:
+        raise ValueError(
+            f"Dimensions can not be promoted. The following dimensions "
+            f"appear in contradicting order: {', '.join((dim.value for dim in in_degree.keys()))}."
+        )
+
+    return topologically_sorted_list
+
+
```
havogt, line 208:
```suggestion
@dataclasses.dataclass(frozen=True)
```

src/gt4py/next/common.py
```diff
@@ -117,10 +116,143 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+        return Domain(broadcast_dims, tuple(intersected_ranges))
+
+    def broadcast_ranges(self, broadcast_dims: tuple[Dimension, ...]) -> tuple[UnitRange, ...]:
+        if len(self.dims) == len(broadcast_dims):
+            return self.ranges
+
+        # Broadcast dimensions with infinite sizes for missing ranges
+        broadcasted_ranges = list(self.ranges)
+        for i in range(len(broadcast_dims) - len(self.dims)):
+            broadcasted_ranges.append(UnitRange(float("-inf"), float("inf")))
```
havogt, line 251: What does mypy say to that? Maybe `start = None` could be -Inf and `stop=None` +Inf

src/gt4py/next/common.py
tehrengruber: This function is not just topologically sorting, but also ensuring the order is unique. This is a rather special requirement and one would need to enhance the error messages again (they are user facing) for the dimension case. As long as it is not used in a more generic context it seams like a violation of KISS to generalize this.

havogt: so where do you want to put it?

samkellerhals: Could we leave it in `common` for now? Placing it in `gt4py._core.utils` still results in circular imports due to importing `Dimension` for the type annotations.

src/gt4py/next/common.py
samkellerhals: For now I have left it as part of `Domain` as discussed

src/gt4py/next/common.py
samkellerhals: Since `__len__()` always has to return an `int` this is not possible. I've implemented my own `Infinity` class with its own positive and negative infinities, which satisfies this constraint and mypy.

src/gt4py/next/common.py
```diff
@@ -117,10 +129,102 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+        return UnitRange(start, stop)
+
 
 DomainT: TypeAlias = tuple[tuple[Dimension, UnitRange], ...]
 
 
+from dataclasses import dataclass
+from collections.abc import Sequence
+
+
```
havogt, line 144: maybe we need something like 

```
NamedRange: TypeAlias = tuple[Dimension, UnitRange]
NamedPosition: TypeAlias = tuple[[Dimension, int], ...]
```

src/gt4py/next/common.py
```diff
@@ -117,10 +129,102 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+        elif isinstance(index, Dimension):
+            dim_range_subset = [
+                (dim, range_) for dim, range_ in zip(self.dims, self.ranges) if dim == index
+            ]
+            if len(dim_range_subset) > 1:
+                return Domain(
+                    [dim_range[0] for dim_range in dim_range_subset],
+                    [dim_range[1] for dim_range in dim_range_subset],
+                )
+            elif len(dim_range_subset) == 1:
```
havogt, line 194: just assert that Dimension is unique for now

src/gt4py/next/common.py
```diff
@@ -117,10 +129,102 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+        start = max(self.start, other.start)
+        stop = min(self.stop, other.stop)
+        return UnitRange(start, stop)
+
 
 DomainT: TypeAlias = tuple[tuple[Dimension, UnitRange], ...]
 
 
+from dataclasses import dataclass
+from collections.abc import Sequence
```
havogt, line 142: move to top

src/gt4py/next/common.py
```diff
@@ -117,10 +129,102 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
             else:
                 raise IndexError("UnitRange index out of range")
 
+    def __and__(self, other: UnitRange) -> UnitRange:
+        start = max(self.start, other.start)
+        stop = min(self.stop, other.stop)
+        return UnitRange(start, stop)
+
 
 DomainT: TypeAlias = tuple[tuple[Dimension, UnitRange], ...]
```
havogt, line 138: Delete and change all uses of DomainT to the new domain, e.g. test_nd_array_field

src/gt4py/next/ffront/type_info.py
```diff
@@ -17,7 +17,7 @@
 
 import gt4py.next.ffront.type_specifications as ts_ffront
 import gt4py.next.type_system.type_specifications as ts
-from gt4py.next.common import Dimension
+from gt4py.next.common import Dimension, promote_dims
```
havogt, line 20:
```suggestion
from gt4py.next import common
```
to follow our style guide (please update all usages of Dimension, too)

src/gt4py/next/type_system/type_info.py
```diff
@@ -19,7 +19,7 @@
 import numpy as np
 
 from gt4py.eve.utils import XIterable, xiter
-from gt4py.next.common import Dimension, DimensionKind
+from gt4py.next.common import Dimension, DimensionKind, promote_dims
```
havogt, line 22:
```suggestion
from gt4py.next import common
```
as above

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_type_deduction.py
```diff
@@ -36,6 +36,7 @@
 from gt4py.next.ffront.experimental import as_offset
 from gt4py.next.ffront.func_to_foast import FieldOperatorParser
 from gt4py.next.type_system import type_info, type_specifications as ts
+from gt4py.next.common import promote_dims
```
havogt, line 39:
```suggestion
from gt4py.next import common
```

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_type_deduction.py
```diff
@@ -498,10 +499,10 @@ def test_dimension_promotion(
     expected_error_msg: Optional[str],
 ):
     if expected_result:
-        assert type_info.promote_dims(*dim_list) == expected_result
+        assert promote_dims(*dim_list) == expected_result
     else:
         with pytest.raises(Exception) as exc_info:
```
havogt, line 504: this test should now be moved to test_common

tests/next_tests/unit_tests/test_common.py
```diff
@@ -64,3 +82,144 @@ def test_unit_range_index_error(rng):
...
+def test_domain_not_contains_element(domain):
+    assert 100 not in domain
+    assert -5 not in domain
+    assert 16 not in domain
+
+
+def test_positive_infinity_range():
+    pos_inf_range = UnitRange(Infinity.positive(), Infinity.positive())
+    assert len(pos_inf_range) == 0
+
```
havogt, line 134: please add a test that intersection with Infinity works correctly (because if we change Infinity later, its important we don't change the behavior of UnitRange)

tests/next_tests/unit_tests/test_common.py
```diff
@@ -64,3 +82,144 @@ def test_unit_range_index_error(rng):
...
+    assert all(res == exp for res, exp in zip(result, expected))
+
+
+import pytest
+
+
+# Assume you have already defined the Domain class and IDim, JDim, KDim classes.
+
+@pytest.fixture
+def domain_repeat_dims():
```
havogt, line 198: I'd keep this test and check for the exception and document here that we might change it in the future

havogt (changes requested):
Some more comments

As discussed, check for all the usages of domain as `tuple[[Dimension, UnitRange], ...]` currently and update accordingly

src/gt4py/next/common.py
```diff
@@ -117,10 +129,102 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
             else:
                 raise IndexError("UnitRange index out of range")
 
+    def __and__(self, other: UnitRange) -> UnitRange:
+        start = max(self.start, other.start)
+        stop = min(self.stop, other.stop)
+        return UnitRange(start, stop)
```
havogt, line 135: As mypy told me, this violates Liskov substitution principle, we need to cover the case where other is a Set[int], e.g something like
```suggestion
    def __and__(self, other: Set[int]) -> UnitRange:
        if isinstance(other, UnitRange):
            start = max(self.start, other.start)
            stop = min(self.stop, other.stop)
            return UnitRange(start, stop)
        else:
            other_start = min(other)
            other_stop = max(other)
            if len(other) != other_stop - other_start:
                raise ValueError("Intersection with non-contiguous range is not supported.")
            start = max(self.start, other_start)
            stop = min(self.stop, other_stop)
            return UnitRange(start, stop)
```

src/gt4py/next/common.py
```diff
@@ -51,6 +51,16 @@
 NDArrayObject = gt4py_defs.NDArrayObject
 
 
+class Infinity:
```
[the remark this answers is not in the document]
samkellerhals: It's a good point, we can make Infinity a subclass of int.

src/gt4py/next/common.py
```diff
@@ -74,7 +84,7 @@ def __str__(self):
 
 
 @dataclasses.dataclass(frozen=True)
-class UnitRange(Sequence[int], Set[int]):
+class UnitRange(Sequence[int]):
```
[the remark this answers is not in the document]
samkellerhals: Because `UnitRange` is overriding the `__and__` method and providing a custom implementation for that expecting that the type of `other` is `UnitRange`. If I inherit from `Set[int]` I get the following mypy error:

```
src/gt4py/next/common.py:130:23: error: Argument 1 of "__and__" is incompatible with supertype "set"; supertype defines the argument type as "AbstractSet[object]"  [override]
src/gt4py/next/common.py:130:23: note: This violates the Liskov substitution principle
src/gt4py/next/common.py:130:23: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
src/gt4py/next/common.py:130:23: error: Argument 1 of "__and__" is incompatible with supertype "AbstractSet"; supertype defines the argument type as "AbstractSet[Any]"  [override]
``` 

If we inherit from set the `__and__` method expects that `other` has type `AbstractSet` which is not the case for us. It seems to me that UnitRange doesn't need to fully implement the Set interface, as it is mainly meant to represent a numerical range and doesn't behave like a full set.

src/gt4py/next/common.py
```diff
@@ -87,6 +97,8 @@ def __post_init__(self):
             object.__setattr__(self, "stop", 0)
 
     def __len__(self) -> int:
+        if Infinity.positive() in (abs(self.start), abs(self.stop)):
+            raise ValueError("Cannot get length of infinite range.")
```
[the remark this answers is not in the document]
samkellerhals: This is what I thought too initially, but in that case Python complains with `OverflowError: cannot fit 'int' into an index-sized integer`

src/gt4py/next/common.py
samkellerhals: Do we want to support intersections with sets too then or only UnitRange?

src/gt4py/next/common.py
samkellerhals: So for now I just added `NamedRange` as this is what `Domain` will return from `__get_item__` and `__contains__`, as far as I understand we might worry about `NamedPosition` later?

src/gt4py/next/common.py
```diff
@@ -117,10 +129,102 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
 
 
+from dataclasses import dataclass
+from collections.abc import Sequence
+
+
+@dataclass(frozen=True)
+class Domain(Sequence):
+    dims: list[Dimension]
+    ranges: list[UnitRange]
```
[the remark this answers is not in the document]
samkellerhals: I used `list` as otherwise mypy complains since `promote_dims` expects a list of Dimensions so I just adapted to that interface to avoid having to change type hints everywhere.

```
src/gt4py/next/common.py:183:39: error: Argument 1 to "promote_dims" has incompatible type "Tuple[Dimension]"; expected "List[Dimension]"  [arg-type]
```

src/gt4py/next/common.py
```diff
@@ -117,8 +127,81 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+        intersected_ranges = [
+            rng1 & rng2
+            for rng1, rng2 in zip(
+                self._broadcast(broadcast_dims, self.dims, self.ranges),
+                self._broadcast(broadcast_dims, other.dims, other.ranges),
+            )
+        ]
+        return Domain(broadcast_dims, intersected_ranges)
+
+    @staticmethod
```
samkellerhals, line 193: I've added this back to Domain as its just a utility method, not really relevant anywhere else.

tests/next_tests/unit_tests/test_common.py
```diff
@@ -64,3 +81,117 @@ def test_unit_range_index_error(rng):
...
+    assert UnitRange(20, 30) in domain
+    assert UnitRange(25, 28) in domain
+
+
+def test_domain_contains_element(domain):
+    assert 0 in domain
+    assert 5 in domain
+    assert 20 in domain
+    assert 9 in domain
+    assert 14 in domain
```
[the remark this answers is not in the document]
samkellerhals: The default __contains__ method now just checks for the presence of a `NamedRange` inside the Domain.

src/gt4py/next/common.py
samkellerhals: I've added a `__post_init__` check for this which simplifies `__get_item__`

src/gt4py/next/common.py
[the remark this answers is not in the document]
havogt: With the condition you could still return `Infinity.positive()` I guess. Just `Infinity.positive() - Infinity.negative()` (i.e. default implementation will not work)

src/gt4py/next/common.py
havogt: this will solve the problem that you had with adding `Set` to the inheritance

havogt: alternatively we could also just `raise NotImplementedError()` for the case where other is not `UnitRange`

src/gt4py/next/common.py
havogt: Makes sense!

src/gt4py/next/common.py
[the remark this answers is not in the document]
havogt: Then let's change the interface of promote dims to

```
promote_dims(*dims_list: Sequence[Dimension]) -> list[Dimension]
```

src/gt4py/next/common.py
havogt: this is against our style guide https://google.github.io/styleguide/pyguide.html#2174-decision

src/gt4py/next/common.py
```diff
@@ -117,8 +128,81 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
             else:
                 raise IndexError("UnitRange index out of range")
 
+    def __and__(self, other: AbstractSet[Any]) -> UnitRange:
```
havogt, line 131:
```suggestion
    def __and__(self, other: Set[Any]) -> UnitRange:
```

havogt: see https://docs.python.org/3/library/typing.html#typing.AbstractSet

src/gt4py/next/common.py
```diff
@@ -18,9 +18,9 @@
 import dataclasses
 import enum
 import functools
-from collections.abc import Sequence, Set
-from typing import overload
-
+import sys
+from collections.abc import Sequence
+from typing import overload, Set, AbstractSet
```
havogt, line 23:
```suggestion
from collections.abc import Sequence, Set
from typing import overload
```
we want the metaclass such that UnitRange gets comparison operators and stuff

src/gt4py/next/common.py
```diff
@@ -117,8 +128,81 @@ def __getitem__(self, index: int | slice) -> int | UnitRange:
...
+            rng1 & rng2
+            for rng1, rng2 in zip(
+                broadcast_ranges(broadcast_dims, self.dims, self.ranges),
+                broadcast_ranges(broadcast_dims, other.dims, other.ranges),
+            )
+        ]
+        return Domain(broadcast_dims, intersected_ranges)
+
+
+def broadcast_ranges(
```
havogt, line 195:
```suggestion
def _broadcast_ranges(
```

havogt (changes requested):

PR #1311, opened 2023-08-03
refactor[next]: iterator.embedded cleanups

havogt:
- field_set/getitem accept named indices: move the ordering of indices into the field implementation: the field is the abstraction which gives buffer dimensions a tag...
- remove `runtime.Offset` as valid field axis


src/gt4py/next/common.py
```diff
@@ -52,10 +52,14 @@ def __str__(self):
...
-@dataclass(frozen=True)
+@dataclasses.dataclass(frozen=True)
 class Dimension:
     value: str
-    kind: DimensionKind = DimensionKind.HORIZONTAL
+    kind: DimensionKind = dataclasses.field(
+        default=DimensionKind.HORIZONTAL,
+        hash=False,
+        compare=False,  # TODO(havogt): include kind in hash and comparison, disabled because for itir dimension is just the tag
+    )
```
havogt, line 62: If you think this should be immediately removed, I'd propose to discuss if we should extend Iterator IR with this information...

src/gt4py/next/iterator/embedded.py
```diff
@@ -538,7 +538,11 @@ def execute_shift(
 ) -> MaybePosition:
     assert pos is not None
     if isinstance(tag, SparseTag):
-        current_entry = pos[tag]
+        # hack: relies on
+        # (a) Dimensions with same name are equal
+        # (b) sparse dimensions have the same name as the offset applying to them
+        sparse_dim = common.Dimension(tag)
```
havogt, line 544:
```suggestion
        # hack: relies on sparse dimensions have the same name as the offset applying to them
        sparse_dim = common.Dimension(tag, kind=common.DimensionKind.LOCAL)
```

havogt: We already have this weird constraint, because of integration with field view lowering...

src/gt4py/next/iterator/embedded.py
```diff
@@ -538,7 +538,9 @@ def execute_shift(
 ) -> MaybePosition:
     assert pos is not None
     if isinstance(tag, SparseTag):
-        current_entry = pos[tag]
+        # hack: sparse dimensions have the same name as the offset applying to them
```
havogt, line 541: we already have this constraint from lowering from field view...

tests/next_tests/integration_tests/feature_tests/ffront_tests/ffront_test_utils.py
```diff
@@ -36,7 +38,8 @@ def no_backend(program: itir.FencilDefinition, *args: Any, **kwargs: Any) -> Non
         gtfn_cpu.run_gtfn,
         gtfn_cpu.run_gtfn_imperative,
         dace_iterator.run_dace_iterator,
-    ]
+    ],
+    ids=lambda p: next_tests.get_processor_id(p),
```
havogt, line 42: same names as in iterator_tests...

src/gt4py/next/iterator/embedded.py
```diff
@@ -1284,7 +1280,7 @@ def _tuple_assign(field, value, indices):
             _tuple_assign(f, v, indices)
     else:
         assert is_located_field(field)
-        field[indices] = value
+        field.field_setitem(indices, value)
```
tehrengruber, line 1283:
```suggestion
        field.field_setitem(named_indices, value)
```
Those are `named_indices` now right? I would rename the variable then. How about the `_build_tuple_result` function above? Is the `__getitem__` access correct there?

havogt: yes, looks like there was a bug, not sure why we didn't run into it.

PR #1315, opened 2023-08-08
feat[next]: embedded.ndarray_field: field slicing and intersection

samkellerhals:
Enable slicing a `Field`.
- Absolute slicing with a `DomainSlice`, a sequence of `NamedRange` or `NamedIndex`.
- Relative slicing with tuple of `slice` or `int`s.

Enable binary operations on fields with different domain.


src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -254,3 +253,45 @@ class JaxArrayField(_BaseNdArrayField):
         array_ns: ClassVar[ModuleType] = jnp
 
     common.field.register(jnp.ndarray, JaxArrayField.from_array)
+
+
+def _slice_with_domain(field: common.Field, new_domain: common.Domain) -> np.ndarray:
```
havogt, line 258:
```suggestion
def _slice_with_named_ranges(field: common.Field, named_ranges: Sequence[NamedRange]) -> np.ndarray:
```
should already work

havogt, line 258:
```suggestion
def _slice_with_domain(field: common.Field, new_domain: common.Domain) -> tuple[slice| new_axis, ...]:
```
maybe like this to simplify testing

src/gt4py/next/common.py
```diff
@@ -201,6 +206,11 @@ def _broadcast_ranges(
     )
 
 
+NamedRange: TypeAlias = tuple[Dimension, UnitRange]
+NamedIndex: TypeAlias = tuple[Dimension, int]
+NamedIndicesOrRanges = Sequence[NamedRange | NamedIndex, ...]
```
havogt, line 211: move to line 140. `NamedRange` is already there

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -254,3 +253,48 @@ class JaxArrayField(_BaseNdArrayField):
...
+    slice_indices = []
+
+    for new_dim, new_rng in named_indices:
+
+        pos_new = named_indices.dims.index(new_dim)
+
+        if new_dim in old_domain.dims:
+            pos_old = old_domain.dims.index(new_dim)
+
+            if pos_new == pos_old + len(slice_indices):
```
havogt, line 290: what does this `if` do?
do you assume `len(field.domain) == len(named_indices)`? then this should always be true, no?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -254,3 +253,48 @@ class JaxArrayField(_BaseNdArrayField):
...
+                                      of the field. If a dimension is not present in the named indices or ranges, a None
+                                      element is used to indicate expanding the dimension along that axis.
+
+        Note:
+            - The function handles cases where named indices or ranges are provided for dimensions that do not exist in the
+              field's domain.
+            - Slices are generated according to the specified indices or ranges, and can be used to extract sub-arrays using
+              indexing.
+        """
+    old_domain = field.domain
```
havogt, line 280: my understanding is that this works for `len(field.domain) == len(named_indices)`, if that's the case let's add an assert.

havogt: And then in an next step, let's think about the following:
```
f: Field[[I,J],...]
f[(I, UnitRange(...),)]
```
is this legal? should it be?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -254,3 +253,48 @@ class JaxArrayField(_BaseNdArrayField):
...
+    for new_dim, new_rng in named_indices:
+
+        pos_new = named_indices.dims.index(new_dim)
+
+        if new_dim in old_domain.dims:
+            pos_old = old_domain.dims.index(new_dim)
+
+            if pos_new == pos_old + len(slice_indices):
+                slice_indices.append(
+                    slice(
```
havogt, line 292: this is not handling the `NamedIndices` case isn't it?

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -125,3 +131,54 @@ def fma(a: common.Field, b: common.Field, c: common.Field, /) -> common.Field:
 
     result = fma(field_inp_a, field_inp_b, field_inp_c)
     assert np.allclose(result.ndarray, expected)
+
+
+def test_get_slices_with_named_indices_1d_to_2d_missing_dim_right():
+    field_domain = common.Domain(dims=(IDim,), ranges=(UnitRange(0, 10),))
+    new_domain = common.Domain(dims=(IDim, JDim), ranges=(UnitRange(5, 10), UnitRange(5, 10)))
```
havogt, line 138: let's write these testcases with `Sequence[NamedRange]` and then maybe one case with `Domain` to illustrate `Domain` is a `Sequence[NamedRange]`.

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -125,3 +131,54 @@ def fma(a: common.Field, b: common.Field, c: common.Field, /) -> common.Field:
...
+    assert slices == (slice(0, 5, None), None, None)
+
+
+def test_get_slices_with_named_indices_3d_to_1d():
+    field_domain = common.Domain(dims=(IDim, JDim, KDim), ranges=(UnitRange(0, 10), UnitRange(0, 10), UnitRange(0, 10)))
+    new_domain = common.Domain(dims=(IDim,), ranges=(UnitRange(0, 10),))
+    field = common.field(np.ones((10, 10, 10)), domain=field_domain)
+    slices = _get_slices_with_named_indices(field, new_domain)
+    assert slices == (slice(0, 10, None),)
+
```
havogt, line 166: test cases for NamedIndex missing, something like:
```suggestion
def test_get_slices_with_named_indices_and_ranges:
    field_domain = common.Domain(dims=(IDim, JDim, KDim), ranges=(UnitRange(0, 10), UnitRange(0, 10), UnitRange(0, 10)))
    new_domain = ((IDim,UnitRange(0, 10),(IDim, 2), (KDim, 3)) # <--- indices used
    field = common.field(np.ones((10, 10, 10)), domain=field_domain)
    slices = _get_slices_with_named_indices(field, new_domain)
    assert slices == (slice(0, 10, None),)
```

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -125,3 +131,54 @@ def fma(a: common.Field, b: common.Field, c: common.Field, /) -> common.Field:
 
     result = fma(field_inp_a, field_inp_b, field_inp_c)
     assert np.allclose(result.ndarray, expected)
+
+
+def test_get_slices_with_named_indices_1d_to_2d_missing_dim_right():
+    field_domain = common.Domain(dims=(IDim,), ranges=(UnitRange(0, 10),))
+    new_domain = common.Domain(dims=(IDim, JDim), ranges=(UnitRange(5, 10), UnitRange(5, 10)))
+    field = common.field(np.ones((10,)), domain=field_domain)
```
havogt, line 139: you could go one step further and add a FieldMock, like

```python
@dataclasses.dataclass
class FieldMock:
    domain: Domain
```
then you don't have to allocate any array, which is unrelated to this test.

src/gt4py/next/embedded/nd_array_field.py
samkellerhals: I wrote it as a safeguard to ensure proper handling of dimensions, especially when dimensions are not in the expected sequence or there's a possibility of misalignment between named_indices and old_domain, but since they are broadcasted during the intersection I guess its not necessary (the tests also pass without this condition).

src/gt4py/next/embedded/nd_array_field.py
samkellerhals: No not yet

src/gt4py/next/embedded/nd_array_field.py
samkellerhals: Regarding this: 

```
f: Field[[I,J],...]
f[(I, UnitRange(...),)]
```

Currently I have implemented something similar as a test here: https://github.com/GridTools/gt4py/pull/1315/files#diff-7e004bcaf958ca74ad45bb6bfe6c0aaf90bc110f6c466923e5da97c54af16b20R160-R165

Currently it is legal to slice a field with a smaller domain and then it just returns that dimension and corresponding range, if it should not be legal we would have to add a check to check whether the named_indices are smaller than the `field.domain`

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -125,3 +131,54 @@ def fma(a: common.Field, b: common.Field, c: common.Field, /) -> common.Field:
...
+    slices = _get_slices_with_named_indices(field, new_domain)
+    assert slices == (slice(0, 5, None), None, None)
+
+
+def test_get_slices_with_named_indices_3d_to_1d():
+    field_domain = common.Domain(dims=(IDim, JDim, KDim), ranges=(UnitRange(0, 10), UnitRange(0, 10), UnitRange(0, 10)))
+    new_domain = common.Domain(dims=(IDim,), ranges=(UnitRange(0, 10),))
+    field = common.field(np.ones((10, 10, 10)), domain=field_domain)
+    slices = _get_slices_with_named_indices(field, new_domain)
+    assert slices == (slice(0, 10, None),)
```
havogt, line 165: This is wrong (or maybe correct by chance). What do you expect the result to be if you apply this slice to a 3D field. What happens if `new_domain` has JDim instead of IDim?

For safety I would always produce N-tuple for slicing when `shape(field)==N` to make sure we don't accidentally broadcast in the wrong dimension.

havogt: Ok, actually it's more complicated and maybe it makes sense to first write down what we expect to happen.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -254,3 +253,46 @@ class JaxArrayField(_BaseNdArrayField):
         array_ns: ClassVar[ModuleType] = jnp
 
     common.field.register(jnp.ndarray, JaxArrayField.from_array)
+
+
+def _get_slices_with_named_indices(field: common.Field, named_indices: NamedIndicesOrRanges) -> tuple[slice | None, ...]:
```
havogt, line 258: maybe even better to implement this function as 
```suggestion
def _get_slices_with_named_indices(field: Domain, named_indices: NamedIndicesOrRanges) -> tuple[slice | None, ...]:
```
because we only need domain from field, then it's even more generic (just use `None` for array_ns.newaxis with comment).

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -125,3 +131,68 @@ def fma(a: common.Field, b: common.Field, c: common.Field, /) -> common.Field:
 
     result = fma(field_inp_a, field_inp_b, field_inp_c)
     assert np.allclose(result.ndarray, expected)
+
+
+@dataclasses.dataclass
+class FieldMock:
+    domain: common.Domain
+    array_ns = np
```
havogt, line 139: sorry, I have another idea that doesn't require this mock, see signature proposal for _get_slices... function

src/gt4py/next/embedded/nd_array_field.py
samkellerhals: Now it is

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
samkellerhals: Not sure I understand why

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -231,9 +233,68 @@ def restrict(self: _BaseNdArrayField, domain) -> _BaseNdArrayField:
...
+    def _getitem_domain_slice(self, index: DomainSlice) -> common.Field:
+        slices = _get_slices_from_domain_slice(self.domain, index)
+        new = self.ndarray[slices]
+        return common.field(new, domain=index)
+
+    def _getitem_relative_slice(self, index: tuple[slice | int, ...]) -> common.Field:
+        new = self.ndarray[index]
+        new_dims = self.domain[:len(new.shape)].dims
+        new_ranges = [UnitRange(0, s) for s in new.shape]
+        new_domain = Domain(dims=new_dims, ranges=tuple(new_ranges))
```
havogt, line 289:
```suggestion
        # not included: ellipsis, missing dimension in tuple
        new_dims = []
        new_ranges = []
        for i, elem in enumerate(index):
            if isinstance(elem, slice):
                new_dims.append(self.domain.dims[i])
                new_ranges.append(slice_range(self.domain.ranges[i], elem))
        new_domain = Domain(dims=new_dims, ranges=tuple(new_ranges))
```
with 
```
def slice_range(range, slice):
    start = range.start + slice.start if slice.start >= 0 else range.stop + slice.start
    stop = range.start + slice.stop if slice.stop >= 0 else range.stop + slice.stop
    return UnitRange(start, stop)
```

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -194,6 +196,113 @@ def restrict(self: _BaseNdArrayField, domain) -> _BaseNdArrayField:
...
+            slice_stop = 0
+        else:
+            slice_stop = (
+                slice_obj.stop if slice_obj.stop >= 0 else input_range.stop + slice_obj.stop
+            )
+
+        start = input_range.start + slice_start
+        stop = input_range.start + slice_stop
+
+        return UnitRange(start, stop)
```
havogt, line 304: testcase:
input_range = UnitRange(2,10)
slice_obj = slice(2,-2)
expected = UnitRange(4,8)

result:
slice_start = 2
slice_stop = 10 + -2 = 8
start = 2 + 2 = 4 (correct)
stop = 2 + 8 = 10 (wrong)

if I executed the code correctly. If in doubt, test this function separately.

Maybe
```suggestion
        start = (input_range.start if (slice_obj.start or 0) >= 0 else input_range.stop) + slice_obj.start
        stop = (input_range.start if (slice_obj.stop or len(input_range)) >= 0 else input_range.stop) + slice_obj.stop
        return UnitRange(start, stop)
```

havogt (changes requested):
If I didn't make a mistake, I think there are still some algorithmic errors. I didn't go through the tests to check if there is a wrong expected result or a test missing.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -194,6 +197,129 @@ def restrict(self: _BaseNdArrayField, domain) -> _BaseNdArrayField:
...
+
+    def __getitem__(
+            self, index: DomainSlice | Sequence[common.NamedIndex] | tuple[int, ...]
+    ) -> common.Field | core_defs.DType[core_defs.ScalarT]:
+
+        if not isinstance(index, tuple):
+            index = (index,)
+
+        if isinstance(index[0], Domain):
+            index = index[0]
```
havogt, line 225:
```suggestion
        if not isinstance(index, tuple) and not is_domain_slice(index):
            index = (index,)
```

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -194,6 +211,107 @@ def restrict(self: _BaseNdArrayField, domain) -> _BaseNdArrayField:
...
+        for i, elem in enumerate(new_index):
+            if isinstance(elem, slice):
+                new_dims.append(self.domain.dims[i])
+                new_ranges.append(self._slice_range(self.domain.ranges[i], elem))
+            elif isinstance(elem, int):
+                new_dims.append(self.domain.dims[elem])
+                new_ranges.append(self.domain.ranges[elem])
+            elif isinstance(elem, type(Ellipsis)):
+                new_dims.append(self.domain.dims[i])
+                new_ranges.append(self.domain.ranges[i])
```
havogt, line 290: this is not working in case `len(new.shape) - len(index) < 0`, e.g.` f: Field[[I,J,K]]`, `f[1,:, 2]`.
In my branch I wrote an alternative algorithm, but you could already add test cases for this, and check if it really fails.

src/gt4py/next/embedded/nd_array_field.py
havogt: You marked this as resolved, but it still looks wrong to me. Did you test my example?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -194,6 +211,107 @@ def restrict(self: _BaseNdArrayField, domain) -> _BaseNdArrayField:
...
+
+        new_domain = Domain(dims=new_dims, ranges=tuple(new_ranges))
+        return common.field(new, domain=new_domain)
```
havogt, line 293: the case where new is a scalar is not handled

havogt: (add a test first)

src/gt4py/next/common.py
```diff
@@ -95,6 +94,11 @@ def __post_init__(self):
             object.__setattr__(self, "start", 0)
             object.__setattr__(self, "stop", 0)
 
+    @classmethod
+    @property
+    def infinity(cls) -> UnitRange:
```
havogt, line 99: might need a `#type: ignore[xxx]` because classmethod+property is a python 3.9 feature

havogt: maybe also make the `Infinity.positive/negative` a property?

src/gt4py/next/common.py
```diff
@@ -445,3 +458,20 @@ def promote_dims(*dims_list: Sequence[Dimension]) -> list[Dimension]:
...
+def is_named_range(v: Any) -> TypeGuard[NamedRange]:
+    return isinstance(v, tuple) and isinstance(v[0], Dimension) and isinstance(v[1], UnitRange)
+
+
+def is_named_index(v: Any) -> TypeGuard[NamedIndex]:
+    return isinstance(v, tuple) and isinstance(v[0], Dimension) and isinstance(v[1], int)
+
+
+def is_domain_slice(index: Any) -> TypeGuard[DomainSlice]:
+    return all(is_named_range(idx) or is_named_index(idx) for idx in index)
```
havogt, line 472:
```suggestion
    return isinstance(index, Sequence) and all(is_named_range(idx) or is_named_index(idx) for idx in index)
```

src/gt4py/next/common.py
```diff
@@ -445,3 +458,20 @@ def promote_dims(*dims_list: Sequence[Dimension]) -> list[Dimension]:
...
+
+
+def index_tuple_with_indices(target_tuple, indices_to_use):
```
havogt, line 475: unused now?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -194,6 +211,107 @@ def restrict(self: _BaseNdArrayField, domain) -> _BaseNdArrayField:
...
+    def __getitem__(self, index: tuple[slice | int, ...]) -> common.Field:
+        """Relative slicing with ordered dimension access."""
+        ...
+
+    @overload
+    def __getitem__(
+        self, index: Sequence[common.NamedIndex]
+    ) -> common.Field | core_defs.DType[core_defs.ScalarT]:
+        # Value in case len(i) == len(self.domain)
+        ...
```
havogt, line 229: this overload was only to emphasize the construction of lower dimensions, but not correct for mypy because the parameter is a subset of the first overload.
```suggestion
```

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -194,6 +211,107 @@ def restrict(self: _BaseNdArrayField, domain) -> _BaseNdArrayField:
...
 
+    @overload
+    def __getitem__(self, index: DomainSlice) -> common.Field:
+        """Absolute slicing with dimension names."""
+        ...
+
+    @overload
+    def __getitem__(self, index: tuple[slice | int, ...]) -> common.Field:
+        """Relative slicing with ordered dimension access."""
+        ...
```
havogt, line 222:
```suggestion
    @overload
    def __getitem__(self, index: DomainSlice) -> common.Field | core_defs.ScalarT:
        """Absolute slicing with dimension names."""
        ...

    @overload
    def __getitem__(self, index: tuple[slice | int, ...]) -> common.Field | core_defs.ScalarT:
        """Relative slicing with ordered dimension access."""
        ...
```

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -254,3 +371,87 @@ class JaxArrayField(_BaseNdArrayField):
...
+            domain_slice.append(np.newaxis)
+            new_domain_dims.append(dim)
+            new_domain_ranges.append(
+                UnitRange(common.Infinity.negative(), common.Infinity.positive())
+            )
+    return common.field(
+        field.ndarray[tuple(domain_slice)],
+        domain=Domain(tuple(new_domain_dims), tuple(new_domain_ranges)),
+    )
+
```
havogt, line 402:
```suggestion
_BaseNdArrayField.register_builtin_func(fbuiltins.broadcast, broadcast)
```
this registers the broadcast to be used as a field function. please change the broadcast tests to use the `fbuiltins.broadcast`.
Also test broadcasting a value to a field, this might require modifications to the broadcast function.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -254,3 +371,87 @@ class JaxArrayField(_BaseNdArrayField):
...
+
+def _find_index_of_dim(dim: Dimension, domain_slice: DomainSlice) -> Optional[int]:
+    for i, (d, _) in enumerate(domain_slice):
+        if dim == d:
+            return i
+    return None
+
+
+def broadcast(field: common.Field, new_dimensions: tuple[common.Dimension, ...]):
```
havogt, line 383:
```suggestion
def _broadcast(field: common.Field, new_dimensions: tuple[common.Dimension, ...]):
```
sucht that users use `fbuiltins.broadcast` and not accidentally this one

src/gt4py/next/embedded/nd_array_field.py
samkellerhals: Should be fixed now

src/gt4py/next/embedded/nd_array_field.py
samkellerhals: This should also be fixed now

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -194,6 +211,107 @@ def restrict(self: _BaseNdArrayField, domain) -> _BaseNdArrayField:
...
+        else:
+            new_index = index
+
+        for i, elem in enumerate(new_index):
+            if isinstance(elem, slice):
+                new_dims.append(self.domain.dims[i])
+                new_ranges.append(self._slice_range(self.domain.ranges[i], elem))
+            elif isinstance(elem, int):
+                new_dims.append(self.domain.dims[elem])
+                new_ranges.append(self.domain.ranges[elem])
```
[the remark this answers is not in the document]
samkellerhals: I ended up implementing a different solution since I need that branch for when I am just indexing with integers.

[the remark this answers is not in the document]
havogt: I don't understand how this could be used? when you index every dimension, it's just a value, if not, you skip the dimension because it's removed.

src/gt4py/next/embedded/nd_array_field.py
havogt: broadcasting scalars actually doesn't reach this function (and we wouldn't know which ndarray to construct)

havogt (approved):

PR #1316, opened 2023-08-10
feat[next]: add `where` to embedded field view

havogt:
- unifies unary and binary builtin in NdArrayField
- special case for `where` with tuples


src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -95,6 +95,8 @@ def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _R:
 
     def dispatch(self, *args: Any) -> Callable[_P, _R]:
         arg_types = tuple(type(arg) for arg in args)
+        if any(t == tuple for t in arg_types):
+            return self.function
```
havogt, line 99: obviously this is ugly, @tehrengruber do you want to discuss?

havogt: Till's temporary idea: 
try gt_builtin_func from each argument, if all fail, we do the default implementation

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -43,36 +43,31 @@
 from gt4py.next.ffront import fbuiltins
 
 
-def _make_unary_array_field_intrinsic_func(builtin_name: str, array_builtin_name: str) -> Callable:
-    def _builtin_unary_op(a: _BaseNdArrayField) -> common.Field:
-        xp = a.__class__.array_ns
+def _make_nary_intrinsic(builtin_name: str, array_builtin_name: str) -> Callable:
```
havogt, line 46:
```suggestion
def _make_intrinsic(builtin_name: str, array_builtin_name: str) -> Callable:
```

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -95,6 +95,9 @@ def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _R:
 
     def dispatch(self, *args: Any) -> Callable[_P, _R]:
         arg_types = tuple(type(arg) for arg in args)
+        if self.function == where.function and arg_types[1] == tuple and arg_types[2] == tuple:
+            # TODO(havogt): temporary workaround for `where(cond, tuple(...), tuple(...))`
```
havogt, line 99: In the end I didn't like this fall-back proposal, I would like to keep this special case and then revisit the pattern once we have all builtins implemented.

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -95,6 +95,9 @@ def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _R:
 
     def dispatch(self, *args: Any) -> Callable[_P, _R]:
```
havogt, line 96: other approach: subclass BuiltinFunction for `where` to handle the special case

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -133,6 +133,19 @@ def builtin_function(fun: Callable[_P, _R]) -> BuiltInFunction[_R, _P]:
     return BuiltInFunction(fun)
 
 
+class WhereBuiltinFunction(BuiltInFunction[_R, _P]):
+    def dispatch(self, *args: Any) -> Callable[_P, _R]:
+        arg_types = tuple(type(arg) for arg in args)
+        if arg_types[1] == tuple and arg_types[2] == tuple:
+            # TODO(havogt): temporary workaround for `where(cond, tuple(...), tuple(...))`
+            return self.function
+        return super().dispatch(*args)
```
tehrengruber, line 142:
```suggestion
class WhereBuiltinFunction(BuiltInFunction[_R, _P]):
    def __call__(self, *args: Any) -> Callable[_P, _R]:
      if isinstance(true_field, tuple) or isinstance(false_field, tuple):
          if not (isinstance(true_field, tuple) and isinstance(false_field, tuple)):
              raise ValueError(
                  f"Either both or none can be tuple in {true_field=} and {false_field=}."
              )
          if len(true_field) != len(false_field):
              raise ValueError("Tuple of different size not allowed.")
          return tuple(where(mask, t, f) for t, f in zip(true_field, false_field))
      return super()(*args)
```
Why not just have all the implementation here?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -40,40 +41,38 @@
     jnp: Optional[ModuleType] = None  # type:ignore[no-redef]
 
 
-def _make_unary_array_field_intrinsic_func(builtin_name: str, array_builtin_name: str) -> Callable:
-    def _builtin_unary_op(a: _BaseNdArrayField) -> common.Field:
-        xp = a.__class__.array_ns
+def _make_builtin(builtin_name: str, array_builtin_name: str) -> Callable:
+    def _builtin_op(*fields: common.Field) -> common.Field:
```
tehrengruber, line 45:
```suggestion
    def _builtin_op(*fields: common.Field | core_defs.SCALAR_TYPES) -> common.Field:
```
or something like
```suggestion
    def _builtin_op(self: common.Field, *fields: common.Field | core_defs.SCALAR_TYPES) -> common.Field:
```
Not sure if using self outside of a class is good practice, but afterall this will eventually be a method.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -40,40 +41,38 @@
...
-                new_data = op(a_broadcasted.ndarray[a_slices], b_broadcasted.ndarray[b_slices])
-                return a.__class__.from_array(new_data, domain=domain_intersection)
-            new_data = op(a.ndarray, xp.asarray(b.ndarray))
-        else:
-            assert isinstance(b, core_defs.SCALAR_TYPES)
-            new_data = op(a.ndarray, b)
+        transformed = []
+        domain_intersection = functools.reduce(
+            operator.and_,
+            [f.domain for f in fields if common.is_field(f)],
```
tehrengruber, line 54: Isn't every `f` a field anyway?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -40,40 +41,38 @@
...
-                b_broadcasted = _broadcast(b, domain_intersection.dims)
-                a_slices = _get_slices_from_domain_slice(a_broadcasted.domain, domain_intersection)
-                b_slices = _get_slices_from_domain_slice(b_broadcasted.domain, domain_intersection)
-                new_data = op(a_broadcasted.ndarray[a_slices], b_broadcasted.ndarray[b_slices])
-                return a.__class__.from_array(new_data, domain=domain_intersection)
-            new_data = op(a.ndarray, xp.asarray(b.ndarray))
-        else:
-            assert isinstance(b, core_defs.SCALAR_TYPES)
-            new_data = op(a.ndarray, b)
+        transformed = []
```
tehrengruber, line 51: Doesn't it make sense to move this in front of the loop where it is actually populated below? Type annotation would also be nice.

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -133,6 +133,31 @@ def builtin_function(fun: Callable[_P, _R]) -> BuiltInFunction[_R, _P]:
...
+FieldT = TypeVar("FieldT", bound=Union[Field, gt4py_defs.Scalar, Tuple])
+
+
+class WhereBuiltinFunction(
+    BuiltInFunction[_R, [MaskT, FieldT, FieldT]], Generic[_R, MaskT, FieldT]
+):
+    def __call__(self, mask: MaskT, true_field: FieldT, false_field: FieldT) -> _R:
+        if isinstance(true_field, tuple) or isinstance(false_field, tuple):
+            if not (isinstance(true_field, tuple) and isinstance(false_field, tuple)):
+                raise ValueError(
```
tehrengruber, line 146: I think we should have matching errors for embedded and toolchain. Let's discuss.

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -133,6 +133,31 @@ def builtin_function(fun: Callable[_P, _R]) -> BuiltInFunction[_R, _P]:
...
+            if len(true_field) != len(false_field):
+                raise ValueError("Tuple of different size not allowed.")
+            return tuple(where(mask, t, f) for t, f in zip(true_field, false_field))  # type: ignore[return-value] # `tuple` is not `_R`
+        return super().__call__(mask, true_field, false_field)
+
+
+def where_builtin_function(
+    fun: Callable[[MaskT, FieldT, FieldT], _R]
+) -> WhereBuiltinFunction[_R, MaskT, FieldT]:
+    return WhereBuiltinFunction(fun)
```
tehrengruber, line 158: Fine, but doesn't feel very pretty.

tehrengruber (changes requested):

src/gt4py/next/embedded/nd_array_field.py
havogt: not in call cases they will be methods

src/gt4py/next/ffront/fbuiltins.py
havogt: agree

src/gt4py/next/ffront/fbuiltins.py
havogt: Agree, however if we enable parsing (which is current default we will not reach to this point). Should I add a TODO?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -39,40 +41,38 @@
...
-            assert isinstance(b, core_defs.SCALAR_TYPES)
-            new_data = op(a.ndarray, b)
-
-        return a.__class__.from_array(new_data, domain=a.domain)
-
-    _builtin_binary_op.__name__ = builtin_name
-    return _builtin_binary_op
+        domain_intersection = functools.reduce(
+            operator.and_,
+            [f.domain for f in fields if common.is_field(f)],
```
tehrengruber, line 53:
```suggestion
            [f.domain for f in fields if not is_scalar(f)],
```
I see, intuitively I would say it should ignore scalars then and not just take everything that is a field. Both for readability and robustness.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -39,40 +41,38 @@
...
+            common.Domain(dims=tuple(), ranges=tuple()),
+        )
+        transformed: list[core_defs.NDArrayObject | core_defs.Scalar] = []
+        for f in fields:
+            if common.is_field(f):
+                if f.domain == domain_intersection:
+                    transformed.append(xp.asarray(f.ndarray))
+                else:
+                    f_broadcasted = _broadcast(f, domain_intersection.dims)
```
tehrengruber, line 62: It looks like the broadcasted array is materialized here. That seems unnecessary overhead. Can we at least add a todo to clean this up?

src/gt4py/next/ffront/fbuiltins.py
tehrengruber: TODO is fine for now I think.

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -98,6 +98,37 @@ def test_math_function_builtins(builtin_name: str, inputs, nd_array_implementati
...
+    false_ = np.asarray([3.0, 4.0], dtype=np.float32)
+
+    field_inputs = [_make_field(inp, nd_array_implementation) for inp in [cond, true_, false_]]
+    expected = np.where(cond, true_, false_)
+
+    result = fbuiltins.where(*field_inputs)
+    assert np.allclose(result.ndarray, expected)
+
+
+def test_where_builtin_with_tuple(nd_array_implementation):
```
tehrengruber, line 113: I think a test where the true and false branches are on different domains (different sizes to test intersection, and different dimensions to test the broadcast) is needed.

tehrengruber (changes requested):

src/gt4py/next/embedded/nd_array_field.py
havogt: Not sure, typing says `f` is Field or Scalar, and of the 2 we select the one which has `.domain`, i.e. `is_field`.

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -132,6 +132,31 @@ def builtin_function(fun: Callable[_P, _R]) -> BuiltInFunction[_R, _P]:
...
```
egparedes, line 157: Is this function really needed? Wouldn't work exactly the same using the `WhereBuiltinFunction` class as a decorator (which would call the constructor anyway)?

src/gt4py/next/embedded/nd_array_field.py
havogt: How would you do that? Maybe I miss something obvious.

src/gt4py/next/ffront/fbuiltins.py
havogt: right...

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
havogt: good point, the pr is older than that feature 🙃

src/gt4py/next/embedded/nd_array_field.py
tehrengruber: Using `np.newaxis` which is apparently already done in `_broadcast`... I'm a little skeptical that `f_broadcasted.ndarray[f_slices]` works properly in all cases, but we'll see when you write the tests :-P

src/gt4py/next/embedded/nd_array_field.py
tehrengruber: That is technically true (and a valid viewpoint). My perspective is the following: The resulting domain is the intersection of all input domains. For scalars then there is a (conceptual) broadcast to a zero-dimensional field. So for me the domain intersection is everything, but scalars. When I read this I intermediately thought why do only fields participate in the computation of the domain (and not all inputs) and my answer was because they are not scalars. There is an assert below that ensures the elements are either a field or a scalar so I am fine with this.

tehrengruber (approved):

PR #1322, opened 2023-08-16
OUTDATED - [feature][next] GPU backend from Python

petiaccja:
THIS PR HAS BEEN SUPERSEDED BY THIS:  https://github.com/GridTools/gt4py/pull/1325
I'LL CLOSE THIS PR AS SOON AS THE DISCUSSION ARE RESOLVED.


src/gt4py/next/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -39,6 +40,11 @@ def get_param_description(name: str, obj: Any) -> interface.Parameter:
     return interface.Parameter(name, type_translation.from_value(obj))
 
 
+class HardwareAccelerator(enum.IntEnum):
+    CPU = enum.auto()
+    GPU = enum.auto()
```
havogt, line 45: In the meantime there is a `DeviceKind` enum in `gt4py._core.definitions`. Please check if it makes sense to use it here.

src/gt4py/next/program_processors/runners/gtfn_cpu.py
```diff
@@ -20,9 +20,9 @@
 from gt4py.eve.utils import content_hash
```
havogt, line 20: let's rename the module to `gtfn`

src/gt4py/next/otf/compilation/build_systems/cmake_lists.py
```diff
@@ -30,22 +30,31 @@ class LinkDependency(eve.Node):
...
     CMakeListsFile = as_jinja(
         """
-        project({{project_name}})
         cmake_minimum_required(VERSION 3.20.0)
+        
+        project({{project_name}})
 
         # Languages
-        enable_language(CXX)
+        if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
```
havogt, line 54: we could default to the arch of device 0 as we do in cartesian, see https://github.com/petiaccja/gt4py/blob/ecdba0124a0b834e56c22fb8b75a96c2488b8325/src/gt4py/cartesian/backend/pyext_builder.py#L40

havogt: Did you check if `CMAKE_CUDA_ARCHITECTURES` does the right thing at compilation? I never used it (we never upgraded gridtools' cmake to work with it, because we just went through a big cmake refactoring when that feature was introduced).

havogt (commented):
not a full review, just a few comments and questions.

src/gt4py/next/program_processors/runners/gtfn_cpu.py
petiaccja: I wanted to, but `next.program_processors.runners.gtfn` clashes with `next.program_processors.codegens.gtfn`, so we'll have to find a different name.

petiaccja:
PLEASE DON'T ADD ANY MORE COMMENTS TO THIS PR, GO HERE INSTEAD: https://github.com/GridTools/gt4py/pull/1325

src/gt4py/next/otf/compilation/build_systems/cmake_lists.py
petiaccja: It sets the flags for `nvcc`: `--generate-code=arch=compute_60,code=[compute_60,sm_60]`. That looks correct to me.

petiaccja: I'd scan the node for CUDA-compatible devices and either pick the oldest architecture or simply generate the binary for every specific architecture. The problem is, we don't really have infrastructure for specifying command line options to `cmake`, and it would be a mistake to implement this so low in the call stack.

havogt:
moved all comments to the new PR.

PR #1323, opened 2023-08-16
bug[next]: Improve error message on invalid call to field operator and program

tehrengruber:
After #1275 most of the error message given to the user when calling a field operator or program with invalid arguments was only available in verbose mode. This PR shows this information again.

```python
@field_operator
def foo(x: IField):
    return x

@field_operator
def testee(a: IField, b: IField, c: IField) -> IField:
    return foo(1)
```

```
gt4py.next.errors.exceptions.DSLError: Invalid argument types in call to `foo`.
E           Invalid call to function of type `FieldOperatorType(definition=FunctionType(pos_only_args=[], pos_or_kw_args={'x': FieldType(dims=[Dimension(value='IDim', kind=<DimensionKind.HORIZONTAL: 'horizontal'>)], dtype=ScalarType(kind=<ScalarKind.INT32: 32>, shape=None))}, kw_only_args={}, returns=FieldType(dims=[Dimension(value='IDim', kind=<DimensionKind.HORIZONTAL: 'horizontal'>)], dtype=ScalarType(kind=<ScalarKind.INT32: 32>, shape=None))))`:
E             - Expected argument `x` to be of type `Field[[IDim], int32]`, but got `int32`.
E             File ".../gt4py_functional/tests/next_tests/integration_tests/feature_tests/ffront_tests/test_arg_call_interface.py", line 113
E                       return foo(1)
```


egparedes (approved):
Approving because it's a very small fix with potentially high quality of life improvement for some users in the short term. Most likely this should be revisited according to the recent work on standardizing our conventions to communicate errors to users. Mentioning @DropD here just to make sure he is aware of the change in this PR.

PR #1324, opened 2023-08-16
deps: Update GridTools C++ dependency to 2.3.1

havogt:


src/gt4py/next/otf/workflow.py
```diff
@@ -82,7 +82,7 @@ def replace(self, **kwargs: Any) -> Self:
         if not dataclasses.is_dataclass(self):
             raise TypeError(f"{self.__class__} is not a dataclass")
         assert not isinstance(self, type)
-        return dataclasses.replace(self, **kwargs)
```
havogt, line 85: this is the only manual change, mypy doesn't understand anymore that self is a dataclass (as checked in line 82)

petiaccja (approved):

PR #1332, opened 2023-09-05
feat[next]: extend DaCe support of reduction operator

edopao:
## Description

The DaCe support in baseline of reduction operator is limited to builtin-in math operator as reduction function.
The goal of this PR to support more complex lambda functions and also addressing of neighbors in external input fields.

## Requirements


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -625,47 +724,156 @@ def _visit_indirect_addressing(self, node: itir.FunCall) -> IteratorExpr:
...
+                inputs=input_mapping,
+                outputs=output_mapping,
+                symbol_mapping=symbol_mapping,
+                input_nodes={arg.value.data: arg.value for arg in args},
+                output_nodes={result_name: result_access},
+            )
+
+            # map fusion is beneficial for reduction stencils:
+            # the solution was designed to keep code generation simple and let DaCe fuse maps
+            self.context.body.apply_transformations_repeated([MapFusion], validate=False)
```
petiaccja, line 874: I don't think it's a good idea to put any optimization/canonicalization transforms in here unless you deliberately generate IR that doesn't meet post-conditions and you fix that with already existing passes.

petiaccja (changes requested):

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -56,6 +60,21 @@ def itir_type_as_dace_type(type_: next_typing.Type):
     raise NotImplementedError()
 
 
+def reduction_init_value(op_name_: str, type_: Any):
```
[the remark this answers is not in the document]
edopao: No, we do not need. The `init` value is actually the identity value  associated with the reduction operation. I am going to rename this function to `get_reduce_identity_value`. In the case where the field is used as in/out parameter, so there is an initial value, the ITIR explicitly combines the reduction operation with an update of the field.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -135,6 +154,21 @@ class Context:
...
+        self,
+        body: dace.SDFG,
+        state: dace.SDFGState,
+        symbol_map: dict[str, IteratorExpr | ValueExpr | SymbolExpr],
+    ):
+        self.body = body
+        self.state = state
+        self.symbol_map = symbol_map
+        self.reduce_limit = 0
+        self.reduce_wcr = None
```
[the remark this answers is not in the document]
edopao: Correct, I have added the comment on line 157.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: I think this is the case, as I wrote in the comment. Consider this reduction SDFG without applying `MapFusion`:
image

After applying `MapFusion` we get a simplified SDFG:
image

petiaccja: Hmm, so what I meant is for example if you already have a compiler pass that converts AVX instruction to SSE, it may be reasonable to write only the AVX vectorization pass, and that let's you target both ISAs together with the conversion pass. In this case, when you vectorize with SSE, emitting AVX instructions violates the post conditions, so you shouldn't do that, so you first vectorize with AVX, and then call the AVX to SSE conversion.

In your case, you first render the reductions, and then call map fusion. My question is, if you don't call map fusion, will your output have invalid "instructions" (like AVX)? If not, like I think is the case for you, it's probably not a wise idea to run optimization passes. Instead, you can just leave the generated (but still valid) IR as unoptimized as you want, and then once you're completely done with code generation you can run optimization passes over the entire IR, in a well-thought out order, with access to global information.

edopao: I get your point now, to prefer optimization with global knowledge, rather than doing it early inside the nested-SDFG. It is a good point. My intention was to run this transformation here as a cleanup pass rather than an optimization pass. In addition to the example above, there is another use case where we always generate redundant maps, that is deferencing with the neighbor index (`test_external_local_field`). So it is 2 known cases of redundant SDFG code.
In principle,  I would like not to generate redundant maps but this is the result of ITIR representation and traversal of reduction.

edopao: I can remove the `MapFusion` transformation. As you say, the SDFG is valid in any case and produces functionally correct code. We'll postpone it to a future optimization phase with global knowledge.

petiaccja: You're only running it localized to the part of the SDFG that's generated for the reduction, and not the whole SDFG, right? I think in that case it's okay to run the fusion, but can you perhaps comment that it's for cosmetic purposes, so that both the SDFG generating code is simple and the generated SDFG is simple.

petiaccja (approved):

PR #1334, opened 2023-09-07
fix[cartesian]: use distutils from setuptools instead of the standard library

egparedes:
This PR fixes the imports in the module used to compile python extensions to use the `distutils` versions packaged into `setuptools` instead of the version from the standard library, which is deprecated and causes some warnings at compilation which can break other tools (e.g.  spack).


egparedes:
cscs-ci run

egparedes:
cscs-ci run default

havogt (approved):

PR #1336, opened 2023-09-08
test[next]: check for DaCe dependency in test execution

edopao:
## Description
This code change will check if the DaCe module is available in the python environment. If not available, the DaCe backend is skipped in test execution.

## Requirements
N/A


tests/next_tests/__init__.py
```diff
@@ -11,6 +11,77 @@
...
+import enum
+
+import pytest
+
+
+# Skip definitions
+class SkipMark(enum.Enum):
+    XFAIL = pytest.xfail
+    SKIP = pytest.skip
+
```
egparedes, line 23: I don't think the enum is needed anymore, since we directly use the pytest functions here
```suggestion
XFAIL = pytest.xfail
SKIP = pytest.skip
```

tests/next_tests/__init__.py
```diff
@@ -11,6 +11,77 @@
...
+USES_ORIGIN = "uses_origin"
+USES_REDUCTION_OVER_LIFT_EXPRESSIONS = "uses_reduction_over_lift_expressions"
+USES_SCAN_IN_FIELD_OPERATOR = "uses_scan_in_field_operator"
+USES_SPARSE_FIELDS = "uses_sparse_fields"
+USES_STRIDED_NEIGHBOR_OFFSET = "uses_strided_neighbor_offset"
+USES_TUPLE_ARGS = "uses_tuple_args"
+USES_TUPLE_RETURNS = "uses_tuple_returns"
+USES_ZERO_DIMENSIONAL_FIELDS = "uses_zero_dimensional_fields"
+
+# Skip matrix
```
egparedes, line 54: Just a question: what do you think about moving all those definitions to a different module (e.g. `exclusion_matrices.py`) to make them more visible to other developers? Would it make sense for you?

tests/next_tests/integration_tests/feature_tests/ffront_tests/ffront_test_utils.py
```diff
@@ -32,20 +41,33 @@ def no_backend(program: itir.FencilDefinition, *args: Any, **kwargs: Any) -> Non
...
+    ]
+    + OPTIONAL_PROCESSORS,
     ids=lambda p: next_tests.get_processor_id(p),
 )
 def fieldview_backend(request):
+    backend = request.param
+    backend_id = next_tests.get_processor_id(backend)
+
+    if skip_cases := next_tests.BACKEND_SKIP_TEST_MATRIX.get(backend_id, []):
+        for marker, skip_mark, msg in skip_cases:
```
egparedes, line 64: I realized that the `if` is just redundant here:
```suggestion
    for marker, skip_mark, msg in next_tests.BACKEND_SKIP_TEST_MATRIX.get(backend_id, []):
```

tests/next_tests/unit_tests/conftest.py
```diff
@@ -73,19 +89,20 @@ def pretty_format_and_check(root: itir.FencilDefinition, *args, **kwargs) -> str
...
+    + OPTIONAL_PROCESSORS,
     ids=lambda p: next_tests.get_processor_id(p[0]),
 )
 def program_processor(request):
-    return request.param
+    backend, _ = request.param
+    backend_id = next_tests.get_processor_id(backend)
 
+    if skip_cases := next_tests.BACKEND_SKIP_TEST_MATRIX.get(backend_id, []):
+        for marker, skip_mark, msg in skip_cases:
```
egparedes, line 101: Remove the redundant `if`
```suggestion
    for marker, skip_mark, msg in next_tests.BACKEND_SKIP_TEST_MATRIX.get(backend_id, []):
```

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -12,9 +12,18 @@
...
 
+try:
+    import dace
+except ModuleNotFoundError:
+    import sys
+
+    if "pytest" in sys.modules:
+        raise
+    print("Error, module dace not found", file=sys.stderr)
+    exit()
```
egparedes, line 23: Why is the `import dace` needed here? The importer of this package (either a test module or another part of gt4py) should decide how deal with the import exception on its own, right?

tests/next_tests/unit_tests/conftest.py
```diff
@@ -23,12 +24,22 @@
...
-    dace_iterator,
-    double_roundtrip,
-    gtfn_cpu,
-    roundtrip,
-)
+from gt4py.next.program_processors.runners import double_roundtrip, gtfn_cpu, roundtrip
+
+
+try:
+    import dace
```
egparedes, line 31: Is this explicit dace import really needed? The `import dace` inside `dace_iterator` should produce the same exception if it is not installed, right?

tests/next_tests/unit_tests/conftest.py
```diff
@@ -23,12 +24,22 @@
...
+    # import dace_iterator only if dace is available
+    from gt4py.next.program_processors.runners import dace_iterator
+except ModuleNotFoundError as e:
+    if "dace" in str(e):
+        dace_iterator = None
+    else:
+        raise e
+finally:
+    if dace_iterator:
+        del dace
```
egparedes, line 42: If the explicit `import dace` above is removed, this part should also be removed

egparedes (commented):
I did a very quick review but it looks almost ready. Have some comments and questions (mostly about code style).

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
edopao: It seems that the Spack tests are executing this command:
`python3.10 -c 'import gt4py.next.program_processors.runners.dace_iterator'`
I received the following error report from @abishekg7 for Spack build, before I applied the change in this PR:
```
==> Testing package py-gt4py-test_dace-bab6rui
==> [2023-09-12-11:21:09.665161] Running install-time tests
==> [2023-09-12-11:21:09.665311] ('RUN-TESTS: install-time tests [test]',)
==> Error: TestFailure: 1 tests failed.


Command exited with status 1:
    '/project/g110/spack/upstream/daint_v0.18.1.7/icon-dsl/python-3.10.4/gcc-9.3.0/3enasifkvryqtb57gwbaodoxbyobiloa/bin/python3.10' '-c' 'import gt4py.next.program_processors.runners.dace_iterator'
  File "<string>", line 1, in <module>
  File "/scratch/snx3000/agopal/spack_test/test/spack/opt/spack/py-gt4py-test_dace/gcc-9.3.0/bab6ruizwdfnhttn4oyq5jwgh4vzrkhq/lib/python3.10/site-packages/gt4py/next/program_processors/runners/dace_iterator/__init__.py", line 17, in <module>
    import dace
ModuleNotFoundError: No module named 'dace'
```

tests/next_tests/integration_tests/feature_tests/ffront_tests/ffront_test_utils.py
edopao: Yes, I realised it too.. I thought it was to keep the line shorter.

tests/next_tests/__init__.py
edopao: Moving these definitions to a dedicated place is definitely more clear.

tests/next_tests/unit_tests/conftest.py
edopao: This error disappears after deleting the Python venv and creating a new one.

edopao:
cscs-ci run

edopao:
cscs-ci run

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
edopao: I have tried the Spack build locally with this PR and I observed that the change discussed here is not needed. The PR seems to work fine without this change, so I am going to remove it in next commit.

tests/next_tests/integration_tests/multi_feature_tests/iterator_tests/test_fvm_nabla.py
```diff
@@ -396,17 +370,11 @@ def nabla_sign(n_nodes, out_MXX, out_MYY, pp, S_MXX, S_MYY, vol, node_index, is_
 
 
 @pytest.mark.requires_atlas
```
edopao, line 372: @egparedes Can you please explain why we need the marker `requires_atlas` only on this test case? Should we add the `uses_atlas_tables` marker to any other backend, in addition to those 4 in baseline code?

egparedes: I think it is just an oversight. In principle, all tests in the file should need the `requires_atlas` mark to skip them if atlas is not installed but the `importorskip(atlas4py)` statement at the beginning of the file does basically the same automatically, and that's why the tests in this file didn't failed. `importorskip` is useful if we really want to skip ALL the tests from a file if a dependency is missing, but the custom markers give us finer grain control.

BTW, In this case it looks like `uses_atlas_tables` and `requires_atlas` are currently equivalent, right? Maybe we should delete one of them.

egparedes: Good catch, by the way! :facepalm:

edopao:
@egparedes I have added the ADR and removed the duplicated fixture for ATLAS tables. Also tested with Spack on Daint.

egparedes (approved):
It looks very good. Thanks for adding all the new pytest markers.

PR #1337, opened 2023-09-08
feat[next]: DaCe support for floordiv

edopao:
## Description
This is a small PR that adds support in DaCe backend for `floordiv` math built-in function.
It also enables some tests for math built-in execution on DaCe backend.


petiaccja (approved):
Pretty straightforward, I think it looks good!

PR #1338, opened 2023-09-11
feat[next]: Enable field origin in DaCe backend

edopao:
## Description

Small change in DaCe backend to enable setting field origin other than default 0.


edopao:
This is an optional feature for DaCe backend, not currently needed. Its implementation is more complex than expected.

PR #1343, opened 2023-09-29
feat[next]: DaCe support for tuple returns

edopao:
## Description

This PR adds partial support for stencils with tuple returns. The motivation is that tuple returns are used in icon4py stencils, although the internal expressions do not operate on tuples. Tuples are just a mean to aggregate multiple-outputs from one stencil. This is why this PR does not contain support for scan or conditional expressions with tuples.


edopao:
cscs-ci run

edopao:
cscs-ci run

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -124,6 +125,17 @@ def add_storage(self, sdfg: dace.SDFG, name: str, type_: ts.TypeSpec, has_offset
...
+    def get_output_nodes(
+        self, closure: itir.StencilClosure, context: Context
+    ) -> dict[str, dace.nodes.AccessNode]:
+        if isinstance(closure.output, itir.SymRef):
+            name = str(closure.output.id)
+            return {name: context.state.add_access(name)}
+        else:
+            translator = PythonTaskletCodegen(self.offset_provider, context, self.node_types)
+            output_nodes = flatten_list(translator.visit(closure.output))
+            return {node.value.data: node.value for node in output_nodes}
```
petiaccja, line 137: Can you extract common code from the two branches? Why not just `translator.visit` the output anyways, even if it's just a single one? Does `flatten_list` also handle single outputs? If so, you can maybe eliminate the `if` statement completely. I'm not saying you should at all costs, but if you don't have to change any of the rest of the code, you might as well.

petiaccja (changes requested):

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -423,32 +424,42 @@ def visit_Lambda(
...
-        self.context.body.arrays[result.value.data].transient = False
-        self.context = prev_context
+        results: list[ValueExpr] = []
+        """
+        Ideally, we would like the lambdas to be single-output expressions, which implies only one child node.
+        This would simplify SDFG-generation. For this, we need a transformation pass in ITIR preprocessor
+        to unpack tuple expressions into multiple single-output expressions.
+        The same applies to scan and conditional expressions with tuples. However, for lambda expressions
+        it is easy to unpack tuples in the for-loop below, after flattening the nested list of expressions.
+        Consider removing this loop once the ITIR transformation loop is in place.
```
petiaccja, line 434: I think the conclusion yesterday was that multiple-input-multiple-output lambdas are okay, we just don't want nested tuples as output. Can you perhaps update the comment to avoid the term "single-output"?

petiaccja (approved):
I think it's good!

PR #1344, opened 2023-09-29
feat[next]: DaCe support for neighbor strided offset

edopao:
## Description

This PR adds support for neighbor strided offset in DaCe backend, another ITIR feature needed in order to support icon4py stencils.

## Requirements


petiaccja (commented):
To do: choose how to represent max_neighbors:
- Runtime: extract from `StridedNeighborOffsetProvider` instance and pass it as an SDFG symbol (simple and independent of caching SDFGs)
- Compile time: extract from `StridedNeighborOffsetProvider` instance and hardcode it in the SDFG (error prone when caching SDFGs)
- Compile time: create and infer a `StridedNeighborOffsetProviderType` from the instance and hardcode from the `Type` (lots of work but great for caching SDFGs)

edopao:
Decided to go for option 2 to represent `max_neighbors`:
- Compile time: extract from `StridedNeighborOffsetProvider` instance and hardcode it in the SDFG (error prone when caching SDFGs)

The reason is that this option provides several advantages:
- It is compatible with the long-term more desirable solution to create and infer a `StridedNeighborOffsetProviderType`
- It enables compile-time optimizations
- This solution is already used to select the scan dimension (`column_axis`)

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -111,12 +111,35 @@ def get_stride_args(
...
+    arg_types: Sequence[ts.TypeSpec],
+    column_axis: Optional[Dimension],
+    offset_provider: Mapping[str, Any],
+) -> str:
+    max_neighbors = [
+        (k, v.max_neighbors)
+        if isinstance(v, (NeighborTableOffsetProvider, StridedNeighborOffsetProvider))
+        else (k, -1)
+        for k, v in offset_provider.items()
+    ]
```
petiaccja, line 129: Just drop the ones that don't have any compile-time properties. The ones with only runtime properties should manifest in the SDFG structure, that should be enough, right? Let me know if I'm missing something here.
```suggestion
    max_neighbors = [
        (k, v.max_neighbors)
        for k, v in offset_provider.items()
        if isinstance(v, (NeighborTableOffsetProvider, StridedNeighborOffsetProvider))
    ]
```

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -111,12 +111,35 @@ def get_stride_args(
...
+            program,
+            *arg_types,
+            column_axis,
+            *max_neighbors,
+        )
+    ]
+    m = hashlib.sha256()
+    for s in cache_id_args:
+        m.update(s.encode())
+    return m.hexdigest()
```
petiaccja, line 142: Looks good. I hope it runs faster than the SDFG generation :P

petiaccja (changes requested):
Just a small thing with your caching.

petiaccja (approved):

PR #1346, opened 2023-10-06
fix[next-dace]: Preserve ITIR data layout

edopao:
## Description

The DaCe backend is reordering the dimensions of field domain based on alphabetical order - we call this the _canonical order_ of dimensions. This PR corrects indexing of field domain in `get_scan_dim` which was not consistent with the canonical order.

Additional minor edit:
- rename `map_domain` -> `map_ranges`
- replace `dace.Memlet()` with `dace.Memlet.simple()`

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
@@ -31,11 +31,7 @@ from .utility import connectivity_identifier, filter_neighbor_tables
 
 def convert_arg(arg: Any):
     if common.is_field(arg):
-        sorted_dims = sorted(enumerate(arg.__gt_dims__), key=lambda v: v[1].value)
-        ndim = len(sorted_dims)
-        dim_indices = [dim[0] for dim in sorted_dims]
-        assert isinstance(arg.ndarray, np.ndarray)
-        return np.moveaxis(arg.ndarray, range(ndim), dim_indices)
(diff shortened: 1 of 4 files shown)


petiaccja (commented):
Choose dimension ordering and ABI scheme:
- Canonical ordering + no runtime casts: order dimensions lexicographically , shuffle strides, sizes and offsets consistently, everywhere (kinda error prone but simple and no runtime costs)
- Dynamic ordering + runtime casts: generate each callable with whatever dimension order in its ABI, and generate runtime instructions for casting differently ordered arguments to the callable (kinda error prone, kinda complicated, kinda slow, but reuses function bodies)
- Dynamic ordering + no runtime casts: generate each callable once for each different dimension order of its arguments (simple implementation, complicated caching for SDFG generation, not error prone, bloated generated code)

edopao:
Decided to go for option 1, which means keeping the solution already present in main branch:
- Canonical ordering + no runtime casts: order dimensions lexicographically , shuffle strides, sizes and offsets consistently, everywhere (kinda error prone but simple and no runtime costs)

However, a few places need to be updated because they were not consistent:
1. Addressing of output field based on closure domain: the dimensions in the closure domain are not alphabetically sorted, because sorting the closure domain would imply reordering the loops.
2. Addressing of scan dimension, which was using the ITIR order of dimensions in the output field domain.

petiaccja (approved):

PR #1347, opened 2023-10-10
feat[next-dace]: Add support for GPU execution

edopao:
## Description

This PR adds support for GPU execution in DaCe Backend. Additionally, it also introduces a build cache for each visited ITIR program and corresponding binary DaCe program.

N.B. The build cache is an early solution with very limited functionality. For example, there is no check on max entries and the hash function on ITIR program might need to include additional parameters such as DaCe symbols table. It fits the initial need of prototype benchmarking.


src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -85,44 +95,129 @@ def get_stride_args(
...
     }
+
+    if run_on_gpu:
+        import cupy as cp
+
+        # copy input data to GPU memory
+        for label, array in sdfg.arrays.items():
+            if not array.transient:
+                host_array = expected_args[label]
+                expected_args[label] = cp.array(host_array)
```
petiaccja, line 188: If the user passes `run_on_gpu=True`, then you can expect `args` to contain CUDA arrays. No need to copy the array into GPU memory.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -85,44 +95,129 @@ def get_stride_args(
...
         dace.config.Config.set("frontend", "check_args", value=True)
-        sdfg(**expected_args)
+        sdfg_program(**expected_args)
+
+    if run_on_gpu:
+        # copy result to host memory
+        for k, v in expected_args.items():
+            if k in dace_field_args:
+                np.copyto(dace_field_args[k], v.get())
+
```
petiaccja, line 200: This copy is also not needed, the output array is expected to be already on the GPU as well.

petiaccja (changes requested):
Remove the host<->device copying of array, otherwise good.

petiaccja (approved):

PR #1351, opened 2023-10-19
feat[next][dace]: DaCe support for temporaries

edopao:
## Description

Lift operators are removed by an ITIR transformation by introducing temporaries. Temporaries are implemented in DaCe backend as transient arrays. This PR adds extraction of temporaries and generation of corresponding transient arrays in the SDFG representation.


src/gt4py/next/program_processors/runners/dace_iterator/utility.py
```diff
@@ -35,6 +35,20 @@ def as_dace_type(type_: ts.ScalarType):
...
+        return ts.ScalarType(kind=ts.ScalarKind.BOOL)
+    elif dtype == "float32":
+        return ts.ScalarType(kind=ts.ScalarKind.FLOAT32)
+    elif dtype == "float64":
+        return ts.ScalarType(kind=ts.ScalarKind.FLOAT64)
+    elif dtype == "int32":
+        return ts.ScalarType(kind=ts.ScalarKind.INT32)
+    elif dtype == "int64":
+        return ts.ScalarType(kind=ts.ScalarKind.INT64)
+    raise ValueError(f"data type {dtype} not supported")
```
tehrengruber, line 49:
```suggestion
    try:
      return getattr(ts.ScalarKind, dtype.upper())
    except AttributeError:
          raise ValueError(f"Data type {dtype} not supported.")
```

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -628,3 +754,33 @@ def _check_shift_offsets_are_literals(node: itir.StencilClosure):
...
+        def __replace_in_expression(fun: itir.FunCall, p_old: str, p_new: str):
+            for arg in fun.args:
+                if isinstance(arg, itir.FunCall):
+                    __replace_in_expression(arg, p_old, p_new)
+                elif isinstance(arg, itir.SymRef) and arg.id == p_old:
+                    arg.id = eve.SymbolRef(p_new)
+
+        if isinstance(closure.stencil, itir.Lambda):
+            for p in closure.stencil.params:
+                p_new = ""
```
tehrengruber, line 777: It think this will give a collision if both `z_gammaᐞ0ᐞ3` and `z_gamma7_0` are a parameter name.

tehrengruber (commented):
Drive-by review.

src/gt4py/next/program_processors/runners/dace_iterator/utility.py
edopao: Thank you, nice support for reflection in Python language!

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
edopao: The for loop below should transform `z_gammaᐞ0ᐞ3` into `z_gamma7_010_3`

edopao:
@tehrengruber Thank you for the review comments. At last week's DaCe workshop, we were informed that array offsets will be deprecated in DaCe. The reason is that offsets are difficult to translate across nested-SDFG boundaries, and therefore very error prone during SDFG transformation. The SDFG generated by this PR works, but It's not guarantee to still work in the future. In DaCe front-ends, in order to avoid offsets, they are running a pass to normalize all array access to be 0-based. We could discuss how to do it in ITIR as well.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -628,3 +754,33 @@ def _check_shift_offsets_are_literals(node: itir.StencilClosure):
             if not all(isinstance(arg, (itir.Literal, itir.OffsetLiteral)) for arg in shift.args):
                 return False
         return True
+
+    @staticmethod
+    def _replace_ssa_identifiers(closure: itir.StencilClosure):
```
petiaccja, line 759: This solves the problem but I don't think this is the right approach.

We have two separate issues here:

### 1. The SSA pass generates invalid identifiers

See the documentation of the SSA pass:
> Mutates the python AST, variable names will not be valid python names anymore.

Instead of adding invalid identifiers and then later removing them, we could simplify the process by not adding them in the first place, so my opinion is that we should change the SSA pass rather than changing all backends. This could be as simple as replacing the special characters the SSA pass uses with an underscore right after the main SSA pass in a post-pass.

### 2. Python identifiers may not be valid in backends

This has nothing to do with special characters added by the SSA pass, but we it may happen that user-written code has valid identifiers in Python, but not in a backend. We should come up with a scheme that maps these invalid Python identifiers to a more common character set and reuse that pass across all backends. This is unrelated to this PR.

petiaccja (dismissed):

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
tehrengruber: True, but then `z_gammaᐞ0ᐞ3` collides with `z_gamma7_010_3`.

edopao: I could add a check that `z_gamma7_010_3` does not already exist among `closure.stencil.params`, is that what you are trying to say?

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -161,10 +164,22 @@ def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs) -> None:
...
     else:
+        program = preprocess_program(program, offset_provider, LiftMode.FORCE_INLINE)
+        if all([ItirToSDFG._check_no_lifts(node) for node in program.closures]):
+            tmps = []
+        else:
+            program_with_tmps: global_tmps.FencilWithTemporaries = preprocess_program(
+                program, offset_provider, LiftMode.FORCE_TEMPORARIES
+            )
+            program = program_with_tmps.fencil
+            tmps = program_with_tmps.tmps
```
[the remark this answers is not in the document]
edopao: I agree, the reason I was doing it this way is that I get exceptions for `LiftMode.FORCE_TEMPORARIES` in some cases. 

@tehrengruber It's 2 types of exceptions. One is related to type inference:

```
__________________________________________________ TestMoAdvectionTrajBtrajComputeO1Dsl.test_TestMoAdvectionTrajBtrajComputeO1Dsl[simple_mesh-run_dace_cpu] ___________________________________________________
../edopao/gt4py/src/gt4py/next/iterator/transforms/pass_manager.py:145: in apply_common_transforms
    ir = CreateGlobalTmps().visit(ir, offset_provider=offset_provider)
../edopao/gt4py/src/gt4py/eve/visitors.py:122: in visit
    return visitor(node, **kwargs)
../edopao/gt4py/src/gt4py/next/iterator/transforms/global_tmps.py:575: in visit_FencilDefinition
    res = split_closures(node, offset_provider=offset_provider)
../edopao/gt4py/src/gt4py/next/iterator/transforms/global_tmps.py:208: in split_closures
    type_inference.infer_all(node, offset_provider=offset_provider, save_to_annex=True)
../edopao/gt4py/src/gt4py/next/iterator/type_inference.py:994: in infer_all
    unified_types, unsatisfiable_constraints = unify(
../edopao/gt4py/src/gt4py/next/type_inference.py:358: in unify
    return unifier.unify()
../edopao/gt4py/src/gt4py/next/type_inference.py:261: in unify
    handled = self._handle_constraint(constraint)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <gt4py.next.type_inference._Unifier object at 0x13148b130>, constraint = (_Box(value=Location(name='ANYWHERE')), _Box(value=Location(name='Edge')))

    def _handle_constraint(self, constraint: tuple[_Box, _Box]) -> bool:
        """Handle a single constraint."""
        s, t = (c.value for c in constraint)
        if s == t:
            # Constraint is satisfied if LHS equals RHS
            return True
    
        if type(s) is TypeVar:
            assert s not in _free_variables(t)
            # Just replace LHS by RHS if LHS is a type variable
            self._rename(s, t)
            return True
    
        if s.handle_constraint(t, self._add_constraint):
            # Use a custom constraint handler if available
            return True
    
        if type(s) is type(t):
            assert s not in _free_variables(t) and t not in _free_variables(s)
            assert datamodels.fields(s).keys() == datamodels.fields(t).keys()
            for k in datamodels.fields(s).keys():
                sv = getattr(s, k)
                tv = getattr(t, k)
                if isinstance(sv, Type):
                    assert isinstance(tv, Type)
                    self._add_constraint(sv, tv)
                else:
>                   assert sv == tv
E                   AssertionError

../edopao/gt4py/src/gt4py/next/type_inference.py:318: AssertionError
```

and the other one is relate to the temporary pass:
```
___________________________________________________________________ TestCalculateNabla4.test_TestCalculateNabla4[simple_mesh-run_dace_cpu] ____________________________________________________________________
../edopao/gt4py/src/gt4py/next/iterator/transforms/pass_manager.py:145: in apply_common_transforms
    ir = CreateGlobalTmps().visit(ir, offset_provider=offset_provider)
../edopao/gt4py/src/gt4py/eve/visitors.py:122: in visit
    return visitor(node, **kwargs)
../edopao/gt4py/src/gt4py/next/iterator/transforms/global_tmps.py:583: in visit_FencilDefinition
    res = update_domains(res, offset_provider)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

node = FencilWithTemporaries(fencil=FencilDefinition(id=SymbolName('calculate_nabla4'), function_definitions=[], params=[Sym(...abla4_e2_size_0'), kind=None, dtype=None), Sym(id=SymbolName('__z_nabla4_e2_size_1'), kind=None, dtype=None)], tmps=[])
offset_provider = {'C2CE': <gt4py.next.iterator.embedded.NeighborTableOffsetProvider object at 0x140b37130>, 'C2CEC': <gt4py.next.iterat...object at 0x140b36680>, 'C2E2C': <gt4py.next.iterator.embedded.NeighborTableOffsetProvider object at 0x140b34250>, ...}

    def update_domains(node: FencilWithTemporaries, offset_provider: Mapping[str, Any]):
        horizontal_sizes = _max_domain_sizes_by_location_type(offset_provider)
    
        closures: list[ir.StencilClosure] = []
        domains = dict[str, ir.FunCall]()
        for closure in reversed(node.fencil.closures):
            if closure.domain == AUTO_DOMAIN:
                # every closure with auto domain should have a single out field
                assert isinstance(closure.output, ir.SymRef)
                domain = domains[closure.output.id]
                closure = ir.StencilClosure(
                    domain=copy.deepcopy(domain),
                    stencil=closure.stencil,
                    output=closure.output,
                    inputs=closure.inputs,
                )
            else:
                domain = closure.domain
    
            closures.append(closure)
    
            if closure.stencil == ir.SymRef(id="deref"):
                # all closure inputs inherit the domain
                for input_arg in _tuple_constituents(closure.inputs[0]):
                    assert isinstance(input_arg, ir.SymRef)
                    assert domains.get(input_arg.id, domain) == domain
                    domains[input_arg.id] = domain
                continue
    
            local_shifts = trace_shifts.TraceShifts.apply(closure)
            for param, shift_chains in local_shifts.items():
                assert isinstance(param, str)
                consumed_domains: list[SymbolicDomain] = (
                    [SymbolicDomain.from_expr(domains[param])] if param in domains else []
                )
                for shift_chain in shift_chains:
                    consumed_domain = SymbolicDomain.from_expr(domain)
                    for offset_name, offset in _group_offsets(shift_chain):
                        if isinstance(offset_provider[offset_name], gtx.Dimension):
                            # cartesian shift
                            dim = offset_provider[offset_name].value
                            consumed_domain.ranges[dim] = consumed_domain.ranges[dim].translate(offset)
                        elif isinstance(offset_provider[offset_name], gtx.NeighborTableOffsetProvider):
                            # unstructured shift
                            nbt_provider = offset_provider[offset_name]
                            old_axis = nbt_provider.origin_axis.value
                            new_axis = nbt_provider.neighbor_axis.value
                            consumed_domain.ranges.pop(old_axis)
                            assert new_axis not in consumed_domain.ranges
                            consumed_domain.ranges[new_axis] = SymbolicRange(
                                im.literal("0", ir.INTEGER_INDEX_BUILTIN),
                                im.literal(str(horizontal_sizes[new_axis]), ir.INTEGER_INDEX_BUILTIN),
                            )
                        else:
>                           raise NotImplementedError
E                           NotImplementedError

../edopao/gt4py/src/gt4py/next/iterator/transforms/global_tmps.py:493: NotImplementedError
```

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
edopao: @tehrengruber For issue 1, is it possible to paremeterize `_UNIQUE_NAME_SEPARATOR` in the SSA pass? or can we just change `ᐞ` to an underscore?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
tehrengruber: As a bandaid I'm fine with that, but we should find a general algorithm that is able to remove / escape special characters without collisions.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
tehrengruber: Just using an underscore is not an option as it will lead to collisions with other symbols. We need a way to generate collision free symbols. If we are able to do that for arbitrary symbol names and to a common character set as @petiaccja mentions both issues are solved.

As a side-note the documentation is wrong the identifiers are valid python names. I would need to look up the PEP, but `ᐞ` is fine.

src/gt4py/next/program_processors/runners/dace_iterator/utility.py
```diff
@@ -35,6 +35,14 @@ def as_dace_type(type_: ts.ScalarType):
     raise ValueError(f"scalar type {type_} not supported")
 
 
+def as_scalar_type(dtype: str) -> ts.ScalarType:
+    try:
+        kind = getattr(ts.ScalarKind, dtype.upper())
+    except AttributeError:
+        raise ValueError(f"Data type {dtype} not supported.")
+    return ts.ScalarType(kind)
```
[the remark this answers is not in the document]
egparedes: What do you mean exactly by `checked_cast`? You can always write a custom Typeguard function (see: https://peps.python.org/pep-0647/#typeguard-type) to do a runtime check whose result is also available for the static type checker.

In case you're not aware, more detailed data type definitions were added for the ongoing work with the storages and the embedded mode which could be useful here (but I don't really know because I'm not following closely the PR): https://github.com/GridTools/gt4py/blob/main/src/gt4py/_core/definitions.py  
In the incoming PR with the remaining part of the storages there are some changes and fixes to this file

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
egparedes: There are indeed two different questions here as @petiaccja said: the method used by the SSA pass to rename identifiers and the missing sanitation of the IR identifiers in case the source language of some code-generation backends doesn't support the same characters as Python. Previously I also thought there were an associated third question which is how DaCe deals internally with non-valid identifiers from inputs, but if I understood correctly what @edopao said, DaCe checks beforehand that the identifiers are valid, which means that for us the issue with the DaCe backend is exactly the same as with other backends with limited support for unicode symbols like  C++.

Regarding the first question, the current SSA pass generates of course valid Python identifiers because the current separator was explicit chosen from the set of valid python identifiers (check https://docs.python.org/3/reference/lexical_analysis.html#identifiers) as a readable character for humans when inspecting and debugging the ITIR code. If the current implementation is problematic it could be changed to a different one, but since this wouldn't remove the need for a pass to sanitize the identifiers before passing them to some backends, unless there is a new proposal which doesn't harm human readability, I don't think it would make the code better.

For the second question, I do think there is need to add some identifiers sanitation pass to deal with the non-supported characters in the backend language. I don't have a concrete proposal but it doesn't look like something hard to add. One tool could help is the `slugify` method from `boltons` (https://boltons.readthedocs.io/en/latest/strutils.html#boltons.strutils.slugify) directly exposed in `eve.utils` (IIRC).

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -68,28 +70,44 @@ def convert_arg(arg: Any):
...
     else:
-        fencil_definition = apply_common_transforms(
-            program,
-            common_subexpression_elimination=False,
-            force_inline_lambda_args=True,
-            lift_mode=lift_mode,
-            offset_provider=offset_provider,
-            unroll_reduce=True,
+        raise TypeError(
+            f"Expected a `FencilDefinition` or `FencilWithTemporaries`, but got `{type(program).__name__}`."
```
DropD, line 107:
```suggestion
            f"Expected a 'FencilDefinition' or 'FencilWithTemporaries', got '{type(program).__name__}'."
```

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -216,9 +234,16 @@ def build_sdfg_from_itir(
     device = dace.DeviceType.GPU if on_gpu else dace.DeviceType.CPU
 
     # visit ITIR and generate SDFG
-    program = preprocess_program(program, offset_provider, lift_mode)
-    sdfg_genenerator = ItirToSDFG(arg_types, offset_provider, column_axis, on_gpu)
+    program, tmps = preprocess_program(program, offset_provider, lift_mode)
+    sdfg_genenerator = ItirToSDFG(arg_types, offset_provider, tmps, column_axis, on_gpu)
     sdfg = sdfg_genenerator.visit(program)
+    if sdfg is None:
+        raise RuntimeError(f"visit failed for program {program.id}")
```
DropD, line 241:
```suggestion
        raise RuntimeError(f"visit failed for program '{program.id}'")
```

DropD (commented):
Small changes to conform to upcomming error message guidelines.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
edopao: I realized that this pass is not immediately needed. It is possible to enable `lift_mode=FORCE_TEMPORARIES` on gt4py unit tests without this pass. I can introduce it in a separate PR, later, if needed.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
edopao: I realized that this pass is not immediately needed. It is possible to enable `lift_mode=FORCE_TEMPORARIES` on gt4py unit tests without this pass. I can introduce it in a separate PR, if needed, as a generic ITIR pass to sanitize symbols.

edopao:
I would like to merge this PR, in order to stop overriding `lift_mode` in gt4py unit tests, which is misleading. The support for temporaries will be extended once the improvement on the temporary pass is merged. This extension includes, for example, a generic ITIR pass to sanitize unicode symbols, if we still see the need for it. Another extension could be the normalization of field domain in order to avoid array offsets, if we see that DaCe cannot handle it in SDFG transformations.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -189,6 +197,103 @@ def add_storage(
             raise NotImplementedError()
         self.storage_types[name] = type_
 
+    def generate_temporaries(
```
philip-paul-mueller, line 200: I am not sure about the name.
What confuses me is that `self.tmps` are the temporaries, or not?
And later you call this function only if `self.tmps` is not empty, which seems a bit contradictory.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -189,6 +197,103 @@ def add_storage(
...
+        for sym in node_params:
+            if all([sym.id != tmp.id for tmp in self.tmps]) and sym.kind != "Iterator":
+                name_ = str(sym.id)
+                type_ = self.storage_types[name_]
+                assert isinstance(type_, ts.ScalarType)
+                symbol_map[name_] = SymbolExpr(name_, as_dace_type(type_))
+
+        symbol_dtype = dace.int64
+        tmp_symbols: dict[str, TaskletExpr] = {}
+        for tmp in self.tmps:
```
philip-paul-mueller, line 216: You spend ~70 lines on filling `tmp_symbols` and does not explains what the loop does or the variable represents.
But for the small loop above you write a description, this seems a bit strange.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -189,6 +197,103 @@ def add_storage(
...
 
+    def generate_temporaries(
+        self, node_params: list[Sym], defs_state: dace.SDFGState, program_sdfg: dace.SDFG
+    ) -> dict[str, TaskletExpr]:
+        """Create a table of symbols which are used to define array shape, stride and offset for temporaries."""
+        symbol_map: dict[str, TaskletExpr] = {}
+        # The shape of temporary arrays might be defined based on the shape of other input/output fields.
+        # Therefore, here we collect the symbols used to define data-field parameters that are not temporaries.
+        for sym in node_params:
+            if all([sym.id != tmp.id for tmp in self.tmps]) and sym.kind != "Iterator":
```
philip-paul-mueller, line 208: This seems very indirect to me, and not that reliable, because what is with a scalar input, such as the timestep, wouldn't that be matched here as well.
Since all strides/shapes are generated by one function, thy all have a predicable name, which you could match using regex.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -189,6 +197,103 @@ def add_storage(
...
+            tmp_shape, _ = new_array_symbols(tmp_name, len(dims))
+            tmp_offset = [
+                dace.symbol(unique_name(f"{tmp_name}_offset{i}")) for i in range(len(dims))
+            ]
+            _, tmp_array = program_sdfg.add_array(
+                tmp_name, tmp_shape, as_dace_type(type_.dtype), offset=tmp_offset, transient=True
+            )
+
+            # Loop through all dimensions to initialize the array parameters for shape and offsets
+            for (_, (begin, end)), offset_sym, shape_sym in zip(
```
philip-paul-mueller, line 252: It looks as if `begin.value` and `end.value` are constant and known during translation time, as the tasklets have no inputs.
So why dou you creates scalars in the first place, you should just create them as symbols (what you later do but you could compute their value here.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -214,6 +319,9 @@ def visit_FencilDefinition(self, node: itir.FencilDefinition):
         for param, type_ in zip(node.params, self.param_types):
             self.add_storage(program_sdfg, str(param.id), type_, neighbor_tables)
 
+        if self.tmps:
```
philip-paul-mueller, line 322: Since the usage of the symbol is 60 lines down, I would here make a note about its usage.

philip-paul-mueller (changes requested):

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
edopao: Good point, code updated.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
philip-paul-mueller: But you should be able to feed that to symbolic assignements in edges.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -189,6 +197,106 @@ def add_storage(
...
+                shape_tasklet = defs_state.add_tasklet(
+                    "shape",
+                    code=f"__result = {end.value} - {begin.value}",
+                    inputs={},
+                    outputs={"__result"},
+                )
+                shape_var = unique_var_name()
+                program_sdfg.add_scalar(shape_var, symbol_dtype, transient=True)
+                shape_node = defs_state.add_access(shape_var)
+                defs_state.add_edge(
```
philip-paul-mueller, line 286: According to my understanding, you should be able to put even complex expressions in assignments on interstate edge.

philip-paul-mueller (approved):
There are two suggestions, however, they are not mandatory in my view, so I will approve it.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
edopao: I remember I tried and it din't work, at that time. I think the problem was in some symbolic expression like `some_symbol - 1`, but I could be wrong. Anyway, it seems to work for now so I will change as you suggest.

edopao: Oh no, here we go:
```
../../../../../src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py:363: in visit_FencilDefinition
    program_sdfg.validate()
/Users/epaone/repo/dace/dace/sdfg/sdfg.py:2314: in validate
    validate_sdfg(self, references, **context)
/Users/epaone/repo/dace/dace/sdfg/validation.py:264: in validate_sdfg
    validate_control_flow_region(sdfg, sdfg, initialized_transients, symbols, references, **context)
/Users/epaone/repo/dace/dace/sdfg/validation.py:79: in validate_control_flow_region
    undef_syms = set(edge.data.free_symbols) - set(symbols.keys())
/Users/epaone/repo/dace/dace/sdfg/sdfg.py:287: in free_symbols
    return self.used_symbols(all_symbols=True)
/Users/epaone/repo/dace/dace/sdfg/sdfg.py:276: in used_symbols
    rhs_symbols |= set(map(str, dace.symbolic.symbols_in_ast(ast.parse(rhs))))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

source = "SymbolExpr(value='0 - 0', dtype=long long)", filename = '<unknown>'
mode = 'exec'

    def parse(source, filename='<unknown>', mode='exec', *,
              type_comments=False, feature_version=None):
        """
        Parse the source into an AST node.
        Equivalent to compile(source, filename, mode, PyCF_ONLY_AST).
        Pass type_comments=True to get back type comments where the syntax allows.
        """
        flags = PyCF_ONLY_AST
        if type_comments:
            flags |= PyCF_TYPE_COMMENTS
        if isinstance(feature_version, tuple):
            major, minor = feature_version  # Should be a 2-tuple.
            assert major == 3
            feature_version = minor
        elif feature_version is None:
            feature_version = -1
        # Else it should be an int giving the minor version for 3.x.
>       return compile(source, filename, mode, flags,
                       _feature_version=feature_version)
E         File "<unknown>", line 1
E           SymbolExpr(value='0 - 0', dtype=long long)
E                                           ^^^^^^^^^
E       SyntaxError: invalid syntax. Perhaps you forgot a comma?

/opt/homebrew/Cellar/python@3.10/3.10.13_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py:50: SyntaxError
```

edopao: and different error if I change dtype to `int32`

edopao: Right, probably my mistake.

PR #1352, opened 2023-10-19
feat[next]: Extend astype to work with tuples

ninaburg:
This pull request extends the functionality of the built-in `astype` function in the `ffront` module such that we can write `field_float32, field_float32 = astype((field_float64, field_float64), float32)`. The changes specifically impact the `foast_to_itir.py` file. Here's an overview of the modifications:

- The `astype` function's argument and return types have been extended to accept tuples of fields in addition to individual fields.
- In the `foast_to_itir.py` file, when casting tuples, each element is processed individually. The casted elements are then combined into a new tuple using `make_tuple`.
- The code also includes test cases to ensure that the modified `astype` function correctly handles tuples of fields.

The motivation for these changes is to simplify and enhance code readability when performing operations that mix different types. In such scenarios, explicit type promotion is required, resulting in cluttered stencils with casting operations. This update makes the type conversion explicit and visible, reducing the need for extensive casting operations and improving code clarity.

### DIFF
--- a/src/gt4py/next/ffront/fbuiltins.py
+++ b/src/gt4py/next/ffront/fbuiltins.py
@@ -175,7 +175,11 @@ def where(
 
 
 @builtin_function
-def astype(field: Field | gt4py_defs.ScalarT, type_: type, /) -> Field:
+def astype(
+    field: Field | gt4py_defs.ScalarT | Tuple[Field, ...],
+    type_: type,
+    /,
+) -> Field | Tuple[Field, ...]:
     raise NotImplementedError()
 
 
--- a/src/gt4py/next/ffront/foast_passes/type_deduction.py
+++ b/src/gt4py/next/ffront/foast_passes/type_deduction.py
@@ -837,18 +837,12 @@ class FieldOperatorTypeDeduction(traits.VisitorWithSymbolTableTrait, NodeTransla
                 f"Invalid call to `astype`. Second argument must be a scalar type, but got {new_type}.",
             )
 
-        if isinstance(value, foast.TupleExpr):
-            return_type = type_info.apply_to_primitive_constituents(
-                value.type,
-                lambda primitive_type: with_altered_scalar_kind(
-                    primitive_type, getattr(ts.ScalarKind, new_type.id.upper())
-                ),
-            )
-
-        else:
-            return_type = with_altered_scalar_kind(
-                value.type, getattr(ts.ScalarKind, new_type.id.upper())
-            )
+        return_type = type_info.apply_to_primitive_constituents(
+            value.type,
+            lambda primitive_type: with_altered_scalar_kind(
+                primitive_type, getattr(ts.ScalarKind, new_type.id.upper())
+            ),
+        )
 
         return foast.Call(
             func=node.func,
--- a/src/gt4py/next/ffront/foast_passes/type_deduction.py
+++ b/src/gt4py/next/ffront/foast_passes/type_deduction.py
@@ -840,6 +840,7 @@ class FieldOperatorTypeDeduction(traits.VisitorWithSymbolTableTrait, NodeTransla
         if isinstance(value, foast.TupleExpr):
             element_types_new = []
             for element in value.elts:
+                assert isinstance(element.type, (ts.FieldType, ts.ScalarType))
                 element_types_new.append(
                     with_altered_scalar_kind(
                         element.type, getattr(ts.ScalarKind, new_type.id.upper())
--- a/src/gt4py/next/ffront/foast_passes/type_deduction.py
+++ b/src/gt4py/next/ffront/foast_passes/type_deduction.py
@@ -838,15 +838,12 @@ class FieldOperatorTypeDeduction(traits.VisitorWithSymbolTableTrait, NodeTransla
             )
 
         if isinstance(value, foast.TupleExpr):
-            element_types_new = []
-            for element in value.elts:
-                assert isinstance(element.type, (ts.FieldType, ts.ScalarType))
-                element_types_new.append(
-                    with_altered_scalar_kind(
-                        element.type, getattr(ts.ScalarKind, new_type.id.upper())
-                    )
-                )
-            return_type = ts.TupleType(types=cast(list[ts.DataType], element_types_new))
+            return_type = type_info.apply_to_primitive_constituents(
+                value.type,
+                lambda primitive_type: with_altered_scalar_kind(
+                    primitive_type, getattr(ts.ScalarKind, new_type.id.upper())
+                ),
+            )
 
         else:
             return_type = with_altered_scalar_kind(
--- a/src/gt4py/next/ffront/foast_passes/type_deduction.py
+++ b/src/gt4py/next/ffront/foast_passes/type_deduction.py
@@ -823,10 +823,12 @@ class FieldOperatorTypeDeduction(traits.VisitorWithSymbolTableTrait, NodeTransla
         return self._visit_reduction(node, **kwargs)
 
     def _visit_astype(self, node: foast.Call, **kwargs) -> foast.Call:
+        return_type: ts.TupleType | ts.ScalarType | ts.FieldType
         value, new_type = node.args
         assert isinstance(
-            value.type, (ts.FieldType, ts.ScalarType)
+            value.type, (ts.FieldType, ts.ScalarType, ts.TupleType)
         )  # already checked using generic mechanism
+
         if not isinstance(new_type, foast.Name) or new_type.id.upper() not in [
             kind.name for kind in ts.ScalarKind
         ]:
@@ -835,9 +837,20 @@ class FieldOperatorTypeDeduction(traits.VisitorWithSymbolTableTrait, NodeTransla
                 f"Invalid call to `astype`. Second argument must be a scalar type, but got {new_type}.",
             )
 
-        return_type = with_altered_scalar_kind(
-            value.type, getattr(ts.ScalarKind, new_type.id.upper())
-        )
+        if isinstance(value, foast.TupleExpr):
+            element_types_new = []
+            for element in value.elts:
+                element_types_new.append(
+                    with_altered_scalar_kind(
+                        element.type, getattr(ts.ScalarKind, new_type.id.upper())
+                    )
+                )
+            return_type = ts.TupleType(types=cast(list[ts.DataType], element_types_new))
+
+        else:
+            return_type = with_altered_scalar_kind(
+                value.type, getattr(ts.ScalarKind, new_type.id.upper())
+            )
 
         return foast.Call(
             func=node.func,
--- a/src/gt4py/next/ffront/foast_to_itir.py
+++ b/src/gt4py/next/ffront/foast_to_itir.py
@@ -319,22 +319,7 @@ class FieldOperatorLowering(NodeTranslator):
         assert len(node.args) == 2 and isinstance(node.args[1], foast.Name)
         obj, dtype = node.args[0], node.args[1].id
 
-        def recursive_cast(obj, dtype):
-            if isinstance(obj, foast.TupleExpr):
-                casted_elements = []
-
-                for element in obj.elts:
-                    casted_element = recursive_cast(element, dtype)
-                    casted_elements.append(casted_element)
-
-                return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
-                    *casted_elements
-                )
-
-            else:
-                return self._map(im.lambda_("it")(im.call("cast_")("it", str(dtype))), obj)
-
-        return recursive_cast(obj, dtype)
+        return self._process_elements(obj, lambda x: im.call("cast_")(x, str(dtype)), **kwargs)
 
     def _visit_where(self, node: foast.Call, **kwargs) -> itir.FunCall:
         return self._map("if_", *node.args)
@@ -413,6 +398,33 @@ class FieldOperatorLowering(NodeTranslator):
             op = im.call("map_")(op)
 
         return im.promote_to_lifted_stencil(im.call(op))(*lowered_args)
+
+    def _process_elements(self, obj, process_func, **kwargs):
+        """
+        Recursively applies a processing function to the elements of a structured object, preserving its structure.
+
+        Args:
+            obj: The structured object to be processed.
+            process_func: A function to apply to each element.
+            **kwargs: Additional keyword arguments for the processing function.
+
+        Example:
+            result = process_elements(obj, lambda x: im.call("cast_")(x, str(dtype)), **kwargs)
+
+        Returns:
+            Structured object with the processing function applied to its elements.
+        """
+        if isinstance(obj.type, ts.TupleType):
+            if isinstance(obj, foast.Name) or isinstance(obj, foast.Call):
+                return self.visit(obj)
+            processed_elements = [
+                self._process_elements(el, process_func, **kwargs) for el in obj.elts
+            ]
+            return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
+                *[el for el in processed_elements]
+            )
+        else:
+            return self._map(im.lambda_("it")(process_func("it")), obj)
 
 
 class FieldOperatorLoweringError(Exception):
--- a/src/gt4py/next/ffront/foast_to_itir.py
+++ b/src/gt4py/next/ffront/foast_to_itir.py
@@ -319,7 +319,7 @@ class FieldOperatorLowering(NodeTranslator):
         assert len(node.args) == 2 and isinstance(node.args[1], foast.Name)
         obj, new_type = node.args[0], node.args[1].id
         return self._process_elements(
-            obj, obj.type, lambda x: im.call("cast_")(x, str(new_type)), "expr", **kwargs
+            lambda x: im.call("cast_")(x, str(new_type)), obj, obj.type, **kwargs
         )
 
     def _visit_where(self, node: foast.Call, **kwargs) -> itir.FunCall:
@@ -400,27 +400,31 @@ class FieldOperatorLowering(NodeTranslator):
 
         return im.promote_to_lifted_stencil(im.call(op))(*lowered_args)
 
-    def _process_elements(self, obj, obj_type, process_func, expr, **kwargs):
-        """Recursively applies a processing function to the elements of a structured object, preserving its structure."""
-        if isinstance(obj_type, ts.TupleType):
-            if isinstance(obj, foast.Name) or isinstance(obj, foast.Call):
-                return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
-                    *[
-                        self._process_elements(
-                            obj, obj_type.types[i], process_func, im.tuple_get(i, expr)
-                        )
-                        for i in range(len(obj.type.types))
-                    ]
-                )
-            elif isinstance(obj, foast.TupleExpr):
-                return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
-                    *[
-                        self._process_elements(el, el.type, process_func, expr, **kwargs)
-                        for el in obj.elts
-                    ]
-                )
+    def _process_elements(
+        self,
+        process_func: Callable[[itir.Expr], itir.Expr],
+        obj: foast.Expr,
+        current_el_type: ts.TypeSpec,
+        current_el_expr: itir.Expr = im.ref("expr"),
+    ):
+        """Recursively applies a processing function to all primitive constituents of a tuple."""
+        if isinstance(current_el_type, ts.TupleType):
+            # TODO(ninaburg): Refactor to avoid duplicating lowered obj expression for each tuple element.
+            return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
+                *[
+                    self._process_elements(
+                        process_func,
+                        obj,
+                        current_el_type.types[i],
+                        im.tuple_get(i, current_el_expr),
+                    )
+                    for i in range(len(current_el_type.types))
+                ]
+            )
+        elif type_info.contains_local_field(current_el_type):
+            raise NotImplementedError("Processing fields with local dimension is not implemented.")
         else:
-            return self._map(im.lambda_("expr")(process_func(expr)), obj)
+            return self._map(im.lambda_("expr")(process_func(current_el_expr)), obj)
 
 
 class FieldOperatorLoweringError(Exception):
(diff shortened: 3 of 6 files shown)


ninaburg:
cscs-ci run

ninaburg:
cscs-ci run

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -175,7 +175,11 @@ def where(
 
 
 @builtin_function
-def astype(field: Field | gt4py_defs.ScalarT, type_: type, /) -> Field:
+def astype(
+    field: Field | gt4py_defs.ScalarT | Tuple[Field, ...],
+    type_: type,
+    /,
+) -> Field | Tuple[Field, ...]:
```
nfarabullini, line 182: since this function only returns a `NotImplementedError()`, shouldn't the return type here be `None`?

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -835,9 +837,20 @@ def _visit_astype(self, node: foast.Call, **kwargs) -> foast.Call:
                 f"Invalid call to `astype`. Second argument must be a scalar type, but got {new_type}.",
             )
 
-        return_type = with_altered_scalar_kind(
-            value.type, getattr(ts.ScalarKind, new_type.id.upper())
-        )
+        if isinstance(value, foast.TupleExpr):
+            element_types_new = []
+            for element in value.elts:
```
nfarabullini, line 842: I would add an extra check here to make sure that `new_type.id` is a ScalarKind like it is done above in lines 832-838

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -318,12 +318,22 @@ def visit_Call(self, node: foast.Call, **kwargs) -> itir.Expr:
...
         obj, dtype = node.args[0], node.args[1].id
-
-        # TODO check that we test astype that results in a itir.map_ operation
-        return self._map(
-            im.lambda_("it")(im.call("cast_")("it", str(dtype))),
-            obj,
-        )
+        if isinstance(obj, foast.TupleExpr):
+            casted_elements = []
+            for _, element in enumerate(obj.elts):
```
nfarabullini, line 323: if you don't use both elements of enumerate, you can remove it and directly do:
```suggestion
            for element in obj.elts:
```

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -318,12 +318,22 @@ def visit_Call(self, node: foast.Call, **kwargs) -> itir.Expr:
     def _visit_astype(self, node: foast.Call, **kwargs) -> itir.FunCall:
```
nfarabullini, line 318: This I am not sure about, but it might be that the return type could be `itir.Expr` in some cases, which would supersed `itir.FunCall`. Could you please check?

nfarabullini (approved):

src/gt4py/next/ffront/fbuiltins.py
ninaburg: I hear you but the type here seems to define the return type of the built-in function. If I replaced it by `None`, an error will be raised that the return type is "Illegal". Also, all built-in functions defined in this file have their proper return types.

src/gt4py/next/ffront/foast_to_itir.py
ninaburg: That's a good point. When developing the` _visit_astype` method, `mypy` who checks the return types in `pre-commit` complained that `itir.Expr` was returned when `itir.FunCall` is expected. I wanted to use the `ir_makers.promote_to_lifted_stencil` function but it returns an `itir.Expr`, so I finally used its contents without calling it and the tests are now green. So no `itir.Expr` should be returned to answer your question but doesn't the root of this problem lie in the fact that the return type of `im.promote_to_lifted_stencil` which also is called in the `_map` function should rather be a itir.`FunCall `as it calls a lift rather than `itir.Expr`?

src/gt4py/next/ffront/foast_passes/type_deduction.py
ninaburg: As the code in lines 832-838 already includes the check, wouldn't an extra check here make it redundant?

ninaburg:
cscs-ci run

src/gt4py/next/ffront/foast_passes/type_deduction.py
nfarabullini: yes, the check should be over `element.type`. Sorry

src/gt4py/next/ffront/foast_to_itir.py
nfarabullini: this is more of a question for @tehrengruber

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_type_deduction.py
```diff
@@ -785,8 +785,8 @@ def simple_astype(a: Field[[TDim], float64]):
 
 def test_astype_wrong_value_type():
     def simple_astype(a: Field[[TDim], float64]):
-        # we just use a tuple here but anything that is not a field or scalar works
-        return astype((1, 2), bool)
+        # we just use broadcast here but anything that is not a field, scalar or tuple works
```
tehrengruber, line 788:
```suggestion
        # we just use broadcast here but anything that is not a field, scalar or tuple thereof works
```
Nitpicking.

tests/next_tests/unit_tests/ffront_tests/foast_passes_tests/test_type_alias_replacement.py
```diff
@@ -42,3 +46,22 @@ def fieldop_with_typealias(
         foast_tree.body.stmts[0].value.left.func.id == expected
         and foast_tree.body.stmts[0].value.right.args[1].id == expected
     )
+
+
+def test_type_alias_replacement_astype_with_tuples():
```
tehrengruber, line 51: While from a user perspective the type alias feature is often used in conjunction with where on tuples, from a toolchain perspective the features are orthogonal (i.e. a bug in the type alias replacement has no effect on the where on tuples and vice-versa). I would suggest to remove this test and instead add an integration test (somewhere where `where` is tested) that ensures
```
    def where_on_tuples(
        a: gtx.Field[[TDim], float], b: gtx.Field[[TDim], float]
    ) -> tuple[gtx.Field[[TDim], float], gtx.Field[[TDim], float]]:
        return astype((a, b), float)
```
gives the expected result (i.e. the values are correct). If you want to go the full mile you can also add a unit test in `test_foast_to_itir.py` that ensures the generated ITIR is as expected. Please also add a test for nested tuples.

src/gt4py/next/ffront/foast_to_itir.py
tehrengruber: As a general guideline: A type annotation should not affect the way you write code, but merely help in finding accidental errors. Occasionally (or given the terrible type system in python frequently) the type checker needs some help and a cast or assert is needed, but that's about it. With respect to the actual problem: The annotation in `promote_to_lifted_stencil` is too lose. In fact it only ever returns an `itir.FunCall` (or callable returning one to be precise) , which is why replacing the call to `promote_to_lifted_stencil` with its implementation made mypy happy. If you tighten the annotation in `promote_to_lifted_stencil` to `itir.FunCall` it should work.

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -318,12 +318,22 @@ def visit_Call(self, node: foast.Call, **kwargs) -> itir.Expr:
...
         assert len(node.args) == 2 and isinstance(node.args[1], foast.Name)
         obj, dtype = node.args[0], node.args[1].id
-
-        # TODO check that we test astype that results in a itir.map_ operation
-        return self._map(
-            im.lambda_("it")(im.call("cast_")("it", str(dtype))),
-            obj,
-        )
+        if isinstance(obj, foast.TupleExpr):
+            casted_elements = []
```
tehrengruber, line 322: This will fail for nested tuples. I would suggest to first extend it (using some recursive function) and then try to generalize it to a function that is independent of the `cast` use-case here (analogous to `apply_to_primitive_constituents`). Let's start with a first draft and look at it together again, not sure how well this works out.

tehrengruber (changes requested):

ninaburg:
cscs-ci run

tests/next_tests/unit_tests/ffront_tests/foast_passes_tests/test_type_alias_replacement.py
ninaburg: I suppose you meant `astype` instead of `where`. The tests have been removed from the type alias tests and the integration tests have been added. I want to go the full mile, so the tests in `test_foast_to_itir.py` are to come.

src/gt4py/next/ffront/foast_to_itir.py
ninaburg: The first suggestion (recursive function not generalized) has been added and here a suggestion for the generalized function analogous to `apply_to_primitive_constituents` which has proven to be very useful: 
```python

def process_elements(obj, process_func, **kwargs):
    """
    Example:
    result = process_elements(obj, lambda x: im.call("cast_")(x, str(dtype)), **kwargs)

    """
    if isinstance(obj, foast.TupleExpr):
        processed_elements = [process_elements(el, process_func, **kwargs) for el in obj.elts]
        return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
            *[self.visit(el, **kwargs) for el in processed_elements])
    else:
        return self._map(im.lambda_("it")(process_func("it")), obj)
```

ninaburg:
cscs-ci run

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -835,9 +837,18 @@ def _visit_astype(self, node: foast.Call, **kwargs) -> foast.Call:
...
+                value.type,
+                lambda primitive_type: with_altered_scalar_kind(
+                    primitive_type, getattr(ts.ScalarKind, new_type.id.upper())
+                ),
+            )
+
+        else:
+            return_type = with_altered_scalar_kind(
+                value.type, getattr(ts.ScalarKind, new_type.id.upper())
+            )
```
tehrengruber, line 851:
```suggestion
        return_type = type_info.apply_to_primitive_constituents(
                value.type,
                lambda primitive_type: with_altered_scalar_kind(
                    primitive_type, getattr(ts.ScalarKind, new_type.id.upper())
                ),
            )
```
Almost there :-P. `apply_to_primitive_constituents` also works on a scalar, that way the `with_altered_scalar_kind` call is not duplicated.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -327,6 +327,36 @@ def testee(a: cases.IFloatField) -> gtx.Field[[IDim], int64]:
...
+    @gtx.field_operator
+    def cast_tuple(
+        a: cases.IFloatField, b: cases.IFloatField
+    ) -> tuple[gtx.Field[[IDim], int64], gtx.Field[[IDim], int64]]:
+        return astype((a, b), int64)
+
+    @gtx.field_operator
+    def combine(a: cases.IFloatField, b: cases.IFloatField) -> gtx.Field[[IDim], int64]:
+        packed_tuple = cast_tuple(a, b)
+        return packed_tuple[0] + packed_tuple[1]
```
tehrengruber, line 340:
```suggestion
```
This seems an unnecessary distraction from what the test is supposed to cover. Calling `cast_tuple` directly should be fine. Returning a tuple will likely fail with the dace backend, but just mark the test with `@pytest.mark.uses_tuple_returns` until dace supports it. Additionally I would use `cases.verify` with values that are non-integral (e.g. `1.1`) such that if the cast doesn't happen, but `verify_with_default_data` just uses an integer, it doesn't silently work.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -327,6 +327,36 @@ def testee(a: cases.IFloatField) -> gtx.Field[[IDim], int64]:
...
+
+    cases.verify_with_default_data(cartesian_case, combine, ref=lambda a, b: a + b)
+
+
+def test_astype_on_nested_tuples(cartesian_case):
```
tehrengruber, line 345: Same as above.

src/gt4py/next/ffront/foast_to_itir.py
tehrengruber: Great, this is exactly the pattern I meant. However this will fail in case the tuple is not constructed directly, but for example using a function call or in a previous assignment, e.g. this will not work:
```
tup = (foo, bar)
return `astype(tup, int)`
```
or
```
tup1 = (foo, bar)
tup2 = (baz, tup1)
return `astype(tup2, int)`
```
or
```
return `astype(field_op_returning_a_tuple(...), int)`
```
Please add test cases for the first two ones or adopt the existing ones (latter preferred to reduce code bloat).

tehrengruber (changes requested):

ninaburg:
cscs-ci run

src/gt4py/next/ffront/foast_to_itir.py
ninaburg: I adapted here below the pattern to also work for function calls and previous assignment and integrated it in the code.  The corresponding integration tests were also added. Let me know if the function is ok.
```
    def _process_elements(self, obj, process_func, **kwargs):
        """
        Recursively applies a processing function to the elements of a structured object, preserving its structure.
        Args:
            obj: The structured object to be processed.
            process_func: A function to apply to each element.
            **kwargs: Additional keyword arguments for the processing function.
        Example:
            result = process_elements(obj, lambda x: im.call("cast_")(x, str(dtype)), **kwargs)
        Returns:
            Structured object with the processing function applied to its elements.
        """
        if isinstance(obj.type, ts.TupleType):
            if isinstance(obj, foast.Name) or isinstance(obj, foast.Call):
                return self.visit(obj)
            processed_elements = [
                self._process_elements(el, process_func, **kwargs) for el in obj.elts
            ]
            return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
                *[el for el in processed_elements]
            )
        else:
            return self._map(im.lambda_("it")(process_func("it")), obj)
```

ninaburg: I adapted the function in order to handle the cast of tuples and nested tuples that were not created directly like you mentioned above:
```
    def _process_elements(self, obj, obj_type, process_func, expr, **kwargs):
        """Recursively applies a processing function to the elements of a structured object, preserving its structure."""
        if isinstance(obj_type, ts.TupleType):
            if isinstance(obj, foast.Name) or isinstance(obj, foast.Call):
                return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
                    *[
                        self._process_elements(
                            obj, obj_type.types[i], process_func, im.tuple_get(i, expr)
                        )
                        for i in range(len(obj.type.types))
                    ]
                )
            elif isinstance(obj, foast.TupleExpr):
                return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
                    *[
                        self._process_elements(el, el.type, process_func, expr, **kwargs)
                        for el in obj.elts
                    ]
                )
        else:
            return self._map(im.lambda_("expr")(process_func(expr)), obj)
```
The tests were also adapted in the form here below to avoid that the backend silently does the cast when writing back the float output to an int field.
```
@gtx.field_operator
    def cast_tuple(
        a: cases.IFloatField, b: cases.IFloatField,
        a_casted_to_int_outside_of_gt4py: cases.IIntField,
        b_casted_to_int_outside_of_gt4py: cases.IIntField,
    ) -> tuple[gtx.Field[[IDim], bool], gtx.Field[[IDim], bool]]:
        result = astype(field_op_returning_a_tuple(a, b), int32)
        return result[0] == a_casted_to_int_outside_of_gt4py, result[1] == b_casted_to_int_outside_of_gt4py
```

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -403,6 +400,28 @@ def _map(self, op, *args, **kwargs):
...
+                )
+            elif isinstance(obj, foast.TupleExpr):
+                return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
+                    *[
+                        self._process_elements(el, el.type, process_func, expr, **kwargs)
+                        for el in obj.elts
+                    ]
+                )
+        else:
+            return self._map(im.lambda_("expr")(process_func(expr)), obj)
```
tehrengruber, line 423: Ok so this was tricky to grasp. Here is a simplified version:
```suggestion
     def _process_elements(self,
                          process_func: Callable[[itir.Expr], itir.Expr],
                          obj: foast.Expr,
                          current_el_type: ts.TypeSpec,
                          current_el_expr: itir.Expr = im.ref("expr"),
        ):
        """Recursively applies a processing function to all primitive constituents of a tuple."""
        if isinstance(current_el_type, ts.TupleType):
            return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
                *[
                    self._process_elements(
                        process_func, obj, current_el_type.types[i], im.tuple_get(i, current_el_expr)
                    )
                    for i in range(len(obj.type.types))
                ]
            )
        else:
            return self._map(im.lambda_("expr")(process_func(current_el_expr)), obj)
```
What I changed:
 - I renamed `obj_type` to `current_el_type` because this is not (neccesarily) as the name suggests the type of `obj` but the type of the object the function is currently operating on (e.g. a some element).
 - Renamed `expr` to `current_el_expr`. In the original version I was rather confused by the fact that the lambda function in the last line (`self._map(im.lambda_("expr")(process_func(expr)), obj)
`) takes a argument called `expr` and its body consists of what is inside the variable `expr`, but the two have completely different meaning. The former, the symbol `expr`, refers to the value of the entire tuple, while the latter, the ITIR expression stored inside the variable `expr`, refers to the value which `_process_elements` is currently processing. By changing the name it is clear that the two are different things.
- I made `expr` or with the new naming `current_el_expr` have a default value. The caller shouldn't need to know that this needs to be "expr". That is in implementation detail just stemming from a choice of the name of the argument in the lambda function at the end.
- I removed the `isinstance(obj, foast.TupleExpr)` branch of the if. This is not necessary. It generates slightly simpler ITIR code, but I would suggest to remove it for simplicity (in combination with the argument naming it made the function very hard to read). If you want to keep it, use something like this:
```
            elif isinstance(obj, foast.TupleExpr):
                assert current_el_expr == "expr"
                return im.promote_to_lifted_stencil(lambda *elts: im.make_tuple(*elts))(
                    *[
                        self._process_elements(process_func, el, el.type, "expr", **kwargs)
                        for el in obj.elts
                    ]
                )
```
So that it is clear that the current expression is always the top-level "expr" because you also change the `obj` to be an element.
- Shorter docstring. Your's was fine, but too long for the line-length limit we have.

Now... two problems remain:

- Since you are using `self._map` and just pass `obj` which is the entire tuple, you end up with duplicating the lowered `obj` expression many times (ones for every element). If you look at the result of `_process_elements` in the debugger you will see that this expression occurs as many times as you have elements in the tuple. I would suggest to add a todo to fix this later. If you have time to do it great, but we should at least postpone to after your icon4py PR is merged.
- I'm pretty sure this implementation will fail for fields with local dimension. I would suggest to throw a `NotImplemented` exception when this condition is true `type_info.contains_local_field(current_el_type)` in the second last line. This can be addressed together with the previous problem.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -325,6 +325,80 @@ def testee(a: cases.IFloatField) -> gtx.Field[[IDim], int64]:
...
+        )
+
+    a = cases.allocate(cartesian_case, cast_tuple, "a")()
+    b = cases.allocate(cartesian_case, cast_tuple, "b")()
+    a_casted_to_int_outside_of_gt4py = cases.allocate(
+        cartesian_case, cast_tuple, "a", dtype=int32
+    )()
+    b_casted_to_int_outside_of_gt4py = cases.allocate(
+        cartesian_case, cast_tuple, "b", dtype=int32
+    )()
```
tehrengruber, line 371:
```suggestion
    a_casted_to_int_outside_of_gt4py = gtx.np_as_located_field(np.asarray(a).astype(int32))
    b_casted_to_int_outside_of_gt4py = gtx.np_as_located_field(np.asarray(b).astype(int32))
```
This avoids relying on the `cases` utility to generate the same values (which only works because you specify the name of the argument). This is behavior we shouldn't rely on.

tehrengruber (changes requested):

ninaburg:
cscs-ci run

src/gt4py/next/ffront/foast_to_itir.py
ninaburg: Thanks for the detailled feedback. The function is simpler now and pass the tests. The TODOs were also added.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -325,6 +325,76 @@ def testee(a: cases.IFloatField) -> gtx.Field[[IDim], int64]:
...
+        a: cases.IFloatField, b: cases.IFloatField
+    ) -> tuple[gtx.Field[[IDim], float], gtx.Field[[IDim], float]]:
+        tup = (a, b)
+        return tup
+
+    @gtx.field_operator
+    def cast_tuple(
+        a: cases.IFloatField,
+        b: cases.IFloatField,
+        a_casted_to_int_outside_of_gt4py: cases.IField,
```
tehrengruber, line 341:
```suggestion
        a_asint: cases.IField,
```
The naming was not a serious proposal of mine, but just to signify what it meant. Here it should be clear from the context.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -325,6 +325,76 @@ def testee(a: cases.IFloatField) -> gtx.Field[[IDim], int64]:
...
+        result = astype((a, field_op_returning_a_tuple(a, b)), int32)
+        return (
+            result[0] == a_casted_to_int_outside_of_gt4py,
+            result[1][0] == a_casted_to_int_outside_of_gt4py,
+            result[1][1] == b_casted_to_int_outside_of_gt4py,
+        )
+
+    a = cases.allocate(cartesian_case, cast_tuple, "a")()
+    b = cases.allocate(cartesian_case, cast_tuple, "b")()
+    a_casted_to_int_outside_of_gt4py = gtx.np_as_located_field(IDim)(np.asarray(a).astype(int32))
```
tehrengruber, line 366:
```suggestion
    a_asint = gtx.np_as_located_field(IDim)(np.asarray(a).astype(int32))
```
As above.

PR #1353, opened 2023-10-20
feat[next]: Extend DaCe support for offset providers

edopao:
## Description

Extend support in DaCe backend for offset providers, in order to generate the tasklet code in case of shift expressions with both direct and indirect addressing.

This allows to visit expressions like:

`⟪C2E2Cₒ, 0ₒ, Koffₒ, ·⟪C2CECₒ, 0ₒ⟫(zd_vertoffset)⟫(theta_v)`

which originates from user code:

`theta_v(as_offset(Koff, zd_vertoffset(C2CEC[0]) + int32(1)))`

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
@@ -653,17 +653,13 @@ class PythonTaskletCodegen(gt4py.eve.codegen.TemplatedGenerator):
             iterator = self.visit(node.args[0])
 
         assert isinstance(tail[0], itir.OffsetLiteral)
-        offset = tail[0].value
-        assert isinstance(offset, str)
+        offset_dim = tail[0].value
+        assert isinstance(offset_dim, str)
+        offset_node = self.visit(tail[1])[0]
 
-        if isinstance(tail[1], itir.OffsetLiteral):
-            offset_node = self.visit_OffsetLiteral(tail[1])
-        else:
-            offset_node = self.visit(tail[1])[0]
-
-        if isinstance(self.offset_provider[offset], NeighborTableOffsetProvider):
-            offset_provider = self.offset_provider[offset]
-            connectivity = self.context.state.add_access(connectivity_identifier(offset))
+        if isinstance(self.offset_provider[offset_dim], NeighborTableOffsetProvider):
+            offset_provider = self.offset_provider[offset_dim]
+            connectivity = self.context.state.add_access(connectivity_identifier(offset_dim))
 
             shifted_dim = offset_provider.origin_axis.value
             target_dim = offset_provider.neighbor_axis.value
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -698,9 +698,20 @@ def _visit_indirect_addressing(self, node: itir.FunCall) -> IteratorExpr:
...
+            assert isinstance(element, int)
+            element_var = unique_var_name()
+            self.context.body.add_scalar(element_var, dace.dtypes.int64, transient=True)
+            element_node = self.context.state.add_access(element_var)
+            tlet_node = self.context.state.add_tasklet(
+                "get_element", {}, {"__out"}, f"__out = {element}"
+            )
+            self.context.state.add_edge(
+                tlet_node, "__out", element_node, None, dace.Memlet.simple(element_var, "0")
+            )
```
petiaccja, line 712: Can you extract this into a new visitor `visit_OffsetLiteral`?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -712,20 +723,32 @@ def _visit_indirect_addressing(self, node: itir.FunCall) -> IteratorExpr:
...
+            expr = f"{internals[0]} * {offset_provider.max_neighbors} + {internals[1]}"
+        else:
+            assert isinstance(self.offset_provider[offset], Dimension)
+            shifted_dim = target_dim = self.offset_provider[offset].value
+            args = [
+                ValueExpr(iterator.indices[shifted_dim], dace.int64),
+                element_node,
+            ]
+            internals = [f"{arg.value.data}_v" for arg in args]
+            expr = f"{internals[0]} + {internals[1]}"
```
petiaccja, line 751: It seems to me that this part of the code duplicates the functionality of `_visit_direct_addressing`. The dispatch between the two is made line 488. Due to the dispatch, this code will never actually be called.

You can leave all the different offset providers in one function, like this, delete the `_visit_direct_addressing`, and remove the dispatch, or you can add a `_visit_strided_addressing` function and extend the dispatch. Both would solve the issue.

petiaccja (changes requested):

edopao:
cscs-ci run

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -667,68 +655,50 @@ def _visit_direct_addressing(self, node: itir.FunCall) -> IteratorExpr:
...
-        return IteratorExpr(iterator.field, shifted_index, iterator.dtype, iterator.dimensions)
 
-    def _visit_indirect_addressing(self, node: itir.FunCall) -> IteratorExpr:
-        shift = node.fun
-        assert isinstance(shift, itir.FunCall)
-        tail, rest = self._split_shift_args(shift.args)
-        if rest:
-            iterator = self.visit(self._make_shift_for_rest(rest, node.args[0]))
+        if isinstance(tail[1], itir.OffsetLiteral):
+            offset_node = self.visit_OffsetLiteral(tail[1])
```
petiaccja, line 660: The reason **all** `visit_*` methods return a `list[ValueExpr]` is because that seemed like the most sane way of dealing with tuples. It's not so nice, but necessary when translating from ASTs to DAGs while eliminating tuples.

If you made `visit_OffsetLiteral` return a `list[ValueExpr]`, even if the list's size is always one, you could get rid of this `if` statement and the explicit call to the node visitor as `visit(tail[1])[0]` would cover all cases.

petiaccja (changes requested):
I have one comment on the new `visit_OffsetLiteral`, otherwise I like the way you restructured the `_visit_shift`, and the PR is good to go IMO.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: That makes sense, thank you!

PR #1354, opened 2023-10-27
fix[next]: Improvements in DaCe backend

edopao:
## Description

This PR contains some fixes and code refactoring in DaCe backend:
- (refactoring) Use memlet API for full array subset
- Fix for symbolic analysis of memlet volume: define symbols before visiting the closure domain in order to allow symbolic analysis of memlet volume
- Fix for gpu execution: import cupy for sorting of field dimensions.

Each commit can be reviewed separately.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
@@ -31,6 +31,12 @@ from .itir_to_sdfg import ItirToSDFG
 from .utility import connectivity_identifier, filter_neighbor_tables, get_sorted_dims
 
 
+try:
+    import cupy as cp
+except ImportError:
+    cp = None
+
+
 def get_sorted_dim_ranges(domain: Domain) -> Sequence[UnitRange]:
     sorted_dims = get_sorted_dims(domain.dims)
     return [domain.ranges[dim_index] for dim_index, _ in sorted_dims]
@@ -44,19 +50,16 @@ _cpu_args = (
 )
 
 
-def convert_arg(arg: Any, run_on_gpu: bool):
+def convert_arg(arg: Any):
     if is_field(arg):
         sorted_dims = get_sorted_dims(arg.domain.dims)
         ndim = len(sorted_dims)
         dim_indices = [dim_index for dim_index, _ in sorted_dims]
-        if run_on_gpu:
-            import cupy as cp
-
-            assert isinstance(arg.ndarray, cp.ndarray)
-            return cp.moveaxis(arg.ndarray, range(ndim), dim_indices)
-        else:
-            assert isinstance(arg.ndarray, np.ndarray)
+        if isinstance(arg.ndarray, np.ndarray):
             return np.moveaxis(arg.ndarray, range(ndim), dim_indices)
+        else:
+            assert cp is not None and isinstance(arg.ndarray, cp.ndarray)
+            return cp.moveaxis(arg.ndarray, range(ndim), dim_indices)
     return arg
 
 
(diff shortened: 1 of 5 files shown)


src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -44,13 +44,19 @@ def get_sorted_dim_ranges(domain: Domain) -> Sequence[UnitRange]:
...
-        assert isinstance(arg.ndarray, np.ndarray)
-        return np.moveaxis(arg.ndarray, range(ndim), dim_indices)
+        if run_on_gpu:
+            import cupy as cp
+
+            assert isinstance(arg.ndarray, cp.ndarray)
+            return cp.moveaxis(arg.ndarray, range(ndim), dim_indices)
+        else:
+            assert isinstance(arg.ndarray, np.ndarray)
+            return np.moveaxis(arg.ndarray, range(ndim), dim_indices)
```
petiaccja, line 59: I think it's better to keep `convert_arg` independent of the hardware acceleration. Could you rewrite `convert_arg` in a way such that it no longer expects a `run_on_gpu` flag, but rather just simply recognizes the argument and converts it accordingly? You can first check if it's a numpy array, and if it not, you can import cupy in a `try` block, and handle the `ImportError` or return the processed cupy array. This would save you from passing the `run_on_gpu` flag down the entire call stack.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -64,8 +70,8 @@ def preprocess_program(program: itir.FencilDefinition, offset_provider: Mapping[
     return program
 
 
-def get_args(params: Sequence[itir.Sym], args: Sequence[Any]) -> dict[str, Any]:
-    return {name.id: convert_arg(arg) for name, arg in zip(params, args)}
+def get_args(params: Sequence[itir.Sym], args: Sequence[Any], run_on_gpu: bool) -> dict[str, Any]:
+    return {name.id: convert_arg(arg, run_on_gpu) for name, arg in zip(params, args)}
```
petiaccja, line 74: You could eliminate the `run_on_gpu` flag here.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -195,7 +201,7 @@ def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs) -> None:
         if build_cache is not None:
             build_cache[cache_id] = sdfg_program
 
-    dace_args = get_args(program.params, args)
+    dace_args = get_args(program.params, args, run_on_gpu)
```
petiaccja, line 204: You could eliminate the `run_on_gpu` flag here as well.

petiaccja (changes requested):

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
edopao: Good idea. I have also elaborated it a bit more.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -226,24 +235,22 @@ def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs) -> None:
...
+def run_dace(program: itir.FencilDefinition, *args, **kwargs) -> None:
+    run_on_gpu = any(not isinstance(arg.ndarray, np.ndarray) for arg in args if is_field(arg))
+    if run_on_gpu:
+        if cp is None:
+            raise RuntimeError(
+                f"Non-numpy field argument passed to program {program.id} but module cupy not installed"
+            )
 
+        if not all(isinstance(arg.ndarray, cp.ndarray) for arg in args if is_field(arg)):
+            raise RuntimeError("Execution on GPU requires all fields to be stored as cupy arrays")
```
petiaccja, line 247: Yeah, I like this direction much better. We could do this in a nicer way, perhaps using the array API, because this won't really scale if we want to add support for JAX, tensorflow, PyTorch and the like's own arrays. Anyways, that I think is outside the scope of this PR.

petiaccja (approved):

PR #1355, opened 2023-11-03
feat[next]: Limit use of global type inference in CollapseTuple pass 

nfarabullini:
Make `CollapseTuple` pass configurable such that whether the global ITIR type inference is used to get the tuple size or the simple local inference can be configured using a boolean flag.

Execute with global ITIR type inference once in loop in pass manager and in all subsequent runs use the local inference. This significantly improves the performance of the toolchain.

### DIFF
--- a/src/gt4py/next/iterator/transforms/pass_manager.py
+++ b/src/gt4py/next/iterator/transforms/pass_manager.py
@@ -82,7 +82,7 @@ def apply_common_transforms(
     unconditionally_collapse_tuples=False,
 ):
     if lift_mode is None:
-        lift_mode = LiftMode.FORCE_INLINE
+        lift_mode = LiftMode.FORCE_TEMPORARIES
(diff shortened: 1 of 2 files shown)


src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -82,15 +82,15 @@ def apply_common_transforms(
     unconditionally_collapse_tuples=False,
 ):
     if lift_mode is None:
-        lift_mode = LiftMode.FORCE_INLINE
+        lift_mode = LiftMode.FORCE_TEMPORARIES
```
tehrengruber, line 85:
```suggestion
        lift_mode = LiftMode.FORCE_INLINE
```
This was only meant as a test to see how the pipeline performs.

tehrengruber (changes requested):
Only style suggestions.

tehrengruber (approved):

PR #1356, opened 2023-11-07
feat[next]: DaCe support for can_deref

edopao:
## Description

This small PR adds support for `can_deref` operator in DaCe backend. It also improves the code for preprocess ITIR transformations.


src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -154,11 +169,14 @@ def get_cache_id(
...
     auto_optimize = kwargs.get("auto_optimize", False)
+    build_cache = kwargs.get("build_cache", None)
     build_type = kwargs.get("build_type", "RelWithDebInfo")
     run_on_gpu = kwargs.get("run_on_gpu", False)
-    build_cache = kwargs.get("build_cache", None)
     # ITIR parameters
     column_axis = kwargs.get("column_axis", None)
+    lift_mode = (
+        LiftMode.FORCE_INLINE
+    )  # TODO(edopao): make it configurable once DaCe backend supports temporaries
```
petiaccja, line 179:
```suggestion
    lift_mode = LiftMode.FORCE_INLINE  # TODO(edopao): make it configurable once DaCe backend supports temporaries
```

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -63,14 +63,29 @@ def convert_arg(arg: Any):
...
+    if all([ItirToSDFG._check_no_lifts(closure) for closure in node.closures]):
+        fencil_definition = node
+    else:
+        fencil_definition = apply_common_transforms(
+            program,
+            common_subexpression_elimination=False,
+            lift_mode=lift_mode,
+            offset_provider=offset_provider,
+            unroll_reduce=True,
+        )
```
petiaccja, line 87: Can you explain why you don't unroll in the first place?

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -171,7 +189,7 @@ def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs) -> None:
         sdfg = sdfg_program.sdfg
     else:
         # visit ITIR and generate SDFG
-        program = preprocess_program(program, offset_provider)
+        program = preprocess_program(program, offset_provider, lift_mode)
```
petiaccja, line 192: This looks good! One note is that regarding DaCe I would probably always create a temporary for all lifts. Since eliminating temporaries is one of the key optimizations, I would find it interesting to see how different approaches deal with it, and if you don't give the chance to DaCe to do this optimization we'll never know.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -265,6 +265,29 @@ def builtin_neighbors(
...
+    # first visit shift, to get set of indices for deref
+    can_deref_callable = node_args[0]
+    assert isinstance(can_deref_callable, itir.FunCall)
+    shift_callable = can_deref_callable.fun
+    assert isinstance(shift_callable, itir.FunCall)
+    assert isinstance(shift_callable.fun, itir.SymRef)
+    assert shift_callable.fun.id == "shift"
+    iterator = transformer._visit_shift(can_deref_callable)
+
+    # create tasklet to check that field indices are 0-based (-1 is invalid)
```
petiaccja, line 280:
```suggestion
    # create tasklet to check that field indices are non-negative (-1 is invalid)
```

petiaccja (changes requested):

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
edopao: That's the auto-formatter

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
edopao: This is the approach used in GTFN backend. We could play with this option to see what SDFG representation is most suitable for DaCe optimizations. The non-unrolled version produces a more compact SDFG.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
edopao: I agree, once temporaries work we can make it the default option.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: I can fix this, if there are other changes.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
petiaccja: I think my question was more like, can you always generate the DaCe SDFG with `unroll_reduce=False`? Because you're first running with `False`, then check for lifts, then run with `True`, which does not look too robust.

As for optimizations, I agree with you, we can play around later.

edopao:
cscs-ci run

petiaccja (approved):

PR #1357, opened 2023-11-14
feat[next] Enable GPU backend tests

havogt:


src/gt4py/next/constructors.py
```diff
@@ -189,7 +189,10 @@ def full(
         allocator=allocator,
         device=device,
     )
-    field[...] = field.dtype.scalar_type(fill_value)
+    if hasattr(fill_value, "__array__"):
+        field[...] = field.array_ns.asarray(fill_value)
+    else:
+        field[...] = field.dtype.scalar_type(fill_value)
```
egparedes, line 195: I don't think this is the right solution. `full`, by definition, should only accept scalar fill values. If the provided `fill_value` is an ndarray and not a scalar, then the `as_field()` function should be used.

havogt:
I reviewed @DropD 's changes. Did you review mine?

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -126,7 +139,7 @@ def compilation_hash(otf_closure: stages.ProgramCall) -> int:
     translation=GTFN_GPU_TRANSLATION_STEP,
     bindings=nanobind.bind_source,
     compilation=GTFN_DEFAULT_COMPILE_STEP,
-    decoration=convert_args,
+    decoration=functools.partial(convert_args, for_cp=True),
```
havogt, line 142: use device instead of for_cp

tests/next_tests/integration_tests/cases.py
```diff
@@ -133,6 +133,13 @@ def field(
         sizes: dict[gtx.Dimension, int],
         dtype: np.typing.DTypeLike,
     ) -> FieldValue:
+        if hasattr(self.value, "__array__"):
+            return constructors.as_field(
+                common.domain(sizes),
+                np.full(tuple(sizes.values()), self.value),
+                dtype=dtype,
+                allocator=backend,
+            )
```
havogt, line 142: which case triggers that actually?

DropD (approved):

tests/next_tests/integration_tests/cases.py
DropD: The cases where an array of boolean values is used as the value for ConstInitializer. Example: `test_gt4py_builtins.test_conditional_promotion`

src/gt4py/next/program_processors/runners/gtfn.py
DropD: You mean rename the argument?

DropD: Or also change the type to `DeviceType` and change the logic accordingly?

havogt: sorry, that note was for myself. Enrique, suggested instead of using the boolean, and do the same logic if DeviceType==GPU

tests/next_tests/integration_tests/cases.py
havogt: I don't understand that use case. What's the semantic? promote to 1 dimension more?

DropD: It seems the existing use cases wer all erroneous and were not actually using arrays as point values.

src/gt4py/next/program_processors/runners/gtfn.py
DropD: I can quickly do that.

tests/next_tests/integration_tests/cases.py
```diff
@@ -117,12 +117,19 @@ def from_case(
...
 class ConstInitializer(DataInitializer):
     """Initialize with a given value across the coordinate space."""
 
     value: ScalarValue
 
+    def __init__(self, value: ScalarValue):
+        if hasattr(value, "__array__") or hasattr(value, "__getitem__"):
+            raise ValueError(
+                "`ConstInitializer` can not be used with non-scalars. Use `Case.as_field` instead."
+            )
```
havogt, line 130:
```suggestion
        if not core_defs.is_scalar_type(value):
            raise ValueError(
                "`ConstInitializer` can not be used with non-scalars. Use `Case.as_field` instead."
            )
```

tests/next_tests/integration_tests/cases.py
```diff
@@ -117,12 +117,19 @@ def from_case(
...
 
 
-@dataclasses.dataclass
+@dataclasses.dataclass(init=False)
 class ConstInitializer(DataInitializer):
     """Initialize with a given value across the coordinate space."""
 
     value: ScalarValue
 
+    def __init__(self, value: ScalarValue):
```
havogt, line 126:
```suggestion
    def __init__(self, value: core_defs.Scalar):
```

havogt: and remove `ScalarValue` completely

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_gt4py_builtins.py
```diff
@@ -137,10 +138,7 @@ def conditional_nested_tuple(
         return where(mask, ((a, b), (b, a)), ((5.0, 7.0), (7.0, 5.0)))
 
     size = cartesian_case.default_sizes[IDim]
-    bool_field = np.random.choice(a=[False, True], size=(size))
-    mask = cases.allocate(cartesian_case, conditional_nested_tuple, "mask").strategy(
-        cases.ConstInitializer(bool_field)
-    )()
+    mask = cartesian_case.as_field([IDim], np.random.choice(a=[False, True], size=size))
```
havogt, line 141: could we built in to cases that boolean fields are allocated with alternating values by default?

havogt: maybe in a follow-up PR to get the GPU support in

tests/next_tests/integration_tests/cases.py
havogt: I changed that...

tests/next_tests/integration_tests/cases.py
havogt: ... but kept `ScalarValue` pointing to `core_defs.Scalar` as this seems to be the pattern that is used in this field

havogt:
will merge now as no one complained about my last changes...

PR #1360, opened 2023-11-16
feat[next]: DaCe backend - enable GPU test execution

edopao:
## Description

- Small fix for GPU execution. The fix is about handling of in/out fields, for which the program argument is copied to a transient array. We need to use for the transient array the same storage as the program argument (i.e. gpu storage), otherwise code generation will throw an error because of mixed storage for inputs to the closure map.
- Enable test execution with DACE_GPU backend
- Minor code refactoring (test exclusion matrix, dace backend processor interface)
- Cleanup test exclusion matrix (some left-overs after rebase of previous dace PR)


kotsaloscv (approved):
Looks good to me.

PR #1367, opened 2023-11-20
fix[next]: DaCe backend - symbol propagation in lambda scope

edopao:
## Description

Full-inling of ITIR `lift` operator can result in nested lambda calls, which are translated to nested SDFGs in DaCe backend. The problem was that inner lambda SDFGs were not inheriting the symbols in scope from the parent tasklet, instead the DaCe backend was only mapping the lambda arguments.
The solution implemented in this PR is to run a pre-pass to discover all symbols used in the nested lambdas, and propagate the required data containers from outer to inner scope.

The legacy `visit_SymRef` function is also cleaned up. In baseline code, it was supposed to create access nodes whenever a symbol was used. With this PR, this operation is delegated to a pre-pass.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
@@ -11,7 +11,7 @@
 # distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
-import copy
+
 import dataclasses
 import itertools
 from collections.abc import Sequence
@@ -23,7 +23,6 @@ from dace.transformation.dataflow import MapFusion
 from dace.transformation.passes.prune_symbols import RemoveUnusedSymbols
 
 import gt4py.eve.codegen
-from gt4py import eve
 from gt4py.next import Dimension, StridedNeighborOffsetProvider, type_inference as next_typing
 from gt4py.next.iterator import ir as itir, type_inference as itir_typing
 from gt4py.next.iterator.embedded import NeighborTableOffsetProvider
@@ -156,7 +155,6 @@ class IteratorExpr:
 class Context:
     body: dace.SDFG
     state: dace.SDFGState
-    # TODO(edopao): use a stack for symbols in different nested scope, needed for lambdas
     symbol_map: dict[str, IteratorExpr | ValueExpr | SymbolExpr]
     # if we encounter a reduction node, the reduction state needs to be pushed to child nodes
     reduce_limit: int
@@ -167,64 +165,12 @@ class Context:
         body: dace.SDFG,
         state: dace.SDFGState,
         symbol_map: dict[str, IteratorExpr | ValueExpr | SymbolExpr],
-        **kwargs,
     ):
         self.body = body
         self.state = state
         self.symbol_map = symbol_map
-        self.reduce_limit = kwargs.get("reduce_limit", 0)
-        self.reduce_wcr = kwargs.get("reduce_wcr", None)
-
-    def get_scope_symbols(self) -> dict[str, IteratorExpr | ValueExpr | SymbolExpr]:
-        # TODO(edopao): flatten symbols across nested scopes
-        return self.symbol_map
-
-    def push_scope(self, symbols: dict[str, IteratorExpr | ValueExpr | SymbolExpr]):
-        # TODO(edopao)
-        pass
-
-    def pop_scope(self) -> dict[str, IteratorExpr | ValueExpr | SymbolExpr]:
-        # TODO(edopao)
-        pass
-
-
-class GatherSymbolsPass(eve.NodeVisitor):
-    sdfg: dace.SDFG
-
-    def __init__(
-        self,
-            sdfg: dace.SDFG,
-            ctx: Context
-    ):
-        self.sdfg = sdfg
-        self.ctx = ctx
-
-    def visit_SymRef(self, node: itir.SymRef):
-        symbol_name = str(node.id)
-        scope_symbols = self.ctx.get_scope_symbols()
-        if (symbol_name in scope_symbols):
-            symbol_value = scope_symbols[symbol_name]
-        else:
-            dtype = self.node_types[id(node)]
-            symbol_value = SymbolExpr(symbol_name, dtype)
-
-        if isinstance(symbol_value, ValueExpr):
-            self.sdfg.add_scalar(symbol_name, dtype=symbol_value.dtype)
-        elif isinstance(symbol_value, IteratorExpr):
-            ndims = len(symbol_value.dimensions)
-            shape = tuple(
-                dace.symbol(unique_var_name() + "__shp", dace.int64) for _ in range(ndims)
-            )
-            strides = tuple(
-                dace.symbol(unique_var_name() + "__strd", dace.int64) for _ in range(ndims)
-            )
-            self.sdfg.add_array(symbol_name, shape=shape, strides=strides, dtype=symbol_value.dtype)
-            index_names = {dim: f"__{symbol_name}_i_{dim}" for dim in symbol_value.indices.keys()}
-            for _, index_name in index_names.items():
-                self.sdfg.add_scalar(index_name, dtype=dace.int64)
-
-    def visit_Lambda(self, node: itir.Lambda):
-        return
+        self.reduce_limit = 0
+        self.reduce_wcr = None
 
 
 def builtin_neighbors(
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -350,6 +351,104 @@ def builtin_undefined(*args: Any) -> Any:
...
+        param = str(node.id)
+        if param not in _GENERAL_BUILTIN_MAPPING and param not in self.symbol_map:
+            node_type = self.node_types[id(node)]
+            assert isinstance(node_type, Val)
+            access_node = self.state.add_access(param)
+            self.symbol_map[param] = ValueExpr(
+                access_node, dtype=itir_type_as_dace_type(node_type.dtype)
+            )
+
+    def visit_FunCall(self, node: itir.FunCall):
```
edopao, line 447: Try to remove `visit_FunCall` and use default visitor.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -350,6 +351,104 @@ def builtin_undefined(*args: Any) -> Any:
...
+            assert isinstance(arg, SymbolExpr)
+            self.symbol_map[param] = arg
+
+    def visit_SymRef(self, node: itir.SymRef):
+        name = str(node.id)
+        if name in self.parent_symbol_map and name not in self.symbol_map:
+            arg = self.parent_symbol_map[name]
+            self._add_symbol(name, arg)
+
+    def visit_FunCall(self, node: itir.FunCall):
```
edopao, line 407: Try to remove `visit_FunCall`  and use default visitor.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -165,12 +165,13 @@ def __init__(
         body: dace.SDFG,
         state: dace.SDFGState,
         symbol_map: dict[str, IteratorExpr | ValueExpr | SymbolExpr],
+        **kwargs,
```
petiaccja, line 168: I'd prefer defaulted parameters instead of `**kwargs`:
```suggestion
        reduce_limit = 0,
        reduce_wcr = None,
```

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -350,6 +351,104 @@ def builtin_undefined(*args: Any) -> Any:
...
+    sdfg: dace.SDFG
+    state: dace.SDFGState
+    symbol_map: dict[str, IteratorExpr | SymbolExpr | ValueExpr]
+    parent_symbol_map: dict[str, IteratorExpr | SymbolExpr | ValueExpr]
+
+    def __init__(
+        self,
+        sdfg,
+        state,
+        symbol_map,
```
petiaccja, line 364: You pass `symbol_map` as a `&` to the constructor and in the code after calling `visit` you rely on the fact that `GatherLambdaSymbolsPass` will modify the original `symbol_map`. If `GatherLambdaSymbolsPass` made a (shallow) copy of your `symbol_map`, your code would break despite no API changes as the user of the *pass* would not see the changes to the dict. I think it'd better to remove this parameter from the ctor, create an empty dict inside the ctor, and retrieve the filled dict through a property after calling `visit`.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -350,6 +351,104 @@ def builtin_undefined(*args: Any) -> Any:
...
+            for _, index_name in index_names.items():
+                self.sdfg.add_scalar(index_name, dtype=dace.int64)
+            # update table of lambda symbol
+            field = self.state.add_access(param)
+            indices = {
+                dim: self.state.add_access(index_arg) for dim, index_arg in index_names.items()
+            }
+            self.symbol_map[param] = IteratorExpr(field, indices, arg.dtype, arg.dimensions)
+        else:
+            assert isinstance(arg, SymbolExpr)
```
petiaccja, line 398: +1 for defensive coding

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -350,6 +351,104 @@ def builtin_undefined(*args: Any) -> Any:
...
+        name = str(node.id)
+        if name in self.parent_symbol_map and name not in self.symbol_map:
+            arg = self.parent_symbol_map[name]
+            self._add_symbol(name, arg)
+
+    def visit_FunCall(self, node: itir.FunCall):
+        self.visit(node.args)
+        self.visit(node.fun)
+
+    def visit_Lambda(self, node: itir.Lambda, args=None):
```
petiaccja, line 411: Could you add type annotations for `args`?

I'm also thinking if we should make this parameter mandatory and pass an empty list/dict/whatever it is.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -369,70 +468,46 @@ def visit_FunctionDefinition(self, node: itir.FunctionDefinition, **kwargs):
...
-        context.reduce_wcr = prev_context.reduce_wcr
-        self.context = context
+        connectivity_names = [connectivity_identifier(offset) for offset, _ in neighbor_tables]
+
+        # Create the SDFG for the lambda's body
+        lambda_sdfg = dace.SDFG(func_name)
+        lambda_state = lambda_sdfg.add_state(f"{func_name}_entry", True)
 
-        # Add input parameters as arrays
+        lambda_symbol_map: dict[str, IteratorExpr | SymbolExpr | ValueExpr] = {}
```
petiaccja, line 485: Here, you shouldn't rely on `GatherLambdaSymbolsPass` modifying the original `lambda_symbol_map ̇, its implementation may perform a copy and leave the original dict untouched.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -441,50 +516,53 @@ def visit_Lambda(
...
-        self.context = prev_context
-        for node in context.state.nodes():
-            if isinstance(node, dace.nodes.AccessNode):
-                if context.state.out_degree(node) == 0 and context.state.in_degree(node) == 0:
-                    context.state.remove_node(node)
+        # remove isolated access nodes for connectivity arrays not consumed by lambda
+        for sub_node in lambda_state.nodes():
+            if isinstance(sub_node, dace.nodes.AccessNode):
+                if lambda_state.out_degree(sub_node) == 0 and lambda_state.in_degree(sub_node) == 0:
+                    lambda_state.remove_node(sub_node)
```
petiaccja, line 559: Is this still necessary, or are we now only passing symbols or access nodes that are actually used?

petiaccja (changes requested):

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: Right, it makes sense

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: I agree

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: Unfortunately it is still necessary because 1. we pass all connectivity arrays to the lambda and 2. the lambda arguments are always added to the symbol map, without checking if they are used.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: I have added the type annotation for now. I will investigate the other proposal.

edopao: We need to keep `args` defaulted to `None`. We could enter this function either from the root lambda node, for which we have argument nodes, or from a nested lambda, for which we do not want to pass the arguments because the argument nodes for a nested lambda should not appear in the symbol map of the root lambda.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
petiaccja: Shame, I don't like this code, but it'll be like this for now then.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
petiaccja: Okay, I see

PR #1369, opened 2023-11-24
test[next]: Fix warnings that cause Spack to crash

edopao:
## Description

Solves warning about invalid escape sequence in some regex strings. These warnings cause spack build to crash.

### DIFF
--- a/tests/next_tests/unit_tests/ffront_tests/test_func_to_past.py
+++ b/tests/next_tests/unit_tests/ffront_tests/test_func_to_past.py
@@ -206,7 +206,9 @@ def test_domain_exception_3(identity_def):
     assert exc_info.match("Invalid call to `domain_format_3`")
 
     assert (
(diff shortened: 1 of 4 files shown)


tests/next_tests/unit_tests/ffront_tests/test_func_to_past.py
```diff
@@ -206,7 +206,9 @@ def domain_format_3_program(in_field: gtx.Field[[IDim], float64]):
     assert exc_info.match("Invalid call to `domain_format_3`")
 
     assert (
-        re.search("Missing required keyword argument\(s\) `out`.", exc_info.value.__cause__.args[0])
+        re.search(
```
egparedes, line 209: Just a question: why has the format of this line been changed? It looks exactly the same as the change below which kept the same format...

egparedes (approved):
It looks good to me.

PR #1372, opened 2023-11-28
feat[next]: Enable tests for embedded with cupy

havogt:
Introduces mechanism in tests for having different allocators for the same (`None`) backend.

Fixes:
- The resulting buffer for scan is deduced from the buffer type of the arguments, if there are no arguments we fallback to numpy (maybe break). We need to find a mechanism for this corner case. Currently these tests are excluded with `pytest.mark.uses_scan_without_field_args` for cupy embedded execution.

Refactoring:
- make common.field and common.connectivity private
- rename next_tests.exclusion_matrices to definitions

TODOs for later:
- `broadcast` of scalar ignores the broadcast


tests/next_tests/integration_tests/cases.py
```diff
@@ -482,17 +482,18 @@ def verify_with_default_data(
 @pytest.fixture
 def cartesian_case(fieldview_backend):  # noqa: F811 # fixtures
     yield Case(
-        fieldview_backend,
+        fieldview_backend if isinstance(fieldview_backend, ppi.ProgramExecutor) else None,
```
havogt, line 485: this is weird and dangerous, let's discuss a better strategy

src/gt4py/next/common.py
```diff
@@ -844,7 +844,7 @@ def is_connectivity_field(
 
 
 @functools.singledispatch
-def field(
+def _field(
```
havogt, line 847: add comment for what this is used

src/gt4py/next/embedded/operators.py
```diff
@@ -141,10 +150,21 @@ def _intersect_scan_args(
...
 
 
-def _construct_scan_array(domain: common.Domain):
+def _get_array_ns(
+    *args: core_defs.Scalar | common.Field | tuple[core_defs.Scalar | common.Field | tuple, ...]
+) -> ModuleType:
+    for arg in utils.flatten_nested_tuple(args):
+        if hasattr(arg, "array_ns"):
+            return arg.array_ns
+    return np
```
havogt, line 159: instead of returning np, maybe the save thing would be to error for now.

docs/development/ADRs/0015-Test_Exclusion_Matrices.md
```diff
@@ -1,13 +1,13 @@
 ---
-tags: []
+tags: [testing]
 ---
 
 # Test-Exclusion Matrices
 
 - **Status**: valid
 - **Authors**: Edoardo Paone (@edopao), Enrique G. Paredes (@egparedes)
```
egparedes, line 8: Should you add your name here somehow?

src/gt4py/next/common.py
```diff
@@ -844,18 +844,20 @@ def is_connectivity_field(
...
     definition: Any,
     /,
     *,
     domain: Optional[DomainLike] = None,
     dtype: Optional[core_defs.DType] = None,
 ) -> Field:
+    # Utility function to construct a `Field` from different buffer representations.
+    # Consider removing this function and using `Field` constructor directly.
```
egparedes, line 855: Put the comment above the function (and optionally mention also `_connectivity`)

tests/next_tests/definitions.py
```diff
@@ -55,6 +59,32 @@ class ProgramBackendId(_PythonObjectIdMixin, str, enum.Enum):
...
+
+    @property
+    def allocator(self) -> next_allocators.FieldBufferAllocatorProtocol:
+        ...
+
+
+@dataclasses.dataclass(frozen=True)
+class EmbeddedExecutionDescriptor:
+    allocator: next_allocators.FieldBufferAllocatorProtocol
+    executor: NoneType = dataclasses.field(default=None, init=False)
```
egparedes, line 76: I think it should be already enough to define the executor at the class level.
```suggestion
    executor: Final = None
```

egparedes (commented):
Minor comments and questions. It should be ready to merge in the next iteration.

docs/development/ADRs/0015-Test_Exclusion_Matrices.md
egparedes: :+1:

PR #1373, opened 2023-11-28
fix[cartesian]: DaceIR bridge for DaCe v0.15

FlorianDeconinck:
## Description

Going to the next major version of DaCe (v15) has led a couple of breakage.
The current codegen used a static naming for state struct. The default has changed and an option has been introduce in the dace config, we replicate those changes here.
Missing DaCe schedule to GT4Py schedule mapping.

Those changes will _not_ affect `next`.
Those changes were tested internally at NASA on the GEOS model.

### DIFF
--- a/src/gt4py/cartesian/backend/dace_backend.py
+++ b/src/gt4py/cartesian/backend/dace_backend.py
@@ -451,7 +451,7 @@ class DaCeComputationCodegen:
                 const int __I = domain[0];
                 const int __J = domain[1];
                 const int __K = domain[2];
-                ${name}_{state_prefix} dace_handle;
+                ${name}_${state_prefix} dace_handle;
                 ${backend_specifics}
                 auto allocator = gt::sid::cached_allocator(&${allocator}<char[]>);
                 ${"\\n".join(tmp_allocs)}
--- a/src/gt4py/cartesian/backend/dace_backend.py
+++ b/src/gt4py/cartesian/backend/dace_backend.py
@@ -575,7 +575,7 @@ class DaCeComputationCodegen:
             functor_args=self.generate_functor_args(sdfg),
             tmp_allocs=self.generate_tmp_allocs(sdfg),
             allocator="gt::cuda_util::cuda_malloc" if is_gpu else "std::make_unique",
-            state_prefix=dace_state_suffix,
+            state_suffix=dace_state_suffix,
         )
         generated_code = textwrap.dedent(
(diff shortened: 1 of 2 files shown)


src/gt4py/cartesian/backend/dace_backend.py
```diff
@@ -561,13 +561,21 @@ def apply(cls, stencil_ir: gtir.Stencil, builder: "StencilBuilder", sdfg: dace.S
         else:
             omp_threads = ""
             omp_header = ""
+
+        # Backward compatible state struct name change in dave >=0.15.x
```
gronerl, line 565:
```suggestion
        # Backward compatible state struct name change in dace >=0.15.x
```

gronerl (approved):
Make sure to fix squashed commit message to say v0.15, e.g. by fixing PR title (would make looking for the commit easier later).

src/gt4py/cartesian/backend/dace_backend.py
```diff
@@ -451,7 +451,7 @@ class DaCeComputationCodegen:
                 const int __I = domain[0];
                 const int __J = domain[1];
                 const int __K = domain[2];
-                ${name}_t dace_handle;
+                ${name}_${state_prefix} dace_handle;
```
havogt, line 454: any reason why you call it prefix instead of suffix?

src/gt4py/cartesian/backend/dace_backend.py
```diff
@@ -561,13 +561,21 @@ def apply(cls, stencil_ir: gtir.Stencil, builder: "StencilBuilder", sdfg: dace.S
         else:
             omp_threads = ""
             omp_header = ""
+
+        # Backward compatible state struct name change in dave >=0.15.x
+        try:
```
havogt, line 566: Would it be ok from your side if we drop support for dace < 0.15.0? If not, we can bump the minimum required version to 0.15.x

havogt:
cscs-ci run

src/gt4py/cartesian/backend/dace_backend.py
FlorianDeconinck: Not quite yet, unfortunately. I will log it as a task on our end for near-term when we flush out the first GEOS GPU version, then revisit this code and kill it.

FlorianDeconinck: Logged NASA side

havogt:
cscs-ci run

PR #1374, opened 2023-11-29
fix[next]: Proper calling signature in DaCe

philip-paul-mueller:
I fixed a bug in [Dace](https://github.com/spcl/dace/pull/1457), which caused some memory issue to surface, this commit addresses and fixes them.


edopao:
I am able to run all icon4py stencils on `dace_cpu` backend, using latest dace master branch, without this change.

philip-paul-mueller:
In the mean time I have realized what the issue is/was.
After I perform the translation (ITIR → SDFG) I am doing a deepcopy of the arguments. However, the fields use some special allocation policy (non minimal strides) but a copy will shrink them to the minimum. Finally when I make the call to benchmark it, I do not adapt these strides, which then leads to the out of bound access.

However, I think the change should still be merged, since:
- It allows to call the SDFG with positional argument, currently only keyword-only arguments are supported.
- It updates the names of the variables, so instead of just `__var15` it will now create `__var15_a_stride1` which tells you immediately that this variable is the second stride for variable `a`.

I will push soon a new commit that removes the copy.
Furthermore, we should also think about removing the two underscores from the variable names, since DaCe makes it clear that "everything starting with two underscore belongs to DaCe".

@edopao @kotsaloscv

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -184,12 +185,15 @@ def get_cache_id(
...
-def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs) -> None:
+def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs) -> Optional[dace.SDFG]:
     # build parameters
     auto_optimize = kwargs.get("auto_optimize", False)
     build_cache = kwargs.get("build_cache", None)
     build_type = kwargs.get("build_type", "RelWithDebInfo")
     run_on_gpu = kwargs.get("run_on_gpu", False)
+    # Return parameter
+    return_sdfg = kwargs.get('return_sdfg', False)
+    run_sdfg    = kwargs.get('run_sdfg', True)
```
havogt, line 196: I think the function should be split in one that creates the sdfg and one that runs, then there is no need for extra arguments.

philip-paul-mueller: I like that idea. Should I create an extra PR for that? Since everything else belongs to make a proper calling signature.

edopao:
Looks good, thank you.

philip-paul-mueller:
@edopao I have rewrite the comment.

edopao (approved):

PR #1375, opened 2023-11-30
refactor[next]: Unified type system

petiaccja:


src/gt4py/next/type_system_2/types.py
```diff
@@ -0,0 +1,257 @@
...
+    def is_signed(self) -> bool:
+        return True
+
+    def is_constructible_from(self, ty: Type) -> bool:
+        return isinstance(ty, IntegerType) or isinstance(ty, FloatType)
+
+    def is_implicitly_constructible_from(self, ty: Any) -> bool:
+        if isinstance(ty, IntegerType):
+            return ty.width < self.width
+        elif isinstance(ty, FloatType):
```
havogt, line 91: Note to myself: it seems here all the possible cases are explicitly listed, if a language needs extension for a custom type, how could we handle that?

src/gt4py/next/type_system_2/types.py
```diff
@@ -0,0 +1,257 @@
...
+            return ty.width <= self.width
```
havogt, line 92: At least for now, we want to be explicit of the floating point width, otherwise users might very easily cast to a larger floating point type without noticing.

src/gt4py/next/type_system_2/types.py
```diff
@@ -0,0 +1,257 @@
...
+                    return False, "too many positional arguments supplied"
+                param = positionals[arg.location]
+            else:
+                if arg.location not in keywords:
+                    return False, f"unexpected keyword argument '{arg.location}'"
+                param = keywords[arg.location]
+            if supplied[param]:
+                return False, f"argument '{param.name}' supplied multiple times"
+            if not traits.is_implicitly_convertible(arg.ty, param.ty):
+                return False, f"argument cannot be implicitly converted from '{arg.ty}' to '{param.ty}'  for parameter '{param.name}'"
```
havogt, line 176: In the current type deduction, we provided a full list of which args did or did not match the parameters. I think this is very useful, because this is one of the first errors users run into.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -166,7 +166,7 @@ def test_tuples(cartesian_case):  # noqa: F811 # fixtures
     @gtx.field_operator
     def testee(a: cases.IJKFloatField, b: cases.IJKFloatField) -> cases.IJKFloatField:
         inps = a, b
-        scalars = 1.3, float64(5.0), float64("3.4")
+        scalars = 1.3, float64(5.0), float64(3.4)
```
havogt, line 169: This change is ok, here, but we need the functionality in general (and hopefully there is another test covering this). The use-case is `float32("0.1")` which means take the closest float32 of the string literal, while `float32(0.1)` would create a float64 literal and then cast to float32, which is different (if not in this case, then you can google for the sequence where it matters).

src/gt4py/next/type_system_2/types.py
```diff
@@ -0,0 +1,257 @@
...
+
+from .traits import FunctionArgument
+
+
+class Type:
+    ...
+
+
+@dataclasses.dataclass
+class IntegerType(
```
havogt, line 13: just leaving it here as a possible extension: instead of having separate types used in the type system and at runtime, we could merge this stuff with the (d)types defined in `_core.definitions`.

src/gt4py/next/type_system_2/types.py
```diff
@@ -0,0 +1,257 @@
...
+@dataclasses.dataclass
+class BoolType(IntegerType):
+    def __init__(self):
+        super().__init__(1, False)
+
+
+@dataclasses.dataclass
+class Int8Type(IntegerType):
+    def __init__(self):
+        super().__init__(8, True)
```
havogt, line 197: Maybe this could just be instances
```suggestion
Int8Type = IntegerType(width=8, signed=True)
```

src/gt4py/next/type_system_2/traits.py
```diff
@@ -0,0 +1,115 @@
...
+from collections.abc import Sequence
+
+import dataclasses
+
+
+class Trait:
+    ...
+
+
+class SignednessTrait(Trait):
```
havogt, line 14: Not sure, but maybe `Protocol`s could be more appropriate for this pattern.

src/gt4py/next/type_system_2/traits.py
```diff
@@ -0,0 +1,115 @@
...
+class FromTrait(Trait):
+    @abc.abstractmethod
+    def is_constructible_from(self, ty: Any) -> bool:
+        ...
+
+
+class ToTrait(Trait):
+    @abc.abstractmethod
+    def is_convertible_to(self, ty: Any) -> bool:
+        ...
```
havogt, line 29: Note to myself: check why both directions exist? what happens if they contradict?

src/gt4py/next/type_system_2/traits.py
```diff
@@ -0,0 +1,115 @@
...
+
+class SignednessTrait(Trait):
+    @abc.abstractmethod
+    def is_signed(self) -> bool:
+        ...
+
+
+class FromTrait(Trait):
+    @abc.abstractmethod
+    def is_constructible_from(self, ty: Any) -> bool:
```
havogt, line 22:
```suggestion
    def is_constructible_from(self, ty: types.Type) -> bool:
```
probably, but you might have done this later anyway

src/gt4py/next/type_system_2/inference.py
```diff
@@ -0,0 +1,89 @@
...
+
+import numpy as np
+from . import types
+from typing import Any, Optional, Callable
+
+
+@dataclasses.dataclass
+class Pattern:
+    annotation: Optional[Callable[[TypeInferrer, Any], Optional[types.Type]]]
+    instance: Optional[Callable[[TypeInferrer, Any], Optional[types.Type]]]
```
havogt, line 14: note to myself: think about this pattern

src/gt4py/next/type_system_2/inference.py
```diff
@@ -0,0 +1,89 @@
+from __future__ import annotations
```
havogt, line 1: probably handling annotations should use the eve utilities, because there are often changes between Python versions and we need to canonicalize the representation

havogt (commented):
Some comments, some to myself.

src/gt4py/next/type_system_2/types.py
petiaccja: There is also `is_implicitly_convertible_to`, which allows you to make A constructible from B without modifying A.

src/gt4py/next/type_system_2/types.py
petiaccja: The purpose of this function is to express if a type can be losslessly converted to another. The IR does not have to use this, it can also use equality comparison if we want to disallow lossless implicit type conversions as well.

src/gt4py/next/type_system_2/types.py
petiaccja: We don't currently have a mechanism for displaying batched error messages or notes attached to compiler messages. If we did, I agree that it's a good idea to display error messages eagerly, but until than, I'd rather not make a mess to bolt on a feature that has no proper foundation.

src/gt4py/next/type_system_2/traits.py
petiaccja: I don't really know how to avoid the circular import with `types.py` and `traits.py`, but yes, it should definitely not be `Any`.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
petiaccja: It's possible to add a `StringType`, but since this is the one and only case where it would be used, it may be better to address this in the lowering so as to not litter the type system and mislead people into thinking there are actual strings in the DSL. Agreed, adding a `StringType` is probably cleaner in code.

If I understand correctly, this difference shows up for less than 0.000000186264514923% of the numbers, and all it does is it flips the direction in which a 0.5 ULP error is introduced, which won't make any difference in the overwhelming majority of cases.

The return on investment on this feature is just so low that I don't see it worth to spend time on it. IMO we should focus on much more pressing issues, like the design of the core IRs.

src/gt4py/next/type_system_2/types.py
petiaccja: I don't know the purpose of the types in `_core.definitions`, so I don't have an opinion on this.

src/gt4py/next/type_system_2/types.py
petiaccja: Yes, it would be possible, but then I think I'd rather define instances like `int8 = IntegerType(8, True)`. We could also have both, but then people would mix them and that would likely lead to inconsistent code and confusion over which ones to use.

src/gt4py/next/type_system_2/traits.py
petiaccja: Do you have a minimal example of how you think we could use protocols, or any other constructs, for that matter? I haven't spent much time exploring other options but I can imagine another construct could be a better fit.

src/gt4py/next/type_system_2/traits.py
petiaccja: It's so that the conversion A -> B can be defined in the code of either A or B, but normally not both so that you avoid the contradictions. You'll choose the implementation site depending on whether modA imports modB, or the other way around. If modA imports modB, then you'd implement `A::is_convertible_to(B)`, otherwise you would implement `B::is_constructible_from(A)`.

PR #1377, opened 2023-11-30
feat[next][dace]: Support for sparse fields and reductions over lift expressions

edopao:
## Description

This PR adds support to DaCe backend for sparse fields and reductions over lift expressions.

The DaCe backend is configured with full-inlining of lifts in ITIR. An additional option is enabled, in combination with reduce-unroll, to fully inline lambda-like arguments. This option is required because the ITIR visitor in the DaCe backend is not able to translate reductions over lift expressions.

The inlined ITIR could contain the `list_get` operator. This PR adds translation capability for the`list_get` operator.

In icon4py stencils, `list_get` is proceeded by the `deref` operator, in order to extract a 1D slice out of a multi-dimensional field. Baseline only supported `deref` operator on full-domain index, that returns a scalar value. When the `deref` indexing is partial, `deref` will return a 1D or ND slice from the original field, and the return type should be array. Therefore, this PR also adds support for `deref` with array return.

## Requirements

- [x] #1370

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
@@ -289,8 +289,10 @@ def builtin_can_deref(
     assert shift_callable.fun.id == "shift"
     iterator = transformer._visit_shift(can_deref_callable)
 
+    # this iterator is accessing a neighbor table, so it should return an index
+    assert iterator.dtype in dace.dtypes.INTEGER_TYPES
     # create tasklet to check that field indices are non-negative (-1 is invalid)
-    args = [ValueExpr(access_node, _INDEX_DTYPE) for access_node in iterator.indices.values()]
+    args = [ValueExpr(access_node, iterator.dtype) for access_node in iterator.indices.values()]
     internals = [f"{arg.value.data}_v" for arg in args]
     expr_code = " and ".join([f"{v} >= 0" for v in internals])
 
@@ -844,6 +846,7 @@ class PythonTaskletCodegen(gt4py.eve.codegen.TemplatedGenerator):
         offset_dim = tail[0].value
         assert isinstance(offset_dim, str)
         offset_node = self.visit(tail[1])[0]
+        assert offset_node.dtype in dace.dtypes.INTEGER_TYPES
 
         if isinstance(self.offset_provider[offset_dim], NeighborTableOffsetProvider):
             offset_provider = self.offset_provider[offset_dim]
@@ -853,7 +856,7 @@ class PythonTaskletCodegen(gt4py.eve.codegen.TemplatedGenerator):
             target_dim = offset_provider.neighbor_axis.value
             args = [
                 ValueExpr(connectivity, offset_provider.table.dtype),
-                ValueExpr(iterator.indices[shifted_dim], _INDEX_DTYPE),
+                ValueExpr(iterator.indices[shifted_dim], offset_node.dtype),
                 offset_node,
             ]
             internals = [f"{arg.value.data}_v" for arg in args]
@@ -864,7 +867,7 @@ class PythonTaskletCodegen(gt4py.eve.codegen.TemplatedGenerator):
             shifted_dim = offset_provider.origin_axis.value
             target_dim = offset_provider.neighbor_axis.value
             args = [
-                ValueExpr(iterator.indices[shifted_dim], _INDEX_DTYPE),
+                ValueExpr(iterator.indices[shifted_dim], offset_node.dtype),
                 offset_node,
             ]
             internals = [f"{arg.value.data}_v" for arg in args]
@@ -875,14 +878,14 @@ class PythonTaskletCodegen(gt4py.eve.codegen.TemplatedGenerator):
             shifted_dim = self.offset_provider[offset_dim].value
             target_dim = shifted_dim
             args = [
-                ValueExpr(iterator.indices[shifted_dim], _INDEX_DTYPE),
+                ValueExpr(iterator.indices[shifted_dim], offset_node.dtype),
                 offset_node,
             ]
             internals = [f"{arg.value.data}_v" for arg in args]
             expr = f"{internals[0]} + {internals[1]}"
 
         shifted_value = self.add_expr_tasklet(
-            list(zip(args, internals)), expr, _INDEX_DTYPE, "shift"
+            list(zip(args, internals)), expr, offset_node.dtype, "shift"
         )[0].value
 
         shifted_index = {dim: value for dim, value in iterator.indices.items()}


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -726,14 +750,90 @@ def _visit_deref(self, node: itir.FunCall) -> list[ValueExpr]:
...
+            )
+            result_name = unique_var_name()
+            self.context.body.add_array(result_name, result_shape, iterator.dtype, transient=True)
+            result_array = self.context.body.arrays[result_name]
+            result_node = self.context.state.add_access(result_name)
+
+            deref_connectors = ["_inp"] + [
+                f"_i_{dim}" for dim in sorted_dims if dim in iterator.indices
+            ]
+            deref_nodes = [iterator.field] + [
```
edopao, line 779: The solution proposed here is to build a nested sdfg and map the index values to sdfg symbols. This allows inside the nested sdfg to slice the source field by specifying the subset in a memlet copy. The reason why we need a nested sdfg is because the memlet cannot access the scalar values directly, from the index nodes.

This solution is inefficient, unfortunately. It will probably be impossible for DaCe to inline this sdfg. Another solution which was considered is to use a map and copy the requested slice element-by-element from the source array. This solution does not require a nested sdfg, however it will probably be impossible for DaCe to optimize away the map and instead generate memory-copy along each slice dimension.

Having to decide between sdfg nesting and element-by-element copy, I chose the first one. SDFG nesting results in one additional level of function call, but it should preserve memory-copy in each slice dimension, at least from a representation point-of-view. From the dace templates used for code generation, it looks like the dynamic copy will anyway result in an element-by-element copy.
```
inline void deref_1_0_83(my_stencil_state_t *__state, const long long&  _i_Edge, double* __restrict__ _inp, double* __restrict__ _out, long long __var_272__shp, long long __var_274__strd, long long __var_275__strd) {
    long long _sym_i_Edge;

    _sym_i_Edge = _i_Edge;
    {
        dace::CopyNDDynamic<double, 1, false, 1>::template ConstDst<1>::Copy(
        _inp + (__var_275__strd * _sym_i_Edge), _out, __var_272__shp, __var_274__strd);
    }
}
```

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -131,6 +132,11 @@ def get_reduce_identity_value(op_name_: str, type_: Any):
...
 
 
+# Define type of index variables and symbols used in definitions for array shape/strides
+_INDEX_DTYPE = _TYPE_MAPPING["int"]
+_SYMBOL_DTYPE = _TYPE_MAPPING["int64"]
+
+
 @dataclasses.dataclass
 class SymbolExpr:
     value: str | dace.symbolic.sympy.Basic
```
philip-paul-mueller, line 142: Out of curiosity shouldn't it also be `dace.symbolic.SymExpr`?

edopao: Either `dace.symbolic.SymExpr` or the union type `dace.symbolic.SymbolicType`, but probably better `dace.symbolic.SymExpr`, as you suggest, because it wraps `sympy`.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -388,15 +411,15 @@ def _add_symbol(self, param, arg):
             # create storage in lambda sdfg
             ndims = len(arg.dimensions)
             shape = tuple(
-                dace.symbol(unique_var_name() + "__shp", dace.int64) for _ in range(ndims)
+                dace.symbol(unique_var_name() + "__shp", _SYMBOL_DTYPE) for _ in range(ndims)
```
philip-paul-mueller, line 414: You should switch to `unique_name(...)` further you should include the name and which stride it is meaning:
```
dace.symbol(unique_name(f"__{param}_shp{i}"), _SYMBOL_DTYPE) for i in range(ndims) 
```

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -226,7 +232,7 @@ def builtin_neighbors(
         outputs={"__result"},
     )
     idx_name = unique_var_name()
-    sdfg.add_scalar(idx_name, dace.int64, transient=True)
+    sdfg.add_scalar(idx_name, _INDEX_DTYPE, transient=True)
```
philip-paul-mueller, line 235: I am not sure if it is an issue, but here you go down with the precision, since `_TYPE_MAPPING["int"]` is [not necessarily 64bits](https://github.com/edopao/gt4py/blob/52f2abb24f3ee3007234c5638b6484b3076e12fd/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py#L52C30-L52C70). But it might be irrelevant.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -697,7 +721,7 @@ def _visit_deref(self, node: itir.FunCall) -> list[ValueExpr]:
                 for dim in sorted_dims
             ]
             args = [ValueExpr(iterator.field, iterator.dtype)] + [
-                ValueExpr(iterator.indices[dim], iterator.dtype) for dim in iterator.indices
+                ValueExpr(iterator.indices[dim], _INDEX_DTYPE) for dim in iterator.indices
```
philip-paul-mueller, line 724: Couldn't this be an issue? I would add some test if `iterator.dtype` is appropriate.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -726,14 +750,90 @@ def _visit_deref(self, node: itir.FunCall) -> list[ValueExpr]:
...
+        else:
+            # Not all dimensions are included in the deref index list:
+            # this means the ND-field will be sliced along one or more dimensions and the result will be an array
+            field_array = self.context.body.arrays[iterator.field.data]
+            result_shape = tuple(
+                dim_size
+                for dim, dim_size in zip(sorted_dims, field_array.shape)
+                if dim not in iterator.indices
+            )
+            result_name = unique_var_name()
```
philip-paul-mueller, line 771: I would give it a more expressiveness name if possible.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
philip-paul-mueller: Just for a reference:
- My [Gather](https://github.com/philip-paul-mueller/jax_to_dace/blob/philip.dev/JaxprToSDFG/translators/gatherTranslator.py) instruction that basically performs indirection, i.e. reads indexes from an array and performs the accesses in another array.
- My [Slicer](https://github.com/philip-paul-mueller/jax_to_dace/blob/c5db346ae9793bb226cd5e1ec5e4bab79de4bcf6/JaxprToSDFG/translators/slicingTranslator.py#L115).

The code above assumes fixed sizes at compile time, which is simpler.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -726,14 +750,90 @@ def _visit_deref(self, node: itir.FunCall) -> list[ValueExpr]:
...
+                dace.InterstateEdge(
+                    assignments={f"_sym{inp}": inp for inp in deref_connectors[1:]}
+                ),
+            )
+            # we access the size in source field shape as symbols set on the nested sdfg
+            source_subset = tuple(
+                f"_sym_i_{dim}" if dim in iterator.indices else f"0:{size}"
+                for dim, size in zip(sorted_dims, field_array.shape)
+            )
+            deref_access_state.add_edge(
```
philip-paul-mueller, line 808: you can use `add_nedge()` which includes the two `None` already.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -906,7 +1006,7 @@ def _visit_reduce(self, node: itir.FunCall):
 
             # initialize the reduction result based on type of operation
             init_value = get_reduce_identity_value(op_name.id, result_dtype)
-            init_state = self.context.body.add_state_before(self.context.state, "init")
+            init_state = self.context.body.add_state_before(self.context.state, "init", True)
```
philip-paul-mueller, line 1009: I know it does not belong to the PR, but why not using a library node?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -1076,8 +1177,10 @@ def closure_to_tasklet_sdfg(
             body.add_scalar(name, dtype=dtype)
             symbol_map[name] = ValueExpr(state.add_access(name), dtype)
     for arr, name in connectivities:
-        shape = [dace.symbol(f"{unique_var_name()}_shp{i}", dtype=dace.int64) for i in range(2)]
-        stride = [dace.symbol(f"{unique_var_name()}_strd{i}", dtype=dace.int64) for i in range(2)]
+        shape = [dace.symbol(f"{unique_var_name()}_shp{i}", dtype=_SYMBOL_DTYPE) for i in range(2)]
+        stride = [
+            dace.symbol(f"{unique_var_name()}_strd{i}", dtype=_SYMBOL_DTYPE) for i in range(2)
+        ]
```
philip-paul-mueller, line 1183: Same as above.

philip-paul-mueller (commented):
This is my first one. Further, I have one or two off topic questions, things that just pop up in my mind and would like to ask.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: I agree, I will change it to `dace.symbolic.SymbolicType`

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: I agree. I will create a utility function to do this, it could also be used in #1379

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: We are building a list of array indices. As per my understanding, `iterator.dtype` is the type of the field being accessed (typically float).

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: This is always very challenging.. it's indeed the result of deref :)

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: I'll use it, thanks.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: Do you mean for reduction. That's definitely something to explore, DaCe already provides it:
https://github.com/spcl/dace/blob/master/dace/libraries/standard/nodes/reduce.py

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: Done

edopao:
@philip-paul-mueller Thank you for the review, I have tried to address your comments. Please check again.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -513,14 +533,7 @@ def visit_Lambda(
 
         # Add connectivities as arrays
         for name in connectivity_names:
-            shape = (
-                dace.symbol(unique_var_name() + "__shp", dace.int64),
-                dace.symbol(unique_var_name() + "__shp", dace.int64),
-            )
-            strides = (
-                dace.symbol(unique_var_name() + "__strd", dace.int64),
-                dace.symbol(unique_var_name() + "__strd", dace.int64),
```
philip-paul-mueller, line 522: In the `add_storage` function we also create such variables, but we name them `shape` and `stride`.
We should use the same words in both functions. However, since they were part of #1379 I will change them there.

PR #1379, opened 2023-11-30
feat[next]: Separates ITIR -> SDFG translation from running

philip-paul-mueller:
Before it was only possible to translate ITIR to SDFG and execute it and it was not possible to extract the SDFG.
This commits splits this task into two parts and thus allows to perform the ITIR to SDFG translation without executing it.

Should be merged after #1374 .


src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -184,11 +184,9 @@ def get_cache_id(
     return m.hexdigest()
 
 
-def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs) -> None:
+def build_sdfg_from_itir(program: itir.FencilDefinition, *args, **kwargs) -> dace.SDFG:
```
edopao, line 187: What do you think about changing the signature of this function and explicitly passing the parameters that were already extracted by the caller?
`auto_optimize: bool`
`device: dace.DeviceType` (we can skip run_on_gpu)
`column_axis: Optional[Dimension]`
`offset_provider: dict[str, Any]`

philip-paul-mueller: @edopao Yes this sounds like a good idea. Could you review the changes, please.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -259,6 +305,7 @@ def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs) -> None:
         dace.config.Config.set("compiler", "allow_view_arguments", value=True)
         dace.config.Config.set("frontend", "check_args", value=True)
         sdfg_program(**expected_args)
+    return None
```
edopao, line 308: `return None` not needed

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -184,43 +184,89 @@ def get_cache_id(
...
+            If `for_gpu` is `None`, the default, the function will consult `run_on_gpu`.
+        Currently only the `FORCE_INLINE` liftmode is supported and an error is generated.
+    """
+    if for_gpu is None and "run_on_gpu" in kwargs:
+        for_gpu = kwargs.pop("run_on_gpu")
+    elif "run_on_gpu" in kwargs and for_gpu != kwargs["run_on_gpu"]:
+        raise ValueError(
+            "Passed conflicting arguments, for `for_gpu` you passed"
+            f"`{for_gpu}` but `run_on_gpu` was `{kwargs['run_on_gpu']}`."
+        )
```
edopao, line 220: If the problem is that we don't like "run_" in the name of the parameter, we should rename `run_on_gpu` to `for_gpu` in the original parameter list. It's a change local to this file.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -205,6 +208,16 @@ def visit_FencilDefinition(self, node: itir.FencilDefinition):
...
                 last_state.add_edge(nsdfg_node, inner_name, access_node, None, memlet)
 
+        # Create the call signature for the SDFG.
+        #  All arguments required by the SDFG, regardless if explicit and implicit, are added
+        #  as positional arguments. In the front are all arguments to the Fencil, in that
+        #  order, they are followed by the arguments created by the translation process,
+        arg_list = [str(a) for a in node.params]
+        sig_list = program_sdfg.signature_arglist(with_types=False)
+        implicit_args = set(sig_list) - set(arg_list)
+        call_params = arg_list + [ia for ia in sig_list if ia in implicit_args]
```
edopao, line 218: I don't know how this works, so I am just wondering in which order we should present the fencil arguments (`arg_list`). Should we maybe make sure that the arguments are in the same order as they appear in `sig_list`? 
I mean:
`call_params = [fa for fa in sig_list if fa in arg_list] + [ia for ia in sig_list if ia in implicit_args]`

From the API it is not clear if `sig_list` will follow the same order as `node.params`.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -205,6 +208,16 @@ def visit_FencilDefinition(self, node: itir.FencilDefinition):
...
+        program_sdfg.arg_names.extend(call_params)
```
edopao, line 219: Why `extend` and not just assign?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
philip-paul-mueller: This part is extremely strange. `sig_list` is the signature of the actuall `C++` function and `arg_names` is the (ordered) list of arguments how you can call the SDFG.
When you call `CompiledSDFG._call__()` the function expects that (positional) arguments are passed in the order  specified by `arg_names` before it will then perform the call to the `C++` function it will rearanging the passed arguments.
However, changing the order of `arg_names` directly to the one given by `sig_list` has no effect because, `CompiledSDFG.__call__()` will pack everything in a `dict` before it does the processing.

So I think that the best choice is to preserving the order given by the fencil, since it that what a user would expect. However, the additional arguments, that are produced by the process could be in any order.

As a side note, the order in `sig_list` is first all array arguments then all scalar, inside these two groups they are ordered alphabetical.

philip-paul-mueller: This part is extremely strange. `sig_list` is the signature of the actuall `C++` function and `arg_names` is the (ordered) list of arguments how you can call the SDFG.
When you call `CompiledSDFG._call__()` the function expects that (positional) arguments are passed in the order  specified by `arg_names` before it will then perform the call to the `C++` function it will rearanging the passed arguments.
However, changing the order of `arg_names` directly to the one given by `sig_list` has no effect because, `CompiledSDFG.__call__()` will pack everything in a `dict` before it does the processing.

So I think that the best choice is to preserving the order given by the fencil, since it that what a user would expect. However, the additional arguments, that are produced by the process could be in any order.

As a side note, the order in `sig_list` is first all array arguments then all scalar, inside these two groups they are ordered alphabetical.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
philip-paul-mueller: It is more general, since a preprocesser could already have added some arguments. However, we could also do an assignment, I think I will change it.

philip-paul-mueller:
No I did something stupid since these changes descend from #1374. Further one of your comments is basically everything what #1374 is all about.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -201,36 +201,26 @@ def build_sdfg_from_itir(
         *args:		        Arguments for which the fencil should be called.
         offset_provider:	The set of offset providers that should be used.
         auto_optimize:	    Apply DaCe's `auto_optimize` heuristic.
-        for_gpu:		    Performs the translation for GPU, defaults to `None`.
+        on_gpu:		        Performs the translation for GPU, defaults to `False`.
         column_axis:		The column axis to be used, defaults to `None`.
         lift_mode:		    Which lift mode should be used, defaults `FORCE_INLINE`.
         **kwargs:           Except `run_on_gpu` all are ignored.
```
edopao, line 207: This doc line should be updated. Can we remove `kwargs` from this function?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
edopao: I see. Thank you for looking into this and for the explanation.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
philip-paul-mueller: The docstring definitely has to be updated.

It is possible to remove `**kwargs` from function, but this has some implications, especially at where we call the function, since we have to explicitly filter out arguments the `build_sdfg_from_itir()` (see 014e1bae).
From an aesthetic pov I would revert that commit, but I am fine if we keep the change, since it is more explicit.

edopao:
cscs-ci run

philip-paul-mueller:
cscs-ci run

edopao:
cscs-ci run

edopao:
cscs-ci run

PR #1384, opened 2023-12-04
feat[next][dace]: Add more debug info to DaCe

kotsaloscv:
## Description

Add more debug info to DaCe (pass SourceLocation from past/foast to itir, and from itir to the SDFG)


petiaccja (changes requested):
The PR is pretty straightforward (that's a good thing), I left a few comments, but other than that I think there isn't much to change.

However, one thing I'm really missing is tests or any way to ensure that location info is properly transmitted. For example, if someone writes a new iterator IR pass and they don't pass the location, we won't know until somebody tries to look at the generated SDFGs while doing something else. That may be weeks or months down the line and finding the culprit causing the loss of location info will be difficult and annoying.

If you make the location field mandatory in iterator IR, that would do the trick. You could also enforce that location is not None in the SDFG generation, though that's less clear. You could also add a test that verifies if the SDFG has debug info.

kotsaloscv:
> The PR is pretty straightforward (that's a good thing), I left a few comments, but other than that I think there isn't much to change.
> 
> However, one thing I'm really missing is tests or any way to ensure that location info is properly transmitted. For example, if someone writes a new iterator IR pass and they don't pass the location, we won't know until somebody tries to look at the generated SDFGs while doing something else. That may be weeks or months down the line and finding the culprit causing the loss of location info will be difficult and annoying.
> 
> If you make the location field mandatory in iterator IR, that would do the trick. You could also enforce that location is not None in the SDFG generation, though that's less clear. You could also add a test that verifies if the SDFG has debug info.

FYI, I have integrated Peter's comments in my PR.
I also experimented on making the SourceLocation a mandatory field in the itir node, but it turns out that we will have to modify the code in countless places. There are so many tests that experiment with itir nodes for which we will have to pass a valid location. I do not think that it worths the investment. As is right now, I find it less intrusive and it is up to us to decide what we want to pass from past/foast to itir.
Regarding the testing of the genereted sdfgs to have a valid bebuginfo field, I have added this check in the code:
https://github.com/kotsaloscv/gt4py/blob/more_debinfo/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py#L222
At least, the developer will always be warned about this field and add it appropriately.

cc: @havogt @tehrengruber

src/gt4py/next/program_processors/runners/dace_iterator/utility.py
```diff
@@ -12,15 +12,33 @@
...
+        di = dace.dtypes.DebugInfo(
+            start_line=location.line,
+            start_column=location.column if location.column else 0,
+            end_line=location.end_line if location.end_line else -1,
+            end_column=location.end_column if location.end_column else 0,
+            filename=location.filename,
+        )
+    else:
+        di = debuginfo
+    return di
```
petiaccja, line 39: Simplify control flow:
```suggestion
    if location:
        return dace.dtypes.DebugInfo(
            start_line=location.line,
            start_column=location.column if location.column else 0,
            end_line=location.end_line if location.end_line else -1,
            end_column=location.end_column if location.end_column else 0,
            filename=location.filename,
        )
    return debuginfo
```

src/gt4py/eve/traits.py
```diff
@@ -172,3 +172,11 @@ def visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
...
 
         return result
+
+
+class PreserveLocationWithSymbolTableTrait(VisitorWithSymbolTableTrait):
+    def visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
+        result = super().visit(node, **kwargs)
+        if hasattr(node, "location") and hasattr(result, "location"):
+            result.location = node.location
+        return result
```
tehrengruber, line 182:
```suggestion
```
Beside the code duplication I don't think this is needed. Just inheriting from `PreserveLocation` should be fine right?

src/gt4py/eve/visitors.py
```diff
@@ -196,3 +196,11 @@ def generic_visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
             )
 
         return copy.deepcopy(node, memo=memo)
+
+
+class PreserveLocation(NodeVisitor):
```
tehrengruber, line 201:
```suggestion
class PreserveLocationVisitor(NodeVisitor):
```

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -61,15 +61,27 @@ class FieldOperatorLowering(NodeTranslator):
...
+    def apply(cls, node: foast.LocatedNode, preserve_location: bool = True) -> itir.Expr:
+        return cls(preserve_location=preserve_location).visit(node)
+
+    def visit(self, node: concepts.RootNode, **kwargs: extended_typing.Any) -> extended_typing.Any:
+        result = super().visit(node, **kwargs)
+        if (
+            hasattr(node, "location")
+            and hasattr(result, "location")
+            and not isinstance(node, foast.Name)
+            and self.preserve_location
```
tehrengruber, line 81: Why is the location preservation disabled for `foast.Name` nodes?
```suggestion
            and not isinstance(node, foast.Name)
            and self.preserve_location
```
If this wouldn't be needed the `preserve_location` flag could easily be added to `PreserveLocation` class making this custom handling unneeded.

src/gt4py/next/ffront/past_to_itir.py
```diff
@@ -247,7 +255,9 @@ def _construct_itir_domain_arg(
         else:
             raise AssertionError()
 
-        return itir.FunCall(fun=itir.SymRef(id=domain_builtin), args=domain_args)
+        return itir.FunCall(
+            fun=itir.SymRef(id=domain_builtin), args=domain_args, location=out_field.location
```
tehrengruber, line 259: Such that the location of the `node_domain` is used when available.
```suggestion
            fun=itir.SymRef(id=domain_builtin), args=domain_args, location=(node_domain or out_field).location
```

src/gt4py/next/ffront/past_to_itir.py
```diff
@@ -263,7 +273,10 @@ def _construct_itir_initialized_domain_arg(
                 f"expected '{dim}', got '{keys_dims_types}'."
             )
 
-        return [self.visit(bound) for bound in node_domain.values_[dim_i].elts]
+        itir_node = [self.visit(bound) for bound in node_domain.values_[dim_i].elts]
```
tehrengruber, line 276: Why is not taken care of by the `PreserveLocationWithSymbolTableTrait`? Aren't you just manually copying the location of the bound you are also visiting?

tehrengruber (approved):
Changes as discussed in person.

Issue #1292, opened 2023-07-21
Finalize error message formatting


petiaccja:
Error messages in gt4py are mainly formatted as a full sentence:
```
module.Error: The IR is at object `name` is invalid because of x and y.
```
Python, however, formats error messages without punctuation:
```
module.Error: the IR at object 'name' is invalid because of x and y
```
We should decide which style, if either of these, we follow, and adjust error messages across the code accordingly.

havogt:
Done in #1386

PR #1386, opened 2023-12-05
style[next]: standardize error messages.

DropD:
## Description

- add style guide to the coding guidelines
- fix existing error messages in next
- deal with ensuing qa errorrs / test updates
- unshadow one test and fix the code it wasn't testing

## Scope

The scope of this PR is purely the styling of existing error messages in `gt4py.next`.

## Out of Scope

Refactoring exception types, automatically composing recurring types of error messages etc is explicitly deferred to future PRs.

## Requirements


src/gt4py/_core/definitions.py
```diff
@@ -491,22 +499,28 @@ def __rtruediv__(self, other: Any) -> NDArrayObject:
     def __pow__(self, other: NDArrayObject | Scalar) -> NDArrayObject:
         ...
 
-    def __eq__(self, other: NDArrayObject | Scalar) -> NDArrayObject:  # type: ignore[override] # mypy want to return `bool`
+    # mypy wants to return `bool`
```
havogt, line 502: why this change? the comment explains the type ignore, therefore I prefer the old location.

CODING_GUIDELINES.md
```diff
@@ -51,6 +51,44 @@ We deviate from the [Google Python Style Guide][google-style-guide] only in the
   - Client code (like tests, doctests and examples) should use the above style for public FieldView API
   - Library code should always import the defining module and use qualified names.
 
+### Error messages
+
+Error messages should be written as sentences, starting with a capital letter and ending with a period (avoid exclamation marks). Try to be informative without being verbose. Code object like 'ClassNames' and 'function_names' etc should be single-quoted and so should string values used to interpolate the message.
```
havogt, line 56: Is this common practice to use full sentence? @petiaccja pointed out in #1292 that Python does formatting different. What does numpy and other big libraries do?  Also not explicitly stated, the google style guide implicitly (by examples) suggests what you propose.

src/gt4py/_core/definitions.py
DropD: The automatic formatting broke it. If pre-commit black is fine with the old formatting then I will have to investigate why my ruff-lsp differs from it.

CODING_GUIDELINES.md
DropD: I tried to find the reasoning behind the Python builtins convention but there doesn't even seem to be a conscious decision on record (at least not easily findable).

DropD:
> I only reviewed the changes in the coding conventions document and they look good to me (just have some style suggestions typo fixes).
> 
> As an additional comment (probably outside of the scope of this PR), I think we should look into refactor a bit our code to add more custom exceptions in the places where it makes sense and also to make sure we catch internal exceptions and wrapped them into informative DSLErrors in all places visible to users.

Out of scope for this PR but not for the project. Expect another PR along those lines.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -1108,7 +1108,7 @@ def unpack(
 def test_tuple_unpacking_too_many_values(cartesian_case):
```
DropD, line 1108: This test was previously shadowed by the succeeding one (now renamed), and turned out to fail. The changes to `foast_passes.type_deduction` are to make this one pass.

havogt (approved):
Thanks, looks good.

PR #1388, opened 2023-12-08
fix[next]:Python>=3.11

kotsaloscv:
## Description
Fix for Python>=3.11

### DIFF
--- a/src/gt4py/eve/extended_typing.py
+++ b/src/gt4py/eve/extended_typing.py
@@ -526,6 +526,14 @@ else:
 
 def has_type_parameters(cls: Type) -> bool:
     """Return ``True`` if obj is a generic class with type parameters."""
+    if len(getattr(cls, "__parameters__", [])) == 0:
+        cls.__parameters__ = []
+        annotations = getattr(cls, "__annotations__", {})
+        type_ = annotations.get("type")
+        if annotations and type_:
+            for t in get_args(type_):
+                if isinstance(t, _typing.TypeVar):
+                    cls.__parameters__.append(t)
     return issubclass(cls, Generic) and len(getattr(cls, "__parameters__", [])) > 0  # type: ignore[arg-type]  # Generic not considered as a class
 
 
--- a/.github/workflows/daily-ci.yml
+++ b/.github/workflows/daily-ci.yml
@@ -14,7 +14,7 @@ jobs:
   daily-ci:
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
         tox-module-factor: ["cartesian", "eve", "next", "storage"]
         os: ["ubuntu-latest"]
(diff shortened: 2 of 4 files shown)


src/gt4py/eve/extended_typing.py
```diff
@@ -526,6 +526,14 @@ def is_Any(obj: Any) -> bool:
...
 def has_type_parameters(cls: Type) -> bool:
     """Return ``True`` if obj is a generic class with type parameters."""
+    if len(getattr(cls, "__parameters__", [])) == 0:
+        cls.__parameters__ = []
+        annotations = getattr(cls, "__annotations__", {})
+        type_ = annotations.get("type")
+        if annotations and type_:
+            for t in get_args(type_):
+                if isinstance(t, _typing.TypeVar):
+                    cls.__parameters__.append(t)
```
egparedes, line 536: I don't really understand these lines, but they are adding a new `__parameters__` member to any class passed as parameter, which is effectively monkey-patching the class for no reason.

egparedes (changes requested):
Thanks for opening this PR! I think we still need more work to fully understand the cause of the error and find a cleaner the solution, but it's a good first try.

src/gt4py/eve/extended_typing.py
kotsaloscv: Hi @egparedes ,
FYI, I checked the python 3.10 and the `__parameters__` member is missing as well. So it seems to be a recurring issue not related to 3.11 only. It does not crash because it never enters this part of the code. I tried to understand where they diverge (3.10 vs 3.11) but I could not find it.
Essentially what the above code does is to check if this field exists and if not to properly construct it. Another place that would make sense to put it, would be this part of the code: https://github.com/kotsaloscv/gt4py/blob/py311_fix/src/gt4py/eve/datamodels/core.py#L1071
What do you think?

kotsaloscv:
FYI,
cc @egparedes 
We have used this branch with Will Sawyer to install icon4py and we got the following error:
```
_external_src/gt4py/src/gt4py/eve/type_validation.py:362: in _is_instance_of
    raise TypeError(
E   TypeError: 'Program.params' must be <class 'gt4py.next.ffront.program_ast.Symbol__datatypet__datatypet'> (got 'Symbol__datatypet(location=SourceLocation(filename='/Users/kotsaloc/repos/will/icon4py/model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/stencils/apply_nabla2_and_nabla4_global_to_vn.py', line=41, column=5, end_line=41, end_column=41), id=SymbolName('area_edge'), namespace=<Namespace.LOCAL: 'local'>, type=FieldType(dims=[Dimension(value='Edge', kind=<DimensionKind.HORIZONTAL: 'horizontal'>)], dtype=ScalarType(kind=<ScalarKind.FLOAT64: 1064>, shape=None)))' which is a <class 'gt4py.next.ffront.program_ast.Symbol__datatypet'>).
```
While the tests here are passing (even with this hack for Python 3.11), icon4py fails.
I am not sure if this is related to my fix/work-around or another issue with Python 3.11.

PR #1398, opened 2023-12-18
feat[dace]: Computing SDFG call arguments

philip-paul-mueller:
This commit adds a function that allows to generate the arguments needed to call an SDFG, before this was part of `run_dace_iterator()`. This made it very complex to run an SDFG outside this function.

One should consider this as an amend to [PR #1379](https://github.com/GridTools/gt4py/pull/1379).

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
@@ -13,7 +13,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import hashlib
 import warnings
-from typing import Any, Mapping, Optional, Sequence
+from typing import Any, Mapping, Optional, Sequence, Union
 
 import dace
 import numpy as np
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -94,8 +94,12 @@ def preprocess_program(
     return fencil_definition
 
 
-def get_args(params: Sequence[itir.Sym], args: Sequence[Any]) -> dict[str, Any]:
-    return {name.id: convert_arg(arg) for name, arg in zip(params, args)}
+def get_args(params: Union[Sequence[itir.Sym], dace.SDFG], args: Sequence[Any]) -> dict[str, Any]:
```
edopao, line 97: `get_args` and `get_offset_args` are used only in one place. Do we need to keep `Sequence[itir.Sym]` as input type?

philip-paul-mueller: I was also thinking that, but I kept it in the end for backwards compatibility, but you are right, we should remove it and only accept SDFG.

edopao (approved):

philip-paul-mueller:
cscs-ci run

PR #1399, opened 2023-12-18
fix[next][dace]: Fix memlet for array slicing

edopao:
The memlet used for array slicing was configured in the wrong direction, which caused an error in memlet propagation when targeting a GPU node.

This problem was analyzed in DaCe issue [#1481](https://github.com/spcl/dace/issues/1481), and resolved by @philip-paul-mueller and @edopao. However, DaCe is still not able of generating CUDA code for such representation.

As a workaround, this PR proposes an implementation of array slicing based on a mapped tasklet.


philip-paul-mueller:
While the fix proposed by @edopao helps, I now get an error regarding some missed symbols in an nested SDFG.

edopao:
@philip-paul-mueller I have pushed a different implementation, based on a mapped tasklet. This implementation works on GPU (tested on `TestMoVelocityAdvectionStencil20`).

philip-paul-mueller (approved):

PR #1400, opened 2023-12-19
fix[dace]: Fixed SDFG args

philip-paul-mueller:
Modified how the SDFG arguments are computed.

It was noticed that some transformations, especially the `SDFG.apply_gpu_transformation()`, to the SDFG, added new arguments to the SDFG.
But, since a lot of functions build on the `SDFG.arg_names` member and this member was populated before the transformation, an error occurred.
Thus it was changed such that `SDFG.arg_names` was only populated with the arguments also known to the Fencil.


philip-paul-mueller:
cscs-ci run

edopao (approved):

philip-paul-mueller:
cscs-ci run

PR #1405, opened 2023-12-23
bug[next]: Fix size check in CollapseTuple pass

tehrengruber:
Just a small bugfix for the CollapseTuple pass such that Tuples where the tail is a TypeVar are supported (we don't know the length for them).


havogt (approved):
lgtm

PR #1406, opened 2024-01-04
feature[next]: Make gtfn cache path configurable

tehrengruber:
Description to follow


DropD (commented):
The benefit of this design is obvious: a custom path can now be set.

On the other hand, this does not tick any of the boxes in https://hackmd.io/2l_ut1O-Q3ilK1IZboSryA and that project exists to avoid heaping more ad-hoc customizability on top of this system.

DropD:
Also, changing the interface from explicitly stating the desire for a cache that does or doesn't stick around after the run (and get in the way of debugging) to passing a path and implicitly determining the lifetime by where that path is on the FS could be seen as a usability degradation.

tehrengruber:
Superseeded by #1438

PR #1411, opened 2024-01-09
bug[next]: Bound args kwargs edit

### LINKED ISSUE #1402: ProgramWithBoundArgs: pos_or_kw_args need to go to kwargs sometimes
in decorator.py:455, we need to do something like

```
        full_args = [*args]
        full_kwargs = {**kwargs}
        for index, param in enumerate(self.past_node.params):
            if param.id in self.bound_args.keys():
                if index < len(full_args):
                    full_args.insert(index, self.bound_args[param.id])
                else:
                    full_kwargs[param.id] = self.bound_args[param.id]

        return super()._process_args(tuple(full_args), full_kwargs)
```
in case we have a bound_arg which need to be provided as kwarg, because all previous pos_or_kw_args are passed as kwargs.

Test case would be something like
```python
@program
def foo(a, b, c):
    ...

bound = foo.with_bound_args(b = True)

bound(a=1,c=2)
```
currently we try to call foo as `foo(True, a=1,c=2)` which messes up the order.

nfarabullini:
Fixes #1402

### DIFF
--- a/src/gt4py/next/ffront/decorator.py
+++ b/src/gt4py/next/ffront/decorator.py
@@ -453,11 +453,24 @@ class ProgramWithBoundArgs(Program):
             ) from err
 
         full_args = [*args]
+        full_kwargs = list(kwargs.items())
         for index, param in enumerate(self.past_node.params):
             if param.id in self.bound_args.keys():
-                full_args.insert(index, self.bound_args[param.id])
+                if index < len(full_args):
+                    full_args.insert(index, self.bound_args[param.id])
+                else:
+                    pos = list(self.past_node.type.definition.pos_or_kw_args).index(str(param.id))
+                    full_kwargs.insert(pos, (str(param.id), self.bound_args[param.id]))
+                    #full_kwargs[str(param.id)] = self.bound_args[param.id]
 
-        return super()._process_args(tuple(full_args), kwargs)
+        return super()._process_args(tuple(full_args), dict(full_kwargs))
+
+        # full_args = [*args]
+        # for index, param in enumerate(self.past_node.params):
+        #     if param.id in self.bound_args.keys():
+        #         full_args.insert(index, self.bound_args[param.id])
+        #
+        # return super()._process_args(tuple(full_args), kwargs)
 
     @functools.cached_property
     def itir(self):
--- a/src/gt4py/next/ffront/decorator.py
+++ b/src/gt4py/next/ffront/decorator.py
@@ -461,16 +461,8 @@ class ProgramWithBoundArgs(Program):
                 else:
                     pos = list(self.past_node.type.definition.pos_or_kw_args).index(str(param.id))
                     full_kwargs.insert(pos, (str(param.id), self.bound_args[param.id]))
-                    #full_kwargs[str(param.id)] = self.bound_args[param.id]
 
         return super()._process_args(tuple(full_args), dict(full_kwargs))
-
-        # full_args = [*args]
-        # for index, param in enumerate(self.past_node.params):
-        #     if param.id in self.bound_args.keys():
-        #         full_args.insert(index, self.bound_args[param.id])
-        #
-        # return super()._process_args(tuple(full_args), kwargs)
 
     @functools.cached_property
     def itir(self):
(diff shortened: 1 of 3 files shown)


src/gt4py/next/ffront/decorator.py
```diff
@@ -453,11 +453,16 @@ def _process_args(self, args: tuple, kwargs: dict):
...
         full_args = [*args]
+        full_kwargs = list(kwargs.items())
         for index, param in enumerate(self.past_node.params):
             if param.id in self.bound_args.keys():
-                full_args.insert(index, self.bound_args[param.id])
+                if index < len(full_args):
+                    full_args.insert(index, self.bound_args[param.id])
+                else:
+                    pos = list(self.past_node.type.definition.pos_or_kw_args).index(str(param.id))
+                    full_kwargs.insert(pos, (str(param.id), self.bound_args[param.id]))
```
havogt, line 463: why do you operate on a list and not directly on the dict?

src/gt4py/next/ffront/decorator.py
```diff
@@ -472,8 +477,9 @@ def itir(self):
                 new_clos.inputs.pop(index)
             new_args = [ref(inp.id) for inp in new_clos.inputs]
             params = [sym(inp.id) for inp in new_clos.inputs]
-            for value in self.bound_args.values():
-                new_args.append(promote_to_const_iterator(literal_from_value(value)))
+            for key, value in self.bound_args.items():
+                pos = list(self.past_node.type.definition.pos_or_kw_args).index(key)
+                new_args.insert(pos, promote_to_const_iterator(literal_from_value(value)))
```
havogt, line 482: looking at this I guess the answer to the comment above is: order matters?

Can't we reconstruct the order here?
When we go through the original closure.inputs, can't we just directly build the new_args with the correct order?

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -918,6 +918,26 @@ def program_bound_args(a: cases.IField, scalar: int32, condition: bool, out: cas
     cases.verify(cartesian_case, prog_bounds, a, out, inout=out, ref=ref)
 
 
+def test_with_bound_args_order_args(cartesian_case):
```
havogt, line 921: let's move all bound_arg tests to a new module test_bound_args

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -918,6 +918,26 @@ def program_bound_args(a: cases.IField, scalar: int32, condition: bool, out: cas
...
+    @gtx.field_operator
+    def fieldop_args(a: cases.IField, condition: bool, scalar: int32) -> cases.IField:
+        if not condition:
+            scalar = 0
+        return a + scalar
```
havogt, line 926: I would simplify the test case to the minimum needed to test the problem (the condition is just distraction from the core problem).

havogt (changes requested):
If I understand correctly, the bound_args dictionary needs to be ordered, which seems to be an overspecification, because the order is already defined by the params of the original program. Can you double-check if that's the case and if the code could be simplified.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
nfarabullini: I would disagree here because the boolean parameter is good to asses the `type_inference` passes correctly

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_bound_args.py
```diff
@@ -0,0 +1,64 @@
...
+    reduction_setup,
+)
+
+
+def test_with_bound_args(cartesian_case):
+    @gtx.field_operator
+    def fieldop_bound_args(a: cases.IField, scalar: int32, condition: bool) -> cases.IField:
+        if not condition:
+            scalar = 0
+        return a + a + scalar
```
havogt, line 34: > I would disagree here because the boolean parameter is good to asses the `type_inference` passes correctly

A test should be as minimal as possible to illustrate what it's testing. Using different types in the input is one way. But in the body (in the runtime part) the order of the arguments is not tested in the simplest way. E.g. what does `a+a` do to this test?

nfarabullini: this I agree with

nfarabullini:
cscs-ci run

havogt (approved):

PR #1420, opened 2024-01-23
fix[next]: Default to Release in CMake

havogt:


src/gt4py/next/otf/compilation/build_systems/cmake.py
```diff
@@ -44,7 +44,7 @@ class CMakeFactory(
     """Create a CMakeProject from a ``CompilableSource`` stage object with given CMake settings."""
 
     cmake_generator_name: str = "Ninja"
-    cmake_build_type: BuildType = BuildType.DEBUG
```
havogt, line 47: @DropD please check if the default make sense in 3 different places and consider this example in the configuration project

DropD: Well, sort of. I would argue that `CMakeProject` does not need to have a default value, since we never instanciate it directly. It makes sense though, that the factories can have separate default values. 
Consider that using pure cmake leads to slightly simpler debugging, and in debugging the compilation times might not be as crucial. 

There is, therefore an argument for leaving the default for `CMakeFactory` at `DEBUG`, while changing it to `RELEASE` for `CompiledbFactory`... However, in practice this might be more consfusing than helpful...

havogt:
Fixed in #1438

PR #1422, opened 2024-01-24
fix[next][dace]: Use constant shape for neighbor tables in local dimension

edopao:
## Description

Main purpose of this PR is to avoid the definition of shape symbols for array dimensions known at compile time. The local size of neighbor connectivity tables falls into this category. For each element in the origin dimension, the number of elements in the target dimension is defined by the attribute `max_neighbors` in the offset provider.

Some side changes included in the PR:

- In the hash function for the build cache, include `max_neighbors` and other attributes of NeighborTable offset providers.
- In the hash function for the build cache, include the build type (debug, release, etc.).
- Replace `itir_embedded.NeighborTableOffsetProvider` with generic type `common.NeighborTable` in type annotations.
- Optimize `list_get` for the case where the index is a `SymbolExpr`, simply use a memlet.
- Optimize neighbor reduction for the case with `has_skip_values == False`, do not generate the `can_deref` check.
- Rename parameter `conn_names` to `connectivity_names`.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
@@ -108,12 +108,12 @@ def _ensure_is_on_device(
 
 
 def get_connectivity_args(
-    neighbor_tables: Sequence[tuple[str, itir_embedded.NeighborTableOffsetProvider]],
+    neighbor_tables: Mapping[str, itir_embedded.NeighborTableOffsetProvider],
     device: dace.dtypes.DeviceType,
 ) -> dict[str, Any]:
     return {
         connectivity_identifier(offset): _ensure_is_on_device(table.table, device)
-        for offset, table in neighbor_tables
+        for offset, table in neighbor_tables.items()
     }
 
 
@@ -124,6 +124,7 @@ def get_shape_args(
         str(sym): size
         for name, value in args.items()
         for sym, size in zip(arrays[name].shape, value.shape)
+        if isinstance(sym, dace.symbol)
     }
 
 
@@ -271,7 +272,7 @@ def build_sdfg_from_itir(
     # visit ITIR and generate SDFG
     program = preprocess_program(program, offset_provider, lift_mode)
     sdfg_genenerator = ItirToSDFG(arg_types, offset_provider, column_axis)
-    sdfg = sdfg_genenerator.visit(program)
+    sdfg: dace.SDFG = sdfg_genenerator.visit(program)
     if sdfg is None:
         raise RuntimeError(f"Visit failed for program {program.id}.")
 
--- a/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
@@ -108,19 +108,12 @@ def _ensure_is_on_device(
 
 
 def get_connectivity_args(
-    sdfg_sig: Sequence[str],
     neighbor_tables: Mapping[str, common.NeighborTable],
     device: dace.dtypes.DeviceType,
 ) -> dict[str, Any]:
-    # keep only neighbor tables that are used by the SDFG to reduce the copy to device
-    sdfg_neighbor_tables = [
-        (offset, offset_provider)
-        for offset, offset_provider in neighbor_tables.items()
-        if connectivity_identifier(offset) in sdfg_sig
-    ]
     return {
         connectivity_identifier(offset): _ensure_is_on_device(offset_provider.table, device)
-        for offset, offset_provider in sdfg_neighbor_tables
+        for offset, offset_provider in neighbor_tables.items()
     }
 
 
@@ -227,7 +220,7 @@ def get_sdfg_args(sdfg: dace.SDFG, *args, **kwargs) -> dict[str, Any]:
     sdfg_sig = sdfg.signature_arglist(with_types=False)
     dace_args = get_args(sdfg, args)
     dace_field_args = {n: v for n, v in dace_args.items() if not np.isscalar(v)}
-    dace_conn_args = get_connectivity_args(sdfg_sig, neighbor_tables, device)
+    dace_conn_args = get_connectivity_args(neighbor_tables, device)
     dace_shapes = get_shape_args(sdfg.arrays, dace_field_args)
     dace_conn_shapes = get_shape_args(sdfg.arrays, dace_conn_args)
     dace_strides = get_stride_args(sdfg.arrays, dace_field_args)
@@ -320,15 +313,21 @@ def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs):
         sdfg = sdfg_program.sdfg
 
     else:
-        sdfg = build_sdfg_from_itir(
-            program,
-            *args,
-            offset_provider=offset_provider,
-            auto_optimize=auto_optimize,
-            on_gpu=on_gpu,
-            column_axis=column_axis,
-            lift_mode=lift_mode,
-        )
+        generate_sdfg = True
+        sdfg_filename = f"/repo/icon4py/{program.id}.sdfg"
+
+        if generate_sdfg:
+            sdfg = build_sdfg_from_itir(
+                program,
+                *args,
+                offset_provider=offset_provider,
+                auto_optimize=auto_optimize,
+                on_gpu=on_gpu,
+                column_axis=column_axis,
+                lift_mode=lift_mode,
+            )
+        else:
+            sdfg = dace.SDFG.from_file(sdfg_filename)
 
         sdfg.build_folder = compilation_cache._session_cache_dir_path / ".dacecache"
         with dace.config.temporary_config():
--- a/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
@@ -313,8 +313,9 @@ def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs):
         sdfg = sdfg_program.sdfg
 
     else:
+        # debug: test large icon4py stencils without regenerating the SDFG at each run
         generate_sdfg = True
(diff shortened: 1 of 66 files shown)


src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -109,23 +109,29 @@ def _ensure_is_on_device(
...
 def get_connectivity_args(
-    neighbor_tables: Sequence[tuple[str, itir_embedded.NeighborTableOffsetProvider]],
+    neighbor_tables: Mapping[str, common.NeighborTable],
     device: dace.dtypes.DeviceType,
 ) -> dict[str, Any]:
     return {
-        connectivity_identifier(offset): _ensure_is_on_device(table.table, device)
-        for offset, table in neighbor_tables
+        connectivity_identifier(offset): _ensure_is_on_device(offset_provider.table, device)
+        for offset, offset_provider in neighbor_tables.items()
```
philip-paul-mueller, line 117: Just for my curiosity, did you found out what the problem here was?

src/gt4py/next/program_processors/runners/dace_iterator/utility.py
```diff
@@ -181,6 +181,24 @@ def unique_var_name():
     return unique_name("_var")
 
 
+def map_field_dimensions_to_sdfg_symbols(
```
philip-paul-mueller, line 184: For some reason I do not like the name of this function, since it does not express what the function is actually doing.
My suggestion would be something  `_create_shape_stride()`.
- Since it is internal to the package we would prefix it with `_`.
- I do not like the word mapping in this context, since for me a mapping, as it is used in the context here (to my knowledge; I know there is `map`), establishes a connection between two "things". However, here if needed we create the symbols, so I think we should include [creat](https://linux.die.net/man/2/creat) or make in the function name.

If you want to keep the name, then you can ignore the stuff above.


However, the function does something highly non trivial, so we should add a meaningful docstring or at least some comments.
I do not believe in the ["the code is the ultimate documentation" mantra](https://web.mit.edu/~simsong/www/ugh.pdf).

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -310,22 +322,31 @@ def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs):
...
+                *args,
+                offset_provider=offset_provider,
+                auto_optimize=auto_optimize,
+                on_gpu=on_gpu,
+                column_axis=column_axis,
+                lift_mode=lift_mode,
+            )
+            sdfg.save(sdfg_filename)
+        else:
+            sdfg = dace.SDFG.from_file(sdfg_filename)
```
philip-paul-mueller, line 349: It looks like you have now two caches, one that is coming from the outside, `build_cache`  and one you create inside the `_dacegraphs` folder?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -222,7 +239,9 @@ def visit_StencilClosure(
 
         input_names = [str(inp.id) for inp in node.inputs]
         neighbor_tables = filter_neighbor_tables(self.offset_provider)
-        connectivity_names = [connectivity_identifier(offset) for offset, _ in neighbor_tables]
+        connectivity_names = [
+            connectivity_identifier(offset) for offset, _ in neighbor_tables.items()
```
philip-paul-mueller, line 243: Use `.keys()`

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -222,7 +239,9 @@ def visit_StencilClosure(
...
+        ]
 
         output_nodes = self.get_output_nodes(node, closure_sdfg, closure_state)
         output_names = [k for k, _ in output_nodes.items()]
```
philip-paul-mueller, line 247: Use `.keys()`

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -570,7 +589,9 @@ def _visit_parallel_stencil_closure(
     ) -> tuple[dace.SDFG, dict[str, str | dace.subsets.Subset], list[str]]:
         neighbor_tables = filter_neighbor_tables(self.offset_provider)
         input_names = [str(inp.id) for inp in node.inputs]
-        conn_names = [connectivity_identifier(offset) for offset, _ in neighbor_tables]
+        connectivity_names = [
+            connectivity_identifier(offset) for offset, _ in neighbor_tables.items()
```
philip-paul-mueller, line 593: Use `.keys()`

philip-paul-mueller (changes requested):
The changes looks very good, however, there are some small changes and clarification needed.
The majority is that you should change expressions like `for k, _ in dict().item()` to `for k in dict().keys()`, which makes things clearer.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
edopao: Ok

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
edopao: Ok

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -372,20 +377,27 @@ def builtin_list_get(
...
-    expr = f"{internals[1]}[{internals[0]}]"
-    return transformer.add_expr_tasklet(
-        expr_args, expr, args[1].dtype, "list_get", dace_debuginfo=di
-    )
+    if isinstance(args[0], SymbolExpr):
+        index_value = args[0].value
+        result_name = unique_var_name()
+        transformer.context.body.add_scalar(result_name, args[1].dtype, transient=True)
+        result_node = transformer.context.state.add_access(result_name)
+        transformer.context.state.add_edge(
```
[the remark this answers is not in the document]
edopao: Changed.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -562,9 +574,11 @@ def visit_Lambda(
     ]:
         func_name = f"lambda_{abs(hash(node)):x}"
         neighbor_tables = (
-            filter_neighbor_tables(self.offset_provider) if use_neighbor_tables else []
+            filter_neighbor_tables(self.offset_provider) if use_neighbor_tables else {}
         )
-        connectivity_names = [connectivity_identifier(offset) for offset, _ in neighbor_tables]
+        connectivity_names = [
+            connectivity_identifier(offset) for offset, _ in neighbor_tables.items()
```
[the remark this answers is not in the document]
edopao: Ok

src/gt4py/next/program_processors/runners/dace_iterator/utility.py
edopao: I agree, function name changed to `_make_array_shape_and_strides`, doc string added and function moved to file scope.

src/gt4py/next/program_processors/runners/dace_iterator/utility.py
```diff
@@ -181,6 +181,24 @@ def unique_var_name():
...
+        neighbor_tables[dim.value].max_neighbors
+        if dim.kind == DimensionKind.LOCAL
+        else dace.symbol(unique_name(f"{name}_shape{i}"), dtype)
+        for i, dim in enumerate(sorted_dims)
+    ]
+    strides = [dace.symbol(unique_name(f"{name}_stride{i}"), dtype) for i, _ in enumerate(shape)]
+    return shape, strides
+
+
 def new_array_symbols(name: str, ndim: int) -> tuple[list[dace.symbol], list[dace.symbol]]:
```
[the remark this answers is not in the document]
edopao: I will keep it for now, because it is still used when we allocate arrays (for both fields and connectivities) in nested scope. The constant dimensions are propagated through the levels of nesting by the simplify pass.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -310,22 +322,31 @@ def run_dace_iterator(program: itir.FencilDefinition, *args, **kwargs):
...
-            on_gpu=on_gpu,
-            column_axis=column_axis,
-            lift_mode=lift_mode,
-        )
+        # debug: test large icon4py stencils without regenerating the SDFG at each run
+        generate_sdfg = True
+        target = "gpu" if on_gpu else "cpu"
+        sdfg_filename = f"_dacegraphs/{target}/{program.id}.sdfg"
+
+        if generate_sdfg:
```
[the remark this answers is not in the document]
philip-paul-mueller: Okay this makes sense (and would be in fact also useful for me).
However, there should be some extranall way of doing it.
An idea would be to add an argument `_force_dace_cache` to the function.

src/gt4py/next/program_processors/runners/dace_iterator/utility.py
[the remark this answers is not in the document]
philip-paul-mueller: This make sense.
However, I do not like this "array creation in the wild", there should be probably something like `add_storage()`.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
philip-paul-mueller: Somehow, my comment got missing so I repeat it here:
I very much like the idea, but I think that it should be possible to influence the value from the outside, for example by an ENV or an explicit argument.
Because this is plain weird, you should at least put a comment to indicate that it must be changed manually.

philip-paul-mueller (approved):

PR #1423, opened 2024-01-24
build: Update deployment action with trusted publisher

havogt:
This can only be tested once merged and a new release is created.

### DIFF
--- a/.github/workflows/deploy-release.yml
+++ b/.github/workflows/deploy-release.yml
@@ -2,9 +2,9 @@ name: Deploy Python Distribution
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
   release:
     types: [published]
   workflow_dispatch:
--- a/.github/workflows/deploy-release.yml
+++ b/.github/workflows/deploy-release.yml
@@ -6,8 +6,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  build-n-publish:
-    name: Build and publish Python distribution
+  build:
+    name: Build Python distribution
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@master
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-release.yml
```diff
@@ -21,14 +21,29 @@ jobs:
...
       run: |
         python -m build --sdist --wheel --outdir dist/
-    - name: Publish distribution to Test PyPI
-      if: ${{ github.event_name == 'release' }}
-      uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
-        repository_url: https://test.pypi.org/legacy/
+  publish-pypi:
+    name: Publish Python distribution to test.pypi.org
```
DropD, line 25:
```suggestion
    name: Publish Python distribution to pypi.org
```

It seems to me that this one is for the main PyPI.

.github/workflows/deploy-release.yml
```diff
@@ -21,14 +21,29 @@ jobs:
...
+      url: https://pypi.org/project/gt4py
+    permissions:
+      id-token: write
+    steps:
     - name: Publish distribution to PyPI
-      if: ${{ github.event_name == 'workflow_dispatch' }}
       uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        password: ${{ secrets.PYPI_API_TOKEN }}
+  publish-test-pypi:
```
DropD, line 37: What is the value of publishing to test as well? If there is one, should this be tried first, so that if it fails we don't even try to upload to main PyPI?

**edit**
This is clear after reading the documentation

.github/workflows/deploy-release.yml
```diff
@@ -21,14 +21,29 @@ jobs:
...
+    name: Publish Python distribution to test.pypi.org
+    runs-on: ubuntu-latest
+    needs: build
+    if: ${{ github.event_name == 'release' }}
```
DropD, line 41: This is confusing me: TestPyPI is uploaded on "release", while main PyPI is uploaded on "workflow_dispatch". The name "release" seems to suggest it should be the other way around.

**edit** 
The release documentation explains this. Maybe a comment here would help someone casually skimming through this without the release docs open at the same time.

.github/workflows/deploy-release.yml
```diff
@@ -21,14 +21,29 @@ jobs:
...
-      if: ${{ github.event_name == 'release' }}
-      uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
-        repository_url: https://test.pypi.org/legacy/
+  publish-pypi:
+    name: Publish Python distribution to test.pypi.org
+    runs-on: ubuntu-latest
+    needs: build
+    if: ${{ github.event_name == 'workflow_dispatch' }}
```
DropD, line 28:
```suggestion
    if: ${{ github.event_name == 'workflow_dispatch' }} # the action was triggered manually
```

.github/workflows/deploy-release.yml
DropD, line 41:
```suggestion
    if: ${{ github.event_name == 'release' }} # triggered by releasing on github, test first before manually triggering the deployment to PyPI (see release documentation)
```

DropD (changes requested):
The only mandatory change is the name of the full PyPI release step, the other ones will help whoever needs to edit this in the future though.

DropD (approved):

egparedes (approved):
ILGTM

.github/workflows/deploy-release.yml
```diff
@@ -21,14 +21,29 @@ jobs:
...
+    needs: build
+    if: ${{ github.event_name == 'release' }} # triggered by releasing on github, test first before manually triggering the deployment to PyPI (see release documentation)
+    environment:
+      name: testpypi
+      url: https://test.pypi.org/project/gt4py/
+    permissions:
+      id-token: write
+    steps:
+    - name: Publish distribution to Test PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
```
havogt, line 49: missing url?

egparedes: Based on the documentation here (https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#separate-workflow-for-publishing-to-testpypi), it looks like the `repository-url` key might be missing:
```
      with:
        repository-url: https://test.pypi.org/legacy/
```

havogt: yes, sorry, that comment was for me to add exactly that

.github/workflows/deploy-release.yml
```diff
@@ -21,14 +21,29 @@ jobs:
     - name: Build a wheel and a source tarball
       run: |
         python -m build --sdist --wheel --outdir dist/
```
havogt, line 23: store artifact

PR #1427, opened 2024-01-25
feat[next]: Slicing field to 0d to return field not scalar

nfarabullini:
Extracting a single value from a field should return an empty field, not a scalar

Given the following example:
```python
    @gtx.field_operator
    def testee_fo(a: cases.IKField) -> cases.IKField:
        return a

    @gtx.program
    def testee_prog(a: cases.IKField):
        testee_fo(a, out=a[1:2, 3:4])
```

The `program` return type should be:
```python
     gtx.Field[[], int]
```

Additionally, a small fix was made in `past_passes.type_deduction` such that `arg_types` is used instead of `nodes.args.type`


src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -203,15 +203,13 @@ def remap(
 
     __call__ = remap  # type: ignore[assignment]
 
-    def restrict(self, index: common.AnyIndexSpec) -> common.Field | core_defs.ScalarT:
+    def restrict(self, index: common.AnyIndexSpec) -> common.Field:
```
egparedes, line 206: The return type annotation should also be changed in all the related protocols like `Field`, `MutableField`, ...

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -203,15 +203,13 @@ def remap(
...
         new_domain, buffer_slice = self._slice(index)
-
         new_buffer = self.ndarray[buffer_slice]
-        if len(new_domain) == 0:
-            # TODO: assert core_defs.is_scalar_type(new_buffer), new_buffer
-            return new_buffer  # type: ignore[return-value] # I don't think we can express that we return `ScalarT` here
-        else:
-            return self.__class__.from_array(new_buffer, domain=new_domain)
+        if new_domain.ndim == 0:
+            assert self.array_ns.dtype(new_buffer) in core_defs.SCALAR_TYPES
```
egparedes, line 210: The assert is not really necessary, since we are restricting an existing field. I would remove it.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -203,15 +203,13 @@ def remap(
...
+            return self._scalar_to_field(new_buffer)  # type: ignore[return-value, arg-type]
```
egparedes, line 211: This is returning an array not a field. I think `NdArrayField.from_array()` could be used here, since it should work for both scalars and ndarrays.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -304,6 +302,12 @@ def _slice(
         assert common.is_relative_index_sequence(slice_)
         return new_domain, slice_
 
+    def _scalar_to_field(self, value: core_defs.Scalar) -> np.ndarray:
+        if self.array_ns == cp:
+            return cp.asarray(value)
+        else:
+            return np.asarray(value)
+
```
egparedes, line 310: Not needed, see comment above.
 ```suggestion
```

File: src/gt4py/next/ffront/past_passes/type_deduction.py [hunk 5]
```diff
@@ -217,12 +217,12 @@ def visit_Call(self, node: past.Call, **kwargs):
                         f"'{new_kwargs['out'].type}'."
                     )
             elif new_func.id in ["minimum", "maximum"]:
-                if new_args[0].type != new_args[1].type:
+                if arg_types[0] != arg_types[1]:
```
egparedes, line 220: Unrelated fix, so please don't forget to mention it in the PR/commit description.

tests/next_tests/integration_tests/multi_feature_tests/iterator_tests/test_anton_toy.py
```diff
@@ -68,8 +68,11 @@ def naive_lap(inp):
     for i in range(1, shape[0] + 1):
         for j in range(1, shape[1] + 1):
             for k in range(0, shape[2]):
-                out[i - 1, j - 1, k] = -4 * inp[i, j, k] + (
-                    inp[i + 1, j, k] + inp[i - 1, j, k] + inp[i, j + 1, k] + inp[i, j - 1, k]
+                out[i - 1, j - 1, k] = -4 * inp.asnumpy()[i, j, k] + (
```
egparedes, line 71: Couldn't we add an alias to the input numpy array outside of the loop nest and then use the alias inside?
```
inp_data = inp.asnumpy()
for ...
   for ...
      for ...
         out ... = inp_data[...]
```

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -320,6 +320,21 @@ def testee(inp: gtx.Field[[KDim], float]) -> gtx.Field[[KDim], float]:
...
+    @gtx.field_operator
+    def testee_fo(a: cases.IKField) -> cases.IKField:
+        return a
+
+    @gtx.program
+    def testee_prog(a: cases.IKField):
+        testee_fo(a, out=a[1:2, 3:4])
+
+    a = cases.allocate(cartesian_case, testee_prog, "a")()
+    ref = np.asarray(a.asnumpy()[1, 3])
```
egparedes, line 333: Just a question, can't we directly slice the field and then extract the data as a numpy ndarray or would it fail?
```suggestion
    ref = a[1, 3].asnumpy()
```

tests/next_tests/integration_tests/feature_tests/iterator_tests/test_horizontal_indirection.py
```diff
@@ -101,7 +101,7 @@ def test_direct_offset_for_indirection(program_processor):
 
     ref = np.zeros(shape)
     for i in range(shape[0]):
-        ref[i] = inp[i + cond[i]]
+        ref[i] = inp.ndarray[i + cond[i].item()]
```
egparedes, line 104: If getitem returns a field, it should be explicitly transformed first into a ndarray
```suggestion
        ref[i] = inp.ndarray[i + cond[i].ndarray.item()]
```

egparedes (changes requested):
It looks good overall but there are some changes that need to be done before merging.

tests/next_tests/integration_tests/feature_tests/iterator_tests/test_horizontal_indirection.py
nfarabullini:
```suggestion
        ref[i] = inp.ndarray[i + cond.ndarray[i]]
```

src/gt4py/next/embedded/nd_array_field.py
nfarabullini: `ConstantField` returns a scalar for `restrict` and `_tuple_assign_field` uses `MutableField` to return scalars. I would leave these as they are

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -203,15 +203,12 @@ def remap(
...
-
         new_buffer = self.ndarray[buffer_slice]
-        if len(new_domain) == 0:
-            # TODO: assert core_defs.is_scalar_type(new_buffer), new_buffer
-            return new_buffer  # type: ignore[return-value] # I don't think we can express that we return `ScalarT` here
-        else:
-            return self.__class__.from_array(new_buffer, domain=new_domain)
+        if new_domain.ndim == 0:
+            return self.from_array(new_buffer, domain=new_domain)
+        return self.__class__.from_array(new_buffer, domain=new_domain)
```
egparedes, line 211: I think the `if` is redundant, aren't the two branches doing the same thing?

src/gt4py/next/ffront/past_to_itir.py
```diff
@@ -382,12 +382,11 @@ def visit_BinOp(self, node: past.BinOp, **kwargs) -> itir.FunCall:
...
+            assert len(node.args) == 2
             return itir.FunCall(
                 fun=itir.SymRef(id=node.func.id),
                 args=[self.visit(node.args[0]), self.visit(node.args[1])],
             )
         else:
-            raise AssertionError(
-                "Only 'minimum' and 'maximum' builtins supported supported currently."
-            )
+            raise AssertionError("Only 'minimum', and 'maximum' builtins supported currently.")
```
egparedes, line 392: Completely optional suggestion since this is unrelated to this PR but since the code has been touched anyway maybe a `NotImplementedError` would make more sense here. Feel free to ignore.
```suggestion
            raise NotImplementedError("Only 'minimum', and 'maximum' builtins supported currently.")
```

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -468,10 +468,10 @@ def test_absolute_indexing_value_return():
     field = common._field(np.reshape(np.arange(100, dtype=np.int32), (10, 10)), domain=domain)
 
     named_index = ((IDim, 12), (JDim, 6))
-    value = field[named_index]
+    value = field[named_index].ndarray
 
-    assert isinstance(value, np.int32)
-    assert value == 21
+    assert isinstance(value, np.ndarray)
+    assert value == np.asarray([21])
```
egparedes, line 474: If the expected value is a 0d field, shouldn't it be expressed as below?
```suggestion
    assert value == np.asarray(21)
```
By automatic shape broadcasting most likely works in both ways, but it feels it would be more correct to use a 0d array as reference, or am I misinterpreting the test?

egparedes (commented):
It looks clean now. I only have a couple of issues and questions

src/gt4py/next/embedded/nd_array_field.py
nfarabullini: Damn you're right

src/gt4py/next/embedded/nd_array_field.py
havogt: They should be changed then. In the end if you slice any field, you want to get a field back.

egparedes (approved):
ILGTM.

src/gt4py/next/embedded/operators.py
```diff
@@ -188,7 +188,9 @@ def _tuple_at(
     @utils.tree_map
     def impl(field: common.Field | core_defs.Scalar) -> core_defs.Scalar:
         res = field[pos] if common.is_field(field) else field
-        res = res.item() if hasattr(res, "item") else res  # extract scalar value from array
+        res = (
+            res.ndarray.item() if hasattr(res, "ndarray") else res  # type: ignore[union-attr]
+        )  # extract scalar value from array
```
havogt, line 193: I think we should enable something like 
```
res = field[pos].item() if common.is_field(field) else field
```
or `.asscalar()` as we previously discussed. This is different from having the built-in function.

tests/next_tests/integration_tests/feature_tests/iterator_tests/test_horizontal_indirection.py
```diff
@@ -70,7 +70,7 @@ def test_simple_indirection(program_processor):
 
     ref = np.zeros(shape, dtype=inp.dtype)
     for i in range(shape[0]):
-        ref[i] = inp.ndarray[i + 1 - 1] if cond[i] < 0.0 else inp.ndarray[i + 1 + 1]
+        ref[i] = inp.ndarray[i + 1 - 1] if cond.ndarray[i] < 0.0 else inp.ndarray[i + 1 + 1]
```
havogt, line 73:
```suggestion
        ref[i] = inp.asnumpy()[i + 1 - 1] if cond.asnumpy()[i] < 0.0 else inp.asnumpy()[i + 1 + 1]
```
to use the normal pattern (and do a first step in allocating for the proper device)

tests/next_tests/integration_tests/feature_tests/iterator_tests/test_horizontal_indirection.py
```diff
@@ -101,7 +101,7 @@ def test_direct_offset_for_indirection(program_processor):
 
     ref = np.zeros(shape)
     for i in range(shape[0]):
-        ref[i] = inp[i + cond[i]]
+        ref[i] = inp.ndarray[i + cond.ndarray[i]]
```
havogt, line 104:
```suggestion
        ref[i] = inp.asnumpy()[i + cond.asnumpy()[i]]
```

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -468,10 +468,10 @@ def test_absolute_indexing_value_return():
     field = common._field(np.reshape(np.arange(100, dtype=np.int32), (10, 10)), domain=domain)
 
     named_index = ((IDim, 12), (JDim, 6))
-    value = field[named_index]
+    value = field[named_index].ndarray
```
havogt, line 471: make sure the result is still a field
```suggestion
    assert common.is_field(field[named_index])
    value = field[named_index].ndarray
```

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -468,10 +468,10 @@ def test_absolute_indexing_value_return():
...
 
-    assert isinstance(value, np.int32)
-    assert value == 21
+    assert isinstance(value, np.ndarray)
```
havogt, line 473: Maybe we should clarify this (also @egparedes):
do we care what kind the buffer is? I think we only care that the field is still a field, but the buffer could be np.int32 conceptually.

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -568,14 +568,17 @@ def test_relative_indexing_slice_3D(index, expected_shape, expected_domain):
...
+    [
+        ((1, 0), 10),
+        ((0, 1), 1),
+    ],
 )
 def test_relative_indexing_value_return(index, expected_value):
     domain = common.Domain(dims=(IDim, JDim), ranges=(UnitRange(5, 15), UnitRange(2, 12)))
     field = common._field(np.reshape(np.arange(100, dtype=int), (10, 10)), domain=domain)
-    indexed_field = field[index]
+    indexed_field = field[index].ndarray
```
havogt, line 579: I would revert this...

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -568,14 +568,17 @@ def test_relative_indexing_slice_3D(index, expected_shape, expected_domain):
...
 
-    assert indexed_field == expected_value
+    assert indexed_field == np.asarray(expected_value)
```
havogt, line 581: and do 
```suggestion
    assert indexed_field.asscalar() == expected_value
```

havogt (commented):
I had another look. I think we should introduce that function to extract the scalar value, because we have places where we need it. Going via `ndarray` seems wrong, because it doesn't have to be a ndarray field in some cases.

src/gt4py/next/embedded/operators.py
nfarabullini:
```python
field[pos].ndarray.asscalar()
Python 3.10.0 (v3.10.0:b494f5935c, Oct  4 2021, 14:59:20) [Clang 12.0.5 (clang-1205.0.22.11)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.12.3 -- An enhanced Interactive Python. Type '?' for help.
PyDev console: using IPython 8.12.3
  File "/Users/nicolettafarabullini/PycharmProjects/gt4py/.venv/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-1-0aa42db6c80a>", line 1, in <module>
    field[pos].ndarray.asscalar()
AttributeError: 'numpy.ndarray' object has no attribute 'asscalar'
```

egparedes: @havogt  means that `asscalar` or `as_scalar` should be a new member function of Field doing conceptually the same  as `ndarray.item()`.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -203,15 +203,13 @@ def remap(
...
         new_buffer = self.ndarray[buffer_slice]
-        if len(new_domain) == 0:
-            # TODO: assert core_defs.is_scalar_type(new_buffer), new_buffer
-            return new_buffer  # type: ignore[return-value] # I don't think we can express that we return `ScalarT` here
-        else:
-            return self.__class__.from_array(new_buffer, domain=new_domain)
+        if new_domain.ndim == 0:
+            assert self.array_ns.dtype(new_buffer) in core_defs.SCALAR_TYPES
+            return self._scalar_to_field(new_buffer)  # type: ignore[return-value, arg-type]
+        return self.__class__.from_array(new_buffer, domain=new_domain)
```
havogt, line 212:
```suggestion
        new_buffer = self.__class__.array_ns.asarray(new_buffer)
        return self.__class__.from_array(new_buffer, domain=new_domain)
```
would that work?

src/gt4py/next/embedded/nd_array_field.py
havogt: Shouldn't this be removed already?

src/gt4py/next/iterator/embedded.py
```diff
@@ -1086,12 +1089,12 @@ def remap(self, index_field: common.ConnectivityField | fbuiltins.FieldOffset) -
...
         raise NotImplementedError()
 
-    def restrict(self, item: common.AnyIndexSpec) -> common.Field | core_defs.int32:
+    def restrict(self, item: common.AnyIndexSpec) -> common.Field:
         if common.is_absolute_index_sequence(item) and all(common.is_named_index(e) for e in item):  # type: ignore[arg-type] # we don't want to pollute the typing of `is_absolute_index_sequence` for this temporary code # fmt: off
             d, r = item[0]
             assert d == self._dimension
             assert isinstance(r, int)
-            return self.dtype.scalar_type(r)
+            return np.asarray(r)  # type: ignore[return-value] # Field is a superset
```
havogt, line 1097: this is not correct (ndarray is not a Field), but I have to think about how to do it...

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -468,10 +468,11 @@ def test_absolute_indexing_value_return():
     field = common._field(np.reshape(np.arange(100, dtype=np.int32), (10, 10)), domain=domain)
 
     named_index = ((IDim, 12), (JDim, 6))
+    assert common.is_field(field)
     value = field[named_index]
 
-    assert isinstance(value, np.int32)
-    assert value == 21
+    assert isinstance(value, np.ndarray)
```
havogt, line 474:
```suggestion
    assert common.is_field(value)
```

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -468,10 +468,11 @@ def test_absolute_indexing_value_return():
...
+    assert value == np.asarray(21)
```
havogt, line 475:
```suggestion
    assert value.item() == 21
```

src/gt4py/next/embedded/operators.py
havogt: yes

havogt (changes requested):
I have to look at the details again tomorrow. We can talk in the morning...

src/gt4py/next/common.py
```diff
@@ -896,9 +900,20 @@ def ndarray(self) -> Never:
...
+        assert self._cur_index is not None
+        return self._cur_index
+
     @functools.cached_property
     def domain(self) -> Domain:
-        return Domain(dims=(self.dimension,), ranges=(UnitRange.infinite(),))
+        if self._cur_index is None:
+            return Domain(dims=(self.dimension,), ranges=(UnitRange.infinite(),))
+        else:
+            return Domain()
```
havogt, line 916: @egparedes let's discuss these changes...

src/gt4py/next/common.py
```diff
@@ -623,14 +623,17 @@ def asnumpy(self) -> np.ndarray: ...
     def remap(self, index_field: ConnectivityField | fbuiltins.FieldOffset) -> Field: ...
 
     @abc.abstractmethod
-    def restrict(self, item: AnyIndexSpec) -> Field | core_defs.ScalarT: ...
+    def restrict(self, item: AnyIndexSpec) -> Field: ...
+
+    @abc.abstractmethod
+    def as_scalar(self) -> core_defs.ScalarT: ...
```
havogt, line 629: @egparedes Wondering if Field should not be a protocol (or we should separate an ABC), because then we could do the check for 0d field and put the concrete implementation in a separate function. but would become obsolete if we refactor to a single field implementation

egparedes: I don't completely follow because in any case a Protocol is an ABC, so you can put implementation details in the protocol for the case where other specific implementations want to use it as an ABC and inherit from it. With Protocols the inheritance is optional if implementations provide exactly the same interface.

src/gt4py/next/common.py
egparedes: ok, let's discuss it. Maybe the 0d-index should be a special case of the Domain class, not of CartesianConnectivity...

egparedes (approved):
ILGTM

havogt (approved):

PR #1430, opened 2024-01-29
fix[next][dace]: Bugfix in deref (dynamic memory allocation)

edopao:
## Description

Baseline contained a bug in the lowering of `deref` in the context of neighbor reduction. The data container should be statically allocated with size equal to the `max_neighbors` attribute in the offset provider.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
@@ -763,7 +763,6 @@ class PythonTaskletCodegen(gt4py.eve.codegen.TemplatedGenerator):
             # already a list of ValueExpr
             return iterator
 
-        args: list[ValueExpr]
         sorted_dims = sorted(iterator.dimensions)
         if all([dim in iterator.indices for dim in iterator.dimensions]):
             # The deref iterator has index values on all dimensions: the result will be a scalar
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -781,16 +780,16 @@ def _visit_deref(self, node: itir.FunCall) -> list[ValueExpr]:
...
-            # Not all dimensions are included in the deref index list:
-            # this means the ND-field will be sliced along one or more dimensions and the result will be an array
-            field_array = self.context.body.arrays[iterator.field.data]
-            result_shape = tuple(
-                dim_size
-                for dim, dim_size in zip(sorted_dims, field_array.shape)
-                if dim not in iterator.indices
-            )
+            dims_not_indexed = [dim for dim in iterator.dimensions if dim not in iterator.indices]
+            assert len(dims_not_indexed) == 1
```
philip-paul-mueller, line 784: I would suggest to throw a `NotImplementedError` here with the description `Reduction of more than 1D fields is currently not implemented.`
Or does this even make sense?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -800,19 +799,17 @@ def _visit_deref(self, node: itir.FunCall) -> list[ValueExpr]:
...
+            ] + [dace.Memlet(data=node.data, subset="0") for node in deref_nodes[1:]]
 
             # we create a mapped tasklet for array slicing
+            index_name = unique_name(f"_i_{neighbor_dim}")
             map_ranges = {
-                f"_i_{dim}": f"0:{size}"
-                for dim, size in zip(sorted_dims, field_array.shape)
-                if dim not in iterator.indices
+                index_name: f"0:{offset_provider.max_neighbors}",
             }
```
philip-paul-mueller, line 810: I do not get how you handle the case that you have less than `max_neighbors` to reduce over?

philip-paul-mueller (commented):
It generally looks good, but there is something I do not get.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: The length check is preparatory to use the dimension to select the offset provider, on next line. I have not seen so far expressions with multiple offset providers at the same time.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: This is the case where we dereference a connectivity table, given by the offset provider. We can always dereference the connectivity table, but we might get `-1` if no neighbor is present. The gathering of neighbors is handled by `builtin_neighbors`, where we perform the check on connectivity index != -1.

philip-paul-mueller: Sorry forgot that.

philip-paul-mueller (approved):

PR #1431, opened 2024-01-30
feat[next][dace]: Add support for lift expressions in neighbor reductions (no unrolling)

edopao:
## Description

Baseline dace backend forced unroll of neighbor reductions, in the ITIR pass, in order to eliminate all lift expressions. This PR adds support for lowering of lift expressions in neighbor reductions, thus avoiding the need to unroll reduce expressions. The result is a more compact SDFG, which leaves to the optimization backend the option of unrolling neighbor reductions.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
@@ -260,8 +260,6 @@ class ItirToSDFG(eve.NodeVisitor):
     def visit_StencilClosure(
         self, node: itir.StencilClosure, array_table: dict[str, dace.data.Array]
     ) -> tuple[dace.SDFG, list[str], list[str]]:
-        assert ItirToSDFG._check_no_lifts(node)
-
         # Create the closure's nested SDFG and single state.
         closure_sdfg = dace.SDFG(name="closure")
         closure_sdfg.debuginfo = dace_debuginfo(node)
@@ -461,7 +459,7 @@ class ItirToSDFG(eve.NodeVisitor):
         assert isinstance(node.output, SymRef)
         neighbor_tables = filter_neighbor_tables(self.offset_provider)
         input_names = [str(inp.id) for inp in node.inputs]
-        connectivity_names = [connectivity_identifier(offset) for offset, _ in neighbor_tables]
+        connectivity_names = [connectivity_identifier(offset) for offset in neighbor_tables.keys()]
 
         # find the scan dimension, same as output dimension, and exclude it from the map domain
         map_ranges = {}
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
@@ -260,6 +260,8 @@ class ItirToSDFG(eve.NodeVisitor):
     def visit_StencilClosure(
         self, node: itir.StencilClosure, array_table: dict[str, dace.data.Array]
     ) -> tuple[dace.SDFG, list[str], list[str]]:
+        assert ItirToSDFG._check_no_lifts(node)
+
         # Create the closure's nested SDFG and single state.
         closure_sdfg = dace.SDFG(name="closure")
         closure_sdfg.debuginfo = dace_debuginfo(node)
@@ -673,11 +675,14 @@ class ItirToSDFG(eve.NodeVisitor):
 
     @staticmethod
     def _check_no_lifts(node: itir.StencilClosure):
-        if any(
-            getattr(fun, "id", "") == "lift"
-            for fun in eve.walk_values(node).if_isinstance(itir.FunCall).getattr("fun")
-        ):
-            return False
+        # allow lift only to appear as child node in neighbor reductions
+        neighbors_call_count = 0
+        for fun in eve.walk_values(node).if_isinstance(itir.FunCall).getattr("fun"):
+            if getattr(fun, "id", "") == "neighbors":
+                neighbors_call_count = 3
+            elif getattr(fun, "id", "") == "lift" and neighbors_call_count != 1:
+                return False
+            neighbors_call_count = max(0, neighbors_call_count - 1)
         return True
 
     @staticmethod
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
@@ -188,32 +188,56 @@ def _visit_lift_in_neighbors_reduction(
     offset_provider: NeighborTableOffsetProvider,
     map_entry: dace.nodes.MapEntry,
     map_exit: dace.nodes.MapExit,
-    index_node: dace.nodes.AccessNode,
+    neighbor_index_node: dace.nodes.AccessNode,
     result_node: dace.nodes.AccessNode,
 ) -> list[ValueExpr]:
     neighbor_dim = offset_provider.neighbor_axis.value
-    lifted_args = [
-        IteratorExpr(
-            arg.field,
-            {
-                neighbor_dim: index_node,
-            },
-            arg.dtype,
-            arg.dimensions,
-        )
-        if isinstance(arg, IteratorExpr)
-        else arg[0]
-        for arg in node_args
-    ]
+    origin_dim = offset_provider.origin_axis.value
+
+    lifted_args: list[IteratorExpr | ValueExpr] = []
+    iterator_index_nodes: list[dace.nodes.AccessNode] = []
+    for arg in node_args:
+        if isinstance(arg, IteratorExpr):
+            iterator_index_nodes += arg.indices.values()
+            if origin_dim in arg.indices:
+                lifted_indices = arg.indices.copy()
+                lifted_indices.pop(origin_dim)
+                lifted_indices[neighbor_dim] = neighbor_index_node
+                lifted_args.append(
+                    IteratorExpr(
+                        arg.field,
+                        lifted_indices,
+                        arg.dtype,
+                        arg.dimensions,
+                    )
+                )
+            else:
+                lifted_args.append(arg)
+        else:
+            lifted_args.append(arg[0])
+
     lift_context, inner_inputs, inner_outputs = transformer.visit(node.args[0], args=lifted_args)
     assert len(inner_outputs) == 1
     inner_out_connector = inner_outputs[0].value.data
 
-    inner_index_connectors = [
-        inner_index_table[neighbor_dim] for (_, inner_index_table), _ in inner_inputs
-    ]
+    iterator_field_nodes = {}
+    iterator_index_mapping = {}
+    lifted_index_connectors = []
+    for (field_connector, inner_index_table), iterator in filter(
+        lambda x: isinstance(x[1], IteratorExpr), inner_inputs
+    ):
+        iterator_field_nodes[field_connector] = iterator.field
+        for dim, connector in inner_index_table.items():
+            if dim == neighbor_dim:
+                lifted_index_connectors.append(connector)
+            iterator_index_mapping[connector] = iterator.indices[dim]
 
-    arg_field_nodes = [inner_iterator.field for _, inner_iterator in inner_inputs]
+    input_value_nodes = {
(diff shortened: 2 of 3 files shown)


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -419,6 +570,45 @@ def builtin_cast(
...
 
 
+def builtin_make_const_list(
+    transformer: "PythonTaskletCodegen", node: itir.Expr, node_args: list[itir.Expr]
+) -> list[ValueExpr]:
+    args = [transformer.visit(arg)[0] for arg in node_args]
+    assert all([isinstance(x, (SymbolExpr, ValueExpr)) for x in args])
+    args_dtype = set([x.dtype for x in args])
+    assert len(args_dtype) == 1
+    dtype = next(iter(args_dtype))
```
philip-paul-mueller, line 580:
```python
args_dtype = [x.dtype for x in args]
assert len(set(args_dtype)) == 1
dtype = args_dtype[0]
```
- Much nicer to read
- If asserts are disabled, also faster.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -419,6 +570,45 @@ def builtin_cast(
...
+    tasklet_input_mapping = {}
+    for i, arg in enumerate(args):
+        rhs = "out" if len(args) == 1 else f"out[{i}]"
+        if isinstance(arg, SymbolExpr):
+            code += f"{rhs} = {arg.value}\n"
+        else:
+            connector = f"{arg.value.data}_v"
+            code += f"{rhs} = {connector}\n"
+            tasklet_input_mapping[connector] = arg.value
+
```
philip-paul-mueller, line 597: If I understand that code correctly, you add multiple writes in a single Tasklet?
I am not saying that you should change it, but I do not like it.
Although it would blow up the SDFG, but you could generate individual Tasklets.
It would allow DaCe move them arround, because now it has to keep the lhs alive until all elements are computed.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -419,6 +570,45 @@ def builtin_cast(
...
+    )
+    for connector, access_node in tasklet_input_mapping.items():
+        assert access_node.desc(transformer.context.body).shape == (1,)
+        transformer.context.state.add_edge(
+            access_node, None, tasklet, connector, dace.Memlet(access_node.data, "0")
+        )
+    transformer.context.state.add_edge(
+        tasklet, "out", var_node, None, dace.Memlet.from_array(var_name, var_array)
+    )
+    return [ValueExpr(var_node, dtype)]
```
philip-paul-mueller, line 609: Before you were adding Memlet Paths, but why not.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -973,10 +1164,15 @@ def _visit_reduce(self, node: itir.FunCall):
...
-                [arg.value.desc(self.context.body).shape == nreduce_shape for arg in args[2:]]
+            args_shape = set(
+                [
+                    arg.value.desc(self.context.body).shape
+                    for arg in args
+                    if arg.value.desc(self.context.body).shape != (1,)
+                ]
             )
+            assert len(args_shape) == 1
+            nreduce_shape = next(iter(args_shape))
```
philip-paul-mueller, line 1175: See comment above.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -994,7 +1190,9 @@ def _visit_reduce(self, node: itir.FunCall):
             )
 
             input_mapping = {
-                param: create_memlet_at(arg.value.data, nreduce_index)
+                param: dace.Memlet(data=arg.value.data, subset="0")
+                if arg.value.desc(self.context.body).shape == (1,)
+                else create_memlet_at(arg.value.data, nreduce_index)
```
philip-paul-mueller, line 1195: I think we should patch `Memlet` to allow pass either a string or a list of strings which is then interpreted as `', '.join()` because I have seed that a lot as well and it is really cumbersome to write.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -419,6 +570,45 @@ def builtin_cast(
...
+    code = ""
+    tasklet_input_mapping = {}
+    for i, arg in enumerate(args):
+        rhs = "out" if len(args) == 1 else f"out[{i}]"
+        if isinstance(arg, SymbolExpr):
+            code += f"{rhs} = {arg.value}\n"
+        else:
+            connector = f"{arg.value.data}_v"
+            code += f"{rhs} = {connector}\n"
+            tasklet_input_mapping[connector] = arg.value
```
philip-paul-mueller, line 596: Couldn't you use a Memlet here instead of a tasklet?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -675,11 +675,14 @@ def _visit_domain(
 
     @staticmethod
```
philip-paul-mueller, line 676: Do you need to keep it a static function?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -181,45 +181,179 @@ def __init__(
...
-    assert isinstance(iterator, IteratorExpr)
-    field_desc = iterator.field.desc(transformer.context.body)
+    neighbor_dim = offset_provider.neighbor_axis.value
+    origin_dim = offset_provider.origin_axis.value
+
+    lifted_args: list[IteratorExpr | ValueExpr] = []
+    iterator_index_nodes: list[dace.nodes.AccessNode] = []
+    for arg in node_args:
+        if isinstance(arg, IteratorExpr):
+            iterator_index_nodes += arg.indices.values()
```
philip-paul-mueller, line 201: It is a matter of choice, but did you considered using `.extend()`?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -181,45 +181,179 @@ def __init__(
...
+        iterator_field_nodes[field_connector] = iterator.field
+        for dim, connector in inner_index_table.items():
+            if dim == neighbor_dim:
+                lifted_index_connectors.append(connector)
+            iterator_index_mapping[connector] = iterator.indices[dim]
+
+    input_value_nodes = {
+        connector: expr.value
+        for connector, expr in filter(lambda x: isinstance(x[1], ValueExpr), inner_inputs)
+    }
```
philip-paul-mueller, line 238: Why not
```python
for x, y in inner_inputs:
    if isinstance(y, IteratorExpr):
        ...
    else:
        assert isinstance(y, ValueExpr)
        ...
``` 
Then you could also merge in the `input_nodes` construction.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -181,45 +181,179 @@ def __init__(
...
+        parent_state.add_memlet_path(
+            access_node,
+            map_entry,
+            nested_sdfg_node,
+            dst_conn=inner_connector,
+            memlet=input_mapping[inner_connector],
+        )
+
+    for inner_connector, access_node in iterator_index_mapping.items():
+        if inner_connector in lifted_index_connectors:
```
philip-paul-mueller, line 285: `lifted_index_connectors` is a `list` lokking might be inefficient, could you use a `set` instead?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -257,34 +377,84 @@ def builtin_neighbors(
...
+                f" if {neighbor_check} else {transformer.context.reduce_identity.value}"
+                if offset_provider.has_skip_values
+                else ""
+            ),
+            inputs={"__field", field_index},
+            outputs={"__result"},
+            debuginfo=di,
+        )
+        state.add_edge(
+            neighbor_index_node,
```
philip-paul-mueller, line 431: I do not get who does the iteration, i.e. who is gonna increase this value.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -299,32 +469,13 @@ def builtin_can_deref(
...
-            None,
-            dace.Memlet(data=result_name, subset="0", debuginfo=di),
-        )
-        return [ValueExpr(result_node, dace.dtypes.bool)]
+    assert isinstance(iterator, IteratorExpr)
 
     # create tasklet to check that field indices are non-negative (-1 is invalid)
     args = [ValueExpr(access_node, _INDEX_DTYPE) for access_node in iterator.indices.values()]
     internals = [f"{arg.value.data}_v" for arg in args]
     expr_code = " and ".join([f"{v} >= 0" for v in internals])
```
philip-paul-mueller, line 477: You could optimize this by querying `has_skip_values`.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -257,34 +377,84 @@ def builtin_neighbors(
...
+            neighbor_index_node,
+            None,
+            data_access_tasklet,
+            field_index,
+            dace.Memlet(data=neighbor_index_name, subset="0"),
+        )
+        # select full shape only in the neighbor-axis dimension
+        field_subset = tuple(
+            f"0:{shape}" if dim == neighbor_dim else f"i_{dim}"
+            for dim, shape in zip(sorted(iterator.dimensions), field_desc.shape)
```
philip-paul-mueller, line 440: If you sort, i.e. change the order of the dimensions, should then not also the order of the shape changes as well?

philip-paul-mueller (changes requested):
Also at some points you have forgotten to add the debug information.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
philip-paul-mueller: Thanks for the explanation, but I was more looking for the map itself.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
philip-paul-mueller: Right the dimensions of the dace array are already in the correct order.
Maybe you should add a comment or something, but on the other hand, probably not.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -419,6 +579,40 @@ def builtin_cast(
...
+                ),
+                "val",
+                var_node,
+                None,
+                dace.Memlet(data=var_name, subset=f"{i}"),
+            )
+        else:
+            assert arg.value.desc(transformer.context.body).shape == (1,)
+            transformer.context.state.add_nedge(
+                arg.value, var_node, dace.Memlet(data=var_name, subset=f"{i}", other_subset="0")
```
philip-paul-mueller, line 610: Correct me if I am wrong but I think that the Memlet is wrong, shouldn't `subset` and `other_subset` be switched?
And data should be `arg`?

philip-paul-mueller (commented):
There is one point where I am not sure, otherwise it looks fine.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: Good catch, fixed.

philip-paul-mueller (approved):

PR #1436, opened 2024-01-31
test[next]: fix obsolete asarray

havogt:


tehrengruber (approved):
I now understand my recollection of this error again. This test unconditionally passes: https://github.com/GridTools/gt4py/blob/182dfa1e5c6a2f81868f7886bfb5ce8a202794ae/tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py#L476 It's a minor thing though... Just doing `np.full_like(offset_field.ndarray, True, dtype=bool)`  should be enough. Other than that it looks good.

PR #1437, opened 2024-01-31
build: Update gridtools-cpp version to 2.3.2

havogt:


pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = [
   'deepdiff>=5.6.0',
   'devtools>=0.6',
   'frozendict>=2.3',
-  'gridtools-cpp>=2.3.1,==2.*',
```
havogt, line 37: not strictly needed to bump the min version, but since gridtools-cpp doesn't have external dependencies it's a safe choice.

egparedes (approved):
ILGTM

PR #1438, opened 2024-01-31
feature[next]: toolchain configuration interfaces

DropD:
## Description

Simplify workflow configuration / creation by providing composable factories (in this case using [factoryboy](https://factoryboy.readthedocs.io/en/stable/index.html)).

I recomend starting your assessment by comparing how the various `run_gtfn_XX` backends are created in `src/gt4py/next/program_processors/runners/gtfn.py`. Then look at the factories and optionally think about providing the same or better usage with a different library or pattern.

## Requirements

- [X] All the authors are covered by a valid contributor assignment agreement provided to ETH Zurich and signed by the employer if needed.
- [X] The PR contains an updated version of the `AUTHORS.md` file adding the names of all the new contributors.


src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -113,103 +114,84 @@ def compilation_hash(otf_closure: stages.ProgramCall) -> int:
...
-    device_type=core_defs.DeviceType.CUDA,
-)
+    class Params:
+        device_type: core_defs.DeviceType = core_defs.DeviceType.CPU
 
-GTFN_DEFAULT_COMPILE_STEP: step_types.CompilationStep = compiler.Compiler(
-    cache_strategy=cache.Strategy.SESSION, builder_factory=compiledb.CompiledbFactory()
-)
+    translation = factory.SubFactory(
+        gtfn_module.GTFNTranslationStepFactory, device_type=factory.SelfAttribute("..device_type")
```
edopao, line 125: What is the meaning of 2 dots `..device_type`?

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -113,103 +114,84 @@ def compilation_hash(otf_closure: stages.ProgramCall) -> int:
...
-        name="run_gtfn_with_temporaries",
-        otf_workflow=gtfn_executor.otf_workflow.replace(
-            translation=gtfn_executor.otf_workflow.translation.replace(
-                lift_mode=LiftMode.FORCE_TEMPORARIES
-            ),
-        ),
-    ),
-    allocator=next_allocators.StandardCPUFieldBufferAllocator(),
+run_gtfn_imperative = GTFNBackendFactory(
+    otf_workflow__translation__use_imperative_backend=True, **__user_defaults
```
edopao, line 184: Is there a way to shorten these names? (`otf_workflow__translation__use_imperative_backend`)

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -113,103 +114,84 @@ def compilation_hash(otf_closure: stages.ProgramCall) -> int:
...
-    allocator=next_allocators.StandardGPUFieldBufferAllocator(),
+run_gtfn_cached = GTFNBackendFactory(**__user_defaults | {"cached": True})
+
+run_gtfn_with_temporaries = GTFNBackendFactory(
+    name="run_gtfn_with_temporaries",
+    otf_workflow__translation__lift_mode=LiftMode.FORCE_TEMPORARIES,
+    **__user_defaults,
 )
 
+run_gtfn_gpu = GTFNBackendFactory(**__user_defaults | {"gpu": True})
```
edopao, line 195: I like this syntax.

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -113,103 +114,84 @@ def compilation_hash(otf_closure: stages.ProgramCall) -> int:
...
+                )
+            ),
+            cached_name="_cached",
+        )
+        device_type = core_defs.DeviceType.CPU
+        hash_function = compilation_hash
+        otf_workflow = factory.SubFactory(
+            GTFNCompileWorkflowFactory, device_type=factory.SelfAttribute("..device_type")
+        )
+        name = factory.LazyAttribute(lambda o: f"run_gtfn_{o.device_name}{o.cached_name}")
```
edopao, line 166: Maybe we do not want to construct a name, do we? We should just use the object created by the factory.

src/gt4py/next/program_processors/runners/gtfn.py
DropD: In this context it means "look up the attribute on the parent of the SubFactory": https://factoryboy.readthedocs.io/en/stable/reference.html#parents

src/gt4py/next/program_processors/runners/gtfn.py
DropD: It's not a name, it's a path equivalent to `otf_workflow.translation.use_imperative_backend`.  However, that would be invalid python syntax in a keyword arg, so the authors of factoryboy went with `__` instead.

DropD: The answer is still yes though, we can shorten them in two ways:
- make the individual attribute names shorter
- provide a parameter (under `class Params:`) in the parent factory with a shorter name and pass that on to the subfactory. (The full path attribute would still be available though.)

src/gt4py/next/program_processors/runners/gtfn.py
DropD: The factory has to give a name to the object it creates. In practice the name attribute of the executor is mainly used in pytest to see at a glance which test was parametrized with what type of gtfn backend.

src/gt4py/next/config.py
```diff
@@ -0,0 +1,47 @@
...
+    def _generate_next_value_(name, start, count, last_values):
+        return "".join(part.capitalize() for part in name.split("_"))
+
+    DEBUG = enum.auto()
+    RELEASE = enum.auto()
+    REL_WITH_DEB_INFO = enum.auto()
+    MIN_SIZE_REL = enum.auto()
+
+
+PREFIX = "GT4PY_GTX"
```
havogt, line 37:
```suggestion
_PREFIX = "GT4PY"
```
I prefer to use the same variables as in gt4py.cartesian where possible (and their names are good, if they are broken we should fix them here and maybe fallback to them in cartesian).

src/gt4py/next/config.py
```diff
@@ -0,0 +1,47 @@
...
+
```
havogt, line 38: My main driver for this project is the debug mode. I'd like to propose to add it here.
```suggestion
DEBUG = ...
```
which then switches some defaults, e.g. cache dir to persistent (but overwrite with env if provided) and built_type to debug.

src/gt4py/next/config.py
```diff
@@ -0,0 +1,47 @@
...
+BUILD_CACHE_DIR = (
+    pathlib.Path(os.environ.get(f"{PREFIX}_BUILD_CACHE_DIR", tempfile.gettempdir())) / "gt4py_cache"
+)
+
+BUILD_CACHE_LIFETIME = getattr(
+    BuildCacheLifetime, os.environ.get(f"{PREFIX}_BUILD_CACHE_LIFETIME", "session").upper()
+)
+
+CMAKE_BUILD_TYPE = getattr(CMakeBuildType, os.environ.get(f"{PREFIX}_BUILD_TYPE", "debug").upper())
```
havogt, line 47:
```suggestion
CMAKE_BUILD_TYPE = getattr(CMakeBuildType, os.environ.get(f"{PREFIX}_BUILD_TYPE", "debug" if DEBUG else "release").upper())
```

docs/development/ADRs/0017-Toolchain-Configuration.md
```diff
@@ -0,0 +1,101 @@
...
+
+- an internal name (a module level variable)
+- an external name used to load from environment variables (possibly with a common prefix)
+- a fallback default value in case no environment variable is defined
+
+Any other toolchain option is considered an implementation detail.
+
+### Limit the times when configuration can change
+
+By making the `gt4py.next.config` module contain module level variables with user configuration, we ensure user configuration can only be changed between python interpreter runs (after `from gt4py import next` the configuration is fixed). Of course there are ways around it but they should be considered unsupported as they are difficult to make reliable (consider monkey patching as an example).
```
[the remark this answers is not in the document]
havogt: I found the following comment in the test
>Because monkey patching the config variables is not enough, as
> other variables are computed at import time based on them.

which explains it partly. Basically if changing is limited depends on how it's used. That's potentially dangerous.

Also a question: do you consider changing module level variables monkey patching?

havogt (commented):
First round of comments. Didn't look at the factories yet.

docs/development/ADRs/0017-Toolchain-Configuration.md
tehrengruber, line 49:
```suggestion
By making the `gt4py.next.config` module contain module level variables with user configuration, we ensure user configuration can only be changed between python interpreter runs (after `from gt4py import next` the configuration is fixed).
```
I don't agree with this sentence. This doesn't ensure anything. I do agree with the `unsupported` phrasing later on though.

docs/development/ADRs/0017-Toolchain-Configuration.md
```diff
@@ -0,0 +1,101 @@
+---
```
tehrengruber, line 1: I find a good part of this document not useful to read, but also I don't care too much.

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -113,104 +114,85 @@ def compilation_hash(otf_closure: stages.ProgramCall) -> int:
...
-        otf_workflow=gtfn_executor.otf_workflow.replace(
-            translation=gtfn_executor.otf_workflow.translation.replace(
-                lift_mode=LiftMode.FORCE_TEMPORARIES,
-                temporary_extraction_heuristics=global_tmps.SimpleTemporaryExtractionHeuristics,
-            ),
-        ),
-    ),
-    allocator=next_allocators.StandardCPUFieldBufferAllocator(),
+run_gtfn_with_temporaries = GTFNBackendFactory(
+    name_postfix="_with_temporaries",
```
tehrengruber, line 191: Lets make this a trait.

docs/development/ADRs/0017-Toolchain-Configuration.md
```diff
@@ -0,0 +1,101 @@
...
+
+- when a toolchain instance is created, which disregards user config, or
+- when such a toolchain is used
+
+The latter was never tried due to the obvious run time overhead in checking before every use. The former turned out not to be very useful for two reasons:
+
+- such a toolchain may have been created but not used to do work for the end user
+- without proper tracking of where configuration comes from, false positives as well as false negatives could not be eliminated.
+
+Implementing tracking was briefly considered but looked like it would be too heavy weight to justify the maintenance burden.
```
tehrengruber, line 93: This paragraph doesn't really sum up the experiment we did. Also where is the
```
class MyStep:
    config_flag = True

os.environ["GT4PY_MYSTEP_CONFIG_FLAG"] = True

my_backend = MyStepFactory(mystep_config_flag=False)
```
part or the sketch of the merging algorithm?

docs/development/ADRs/0017-Toolchain-Configuration.md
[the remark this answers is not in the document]
DropD: That depends a bit. If I am changing something that is assumed to change throughout the code and I want it to stay changed, probably not. If, like in this case changing it might leave things in an inconsistent state (basically undefined behaviour) and I want the change to only happen locally and revert afterwards, I would describe it as monkey patching.

docs/development/ADRs/0017-Toolchain-Configuration.md
```diff
@@ -0,0 +1,101 @@
...
+- an external name used to load from environment variables (possibly with a common prefix)
+- a fallback default value in case no environment variable is defined
+
+Any other toolchain option is considered an implementation detail.
+
+### Limit the times when configuration can change
+
+By making the `gt4py.next.config` module contain module level variables with user configuration, we ensure user configuration can only be changed between python interpreter runs (after `from gt4py import next` the configuration is fixed). Of course there are ways around it but they should be considered unsupported as they are difficult to make reliable (consider monkey patching as an example).
+
+### Environment variables are the primary end user interface
```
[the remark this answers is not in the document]
DropD: All of these are choices, nothing is needed. The only choice which is fairly well justified is to not try to warn users of "rogue" toolchains, because most of the cycle went into experimentation for that.

src/gt4py/next/config.py
```diff
@@ -0,0 +1,68 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+from __future__ import annotations
+
+import enum
+import os
+import pathlib
+import tempfile
+
+
+class BuildCacheLifetime(enum.Enum):
```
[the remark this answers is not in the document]
DropD: Perhaps, if they start accumulating they should get a place of their own.

docs/development/ADRs/0017-Toolchain-Configuration.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Limit configuration options exposed to the end user
+
+Any option that the end user can change in order to influence the toolchain behavior must be defined in `gt4py.next.config` with
+
+- an internal name (a module level variable)
+- an external name used to load from environment variables (possibly with a common prefix)
+- a fallback default value in case no environment variable is defined
+
+Any other toolchain option is considered an implementation detail.
```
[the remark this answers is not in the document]
DropD: It is not a requirement, simply a description of the status quo. The options not exposed to the end user are implementation details from the point of view of the end user. The end user is someone who runs code that uses GT4Py internally but may not be aware of that.

docs/development/ADRs/0017-Toolchain-Configuration.md
DropD: I guess the suggestion got a bit jumbled up then?

It does indeed not ensure anything, except that changing configuration during a python interpreter run might leave the configuration in an inconsistent state.

src/gt4py/next/program_processors/runners/gtfn.py
DropD: Either the suggestion got jumbled up again or there is a misunderstanding about what factoryboy traits are.

Assuming the former:
You might be talking about switching `lift_mode` and `temporary_extraction_heuristics` based on a trait?

Assuming the latter, you would be talking about something like:
```python
class GTFNBackendFactory(factory.Factory):
    ...
    class Params:
        name_postfix = factory.Trait(
            ... and now what?
        )
        ...
        
# usage
GTFNBackendFactory(name_postfix=True)
```

The effect is to negate the purpose of the `name_postfix` parameter, which is to give custom name postfixes. In either case, this would be an easy post-project cooldown PR. The scope of this PR is now fixed.

docs/development/ADRs/0017-Toolchain-Configuration.md
DropD: The usefulness of this document depends heavily on whether one is thinking about changing any of this design without re-covering the same ground.

Is the lack of interest and perceived usefulness due to that you have no desire to do so? Or do you think it is useless under the premise of wanting to change the design. Or does the writing style distract from the actual information?

src/gt4py/next/config.py
```diff
@@ -0,0 +1,68 @@
...
+        return "".join(part.capitalize() for part in name.split("_"))
+
+    DEBUG = enum.auto()
+    RELEASE = enum.auto()
+    REL_WITH_DEB_INFO = enum.auto()
+    MIN_SIZE_REL = enum.auto()
+
+
+def env_flag_to_bool(flag_value: str) -> bool:
+    """Like in gt4py.cartesian, env vars for flags should be set to '0' or '1'."""
```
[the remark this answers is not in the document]
DropD: Thank you, that would indeed be helpful.

src/gt4py/next/config.py
```diff
@@ -0,0 +1,68 @@
...
+#: - PERSISTENT: generated code projects are written to BUILD_CACHE_DIR and persist between runs
+BUILD_CACHE_LIFETIME = getattr(
+    BuildCacheLifetime,
+    os.environ.get(f"{_PREFIX}_BUILD_CACHE_LIFETIME", "persistent" if DEBUG else "session").upper(),
+)
+
+#: Build type to be used when CMake is used to compile generated code.
+#: Might have no effect when CMake is not used as part of the toolchain.
+CMAKE_BUILD_TYPE = getattr(
+    CMakeBuildType, os.environ.get(f"{_PREFIX}_BUILD_TYPE", "debug" if DEBUG else "release").upper()
```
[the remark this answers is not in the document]
DropD: I will go with the former for now, if people who use it often feel like there is a strong case for the latter, it can be changed at any point. It would also be thinkable to use one as a fallback for the other.

docs/development/ADRs/0017-Toolchain-Configuration.md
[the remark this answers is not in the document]
tehrengruber: But the user can change the option I linked [see here for an actual case](https://github.com/C2SM/icon4py/blob/a32640c8bacc65bbafe41ec9a68b772c7a3c1853/tools/src/icon4pytools/icon4pygen/backend.py#L139), so in that sense it is not a proper description of the status quo in my opinion.

docs/development/ADRs/0017-Toolchain-Configuration.md
tehrengruber: Wasn't meant to be a suggestion, but a quote.

docs/development/ADRs/0017-Toolchain-Configuration.md
tehrengruber: No it's not a lack of interest I just don't feel addressed neither as a user trying to understand how things work (not really the purpose of an ADR sure), but also not as someone trying to understand design decisions (e.g. why would I care about the definition of the term toolchain to mention just one thing).

src/gt4py/next/program_processors/runners/gtfn.py
tehrengruber: Not sure why you mention the `name_postfix` parameter. I want something like this:
```suggestion
run_gtfn_with_temporaries = GTFNBackendFactory(use_temporaries=True)
```
because for many purposes the lift_mode and heuristics is a low-levle detail I don't care about as a library user.

docs/development/ADRs/0017-Toolchain-Configuration.md
tehrengruber: The merging algorithm I proposed has traceability:
```
env_defined_vars: dict[str, Any]
config_file_defined_vars: dict[str, Any]
code_defined_vars: dict[str, Any]

var_sources = {
  "env": env_defined_vars,
  "config": config_file_defined_vars,
  "in-code": code_defined_vars
}

effects_per_var_source: dict[str, dict[str, list[str]]] = []

for var_source, vars in var_sources.items():
  res = factory(**vars)
  # compute what attributes are different in res from default, i.e. factory()
  effect = ...
  effects_per_var_source[var_source][var] = effect

for (source_a, effect_a), (source_b, effect_b) in itertools.product(effects_per_var_source.items(), effects_per_var_source.items())
  for name_outer, effect_outer in user_defined_vars_effects.items():
    for name_inner, effect_inner in code_defined_vars_effects.items():
      if intersection(effect_outer, effect_inner):
        print("{source_a}:{name_outer} conflicts with {source_b}:{name_inner}")
```
It prints both source of the conflict and can straightforward be extended to give even more (e.g. line numbers if the sources are from config files). But my main point is I don't see this paragraph to reflect what we concluded the experiment with:

> Algorithmically, warnings are generally possible, but not with FactoryBoy for Defaults. Extensions to multiple SOT are generally also possible and in the simple version also feasible, but whether that is still satisfactory is questionable (because other projects also do not do that at all).

DropD: Ah, I will add another paragraph about black box analysis of the effects of different configuration sources and clarify that this one is about **tracking** where which value comes from.

DropD: Also, I don't undestand why you would mention `factory-boy`. The black-box analysis is agnostic, it can always only easily track to the highest level of toolchain building logic. As soon as the user configuration does not apply directly anymore (but only indirectly via toolchain building logic), it would involve replicating some of that logic, or tracking (as in annotating any configuration with it's source and keeping and updating that annotation at every level of logic).

src/gt4py/next/program_processors/runners/gtfn.py
DropD: The github view makes it look like you are talking specifically about the line containing `name_postfix`. As I thought, this turns out to be accidental.

The need to change "lift_mode" and "heuristics" in synch arose only as a result of merging with main after the end of the project. Therefore it remains out of scope for this PR.

docs/development/ADRs/0017-Toolchain-Configuration.md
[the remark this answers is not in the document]
DropD: Since your example is in code, it can by definition not have been done by an end user.

docs/development/ADRs/0017-Toolchain-Configuration.md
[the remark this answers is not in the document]
DropD: I reworded this to be more clear (hopefully)

docs/development/ADRs/0017-Toolchain-Configuration.md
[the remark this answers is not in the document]
DropD: I have hopefully made it more clear that this is not a requirement, but a decision that was made for now. I have outlined a sketch of what the alternative could look like (in the alternatives section).

docs/development/ADRs/0017-Toolchain-Configuration.md
DropD: I have changed the wording to be more careful and precise.

docs/development/ADRs/0017-Toolchain-Configuration.md
DropD: The definition is provided so that it is clear what the design decisions have been taken for. It is meant to be skipped but then double checked before taking misguided action based on something one has read in the document. Same as all other definitions.

If we had clearly defined terms for components of GT4Py this would not be necessary.

tehrengruber (approved):

PR #1439, opened 2024-01-31
fix[cartesian]: Add missing cstdint header in gtcpp codegen

tehrengruber:
Fixes some compilation errors I got on my machine with gcc `13.2.1`
```
error: ‘int64_t’ in namespace ‘std’ does not name a type
```


havogt (approved):
lgtm

PR #1442, opened 2024-02-01
bug[next]: allow fields of different sizes in tuple in itir embedded

havogt:
Undo an unintended change in #1202 to re-enable an icon4py pattern.

Longer term, probably, only transposable tuples of fields make sense, e.g. by intersecting.


egparedes (approved):
ILGTM

PR #1443, opened 2024-02-01
fix[next][dace]: Fix for neighbor reduction with skip values

edopao:
This PR provides a bugfix for the case of neighbor reductions with lambda function as reduction operation and connectivity table containing skip values. This specific case is verified by `test_tuple_with_local_field_in_reduction_shifted` and the corresponding field operator ITIR looks like:
```
  out ← (
      λ(e, v) → 
        reduce(λ(_fuse_maps_1, _fuse_maps_3, _fuse_maps_4) → _fuse_maps_1 + (_fuse_maps_3 + _fuse_maps_4), 0)(
          neighbors(V2Eₒ, ⟪E2Vₒ, 0ₒ⟫(e)), make_const_list(·⟪E2Vₒ, 0ₒ⟫(v))
        )
    )(__sym_1, __sym_2) @ u⟨ Edge: [0, __out_size_0) ⟩;
```
The lambda function should only accumulate the results for the valid  neighbors. On the contrary, the baseline implementation was using the reduction identity value for the missing neighbors, resulting in invalid result.

The fix consists of producing an array of boolean flags to determine if the neighbor value is valid or not. If not valid, the call to the lambda function is by-passed.

image

Code documentation is missing. The intention is to add it after rebase of #1431 (support lowering of lift expressions in neighbor reductions) which will conflict with these changes.


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -990,13 +1080,46 @@ def _visit_reduce(self, node: itir.FunCall):
...
             )
             lambda_context, inner_inputs, inner_outputs = self.visit(
-                lambda_node, args=args, use_neighbor_tables=False
+                lambda_node, args=input_args, use_neighbor_tables=False
             )
 
+            if input_valid:
+                # check neighbor validity on if/else inter-state edge
+                start_state = lambda_context.body.add_state(is_start_block=True)
+                skip_neighbor_state = lambda_context.body.add_state()
```
philip-paul-mueller, line 1089: You should give them a name.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -419,6 +466,42 @@ def builtin_cast(
     )
 
 
+def builtin_make_const_list(
+    transformer: "PythonTaskletCodegen", node: itir.Expr, node_args: list[itir.Expr]
+) -> list[ValueExpr]:
+    di = dace_debuginfo(node, transformer.context.body.debuginfo)
+    args = [transformer.visit(arg)[0] for arg in node_args]
+    assert all([isinstance(x, (SymbolExpr, ValueExpr)) for x in args])
```
philip-paul-mueller, line 474: I recently learned that you also could write
```python
all(isinstance(x, (SymbolExpr, ValueExpr)) for x in args)
```
which is actually faster, since the list has not to be created.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -184,49 +184,49 @@ def __init__(
...
     me, mx = state.add_map(
-        f"{offset_dim}_neighbors_map",
-        ndrange={neighbor_index: f"0:{offset_provider.max_neighbors}"},
+        f"{offset_dim}_neighbor_map",
+        ndrange={neighbor_map_index: f"0:{offset_provider.max_neighbors}"},
         debuginfo=di,
     )
     table_name = connectivity_identifier(offset_dim)
-    table_subset = (f"0:{sdfg.arrays[table_name].shape[0]}", neighbor_index)
+    table_subset = (f"0:{sdfg.arrays[table_name].shape[0]}", neighbor_map_index)
```
philip-paul-mueller, line 229: I do not get why this is `.shape[0]` since this are all cells and not the neighbors.

philip-paul-mueller (commented):
There are a few small things but after that we should be able to merge it.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: Nice!

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: Fixed.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -990,13 +1080,46 @@ def _visit_reduce(self, node: itir.FunCall):
...
+                    dace.Memlet(data=arg.value.data, subset="0")
+                    if arg.value.desc(self.context.body).shape == (1,)
+                    else create_memlet_at(arg.value.data, nreduce_index)
+                )
+                for (param, _), arg in zip(inner_inputs, input_args)
+            } | (
+                {"is_valid": create_memlet_at(input_valid[0].value.data, nreduce_index)}
+                if input_valid
+                else {}
+            )
```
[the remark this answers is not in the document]
edopao: Yes, simple and clear.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -1013,7 +1136,8 @@ def _visit_reduce(self, node: itir.FunCall):
                 inputs=input_mapping,
                 outputs=output_mapping,
                 symbol_mapping=symbol_mapping,
-                input_nodes={arg.value.data: arg.value for arg in args},
+                input_nodes={arg.value.data: arg.value for arg in input_args}
+                | ({input_valid[0].value.data: input_valid[0].value} if input_valid else {}),
```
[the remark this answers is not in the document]
edopao: That was the formatting tool.Anyway, I have changed to different code.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
edopao: We let the memlet cut the slice for `neighbor_map_index` and then `shift_tasklet` will do access based on cell index. This could be an issue for DaCe, I see now. The tasklet should receive the array full shape, so it can treat the array as a C pointer and access it with all explicit indexes.

philip-paul-mueller (approved):

PR #1448, opened 2024-02-08
fix[next]: Detection of ROCM vs CUDA gpu runtime

edopao:
Call to `cp.cuda.get_hipcc_path()` seems not to work on Clariden to detect the gpu runtime used by CUPy. It returns ROCm on an Nvidia A100 node. Therefore, the allocator code was updated to use `cp.cuda.runtime.is_hip`.


edopao:
cscs-ci run default

havogt (approved):

PR #1451, opened 2024-02-12
feat[cartesian]: read-only data dims direct access & Fields

FlorianDeconinck:
## Description

We introduce a `ROField` which is a data dimensions only Field. Access is provided via `.at[]` which also becomes a way to access data dimensions on regular Field.

Technically, we re-use the data dims concept, empty out the cartesian axes and make sure we have protection in place against write to the `ROField`. Definition follows the already established data dimensions format for now: `[(type, (axes...))]`

## Requirements


FlorianDeconinck:
Technically the `ROField` is exactly that: a no spatial dim, only data dim `Field`. I made a different `FieldDescriptor` to not have spaghetti code and be more verbose user side with those kind of field which are not on the grid

FlorianDeconinck:
E.g.
```
class _ReadOnlyFieldDescriptorMaker(_FieldDescriptorMaker):
    def __getitem__(self, field_spec):
        if not isinstance(field_spec, collections.abc.Collection) and not len(field_spec) == 2:
            raise ValueError("ROField is defined by a tuple (type, [axes_size..])")

        dtype, data_dims = field_spec

        return _FieldDescriptor(dtype, [], data_dims)
```

FlorianDeconinck:
NB: One side effect of the way I coded up `.at` is that you can always access data dimension that way for read operation, including on a normal Field with spatial dimensions (you get centered spatial dim)

FlorianDeconinck:
Yes, I did think about that. And technically the read-only Field is enforced by the fact you can only do `.at` for access which is then blocked from `write` at the `gtscript` level. So the Field itself is not RO, which is confusing for devs.

At first I wrote it as a "DataField".

Let's exchange tomorrow at community meeting

FlorianDeconinck:
I renamed `ROField` to `OffgridField`.

I looked into using `Field` as-is, but Field by default have a `IJK` spatial axes which is used to be able to write `Field[np.float64]` and have it default to a 3d field.

I didn't come up with an alternate way to write the definition that was clean enough. So out of abundance of caution around the feature, I opted again for the `OffgridField` which forces no spatial dimensions

src/gt4py/cartesian/gtscript.py
```diff
@@ -694,10 +694,23 @@ def __getitem__(self, field_spec):
...
+        dtype, data_dims = field_spec
+
+        return _FieldDescriptor(dtype, [], data_dims)
+
+
 # GTScript builtins: variable annotations
 Field = _FieldDescriptorMaker()
 """Field descriptor."""
 
+OffgridField = _ReadOnlyFieldDescriptorMaker()
```
havogt, line 711: I personally like ConstantField (constant with respect to the grid), but I think not everyone likes it for confusion with non-mutable. Here it both meanings coincide.

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -692,6 +692,18 @@ class ParsingContext(enum.Enum):
     COMPUTATION = 2
 
 
+def _is_absolute_indexing_name(name: str):
```
havogt, line 695: maybe
```suggestion
def _is_datadim_indexing_name(name: str):
```
because otherwise it might be understood as absolute indexing in i,j,k

havogt (commented):
Looks good. I'll discuss with @egparedes when he is back from vacation.

src/gt4py/cartesian/frontend/gtscript_frontend.py
FlorianDeconinck: Renamed the code to follow the `datadims` to be cleaner. Also renamed PR to showcase the "data dims direct access" which is a big part of the feature

egparedes:
> The generic `Array` or `NdArray` could be confusing for users as a generic Array that transfers to any other framework (same for ndarray). I reckon we'd rather limit that concept in scope. What about `GlobalMemory` ?

Yes, I see your point and it makes sense. What about `GlobalParameter` or `GlobalTable`?

FlorianDeconinck:
`GlobalTable`, I like that. It conveys both the array-ness and a form of read-only.

egparedes:
From my side at least, you can go ahead and change the names of the two features, and I think the PR would already be ready to be merged.

FlorianDeconinck:
Done. `GlobalTable` and `.A` for access

FlorianDeconinck:
Ready to be merged.

egparedes (approved):
It looks good to me.

egparedes (changes requested):
Please update the PR to the new formatting tools (it should be very easy)

havogt:
cscs-ci run

egparedes (approved):
ILGTM

PR #1457, opened 2024-02-16
fix[next][dace]: Bugfix for nested neighbor reduction

edopao:
In case of nested neighbor reduction with lift expression on inner node, the DaCe backend should generate a conditional state transition to field access, based on the value of neighbor index provided by the outer connectivity table.

The test case `test_nested_reduction` was modified by swapping the inner and outer connectivity tables, in order to have V2E on the outer neighbor expression. This because V2E is setup with skip values in `skip_value_mesh` tests.

Additional change. The previous selection of valid neighbors implemented as conditional inter-state edge is replaced by a select tasklet, which makes the SDFG easier to read.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
@@ -458,7 +458,7 @@ def builtin_neighbors(
         neighbor_valid_node = state.add_access(neighbor_valid_var, debuginfo=di)
 
         neighbor_valid_tasklet = state.add_tasklet(
-            "check_valid_neighbor",
+            f"check_valid_neighbor_{offset_dim}",
             {"__idx"},
             {"__valid"},
             f"__valid = True if __idx != {neighbor_skip_value} else False",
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -1246,41 +1271,52 @@ def _visit_reduce(self, node: itir.FunCall):
...
+                    {"_out"},
+                    f"_out = _inp if _valid else {reduce_identity}",
+                )
+                lambda_context.state.add_edge(
+                    output_edge.src,
                     None,
-                    dace.Memlet(data=inner_outputs[0].value.data, subset="0"),
+                    select_tasklet,
+                    "_inp",
+                    dace.Memlet(data=output_edge.src.data, subset="0"),
```
philip-paul-mueller, line 1303: I do not understand why you have here `subset="0"`, doesn't this mean that you always read the first element of the array or I am wrong?

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_tasklet.py
```diff
@@ -298,6 +299,30 @@ def _visit_lift_in_neighbors_reduction(
...
+            skip_neighbor_state,
+            dace.InterstateEdge(condition=f"{lifted_index_connectors[0]} == {neighbor_skip_value}"),
+        )
+        lift_context.body.add_edge(
+            start_state,
+            lift_context.state,
+            dace.InterstateEdge(condition=f"{lifted_index_connectors[0]} != {neighbor_skip_value}"),
+        )
+
     return [ValueExpr(neighbor_value_node, inner_outputs[0].dtype)]
```
philip-paul-mueller, line 326: I am just wondering, how your changes from above propagate outside.

philip-paul-mueller (approved):

PR #1460, opened 2024-02-19
feat[next][dace]: Remove symbols for connectivity arrays

edopao:
Remove stride and offset symbols for connectivity arrays.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
@@ -140,16 +140,23 @@ def get_shape_args(
     return shape_args
 
 
-def get_offset_args(
-    sdfg: dace.SDFG,
-    args: Sequence[Any],
-) -> Mapping[str, int]:
+def get_offset_args(sdfg: dace.SDFG, args: Sequence[Any]) -> Mapping[str, int]:
     sdfg_arrays: Mapping[str, dace.data.Array] = sdfg.arrays
     sdfg_params: Sequence[str] = sdfg.arg_names
+    fied_args = {param: arg for param, arg in zip(sdfg_params, args) if common.is_field(arg)}
+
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -235,15 +242,13 @@ def get_sdfg_args(sdfg: dace.SDFG, *args, check_args: bool = False, **kwargs) ->
...
     dace_strides = get_stride_args(sdfg.arrays, dace_field_args)
-    dace_conn_strides = get_stride_args(sdfg.arrays, dace_conn_args)
     dace_offsets = get_offset_args(sdfg, args)
     all_args = {
         **dace_args,
         **dace_conn_args,
         **dace_shapes,
         **dace_conn_shapes,
         **dace_strides,
-        **dace_conn_strides,
```
philip-paul-mueller, line 246: While the removal of the offsets of the connectivities sounds reasonable, I do not get why you also remove the strides here.
I mean bellow you make the change in `add_storage()` to no longer generate the offsets, but I do not see a similar change for the strides (which I thing we should keep).

edopao: Right, it was half left in my head. Ok, let's keep the strides for the connectivity arrays.

philip-paul-mueller: What I find a bit strange is, how the hell did it run successfully?

edopao: Probably because we hard-code stride0 to the local size `max_neighbors`, and we do not use stride1 for indexing because the connectivity table is always 2D.

philip-paul-mueller (approved):

edopao:
cscs-ci run default

PR #1462, opened 2024-02-19
feat[cartesian]: Setting extra compile args and DaCe block size via env variables

stubbiali:
## Description

This PR adds the ability to read extra compile arguments and the DaCe block size from the environment. When the `dace:gpu` backend is used, the block size is also included in the stencil fingerprint. These features are quite useful for the PMAP model.

## Requirements


src/gt4py/cartesian/backend/dace_backend.py
```diff
@@ -546,6 +546,12 @@ def apply(cls, stencil_ir: gtir.Stencil, builder: "StencilBuilder", sdfg: dace.S
         with dace.config.temporary_config():
```
FlorianDeconinck, line 546:
```suggestion
        # To prevent conflict with 3rd party usage of DaCe config always make sure that any changes
        #  be under the temporary_config manager
        with dace.config.temporary_config():
```

src/gt4py/cartesian/config.py
```diff
@@ -83,3 +84,5 @@
 code_settings: Dict[str, Any] = {"root_package_name": "_GT_"}
 
 os.environ.setdefault("DACE_CONFIG", os.path.join(os.path.abspath("."), ".dace.conf"))
+
+DACE_DEFAULT_BLOCK_SIZE: str = os.environ.get("DACE_DEFAULT_BLOCK_SIZE", "32,1,1")
```
FlorianDeconinck, line 88: That sounds a bit low. Unless we care about old hardware we can easily bump it to `64,8,1`

FlorianDeconinck (approved):
LGTM.

Couple of comments to verbose the `temp` usage which is critical for external use and maybe a more aggressive default.

src/gt4py/cartesian/config.py
FlorianDeconinck: Indeed it is, but since we have the liberty to reset a default, we can update it to something more likely. A part of me would like the system to crash if you are using an antic GPU ;)

stubbiali: Both parts make sense ;)

havogt:
cscs-ci run

stubbiali:
@FlorianDeconinck. I cannot merge this myself as I don't have write access, so could you please merge. Thanks a lot! 🙏🏻

PR #1463, opened 2024-02-19
fix[next][dace]: Fix for GPU illegal memory access

edopao:
Add check for all iterator indices when performing shift inside lift expression in neighbor reduction. Baseline was only checking the index in the neighbor dimension, which caused an illegal memory access in GPU execution.

For example, provided this ITIR:
`neighbors(C2Eₒ, (↑(λ(__arg0) → cast_(·__arg0, float64)))(⟪Koffₒ, -1ₒ⟫(__stencil_arg2))))`

we now check validity of both neighbor index (C2E) and cartesian shift (Koff). The unrolled ITIR (which was working) is actually doing the same.


edopao:
This change is not required. The crash is caused by a bug in the stencil program. The bug was caused by the assumption of cyclic access for out-of-boundary array access. More specifically, a vertical offset `Koff[-1]` was applied without limiting the closure domain to start from index 1. Setting `vertical_start=1` in the stencil program solves the problem

PR #1449, opened 2024-02-08
refactor[next]: Change foast lowering from iterator of tuple to tuple of iterator

tehrengruber:
Changes lowering from FOAST to ITIR to use tuples of iterators instead of iterators of tuples.
- Allows the collapse tuple pass to better optimize as what was previously lowered to `↑(λ(it) → it[0])(↑(λ() → {1, 2})` is now `({↑(λ() → 1)(), ↑(λ() → 2)()})[0]`
- Fixes #1413. In particular in combination with temporaries creating an iterator of tuples is problematic in case the respective backing fields are defined on different domains. This did not surface before as inlining all lifts everything always removed the invalid derefs.

PR #1473, opened 2024-02-28
feature[next]: Fix caching stability

tehrengruber:
#1449 introduced a hashing function for expressions that relied on Python's builtin `hash`. As `hash` is not stable across interpreter runs this broke caching in some cases. This PR fixes that.


egparedes (approved):
ILGTM

PR #1474, opened 2024-02-28
feature[next]: ITIR transformation & gtfn codegen caching

tehrengruber:
Significantly improves time to first computed value by caching the ITIR transformations and GTFN code generation.

This is not meant to be merged. Just an intermediate solution to make debugging more practical. In PMAP-G this improves the time for the first time-step for a larger test-case from ~10 minutes to ~40s.


tehrengruber:
Absorbed in #1690

PR #1477, opened 2024-02-29
feat[next][dace]: Add gt4py workflow for the DaCe backend

edopao:
Integrate the DaCe backend with the workflow API.
Still keeping the old backend interface for backwards-compatibility with icon4py. The old interface will be removed later.

### DIFF
--- a/src/gt4py/next/otf/languages.py
+++ b/src/gt4py/next/otf/languages.py
@@ -57,6 +57,11 @@ class Python(LanguageTag):
     ...
 
 
+class JSON(LanguageTag):
+    settings_class = LanguageSettings
+    ...
+
+
 class NanobindSrcL(LanguageTag): ...
 
 
--- a/src/gt4py/next/otf/languages.py
+++ b/src/gt4py/next/otf/languages.py
@@ -57,7 +57,7 @@ class Python(LanguageTag):
     ...
 
 
-class JSON(LanguageTag):
+class SDFG(LanguageTag):
     settings_class = LanguageSettings
     ...
 
--- a/src/gt4py/next/program_processors/runners/dace.py
+++ b/src/gt4py/next/program_processors/runners/dace.py
@@ -40,7 +40,9 @@ class DaCeCompilationStepFactory(factory.Factory):
 
 
 def _convert_args(
-    inp: stages.CompiledProgram, device: core_defs.DeviceType = core_defs.DeviceType.CPU
+    inp: stages.CompiledProgram,
+    device: core_defs.DeviceType = core_defs.DeviceType.CPU,
+    use_field_canonical_representation: bool = False,
 ) -> stages.CompiledProgram:
     sdfg_program = cast(CompiledSDFG, inp)
     on_gpu = True if device == core_defs.DeviceType.CUDA else False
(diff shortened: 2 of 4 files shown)


src/gt4py/next/program_processors/runners/dace.py
```diff
@@ -0,0 +1,128 @@
...
+            return inp(**sdfg_args)
+
+    return decorated_program
+
+
+def _no_bindings(inp: stages.ProgramSource) -> stages.CompilableSource:
+    return stages.CompilableSource(program_source=inp, binding_source=None)
+
+
+class DaCeWorkflowFactory(factory.Factory):
```
edopao, line 74: I should probably move this class to `dace_iterator/workflow.py`, it makes more sense.

src/gt4py/next/otf/languages.py
```diff
@@ -57,6 +57,11 @@ class Python(LanguageTag):
     ...
 
 
+class JSON(LanguageTag):
```
DropD, line 60: These type tags are about making it clear which steps can be mixed with which ones. Therefore it is important to describe what semantics are used. For your purposes "SDGF" describes the semantics, "JSON" only the format.

```suggestion
class SDFG_JSON(LanguageTag):
```

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -0,0 +1,141 @@
...
+class DaCeCompiler(
+    workflow.ChainableWorkflowMixin[
+        stages.CompilableSource[SourceLanguageType, LanguageSettingsType, languages.Python],
+        stages.CompiledProgram,
+    ],
+    workflow.ReplaceEnabledWorkflowMixin[
+        stages.CompilableSource[SourceLanguageType, LanguageSettingsType, languages.Python],
+        stages.CompiledProgram,
+    ],
+    step_types.CompilationStep[SourceLanguageType, LanguageSettingsType, languages.Python],
```
DropD, line 111: It would make sense to be more specific here and use `languages.SDFG_JSON` instead of `SourceLanguageType` and `languages.LanguageSettings` instead of `LanguageSettingsType`. Unless the `DaCeCompiler` can work with other languages (the one for gtfn can work with any generated language supported by CMake).

src/gt4py/next/program_processors/runners/dace.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class DaCeTranslationStepFactory(factory.Factory):
+    class Meta:
+        model = dace_workflow.DaCeTranslator
+
+
+class DaCeCompilationStepFactory(factory.Factory):
+    class Meta:
+        model = dace_workflow.DaCeCompiler
```
DropD, line 39: Is there a reason to prefer having the factories here, rather than next to their "model"-classes? Think about if you want to add a config option to one of the classes and then update the factory accordingly.

DropD (changes requested):
Overall almost ready, just some minor things.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
DropD: I didn't check if there were other related occurrences

src/gt4py/next/otf/languages.py
edopao: I see, it makes sense. I will use "SDFG" in this case.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
edopao: Thanks, I'll fix this.

DropD (approved):

PR #1483, opened 2024-03-06
bug[next]: Mark cupy tests as requires gpu

tehrengruber:
It noticed we execute some tests that use cupy in the cpu CSCS-CI. Tese tests are always executed when cupy is installed (which usually means we have a gpu, but not necessarily). This PR changes this such that they are marked with `requires_gpu` and as such can be disabled.


havogt (approved):

havogt:
I don't understand the failing tests in cartesian. The failing tests in next are probably because the tests where never run so far in ci. (They work locally on my machine.) Seems it's trying to build cub.

PR #1484, opened 2024-03-08
feature[next]: Extend cartesian offset syntax

SF-N:
## Description
Add support for replacing the syntax for Cartesian offsets `field(Ioff[1])(Joff[-1])` by `field(I+1,J-1)`.

Current status: finished for embedded execution

## TODOs
- [x] Move transformation to foast_to_itir and extend for other backends
- [x] merge main and fix conflicts


src/gt4py/next/common.py
```diff
@@ -901,6 +915,18 @@ def __gt_builtin_func__(cls, _: fbuiltins.BuiltInFunction) -> Never:  # type: ig
...
+        if isinstance(other, int):
+            return CartesianConnectivity(dimension=self.codomain, offset=self.offset + other)
+        else:
+            raise TypeError("'ConnectivityField' does not support this operation.")
+
+    def __sub__(self, other: Field | core_defs.IntegralScalar) -> Field:
+        if isinstance(other, int):
+            return CartesianConnectivity(dimension=self.codomain, offset=self.offset - other)
+        else:
+            raise TypeError("'ConnectivityField' does not support this operation.")
```
tehrengruber, line 928:
```suggestion
    def __add__(self, other: Field | core_defs.IntegralScalar) -> Field:
        assert isinstance(other, int)
        return CartesianConnectivity(dimension=self.codomain, offset=self.offset + other)

    def __sub__(self, other: Field | core_defs.IntegralScalar) -> Field:
        assert isinstance(other, int)
        return CartesianConnectivity(dimension=self.codomain, offset=self.offset - other)
```

Should be moved to CartesianConnectivity class. An even cleaner solution that would avoid issues when the attributes of CartesianConnectivity change would be to use [dataclasses.replace](https://docs.python.org/3/library/dataclasses.html):

```
dataclasses.replace(self, offset=self.offset+other)
```

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -620,6 +620,19 @@ def _deduce_binop_type(
         right: foast.Expr,
         **kwargs: Any,
     ) -> Optional[ts.TypeSpec]:
+        if (
```
tehrengruber, line 623:
```suggestion
        # e.g. `IDim+1`
        if (
```

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -620,6 +620,19 @@ def _deduce_binop_type(
...
+            isinstance(left.type, ts.DimensionType)
+            and isinstance(right.type, ts.ScalarType)
+            and type_info.is_integral(right.type)
+        ):
+            return ts.OffsetType(source=left.type.dim, target=(left.type.dim,))
```
tehrengruber, line 628:
```suggestion
            # this relies on the convention that for each dimension we have an offset provider 
            # with the same name as the dimension
            return ts.OffsetType(source=left.type.dim, target=(left.type.dim,))
```

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -620,6 +620,19 @@ def _deduce_binop_type(
...
+        if (
```
tehrengruber, line 629: Why is this cases need? Do we want something like `Ioff+1` which I assume is what this is covering?

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -295,23 +295,44 @@ def visit_Compare(self, node: foast.Compare, **kwargs: Any) -> itir.FunCall:
...
-                assert isinstance(offset_dim, foast.Name)
-                shift_offset = im.shift(
-                    offset_dim.id, im.deref(self.visit(func_args.args[1], **kwargs))
-                )
-            case _:
-                raise FieldOperatorLoweringError("Unexpected shift arguments!")
-        return im.lift(im.lambda_("it")(im.deref(shift_offset("it"))))(
-            self.visit(node.func, **kwargs)
-        )
+        shift_offsets = []
```
tehrengruber, line 298: Note to myself: will look at this later again.

tehrengruber (commented):
Partial review.

src/gt4py/next/ffront/foast_passes/type_deduction.py
SF-N: This is for IDim+1+2, which we would like to support for embedded. I added the respective case to _visit_shift in foast_to_itir.py again as one otherwise gets the "Unexpected shift arguments!" error.

src/gt4py/next/ffront/decorator.py
```diff
@@ -215,6 +215,13 @@ def itir(self) -> itir.FencilDefinition:
         return past_to_itir.PastToItirFactory()(no_args_past).program
 
     def __call__(self, *args, offset_provider: dict[str, Dimension], **kwargs: Any) -> None:
+        params = self.past_stage.past_node.params
```
tehrengruber, line 218: This will break the caching of the gtfn backend. Let's fix this together. Since this is such a performance critical part and for readability I would suggest to outsource the loop into a cached property that computes all the artifically added offset provider and only do a single
```suggestion
        offset_provider = offset_provider | self._implicit_offset_provider
```
here.

src/gt4py/next/ffront/foast_passes/type_deduction.py
tehrengruber: Makes sense. I wondered how this actually works now in embedded since I don't see an `__add__` in `FieldOffset`.

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -295,23 +295,56 @@ def visit_Compare(self, node: foast.Compare, **kwargs: Any) -> itir.FunCall:
...
+        for i in range(len(node.args)):
```
tehrengruber, line 299:
```suggestion
        for arg in node.args:
```

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -295,23 +295,56 @@ def visit_Compare(self, node: foast.Compare, **kwargs: Any) -> itir.FunCall:
...
+                        left=foast.Name(id=dimension),
+                        right=foast.Constant(value=offset_index_left),
+                    ),
+                    right=foast.Constant(value=offset_index),
+                ):
+                    if node.args[i].op == dialect_ast_enums.BinaryOperator.SUB:  # type: ignore[attr-defined] # ensured by pattern
+                        offset_index *= -1
+                    shift_offsets.append(
+                        im.shift(f"{dimension}off", offset_index_left + offset_index)
+                    )
```
tehrengruber, line 327: This is for the `IDim+1+1` case right? Let's remove it for now since it is not generic anyway (i.e. doesn't work for `IDim+1+1+1`).

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -295,23 +295,56 @@ def visit_Compare(self, node: foast.Compare, **kwargs: Any) -> itir.FunCall:
...
+                case foast.Name(id=offset_name):
+                    assert len(node.args) == 1
+                    return im.lifted_neighbors(
+                        str(offset_name), self.visit(node.func, **kwargs)
+                    )  # Todo: fix return statement to take care of several args
```
tehrengruber, line 332:
```suggestion
                    # only a single unstructured shift is supported so returning here is fine even though we
                    # are in a loop.
                    assert len(node.args) == 1
                    return im.lifted_neighbors(
                        str(offset_name), self.visit(node.func, **kwargs)
                    )
```
Please also add an assert that checks that the offset is indeed unstructured.

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -295,23 +295,56 @@ def visit_Compare(self, node: foast.Compare, **kwargs: Any) -> itir.FunCall:
...
+                case foast.Call(func=foast.Name(id="as_offset")):
+                    func_args = node.args[i]
+                    offset_dim = func_args.args[0]  # type: ignore[attr-defined] # ensured by pattern
+                    assert isinstance(offset_dim, foast.Name)
```
tehrengruber, line 336:
```suggestion
                    assert isinstance(offset_dim, foast.Name)  # TODO(tehrengruber): fixme. use type instead.
```
This is actually not correct and breaks when one does `I = Dimension("IDim")`.

src/gt4py/next/ffront/foast_to_itir.py
```diff
@@ -295,23 +295,56 @@ def visit_Compare(self, node: foast.Compare, **kwargs: Any) -> itir.FunCall:
...
+                    shift_offsets.append(
+                        im.shift(offset_dim.id, im.deref(self.visit(func_args.args[1], **kwargs)))  # type: ignore[attr-defined]  # ensured by pattern
+                    )
+                case _:
+                    raise FieldOperatorLoweringError("Unexpected shift arguments!")
+
+        ret = im.lift(im.lambda_("it")(im.deref(shift_offsets[0]("it"))))(self.visit(node.func, **kwargs))
+        for i in range(len(shift_offsets) - 1):
+            ret = im.lift(im.lambda_("it")(im.deref(shift_offsets[i + 1]("it"))))(ret)
```
tehrengruber, line 345: This looks like a case for functools.reduce.

tehrengruber (changes requested):

src/gt4py/next/ffront/foast_passes/type_deduction.py
SF-N: The `__add__` in CartesianConnectivity takes care of that.

src/gt4py/next/common.py
```diff
@@ -901,6 +914,14 @@ def __gt_builtin_func__(cls, _: fbuiltins.BuiltInFunction) -> Never:  # type: ig
     def ndarray(self) -> Never:
         raise NotImplementedError()
 
+    def __add__(self, other: Field | core_defs.IntegralScalar) -> Field:
+        assert isinstance(other, int)
+        return dataclasses.replace(self, offset=self.offset + other)
```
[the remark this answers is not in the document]
SF-N: In _core/devinitions.py `numbers.Integral` ist used. Please let me know if that is also not correct.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -226,7 +226,14 @@ def remap(
...
         )
 
-    __call__ = remap  # type: ignore[assignment]
+    def __call__(
+        self,
+        index_field: common.ConnectivityField | fbuiltins.FieldOffset,
+        *args: common.ConnectivityField | fbuiltins.FieldOffset,
+    ) -> common.Field:
+        return functools.reduce(
+            lambda field, connectivity: field.remap(connectivity), [index_field, *args], self
```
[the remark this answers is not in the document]
SF-N: I would keep `connectivity` to avoid confusion.

src/gt4py/next/ffront/foast_to_itir.py
SF-N: Is `assert (len(node.args) == 1 and len(arg.type.target) > 1)` the correct assert?

src/gt4py/next/ffront/foast_to_itir.py
SF-N: As discussed, I removed all sopport for (IDim +1) + 1 again and raise an error in `type_deduction.py` instaed.

src/gt4py/next/ffront/foast_to_itir.py
SF-N: Yes, I used a `lambda` function there, but mypy wants a def `instead` which needs type annotations. Could you please check the annotations of `func`?

SF-N:
I tried to resolve the conflict with `next/program_processors/runners/gtfn.py` but pre-commit reverts it back. Might it need to be adapted in `main`?

tehrengruber (approved):

PR #1489, opened 2024-03-13
feat[next]: workshop material and docs field Dims type hint update 

nfarabullini:
edited type hints in `gt4py/docs/user/next` from e.g.:
```Field[[CellDim, KDim], dtype]```
to:
```Field[Dims[CellDim, KDim], dtype]```

Addition of all workshop material under the same folder

### CODE REVIEW

**Review by egparedes — CHANGES_REQUESTED** [review]
I think the workshop materials should go in a folder on its own to avoid confusion. Something like:
```
 docs/user/next
    QuickstartGuide.md
    Workshop/
        exercises/
        images/
            logos/
        slides/
```
Also we can move the `logos` folder inside `images` `jupyter_intro` into `slides`, to keep the other folders cleaner

egparedes (commented):
All the changes look good to me. The only missing part is to add automatic checking of all the new notebooks (and the quickstart guide if it's not already tested) to make sure they don't go out of sync with the code.

egparedes (approved):
It looks good to me, but please extend the current PR title and description to include the description of all the added changes in the commit.

PR #1492, opened 2024-03-14
style[next]: stricter typing in ffront

havogt:
Requires full type annotations in next.ffront.


src/gt4py/next/ffront/decorator.py
```diff
@@ -72,25 +72,6 @@
 DEFAULT_BACKEND: Callable = None
 
 
-def _field_constituents_shape_and_dims(
```
havogt, line 75: was moved, but not deleted

PR #1494, opened 2024-03-15
style[next]: more strict typing

havogt:


pyproject.toml
```diff
@@ -203,13 +203,16 @@ warn_unused_ignores = false
 
 [[tool.mypy.overrides]]
 # # TODO: this should be changed to true after a transition period
```
egparedes, line 205:
```suggestion
```
Another option could be to set this option true in the global settings and the override the subpackages which still do not work (`cartesian` and a couple of modules in `next`, I guess)

src/gt4py/eve/codegen.py
```diff
@@ -676,6 +676,23 @@ def generic_dump(cls, node: RootNode, **kwargs: Any) -> str:
         """
         return str(node)
 
+    @overload
```
egparedes, line 679: Nice addition!

src/gt4py/next/embedded/common.py
```diff
@@ -148,7 +148,9 @@ def restrict_to_intersection(
     )
 
 
-def iterate_domain(domain: common.Domain):
+def iterate_domain(
+    domain: common.Domain,
+) -> Generator[tuple[tuple[common.Dimension, int], ...], None, None]:
```
egparedes, line 153: Just a question: would it be enough with `Iterator` instead of `Generator`?  In any case, use the actual ABCs from `collections.abc` instead of the deprecated aliases in `typing`

src/gt4py/next/embedded/operators.py
```diff
@@ -118,9 +118,10 @@ def field_operator_call(op: EmbeddedOperator, args: Any, kwargs: Any):
             res = ctx.run(op, *args, **kwargs)
             _tuple_assign_field(
                 out,
-                res,
+                res,  # type: ignore[arg-type] # maybe can't inferred properly because decorator.py is not properly typed yet
```
egparedes, line 121:
```suggestion
                res,  # type: ignore[arg-type] # maybe can't be inferred properly because decorator.py is not properly typed yet
```

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -891,6 +890,7 @@ def _visit_astype(self, node: foast.Call, **kwargs: Any) -> foast.Call:
                 primitive_type, getattr(ts.ScalarKind, new_type.id.upper())
             ),
         )
+        assert isinstance(return_type, (ts.TupleType, ts.ScalarType, ts.FieldType))
```
egparedes, line 893: Just a question, why is it needed to replace the static type check by a runtime check here?

src/gt4py/next/ffront/type_info.py
```diff
@@ -197,7 +197,9 @@ def _as_field(dtype: ts.TypeSpec, path: tuple[int, ...]) -> ts.FieldType:
             # TODO: we want some generic field type here, but our type system does not support it yet.
             return ts.FieldType(dims=[common.Dimension("...")], dtype=dtype)
 
-    return type_info.apply_to_primitive_constituents(param, _as_field, with_path_arg=True)
+    res = type_info.apply_to_primitive_constituents(param, _as_field, with_path_arg=True)
+    assert isinstance(res, (ts.FieldType, ts.TupleType))
```
egparedes, line 201: Why is the runtime check needed now? There are no changes in this function type hints..

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -191,6 +191,16 @@ def _literal_as_integral_constant(node: itir.Literal) -> IntegralConstant:
     return IntegralConstant(value=int(node.value))
 
 
+def _is_scan(node: itir.Node) -> bool:
```
egparedes, line 194: Why not a `Typeguard[itir.Node]`?

src/gt4py/next/type_inference.py
```diff
@@ -346,7 +346,9 @@ def unify(
     # Deduplicate type nodes, this can speed up later things a bit
     memo = dict[Type, Type]()
     dtypes = [_Dedup().visit(dtype, memo=memo) for dtype in dtypes]
-    constraints = {_Dedup().visit(c, memo=memo) for c in constraints}
+    constraints = {
+        _Dedup().visit(c, memo=memo) for c in constraints
+    }  # TODO this seems broken as _Dedup only works on `Type` not `Tuple[Type,Type]`. but also contraints is already a set?
```
egparedes, line 351: Is the TODO up to date?

src/gt4py/next/embedded/context.py
```diff
@@ -16,7 +16,7 @@
 
 import contextlib
 import contextvars as cvars
-from typing import Any
+from typing import Any, Generator
```
egparedes, line 19: Use Generator from collections.abc

src/gt4py/next/errors/excepthook.py
```diff
@@ -21,7 +21,8 @@
 """
 
 import sys
-from typing import Callable
+import types
+from typing import Callable, Optional
```
egparedes, line 25: We should use `collections.abc.Callable` since Python 3.9 but maybe we can skip it for now until we fully deprecate 3.8 and 3.9 support...

src/gt4py/next/type_system/type_info.py
```diff
@@ -15,7 +15,7 @@
 import functools
 import types
 import typing
-from typing import Any, Callable, Iterator, Type, TypeGuard, cast
+from typing import Any, Callable, Generator, Iterator, Protocol, Type, TypeGuard, cast
```
egparedes, line 18: use the proper (Callable), Generator and ITerator classes from collections.abc

egparedes (changes requested):
I have some comments to improve the typing fixes.

pyproject.toml
havogt: this small comment cost me 3h, because I realized that so far I didn't enable `disallow_untyped_defs`...

src/gt4py/next/embedded/common.py
havogt: changed to Iterator

src/gt4py/next/ffront/type_info.py
havogt: `apply_to_primitive_constituents` was not typed, but the return is not precise enough for how it's used here

src/gt4py/next/ffront/foast_passes/type_deduction.py
havogt: `apply_to_primitive_constituents` was not typed, but the return is not precise enough for how it's used here. Didn't check if it should be loosened here.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
havogt: upgraded to TypeGuard and which allowed to remove an assert

src/gt4py/next/ffront/lowering_utils.py
```diff
@@ -79,7 +79,7 @@ def to_iterator_of_tuples(expr: itir.Expr | str, arg_type: ts.TypeSpec) -> itir.
         for type_ in type_constituents
     )
 
-    def fun(_, path):
+    def fun(_: Any, path: tuple[int, ...]) -> itir.FunCall:
```
egparedes, line 82: Just a question: are we sure `path` is a tuple of ints? Shouldn't also be allowed to contain `str`?

src/gt4py/next/iterator/embedded.py
```diff
@@ -1088,7 +1090,7 @@ def remap(self, index_field: common.ConnectivityField | fbuiltins.FieldOffset) -
         # TODO can be implemented by constructing and ndarray (but do we know of which kind?)
         raise NotImplementedError()
 
-    def restrict(self, item: common.AnyIndexSpec) -> common.Field:
+    def restrict(self, item: common.AnyIndexSpec) -> xtyping.Self:
```
egparedes, line 1093:
```suggestion
    def restrict(self, item: common.AnyIndexSpec) -> Self:
```
and then add the proper import header (ideally, replace the whole `from typing import ...` block by a `from gt4py.eve.xtyping import ...`

src/gt4py/next/iterator/embedded.py
```diff
@@ -1207,7 +1209,7 @@ def remap(self, index_field: common.ConnectivityField | fbuiltins.FieldOffset) -
...
```
egparedes, line 1212: Same comment as above

egparedes (commented):
Looks good. Final round of questions before approving.

src/gt4py/next/embedded/operators.py
```diff
@@ -46,7 +46,7 @@ def __call__(  # type: ignore[override]
         self,
         *args: common.Field | core_defs.Scalar,
         **kwargs: common.Field | core_defs.Scalar,  # type: ignore[override]
-    ) -> common.Field:
+    ) -> common.Field | tuple[common.Field | tuple, ...]:
```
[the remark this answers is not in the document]
havogt: `init` was missing the correct typevar and here it gets a bit more complicated, see solution.

src/gt4py/next/ffront/lowering_utils.py
havogt: my understanding is that it is only `int`s

src/gt4py/next/iterator/embedded.py
havogt: done

egparedes (approved):
Some of the typing ignores in return types could also be solved using `cast()` but it's not a big deal. ILGTM.

PR #1502, opened 2024-03-20
feat[next][dace]: Filter unused connectivity tables

edopao:
During SDFG construction, only include the connectivity tables that are used by the stencil program. By convention, DaCe will not remove unnecessary data dependencies from program arguments.

Before this PR:
Screenshot 2024-03-20 at 08 50 30

with this PR:
Screenshot 2024-03-20 at 08 50 49

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
@@ -32,7 +32,7 @@ from gt4py.next.otf.compilation import cache as compilation_cache
 from gt4py.next.type_system import type_specifications as ts, type_translation
 
 from .itir_to_sdfg import ItirToSDFG
-from .utility import connectivity_identifier, filter_neighbor_tables, get_sorted_dims
+from .utility import connectivity_identifier, get_sorted_dims
 
 
 try:
@@ -129,12 +129,13 @@ def _ensure_is_on_device(
 
 
 def get_connectivity_args(
-    neighbor_tables: Mapping[str, common.NeighborTable],
+    offset_provider: dict[str, Any],
     device: dace.dtypes.DeviceType,
 ) -> dict[str, Any]:
     return {
-        connectivity_identifier(offset): _ensure_is_on_device(offset_provider.table, device)
-        for offset, offset_provider in neighbor_tables.items()
+        connectivity_identifier(offset): _ensure_is_on_device(table.table, device)
+        for offset, table in offset_provider.items()
+        if isinstance(table, common.NeighborTable)
     }
 
 
@@ -234,12 +235,13 @@ def get_sdfg_args(sdfg: dace.SDFG, *args, check_args: bool = False, **kwargs) ->
         "use_field_canonical_representation", _default_use_field_canonical_representation
     )
 
-    neighbor_tables = filter_neighbor_tables(offset_provider)
     device = dace.DeviceType.GPU if on_gpu else dace.DeviceType.CPU
 
     dace_args = get_args(sdfg, args, use_field_canonical_representation)
     dace_field_args = {n: v for n, v in dace_args.items() if not np.isscalar(v)}
-    dace_conn_args = get_connectivity_args(neighbor_tables, device)
+    dace_conn_args = get_connectivity_args(offset_provider, device)
+    # keep only connectivity tables that are used in the sdfg
+    dace_conn_args = {n: v for n, v in dace_conn_args.items() if n in sdfg.arrays}
     dace_shapes = get_shape_args(sdfg.arrays, dace_field_args)
     dace_conn_shapes = get_shape_args(sdfg.arrays, dace_conn_args)
     dace_strides = get_stride_args(sdfg.arrays, dace_field_args)
--- a/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
@@ -27,7 +27,7 @@ from gt4py.next.iterator import transforms as itir_transforms
 from gt4py.next.type_system import type_translation
 
 from .itir_to_sdfg import ItirToSDFG
-from .utility import connectivity_identifier, get_sorted_dims
+from .utility import connectivity_identifier, filter_neighbor_tables, get_sorted_dims
 
 
 try:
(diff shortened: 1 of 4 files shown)


havogt:
This is mainly a cosmetic change, right?

edopao:
It might have some impact, indeed. If we reduce the number of arguments, we reduce the overhead of sorting and mapping the program arguments to the SDFG binary interface. Besides, we could test this simplified interface in the fused SDFG prototype and see if it solves the problem of closure propagation (for some reason, the connectivity tables are not automatically parsed by the dace frontend: one theory is that the dace frontend is expecting all tables with the current SDFG, while the diffusion modules is passing only the tables that are actually used by the stencil).

edopao:
Moreover, a simpler SDFG seems to trigger a different optimization strategy. The resulting SDFG is different.

havogt:
Interesting, this should probably also be fixed in DaCe then?

edopao:
Adding @philip-paul-mueller and @kotsaloscv so they are aware of this discussion. My understanding is that the non-transient arrays allocated in the outermost SDFG are supposed to be program arguments. You can also see this in the other way: you define the program interface by allocating non-transient arrays in the outermost SDFG. Should DaCe then silently modify the program interface, in case some of the arguments are not used? This we can discuss with the DaCe team, but I see some risks for confusion there. What DaCe does already is to remove data dependencies between a nested SDFG and its parent.

philip-paul-mueller:
> Sorry for the wall of text.

What do you mean by 'unnecessary data dependencies from program arguments' exactly?
Do you mean that the unused connectivity tables are not removed or that the global memory is first copied in a transient (2 left most inputs; I can not read their names in the image).

If it is the first one, then DaCe will never touch global memory, so it will by design (or rather convention) not remove them.

If it is the second, then I have experienced the same issue in the past, when I looked at advenction stencil 13, i.e.
```python
def adv13(A, B):
	A -= B
	return A
```
I can not read the names of the names of the access nodes, but the following assumes that the inputs, that have these intermediate transient, are also used as output.
My current understanding, i.e. working hypothesis, for this behaviour is, that it is done to "ensure correctness", with respect to the [guarantees that the `__restrict__` keywords in `C++`](https://stackoverflow.com/questions/8290666/how-restrict-restrict-works-in-c-c) has to uphold.

In my above example it is trivial to see that there is not issue, since everything is done pointwise, but I am sure that determining this in general is hard, thus DaCe play it safe by adding some transient (if you want I can give you some `C` example that illustrates it).

Since we porting ICON stencils, we _knew_ from the get go that this will _never_ happen.
For that reason I wrote a [transformation](https://github.com/philip-paul-mueller/jax_to_dace/blob/f75b9a52d0022f2328386be4e0c1a3a1cf34242d/code/JaxprToSDFG/transformations/redundand_array_remover.py#L21) to get rid of it.

edopao:
We were discussing here the first case you described, that the unused connectivity tables are not removed.

philip-paul-mueller:
Okay then the explanation is: They are global hence they are not removed.

havogt:
I totally agree with you, @edopao. They should not be removed, because they are the interface. But unused parameters should not lead to different optimizations. That's the part that should be fixed, no?

edopao:
Maybe I used the word _optimization_ inappropriately. I should have used _transformations_.
And also, this PR is not only pruning the program arguments, but at each SDFG nesting-level it is removing unnecessary non-transient arrays. This second kind of pruning was already done by DaCe, but maybe at a different stage in the simplify pass.
There is no guarantee for commutativity of SDFG transformations: the pruning of unnecessary non-transient arrays on nested SDFGs at early stage could be the trigger for different transformation results. The differences are small, at least at SDFG representation level.

Also, there is a DaCe bug which shows up in icon4py stencils and that prevents this PR from being merged.

Without this PR:
Screenshot 2024-03-20 at 12 36 14

With this PR:
Screenshot 2024-03-20 at 12 38 04

philip-paul-mueller:
I thought about it a bit more and came to the conclusion that at least from DaCe's perspective they are "used", at least in a sense.
If you think of a nested SDFG as a glorified tasklet, then the connectivity tables are all used, as input to the "tasklet" in the outermost SDFG.
To know that they are unused DaCe would have to decent into the nested SDFG and look if the connectors are used on the inside.
And I am pretty sure that there is no DaCe transformation that does this.

@edopao could you post an example where the removal triggered a different optimization path?
Based on our discussion this morning I guess you are using auto optimize.
As an information for you, I noticed in the past that auto optimize is not necessaraly idempotent, i.e. applying it again again changed the SDFG.
In that regard it does not really surprises me that a different route is chosen.

edopao:
@philip-paul-mueller I forgot to answer you, the example of different transformation result is shown above without and with this PR. I agree with your comment.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -118,12 +118,13 @@ def _ensure_is_on_device(
...
-    neighbor_tables: Mapping[str, common.NeighborTable],
+    offset_provider: dict[str, Any],
     device: dace.dtypes.DeviceType,
 ) -> dict[str, Any]:
     return {
-        connectivity_identifier(offset): _ensure_is_on_device(offset_provider.table, device)
-        for offset, offset_provider in neighbor_tables.items()
+        connectivity_identifier(offset): _ensure_is_on_device(table.table, device)
+        for offset, table in offset_provider.items()
+        if isinstance(table, common.NeighborTable)
```
philip-paul-mueller, line 127: OPTIONAL:
I do not like the idea of filtering out some arguments.
I mean what could `offset_provider` by than _the_ offset provider.
Also the function now has some overlap with `filter_neighbor_tables()`.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -125,6 +125,7 @@ def _make_array_shape_and_strides(
     """
     dtype = dace.int32
     sorted_dims = get_sorted_dims(dims) if sort_dims else list(enumerate(dims))
+    neighbor_tables = {n: v for n, v in offset_provider.items() if isinstance(v, NeighborTable)}
```
philip-paul-mueller, line 128: OPTIONAL:
Why not using `filter_neighbor_tables()`.

src/gt4py/next/program_processors/runners/dace_iterator/utility.py
```diff
@@ -59,11 +60,14 @@ def as_scalar_type(typestr: str) -> ts.ScalarType:
     return ts.ScalarType(kind)
 
 
-def filter_neighbor_tables(offset_provider: dict[str, Any]) -> dict[str, NeighborTable]:
```
philip-paul-mueller, line 62: Before this function only filtered out all values that were not a negbour table.
But now it makes some AST walk walk (which is potentially costly) and then filter them out with some additional condition.
I strongly suggest to you that you change its name to something like `get_used_neigbour_tables()` or something like that to clearly indicate this behaviour.

philip-paul-mueller (approved):
I approve it, but please see my comment to to `filter_neighbor_table()`.

PR #1511, opened 2024-03-21
bug[next]: Fix gtfn code generation for negative literals

tehrengruber:
Add an extra space between the operators of BinOp such that `minus(1, -1)` does not get translated into `1--1`, but `1 - -1`.

While looking at that part of the code I was wondering whether there is another bug in that function namely that we don't case int32 literals to the respective C++ type.

I would add tests, but there seems to be no infrastructure for this. Should we leave it like that for now?

### DIFF
--- a/src/gt4py/next/program_processors/codegens/gtfn/codegen.py
+++ b/src/gt4py/next/program_processors/codegens/gtfn/codegen.py
@@ -98,15 +98,27 @@ class GTFNCodegen(codegen.TemplatedGenerator):
         return value
 
     def visit_Literal(self, node: gtfn_ir.Literal, **kwargs: Any) -> str:
+        # TODO(tehrengruber): isn't this wrong and int32 should be casted to an actual int32?
         match pytype_to_cpptype(node.type):
             case "float":
-                return self.asfloat(node.value) + "f"
+                result = self.asfloat(node.value) + "f"
             case "double":
-                return self.asfloat(node.value)
+                result = self.asfloat(node.value)
             case "bool":
-                return node.value.lower()
+                result = node.value.lower()
             case _:
-                return node.value
+                result = node.value
+        if node.type in ["bool", "int32", "int64", "float32", "float64"]:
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -98,15 +98,27 @@ def asfloat(value: str) -> str:
...
+                result = node.value
+        if node.type in ["bool", "int32", "int64", "float32", "float64"]:
+            # wrap into parenthesis such that `minus(1, -1)` does not get translated into `1--1`,
+            # but `(1)-(-1)`
+            result = f"({result})"
+        elif node.type == "axis_literal":
+            pass
+        else:
+            raise NotImplementedError(
+                f"Literal type '{node.type}' is not supported in the code " f"generator"
```
havogt, line 119: formatting error? `" f"generator`

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -98,15 +98,27 @@ def asfloat(value: str) -> str:
...
             case "double":
-                return self.asfloat(node.value)
+                result = self.asfloat(node.value)
             case "bool":
-                return node.value.lower()
+                result = node.value.lower()
             case _:
-                return node.value
+                result = node.value
+        if node.type in ["bool", "int32", "int64", "float32", "float64"]:
```
havogt, line 111: maybe we can just make sure that there is a space in the binary operations? `1 - -1` should be ok

`BinaryExpr = as_fmt("({lhs} {op} {rhs})")`

havogt (changes requested):
To keep code more readable, I'd go for adding a space in the BinaryExpr instead

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
tehrengruber: Much nicer. Thanks.

havogt (approved):

PR #1513, opened 2024-03-25
fix[next]: fix protocol metaclasses conflicts appearing in some combination of requirement versions

egparedes:
Fix to try to inherit from the right Protocol metaclass


egparedes:
Closed in favor of #1519

PR #1516, opened 2024-04-02
refactor[next]: workflowify step3

DropD:
### New:

- `ffront.stages.FieldOperatorDefinition`
  - all the data to start the toolchain from a field operator dsl definition

- `ffront.stages.FoastOperatorDefinition`
  - data after lowering from field operator dsl code

### Changed:

- `decorator.Program.__post_init__`
  - implementation moved to `past_passes.linters` workflow steps
  - linting stage added to program transforms

- `decorator.FieldOperator.from_function`
  - implementation moved to workflow step in `ffront.func_to_foast`

- `decorator.FieldOperator` data attributes
  - added: `definition_stage`
  - removed:
    - `.foast_node`: replaced with `.foast_stage.foast_node`
    - `.definition`: replaced with `.definition_stage.definition`

## Requirements

### DIFF
--- a/src/gt4py/next/backend.py
+++ b/src/gt4py/next/backend.py
@@ -20,12 +20,30 @@ from typing import Any, Generic
 from gt4py._core import definitions as core_defs
 from gt4py.next import allocators as next_allocators
 from gt4py.next.ffront import func_to_past, past_process_args, past_to_itir, stages as ffront_stages
-from gt4py.next.otf import recipes
+from gt4py.next.ffront.past_passes import linters as past_linters
+from gt4py.next.otf import recipes, workflow
 from gt4py.next.program_processors import processor_interface as ppi
+
+
+@dataclasses.dataclass(frozen=True)
+class ArgsInjector(workflow.Workflow):
+    args: tuple[Any, ...] = dataclasses.field(default_factory=tuple)
+    kwargs: dict[str, Any] = dataclasses.field(default_factory=dict)
+
+    def __call__(self, inp: ffront_stages.PastProgramDefinition) -> ffront_stages.PastClosure:
+        return ffront_stages.PastClosure(
+            past_node=inp.past_node,
+            closure_vars=inp.closure_vars,
+            grid_type=inp.grid_type,
+            args=self.args,
+            kwargs=self.kwargs,
+        )
 
 
 DEFAULT_TRANSFORMS = recipes.ProgramTransformWorkflow(
     func_to_past=func_to_past.OptionalFuncToPastFactory(cached=True),
+    past_lint=past_linters.LinterFactory(),
+    past_inject_args=ArgsInjector(),
     past_transform_args=past_process_args.past_process_args,
     past_to_itir=past_to_itir.PastToItirFactory(),
 )
@@ -40,7 +58,9 @@ class Backend(Generic[core_defs.DeviceTypeT]):
     def __call__(
         self, program: ffront_stages.ProgramDefinition, *args: tuple[Any], **kwargs: dict[str, Any]
     ) -> None:
-        transformer = self.transformer.replace(args=args, kwargs=kwargs)
+        transformer = self.transformer.replace(
+            past_inject_args=ArgsInjector(args=args, kwargs=kwargs)
+        )
         program_call = transformer(program)
         self.executor(program_call.program, *program_call.args, **program_call.kwargs)
 
--- a/src/gt4py/next/backend.py
+++ b/src/gt4py/next/backend.py
@@ -19,7 +19,14 @@ from typing import Any, Generic
 
 from gt4py._core import definitions as core_defs
 from gt4py.next import allocators as next_allocators
-from gt4py.next.ffront import func_to_past, past_process_args, past_to_itir, stages as ffront_stages
+from gt4py.next.ffront import (
+    foast_to_itir,
+    func_to_foast,
+    func_to_past,
+    past_process_args,
+    past_to_itir,
+    stages as ffront_stages,
+)
 from gt4py.next.ffront.past_passes import linters as past_linters
 from gt4py.next.otf import recipes, workflow
 from gt4py.next.program_processors import processor_interface as ppi
@@ -40,6 +47,14 @@ class ArgsInjector(workflow.Workflow):
         )
 
 
+DEFAULT_FIELDOP_TRANSFORMS = recipes.FieldopTransformWorkflow(
+    func_to_foast=func_to_foast.OptionalFuncToFoastFactory(cached=True),
+    foast_to_itir=workflow.CachedStep(
+        step=foast_to_itir.foast_to_itir, hash_function=ffront_stages.hash_foast_operator_definition
+    ),
+)
+
+
 DEFAULT_TRANSFORMS = recipes.ProgramTransformWorkflow(
     func_to_past=func_to_past.OptionalFuncToPastFactory(cached=True),
     past_lint=past_linters.LinterFactory(),
@@ -53,12 +68,12 @@ DEFAULT_TRANSFORMS = recipes.ProgramTransformWorkflow(
 class Backend(Generic[core_defs.DeviceTypeT]):
     executor: ppi.ProgramExecutor
     allocator: next_allocators.FieldBufferAllocatorProtocol[core_defs.DeviceTypeT]
-    transformer: recipes.ProgramTransformWorkflow = DEFAULT_TRANSFORMS
+    transforms_prog: recipes.ProgramTransformWorkflow = DEFAULT_TRANSFORMS
 
     def __call__(
         self, program: ffront_stages.ProgramDefinition, *args: tuple[Any], **kwargs: dict[str, Any]
     ) -> None:
-        transformer = self.transformer.replace(
+        transformer = self.transforms_prog.replace(
             past_inject_args=ArgsInjector(args=args, kwargs=kwargs)
         )
         program_call = transformer(program)
--- a/src/gt4py/next/ffront/decorator.py
+++ b/src/gt4py/next/ffront/decorator.py
@@ -48,15 +48,11 @@ from gt4py.next.ffront import (
     transform_utils,
     type_specifications as ts_ffront,
 )
-from gt4py.next.ffront.foast_passes.type_deduction import FieldOperatorTypeDeduction
-from gt4py.next.ffront.foast_to_itir import FieldOperatorLowering
-from gt4py.next.ffront.func_to_foast import FieldOperatorParser
 from gt4py.next.ffront.gtcallable import GTCallable
 from gt4py.next.ffront.past_passes.closure_var_type_deduction import (
     ClosureVarTypeDeduction as ProgramClosureVarTypeDeduction,
 )
 from gt4py.next.ffront.past_passes.type_deduction import ProgramTypeDeduction
-from gt4py.next.ffront.source_utils import SourceDefinition, get_closure_vars_from_function
 from gt4py.next.iterator import ir as itir
 from gt4py.next.iterator.ir_utils.ir_makers import (
     literal_from_value,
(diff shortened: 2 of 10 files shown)


src/gt4py/next/otf/recipes.py
```diff
@@ -14,12 +14,39 @@
...
+    foast_inject_args: workflow.Workflow[
+        ffront_stages.FoastOperatorDefinition, ffront_stages.FoastClosure
+    ]
+    foast_to_past_closure: workflow.Workflow[ffront_stages.FoastClosure, ffront_stages.PastClosure]
+    past_transform_args: workflow.Workflow[ffront_stages.PastClosure, ffront_stages.PastClosure]
+    past_to_itir: workflow.Workflow[ffront_stages.PastClosure, stages.ProgramCall]
+    foast_to_itir: workflow.Workflow[ffront_stages.FoastOperatorDefinition, itir.Expr]
+
+    @property
+    def step_order(self) -> list[str]:
```
havogt, line 40: where is this used?

havogt (commented):
I feel you need to walk me through these changes, specifically the 2 different DEFAULT_X_TRANSFORMS and backend specifying 2 kinds of transforms.

src/gt4py/next/otf/recipes.py
DropD: In the superclass. It determines in which order the steps are run.

DropD: In the superclass. It determines in which order the steps are run.

src/gt4py/next/ffront/decorator.py
```diff
@@ -111,33 +100,14 @@ def definition(self):
 
     @functools.cached_property
     def past_stage(self):
-        if self.backend is not None and self.backend.transformer is not None:
-            return self.backend.transformer.func_to_past(self.definition_stage)
-        return next_backend.DEFAULT_TRANSFORMS.func_to_past(self.definition_stage)
+        if self.backend is not None and self.backend.transforms_prog is not None:
+            return self.backend.transforms_prog.func_to_past(self.definition_stage)
+        return next_backend.DEFAULT_PROG_TRANSFORMS.func_to_past(self.definition_stage)
```
havogt, line 105: I should have asked this question, in the previous refactoring: what's happening here? Can you document it? What's the case when the `DEFAULT_PROG_TRANSFORMS` should be used?

src/gt4py/next/ffront/decorator.py
```diff
@@ -111,33 +100,14 @@ def definition(self):
...
-            for symbol in self.past_stage.past_node.closure_vars
-            if symbol.id not in self.past_stage.closure_vars
-        ]
-        if undefined_symbols:
-            raise RuntimeError(
-                f"The following closure variables are undefined: {', '.join(undefined_symbols)}."
-            )
+        if self.backend is not None and self.backend.transforms_prog is not None:
+            self.backend.transforms_prog.past_lint(self.past_stage)
+        return next_backend.DEFAULT_PROG_TRANSFORMS.past_lint(self.past_stage)
```
havogt, line 110: Is the answer to the question above this line? I.e. when linting, but the backend doesn't have transforms_prog and therefore no linting, then we fall back to the linting via `DEFAULT_PROG_TRANSFORMS` and therefore need a `past_stage`?

src/gt4py/next/ffront/decorator.py
```diff
@@ -411,138 +382,83 @@ def from_function(
...
+                attributes=operator_attributes or {},
+            ),
             backend=backend,
-            grid_type=grid_type,
-            operator_attributes=operator_attributes,
         )
 
+    def __post_init__(self):
+        """This ensures that DSL linting occurs at decoration time."""
+        _ = self.foast_stage
```
havogt, line 397: maybe add a remark that we want to get rid of this hard-coded post_init linting in a next step

src/gt4py/next/ffront/decorator.py
```diff
@@ -554,36 +470,48 @@ def __call__(self, *args, **kwargs) -> None:
...
+                    self.definition_stage.definition, forward, init, axis
+                )
             else:
-                op = embedded_operators.EmbeddedOperator(self.definition)
+                op = embedded_operators.EmbeddedOperator(self.definition_stage.definition)
             return embedded_operators.field_operator_call(op, args, kwargs)
 
 
+@dataclasses.dataclass(frozen=True)
+class FieldOperatorFromFoast(FieldOperator):
```
havogt, line 508: Maybe add a hint how this could be used (e.g. refer to the test usage).

src/gt4py/next/ffront/stages.py
```diff
@@ -15,29 +15,134 @@
...
+
+
+OperatorNodeT = TypeVar("OperatorNodeT", bound=foast.LocatedNode)
+
+
+class ContentHashableMixin:
+    """
+    Allows deriving dataclasses to modify what goes into the content hash per-field.
+
+    Warning: Using this will modify how the class gets pickled. If unpickling is desired,
```
havogt, line 34: @egparedes Can you have a look at this part? I am not sure I understand its implication and it seems like a second pair of eyes make sense.

havogt (changes requested):
Let's talk with Enrique about the hashing part.

src/gt4py/next/backend.py
```diff
@@ -18,14 +18,119 @@
...
+    ),
+    past_transform_args=past_process_args.past_process_args,
+    past_to_itir=past_to_itir.PastToItirFactory(),
+    foast_to_itir=workflow.CachedStep(
+        step=foast_to_itir.foast_to_itir, hash_function=eve_utils.content_hash
+    ),
+)
+
+
+DEFAULT_PROG_TRANSFORMS = ProgramTransformWorkflow(
```
havogt, line 130: maybe makes sense to put the defaults in the `ProgramTransformWorkflow` definition as you hinted in the overview. Feels cleaner to me. also would have the advantage of being a maker implicitly.

src/gt4py/next/ffront/stages.py
egparedes: IIUC, the `Mixin` classes and the functions have been added here to support hashing the Stage classes with `eve.content_hash` but I don't really see why this is needed. I mean, if these classes need to be hashable and there is an unique way to implement the hash that makes sense in all use cases, then we could just write a custom `__hash__` instead of hacking around with `__setstate__` and `__getstate__` methods (which in addition of breaking the pickleablity of the instances, it'd only work for the `content_hash` function).

An even better and more general solution, in my opinion, would be to define custom hash functions as required for different purposes as `singledispatch` free functions, and then register specific implementations for the classes that need to be hashable in that specific way.

DropD: I think the singledispatch function is probably the solution.

src/gt4py/next/backend.py
DropD: Yes, it wouldn't have made sense where they were previously defined, but now it is the way to go!

src/gt4py/next/ffront/stages.py
DropD: My solution is up for review.

src/gt4py/next/ffront/decorator.py
DropD: Exactly. This is to be backwards compatible. Linting even if there is no backend. Also this ensures it doesn't break when the dummy `no_backend` is used in testing.

DropD: I am not sure where to document that (it occurs in so many places).

DropD: I put a comment on the first occurrence. Is that enough?

src/gt4py/next/backend.py
```diff
@@ -19,29 +19,161 @@
...
+        return [
+            "func_to_foast",
+            "foast_inject_args",
+            "foast_to_past_closure",
+            "past_transform_args",
+            "past_to_itir",
+        ]
+
+
+DEFAULT_FIELDOP_TRANSFORMS = FieldopTransformWorkflow()
```
havogt, line 100: optional: does it make sense to keep the instance here?

src/gt4py/next/backend.py
```diff
@@ -19,29 +19,161 @@
...
+    ] = dataclasses.field(default_factory=ProgArgsInjector)
+    past_transform_args: workflow.Workflow[ffront_stages.PastClosure, ffront_stages.PastClosure] = (
+        dataclasses.field(default=past_process_args.past_process_args)
+    )
+    past_to_itir: workflow.Workflow[ffront_stages.PastClosure, stages.ProgramCall] = (
+        dataclasses.field(default_factory=past_to_itir.PastToItirFactory)
+    )
+
+
+DEFAULT_PROG_TRANSFORMS = ProgramTransformWorkflow()
```
havogt, line 142: see above

havogt (approved):
Looks good to me. Let's wait for @egparedes feedback on the hashing.

src/gt4py/next/ffront/stages.py
```diff
@@ -42,3 +89,103 @@ class PastClosure:
...
+        update_cache_key(item, hasher)
+
+
+@update_cache_key.register
+def update_cache_key_foast(obj: foast.LocatedNode, hasher: Hasher_T) -> None:
+    update_cache_key(obj.location, hasher)
+    update_cache_key(str(obj), hasher)
+
+
+# not registered to avoid circular dependencies
```
egparedes, line 180: I feel like there should be a better way to avoid any circular dependency here but to be honest I don't fully understand the comments or the code. Could you elaborate a bit why this is needed?

egparedes (changes requested):
I only reviewed the changes in `stages.py` related to hashing. The code itself looks good, but I strongly suggest to rename the functions.

src/gt4py/next/ffront/stages.py
DropD: Never mind. The correct solution is to register the dispatch for `decorator.Program` and `decorator.FieldOperator` inside `decorator`.

The problem seemed to be that `decorator` imports `stages`, so specializing on `decorator` classes in `stages` would require a circular dependency. That would be a problem when not using `singledispatch`.

src/gt4py/next/ffront/stages.py
```diff
@@ -42,3 +89,84 @@ class PastClosure:
...
+
+
+@add_content_to_fingerprint.register(int)
+@add_content_to_fingerprint.register(bool)
+@add_content_to_fingerprint.register(float)
+def add_builtin_to_fingerprint(
+    obj: None,
+    hasher: xtyping.HashlibAlgorithm,
+) -> None:
+    hasher.update(str(obj).encode())
```
egparedes, line 137: A suggestion to reduce boilerplate:

```suggestion
@functools.singledispatch
def add_content_to_fingerprint(obj: Any, hasher: xtyping.HashlibAlgorithm) -> None:
    hasher.update(str(obj).encode())

_default_impl = add_content_to_fingerprint[object]
for t in (str, int, bool, float):
    add_content_to_fingerprint.register(t, _default_impl)

@add_content_to_fingerprint.register(FieldOperatorDefinition)
@add_content_to_fingerprint.register(FoastOperatorDefinition)
@add_content_to_fingerprint.register(FoastWithTypes)
@add_content_to_fingerprint.register(FoastClosure)
@add_content_to_fingerprint.register(ProgramDefinition)
@add_content_to_fingerprint.register(PastProgramDefinition)
@add_content_to_fingerprint.register(PastClosure)
def add_content_to_fingerprint_stages(obj: Any, hasher: xtyping.HashlibAlgorithm) -> None:
    add_content_to_fingerprint(obj.__class__, hasher)
    for field in dataclasses.fields(obj):
        add_content_to_fingerprint(getattr(obj, field.name), hasher)
```

DropD: It was necessary for str and int, but only to avoid a max recursion depth error. I am not exactly sure why this helps.

egparedes: For `str` is needed because otherwise it will call the implementation for `Iterable`, not the default implementation, since `str` is `Iterable` (and that's why the infinite recursion appears). For `int`, I don't see why it would be needed. Are you sure that just `str` is not enough? Maybe I'm misunderstanding something...

egparedes (approved):
I still wonder why it is needed to register a specific implementation for `int` in the `add_content_to_fingerprint` singledispatch function, but other than that, the fingerprinting functions look good to  me.

DropD:
> I still wonder why it is needed to register a specific implementation for `int` in the `add_content_to_fingerprint` singledispatch function, but other than that, the fingerprinting functions look good to me.

I also wonder why that should impact the recursion depth.

PR #1527, opened 2024-04-12
feat[next]: GT4Py programs as DaCe SDFGConvertibles

kotsaloscv:
With this fieature, a GT4Py.next Program can be placed inside a DaCe.program region, where its SDFG is automatically extracted and placed as a nested SDFG. This allows fusing multiple stencils under one SDFG -compilation unit-.

This feature is needed for implementing DaCe orchestration, e.g. in ICON4Py.


tests/next_tests/integration_tests/multi_feature_tests/ffront_tests/test_laplacian.py
```diff
@@ -86,3 +92,33 @@ def test_ffront_lap(cartesian_case):
...
+
+def test_sdfgConvertible_laplap(cartesian_case):
+    in_field = cases.allocate(cartesian_case, laplap_program, "in_field")()
+    out_field = cases.allocate(cartesian_case, laplap_program, "out_field")()
+
+    def jit(executor):
+        def decorator(func):
+            def wrapper(*args, **kwargs):
+                if executor == run_dace_cpu:
+                    return dace.program()(func)(*args, **kwargs)
```
edopao, line 105: I would prefer to assume that this test is only executed when the backend is run_dace_cpu, so you can decorate the program directly with `@dace.program`.

So you could put this line above the test function:
`@pytest.mark.skipif(cartesian_case.executor != run_dace_cpu)`

havogt:
Just a general comment, I am wondering if DaCe offers alternatively the option to register the conversion functions, instead of having to make them part of the object. (E.g. like jax pytrees work) If only the current option is available I am wondering if we should go with a monkey patching approach to remove the dace import from unrelated modules. But this is for @egparedes to answer. (And we could think of the other option as a contribution to productice DaCe.)

egparedes:
My first impression is that @havogt is right and the best option would be to enhance DaCe with support for registering SDFG convertible functions defined outside of the object. It would improve both the Gt4Py and DaCe codebases and would make this PR much cleaner. Another option, even simpler to implement in the DaCe side, would be to remove the requirement to inherit from `SDFGConvertible` (which is not even an ABC) and implement it as a typing `Protocol` using `runtime_checkable` or a custom typeguard for `isinstance()` checks.

If we want to merge this PR in the short term before waiting for changes in DaCe, I think it'd be better to add the required DaCe methods and bases conditionally to avoid adding DaCe as a non-optional requirement, as @edopao and @havogt suggested. For example:

```python
class Program:
    ....

if dace:
    class Program(Program, SDFGConvertible):
        def __sdfg__(self, *args, **kwargs) -> Optional[Any]:
             ....

```

src/gt4py/next/ffront/decorator.py
```diff
@@ -65,9 +67,21 @@
...
 
+try:
+    import dace
+    from dace.frontend.python.common import SDFGConvertible
+
+    from gt4py.next.program_processors.runners.dace import run_dace_gpu
+except ImportError:
+    dace: Optional[ModuleType] = None  # type:ignore[no-redef]
+
+    class SDFGConvertible: ...
```
edopao, line 82: This line (`class SDFGConvertible: ...`) is probably not needed anymore after the latest change.

kotsaloscv: good catch

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -45,6 +45,11 @@
 except ImportError:
     jnp: Optional[ModuleType] = None  # type:ignore[no-redef]
 
+try:
+    import dace
+except ImportError:
+    dace: Optional[ModuleType] = None  # type:ignore[no-redef]
```
egparedes, line 51:
```suggestion
    dace: Optional[ModuleType] = None  # type: ignore[no-redef]
```

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -327,6 +332,30 @@ def _slice(
         return new_domain, slice_
 
 
+if dace:
+    # Extension of NdArrayField : Support for SDFGConvertible GT4Py Programs
+    @dataclasses.dataclass(frozen=True)
+    class NdArrayField(  # type: ignore # redefinition
```
egparedes, line 338:
```suggestion
    class NdArrayField(  # type: ignore[no-redef]
```

src/gt4py/next/ffront/decorator.py
```diff
@@ -234,6 +246,62 @@ def __call__(self, *args, offset_provider: dict[str, Dimension], **kwargs: Any)
         )
 
 
+if dace:
+    # Extension of GT4Py Program : Implement SDFGConvertible interface
+    @dataclasses.dataclass(frozen=True)
+    class Program(Program, SDFGConvertible):  # type: ignore # redefinition
```
egparedes, line 252:
```suggestion
    class Program(Program, SDFGConvertible):  # type: ignore[no-redef]
```

src/gt4py/next/embedded/nd_array_field.py
egparedes, line 338: Just a comment: in this case it is not really needed to create a new subclass since it's not needed to add other parent classes, so we could just monkeypatch the original class with the two new extra methods...

tests/next_tests/integration_tests/multi_feature_tests/ffront_tests/test_laplacian.py
```diff
@@ -13,16 +13,26 @@
...
 from next_tests.integration_tests.cases import IDim, Ioff, JDim, Joff, cartesian_case
 from next_tests.integration_tests.feature_tests.ffront_tests.ffront_test_utils import (
     exec_alloc_descriptor,
 )
 
+try:
+    import dace
+    from gt4py.next.program_processors.runners.dace import run_dace_cpu, run_dace_gpu
+except ImportError:
+    dace: Optional[ModuleType] = None  # type:ignore[no-redef]
```
egparedes, line 32:
```suggestion
    dace: Optional[ModuleType] = None  # type: ignore[no-redef]
```

src/gt4py/next/embedded/nd_array_field.py
kotsaloscv: I would prefer to keep it with the subclass as I find it cleaner.
Although, if you think it is a better practice, I could change it.

kotsaloscv: Eventually I followed your suggestion because the redefinition of the NdArrayField class was causing issues (some tests were failing).

havogt (changes requested):
I think we are getting closer. I am still trying to understand how things interact and where stuff belongs. Maybe we could have a chat and you could teach me the interface.

tests/next_tests/integration_tests/multi_feature_tests/ffront_tests/test_laplacian.py
```diff
@@ -86,3 +96,33 @@ def test_ffront_lap(cartesian_case):
...
+        import cupy as xp
+    else:
+        import numpy as xp
+
+    in_field = cases.allocate(cartesian_case, laplap_program, "in_field")()
+    out_field = cases.allocate(cartesian_case, laplap_program, "out_field")()
+
+    @dace.program
+    def sdfg():
+        tmp_field = xp.empty_like(out_field)
```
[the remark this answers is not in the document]
kotsaloscv: I tried your suggestion, but DaCe tries to parse the `cases.allocate` __call__ as an sdfgConvertible and it fails. So, I will keep it as is.

src/gt4py/next/ffront/decorator.py
```diff
@@ -234,6 +246,62 @@ def __call__(self, *args, offset_provider: dict[str, Dimension], **kwargs: Any)
...
+            params = {str(p.id): p.dtype for p in self.itir.params}
+            fields = {str(p.id): p.type for p in self.past_stage.past_node.params}
+            arg_types = [
+                fields[pname]
+                if pname in fields
+                else ts.ScalarType(kind=type_translation.get_scalar_kind(dtype))
+                if dtype is not None
+                else ts.ScalarType(kind=ts.ScalarKind.INT32)
+                for pname, dtype in params.items()
+            ]
```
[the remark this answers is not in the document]
kotsaloscv: To build the sdfg, we just need the arg types and not the args themselves. So, this builds the types of the args and passes it to `translation.generate_sdfg`.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -326,6 +331,23 @@ def _slice(
...
+            obj = self.ndarray
+            if dace.dtypes.is_array(obj) and (
+                hasattr(obj, "__array_interface__") or hasattr(obj, "__cuda_array_interface__")
+            ):
+                if dace.dtypes.is_gpu_array(obj):
+                    return obj.__cuda_array_interface__["data"][0]  # type: ignore
+                else:
+                    return self.ndarray.__array_interface__["data"][0]  # type: ignore
+            else:
+                raise ValueError("Unsupported data container.")
```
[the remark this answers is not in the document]
kotsaloscv: it does not seem to work. FYI, the above code (using `__array_interface__  & __cuda_array_interface__ `) is what they do in DaCe.

egparedes, line 346: This should work for both numpy and cupy arrays. In the future we should try to only dlpack, but we need to write a couple of utility functions with ctypes to access the dlpack data structures so for now we're also using this trick in the storage buffer allocators.
```suggestion
            return self.array_ns.byte_bounds(self.ndarray)[0]
```

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -304,3 +310,96 @@ def build_sdfg_from_itir(
...
+            arg_types,
+            offset_provider=offset_provider,
+            column_axis=kwargs.get("column_axis", None),
+        )
+
+        sdfg.arg_names.extend(self.__sdfg_signature__()[0])
+        sdfg.arg_names.extend(list(self.__sdfg_closure__().keys()))
+
+        # TODO : Keep this metadata here or somewhere else ?
+        # Halo Exchange related info (focus on HORIZONTAL dim)
```
havogt, line 348:
```suggestion
        # TODO(kotsaloscv): Keep halo exchange related metadata (for horizontal dimension) here?
        # Could possibly be computed directly from the SDFG.
```

havogt:
Please
- add a few more details to the PR description (which will become the commit message)
- add yourself to the Authors file

kotsaloscv:
> Please
> 
> * add a few more details to the PR description (which will become the commit message)
> * add yourself to the Authors file

Thank you Hannes! It's done.

havogt (changes requested):

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -307,3 +313,89 @@ def build_sdfg_from_itir(
         sdfg.save(sdfg_filename)
 
     return sdfg
+
+
+@dataclasses.dataclass(frozen=True)
+class Program(Program, SDFGConvertible):  # type: ignore[no-redef]
+    """Extension of GT4Py Program : Implement SDFGConvertible interface"""
+
+    sdfg_convertible: dict[str, Any] = field(default_factory=dict)
```
egparedes, line 322: Can you explain the name of this attribute? If I understand the code correctly, it looks like something along the lines of  `sdfg_closure_vars` would be more appropriate... or am I missing something?

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -307,3 +313,89 @@ def build_sdfg_from_itir(
...
+                if k not in sdfg.GT4Py_Program_input_fields:
+                    continue
+                sdfg.offset_providers_per_input_field.setdefault(k, []).extend(list(v))
+
+        sdfg.stencil_horizontal_start = kwargs.get("horizontal_start", None)
+        sdfg.stencil_horizontal_end = kwargs.get("horizontal_end", None)
+
+        return sdfg
+
+    def __sdfg_closure__(self, reevaluate: Optional[Dict[str, str]] = None) -> Dict[str, Any]:
```
egparedes, line 390: What is the relation between the calls to `__sdfg_closure__()` and `__sdfg__()` from dace at run-time? I'm asking because if `__sdfg_closure__` is not called always after a call  to `__sdfg__` receiving a different offset provider, the closure could get easily out of sync easily.

tests/next_tests/integration_tests/feature_tests/dace/orchestration.py
```diff
@@ -0,0 +1,100 @@
...
+):
+    laplap(in_field, out=out_field[2:-2, 2:-2])
+
+
+def lap_ref(inp):
+    """Compute the laplacian using numpy"""
+    return -4.0 * inp[1:-1, 1:-1] + inp[:-2, 1:-1] + inp[2:, 1:-1] + inp[1:-1, :-2] + inp[1:-1, 2:]
+
+
+def test_sdfgConvertible_laplap(cartesian_case):
```
egparedes, line 73: Consider adding a test calling the same program passing different offset providers

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -419,6 +424,14 @@ def _slice(
         assert common.is_relative_index_sequence(slice_)
         return new_domain, slice_
 
+    if dace:
+        # Extension of NdArrayField : Support for SDFGConvertible GT4Py Programs
```
egparedes, line 428:
```suggestion
        # Extension of NdArrayField  adding SDFGConvertible support in GT4Py Programs
```

egparedes (commented):
I have some comments and questions about the code style and also the features implemented.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
kotsaloscv: This is a nice observation. But, indeed `__sdfg_closure__ ` is always called after a call to `__sdfg__ `.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -307,3 +313,89 @@ def build_sdfg_from_itir(
...
+        itir_tmp = apply_common_transforms(self.itir)
+        for closure in itir_tmp.closures:  # type: ignore[union-attr]
+            shifts = TraceShifts.apply(closure)
+            for k, v in shifts.items():
+                if k not in sdfg.GT4Py_Program_input_fields:
+                    continue
+                sdfg.offset_providers_per_input_field.setdefault(k, []).extend(list(v))
+
+        sdfg.stencil_horizontal_start = kwargs.get("horizontal_start", None)
+        sdfg.stencil_horizontal_end = kwargs.get("horizontal_end", None)
```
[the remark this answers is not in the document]
kotsaloscv: Good catch! Thank you.

tests/next_tests/integration_tests/feature_tests/dace/orchestration.py
kotsaloscv: Could you please elaborate a bit more. I am not sure I get it. Thanks.

egparedes: I meant that the current test is just a single programs without arguments which is called only once, so problems related to capturing too much state in the program closure at compile time cannot be detected.

I would change or extent the test program to receive all the arguments from outside (fields and offset_provider) and then call the program with completely different arguments to effectively test that the closure is not capturing run-time information.

kotsaloscv: I have updated the test. Thanks for the feedback!

Keep in mind that this PR implements the SDFGConvertible interface for the GT4Py programs. Therefore, the test just checks that a program placed inside an DaCe Program can be parsed as a nested SDFG.

The closure features and the runtime behavior will be properly taken into account in the Orchestration PR in icon4py.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -307,3 +312,177 @@ def build_sdfg_from_itir(
...
+                    break
+
+        # TODO(kotsaloscv): Keep halo exchange related metadata here?
+        # Could possibly be computed directly from the SDFG.
+
+        input_fields = [
+            str(inpt.id)
+            for closure in self.itir.closures
+            for inpt in closure.inputs
+            if str(inpt.id) in fields
```
egparedes, line 371:
```suggestion
            str(in_field.id)
            for closure in self.itir.closures
            for in_field in closure.inputs
            if str(in_field.id) in fields
```

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -307,3 +312,177 @@ def build_sdfg_from_itir(
...
+        ]
+        sdfg.gt4py_program_input_fields = {
+            inpt: dim
+            for inpt in input_fields
+            for dim in fields[inpt].dims  # type: ignore[union-attr]
```
egparedes, line 376:
```suggestion
            in_field: dim
            for in_field in input_fields
            for dim in fields[in_field].dims  # type: ignore[union-attr]
```

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -307,3 +312,177 @@ def build_sdfg_from_itir(
...
+    @classmethod
+    def crosscheck_dace_parsing(
+        cls, dace_parsed_args: list[Any], gt4py_program_args: list[Any]
+    ) -> bool:
+        for dace_parsed_arg, gt4py_program_arg in zip(dace_parsed_args, gt4py_program_args):
+            if isinstance(dace_parsed_arg, dace.data.Scalar):
+                assert dace_parsed_arg.dtype == as_dace_type(gt4py_program_arg)
+            elif isinstance(dace_parsed_arg, (bool, int, float, str)) or isinstance(
+                dace_parsed_arg, (np.bool_, np.integer, np.floating, np.str_)
+            ):  # compile-time constant scalar
```
egparedes, line 465:
```suggestion
            elif isinstance(dace_parsed_arg, (bool, int, float, str, np.bool_, np.integer, np.floating, np.str_)):  # compile-time constant scalar
```

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -307,3 +312,177 @@ def build_sdfg_from_itir(
...
+        return closure_dict
+
+    def __sdfg_signature__(self) -> tuple[Sequence[str], Sequence[str]]:
+        args = []
+        for arg in self.past_stage.past_node.params:
+            args.append(arg.id)
+        return (args, [])
+
+    @classmethod
+    def crosscheck_dace_parsing(
```
egparedes, line 457: This shouldn't be a class method but just a private free function in the module.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -307,3 +312,177 @@ def build_sdfg_from_itir(
...
+        for closure in itir_tmp.closures:  # type: ignore[union-attr]
+            shifts = itir_transforms.trace_shifts.TraceShifts.apply(closure)
+            for k, v in shifts.items():  # type: ignore[assignment]
+                if not isinstance(k, str):
+                    continue
+                if k not in sdfg.gt4py_program_input_fields:
+                    continue
+                sdfg.offset_providers_per_input_field.setdefault(k, []).extend(list(v))
+
+        return sdfg
```
egparedes, line 410: It's hard to understand all the stuff that's going on in this function. I think it should be split in small parts implemented as free private functions with meaningful names that help to understand the main idea.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -307,3 +312,177 @@ def build_sdfg_from_itir(
...
+        Keep in mind, that __sdfg_closure__ is called after __sdfg__.
+        """
+        offset_provider = self.connectivities
+
+        symbols = {
+            f"__{connectivity_identifier(k)}_{sym}": dace.symbol(
+                f"__{connectivity_identifier(k)}_{sym}"
+            )
+            for k, v in offset_provider.items()  # type: ignore[union-attr]
+            for sym in ["size_0", "size_1", "stride_0", "stride_1"]
```
egparedes, line 427: What are the literal strings here? Where is their meaning defined?

egparedes (changes requested):
It looks good, although I have a couple of style suggestion and some questions, mostly related to my poor understanding of the SDFG convertible API.

src/gt4py/next/iterator/embedded.py
```diff
@@ -113,6 +121,44 @@ def mapped_index(
...
+            obj = self.table
+            if dace.dtypes.is_array(obj) and (
+                hasattr(obj, "__array_interface__") or hasattr(obj, "__cuda_array_interface__")
+            ):
+                if dace.dtypes.is_gpu_array(obj):
+                    return obj.__cuda_array_interface__["data"][0]  # type: ignore
+                else:
+                    return obj.__array_interface__["data"][0]  # type: ignore
+            else:
+                raise ValueError("Unsupported data container.")
```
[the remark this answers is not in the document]
edopao: Yes, a function is provided although it is marked as private. I used it in `src/gt4py/next/program_processors/runners/dace_iterator/workflow.py`, see this import:
`from dace.codegen.compiled_sdfg import _array_interface_ptr as get_array_interface_ptr`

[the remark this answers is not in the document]
kotsaloscv: `get_array_interface_ptr` demands one more argument, i.e. StorageType. For this reason, I would prefer to stick for now with the simple version that I already have. Once data_ptr is used more ofter, I will create a utility function -waiting also for the GTIR-.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -307,3 +312,177 @@ def build_sdfg_from_itir(
...
+
+@dataclasses.dataclass(frozen=True)
+class Program(decorator.Program, dace.frontend.python.common.SDFGConvertible):
+    """Extension of GT4Py Program implementing the SDFGConvertible interface."""
+
+    sdfg_closure_vars: dict[str, Any] = field(default_factory=dict)
+
+    # Defined as class variable, otherwise the same connectivity table -name & data descriptor- across multiple GT4Py Programs
+    # will be considered as a different table (with subsequent name mangling). Therefore, the memory address of the table's data descriptor needs to be the same.
+    # This affects the case where a DaCe.Program calls multiple GT4Py Programs (fused SDFG).
```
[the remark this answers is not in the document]
kotsaloscv: You are right. The comment is unnecessary long and confusing. This choice about class variable is not at all restrictive, since the connectivitiy tables are defined fully symbolically. So multiple GT4Py programs would just need to set/concretize the symbols in their own run. The detail about dace data descriptors to share the same memory address -to avoid name mangling- is indeed very DaCe specific and not needed at this point.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
kotsaloscv: I find it quite concise as is (~70 lines), and for this reason I would prefer not to break it further. I have added some comments to clarify the different steps and did some reductions/reshuffling ;) .

src/gt4py/next/iterator/embedded.py
```diff
@@ -113,6 +121,44 @@ def mapped_index(
...
+        self.origin_axis = origin_axis
+        self.neighbor_axis = neighbor_axis
+        self.max_neighbors = max_neighbors
+        self.has_skip_values = has_skip_values
+        self.index_type = table.dtype
+
+    def mapped_index(
+        self, cur_index: int | np.integer, neigh_index: int | np.integer
+    ) -> Optional[int | np.integer]: ...
+
```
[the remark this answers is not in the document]
DropD: The preliminary name for mine is `CompileConnectivity`. I think `CompileTimeConnectivity` would be even better. I have been using the same `Compile` prefix for compile-time standins for field arguments as well. I could also imagine `Static` as a prefix instead.

src/gt4py/next/iterator/embedded.py
```diff
@@ -113,6 +121,41 @@ def mapped_index(
...
+            obj = self.table
+            if dace.dtypes.is_array(obj):
+                if hasattr(obj, "__array_interface__"):
+                    return obj.__array_interface__["data"][0]
+                if hasattr(obj, "__cuda_array_interface__"):
+                    return obj.__cuda_array_interface__["data"][0]
+            raise ValueError("Unsupported data container.")
+
+        def __descriptor__(self) -> dace.data.Data:
+            return dace.data.create_datadescriptor(self.table)
```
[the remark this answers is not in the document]
kotsaloscv: Thank you Rico! I like the suugestion and I will modify the code accordingly ;)

src/gt4py/next/iterator/embedded.py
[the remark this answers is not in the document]
kotsaloscv: FYI, I have named it `CompileTimeConnectivity`. As decided with @DropD, since this PR will be merged first, we leave it as is and then Rico removes my `CompileTimeConnectivity` class from embedded.py (TODO in his PR).

tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+@gtx.field_operator
+def _testee(a: gtx.Field[gtx.Dims[Vertex], gtx.float64]):
+    return a(E2V[0])
+
+
+@gtx.program
+def testee(a: gtx.Field[gtx.Dims[Vertex], gtx.float64], b: gtx.Field[gtx.Dims[Edge], gtx.float64]):
+    _testee(a, out=b)
```
egparedes, line 54: I think it would make sense to move these definitions closer to the actual test that is using them

egparedes (commented):
Ok, here are my last very small comments related to the Python style and it should be ready to be merged. I have some questions about the tests, but I think it would be easier to set a quick meeting to talk about it.

src/gt4py/next/iterator/embedded.py
```diff
@@ -113,6 +121,64 @@ def mapped_index(
...
+            max_neighbors=connectivity.max_neighbors,
+            has_skip_values=connectivity.has_skip_values,
+            origin_axis=connectivity.origin_axis,
+            neighbor_axis=connectivity.neighbor_axis,
+            index_type=connectivity.index_type,
+        )
+
+    @property
+    def table(self):
+        return None
```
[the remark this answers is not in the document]
kotsaloscv: exactly, it is required to return something, and None is the most meaningful at this point.

egparedes (approved):
LGTM. Thanks for the hard work to make this work in a clean way.

src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
```diff
@@ -119,11 +121,13 @@ def _make_array_shape_and_strides(
             neighbor_tables[dim.value].max_neighbors
             if dim.kind == DimensionKind.LOCAL
             # we reuse the same gt4py symbol for field size passed as scalar argument which is used in closure domain
-            else dace.symbol(f"__{name}_size_{i}", dtype)
+            else dace.symbol(connectivity_table_size_symbol(name, i), dtype)
```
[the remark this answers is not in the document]
kotsaloscv: Since this is used only for the neighbor_tables, I thought that the name would be a good fit. Also, it is used in the SDFGConvertible interface, for adding the connectivity tables symbols in the fused SDFG.
If you are ok with it, I would prefer to leave it as is.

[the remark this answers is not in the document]
edopao: The function `_make_array_shape_and_strides` is used for all fields, not only connectivity tables.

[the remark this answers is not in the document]
kotsaloscv: You are right then. I will modify the name and ping you.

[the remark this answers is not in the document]
edopao: Thanks, it looks good to me now.

egparedes:
cscs-ci run

egparedes:
cscs-ci run default

kotsaloscv:
cscs-ci run

kotsaloscv:
cscs-ci run default

PR #1532, opened 2024-04-18
refactor[next]: Use is_call_to instead of equality comparison with itir.Ref.

tehrengruber:
In the new ITIR type inference #1531 IR nodes store their type in the node itself. While we initially exclude the attribute from equality comparison we should nonetheless avoid comparison of node that only differ in type. This PR removes many of this occurrences.


src/gt4py/next/iterator/ir_utils/common_pattern_matcher.py
```diff
@@ -32,6 +31,19 @@ def is_let(node: itir.Node) -> TypeGuard[itir.FunCall]:
...
+    Match call expression to a given function.
+
+    >>> from gt4py.next.iterator.ir_utils import ir_makers as im
+    >>> node = im.call("make_tuple")(1, 2)
+    >>> is_call_to(node, "make_tuple")
+    True
+    >>> is_call_to(node, "plus")
+    False
+    """
+    if isinstance(fun, (list, tuple, set)):
```
egparedes, line 45: This `isinstance()` check doesn't match the typying annotation for `fun`.

src/gt4py/next/iterator/ir_utils/common_pattern_matcher.py
```diff
@@ -32,6 +31,19 @@ def is_let(node: itir.Node) -> TypeGuard[itir.FunCall]:
     return isinstance(node, itir.FunCall) and isinstance(node.fun, itir.Lambda)
 
 
-def is_if_call(node: itir.Expr) -> TypeGuard[itir.FunCall]:
-    """Match expression of the form `if_(cond, true_branch, false_branch)`."""
-    return isinstance(node, itir.FunCall) and node.fun == im.ref("if_")
+def is_call_to(node: itir.Node, fun: str | list[str]) -> TypeGuard[itir.FunCall]:
```
egparedes, line 34: Just an **optional** suggestion, feel free to ignore: I'd like better to include the actual name of the node `FunCall` in its typeguard function (e.g. `is_funcall_to`).

src/gt4py/next/iterator/ir_utils/common_pattern_matcher.py
tehrengruber: I switched to `collections.abc.Iterable`, but still included `list`, `tuple`, `set` for performance reasons.

src/gt4py/next/iterator/ir_utils/common_pattern_matcher.py
tehrengruber: We opted for maximum convenience in the ir_makers module, e.g. `im.ref` also creates a `SymRef` so I'll keep `call` for now.

egparedes (approved):
ILGTM.

PR #1533, opened 2024-04-18
refactor[next]: move lift_mode itir test fixture into program_processor

havogt:
Currently we have a mix of specifying the backend which already comes with a lift_mode default and separately the lift_mode fixture in some tests. The default was not overwritten in the roundtrip backend. Now we remove the separate `lift_mode` and add extra backends with the lift_mode set.

Note: we don't run double_roundtrip with temporaries and we don't use the lift_mode==SIMPLE_HEURISTIC in roundtrip, only `USE_TEMPORARIES`.

Longer term we should refactor all itir tests to use the ffront test infrastructure.


tests/next_tests/unit_tests/conftest.py
```diff
@@ -35,18 +35,6 @@
...
-    ],
-    ids=lambda p: f"lift_mode={p.name}",
-)
-def lift_mode(request):
-    return request.param
-
-
 OPTIONAL_PROCESSORS = []
 if dace_iterator:
     OPTIONAL_PROCESSORS.append((next_tests.definitions.OptionalProgramBackendId.DACE_CPU, True))
```
edopao, line 40: Should we enable `run_dace_cpu_with_temporaries` and `run_dace_gpu_with_temporaries`, which are already defined in `src/gt4py/next/program_processors/runners/dace.py`?

havogt: - GPU is not enabled because in itir tests we don't (always) use the test infrastructure that uses the proper allocators.
- If you like you can try adding the DaCe with temporaries in a separate PR. It's not too important from my point of view because we shift focus on the fieldview->DaCe backend.

edopao:
I think with this change we can also remove a couple of TODOs and temporary workarouds:

https://github.com/GridTools/gt4py/blob/main/src/gt4py/next/program_processors/codegens/gtfn/gtfn_module.py
`TODO(tehrengruber): Remove lift_mode from call interface.`

https://github.com/GridTools/gt4py/blob/main/src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
`TODO(tehrengruber): Remove lift_mode from call interface.`

havogt:
Thank you! Is this https://github.com/GridTools/gt4py/blob/2c06bc668aaacb21075fd39b179985d9e7172ae3/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py#L253 still true?

edopao:
> Thank you! Is this
> 
> https://github.com/GridTools/gt4py/blob/2c06bc668aaacb21075fd39b179985d9e7172ae3/src/gt4py/next/program_processors/runners/dace_iterator/__init__.py#L253
> 
> still true?

No, this comment is not true anymore. It can be removed.

edopao (approved):

docs/user/next/QuickstartGuide.md
```diff
@@ -150,10 +150,21 @@ The sign of the edge difference in the sum of the pseudo-laplacian is always suc
...
+      - [Fields](#fields)
+      - [Field operators](#field-operators)
+      - [Programs](#programs)
+      - [Composing field operators and programs](#composing-field-operators-and-programs)
+    - [Operations on unstructured meshes](#operations-on-unstructured-meshes)
+      - [Defining the mesh and its connectivities](#defining-the-mesh-and-its-connectivities)
+      - [Using connectivities in field operators](#using-connectivities-in-field-operators)
+      - [Using reductions on connected mesh elements](#using-reductions-on-connected-mesh-elements)
+      - [Using conditionals on fields](#using-conditionals-on-fields)
+      - [Implementing the pseudo-laplacian](#implementing-the-pseudo-laplacian)
```
havogt, line 167:
```suggestion
- [Defining the mesh and the connectivities (adjacencies) between cells and edges](#Defining-the-mesh-and-its-connectivities)
- [Using connectivities in field operators](#Using-connectivities-in-field-operators)
- [Using reductions on connected mesh elements](#Using-reductions-on-connected-mesh-elements)
- [Implementing the actual pseudo-laplacian](#Implementing-the-pseudo-laplacian)
```

src/gt4py/next/program_processors/runners/double_roundtrip.py
```diff
@@ -22,5 +22,5 @@
     executor=roundtrip.RoundtripExecutorFactory(
         dispatch_backend=roundtrip.RoundtripExecutorFactory()
```
tehrengruber, line 23:
```suggestion
        dispatch_backend=roundtrip.default.executor
```
Just a suggestion.

src/gt4py/next/program_processors/runners/roundtrip.py
```diff
@@ -209,7 +209,7 @@ def execute_roundtrip(
     *args: Any,
     column_axis: Optional[common.Dimension] = None,
     offset_provider: dict[str, embedded.NeighborTableOffsetProvider],
-    debug: bool = False,
+    debug: bool = config.DEBUG,
```
tehrengruber, line 212: Note that this means changing the `config.DEBUG` flag after this module is imported has no effect. Maybe retrieving the value in the function is what we actually want.

src/gt4py/next/program_processors/runners/roundtrip.py
```diff
@@ -230,7 +230,7 @@ def execute_roundtrip(
 
 @dataclasses.dataclass(frozen=True)
 class Roundtrip(workflow.Workflow[stages.ProgramCall, stages.CompiledProgram]):
-    debug: bool = False
+    debug: bool = config.DEBUG
```
tehrengruber, line 233: Same comment as above.

tehrengruber (approved):
Just small things. Since `LiftMode.SIMPLE_HEURISTIC` is not tested anymore I would add either a comment to the heuristic or remove it.

PR #1534, opened 2024-04-18
test[next]: delete cppbackend_tests

havogt:
Remove the cpp_backend_tests. They were from a time when we couldn't run gtfn from python.


egparedes (approved):
ILGTM

PR #1538, opened 2024-04-24
feat[next][dace]: Skeleton of GTIR DaCe backend

edopao:
We introduce a new DaCe backend that translates a GTIR program to SDFG. GTIR is still sometimes referred to as Combined-IR and provides a fieldview representation of the GT4Py program. The GT4Py frontend is not ready to lower a GT4Py program to GTIR, so for now the unit tests use manually written GTIR programs as input.

The new DaCe backend is implemented in the package `gt4py.next.program_processors.runners.dace_fieldview`
This package provides, for now, the method `build_sdfg_from_gtir` to build the DaCe SDFG.

This PR introduces translation capability for the following GTIR types:`ir.Program`, `ir.Stmt`, `ir.SetAt`. And for the following new builtin functions: `as_field_op` and `select`.


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_symbol_ref.py
```diff
@@ -0,0 +1,86 @@
...
+        assert len(access_nodes) == 1
+        return access_nodes[0]
+
+    def _build(self) -> list[tuple[dace.nodes.Node, ts.FieldType | ts.ScalarType]]:
+        sym_node = self._get_access_node()
+        if sym_node:
+            # if already present in current state, share access node
+            pass
+
+        elif isinstance(self._sym_type, ts.FieldType):
```
egparedes, line 65: Is this correct? If the call to `_get_access_node()` returns something a node instead of `None`, I would assume that it would go through the `pass` branch so the `isinstance()` check would never be `True`. Am I reading this wrong?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_field_operator.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # create map range corresponding to the field operator domain
+        map_ranges = {
+            f"i_{dim.value}": f"{lb}:{ub}" for dim, (lb, ub) in self._field_domain.items()
+        }
+
+        input_nodes: dict[str, dace.nodes.AccessNode] = {}
+        input_memlets: dict[str, dace.Memlet] = {}
+        assert len(self._stencil_args) == len(self._stencil_expr.params)
+        for arg, param in zip(self._stencil_args, self._stencil_expr.params):
```
egparedes, line 91: Should this be called with `strict=True`? The same comment applies to all `zip()` calls in this PR

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -0,0 +1,57 @@
...
+    Returns a new dictionary containing only `Connectivity` values.
+    """
+    return {
+        offset: table
+        for offset, table in offset_provider.items()
+        if isinstance(table, Connectivity)
+    }
+
+
+def unique_name(prefix: str) -> str:
```
egparedes, line 52: Use `eve.utils.UIDGenerator`

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -0,0 +1,57 @@
...
+    if type_.kind == ts.ScalarKind.BOOL:
+        return dace.bool_
+    elif type_.kind == ts.ScalarKind.INT32:
+        return dace.int32
+    elif type_.kind == ts.ScalarKind.INT64:
+        return dace.int64
+    elif type_.kind == ts.ScalarKind.FLOAT32:
+        return dace.float32
+    elif type_.kind == ts.ScalarKind.FLOAT64:
+        return dace.float64
```
egparedes, line 34: Just an optional suggestion: I'd use either `match()` or a global read-only mapping with the translation pairs.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_tasklet_codegen.py
```diff
@@ -0,0 +1,118 @@
...
+                raise NotImplementedError(f"'{builtin_name}' not implemented.")
+        raise NotImplementedError(f"Unexpected 'FunCall' node ({node}).")
+
+    @final
+    def visit_Literal(self, node: itir.Literal) -> str:
+        return node.value
+
+    @final
+    def visit_SymRef(self, node: itir.SymRef) -> str:
+        return str(node.id)
```
egparedes, line 118: Optional suggestion: this class doesn't seem to use any special functionality of a `TemplatedGenerator` class so it seems it could be written as a regular visitor. An alternative could be to try to use more features  like for example static templates as supported in the `TemplatedGenerator` class
```suggestion
    Literal = as_fmt("{value}")

    SymRef = as_fmt("{id}")
```

egparedes (changes requested):
First round of reviews. It looks good to me in general. I'm basically adding only style comments. I have a couple of questions about the overall design that maybe we could discuss offline.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_symbol_ref.py
edopao: Correct. If the access node is already present in current state, we can simply return it at the end of this function.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow_builder.py
```diff
@@ -0,0 +1,183 @@
...
+        Specialized visit method for pure stencil expressions.
+
+        Returns a string represnting the Python code to be used as tasklet body.
+        TODO: should we return a list of code strings in case of tuple returns,
+        one for each output value?
+        """
+        return GtirTaskletCodegen().visit(node)
+
+    def visit_FunCall(self, node: itir.FunCall) -> Callable:
+        from gt4py.next.program_processors.runners.dace_fieldview import gtir_builtins
```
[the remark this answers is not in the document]
edopao: It was a way to avoid circular dependency, since `AsFieldOp` and `Select` are derived from `GtirDataflowBuilder`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -0,0 +1,201 @@
...
+        strides = [dace.symbol(f"__{name}_stride_{i}", dtype) for i, _ in enumerate(dims)]
+        return shape, strides
+
+    def _add_storage(self, sdfg: dace.SDFG, name: str, type_: ts.TypeSpec) -> None:
+        """
+        Add external storage (aka non-transient) for data containers passed as arguments to the SDFG.
+
+        For fields, it allocates dace arrays, while scalars are stored as SDFG symbols.
+        """
+        assert isinstance(type_, (ts.FieldType, ts.ScalarType))
```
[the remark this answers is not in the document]
edopao: `ts.FieldType` and `ts.ScalarType` are subtypes of `ts.TypeSpec`

edopao (commented):
Thanks for the review. I will address all codestyle comments. I have answered to the questions.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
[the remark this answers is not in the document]
edopao: Indeed, `mypy` complains about incompatible types.

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
edopao: It requires an instance, right?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow_builder.py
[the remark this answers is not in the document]
edopao: I still see the circular import dependency, unfortunately.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtins/gtir_builtin_field_operator.py
```diff
@@ -0,0 +1,135 @@
...
+        self.field_type = ts.FieldType([dim for dim, _, _ in domain], node_type)
+        self.stencil_expr = stencil_expr
+        self.stencil_args = stencil_args
+
+    def build(self) -> list[tuple[dace.nodes.Node, ts.FieldType | ts.ScalarType]]:
+        # generate a tasklet node implementing the stencil function and represent
+        # the field operator as a mapped tasklet, which will range over the field domain
+        output_connector = "__out"
+        tlet_code = "{var} = {code}".format(
+            var=output_connector, code=GTIRTaskletCodegen().visit(self.stencil_expr.expr)
```
edopao, line 77: Visit the lambda expression as a nested SDFG, in order to generate small tasklets inside one dace node. Alternatively, explore a solution to collect the set of tasklets with internal connections without using a nested SDFG.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtins/gtir_builtin_field_operator.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+
+        # allocate local temporary storage for the result field
+        field_shape = [
```
havogt, line 81: Note to myself: does the data have to flow over a "local storage"

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
egparedes: Yes, it requires an instance so differences instances can be used for different purposes instead of relying on global state. It shouldn't be a problem to use it here, though.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -0,0 +1,207 @@
...
+    connector: str
+
+
+@dataclass(frozen=True)
+class IteratorExpr:
+    """Iterator for field access to be consumed by `deref` or `shift` builtin functions."""
+
+    field: dace.nodes.AccessNode
+    dimensions: list[str]
+    offset: list[dace.symbolic.SymExpr]
```
edopao, line 62: The attribute `offset` can be removed. I will remove it in next update and implement cartesian shift by adding symbolic expressions in `index` attribute.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtins/gtir_builtin_symbol_ref.py
```diff
@@ -0,0 +1,80 @@
...
+                {},
+                {"__out"},
+                f"__out = {self.sym_name}",
+            )
+            sym_node = self.add_local_storage(self.sym_type, shape=[])
+            self.head_state.add_edge(
+                tasklet_node, "__out", sym_node, None, dace.Memlet(data=sym_node.data, subset="0")
+            )
+
+        return [(sym_node, self.sym_type)]
```
egparedes, line 80: Isn't the code equivalent to the following snippet with a simpler control flow?
```suggestion
        sym_node = self._get_access_node()
        if sym_node is None:
            if isinstance(self.sym_type, ts.FieldType):
                # add access node to current state
                sym_node = self.head_state.add_access(self.sym_name)
            else:
                # scalar symbols are passed to the SDFG as symbols: build tasklet node
                # to write the symbol to a scalar access node
                assert self.sym_name in self.sdfg.symbols
                tasklet_node = self.head_state.add_tasklet(
                    f"get_{self.sym_name}",
                    {},
                    {"__out"},
                    f"__out = {self.sym_name}",
                )
                sym_node = self.add_local_storage(self.sym_type, shape=[])
                self.head_state.add_edge(
                    tasklet_node, "__out", sym_node, None, dace.Memlet(data=sym_node.data, subset="0")
                )

        return [(sym_node, self.sym_type)]
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtins/gtir_builtin_symbol_ref.py
```diff
@@ -0,0 +1,80 @@
...
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+from typing import Optional
+
+import dace
+
+from gt4py.next.program_processors.runners.dace_fieldview.gtir_builtins.gtir_builtin_translator import (
+    GTIRBuiltinTranslator,
```
egparedes, line 21: Do not directly import symbols only modules.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtins/gtir_builtin_select.py
```diff
@@ -0,0 +1,125 @@
...
+
+import dace
+
+from gt4py import eve
+from gt4py.next.iterator import ir as itir
+from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm
+from gt4py.next.program_processors.runners.dace_fieldview.gtir_builtins.gtir_builtin_translator import (
+    GTIRBuiltinTranslator,
+)
+from gt4py.next.program_processors.runners.dace_fieldview.utility import get_symbolic_expr
```
egparedes, line 26: Do not directly import symbols only modules.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtins/gtir_builtin_select.py
```diff
@@ -0,0 +1,125 @@
...
+from gt4py.next.type_system import type_specifications as ts
+
+
+class GTIRBuiltinSelect(GTIRBuiltinTranslator):
+    """Generates the dataflow subgraph for the `select` builtin function."""
+
+    true_br_builder: Callable
+    false_br_builder: Callable
```
egparedes, line 34: Just a question: isn't it possible to make these type hints more strict (e.g. defining the return type of the callables)?

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -0,0 +1,96 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from typing import Any, Mapping
+
+import dace
+
+from gt4py.next.common import Connectivity, Dimension
+from gt4py.next.iterator import ir as itir
+from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm
+from gt4py.next.program_processors.runners.dace_fieldview.gtir_python_codegen import (
```
egparedes, line 22: Do not directly import symbols only modules.

egparedes (commented):
It looks good for me. I have some questions and comments about coding and typing style but they are minor.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -0,0 +1,259 @@
...
+    of a GT4Py program as a sequence of `ir.Stmt` (aka statement) expressions.
+    Each statement is translated to a taskgraph inside a separate state. Statement states are chained
+    one after the other: concurrency between states should be extracted by means of SDFG analysis.
+    The translator will extend the SDFG while preserving the property of single exit state:
+    branching is allowed within the context of one statement, but in that case the statement should
+    terminate with a join state; the join state will represent the head state for next statement,
+    from where to continue building the SDFG.
+    """
+
+    data_types: dict[str, ts.FieldType | ts.ScalarType]
```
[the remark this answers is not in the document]
edopao: Yes, the only problem is that there are gt4py symbols and dace symbols.. and they are not the same thing.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -0,0 +1,259 @@
...
+    ) -> list[dace.nodes.AccessNode]:
+        """
+        Specialized visit method for fieldview expressions.
+
+        This method represents the entry point to visit 'Stmt' expressions.
+        As such, it must preserve the property of single exit state in the SDFG.
+
+        TODO: do we need to return the GT4Py `FieldType`/`ScalarType`?
+        """
+        expr_builder = self.visit(node, sdfg=sdfg, head_state=head_state)
```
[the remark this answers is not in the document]
edopao: Changed to:
`assert isinstance(expr_builder, SDFGFieldBuilder)`

where:
`SDFGFieldBuilder: TypeAlias = Callable[[], list[SDFGField]]`

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def visit_Lambda(
+        self, node: itir.Lambda, args: list[SymbolExpr | IteratorExpr | MemletExpr]
+    ) -> tuple[
+        list[InputConnection],
+        TaskletExpr,
+    ]:
+        for p, arg in zip(node.params, args, strict=True):
+            self.symbol_map[str(p.id)] = arg
+        output_expr = self.visit(node.expr)
```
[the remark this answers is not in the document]
edopao: Right. However, based on some discussion with SPCL team, I will unify the return type to always be an access node.

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -0,0 +1,96 @@
...
+    Returns a new dictionary containing only `Connectivity` values.
+    """
+    return {
+        offset: table
+        for offset, table in offset_provider.items()
+        if isinstance(table, Connectivity)
+    }
+
+
+def get_domain(
```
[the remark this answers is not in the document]
edopao: Good point. However, this comment could be applied in theory to any `itir.Expr`. I understand that it makes sense here, particularly, because we are going to encounter exactly the same domain, in multiple places for one program.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtins/gtir_builtin_translator.py
```diff
@@ -0,0 +1,71 @@
...
+from typing import final
+
+import dace
+
+from gt4py.next.program_processors.runners.dace_fieldview.utility import as_dace_type, unique_name
+from gt4py.next.type_system import type_specifications as ts
+
+
+@dataclass(frozen=True)
+class GTIRBuiltinTranslator:
```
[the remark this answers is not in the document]
edopao: Changed to inherit from `abc.ABC`

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtins/gtir_builtin_select.py
edopao: Right. I have introduced the type alias:
`SDFGFieldBuilder: TypeAlias = Callable[[], list[SDFGField]]`

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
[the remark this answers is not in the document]
edopao: I realized I cannot call `isinstance` with a `TypeAlias` argument.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtins/gtir_builtin_field_operator.py
edopao: For the final write to an external field, this internal node is redundant, and dace simplify pass will remove it. On the other hand, it is needed when we write back to a field also used as input (in-out field parameters): in this case, the dace simplify pass will correctly keep it. I have added a testcase `test_gtir_update` to cover this case.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace_fieldview.py
```diff
@@ -0,0 +1,380 @@
...
+        function_definitions=[],
+        params=[itir.Sym(id="x"), itir.Sym(id="size")],
+        declarations=[],
+        body=[
+            itir.SetAt(
+                expr=im.call(
+                    im.call("as_fieldop")(
+                        im.lambda_("a")(im.plus(im.deref("a"), 1.0)),
+                        domain,
+                    )
```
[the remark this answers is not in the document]
havogt: After discussion with enrique, probably we should pass in a `broadcast(1.0, ())`. We can discuss this separately.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -0,0 +1,269 @@
...
+    def visit_FunCall(
+        self, node: itir.FunCall, sdfg: dace.SDFG, head_state: dace.SDFGState
+    ) -> gtir_builtin_translators.SDFGFieldBuilder:
+        # first visit the argument nodes
+        arg_builders = []
+        for arg in node.args:
+            arg_builder: gtir_builtin_translators.SDFGFieldBuilder = self.visit(
+                arg, sdfg=sdfg, head_state=head_state
+            )
+            arg_builders.append(arg_builder)
```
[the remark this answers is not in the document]
edopao: I tried that but it does not work. The reason is that the `generic_visit` method does not aggregate the results:
```
    def generic_visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
        for child in trees.iter_children_values(node):
            self.visit(child, **kwargs)

        return None
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,366 @@
...
+        node_type = ts.ScalarType(kind=ts.ScalarKind.FLOAT64)
+
+        self.field_domain = dace_fieldview_util.get_field_domain(domain_expr)
+        self.field_dtype = node_type
+        self.stencil_expr = stencil_expr
+        self.stencil_args = stencil_args
+
+    def build(self) -> list[SDFGField]:
+        # type of variables used for field indexing
+        index_dtype = dace.int32
```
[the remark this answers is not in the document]
edopao: Agreed.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace_fieldview.py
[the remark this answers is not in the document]
edopao: Test IR updated based on today's discussion.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,322 @@
...
+                        ITERATOR_INDEX_DTYPE,
+                    )
+                    for dim, _, _ in field_domain
+                }
+                iterator_arg = gtir_to_tasklet.IteratorExpr(
+                    data_node,
+                    arg_type.dims,
+                    indices,
+                )
+                stencil_args.append(iterator_arg)
```
egparedes, line 117: Just a suggestion
```suggestion
            if isinstance(arg_type, ts.ScalarType):
                arg_definition = gtir_to_tasklet.MemletExpr(data_node, sbs.Indices([0]))
            else:
                assert isinstance(arg_type, ts.FieldType)
                indices: dict[Dimension, gtir_to_tasklet.IteratorIndexExpr] = {
                    dim: gtir_to_tasklet.SymbolExpr(
                        dace.symbolic.SymExpr(DIMENSION_INDEX_FMT.format(dim=dim.value)),
                        ITERATOR_INDEX_DTYPE,
                    )
                    for dim, _, _ in field_domain
                }
                arg_definition = gtir_to_tasklet.IteratorExpr(
                    data_node,
                    arg_type.dims,
                    indices,
                )
             stencil_args.append(arg_definition)
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,322 @@
...
+        if isinstance(last_node, dace.nodes.Tasklet):
+            # the last transient node can be deleted
+            last_node_connector = state.in_edges(output_expr.node)[0].src_conn
+            state.remove_node(output_expr.node)
+        else:
+            last_node = output_expr.node
+            last_node_connector = None
+
+        # allocate local temporary storage for the result field
+        field_dims = [dim for dim, _, _ in field_domain]
```
egparedes, line 136: Just an idea since there are so many `_` in the following list comprehensions with `field_domain`, maybe unpacking its components first would help:
```python
        dims, ubs, lbs = zip(*field_domain)
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,322 @@
...
+
+class AsFieldOp(PrimitiveTranslator):
+    """Generates the dataflow subgraph for the `as_field_op` builtin function."""
+
+    callable_args: list[itir.Expr]
+
+    def __init__(self, node_args: list[itir.Expr]):
+        self.callable_args = node_args
+
+    def __call__(
```
egparedes, line 71: Just a question since this is a long function: would it be possible to split its different parts into smaller and reusable functions?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,322 @@
...
+            mx,
+            field_node,
+            src_conn=last_node_connector,
+            memlet=dace.Memlet(data=field_node.data, subset=",".join(output_subset)),
+        )
+
+        return [(field_node, field_type)]
+
+
+class Select(PrimitiveTranslator):
```
egparedes, line 196: Should we rename it as `Cond` already in this PR? I'd say yes...

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dace_backend.py
```diff
@@ -0,0 +1,37 @@
...
+    gtir_to_sdfg as gtir_dace_translator,
+)
+
+
+def build_sdfg_from_gtir(
+    program: itir.Program,
+    offset_provider: dict[str, Connectivity | Dimension],
+) -> dace.SDFG:
+    """
+    TODO: enable type inference
```
egparedes, line 29: Please add a proper docstring and/or move the TODO and source code as comments outside of the docstring.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -0,0 +1,248 @@
...
+                target_node,
+                dace.Memlet(data=target_node.data, subset=subset),
+            )
+
+    def visit_FunCall(
+        self, node: itir.FunCall, sdfg: dace.SDFG, head_state: dace.SDFGState
+    ) -> list[gtir_builtin_translators.SDFGField]:
+        # use specialized dataflow builder classes for each builtin function
+        if cpm.is_call_to(node.fun, "as_fieldop"):
+            return gtir_builtin_translators.AsFieldOp(node.args)(node.fun, sdfg, head_state, self)
```
egparedes, line 225: As mentioned above, I think this should just be:
```suggestion
            return gtir_builtin_translators.AsFieldOp(node, sdfg, head_state, self)
```

egparedes (commented):
I like the refactoring work done so far. I have some questions and suggestions to finish the work.

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -0,0 +1,126 @@
...
+) -> dict[Dimension, tuple[dace.symbolic.SymbolicType, dace.symbolic.SymbolicType]]:
+    """
+    Returns domain represented in dictionary form.
+    """
+    domain = get_domain(node)
+
+    return {dim: (lb, ub) for dim, lb, ub in domain}
+
+
+def get_symbolic_expr(node: itir.Expr) -> str:
```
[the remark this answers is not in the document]
edopao: I like the suggestion, but I have implemented it in a slightly different way. I have implemented the free function `get_python_source`  in a new module (`gtir_python_codegen`) together with the class `PythonCodegen` and the helper method to format a builtin function (`format_builtin`).

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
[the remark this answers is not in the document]
edopao: It is an interesting suggestion. Indeed, the caching could be made generic and `itir.Lambda` nodes could also be cached. I would prefer to skip it in this PR, but it could be something to consider to reduce the lowering time.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,322 @@
...
+import dace.subsets as sbs
+
+from gt4py.next.common import Dimension
+from gt4py.next.iterator import ir as itir
+from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm
+from gt4py.next.program_processors.runners.dace_fieldview import (
+    gtir_to_tasklet,
+    utility as dace_fieldview_util,
+)
+from gt4py.next.program_processors.runners.dace_fieldview.sdfg_builder import SDFGBuilder
```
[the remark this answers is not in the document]
edopao: I tried your suggestion but I still had problems with circular imports. I have tried a different approach.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
edopao: I cannot call the primitive translator in the way you suggest. Currently I have to do in this way:
`gtir_builtin_translators.AsFieldOp()(node, sdfg, head_state, self)`

I have the impression that this pattern is not really good. What I need is basically a static class method, or a module function, for each primitive.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: Good idea. I have extracted a couple of helper methods.

edopao:
cscs-ci run default

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,385 @@
...
+            access to the result of the primitive subgraph. The GT4Py data type is useful
+            in the case the returned data is an array, because the type provdes the domain
+            information (e.g. order of dimensions, dimension types).
+        """
+
+
+class AsFieldOp(PrimitiveTranslator):
+    """Generates the dataflow subgraph for the `as_field_op` builtin function."""
+
+    @classmethod
```
egparedes, line 88: These two classmethods can now become private functions in the module

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,385 @@
...
+            mx,
+            field_node,
+            src_conn=last_node_connector,
+            memlet=dace.Memlet(data=field_node.data, subset=",".join(output_subset)),
+        )
+
+        return [(field_node, field_type)]
+
+
+class Cond(PrimitiveTranslator):
```
egparedes, line 257: All translator can be implemented as functions

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dace_backend.py
egparedes, line 1: Optional suggestion: maybe you can find a better name for this module since it doesn't include a proper `backend`....

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -0,0 +1,249 @@
...
+# GT4Py is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or any later
+# version. See the LICENSE.txt file at the top-level directory of this
+# distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+from dataclasses import dataclass
```
egparedes, line 16: Forbidden direct symbol import
```suggestion
import dataclass
```

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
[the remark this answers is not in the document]
egparedes: Ok, let's skip it, but if you think the suggestion makes sense, please add a TODO to not to forget

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -0,0 +1,124 @@
...
+    for named_range in node.args:
+        assert cpm.is_call_to(named_range, "named_range")
+        assert len(named_range.args) == 3
+        axis = named_range.args[0]
+        assert isinstance(axis, gtir.AxisLiteral)
+        bounds = []
+        for arg in named_range.args[1:3]:
+            sym_str = gtir_python_codegen.get_source(arg)
+            sym_val = dace.symbolic.SymExpr(sym_str)
+            bounds.append(sym_val)
```
egparedes, line 101: Just an optional suggestion
```suggestion
        bounds = [
            dace.symbolic.SymExpr(gtir_python_codegen.get_source(arg))
            for arg in named_range.args[1:3]
        ]
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,385 @@
...
+                        IteratorIndexDType,
+                    )
+                    for dim, _, _ in domain
+                }
+                arg_definition = gtir_to_tasklet.IteratorExpr(
+                    data_node,
+                    arg_type.dims,
+                    indices,
+                )
+            stencil_args.append(arg_definition)
```
egparedes, line 124: I think this helper function could just deal with the parsing/transformation of one single argument, and then use it in a list comprehension from outside;
```python
stencil_args = [parse_arg_expr(arg) for arg in node.args]
``` 

```suggestion
def _parse_arg_expr(arg: itir.Expr, ....):
    fields: list[TemporaryData] = sdfg_builder.visit(arg, sdfg=sdfg, head_state=state)
    assert len(fields) == 1
    data_node, arg_type = fields[0]
    # require all argument nodes to be data access nodes (no symbols)
    assert isinstance(data_node, dace.nodes.AccessNode)

    arg_definition: gtir_to_tasklet.IteratorExpr | gtir_to_tasklet.MemletExpr
    if isinstance(arg_type, ts.ScalarType):
        arg_definition = gtir_to_tasklet.MemletExpr(data_node, sbs.Indices([0]))
    else:
        assert isinstance(arg_type, ts.FieldType)
        indices: dict[gtx_common.Dimension, gtir_to_tasklet.IteratorIndexExpr] = {
            dim: gtir_to_tasklet.SymbolExpr(
                dace_fieldview_util.get_map_variable(dim),
                IteratorIndexDType,
            )
            for dim, _, _ in domain
        }
        arg_definition = gtir_to_tasklet.IteratorExpr(
            data_node,
            arg_type.dims,
            indices,
        )
    return arg_definition
```

egparedes (commented):
Almost ready to go, just the final changes...

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: Agreed, much better.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dace_backend.py
edopao: Agree, renamed to `workflow` because it contains the definitions of the workflow steps (so far only one).

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
@egparedes I have now updated the PR with the change for unique map/tasklet names, which makes debugging the SDFG transformations easier.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,360 @@
...
+    temp_name, _ = sdfg.add_temp_transient(
+        field_shape, dace_fieldview_util.as_dace_type(node_type), offset=field_offset
+    )
+    field_node = state.add_access(temp_name)
+    field_type = ts.FieldType(field_dims, node_type)
+
+    return field_node, field_type
+
+
+def visit_AsFieldOp(
```
egparedes, line 149:
```suggestion
def translate_as_field_op(
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,360 @@
...
+        mx,
+        field_node,
+        src_conn=last_node_connector,
+        memlet=dace.Memlet(data=field_node.data, subset=",".join(output_subset)),
+    )
+
+    return [(field_node, field_type)]
+
+
+def visit_Cond(
```
egparedes, line 231:
```suggestion
def translate_cond(
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,360 @@
...
+        false_state.add_nedge(
+            false_br_node,
+            false_br_output_node,
+            dace.Memlet.from_array(data_name, desc),
+        )
+
+    return output_nodes
+
+
+def visit_SymbolRef(
```
egparedes, line 308:
```suggestion
def translate_symbol_ref(
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -0,0 +1,360 @@
...
+            sym_node,
+            None,
+            dace.Memlet(data=sym_node.data, subset="0"),
+        )
+
+    return [(sym_node, data_type)]
+
+
+if TYPE_CHECKING:
+    # Use type-checking to assert that all visitor functions implement the `PrimitiveTranslator` protocol
```
egparedes, line 355:
```suggestion
    # Use type-checking to assert that all translator functions implement the `PrimitiveTranslator` protocol
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_python_codegen.py
```diff
@@ -0,0 +1,128 @@
...
+        raise NotImplementedError(f"Unexpected 'FunCall' node ({node}).")
+
+
+"""
+Specialized visit method for symbolic expressions.
+
+Returns:
+    A string containing the Python code corresponding to a symbolic expression
+"""
+get_source = PythonCodegen.apply
```
egparedes, line 128: It's not intuitive, but the docstring goes after the definition not before
```suggestion
get_source = PythonCodegen.apply
"""
Specialized visit method for symbolic expressions.

Returns:
    A string containing the Python code corresponding to a symbolic expression
"""
```

src/gt4py/next/program_processors/runners/dace_fieldview/workflow.py
egparedes, line 1: Is this module really needed? I mean, to me it looks like this function should be part of the `gtir_to_sdfg` module...

egparedes (commented):
Except for a couple of renaming and syntactic issues, it looks good for me.

src/gt4py/next/program_processors/runners/dace_fieldview/workflow.py
edopao: Agreed, it makes sense.

egparedes (approved):
LGTM

edopao:
cscs-ci run default

PR #1498, opened 2024-03-15
feat[cartesian]: gtc `cuda` backend deprecation

FlorianDeconinck:
## Description

GTC `cuda` backend was made available a few years ago for AI2 team research. It has been kept updated but a recent poll shows that it is not in use. Recent new features break the backend and we propose here to hard deprecate it rather than keep spending time maintaining it.

`GT4PY_GTC_CUDA_USE=1` can be used to force the use of the backend, but will warn that any feature from February 2024 are not available/not tested.

## Requirements

- [x] All fixes and/or new features come with corresponding tests.
- [ ] Important design decisions have been documented in the approriate ADR inside the [docs/development/ADRs/](docs/development/ADRs/Index.md) folder.

PR #1541, opened 2024-05-03
ci[cartesian]: env var on CSCS-CI to suppress cuda backend deprecation error

DropD:
## Description

Suppress the deprecation error for cartesian "CUDA" backend on CSCS-CI.

The following open questions remain (should be answered for a more long-term solution):

- Why is there a deprecation **error**, not a **warning** by default (warning seems to be the convention in GT4Py)?
- Why did CSCS-CI not fail on #1498, only on main after merging?
- Why does only one test fail (`test_bad_layout_warns[cuda]`)? Is the CUDA backend otherwise untested now?

## Requirements

### DIFF
--- a/ci/cscs-ci.yml
+++ b/ci/cscs-ci.yml
@@ -96,6 +96,7 @@ test py38:
     SLURM_TIMELIMIT: 120
     NUM_PROCESSES: auto
     VIRTUALENV_SYSTEM_SITE_PACKAGES: 1
+    GT4PY_GTC_ENABLE_CUDA: 1
     <<: *py38
 
 test py39:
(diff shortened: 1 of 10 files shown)


FlorianDeconinck:
I should have seen this issue in my PR, I am not sure why. Apologies.

`cuda` backend on cartesian not being tested is not an issue, considering we established it wasn't used by anyone when we poked the main users. We keep in there for silent users, if they exists.

The error over warning, comes from the fact that `cuda` is now going to misbehave on new features, therefore it shouldn't be used.

FlorianDeconinck (approved):

DropD:
> I should have seen this issue in my PR, I am not sure why. Apologies.
> 
> `cuda` backend on cartesian not being tested is not an issue, considering we established it wasn't used by anyone when we poked the main users. We keep in there for silent users, if they exists.
> 
> The error over warning, comes from the fact that `cuda` is now going to misbehave on new features, therefore it shouldn't be used.

 I checked and the CSCS-CI did pass on your PR before merging. I don't understand why at the moment.

PR #1542, opened 2024-05-10
ci: define pytest addopts settings as env var in tox config

egparedes:
Move pytest addopts setting from pytest config in pyproject.toml to an enviroment setting in tox to keep pytest CLI invocation clean during local development.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -233,10 +233,6 @@ module = 'gt4py.next.iterator.runtime'
 [tool.pytest]
 
 [tool.pytest.ini_options]
-addopts = "--instafail"
-filterwarnings = [
-  "ignore::UserWarning"
-]
 markers = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -233,10 +233,6 @@ module = 'gt4py.next.iterator.runtime'
 [tool.pytest]
 
 [tool.pytest.ini_options]
-addopts = "--instafail"
-filterwarnings = [
-  "ignore::UserWarning"
```
DropD, line 238: Is there a better way to ignore the embedded backend warnings? They are completely useless (not only on the CI) and they locally on most terminal emulators on default settings they will push the actual failures out of the terminal history.

DropD: If we could instead make sure that warning is only emitted max once per run, that would be preferrable

PR #1543, opened 2024-05-14
build: update frozen dependencies

havogt:
Additional: fix a generator in pretty_printer after mypy update showed the error.

### DIFF
--- a/src/gt4py/next/iterator/transforms/inline_lifts.py
+++ b/src/gt4py/next/iterator/transforms/inline_lifts.py
@@ -138,7 +138,7 @@ class InlineLifts(
         #: when we see that it is not required.
         INLINE_LIFTED_ARGS = 16
 
-    predicate: Callable[[ir.Expr, bool], bool] = lambda _1, _2: True  # [lambda-assignment]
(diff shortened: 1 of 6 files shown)


src/gt4py/next/iterator/transforms/inline_lifts.py
```diff
@@ -138,7 +138,7 @@ class Flag(enum.IntEnum):
         #: when we see that it is not required.
         INLINE_LIFTED_ARGS = 16
 
-    predicate: Callable[[ir.Expr, bool], bool] = lambda _1, _2: True  # noqa: E731 [lambda-assignment]
+    predicate: Callable[[ir.Expr, bool], bool] = lambda _1, _2: True  # [lambda-assignment]
```
egparedes, line 141: Why is the `noqa` not longer needed? In any case, the comment with the error code should be removed.
```suggestion
    predicate: Callable[[ir.Expr, bool], bool] = lambda _1, _2: True
```

egparedes (approved):
I approve it but there is a leftover from a partially removed `noqa` comment that should be completely removed and the PR title description should include a note about the changes in the pretty printer.

PR #1545, opened 2024-05-23
feat[next]: AOT toolchain

DropD:
## Description

Partial implementation of https://hackmd.io/3EMvJ7zJQPepDd18qDU3sg

Missing:
- Integrate transforms and compilation steps into one workflow. This PR does prepare the ground for such a design change however.

### Background:
The `gt4py.next` machinery to transform DSL functions into executable code has so far had unrestricted access to the arguments the user wished to pass to the executable function, including full fields and their contents, as well as connectivity tables.  In other words, only JIT compilation was possible, even though in principle AOT compilation should also work.

In principle this should not be necessary, and for most of the reusable, pluggable steps in this machinery this access was restricted in a previous PR. However, the restrictions were introduced on a per-step level to make the changes easier to integrate.

In order to get pre-compiled programs, `icon4py` has used a workaround to store the result of the first JIT compilation.

### Changes
This PR makes introduces the restriction that the machinery should only have access to the types of field and scalar arguments, and everything but the connectivity tables for connectivityes on a toolchain level. With the exception that one single ITIR pass still requires access to connectivity tables, however it has been tested that nothing else does.

This also enabled the following design changes:
- The field operator / program stages are now decoupled from the args and kwargs, which means less of them are required
- The transformation steps could be simplified, because they only need to deal with the part of the data they actually require and an adapter can ensure interoperability
- "optional" workflow steps have been replaced with step orders that depend on input type (see `next.otf.workflow.MultiWorkflow` and `next.backend.Transforms` for details.
- `next.backend.Backend` now requires a `ModularExecutor` and will call it's `.executor.otf_workflow` directly and not `.executor.__call__`. This is what prepares for unifying transform and compile workflows. This also means that any custom code in `.executor.__call__` is not going to be called by backends. For new `ProgramExecutors` it is highly recommended to put all such code into their `.otf_workflow.decorate` step. The alternative is to only be usable with a custom backend.
- `decorator.FrozenProgram` JIT compiles on first call and subsequently skips the whole toolchain. It is accessible via `@gtx.program(frozen=True, backend=...) or `decorator.Program.freeze()`. It is in principle also compatible with field operators.

### Required steps before merging
- [x] test with `icon4py` (passed: https://github.com/C2SM/icon4py/pull/513) and add required changes

### Future steps
It would be convenient to create a nice API to automatically create compile time arguments from type annotations to DSL functions. But this should probably be done in a follow-up PR.


src/gt4py/next/backend.py
```diff
@@ -26,143 +27,138 @@
...
-            "past_transform_args",
-            "past_to_itir",
-        ]
+INPUT_DATA: typing.TypeAlias = DSL_FOP | FOP | DSL_PRG | PRG | IT_PRG
+INPUT_PAIR: typing.TypeAlias = workflow.DataArgsPair[INPUT_DATA, ARGS | CARG]
 
 
-DEFAULT_FIELDOP_TRANSFORMS = FieldopTransformWorkflow()
+@dataclasses.dataclass(frozen=True)
+class Transforms(workflow.MultiWorkflow[INPUT_PAIR, stages.AOTProgram]):
```
DropD, line 58: @tehrengruber Could you please review the new "Transforms" toolchain with a view to the tradeoff between "pluggability" of steps (the ability to combine them without copy-pasting glue code) and what's necessary to achieve it? I believe that in this version the tradeoff is much more favorable than previously.

src/gt4py/next/backend.py
```diff
@@ -26,143 +27,138 @@
...
-                program_call, kwargs=program_call.kwargs | {"offset_provider": offset_provider}
-            )
-        else:
-            program_call = self.transforms_prog(workflow.InputWithArgs(program, args, kwargs))
-        self.executor(program_call.program, *program_call.args, **program_call.kwargs)
+        if not isinstance(program, IT_PRG):
+            args, kwargs = signature.convert_to_positional(program, *args, **kwargs)
+        self.jit(program, *args, **kwargs)(*args, **kwargs)
+
+    def jit(self, program: INPUT_DATA, *args: Any, **kwargs: Any) -> stages.CompiledProgram:
```
DropD, line 152: @samkellerhals Could you please review the new `.jit` and `.aot` methods whether they are useful enough for the `icon4py` usecase or if you see something missing (which we might implement now or in a follow-up PR).

src/gt4py/next/iterator/embedded.py
```diff
@@ -150,36 +150,6 @@ def _dace_descriptor(self) -> NoReturn:  # type: ignore[misc]
     __descriptor__ = _dace_descriptor
 
 
-@dataclasses.dataclass(frozen=True)
```
DropD, line 153: Moved to `next.otf.arguments`, it seemed in better company there (more cohesion).

src/gt4py/next/ffront/signature.py
```diff
@@ -0,0 +1,138 @@
...
+def signature_from_foast_stage(func: ffront_stages.FoastOperatorDefinition) -> inspect.Signature:
+    return make_signature(func.foast_node)
+
+
+@make_signature.register
+def signature_from_past_stage(func: ffront_stages.PastProgramDefinition) -> inspect.Signature:
+    return make_signature(func.past_node)
+
+
+def convert_to_positional(
```
DropD, line 88: @tehrengruber I would also particularly like your opinion on this. Do you think it could / should be merged with something else? Do you see a corner case where this might fail (which needs to be covered with more tests)?

src/gt4py/next/iterator/ir.py
```diff
@@ -203,6 +203,7 @@ class FencilDefinition(Node, ValidatedSymbolTableTrait):
     function_definitions: List[FunctionDefinition]
     params: List[Sym]
     closures: List[StencilClosure]
+    implicit_domain: bool = False
```
DropD, line 206: This was added to make it possible for backends to decide whether their `decorate` step needs to generate size arguments. I'm glad to talk about better ways, though preferably in terms of a reference implementation and as a topic for a follow-up PR.

src/gt4py/next/otf/compilation/compiler.py
```diff
@@ -85,10 +85,17 @@ def __call__(
...
+        compiled_prog = getattr(
             importer.import_from_path(src_dir / new_data.module), new_data.entry_point_name
         )
 
+        @dataclasses.dataclass(frozen=True)
+        class Wrapper(stages.ExtendedCompiledProgram):
+            implicit_domain: bool = inp.program_source.implicit_domain
+            __call__: stages.CompiledProgram = compiled_prog
+
+        return Wrapper()
```
DropD, line 97: This is to propagate the information whether the compiler needs to put size args generation into it's decorate step (if `implicit_domain == True`) or not (in which case explicit domain size args are part of the program's signature).

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -46,15 +46,32 @@ def convert_arg(arg: Any) -> Any:
...
+        offset_provider: dict[str, common.Connectivity | common.Dimension],
+        out: Any = None,
     ) -> None:
+        if out is not None:
+            args = (*args, out)
         converted_args = [convert_arg(arg) for arg in args]
         conn_args = extract_connectivity_args(offset_provider, device)
-        return inp(*converted_args, *conn_args)
+        return inp(
+            *converted_args, *(iter_size_args(args) if inp.implicit_domain else ()), *conn_args
```
DropD, line 73: This is where the the `implicit_domain` flag is consumed and size arguments are conditionally generated in the GTFN backend.

src/gt4py/next/type_system/type_translation.py
```diff
@@ -28,6 +28,16 @@
 from gt4py.next.type_system import type_info, type_specifications as ts
 
 
+class SizeArg:
```
DropD, line 31: This is a compile-time standin for a dimension size argument (as used in the implicit domain case during AOT transforms). For reviewers: let me know if you think this should move to `next.otf.arguments`.

src/gt4py/next/otf/arguments.py
```diff
@@ -0,0 +1,175 @@
...
+        gt_type = type_translation.from_value(value)
+        match gt_type:
+            case ts.TupleType():
+                return tuple(cls.from_concrete(element) for element in value)
+            case _:
+                return cls(gt_type)
+
+
+@dataclasses.dataclass(frozen=True)
+class CompileTimeConnectivity(common.Connectivity):
```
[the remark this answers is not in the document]
kotsaloscv: @DropD The class is ok and the location as well.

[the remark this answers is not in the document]
DropD: Thanks!

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -204,8 +206,12 @@ def convert_args(
     on_gpu = True if device == core_defs.DeviceType.CUDA else False
 
     def decorated_program(
```
edopao, line 208: This function is where the PR could have an impact. Question is if - with this PR - `decorated_program` will still be called every time the program is run, even when the program is AOT-compiled and cached?

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -204,8 +206,12 @@ def convert_args(
...
-        *args, offset_provider: dict[str, common.Connectivity | common.Dimension]
+        *args, offset_provider: dict[str, common.Connectivity | common.Dimension], out: Any = None
     ):
+        if out is not None:
+            args = (*args, out)
+        if len(sdfg.arg_names) > len(args):
+            args = (*args, *gtfn.iter_size_args(args))
         if sdfg_program._lastargs:
```
edopao, line 215: For everybody's knowledge, `sdfg_program._lastargs` is the list of program arguments in the order expected by dace. I use the array `inp.sdfg_arg_position` to map the GT4Py position to the position in the DaCe expected arguments list. Therefore, as I verbally mentioned once or twice, the dace backend will reorder the program arguments a second time. Is it possible or desirable to let the backend setup the expected order?
Besides, as you can see in the lines below, we distinguish scalar arguments from field arguments: scalar arguments are written always, while field arguments are written only when the field is reallocated. Is there a way (an execution path) to easily detect when the fields are re-allocated?

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -46,15 +46,32 @@ def convert_arg(arg: Any) -> Any:
         return arg
 
 
+def iter_size_args(args: tuple[Any, ...]) -> Iterator[int]:
+    for arg in args:
```
samkellerhals, line 50: A short docstring would be nice explaining that we extract size arguments when the domain is implicit and else do not (or a comment in `convert_args`).

src/gt4py/next/backend.py
samkellerhals: `.jit` is definitely useful in the context of the `CachedProgram` in icon4py. I am not sure what the intended use of `.aot` is as a standalone method, maybe you could help me understand. Also it would be useful if GT4Py had a class similar to the `CachedProgram` so that we can use that instead in icon4py. Another option could be to introduce a keyword argument in the `gtx.program` decorator e.g. `frozen=true`.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -204,8 +206,12 @@ def convert_args(
     on_gpu = True if device == core_defs.DeviceType.CUDA else False
 
     def decorated_program(
-        *args, offset_provider: dict[str, common.Connectivity | common.Dimension]
+        *args, offset_provider: dict[str, common.Connectivity | common.Dimension], out: Any = None
     ):
+        if out is not None:
+            args = (*args, out)
+        if len(sdfg.arg_names) > len(args):
+            args = (*args, *gtfn.iter_size_args(args))
```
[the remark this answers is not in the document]
DropD: Thanks for catching this. I had meant to move `iter_size_args` to a neutral location.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
DropD: Yes, it is the only way to provide the same interface. Without this, the user would have to
- not pass "out" as a kwarg to programs generated from field operators
- pass implicit domain size arguments manually

DropD: Functionality could be added to make it possible to circumvent if the user can guarantee the above.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
DropD: If we can hand over recognizing the generated "out" keyword argument as well as the implicit domain size arguments to DaCe in some way, it would probably be beneficial. I don't know whether that's possible though.

src/gt4py/next/backend.py
DropD: `.aot` could be used, together with reading the argument types from the DSL type hints, to compile ahead-of-time. This will only become generally feasible if the offset provider is already given to the decorator (or, in future, the compile-time offset provider). Except in the cartesian case, where it could be done even now.

This would represent the absolute minimum of overhead, because no extra data structure is needed to trigger compilation on first call. Imagine in

```python
@gtx.program(precompile=True, offset_provider=...)
def foo(...) -> ...:
    ...
```

What you'd get stored as `foo` would already be the compiled program.

DropD: Alternatively, if you have some "initialization step", that could call `backend.aot()` on all the programs.

DropD: I added a `decorator.FrozenProgram` with an flag in `decorator.program`, as well as a `decorator.Program.freeze` method.

src/gt4py/next/otf/arguments.py
```diff
@@ -0,0 +1,196 @@
...
+from gt4py.next import common
+from gt4py.next.otf import workflow
+from gt4py.next.type_system import type_specifications as ts, type_translation
+
+
+DATA_T = typing.TypeVar("DATA_T")
+
+
+@dataclasses.dataclass(frozen=True)
+class JITArgs:
```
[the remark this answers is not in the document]
DropD: Do you have any suggestions?

src/gt4py/next/ffront/decorator.py
```diff
@@ -180,39 +189,95 @@ def _all_closure_vars(self) -> dict[str, Any]:
...
+        return self.__class__(
+            program=dataclasses.replace(self.program, grid_type=grid_type), backend=self.backend
+        )
+
+    def jit(
+        self, *args: Any, offset_provider: dict[str, Dimension | Connectivity], **kwargs: Any
+    ) -> stages.CompiledProgram:
+        if self.backend is not None:
+            return self.backend.jit(self.program, *args, offset_provider=offset_provider, **kwargs)
+        raise ValueError("Can not JIT-compile programs without backend (embedded execution).")
```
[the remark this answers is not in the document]
DropD: I moved the check into the `__post_init__` method. No need for the assertion.

src/gt4py/next/ffront/decorator.py
```diff
@@ -180,39 +189,95 @@ def _all_closure_vars(self) -> dict[str, Any]:
...
 
-    def __call__(self, *args, offset_provider: dict[str, Dimension], **kwargs: Any) -> None:
+    def __call__(
+        self, *args: Any, offset_provider: dict[str, Dimension | Connectivity], **kwargs: Any
+    ) -> None:
         if self.backend is None:
             warnings.warn(
                 UserWarning(
-                    f"Field View Program '{self.itir.id}': Using Python execution, consider selecting a perfomance backend."
+                    f"Field View Program '{self.definition_stage.definition.__name__}': Using Python execution, consider selecting a perfomance backend."
```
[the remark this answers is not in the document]
DropD: Please explain. This is in `Program`, not in `FieldOperator`. If you meant `ProgramFromPast`, that class overrides `.__call__()`.

docs/user/next/advanced/HackTheToolchain.md
```diff
@@ -126,4 +119,12 @@ class PartiallyModularDiamond(
...
-````
+```
+
+```python
+
+```
+
+```python
+
+```
```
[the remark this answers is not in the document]
DropD: I had to use a modified diff, because the suggestion here would have left the last code block unterminated.

src/gt4py/next/otf/workflow.py
```diff
@@ -267,24 +287,61 @@ def skip_condition(self, inp: StartT) -> bool:
         raise NotImplementedError()
 
 
-@dataclasses.dataclass
-class InputWithArgs(Generic[StartT]):
-    data: StartT
-    args: tuple[Any]
-    kwargs: dict[str, Any]
+@dataclasses.dataclass(frozen=True)
+class DataArgsPair(Generic[DataT, ArgT]):
```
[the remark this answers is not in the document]
DropD: Heads up, this will be a slightly bigger refactoring than a simple rename, because I would like to keep the data sctructures in `otf.workflow` usage-agnostic. Or in terms of another comment on this PR "named by what they are, not by the context they are used in".

src/gt4py/next/otf/arguments.py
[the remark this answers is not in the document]
tehrengruber: How about
`JITArgs` -> `Args` (as used in `backend.py` anyway) or `ProgramArgs`
`CompileTimeArgs` -> `ArgsSpec` (inspired by `TypeSpec`), `ProgramArgsSpec` or `ProgramSignature`
`CompileTimeArg` -> class is superfluous altogether right now, it is just a TypeSpec.

src/gt4py/next/ffront/decorator.py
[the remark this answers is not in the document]
tehrengruber: Not sure what happened here, but this indeed looks correct.

tehrengruber (approved):

PR #1546, opened 2024-05-29
Add Gitlab CI/CD targets for gh200

edopao:
This PR adds the CI/CD configuration for GH200 nodes, while keeping the currrent configuration for x86_64+CUDA on PizDaint.

Some differences between GH200 (on Todi vCluster) and x86_64+CUDA (on PizDaint):

- CUDA v11.2.2 / CUDA ARCH=60 on PizDaint vs. CUDA v12.4.1 / CUDA ARCH=90 on GH200
- Support for Python 3.8, 3.9, 3.10, 3.11 on x86_64 Ubuntu base image, while Python 3.8 is not supported on the ARM base image for GH200
- JAX dependency updated from v0.4.13 to v0.4.18 because this is the minimum version available on the ARM base image
- A compiler is allowed to choose if `char` is signed or unsigned. The Python bindings in GT4Py cartesian rely on the signed representation, which was the default for the compiler on the x86_64 Ubuntu base image. This behavior is not the default on ARM base image, so we have to enforce it with the flag `-fsigned-char`


edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

ci/base_cuda12.Dockerfile
```diff
@@ -0,0 +1,57 @@
+FROM docker.io/nvidia/cuda:12.5.0-devel-ubuntu20.04
```
havogt, line 1: maybe with something like
```suggestion
ARG CUDA_VERSION=12.5.0
FROM docker.io/nvidia/cuda:$CUDA_VERSION-devel-ubuntu20.04
```
we can avoid the duplication

edopao: Good idea, thanks

ci/base.Dockerfile
```diff
@@ -53,5 +54,5 @@ RUN pyenv update && \
 
 ENV PATH="/root/.pyenv/shims:${PATH}"
 
-
-RUN pip install --upgrade pip setuptools wheel tox cupy-cuda11x==12.3.0
+ARG CUPY_PACKAGE=cupy-cuda12x
+RUN pip install --upgrade pip setuptools wheel tox ${CUPY_PACKAGE}==12.3.0
```
havogt, line 58:
```suggestion
RUN pip install --upgrade pip setuptools wheel tox
```
Can you try if this works? We install it also in tox by specifying `SUBVARIANT: [-cuda11x]`

edopao: It doesn't work, some new issues appear. I suggest to do this change as a separate PR.

edopao:
cscs-ci run default

src/gt4py/cartesian/backend/pyext_builder.py
```diff
@@ -82,6 +82,9 @@ def get_gt_pyext_build_opts(
             "-ftemplate-depth={}".format(gt_config.build_settings["cpp_template_depth"]),
             "-fvisibility=hidden",
             "-fPIC",
+            # A compiler is allowed to choose if `char` is signed or unsigned. We forse the signed behavior
```
havogt, line 85:
```suggestion
            # A compiler is allowed to choose if `char` is signed or unsigned. We force the signed behavior
```

requirements-dev.txt
```diff
@@ -76,8 +76,10 @@ iniconfig==2.0.0          # via -c constraints.txt, pytest
 ipykernel==6.29.4         # via -c constraints.txt, nbmake
 ipython==8.12.3           # via -c constraints.txt, ipykernel
 isort==5.13.2             # via -c constraints.txt, -r requirements-dev.in
-jax[cpu]==0.4.13          # via -c constraints.txt, gt4py (pyproject.toml)
-jaxlib==0.4.13            # via -c constraints.txt, jax
+jax==0.4.13 ; python_version < "3.9"  # via -c constraints.txt, gt4py (pyproject.toml)
+jax==0.4.18 ; python_version >= "3.9"  # via -c constraints.txt, gt4py (pyproject.toml)
+jaxlib==0.4.13 ; python_version < "3.9"  # via -c constraints.txt, jax
+jaxlib==0.4.18 ; python_version >= "3.9"  # via -c constraints.txt, jax
```
havogt, line 82: Was this a manual change?  It will be overwritten in the next run of updating dependencies.
Probably need something like
```
jax-cpu = ['jax[cpu]>=0.4.13,<0.4.18;python_version<"3.9"', 'jax[cpu]>=0.4.18; python_version>="3.9"']
```
in pyproject.toml, but the update script will not propagate both entries properly.

ci/cscs-ci.yml
```diff
@@ -118,3 +163,61 @@ test py310:
...
+      VARIANT: [-nomesh, -atlas]
+      SUBVARIANT: [-cuda12x, -cpu]
+  before_script:
+  # TODO: remove start of CUDA MPS daemon once CI-CD can handle CRAY_CUDA_MPS
+  - CUDA_MPS_PIPE_DIRECTORY="/tmp/nvidia-mps" nvidia-cuda-mps-control -d
+  variables:
+    # Grace-Hopper gpu architecture is not enabled by default in CUDA build
+    CUDAARCHS: "90"
+    # limit test parallelism to avoid "OSError: too many open files"
+    NUM_PROCESSES: 32
```
havogt, line 184: Do you know where the problem is coming from? Is this something we should report as feedback to the system?

havogt (changes requested):
I don't know how to solve the version constraints problem...

requirements-dev.txt
havogt: Actually we don't need jax for python < 3.10, so we should add 

```
jax-cpu = ['jax[cpu]>=0.4.13; python_version>="3.10"']
```
(or whatever other minimum version we need in jax)

ci/cscs-ci.yml
edopao: The problem was coming from the cmake build stage included in each test case. If we run pytest using the `auto` parallelism, that will run 288 tests in parallel, each test triggering a GT4Py build. I didn't look at this issue in depth, I just thought that we could limit this parallelism.

edopao: I can remove this limitation and see it the same issue still appears.

havogt: No, I think makes sense.

requirements-dev.txt
edopao: Done, thank you.

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

requirements-dev.txt
```diff
@@ -65,19 +66,17 @@ fonttools==4.53.0         # via -c constraints.txt, matplotlib
...
-importlib-metadata==8.0.0  # via -c constraints.txt, build, jax, jupyter-client, sphinx
+importlib-metadata==8.0.0  # via -c constraints.txt, build, jupyter-client, sphinx
 importlib-resources==6.4.0 ; python_version < "3.9"  # via -c constraints.txt, gt4py (pyproject.toml), jsonschema, jsonschema-specifications, matplotlib
 inflection==0.5.1         # via -c constraints.txt, pytest-factoryboy
 iniconfig==2.0.0          # via -c constraints.txt, pytest
-ipykernel==6.29.4         # via -c constraints.txt, nbmake
+ipykernel==6.29.5         # via -c constraints.txt, nbmake
 ipython==8.12.3           # via -c constraints.txt, ipykernel
 isort==5.13.2             # via -c constraints.txt, -r requirements-dev.in
-jax[cpu]==0.4.13          # via -c constraints.txt, gt4py (pyproject.toml)
```
havogt, line 79: We are removing jax from requirements-dev.txt because it's build for python 3.8. For me that's fine as we anyway don't have jax developers atm. Once we drop 3.8 and 3.9 we will have it back. What do you think @egparedes ?

pyproject.toml
```diff
@@ -78,7 +78,7 @@ cuda12 = ['cupy-cuda12x>=12.0']
 dace = ['dace>=0.16.1', 'sympy>=1.9']
 formatting = ['clang-format>=9.0']
 gpu = ['cupy>=12.0']
-jax-cpu = ['jax[cpu]>=0.4.13']
+jax-cpu = ['jax[cpu]>=0.4.18; python_version>="3.10"']
 jax-cuda11 = ['jax[cuda11_pip]>=0.4.13']
 jax-cuda12 = ['jax[cuda12_pip]>=0.4.13']
```
havogt, line 83:
```suggestion
jax-cuda11 = ['jax[cuda11_pip]>=0.4.18; python_version>="3.10"']
jax-cuda12 = ['jax[cuda12_pip]>=0.4.18; python_version>="3.10"']
```

edopao:
cscs-ci run default

edopao:
cscs-ci run default

PR #1548, opened 2024-06-05
feat[next][dace]: Modify name of stride symbols

edopao:
Remove unnecessary unique ID, which makes code generation easier for ICON-DSL bindings.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace_iterator/itir_to_sdfg.py
@@ -46,7 +46,6 @@ from .utility import (
     get_used_connectivities,
     map_nested_sdfg_symbols,
     new_array_symbols,
-    unique_name,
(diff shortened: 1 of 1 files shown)


philip-paul-mueller (approved):
Two underscore have special meaning in DaCe, i.e. `__return`. I am pretty sure that there would not be a collision, but I suggest to you that you prefix the hole thing with something `__gtit` to make sure that it belongs to you.

Nevertheless, I will approve your PR, since it is a suggestion, and there is no need for a second round.

edopao:
@philip-paul-mueller Thanks, I agree with your comment - I was also expecting it, since you are already commented this once. However, the name of the size argument a few lines above (`"__{field_name}_size_{dim_index}"`) has to match the name of the scalar symbol generated by GT4Py. For the strides, there is no such argument, so I could follow your suggestion, but I prefer to follow the same name convention as the size argument. If I see that it stops working, for any reason, I will change it.

PR #1549, opened 2024-06-07
feat[next][dace]: Remove double-buffering for inout fields

edopao:
During integration of dace backend in ICON-DSL it was found that double-buffering for inout fields is not needed: the GT4Py program should guarantee no race condition.


havogt:
Dace guarantees this? Anyway it's correct to remove it, because GT4Py doesn't guarantee it and it would be a user error.

edopao:
No, (plain) dace cannot detect race conditions as far as I know. Maybe some tool for SDFG traversal could do it.

edopao:
I prefer to avoid further changes in ITIR DaCe backend. The new DaCe backend for CombinedIR is under development.

Issue #1553, opened 2024-06-12
over-sized allocation of arrays for dace backend


xyuan:
The original issue comes from the backend c++ code, see for example,

    __state->__0_w1 = new double DACE_ALIGN(64)[((((__J * __K) * (__I - 1)) + (__K * (__J - 1))) + __K)];
    __state->__0_g_rat = new double DACE_ALIGN(64)[((((__J * __K) * (__I - 1)) + (__K * (__J - 1))) + __K)];
    __state->__0_bb = new double DACE_ALIGN(64)[((((__J * __K) * (__I - 1)) + (__K * (__J - 1))) + __K)];
    __state->__0_dd = new double DACE_ALIGN(64)[((((__J * __K) * (__I - 1)) + (__K * (__J - 1))) + __K)];
    __state->__0_bet = new double DACE_ALIGN(64)[((((__J * __K) * (__I - 1)) + (__K * (__J - 1))) + __K)];
    __state->__0_pp = new double DACE_ALIGN(64)[((((__J * __K) * (__I - 1)) + (__K * (__J - 1))) + __K)];
    __state->__0_gam = new double DACE_ALIGN(64)[((((__J * __K) * (__I - 1)) + (__K * (__J - 1))) + __K)];
    __state->__0_aa = new double DACE_ALIGN(64)[((((__J * __K) * (__I - 1)) + (__K * (__J - 1))) + __K)];
    __state->__0_p1 = new double DACE_ALIGN(64)[((((__J * __K) * (__I - 1)) + (__K * (__J - 1))) + __K)];

this bug leads to allocate a larger chunck of memory for array with size of (__I, __J, __K), here __I and __J are the horizontal, and __K is the vertical, when we run large grid size problem, it leads to very poor performance.

the root of this problem comes from the dace data array initialization, which takes shape, strides, and total_size as argument, however, if no total_size is provided, the code calculates the total size of array (see the following):

        if strides is not None and shape is not None and total_size is None:
            # Compute the minimal total_size that could be used with strides and shape
            self.total_size = sum(((shp - 1) * s for shp, s in zip(shape, strides))) + 1
        else:
            self.total_size = total_size or _prod(shape)

To fix this issue, we have to provide the total_size when we call sdfg.add_array function, or we make changes to the data::array constructor. Here is the generated c++ code with this fix,

    __state->__0_w1 = new double DACE_ALIGN(64)[((__I * __J) * __K)];
[...]

PR #1554, opened 2024-06-12
fix array size problem which leads to poor performance

xyuan:
This PR is fix the array size for global and temporary array in the generated c++ code, this bug causes poor performance due to the irregular and over-sized allocation of memory block. see the issues for details of this problem https://github.com/GridTools/gt4py/issues/1553

- PR Title: <type>[<scope>]: <one-line-summary>

    <type>:
        - build: Changes that affect the build system or external dependencies
        - ci: Changes to our CI configuration files and scripts
        - docs: Documentation only changes
        - feat: A new feature
        - fix: A bug fix
        - perf: A code change that improves performance
        - refactor: A code change that neither fixes a bug nor adds a feature
        - style: Changes that do not affect the meaning of the code
        - test: Adding missing tests or correcting existing tests

    <scope>: cartesian | eve | next | storage
    # ONLY if changes are limited to a specific subsytem

- PR Description:

    Description of the main changes with links to appropriate issues/documents/references/...
-->

## Requirements

### DIFF
--- a/src/gt4py/cartesian/gtc/dace/oir_to_dace.py
+++ b/src/gt4py/cartesian/gtc/dace/oir_to_dace.py
@@ -18,6 +18,7 @@ from typing import Dict
 import dace
 import dace.properties
 import dace.subsets
+import dace.data
 
 import gt4py.cartesian.gtc.oir as oir
 from gt4py import eve
@@ -144,14 +145,15 @@ class OirSDFGBuilder(eve.NodeVisitor):
                 dim_strs = [d for i, d in enumerate("IJK") if param.dimensions[i]] + [
                     f"d{d}" for d in range(len(param.data_dims))
                 ]
+                shape = ctx.make_shape(param.name)
+                strides = [dace.symbolic.pystr_to_symbolic(f"__{param.name}_{dim}_stride") for dim in dim_strs]
+                total_size = dace.data._prod([dace.symbolic.pystr_to_symbolic(s) for s in shape])
                 ctx.sdfg.add_array(
                     param.name,
-                    shape=ctx.make_shape(param.name),
-                    strides=[
-                        dace.symbolic.pystr_to_symbolic(f"__{param.name}_{dim}_stride")
-                        for dim in dim_strs
-                    ],
+                    shape=shape,
+                    strides=strides,
                     dtype=data_type_to_dace_typeclass(param.dtype),
+                    total_size=total_size,
                     transient=False,
                     debuginfo=dace.DebugInfo(0),
                 )
@@ -162,14 +164,15 @@ class OirSDFGBuilder(eve.NodeVisitor):
             dim_strs = [d for i, d in enumerate("IJK") if decl.dimensions[i]] + [
                 f"d{d}" for d in range(len(decl.data_dims))
             ]
+            shape = ctx.make_shape(decl.name)
+            strides = [dace.symbolic.pystr_to_symbolic(f"__{decl.name}_{dim}_stride") for dim in dim_strs]
+            total_size = dace.data._prod([dace.symbolic.pystr_to_symbolic(s) for s in shape])
             ctx.sdfg.add_array(
                 decl.name,
-                shape=ctx.make_shape(decl.name),
-                strides=[
-                    dace.symbolic.pystr_to_symbolic(f"__{decl.name}_{dim}_stride")
-                    for dim in dim_strs
-                ],
+                shape=shape,
+                strides=strides,
                 dtype=data_type_to_dace_typeclass(decl.dtype),
+                total_size=total_size,
                 transient=True,
                 debuginfo=dace.DebugInfo(0),
             )
--- a/src/gt4py/cartesian/backend/dace_backend.py
+++ b/src/gt4py/cartesian/backend/dace_backend.py
@@ -206,9 +206,9 @@ def _sdfg_add_arrays_and_edges(
             shape = [f"__{name}_{axis}_size" for axis in axes] + [
                 d for d in field_info[name].data_dims
             ]
-
+            total_size = dace.data._prod([dace.symbolic.pystr_to_symbolic(s) for s in shape])
             wrapper_sdfg.add_array(
-                name, dtype=array.dtype, strides=array.strides, shape=shape, storage=array.storage
+                name, dtype=array.dtype, strides=array.strides, shape=shape, total_size=total_size, storage=array.storage
             )
             if isinstance(origins, tuple):
                 origin = [o for a, o in zip("IJK", origins) if a in axes]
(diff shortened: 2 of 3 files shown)


src/gt4py/cartesian/gtc/dace/oir_to_dace.py
```diff
@@ -162,14 +164,15 @@ def visit_Stencil(self, node: oir.Stencil, **kwargs):
...
+            total_size = dace.data._prod([dace.symbolic.pystr_to_symbolic(s) for s in shape])
             ctx.sdfg.add_array(
                 decl.name,
-                shape=ctx.make_shape(decl.name),
-                strides=[
-                    dace.symbolic.pystr_to_symbolic(f"__{decl.name}_{dim}_stride")
-                    for dim in dim_strs
-                ],
+                shape=shape,
+                strides=strides,
```
FlorianDeconinck, line 173: We should move stride/shape/total_size comptuation in a single function as to remove the code duplication. Probably in `cartesian/backend/dace_backend.py` or on a dedicated memory helper file in `cartesian/gtc/dace/`

xyuan: sure, totally agree

edopao:
I got curious about this PR and I checked the issue #1554. The code snippet reported there:

```
    if strides is not None and shape is not None and total_size is None:
        # Compute the minimal total_size that could be used with strides and shape
        self.total_size = sum(((shp - 1) * s for shp, s in zip(shape, strides))) + 1
    else:
        self.total_size = total_size or _prod(shape)
```

suggests me that for allocation of temporaries we could skip passing the stride symbols, and we would achieve the same result. What do you think?

xyuan:
> I got curious about this PR and I checked the issue #1554. The code snippet reported there:
> 
> ```
>     if strides is not None and shape is not None and total_size is None:
>         # Compute the minimal total_size that could be used with strides and shape
>         self.total_size = sum(((shp - 1) * s for shp, s in zip(shape, strides))) + 1
>     else:
>         self.total_size = total_size or _prod(shape)
> ```
> 
> suggests me that for allocation of temporaries we could skip passing the stride symbols, and we would achieve the same result. What do you think?

from the this code, you are right, but we have to provide stride information for the use in c++, so stride information is necessary in this case.

edopao:
> > I got curious about this PR and I checked the issue #1554. The code snippet reported there:
> > ```
> >     if strides is not None and shape is not None and total_size is None:
> >         # Compute the minimal total_size that could be used with strides and shape
> >         self.total_size = sum(((shp - 1) * s for shp, s in zip(shape, strides))) + 1
> >     else:
> >         self.total_size = total_size or _prod(shape)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > suggests me that for allocation of temporaries we could skip passing the stride symbols, and we would achieve the same result. What do you think?
> 
> from the this code, you are right, but we have to provide stride information for the use in c++, so stride information is necessary in this case.

I see, the stride symbols are used in the cartesian backend in combination with tiling. However, the problem described in #1553 is only for allocation of temporaries, that is arrays with `transient=True`. The non-transient arrays should be allocated externally, so the generated C++ code should not contain memory allocation for such arrays. If this is the case, it is maybe possible to avoid some of the changes, for the arrays with `transient=False`?

src/gt4py/cartesian/gtc/dace/oir_to_dace.py
xyuan: add function in cartesian/gtc/dace/utils.py to calculate the array strides and array total size in order to reduce the duplication of code

src/gt4py/cartesian/gtc/dace/utils.py
```diff
@@ -26,6 +26,14 @@
 from gt4py.cartesian.gtc.passes.oir_optimizations.utils import compute_horizontal_block_extents
 
 
+def array_strides(fld: str, dims: List[str]):
```
FlorianDeconinck, line 29: I know those functions (`array_strides` and `array_total_size`) where lifted from the original code, but they seem very brittle, considering we are using the internal of the dace symbolic parser. 

Can we compute this more cleanly and/or use an existing function within DaCe?

FlorianDeconinck (changes requested):
Good catch and I think it's better to live under a "arrays size are always fully defined in the SDFG, transient or not" paradigm. I'd just would like to search for a better way to compute strides/size that doesn't rely so heavily on internals

FlorianDeconinck:
> > > I got curious about this PR and I checked the issue #1554. The code snippet reported there:
> > > ```
> > >     if strides is not None and shape is not None and total_size is None:
> > >         # Compute the minimal total_size that could be used with strides and shape
> > >         self.total_size = sum(((shp - 1) * s for shp, s in zip(shape, strides))) + 1
> > >     else:
> > >         self.total_size = total_size or _prod(shape)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > suggests me that for allocation of temporaries we could skip passing the stride symbols, and we would achieve the same result. What do you think?
> > 
> > 
> > from the this code, you are right, but we have to provide stride information for the use in c++, so stride information is necessary in this case.
> 
> I see, the stride symbols are used in the cartesian backend in combination with tiling. However, the problem described in #1553 is only for allocation of temporaries, that is arrays with `transient=True`. The non-transient arrays should be allocated externally, so the generated C++ code should not contain memory allocation for such arrays. If this is the case, it is maybe possible to avoid some of the changes, for the arrays with `transient=False`?

My understanding is that the code doesn't allocate as much as it describes fully to the SDFG the size of the arrays. I think it's better to give as much as information as DaCe as we hold rather than the inverse. DaCe should be taking care of not double allocating arrays that are externals.

src/gt4py/cartesian/gtc/dace/utils.py
xyuan: I didn't see the parse strides function in DaCe internal, maybe we could keep the original in the code?

xyuan:
> > > > I got curious about this PR and I checked the issue #1554. The code snippet reported there:
> > > > ```
> > > >     if strides is not None and shape is not None and total_size is None:
> > > >         # Compute the minimal total_size that could be used with strides and shape
> > > >         self.total_size = sum(((shp - 1) * s for shp, s in zip(shape, strides))) + 1
> > > >     else:
> > > >         self.total_size = total_size or _prod(shape)
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > suggests me that for allocation of temporaries we could skip passing the stride symbols, and we would achieve the same result. What do you think?
> > > 
> > > 
> > > from the this code, you are right, but we have to provide stride information for the use in c++, so stride information is necessary in this case.
> > 
> > 
> > I see, the stride symbols are used in the cartesian backend in combination with tiling. However, the problem described in #1553 is only for allocation of temporaries, that is arrays with `transient=True`. The non-transient arrays should be allocated externally, so the generated C++ code should not contain memory allocation for such arrays. If this is the case, it is maybe possible to avoid some of the changes, for the arrays with `transient=False`?
> 
> My understanding is that the code doesn't allocate as much as it describes fully to the SDFG the size of the arrays. I think it's better to give as much as information as DaCe as we hold rather than the inverse. DaCe should be taking care of not double allocating arrays that are externals.

I bought up this issue to DaCe develop community to see whether they want to have a fix in DaCe, if so, we can close this PR and keep our application code same, https://github.com/spcl/dace/issues/1628 @FlorianDeconinck

xyuan:
close this pr, as it is very easy to get it right in DaCe

PR #1562, opened 2024-06-26
feat[next][dace]: Use DaCe SDFG fastcall

edopao:
Starting from DaCe v0.16, we can use the SDFG fastcall API in GT4Py-next DaCe backend. It provides a fast way to call the SDFG, which bypasses parsing and construction of the arguments list. This is useful when the GT4Py program is invoked multiple times, with the same field buffers as arguments (not necessarily the same content in the buffer). This PR provides the logic to use the fastcall API while allowing for override of scalar arguments.


edopao:
cscs-ci run default

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -115,15 +119,27 @@ class Meta:
         model = DaCeTranslator
 
 
+class CompiledDaceProgram(stages.CompiledProgram):
+    sdfg_program: dace.CompiledSDFG
+    sdfg_scalar_args: dict[str, int]  # map arg to its position in program ABI
```
philip-paul-mueller, line 124: I am not sure why all scalar arguments should be integers?
For example the timestep size is a scalar but not an integer (usually).

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -164,17 +199,41 @@ class Meta:
...
     def decorated_program(
         *args, offset_provider: dict[str, common.Connectivity | common.Dimension]
     ):
+        if sdfg_program._lastargs:
+            # The scalar arguments should be replaced with the actual value; for field arguments,
+            # the data pointer should remain the same otherwise fast-call cannot be used and
+            # the args list needs to be reconstructed.
+            use_fast_call = True
+            scalar_args_table: dict[str, ctypes._SimpleCData] = {}
+            for pos, (param, arg) in enumerate(zip(sdfg.arg_names, args), start=1):
```
philip-paul-mueller, line 219: I am very sure that this is not fully correct.
To use fast call you have to bring the arguments in the same order as the C interface, which is `<pointer-arguments>, <scalar.arguments>` each in alphabetical order. `sdfg.arg_names` is the names of the positional arguments, which can have a different order.

Furthermore, I do not see where you construct the arguments to the SDFG's init function.
As a side note here, calling the SDFG from Python uses a little known fact in the C standard:
```C
int foo();
{
        return 42;
}
```
In `C++` this is a function that does not take any arguments, however, in `C` the function takes an unknown number of arguments.
This works because all arguments are passed on the stack and it is the callers responsibility to put them there.
Thus you are not get an error.

The simplest way I know is to use `CompiledSDFG._construct_args()` which will do everything for you and update the internals.
The second possibility is to call `SDFG.arg_list()` which will give you the arguments in the corrected order.
Another option is the first time to simply use the normal `CompiledSDFG.__call__()` function, which will do it for you.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -164,17 +199,41 @@ class Meta:
...
+                if isinstance(arg, common.Field):
+                    storage_type = sdfg.arrays[param].storage
+                    assert isinstance(sdfg_program._lastargs[0][pos], ctypes.c_void_p)
+                    if sdfg_program._lastargs[0][pos].value != get_array_interface_ptr(
+                        arg.ndarray, storage_type
+                    ):
+                        use_fast_call = False
+                        break
+                else:
+                    scalar_args_table[param] = arg
```
philip-paul-mueller, line 229: Are you sure that this works?
Are you sure that this is always a correct ctypes type and not some plain Python `int`?

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -164,17 +199,41 @@ class Meta:
...
         *args, offset_provider: dict[str, common.Connectivity | common.Dimension]
     ):
+        if sdfg_program._lastargs:
+            # The scalar arguments should be replaced with the actual value; for field arguments,
+            # the data pointer should remain the same otherwise fast-call cannot be used and
+            # the args list needs to be reconstructed.
+            use_fast_call = True
+            scalar_args_table: dict[str, ctypes._SimpleCData] = {}
+            for pos, (param, arg) in enumerate(zip(sdfg.arg_names, args), start=1):
+                if isinstance(arg, common.Field):
```
philip-paul-mueller, line 220: Here you also have to check the underlying SDFG type in the signature.

philip-paul-mueller (changes requested):
There are some issues, especially regarding where to get the sources from and the proper argument names.
However, despite its hanky nature it looks promising.

I would also recommend to add some tests, where you explicitly inspect the arguments if they are correctly constructed.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
edopao: This integer value represents the position of a scalar arguments in the ABI call signature.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -155,7 +171,26 @@ def __call__(
...
+            itertools.chain(
+                s.free_symbols
+                for x in sdfg.arrays.values()
+                for s in x.strides
+                if not isinstance(s, int)
+            ),
+        )
+        stride_symbols = {str(x) for x in stride_symbols}
+
+        # extract position of scalar arguments from program ABI
```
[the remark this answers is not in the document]
edopao: It still works because the member `strides` of a `Scalar` contains `[1]`, so it is filtered out. But the way you suggest is more clear, I agree.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
edopao: It works because I am using `SDFG.signature_arglist`, which internally calls `SDFG.arg_list`, to retrieve the position of the arguments in the program ABI.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
edopao: I see the point, but how likely it is that we pass the same underlying buffer but with different numpy data type? If we have to account for all cases, it is better to not use fastcall.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
edopao: I see, I should do something like in dace code:
```
                elif not isinstance(arg, (ctypes._SimpleCData, ctypes._Pointer)):
                    newargs[i] = actype(arg)
                else:
                    newargs[i] = arg
```

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
edopao: One part might be incorrect, I see it now. The order of fields in `arg_names` is not necessarily the same as in the program ABI. I will fix this.
My answer applied to the order of scalars (represented as SDFG symbols): for scalar arguments I was already using the ABI order.

edopao:
@philip-paul-mueller Thanks for the review. I have addressed all your change requests, except for the check of the underlying SDFG type for field buffer reuse: if a GT4Py program is called with different field data type, the type system will trigger a recompilation of the program, and therefore it will regenerate the SDFG. So we do not this extra check.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
philip-paul-mueller: As far as I know/see NumPy uses something like pool allocation, as it might reuse memory of arrays that have gone out of scope.
However, you are right it is quite far fetched.

But it should be still better to use fast call, since we are in a bit different setting.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -121,11 +118,23 @@ class Meta:
...
+        sdfg_used_symbols = program.sdfg.used_symbols(all_symbols=False)
 
-    def __init__(self, program, scalar_args):
         self.sdfg_program = program
-        self.sdfg_scalar_args = scalar_args
+        self.sdfg_arg_position = [
+            sdfg_arg_pos_mapping[param]
+            if param in program.sdfg.arrays or param in sdfg_used_symbols
+            else -1
+            for param in program.sdfg.arg_names
```
philip-paul-mueller, line 136: Totally optional:
- Instead of using `-1` to denote invalid (I guess) use `None`, because `-1` could be understood at "at the end".
- And what does `-1` mean anyway, an argument that is neither a symbol or an array?
- You could construct `arg_names` directly in the correct order, see [PR#1374](https://github.com/GridTools/gt4py/pull/1374) for more, this would simplify a lot. As far as I know nothing depends on it, as I was the one who introduced that feature and if you did not use it in the mean time there should be no problems. (But in that case I would add a big fat note for people like me.)

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -115,15 +116,39 @@ class Meta:
         model = DaCeTranslator
 
 
+class CompiledDaceProgram(stages.CompiledProgram):
+    sdfg_program: dace.CompiledSDFG
+    sdfg_arg_position: list[
+        int
+    ]  # map SDFG argument to its position in program ABI, -1 if argument is not used
```
philip-paul-mueller, line 123: I try to never mention formatting and I follow the doctrine "the output of ruff is final", but consider to put the comment on its own line as it looks awful.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
```diff
@@ -115,15 +116,39 @@ class Meta:
         model = DaCeTranslator
 
 
+class CompiledDaceProgram(stages.CompiledProgram):
+    sdfg_program: dace.CompiledSDFG
+    sdfg_arg_position: list[
```
philip-paul-mueller, line 121: Totally optional:
It might be more elegant to use the type `list[str]` thus the position of the argument, i.e. its index, is directly mapped to its name.
Furthermore if you modify `arg_names` then you could use that instead.

philip-paul-mueller (commented):
The current code is a bit complicated and could be simplified if we would construct the `SDFG::arg_names` property differently (see a review comment).
However, I am open to approve the PR without these changes, but I highly recommend them.

Nevertheless, this PR introduces some big internal dependencies on DaCe, thus I think that there should be some unit tests that tests if the SDFG arguments are updated properly, i.e. that a change in the arguments is actually detected.

src/gt4py/next/program_processors/runners/dace_iterator/workflow.py
edopao: I am interested in reducinf the call overhead, here. Looking for the index of an argument, at each SDFG call, is too expensive.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace.py
```diff
@@ -0,0 +1,121 @@
...
+    verify_testee()
+    mock_construct_args.assert_called_once()
+    fast_call_args = mock_fast_call.call_args.args
+
+    # Now modify the scalar arguments, used and unused ones: reuse previous SDFG arguments
+    for i in range(4):
+        a_offset[i] += 1
+        verify_testee()
+        mock_construct_args.assert_not_called()
+        assert mock_fast_call.call_args.args == fast_call_args
```
philip-paul-mueller, line 109: Excuse my ignorance, but why should the stored arguments not been modified?
Below you change the content of an array, this should not update the processed arguments, since "the argument itself" has not changed only its content.

However, the value of scalars is backed into the processed arguments and thus, changing the arguments has to modify the processed arguments, otherwhise the value will never be updated.

Also since one scalar is unused, I would also expect different behaviour there.
This test is also missing for the unused buffer.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # Modify content of current buffer: reuse previous SDFG arguments
+    a[0] += 1
+    verify_testee()
+    mock_construct_args.assert_not_called()
+    assert mock_fast_call.call_args.args == fast_call_args
```
philip-paul-mueller, line 115: Here I agree that the arguments (the address to the memory storing the argument) do not change as you modify the content, which is different than above.

philip-paul-mueller (changes requested):
There is something in the tests I do not understand.
Beside that the PR is ready to merge.

philip-paul-mueller:
BTW: Shape and stride changes are handled because they are always passed as symbols?

edopao:
> BTW: Shape and stride changes are handled because they are always passed as symbols?

Yes, correct.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace.py
```diff
@@ -0,0 +1,163 @@
...
+        new_scalar_args = get_scalar_values_from_sdfg_args(mock_fast_call.call_args.args)
+        diff = np.array(new_scalar_args) - np.array(prev_scalar_args)
+        assert np.count_nonzero(diff) == 1
+
+    def check_scalar_args_all_same(prev_scalar_args):
+        new_scalar_args = get_scalar_values_from_sdfg_args(mock_fast_call.call_args.args)
+        diff = np.array(new_scalar_args) - np.array(prev_scalar_args)
+        assert np.count_nonzero(diff) == 0
+
+    def chack_pointer_args_all_same():
```
philip-paul-mueller, line 128:
```suggestion
    def check_pointer_args_all_same():
```

tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace.py
```diff
@@ -0,0 +1,163 @@
...
+            check_scalar_args_all_same(fast_call_scalar_values)
+
+    # Modify content of current buffer: reuse previous SDFG arguments
+    for buff in (a, unused_field):
+        buff[0] += 1
+        verify_testee()
+        mock_construct_args.assert_not_called()
+        # same arguments tuple object and same content
+        assert mock_fast_call.call_args.args == fast_call_args
+        chack_pointer_args_all_same()
```
philip-paul-mueller, line 156:
```suggestion
        check_pointer_args_all_same()
```

tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace.py
```diff
@@ -0,0 +1,163 @@
...
+            if isinstance(arg, ctypes._Pointer):
+                assert arg == prev
+
+    # Now modify the scalar arguments, used and unused ones: reuse previous SDFG arguments
+    for i in range(4):
+        a_offset[i] += 1
+        verify_testee()
+        mock_construct_args.assert_not_called()
+        assert mock_fast_call.call_args.args == fast_call_args
+        chack_pointer_args_all_same()
```
philip-paul-mueller, line 139:
```suggestion
        check_pointer_args_all_same()
```

philip-paul-mueller (approved):
There is one typo in a function name.
Apart from that LGTM.

PR #1566, opened 2024-07-01
bug[next]: Fix missing visit call in call to cast builtin in ITIR type inference

tehrengruber:
Since `cast_` is a grammar builtin whose return type is given by its second argument it is easy to forget inferring the types of the first argument and its children. This surfaced in icon4py and is fixed here.

### DIFF
--- a/tests/next_tests/unit_tests/iterator_tests/test_type_inference.py
+++ b/tests/next_tests/unit_tests/iterator_tests/test_type_inference.py
@@ -222,6 +222,18 @@ def test_late_offset_axis():
     assert result.type == it_on_e_of_e_type
 
 
+def test_type_inference_in_cast():
+    # since cast_ is a grammer builtin whose return type is given by its second argument it is
+    # easy to forget inferring the types of the first argument and its children. Simply check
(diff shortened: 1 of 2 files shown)


tests/next_tests/unit_tests/iterator_tests/test_type_inference.py
```diff
@@ -222,6 +222,18 @@ def test_late_offset_axis():
...
+def test_cast_first_arg_inference():
+    # since cast_ is a grammar builtin whose return type is given by its second argument it is
+    # easy to forget inferring the types of the first argument and its children. Simply check
+    # if the first argument has a type inferred correctly here.
+    testee = im.call("cast_")(
+        im.plus(im.literal_from_value(1), im.literal_from_value(2)), "float64"
+    )
+    result = itir_type_inference.infer(testee, offset_provider={}, allow_undeclared_symbols=True)
+
+    assert result.args[0].type == int_type
```
edopao, line 234: For completeness, do also `assert result.type == float64_type`

edopao (approved):

PR #1571, opened 2024-07-05
bug[next]: Add assert for OOB access

tehrengruber:
In case a stencil contains an Out-Of-Bounds accesses the derivation of the dtype in ITIR embedded can fail with an obscure error like
```
ValueError: DType 'DType(scalar_type=<class 'numpy.object_'>, tensor_shape=())' not supported.
```
This PR adds an assert to catch this earlier and fail more gracefully.


havogt (commented):
Can you reach that in a valid computation?

tehrengruber:
I think no, that's why I picked an assert. The position is taken from the domain so the result should never be `_UNDEFINED` unless the user wrote something unvalid.

havogt (approved):

PR #1572, opened 2024-07-08
build: update frozen requirements

havogt:
for resolving https://github.com/GridTools/gt4py/security/dependabot/58


egparedes (approved):
ILGTM

PR #1574, opened 2024-07-09
bug[next]: Fix func to past caching

tehrengruber:
A small indentation error in the `FuncToPast` factory prevented caching of the step, which is fixed here. Additionally the hashing algorithm is changed to `id`.

### DIFF
--- a/src/gt4py/next/backend.py
+++ b/src/gt4py/next/backend.py
@@ -106,7 +106,7 @@ class ProgramTransformWorkflow(workflow.NamedStepSequenceWithArgs):
         ffront_stages.ProgramDefinition | ffront_stages.PastProgramDefinition,
         ffront_stages.PastProgramDefinition,
     ] = dataclasses.field(
-        default_factory=lambda: func_to_past.OptionalFuncToPastFactory(cached=False, workflow="123")
+        default_factory=lambda: func_to_past.OptionalFuncToPastFactory(cached=True)
     )
     past_lint: workflow.Workflow[
         ffront_stages.PastProgramDefinition, ffront_stages.PastProgramDefinition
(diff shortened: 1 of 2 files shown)


DropD:
The failures in CI seem unrelated to the problem that caused me to disable caching by default in #1545 [there is a comment explaining it here](https://github.com/GridTools/gt4py/pull/1545/files#diff-c29d58834223753a25911607f39244f0c746ee485436e08121ef6da3fe163d23R74).

It looks to me like this might be python reusing addresses of de-allocated dataclasses for new ones. I am thinking:

1) some test runs, during which a program is compiled. In the process of this a ProgramDefinition stage is created at address `A`
2) the test finishes, the stage's reference count goes to 0
3) eventually the stage is cleaned up
4) another test starts, a new ProgramDefinition stage is created and happens to be allocated at `A` as well.
5) the cache never got deallocated, because there was always a reference to it, so now it already has a result for address `A`, but it is the result for the earlier ProgramDefinition.

This would explain why the failing tests are not the same between GithubActions and the CSCS-CI pipeline. It would explain why running each separately will not typically fail. It would also explain why the failing tests are not the same between CI runs (which they don't seem to be). I would expect random results with tests that are parametrized with more backends having higher probability of failing (simply because they run more often).

If this is the case, then in general it would be unsafe to use `id` as a `hash_function`. It would however be safe in the case where no program is created after another one has gone out of scope. This would be impossible to guarantee in the GT4Py tests without sacrificing test isolation. However, it could be guaranteed in applications of GT4Py, as long as those applications do not have a similar test suite where programs are generated per test method.

tehrengruber:
Not relevant anymore.

PR #1580, opened 2024-07-17
test[next]: Skip all dace tests until new DaCe backend lands (GTIR branch)

tehrengruber:
CC: @edopao

### DIFF
--- a/tests/next_tests/definitions.py
+++ b/tests/next_tests/definitions.py
@@ -97,6 +97,8 @@ class ProgramFormatterId(_PythonObjectIdMixin, str, enum.Enum):
 
 
 # Test markers
+# Special marker that skips all tests. This is not a regular pytest marker, but handled explicitly
+# to avoid needing to mark all tests.
 ALL = "all"
 REQUIRES_ATLAS = "requires_atlas"
 # TODO(havogt): Remove, skipped during refactoring to GTIR
--- a/tests/next_tests/definitions.py
+++ b/tests/next_tests/definitions.py
@@ -149,7 +149,7 @@ COMMON_SKIP_TEST_LIST = [
     (USES_SPARSE_FIELDS_AS_OUTPUT, XFAIL, UNSUPPORTED_MESSAGE),
 ]
 DACE_SKIP_TEST_LIST = COMMON_SKIP_TEST_LIST + [
-    # TODO(tehrengruber): disable all DaCe tests until new version of the backend using GTIR lands
+    # FIXME[#1582](tehrengruber): disable all DaCe tests until new version of the backend using GTIR lands
     (ALL, XFAIL, UNSUPPORTED_MESSAGE),
     (USES_INDEX_FIELDS, XFAIL, UNSUPPORTED_MESSAGE),
     (USES_LIFT_EXPRESSIONS, XFAIL, UNSUPPORTED_MESSAGE),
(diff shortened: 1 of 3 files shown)


tests/next_tests/definitions.py
```diff
@@ -146,6 +149,8 @@ class ProgramFormatterId(_PythonObjectIdMixin, str, enum.Enum):
     (USES_SPARSE_FIELDS_AS_OUTPUT, XFAIL, UNSUPPORTED_MESSAGE),
 ]
 DACE_SKIP_TEST_LIST = COMMON_SKIP_TEST_LIST + [
+    # TODO(tehrengruber): disable all DaCe tests until new version of the backend using GTIR lands
```
egparedes, line 152: Shouldn't we mark this with the issue tracking the temporary changes for the ongoing GTIR work? I propose to use a `FIXME` tag instead of a `TODO` to avoid mistakes. Something like: 
```suggestion
    # FIXME(#1582): disable all DaCe tests until new version of the backend using GTIR lands
```

edopao:
Just for me to understand, what is the reason for this PR? are we going to break compatibility with legacy ITIR-backends? if we skip ITIR-dace why do we keep ITIR-gtfn?

tests/next_tests/unit_tests/conftest.py
```diff
@@ -82,7 +82,7 @@ def program_processor(request) -> tuple[ppi.ProgramProcessor, bool]:
     for marker, skip_mark, msg in next_tests.definitions.BACKEND_SKIP_TEST_MATRIX.get(
         processor_id, []
     ):
-        if request.node.get_closest_marker(marker):
+        if marker == next_tests.definitions.ALL or request.node.get_closest_marker(marker):
```
edopao, line 85: I would prefer a different approach. I would change line 39 to:
`if dace_iterator and False:`

tests/next_tests/integration_tests/feature_tests/ffront_tests/ffront_test_utils.py
```diff
@@ -96,7 +96,7 @@ def exec_alloc_descriptor(request):
     for marker, skip_mark, msg in next_tests.definitions.BACKEND_SKIP_TEST_MATRIX.get(
         backend_id, []
     ):
-        if request.node.get_closest_marker(marker):
+        if marker == next_tests.definitions.ALL or request.node.get_closest_marker(marker):
```
edopao, line 99: I would prefer a different approach. I would change line 56 to:
`if dace_iterator and False:`

edopao (changes requested):

tests/next_tests/integration_tests/feature_tests/ffront_tests/ffront_test_utils.py
tehrengruber: I don't have a strong opinion on this, but you are aware that this means that the tests are not skipped anymore, but simply not executed and hence invisible in the output? No question that this `ALL` thing is ugly, but doing `if dace_iterator and False:` is very hard to find, I honestly completely forgot about this `OPTIONAL_PROCESSORS` altogether.

edopao: Right, the advantage of the `ALL` marker is visibility. Additionally, it is reusable for other backends.
The reason behind the `OPTIONAL_PROCESSORS` is that dace is an optional dependency for GT4Py, so dace tests are not executed when the dace package is not installed in the python environment.

edopao (approved):

PR #1568, opened 2024-07-03
GTIR domain inference

SF-N:
#### Uses TraceShifts to infer the minimal domain of (nested) `as_fieldop`s and SetAt:
Domain inference of let-statements is not implemented yet and will be done in PR #1591
```
expr = as_fieldop(stencil)(as_fieldop(stencil)(inp_field))

domain = cartesian_domain(...)

# new_expr has transformed all as_fieldop(stencil) into as_fieldop(stencil, domain)
infer_as_fieldop(expr, domain) -> new_expr, dict(referenced_field => accessed domain)
# with new_expr = as_fieldop(stencil, cartesian_domain(...))(inp_field, ...)

# multiple SetAt (no composition)
tmp <- as_fieldop(stencil)(inp_field, ...) @ auto_domain()
out <- as_fieldop(stencil)(tmp, ...) @ ...

# domain is inferred bottom up from last SetAt, only domains of SetAt on temporaries are changed
infer_program(program_expr)
```
### PR Description
- New file gt4py.next.iterator.transforms.infer_domain with function `infer_as_fieldop` and `infer_program`
- Several new tests in test_infer_domain.py to test functionality
- New function SymbolicDomain.translate
- New helper-function `domain` and `as_fieldop` in ir.makers
- Uses constant folding of `itir.Literal`s to simplify resulting domains

### Todos
- [x] Domain inference for nested  `as_fieldop`s (cartesian_domain)
- [x] Domain inference for SetAt
- [x] Extend to unstructured grids

### Note
I agree that my contribution in this PR is additionally licensed under the BSD-3-Clause license (SPDX short identifier).

PR #1584, opened 2024-07-22
fix[next]: Fix size temporary size computation

tehrengruber:
This PR fixes a bug where the size of unstructured temporary fields was wrong in certain cases. This bug surfaced while working on #1568. It took me a while to figure out why this never surfaced in validation tests anywhere:
Right now we compute temporaries for unstructured everywhere, e.g. on all vertices, etc.. What _everywhere_ means is derived from the offset provider in `_max_domain_sizes_by_location_type`. The information is contained either
- explicitly if we have a connectivity with origin axis being the axis whose size we want to know. E.g. if we have a V2E connectivity then we have `table.shape[0]` many vertices.
- or implicitly if we have a connectivity with the neighbor axis whose size we want to know. E.g. if we have a V2E connectivity then we have at least `table.max()+1` many edges. This computation was wrong.

So as long as we have also the E2V connectivity for the examples above, the implicit information is not needed / used and everything works fine. For all our test cases this is indeed the case and since computing temporaries everywhere is only a temporary solution anyway I am fine with just fixing this without increasing test coverage.

### DIFF
--- a/src/gt4py/next/iterator/transforms/global_tmps.py
+++ b/src/gt4py/next/iterator/transforms/global_tmps.py
@@ -407,7 +407,7 @@ def _max_domain_sizes_by_location_type(offset_provider: Mapping[str, Any]) -> di
             )
             sizes[provider.neighbor_axis.value] = max(
                 sizes.get(provider.neighbor_axis.value, 0),
-                provider.table.max(),  # type: ignore[attr-defined] # TODO(havogt): improve typing for NDArrayObject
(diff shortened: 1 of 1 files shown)


egparedes (approved):
LGTM

edopao:
You need to merge `main` into this PR in order to skip the CI on Todi.

edopao:
> You need to merge `main` into this PR in order to skip the CI on Todi.

Done.

PR #1588, opened 2024-07-25
test[next]: Start using requires_dace marker (as cartesian tests already do)

edopao:
Start using `requires_dace` marker in next tests, same as cartesian tests already do.


egparedes (approved):
LGTM

PR #1589, opened 2024-07-25
feat[next][dace]: GTIR-to-SDFG lowering of let-lambdas

edopao:
This PR adds lowering of let-lambdas to DaCe SDFG.


src/gt4py/next/iterator/ir_utils/ir_makers.py
```diff
@@ -423,3 +423,28 @@ def as_fieldop(expr: itir.Expr, domain: Optional[itir.FunCall] = None) -> call:
...
+) -> Callable[..., itir.FunCall]:
+    """
+    Promotes a function `op` to a field_operator.
+    `op` is a function from values to value.
+    Returns:
+        A function from Fields to Field.
+    Examples
+    --------
+    >>> str(op_as_fieldop("op")("a", "b"))
+    '(⇑(λ(__arg0, __arg1) → op(·__arg0, ·__arg1)))(a, b)'
```
egparedes, line 439: Please follow our coding conventions for docstrings  (google-style docstrings):
```suggestion
    Promotes a function `op` to a field_operator.
    
    Args:
        op: is a function from values to value.
    
    Returns:
        A function from Fields to Field.
        
    Examples:
        >>> str(op_as_fieldop("op")("a", "b"))
        '(⇑(λ(__arg0, __arg1) → op(·__arg0, ·__arg1)))(a, b)'
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -324,38 +324,83 @@ def visit_FunCall(
...
+        sdfg: dace.SDFG,
+        head_state: dace.SDFGState,
+        let_symbols: dict[str, gtir_builtin_translators.TemporaryData],
+        args: list[gtir_builtin_translators.TemporaryData],
+    ) -> list[gtir_builtin_translators.TemporaryData]:
         """
-        This visitor class should never encounter `itir.Lambda` expressions
-        because a lambda represents a stencil, which operates from iterator to values.
-        In fieldview, lambdas should only be arguments to field operators (`as_field_op`).
+        All arguments to lambda functions are fields (i.e. `as_fieldop`, field or scalar `gtir.SymRef`,
```
egparedes, line 368: Missing summary line in the docstring.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -324,38 +324,83 @@ def visit_FunCall(
...
+        nested let-lambdas thereof). The dictionary called `let_symbols` maps the lambda parameters
+        to symbols, e.g. temporary fields or program arguments. The lambda parameters can override
+        previously defined symbols, that is why the current dictionary is copied.
         """
-        raise RuntimeError("Unexpected 'itir.Lambda' node encountered in GTIR.")
+        lambda_symbols = let_symbols.copy()
+        lambda_symbols.update({str(p.id): arg for p, arg in zip(node.params, args, strict=True)})
```
egparedes, line 374:
```suggestion
        lambda_symbols = let_symbols | {str(p.id): arg for p, arg in zip(node.params, args, strict=True)}
```

egparedes (commented):
Other than a couple of minor python style issues looks fine to me, but I think someone else should be the main reviewer of this PR (@tehrengruber ? @philip-paul-mueller ?)

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
edopao: Yeah, much nicer!

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -324,38 +324,86 @@ def visit_FunCall(
...
+    ) -> list[gtir_builtin_translators.TemporaryData]:
         """
-        This visitor class should never encounter `itir.Lambda` expressions
-        because a lambda represents a stencil, which operates from iterator to values.
-        In fieldview, lambdas should only be arguments to field operators (`as_field_op`).
+        Translates a `Lambda` node to a tasklet subgraph in the current SDFG state.
+
+        All arguments to lambda functions are fields (i.e. `as_fieldop`, field or scalar `gtir.SymRef`,
+        nested let-lambdas thereof). The dictionary called `let_symbols` maps the lambda parameters
+        to symbols, e.g. temporary fields or program arguments. The lambda parameters can override
```
philip-paul-mueller, line 372: I would write something like. "If the Lambda has an argument whose name is a (previously defined) let symbols, i.e. is also mentioned in `let_symbols` the argument will, during the traversal of the lambda expression, shadow that let symbol."

I also want to conform one thing, this location is the only place where the `let_symbols` get modified?
Furthermore, I am unsure about the behaviour, as I understand it, the current implementation makes all arguments the lambda have available to all of its children, regardless how remote they are and what is between them (with the exception of another Lambda that has the same argument name)?
For this to happen it is not required that a let expression was there in the first place.
I am not sure if this is the intended behaviour or not I just want to confirm it.

My main issue is that it makes the term "let-lambda" meaningless, as every lambda has essentially the same possibilities. On the other hand I see nothing in the IR that allows to create a special let lambda.

Furthermore, another question I had was, if you add all arguments of the lambda to the `let_symbols` are they considered at all?

philip-paul-mueller (commented):
I did not found anything major, I have two suggestions and one question.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
edopao: I will update the docstring as you suggested, here I try to answer your questions.

- Yes, this is the only location where we can override (shadow) previously defined symbols.
- Yes, the current implementation makes all arguments the lambda have available to all of its children. Not only that, a lambda can also access all "global" symbols, that is the program arguments. This behavior can be described, in other words, as the lambda inherits the symbols in the parent scope. I observed this behavior was required in ITIR, I assume it is still valid here @tehrengruber? I mean, a lambda can access symbols defined in the parent scope, not only its own parameters.
- The term "let-lambda" is used to distinguish the regular usage of lambda functions in GTIR (kind of stencils passed as argument to `as_fieldop`) from this particular lambda that creates an intermediate result (a temporary field, indeed) in the current scope and makes it available to the expressions in the child nodes.
- I am not sure I understand your last question. Yes, they are considered because `translate_symbol_ref` first checks if the symbol is present in `let_symbols`

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -309,6 +331,7 @@ def translate_symbol_ref(
     sdfg: dace.SDFG,
     state: dace.SDFGState,
```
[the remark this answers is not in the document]
edopao: As for the reduction PR, my intention is to deliver first this PR and then represent the reduce identity value as let_symbol. In this way, I do not need to add another extra parameter.

It is true that `state` and `let_symbols` belong together: `let_symbols` is a dictionary of `TemporaryData`, and `TemporaryData` is a tuple that contains `AccessNode`. By definition, an `AccessNode` belongs to a state. Yes, I could introduce a dataclass that contains `state` and `let_symbols`. I will consider it but not sure if I want to do it now or later, as refactoring, after the `neghbors` and `reduce` also are in place so I have the full picture.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -273,8 +285,18 @@ def translate_cond(
     sdfg.add_edge(cond_state, false_state, dace.InterstateEdge(condition=(f"not bool({cond})")))
     sdfg.add_edge(false_state, state, dace.InterstateEdge())
 
-    true_br_args = sdfg_builder.visit(true_expr, sdfg=sdfg, head_state=true_state)
-    false_br_args = sdfg_builder.visit(false_expr, sdfg=sdfg, head_state=false_state)
+    true_br_args = sdfg_builder.visit(
+        true_expr,
+        sdfg=sdfg,
+        head_state=true_state,
+        let_symbols=let_symbols,
```
edopao, line 292: After considering @philip-paul-mueller's comment about a dataclass for `state` and `let_symbols`, I realized that here there is a bug. I cannot forward the `AccessNode` in `let_symbols` to the true/false states, because these access nodes belong to the current state. I will fix this.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
philip-paul-mueller: Thanks for your answer.
I was kind of expecting that answer, but thanks for conformation.

Concerning my last point, what I meant by it, since all arguments of a lambda are added to the `let_symbols` will the argument properties of the lambda ever by considered, since you do not have to look at it, since it was evaluated before.
This is more of a question.

edopao: I have now pushed a solution for the problem we discussed on Zoom. I opted for the separate lambda-symbols state, because the direction of SDFG growth (adding states before the head state) is designed for the regular traversal order. The case I showed to you, with let-lambdas as parent of conditional expressions (for which we thought that the conditional node should modify the head state), is specific to let-lambdas and I consider it an exception.
To answer your question: first (in `visit_FunCall`) we evaluate the arguments passed to lambda, and lower it to a dataflow in a separate state that precedes lambda execution; then (in `visit_Lambda`), we map these arguments to the lambda parameters and visit the lambda expression. The lambda expression will contain `SymRef` to the lambda parameters, and at that point we add access nodes to the arguments.
The approach I opted for - the separate lambda-symbols state - relies on the simplify pass to remove redundant states.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -344,8 +344,10 @@ def translate_symbol_ref(
     else:
         sym_value = str(node.id)
         if sym_value in let_symbols:
-            return [let_symbols[sym_value]]
```
philip-paul-mueller: Before you were returning when you encountered a let symbol, but now you run the function.
From the let symbols you get the `sym_node`, however, in the `if` bellow you create a new access node.
I just want to make sure that this is intentional.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -123,8 +123,8 @@ def get_offset_provider(self, offset: str) -> gtx_common.Connectivity | gtx_comm
         return self.offset_provider[offset]
 
     def get_symbol_type(self, symbol_name: str) -> ts.FieldType | ts.ScalarType:
-        assert symbol_name in self.symbol_types
-        return self.symbol_types[symbol_name]
+        assert symbol_name in self.global_symbols
+        return self.global_symbols[symbol_name]
```
philip-paul-mueller, line 127: First of all love `asserts`, but since `dict.__getitem__` will generate a KeyError when the key is missing, that actually tells you what key was not found using the assert here reduces the information, since it does not tell you that.
So either add description to the `assert`, throw your own even more accurate error or get rid of the assert.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -184,7 +184,7 @@ def _add_storage(
 
         # TODO: unclear why mypy complains about incompatible types
         assert isinstance(symbol_type, (ts.FieldType, ts.ScalarType))
-        self.symbol_types[name] = symbol_type
+        self.global_symbols[name] = symbol_type
```
philip-paul-mueller, line 187: Just out of curiosity, why did you rename the `symbol_table` which maps symbols to their type I guess, to `global_symbols` I would not expect that this gives me their type?

philip-paul-mueller (commented):
There are some small things but afterwards we should be able to merge it.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: It is intentional, indeed. Before I was assuming that the let symbols would always be used in the same state where they were created, so I could return and use the access node. With branching and state transitions, I cannot assume it so I always create a new access node in the current state. By design, now, the let symbols are always visited (and therefore initialized) in a state that precedes current state.

philip-paul-mueller: Since it kind of violates the design principles of the SDFG (it is still allowed because simplify will save us) I would add a comment that explains this.

philip-paul-mueller (approved):
LGTM.

PR #1594, opened 2024-07-29
feat[next][dace]: Dace fieldview transformations

philip-paul-mueller:
Initial PR for the Optimization pipeline of the DaCe backend in GT4Py.
The PR is very basic, but contains the basis for later advancement.
Essentially, it defines the `gt_auto_optimize()` function which performs optimization on a basic heuristic.

There are some parts missing, such as:
- There is no way to guide the optimization, i.e. a decision making process that determine if a map should be fused or not.
- There is also no transformation which fuses parallel maps (was not needed for this kernel).


src/gt4py/next/program_processors/runners/dace_fieldview/transformations/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+from .gpu_utils import (
+    GPUSetBlockSize,
+    SerialMapPromoterGPU,
+    gt_gpu_transformation,
+    gt_set_gpu_blocksize,
+)
+from .k_blocking import KBlocking
+from .map_orderer import MapIterationOrder
+from .map_promoter import SerialMapPromoter
+from .map_seriall_fusion import SerialMapFusion
```
edopao, line 31: Is it a typo the double 'l' in serial?

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,372 @@
...
+]
+
+
+def gt_gpu_transformation(
+    sdfg: dace.SDFG,
+    try_removing_trivial_maps: bool = True,
+    use_gpu_storage: bool = True,
+    gpu_block_size: Optional[Sequence[int | str] | str] = None,
+    validate: bool = True,
+    validate_all: bool = False,
```
edopao, line 43: Same comment on `validate_all` as in `auto_opt.py`

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,372 @@
...
+    **kwargs: Any,
+) -> dace.SDFG:
+    """Transform an SDFG into an GPU SDFG.
```
edopao, line 46: an GPU -> a GPU

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,372 @@
...
+
+    The transformation expects a rather optimized SDFG and turn it into an SDFG
+    capable of running on the GPU.
+    The function performs the following steps:
+    - If requested, modify the storage location of the non transient arrays to
+        life on GPU.
```
edopao, line 52: to life -> to live, although I don't like the expression. I suggest "to GPU global memory."

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,372 @@
...
+    - Call the normal GPU transform function followed by simplify.
+    - If requested try to remove trivial kernels.
+    - If given set the `gpu_block_size` parameters of the Maps to the given value.
```
edopao, line 55: A comma would make it easier to read: "If given, ..."
Same above: "If requested, ..."

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,372 @@
...
+
+    Args:
+        sdfg: The SDFG that should be processed.
+        try_removing_trivial_maps: Try to get rid of trivial maps by incorporating them.
+        use_gpu_storage: Assume that the non global memory is already on the GPU.
```
edopao, line 60: I don't understand this: "the non global memory is already on the GPU". After reading the code, I would change it to:
use_gpu_storage: Set to true when the SDFG array arguments are already allocated on GPU global memory. This will avoid the data copy from host to GPU memory.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,372 @@
...
+
+    # You need guru level or above to use these arguments.
+    gpu_launch_factor: Optional[int] = kwargs.get("gpu_launch_factor", None)
+    gpu_launch_bounds: Optional[int] = kwargs.get("gpu_launch_bounds", None)
+
+    # Turn all global arrays (which we identify as input) into GPU memory.
+    #  This way the GPU transformation will not create this copying stuff.
+    if use_gpu_storage:
+        for desc in sdfg.arrays.values():
+            if not (desc.transient or not isinstance(desc, dace.data.Array)):
```
edopao, line 81: For me it would be easier to read as:
`if not desc.transient and isinstance(desc, dace.data.Array)):`

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,372 @@
...
+    sdfg.apply_gpu_transformations(
+        validate=validate,
+        validate_all=validate_all,
+        simplify=False,
+    )
+    # The documentation recommends to run simplify afterwards
+    gtx_transformations.gt_simplify(sdfg)
+
+    if try_removing_trivial_maps:
+        # For reasons a Tasklet can not exist outside a Map in a GPU SDFG. The GPU
```
edopao, line 94: For reasons -> Because of dace design,

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,372 @@
...
+        #  transformation will thus adds trivial maps around them, which translate to
```
edopao, line 95: adds -> add

edopao (commented):
I have only reviewed the transformation driving code. I have not looked yet inside the transformations, which is the most funny part.
Overall it looks very good! It would be nice to include in this PR the nabla4 testcase and the call to the optimize method, in that test.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,311 @@
...
+        # Compute the SDFG to see if something has changed.
+        sdfg_hash = sdfg.hash_sdfg()
+
+        # Phase 2: Kernel Creation
+        #   We will now try to reduce the number of kernels and create big one.
+        #   For this we essentially use map fusion. We do this is a loop because
+        #   after a graph modification followed by simplify new fusing opportunities
+        #   might arise. We use the hash of the SDFG to detect if we have reached a
+        #   fix point.
+        for _ in range(100):
```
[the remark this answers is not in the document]
philip-paul-mueller: I am not sure if the number of maps is a good value.
However, I very much agree that we should use some "informed" value.
I left a todo.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,311 @@
...
+                validate_all=validate_all,
+            )
+
+            # Now do some cleanup task, that may enable further fusion opportunities.
+            #  Note for performance reasons simplify is deferred.
+            phase2_cleanup = []
+            phase2_cleanup.append(dace_dataflow.TrivialTaskletElimination())
+
+            # TODO(phimuell): Should we do this all the time or only once? (probably the later)
+            # TODO(phimuell): More control what we promote.
```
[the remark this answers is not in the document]
philip-paul-mueller: Clarified.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,311 @@
...
+        make_persistent: Turn all transients to persistent lifetime, thus they are
+            allocated over the whole lifetime of the program, even if the kernel exits.
+            Thus the SDFG can not be called by different threads.
+        gpu_block_size: The thread block size for maps in GPU mode, currently only
+            one for all.
+        block_dim: On which dimension blocking should be applied.
+        blocking_size: How many elements each block should process.
+        reuse_transients: Run the `TransientReuse` transformation, might reduce memory footprint.
+        validate: Perform validation during the steps.
+        validate_all: Perform extensive validation.
```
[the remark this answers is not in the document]
philip-paul-mueller: You can not activate `validate_all` just in a specific step.
Furthermore, all DaCe transformations accept the `validate` and `validate_all` argument, so I added them to be compatible.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,311 @@
...
+            )
+
+            # Use the hash to determine if the transformations did modify the SDFG.
+            #  If not we have optimized the SDFG as much as we could, in this phase.
+            old_sdfg_hash = sdfg_hash
+            sdfg_hash = sdfg.hash_sdfg()
+            if old_sdfg_hash == sdfg_hash:
+                break
+
+            # The SDFG was modified by the transformations above. But no
```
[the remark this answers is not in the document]
philip-paul-mueller: No you are right. A transformation could be applied, this is why we have to call simplify.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,311 @@
...
+            gpu_launch_factor: Optional[int] = kwargs.get("gpu_launch_factor", None)
+            gpu_launch_bounds: Optional[int] = kwargs.get("gpu_launch_bounds", None)
+            gtx_transformations.gt_gpu_transformation(
+                sdfg,
+                gpu_block_size=gpu_block_size,
+                gpu_launch_bounds=gpu_launch_bounds,
+                gpu_launch_factor=gpu_launch_factor,
+                validate=validate,
+                validate_all=validate_all,
+                try_removing_trivial_maps=True,
```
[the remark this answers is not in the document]
philip-paul-mueller: The problem is it is very GPU specific.
In a GPU SDFG a Tasklet must be inside a Map, which is different for CPU.
If you call `apply_gpu_transformation()` then DaCe will add a trivial map around every single Tasklet.
And each of them results in its own kernel launch.
This argument instructs the transformation to try to get rid of them.
I agree its name is not that good, and something like `try_resolve_trivial_maps` may be better.
What do you think?

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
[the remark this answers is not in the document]
edopao: You can manually edit the source code and set `validate_all` on the specific step you need to debug. I mean, will you ever want to set `validate_all` on all steps at once? It introduces so much overhead that I do not find it useful, indeed. On the contrary, I completely agree on "validate": it is useful to run all the steps and eventually validate each of them to see which step requires detailed debugging.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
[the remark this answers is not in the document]
edopao: I see now, `apply_gpu_transformation` is called inside this function so trivial maps only exist if we call `gt_gpu_transformation`. Then, it makes sense to remove trivial maps inside. I think the name `try_removing_trivial_maps` is fine

tests/next_tests/unit_tests/program_processor_tests/runners_tests/transformation_tests/__init__.py
```diff
@@ -0,0 +1,14 @@
+# GT4Py - GridTools Framework
```
edopao, line 1: The module name should contain "dace". We could create a subpackage for it and move my existing tests there too:
```
tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace/transformation_tests/__init__.py
tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace/test_gtir_to_sdfg.py (was test_dace_fieldview.py)
```

tests/next_tests/unit_tests/program_processor_tests/runners_tests/transformation_tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+
+from gt4py.next import common as gtx_common
+from gt4py.next.program_processors.runners.dace_fieldview import (
+    transformations as gtx_transformations,
+)
+
+
+@pytest.fixture(autouse=True)
+def _set_dace_settings() -> Generator[None, None, None]:
+    """Enables the correct settings in DaCe."""
```
edopao, line 32: "correct" doesn't tell much.. maybe "Customizes settings for optimization and code generation in DaCe."

edopao (commented):
@philip-paul-mueller I think it would be nice to merge the nabla4 test case, either in this same PR or a separate one (since nabla4 testcase depends on my PR for neighbors reduction). I could help with that.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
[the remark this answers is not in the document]
edopao: Anyway, I do not have a strong opinion against it. If you would like to keep it, please do so.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,374 @@
...
+        launch_factor=gpu_launch_factor,
+    )
+    return sdfg.apply_transformations_once_everywhere([xform])
+
+
+def _gpu_block_parser(
+    self: "GPUSetBlockSize",
+    val: Any,
+) -> None:
+    """Used by the setter ob `GPUSetBlockSize.block_size`."""
```
edopao, line 157: ob -> of

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,374 @@
...
+    org_val = val
+    if isinstance(val, tuple):
+        pass
+    elif isinstance(val, list):
+        val = tuple(val)
+    elif isinstance(val, str):
+        val = tuple(x.replace(" ", "") for x in val.split(","))
```
edopao, line 164: Probably achieves the same result?
`x.strip() for x in val.split(",")`

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,374 @@
...
+            f"Currently only block sizes convertible to int are supported, you passed '{val}'."
+        ) from None
+    self._block_size = val
+
+
+def _gpu_block_getter(
+    self: "GPUSetBlockSize",
+) -> tuple[int, int, int]:
+    """Used as getter in the `GPUSetBlockSize.block_size` property."""
+    assert isinstance(self._block_size, (tuple, list)) and len(self._block_size) == 3
```
edopao, line 188: Here I have a question. I have seen most fields are bi-dimensional. Why should we force a tri-dimensional block? Are we relying on the fact that the second and third dimensions are always the same?

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+@properties.make_properties
+class SerialMapPromoterGPU(transformation.SingleStateTransformation):
+    """Serial Map promoter for empty Maps in case of trivial Maps.
+
+    In CPU mode a Tasklet can be outside of a map, however, this is not
+    possible in GPU mode. For this reason DaCe wraps such Tasklets in a
+    trivial Map.
+    This transformation will look for such Maps and promote them, such
```
edopao, line 297: I see the benefit of this transformation. We are moving program code (I mean computation) to each CUDA thread. I do not know the kind of computation well enough to say whether it could be interesting to instead compute this data on CPU and access it from GPU (implies a device to host read, which eventually is not too expensive on gh200). Removing computation from the GPU is a way to reduce register utilisation, I guess.

edopao (commented):
I reviewed `gpu_utils.py` (I will probably review one file at a time, for SDFG transformations).

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
philip-paul-mueller: You are right most fields are, but GPU Kernels always have three dimensions, thus you need to return three values.

edopao: I think the main reason why this is not an issue is that we assume there will be one single gpu map for the entire program. The time we start defining multiple cuda kernels within a program (by means of global temporaries), not necessarily all maps will have to use the same kernel block. Then, different definitions of 2d and 3d grids could be interesting.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
philip-paul-mueller: I see what you are mean.
But as far as I understand GT4Py having trivial Maps is most likely an extreme edge case.
Currently, Tasklets outside of Maps are only these "literal Tasklets", by moving them inside teh same kernel nvcc will detect them as literal/constant.

if you put them in another kernel/compilation unit, nvcc will never know that it is a literal so it will always result in a memory load, which is slower than a literal (which could blow up binary size) or a register.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_serial_fusion.py
```diff
@@ -0,0 +1,475 @@
...
+                    storage=dtypes.StorageType.Register,
+                    find_new_name=True,
+                )
+
+            else:
+                assert (pre_exit_edge.data.subset.num_elements() > 1) or all(
+                    x == 1 for x in new_inter_shape
+                )
+                is_scalar = False
+                new_inter_name, new_inter_desc = sdfg.add_transient(
```
edopao, line 298: In case the maps cannot be all fused, I guess that the strides of these intermediates will be important. However, I do not have any good suggestion how to handle that (e.g. should they inherit the strides of source/destination nodes?). Maybe write a TODO comment.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_serial_fusion.py
```diff
@@ -0,0 +1,475 @@
...
+                    new_inter_name,
+                    shape=new_inter_shape,
+                    dtype=inter_desc.dtype,
+                    find_new_name=True,
+                )
+            new_inter_node: nodes.AccessNode = state.add_access(new_inter_name)
+
+            # New we will reroute the output Memlet, thus it will no longer going
```
edopao, line 306: will no longer going -> will no longer pass

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_serial_fusion.py
```diff
@@ -0,0 +1,475 @@
...
+            pre_exit_memlet: dace.Memlet = pre_exit_edge.data
+            new_pre_exit_memlet = copy.deepcopy(pre_exit_memlet)
+
+            # We might operate on a different array, but the check below, ensures
+            #  that we do not change the direction of the Memlet.
+            assert pre_exit_memlet.data == inter_name
+            new_pre_exit_memlet.data = new_inter_name
+
+            # Now we have to fix the subset of the Memlet.
+            #  Before the subset of the Memlet dependent on the Map variables,
```
edopao, line 318: dependent -> was dependent

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_serial_fusion.py
```diff
@@ -0,0 +1,475 @@
...
+                # Ensure the correctness of the rerouting below.
+                # TODO(phimuell): Improve the code below to remove the check.
+                assert producer_edge.data.data == inter_name
+
+                # Will not change the direction, because of test above!
+                producer_edge.data.data = new_inter_name
+                producer_edge.data.replace(memlet_repl)
+                if is_scalar:
+                    producer_edge.data.dst_subset = "0"
+                else:
```
edopao, line 353: What about merging else/if ?

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_serial_fusion.py
```diff
@@ -0,0 +1,475 @@
...
+                    if producer_edge.data.dst_subset is not None:
+                        producer_edge.data.dst_subset.pop(squeezed_dims)
+
+            # Now after we have handled the input of the new intermediate node,
+            #  we must handle its output. For this we have to "inject" the temporary
```
edopao, line 358: temporary -> intermediate (just for consistent terminology, assuming I've understood it right)

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_serial_fusion.py
```diff
@@ -0,0 +1,475 @@
...
+                    assert inner_edge.data.data == inter_name  # DIRECTION!!
+
+                    # The create the first Memlet to transmit information, within
+                    #  the second map, we do this again by copying and modifying
+                    #  the original Memlet.
+                    # NOTE: Test above is important to ensure the direction of the
+                    #       Memlet and the correctness of the code below.
+                    new_inner_memlet = copy.deepcopy(inner_edge.data)
+                    new_inner_memlet.replace(memlet_repl)
+                    new_inner_memlet.data = new_inter_name  # Because of the assert above, this will not chenge the direction.
```
edopao, line 389: chenge -> change

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_serial_fusion.py
```diff
@@ -0,0 +1,475 @@
...
+                del sdfg.arrays[inter_name]
+
+            else:
+                # This is the shared mode, so we have to recreate the intermediate
+                #  node, but this time it is at the exit of the second map.
+                state.remove_edge(pre_exit_edge)
+                map_exit_1.remove_in_connector(pre_exit_edge.dst_conn)
+
+                # This is the Memlet that goes from the map internal intermediate
+                #  temporary node to the Map output. This will essentially restore
```
edopao, line 446: intermediate temporary -> intermediate

edopao (commented):
Some comments on `map_serial_fusion.py`

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_orderer.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # This implementation will just swap the variable that is currently the last
+        #  with the one that should be the last.
+        dst_idx = -1
+        src_idx = map_params.index(map_var)
+
+        for to_process in [
+            map_entry.map.params,
+            map_entry.map.range.ranges,
+            map_entry.map.range.tile_sizes,
```
edopao, line 123: I was reviewing `map_orderer.py`. I do not have any comment on this file, however I've seen (I didn't know) that maps also have a `tile_sizes` property. Maybe it could be useful for K-blocking.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_promoter.py
```diff
@@ -0,0 +1,379 @@
...
+
+    By adding certain dimension to a Map, it might became possible to use the Map
+    in more transformations. This class acts as a base and the actual matching and
+    checking must be implemented by a concrete implementation.
+    But it provides some basic check functionality and the actual promotion logic.
+
+    The transformation operates on two Maps, first the "source map". This map
+    describes the Map that should be used as template. The second one is "map to
+    promote". After the transformation the "map to promote" will have the same
+    map parameter than the "source map" has.
```
edopao, line 39: than the "source map" has -> as the "source map"

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_promoter.py
```diff
@@ -0,0 +1,379 @@
...
+
+    To influence what to promote the user must implement the `map_to_promote()`
+    and `source_map()` function. They have to return the map entry node.
+
+    Args:
+        only_inner_maps: Only match Maps that are internal, i.e. inside another Map.
+        only_toplevel_maps: Only consider Maps that are at the top.
+        promote_vertical: If `True` promote vertical dimensions; `True` by default.
+        promote_local: If `True` promote local dimensions; `True` by default.
+        promote_horizontal: If `True` promote horizontal dimensions; `False` by default.
```
edopao, line 54: promote_all is missing

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_promoter.py
```diff
@@ -0,0 +1,379 @@
...
+    # Pattern Matching
+    exit_first_map = transformation.transformation.PatternNode(nodes.MapExit)
+    access_node = transformation.transformation.PatternNode(nodes.AccessNode)
+    entry_second_map = transformation.transformation.PatternNode(nodes.MapEntry)
+
+    @classmethod
+    def expressions(cls) -> Any:
+        """Get the match expressions.
+
+        The function generates two different match expression. The first match
```
edopao, line 314: different? they look exactly the same

edopao (commented):
A few more comments on `map_orderer.py` and `mapPromoter.py`

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
```diff
@@ -0,0 +1,173 @@
...
+
+from typing import Iterable
+
+import dace
+from dace.sdfg import graph as dace_graph, nodes
+
+
+def is_nested_sdfg(sdfg: dace.SDFG) -> bool:
+    """Tests if `sdfg` is a neseted sdfg."""
+    if isinstance(sdfg, dace.SDFGState):
```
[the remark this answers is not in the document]
philip-paul-mueller: You are right the function parameter had the wrong annotation.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace/transformation_tests/dace_fieldview
```diff
@@ -0,0 +1 @@
+/home/quint_essent/git/1_CSCS/gt4py/src/gt4py/next/program_processors/runners/dace_fieldview
```
edopao, line 1: Correct test file to be moved and renamed to `test_gtir_to_sdfg.py`

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
```diff
@@ -0,0 +1,173 @@
...
+    begin: nodes.Node,
+    end: nodes.Node,
+    reverse: bool = False,
+) -> set[nodes.Node] | None:
+    """Find all nodes that are reachable from `begin` but bound by `end`.
+
+    Essentially the function starts a DFS at `begin`, which is never part of the
+    returned set, if at a node an edge is found that lead to `end`, the function
+    will ignore this edge. However, it will find every node that is reachable
+    from `begin` that is reachable by a path that does not visit `end`.
```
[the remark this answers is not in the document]
philip-paul-mueller: I added some of your modification.
The name of this function is badly named it is essentially used as some kind of reachability test.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_serial_fusion.py
```diff
@@ -0,0 +1,475 @@
...
+
+            # Now generate the intermediate data container.
+            if len(new_inter_shape) == 0:
+                assert pre_exit_edge.data.subset.num_elements() == 1
+                is_scalar = True
+                new_inter_name, new_inter_desc = sdfg.add_scalar(
+                    new_inter_name,
+                    dtype=inter_desc.dtype,
+                    transient=True,
+                    storage=dtypes.StorageType.Register,
```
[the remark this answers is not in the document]
philip-paul-mueller: In SDFG register means to not dynamically allocate it.
Otherwise DaCe will allocate it using `new` inside the GPU kernel.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
```diff
@@ -0,0 +1,439 @@
...
+
+@properties.make_properties
+class KBlocking(transformation.SingleStateTransformation):
+    """Applies k-Blocking with separation on a Map.
+
+    This transformation takes a multidimensional Map and performs blocking on a
+    dimension, that is commonly called "k", but identified with `block_dim`.
+
+    All dimensions except `k` are unaffected by this transformation. In the outer
+    Map the will replace the `k` range, currently `k = 0:N`, with
```
edopao, line 36: the will -> it will

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
```diff
@@ -0,0 +1,439 @@
...
+        sdfg: dace.SDFG,
+        permissive: bool = False,
+    ) -> bool:
+        """Test if the map can be blocked.
+
+        The test involves:
+        - Toplevel map.
+        - The map shall not be serial.
+        - The block dimension must be present (exact match).
+        - The map range must have stride one.
```
edopao, line 96: How do you check this requirement? The map itself does not have a stride. From what I see below you require that the range step is 1.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
```diff
@@ -0,0 +1,439 @@
...
+            name=inner_label,
+            ndrange=inner_range,
+            schedule=dace.dtypes.ScheduleType.Sequential,
+        )
+
+        # TODO(phimuell): Investigate if we want to prevent unrolling here
+
+        # Now we modify the properties of the outer map.
+        coarse_block_range = subsets.Range.from_string(
+            f"0:int_ceil(({rng_stop} + 1) - {rng_start}, {self.blocking_size})"
```
edopao, line 170: The start index should not be `0` but `{rng_start}`

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
```diff
@@ -0,0 +1,439 @@
...
+                # The node was already fully handled in the `else` clause.
+                continue
+
+            else:
+                # Relocate the node and make the reconnection.
+                #  Different from the dependent case we will handle all the edges
+                #  of the node in one go.
+                relocated_nodes.add(edge_dst)
+
+                # In order to be useful we have to temporary store the data the
```
edopao, line 234: temporary store -> temporarily store

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_serial_fusion.py
philip-paul-mueller: This is actually a very hard problem.
If you make it fully symbolic, then you have to supply them as arguments.
The only thing that is relatively simple is to change between C and F order.
But I think Daisy Tuner could do something.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_serial_fusion.py
philip-paul-mueller: You were right.
I have adapted the wording.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_orderer.py
philip-paul-mueller: This thing is actually deprecated and should be removed.
I have no idea what it does and when I asked the answer was essentially "do not use it, it is very old".

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/map_promoter.py
philip-paul-mueller: You are right 'different' is probably the wrong word.
You can see the reason in `map_to_promote()` and `source_map()` where the returned value depends on `self.expr_index` (which is defined by the pattern matching mechanism).
This is done to have only one code path.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
philip-paul-mueller: The map has a stride, a range object is essentially a list of length 3 tuple:
- initial value
- end value (including, so it is not pass the end).
- step size (stride).

edopao: Yes, I was just confused by the name (I usually call it step).

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
philip-paul-mueller: This is the coarse block range (shitty name) should start from 0, because it iterates through the blocks, we also give it another name.
If you look above at line 157 where the inner range is defined, you see that it starts at the expected location.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
philip-paul-mueller: Steps sounds actually better.

edopao (approved):
The PR is great! We will find and fix problems when we test more stencils and apply these transformations to different SDFGs. The initial test coverage is very appreciated:
```
dace_fieldview/transformations/__init__.py                 100%
dace_fieldview/transformations/auto_opt.py                  16%
dace_fieldview/transformations/gpu_utils.py                 20%
dace_fieldview/transformations/k_blocking.py                74%
dace_fieldview/transformations/map_fusion_helper.py         64%
dace_fieldview/transformations/map_orderer.py               30%
dace_fieldview/transformations/map_promoter.py              64%
dace_fieldview/transformations/map_serial_fusion.py         92%
dace_fieldview/transformations/util.py                      75%
```
For me we could even merge it and eventually create new PRs in next cycles, or we can keep it open and pull it when necessary, as you prefer.

havogt (changes requested):
I'll do a round off style review.

philip-paul-mueller:
@havogt I am on vacation, so no hurry.

philip-paul-mueller:
@edopao Thanks for spotting it I have removed it.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
```diff
@@ -0,0 +1,168 @@
...
+        elif n in seen:
+            continue
+        seen.add(n)
+        to_visit.extend(next_nodes(n))
+
+    if not found_end:
+        return None
+
+    seen.discard(begin)
+    return seen
```
egparedes, line 87: Just a suggestion to make the code simpler:
```suggestion
    while len(to_visit) > 0:
        n: dace_nodes.Node = to_visit.pop()
        if n not in seen:
            seen.add(n)
            if n != end:
                to_visit.extend(next_nodes(n))

    return seen - {begin, end} if end in seen else None
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
```diff
@@ -0,0 +1,168 @@
...
+        begin: The initial node that from which the search starts.
+        only_tasklets: Return only Tasklets.
+        reverse: Follow the reverse direction.
+    """
+    if isinstance(begin, dace_graph.MultiConnectorEdge):
+        to_visit: list[dace_graph.MultiConnectorEdge[dace.Memlet]] = [begin]
+    elif reverse:
+        to_visit = list(state.in_edges(begin))
+    else:
+        to_visit = list(state.out_edges(begin))
```
egparedes, line 119:
```suggestion
    else:
        to_visit = list(state.in_edges(begin)) if reverse else list(state.out_edges(begin))
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
```diff
@@ -0,0 +1,168 @@
...
+    seen: set[dace_graph.MultiConnectorEdge[dace.Memlet]] = set()
+    found: set[tuple[dace_nodes.Node, dace_graph.MultiConnectorEdge[dace.Memlet]]] = set()
+
+    while len(to_visit) != 0:
```
egparedes, line 123: For coherence with the code above:
```suggestion
    while len(to_visit) > 0:
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
```diff
@@ -0,0 +1,168 @@
...
+                    continue
+                target_conn = curr_edge.dst_conn[3:]
+                new_edges = state.out_edges_by_connector(curr_edge.dst, "OUT_" + target_conn)
+            to_visit.extend(new_edges)
+            del new_edges
+        else:
+            if only_tasklets and (not isinstance(next_node, dace_nodes.Tasklet)):
+                continue
+            found.add((next_node, curr_edge))
+
```
egparedes, line 153:
```suggestion
        elif not only_tasklets and isinstance(next_node, dace_nodes.Tasklet):
            found.add((next_node, curr_edge))

```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
```diff
@@ -0,0 +1,168 @@
...
+                    next_node, dace_nodes.MapEntry
+                ):
+                    # This edge defines a dynamic map range, which is a consumer
+                    if not only_tasklets:
+                        found.add((next_node, curr_edge))
+                    continue
+                target_conn = curr_edge.dst_conn[3:]
+                new_edges = state.out_edges_by_connector(curr_edge.dst, "OUT_" + target_conn)
+            to_visit.extend(new_edges)
+            del new_edges
```
egparedes, line 148: Reorder statements for readability:
```suggestion
            if not reverse:
                # In forward mode a Map entry could also mean the definition of a
                #  dynamic map range.
                if (not curr_edge.dst_conn.startswith("IN_")) and isinstance(
                    next_node, dace_nodes.MapEntry
                ):
                    # This edge defines a dynamic map range, which is a consumer
                    if not only_tasklets:
                        found.add((next_node, curr_edge))
                    continue
                    
                target_conn = curr_edge.src_conn[4:]
                new_edges = state.in_edges_by_connector(curr_edge.src, "IN_" + target_conn)
            else:
                target_conn = curr_edge.dst_conn[3:]
                new_edges = state.out_edges_by_connector(curr_edge.dst, "OUT_" + target_conn)

            to_visit.extend(new_edges)
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
```diff
@@ -0,0 +1,437 @@
...
+import functools
+from typing import Any, Optional, Union
+
+import dace
+from dace import (
+    properties as dace_properties,
+    subsets as dace_subsets,
+    transformation as dace_transformation,
+)
+from dace.sdfg import SDFG, SDFGState, graph as dace_graph, nodes as dace_nodes
```
egparedes, line 19: Direct symbol imports are forbidden in our coding conventions.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
```diff
@@ -0,0 +1,437 @@
...
+    def __init__(
+        self,
+        blocking_size: Optional[int] = None,
+        block_dim: Optional[Union[gtx_common.Dimension, str]] = None,
+    ) -> None:
+        super().__init__()
+        if isinstance(block_dim, str):
+            pass
+        elif isinstance(block_dim, gtx_common.Dimension):
+            block_dim = gtx_dace_fieldview_util.get_map_variable(block_dim)
```
egparedes, line 71:
```suggestion
        if isinstance(block_dim, gtx_common.Dimension):
            block_dim = gtx_dace_fieldview_util.get_map_variable(block_dim)
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
```diff
@@ -0,0 +1,437 @@
...
+        outer_map.params[block_idx] = coarse_block_var
+        outer_map.range[block_idx] = coarse_block_range
+        outer_map.label = f"{outer_map.label}_blocked"
+
+        # Contains the independent nodes that are already relocated.
+        relocated_nodes: set[dace_nodes.Node] = set()
+
+        # Now we iterate over all the output edges of the outer map and rewire them.
+        #  Note that this only handles the entry of the Map.
+        for out_edge in list(graph.out_edges(outer_entry)):
```
egparedes, line 179: Is a list instantiation really needed here? 
```suggestion
        for out_edge in graph.out_edges(outer_entry):
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
```diff
@@ -0,0 +1,437 @@
...
+            f"0:int_ceil(({rng_stop} + 1) - {rng_start}, {self.blocking_size})"
+        ).ranges[0]
+        outer_map.params[block_idx] = coarse_block_var
+        outer_map.range[block_idx] = coarse_block_range
+        outer_map.label = f"{outer_map.label}_blocked"
+
+        # Contains the independent nodes that are already relocated.
+        relocated_nodes: set[dace_nodes.Node] = set()
+
+        # Now we iterate over all the output edges of the outer map and rewire them.
```
egparedes, line 177: From this point on, the function becomes too complex and long to be easily understood. I suggest to split it into readable smaller functions.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
```diff
@@ -0,0 +1,437 @@
...
+        block_independent: set[dace_nodes.Node] = set()  # `\mathcal{I}`
+        block_dependent: set[dace_nodes.Node] = set()  # `\mathcal{D}`
+
+        # Find all nodes that are adjacent to the map entry.
+        nodes_to_partition: set[dace_nodes.Node] = {edge.dst for edge in state.out_edges(map_entry)}
+
+        # Now we examine every node and assign them to one of the sets.
+        #  Note that this is only tentative and we will later inspect the
+        #  outputs of the independent node and reevaluate their classification.
+        for node in nodes_to_partition:
```
egparedes, line 332: Same comment as above. The combination of `return` and `continue` statements in the same loop makes the code very hard to reason about. Please split it in smaller pieces or refactor the logic to avoid the combination of different loop flow statements

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace/transformation_tests/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+from dace.transformation import dataflow as dace_dataflow
+
+from gt4py.next import common as gtx_common
+from gt4py.next.program_processors.runners.dace_fieldview import (
+    transformations as gtx_transformations,
+)
+
+
+@pytest.fixture(autouse=True)
+def _set_dace_settings() -> Generator[None, None, None]:
```
egparedes, line 25: This is a public fixture used everywhere so it should be named accordingly. Also, since this is an important part of the test machinery,  I'd also add a doctring to document which tests are using it and a link to read further documentation about this feature (for other developers reading the code without a deep knowledge of all pytest options).
```suggestion
def set_dace_settings() -> Generator[None, None, None]:
```

egparedes (changes requested):
I've skipped the review of all `*map_*` files since they are likely to be removed in the future.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,341 @@
...
+    sdfg: dace.SDFG,
+    device: dace.DeviceType = dace.DeviceType.CPU,
+    use_gpu_storage: bool = True,
+    **kwargs: Any,
+) -> dace.SDFG:
+    """This is a convenient wrapper arround DaCe's `auto_optimize` function.
+
+    Args:
+        sdfg: The SDFG that should be optimized in place.
+        device: the device for which optimizations should be done, defaults to CPU.
```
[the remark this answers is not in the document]
philip-paul-mueller: Thanks for spotting it.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,341 @@
...
+from dace.transformation import dataflow as dace_dataflow
+from dace.transformation.auto import auto_optimize as dace_aoptimize
+
+from gt4py.next import common as gtx_common
+from gt4py.next.program_processors.runners.dace_fieldview import (
+    transformations as gtx_transformations,
+)
+
+
+__all__ = [
```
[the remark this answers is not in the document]
philip-paul-mueller: I agree, thanks for pointing that out.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,341 @@
...
+        use_gpu_storage: Assumes that the SDFG input is already on the GPU.
+            This parameter is `False` in DaCe but here is changed to `True`.
+        kwargs: Are forwarded to the underlying auto optimized exposed by DaCe.
+    """
+    return dace_aoptimize.auto_optimize(
+        sdfg,
+        device=device,
+        use_gpu_storage=use_gpu_storage,
+        **kwargs,
+    )
```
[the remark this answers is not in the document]
philip-paul-mueller: I added this function for some reason during my experiment and essentially forgot (or not think) to remove it.

Regarding `use_gpu_storage` only takes effect if the device is GPU.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,341 @@
...
+    validate_all: bool = False,
+    **kwargs: Any,
+) -> dace.SDFG:
+    """Performs GT4Py specific optimizations on the SDFG in place.
+
+    The auto optimization works in different phases, that focuses each on
+    different aspects of the SDFG. The initial SDFG is assumed to have a
+    very large number of rather simple Maps.
+
+    1. Some general simplification transformations, beyond classical simplify,
```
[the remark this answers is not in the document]
philip-paul-mueller: There is actually no reason to not do it, however, if you look at the individual phases it does not _yet_ make sense to turn them into separate functions, with the exception of phase 2.

However, I agree that if the function become more complex we should separate the phases where it make sense.
And I will do it for the 2. phase.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,341 @@
...
+        verbose=False,
+        skip=skip,
+    ).apply_pass(sdfg, {})
+
+
+def gt_set_iteration_order(
+    sdfg: dace.SDFG,
+    leading_dim: gtx_common.Dimension,
+    validate: bool = True,
+    validate_all: bool = False,
```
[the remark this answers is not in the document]
philip-paul-mueller: Sorry I forgot them to forward to the underlying implementation.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,311 @@
...
+            sdfg.apply_transformations_once_everywhere(
+                gtx_transformations.MapIterationOrder(
+                    leading_dim=leading_dim,
+                )
+            )
+
+        # Phase 5: Apply blocking
+        if block_dim is not None:
+            sdfg.apply_transformations_once_everywhere(
+                gtx_transformations.KBlocking(
```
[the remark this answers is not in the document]
philip-paul-mueller: K Blocking is the name this technique is known. But I actually do not really care (except that _I_ will have to change it everywhere).
Then please suggest a name for it.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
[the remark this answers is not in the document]
philip-paul-mueller: This `validate_all`is standard in all DaCe call so it is there because it is supposed to be there.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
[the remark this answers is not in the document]
philip-paul-mueller: I now added an argument for that that defaulted to the value 100.
Although we will probably never set it anyway.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,370 @@
...
+    - Call the normal GPU transform function followed by simplify.
+    - If requested try to remove trivial kernels.
+    - If specified, set the `gpu_block_size` parameters of the Maps to the given value.
+
+    Args:
+        sdfg: The SDFG that should be processed.
+        try_removing_trivial_maps: Try to get rid of trivial maps by incorporating them.
+        use_gpu_storage: Assume that the non global memory is already on the GPU.
+        gpu_block_size: Set to true when the SDFG array arguments are already allocated
+            on GPU global memory. This will avoid the data copy from host to GPU memory.
```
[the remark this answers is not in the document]
philip-paul-mueller: You are right it is actually one that would fit the one on top.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,370 @@
...
+        that you should set `try_removing_trivial_maps` to `False`.
+
+    Todo:
+        - Solve the fusing problem.
+        - Currently only one block size for all maps is given, add more options.
+    """
+
+    # You need guru level or above to use these arguments.
+    gpu_launch_factor: Optional[int] = kwargs.pop("gpu_launch_factor", None)
+    gpu_launch_bounds: Optional[int] = kwargs.pop("gpu_launch_bounds", None)
```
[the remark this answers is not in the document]
philip-paul-mueller: They are documented, in the line above it is written that you need "guru level" to use them.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -0,0 +1,370 @@
...
+        val = (*val, 1)
+    elif len(val) != 3:
+        raise ValueError(f"Can not parse block size '{org_val}': wrong length")
+    try:
+        val = [int(x) for x in val]
+    except ValueError:
+        raise TypeError(
+            f"Currently only block sizes convertible to int are supported, you passed '{val}'."
+        ) from None
+    self._block_size = val
```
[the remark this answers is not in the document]
philip-paul-mueller: This is how a custom parser is added to a DaCe property.
I took some of your ideas but essentially kept the code as it is.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
philip-paul-mueller: I see your point but did not fully like your solution, so I made a blend.
I think it is better, but it could be better.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
philip-paul-mueller: I darkly remember that it was, because you are gonna manipulate that very set of edges.
However, I looked at the DaCe source and it seems that they do that already, so it should be technically safe to do do. But it seems that it would not be save if we use the networkx layer directly.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
philip-paul-mueller: I liked your idea and applied it.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
philip-paul-mueller: I have now refactored the class.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# Copyright (c) 2014-2024, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Transformation and optimization pipeline for the DaCe backend in GT4Py.
+
+Please also see [this HackMD document](https://hackmd.io/@gridtools/rklwk4OIR#Requirements-on-SDFG)
```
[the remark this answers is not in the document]
philip-paul-mueller: I added one.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace/transformation_tests/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+from gt4py.next.program_processors.runners.dace_fieldview import (
+    transformations as gtx_transformations,
+)
+
+
+@pytest.fixture(autouse=True)
+def set_dace_settings() -> Generator[None, None, None]:
+    """Sets the common DaCe settings for the tests.
+
+    The function will modify the following settings:
```
egparedes, line 28: Please add a comment mentioning that because of the `autouse=True` option, this fixture will be automatically used by all tests inside this folder (and subfolders). To be honest, I'm not really sure if I like this implicit/magic behavior instead of  renaming the fixture to just `dace_config` and making it clear that it should be explicit called by all tests. Any other opinions about it @edopao @havogt ?

egparedes (commented):
It seems ready to go after fixing the ADR text and a couple of open conversations. It should be possible to merge this PR today/tomorrow from my point of view

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
```diff
@@ -0,0 +1,168 @@
...
+
+    Notes:
+        - The returned set will also contain the nodes of path that starts at
+            `begin` and ends at a node that is not `end`.
+    """
+
+    def next_nodes(node: dace_nodes.Node) -> Iterable[dace_nodes.Node]:
+        if reverse:
+            return (edge.src for edge in graph.in_edges(node))
+        return (edge.dst for edge in graph.out_edges(node))
```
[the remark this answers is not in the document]
philip-paul-mueller: I think in this particular case the `return ... if ... else ...` is much less readable than the current 3 line version.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
philip-paul-mueller: It has to do with what the function does finding consumer. A map is not a consumer but more like a "door" where data passes through, that its why it has these `IN_*` and `OUT_*` connectors.
In the if branch we check if we found such a door if we have not found one, then must have found a consumer.
If we are only interested Tasklets and have not found a Tasklet then we have found a consumer we do not care about and do not want to report.
In all other cases we have found a consumer we want to report.

A solution could be to have nested `if` which would make it a bit clearer in this case.
I.e. something like 
```python
elif only_tasklets:
    if isinstance(next_node, dace_nodes.Tasklet):
        found.add((next_node, curr_edge))
else:
    found.add((next_node, curr_edge))
```
I personally do not care.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -0,0 +1,369 @@
...
+        # TODO(phimuell): Fix the bug, it uses the tile value and not the stack array value.
+        dace_aoptimize.move_small_arrays_to_stack(sdfg)
+        if make_persistent:
+            # TODO(phimuell): Allow to also to set the lifetime to `SDFG`.
+            dace_aoptimize.make_transients_persistent(sdfg, device)
+
+        return sdfg
+
+
+def _gt_auto_optimize_phase_2(
```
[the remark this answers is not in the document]
philip-paul-mueller: I agree that `gt_auto_fuse_top_level_maps()` is more meaningful, however, calling these functions outside of `gt_auto_optimize()` is most likely not useful (okay this is an exception). For that reason I would add a leading underscore.
Never the less, I added a comment to the function that calling it directly is probably not correct.

I also agree that the situation that a public doc string refers to a private function is not ideal.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/k_blocking.py
philip-paul-mueller: I removed them where it is safe to do.
I also asked them and they say that it was not specifically designed in that way. However, they do not expect to change it.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace/transformation_tests/conftest.py
philip-paul-mueller: @egparedes I do not understand what I should put in the comment?

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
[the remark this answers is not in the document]
philip-paul-mueller: You said that I should not mark threads as resolved if I do not fully accept them.
It probably clicked on the wrong button or so.

But can _I_ mark this threads (or other threads such as this one) as resolved?

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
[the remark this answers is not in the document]
egparedes: Ooops, I didn't want to remove the leading underscore, it was just a typo from my side, I'm sorry. I only meant to change the name of the function for something more meaningful, not changing its visibility.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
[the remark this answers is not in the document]
egparedes: In general I think you should only mark threads as resolved if you fully accept the suggestion from the reviewer and don't have further comments, so there is no need for the reviewer to look at it again. In the case you reply something, even if you agree with the suggestion, you should not mark the conversation as resolved because it will be collapsed and the reviewer won't see your comment.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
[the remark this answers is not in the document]
egparedes: Like it or not, that is the syntax of a ternary operator in Python. My point here was trying to make more explicit that there are two exclusive branches in the code, which is simpler to see in any `if-then-else` construction (either a ternary operator or a explicit `if-then-else` block) rather than in a `if-then_with_return; return` construction. But it's not important, feel free to ignore.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/loop_blocking.py
```diff
@@ -0,0 +1,607 @@
...
+            outer_exit=outer_exit,
+            inner_entry=inner_entry,
+            inner_exit=inner_exit,
+            independent_nodes=independent_nodes,
+            dependent_nodes=dependent_nodes,
+            state=graph,
+            sdfg=sdfg,
+        )
+
+    def _rewire_map_scope(
```
egparedes, line 147: If this function doesn't need to be a method because it doesn't use `self`, it should be declared as a module-level function outside of the class. Check here:  https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators
 > Never use staticmethod unless forced to in order to integrate with an API defined in an existing library. Write a module-level function instead.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/loop_blocking.py
```diff
@@ -0,0 +1,607 @@
...
+
+        # After the independent set is computed compute the set of dependent nodes
+        #  as the set of all nodes adjacent to `outer_entry` that are not dependent.
+        dependent_nodes: set[dace_nodes.Node] = {
+            edge.dst for edge in state.out_edges(outer_entry) if edge.dst not in independent_nodes
+        }
+
+        return (independent_nodes, dependent_nodes)
+
+    def classify_node(
```
egparedes, line 473: If this function doesn't need to be a method because it doesn't use `self`, it should be declared as a module-level function outside of the class. Check here:  https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators
 > Never use staticmethod unless forced to in order to integrate with an API defined in an existing library. Write a module-level function instead.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/loop_blocking.py
```diff
@@ -0,0 +1,607 @@
...
+            outer_entry: The entry of the map that should be partitioned.
+            blocking_parameter: The iteration parameter that should be blocked.
+            independent_nodes: The set of nodes that was already classified as
+                independent, in which case it is added to `independent_nodes`.
+            state: The state containing the map.
+            sdfg: The SDFG that is processed.
+        """
+
+        # We are only able to handle certain kind of nodes, so screening them.
+        if isinstance(node_to_classify, dace_nodes.Tasklet):
```
egparedes, line 512: Don't need to change anything here, just a comment for the future: I recommend to use structural pattern matching for this kind of situations where several conditional branches depend on the type of some value (e.g.: 
https://stackoverflow.com/questions/67524641/convert-multiple-isinstance-checks-to-structural-pattern-matching )

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/util.py
egparedes: I still think that reading something like `save this node if it's either a tasklet or we want to save every kind of node ` is the simplest way to put it:
```python
elif if isinstance(next_node, dace_nodes.Tasklet) or not only_tasklets:
    found.add((next_node, curr_edge))
```

egparedes (changes requested):
Final round of comments. Please merge `main` back in the PR.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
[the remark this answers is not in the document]
philip-paul-mueller: Does this mean that, If you made a comment and I replied something, then after the next round, the thread is still open, but you have not written anything to it, I can close it.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/loop_blocking.py
philip-paul-mueller: You are right it does not _need_ to be a method. However, making it a free function feels wrong to me, as the functionality is now split, as some parts are implemented in the class and one (super important one) is a free function. For me this does not make any sense.
For me a class group stuff together and now you tear it apart. It would be different, if the free function could potentially be used somewhere else, but the function is super specific.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/loop_blocking.py
philip-paul-mueller: See my point above regarding `_rewire_map_scope()`. However, there is an inconsistency it should be private.

egparedes (approved):
LGTM.

docs/development/ADRs/0018-Canonical_SDFG_in_GT4Py_Transformations.md
```diff
@@ -0,0 +1,130 @@
...
+   - [Note]: An SDFG can still be constructed using different access node for the same underlying data; _simplify()_ will combine them.
+
+9. Every access node that reads from an array (having an outgoing edge) that was not written to in the same state must be a source node.
+
+   - [Rationale]: Together with rule 1, 4, 5, 6, 7 and 8 this simplifies checking if a transient can be safely removed or if it is used somewhere else.
+     These rules guarantee that the number of "interstate transients" remains constant and this set is given by the _set of source nodes and all access nodes that have an outgoing degree larger than one_.
+   - [Note]: To prevent some issues caused by the violation of rule 4 by _simplify()_, this set is extended with the transient sink nodes and all scalars.
+     Excess interstate transients, that will be kept alive that way, will be removed by later calls to _simplify()_.
+
+10. Every AccessNode within a map scope must refer to a data descriptor whose lifetime must be `dace.dtypes.AllocationLifetime.Scope` and its storage class should _preferably_ be `dace.dtypes.StorageType.Register`.
```
edopao, line 105: Regarding the `storage_type`: I always use default storage while building the SDFG. I believe it is responsibility of the transformations to set the optimal `storage_type`.

docs/development/ADRs/0018-Canonical_SDFG_in_GT4Py_Transformations.md
```diff
@@ -0,0 +1,130 @@
...
+    - [Rationale 1]: This makes optimizations operating inside maps/kernels simpler, as it guarantees that the AccessNode does not propagate outside.
+    - [Rationale 2]: The storage type avoids the need to dynamically allocate memory inside a kernel.
+
+#### Maps
+
+For maps we assume the following:
+
+11. The names of map variables (iteration variables) follow the following pattern.
+
+    - 11.1: All map variables iterating over the same dimension (disregarding the actual range), have the same deterministic name, that includes the `gtx.Dimension.value` string.
```
edopao, line 115: Remove comma: "the actual range),"

docs/development/ADRs/0018-Canonical_SDFG_in_GT4Py_Transformations.md
```diff
@@ -0,0 +1,130 @@
...
+    - 11.2: The name of horizontal dimensions (`kind` attribute) always end in `__gtx_horizontal`.
+    - 11.3: The name of vertical dimensions (`kind` attribute) always end in `__gtx_vertical`.
+    - 11.4: The name of local dimensions always ends in `__gtx_localdim`.
+    - 11.5: No transformation is allowed to modify the name of an iteration variable that follows rules 11.2, 11.3 or 11.4.
+    - [Rationale]: Without this rule it is very hard to tell which map variable does what, this way we can transmit information from GT4Py to DaCe, see also rule 12.
+
+12. Two map ranges, i.e. the pair map/iteration variable and range, can only be fused if they have the same name _and_ cover the same range.
+    - [Rationale 1]: Because of rule 11, we will only fuse maps that actually makes sense to fuse.
+    - [Rationale 2]: This allows to fusing maps without performing a renaming on the map variables.
```
edopao, line 124: allows to fusing -> allows fusing
performing a renaming on -> renaming

docs/development/ADRs/0018-Canonical_SDFG_in_GT4Py_Transformations.md
```diff
@@ -0,0 +1,130 @@
...
+The following rules especially affects transformations and how they operate:
+
+1. Intrastate transformation and interstate transformations must run separately and can not be mixed in the same (DaCe) pipeline.
+
+   - [Rationale]: As a consequence the number of "interstate transients" (transients that are used in multiple states) remains constant during intrastate transformations.
+   - [Note 1]: It is allowed to run them one after another, as long as they are strictly separated.
+   - [Note 2]: It is allowed for an _intrastate_ transformation to act in a way that allows state fusion by later intrastate transformations.
+   - [Note 3]: The DaCe simplification pass violates this rule, for that reason this pass must always be called on its own, see also rule 2.
+
+2. It is invalid to call the simplification pass directly, i.e. the usage of `SDFG.simplify()` is not allowed. The only valid way to call _simplify()_ is to call the `gt_simplify()` function provided by GT4Py.
```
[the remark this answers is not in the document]
philip-paul-mueller: The file you mentioned does not seem to exists. But I have removed all occurrences of simplify.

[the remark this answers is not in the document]
edopao: It exists in the open PR. Ok, we can call the optimization in the next PR.

docs/development/ADRs/0018-Canonical_SDFG_in_GT4Py_Transformations.md
```diff
@@ -0,0 +1,130 @@
...
+- **Status**: valid
+- **Authors**: Philip Müller (@philip-paul-mueller)
+- **Created**: 2024-08-27
+
+In the context of the implementation of the new DaCe fieldview we decided about a particular form of the SDFG.
+Their main intent is to reduce the complexity of the GT4Py specific transformations.
+
+## Context
+
+The canonical is outlined in this document was mainly designed from the perspective of the optimization pipeline.
```
egparedes, line 16:
```suggestion
The canonical form outlined in this document was mainly designed from the perspective of the optimization pipeline.
```

philip-paul-mueller:
cscs-ci run default

egparedes:
It looks like the tests you were trying to fix are still failing...

edopao (approved):

PR #1596, opened 2024-07-29
ci: Re-enable CI on GH200 nodes

edopao:
Additional change: remove local daemon for CUDA MPS in CI test config, because CUDA MPS should now be supported by the CI workflow.


edopao:
cscs-ci run default

tehrengruber (approved):

PR #1600, opened 2024-08-02
ci: GH200 test instability

edopao:
It was observed that the CI tests on GPU randomly hang, and eventually time out. In an attempt to make the CI stable on GH200 nodes, this PR proposes 2 changes:
- Reduce pytest parallelism, by lowering the num processes from 32 to 16 (note that GPU tests exploit CUDA MPS)
- Reduce the SLURM timeout, in order to early detect when jobs hang (longest job takes 10 min)

Additional change:
- Improve the hash on the Dockerfile for caching of base image: now also include the build arguments

### DIFF
--- a/ci/cscs-ci.yml
+++ b/ci/cscs-ci.yml
@@ -31,12 +31,8 @@ stages:
   # Since the base image name is runtime dependent, we need to carry the value of it to
   # the following jobs via a dotenv file.
   before_script:
-  - DATA_TAG=`mktemp`
-  - cat $DOCKERFILE > $DATA_TAG
   # include build arguments in hash since we use a parameterized Docker file
-  - echo $DOCKER_BUILD_ARGS >> $DATA_TAG
-  - DOCKER_TAG=`sha256sum $DATA_TAG | head -c 16`
-  - rm -f $DATA_TAG
+  - DOCKER_TAG=`echo "$(cat $DOCKERFILE) $DOCKER_BUILD_ARGS" | sha256sum | head -c 16`
   - export PERSIST_IMAGE_NAME=$CSCS_REGISTRY_PATH/public/$ARCH/base/gt4py-ci:$DOCKER_TAG-$PYVERSION
   - echo "BASE_IMAGE_${PYVERSION_PREFIX}=$PERSIST_IMAGE_NAME" >> build.env
   artifacts:
--- a/ci/cscs-ci.yml
+++ b/ci/cscs-ci.yml
@@ -182,8 +182,9 @@ build_py38_image_x86_64:
   variables:
     # Grace-Hopper gpu architecture is not enabled by default in CUDA build
     CUDAARCHS: "90"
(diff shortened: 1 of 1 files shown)


ci/cscs-ci.yml
```diff
@@ -149,7 +149,7 @@ build_py38_image_x86_64:
   variables:
     CRAY_CUDA_MPS: 1
     SLURM_JOB_NUM_NODES: 1
-    SLURM_TIMELIMIT: 120
+    SLURM_TIMELIMIT: 30
```
havogt, line 152: probably can be even 15 min?

edopao: Yes, the longest job is 10 min

edopao:
I need to check with @finkandreas if CUDA MPS on GH200 is configured with multi-GPU support, because that would be desirable in our configuration: https://docs.nvidia.com/deploy/mps/#mps-on-multi-gpu-systems

edopao:
> I need to check with @finkandreas if CUDA MPS on GH200 is configured with multi-GPU support, because that would be desirable in our configuration: https://docs.nvidia.com/deploy/mps/#mps-on-multi-gpu-systems

@finkandreas Please ignore this question, No, it is not relevant to us. From the docs:

> When CUDA_VISIBLE_DEVICES is set before launching the control daemon, the devices will be remapped by the MPS server.

I was hoping that MPS could automatically dispatch the CUDA kernels across different devices, but this is not the case.

ci/cscs-ci.yml
```diff
@@ -178,7 +183,8 @@ build_py38_image_x86_64:
     # Grace-Hopper gpu architecture is not enabled by default in CUDA build
     CUDAARCHS: "90"
     # limit test parallelism to avoid "OSError: too many open files" in the gt4py build stage
-    NUM_PROCESSES: 32
+    # observed gpu tests hang in combination with CUDA MPS
```
havogt, line 186: I don't understand the comment

edopao: Comment updated

ci/cscs-ci.yml
```diff
@@ -31,7 +31,12 @@ stages:
   # Since the base image name is runtime dependent, we need to carry the value of it to
   # the following jobs via a dotenv file.
   before_script:
-  - DOCKER_TAG=`sha256sum $DOCKERFILE | head -c 16`
+  - DATA_TAG=`mktemp`
+  - cat $DOCKERFILE > $DATA_TAG
+  # include build arguments in hash since we use a parameterized Docker file
+  - echo $DOCKER_BUILD_ARGS >> $DATA_TAG
+  - DOCKER_TAG=`sha256sum $DATA_TAG | head -c 16`
+  - rm -f $DATA_TAG
```
havogt, line 39: maybe 
```suggestion
  - DOCKER_TAG=`echo "$(cat $DOCKERFILE) $DOCKER_BUILD_ARGS" | sha256sum | head -c 16`
```

havogt (approved):
Thanks!

PR #1601, opened 2024-08-05
feat[next][dace]: GTIR-to-SDFG lowering of let-lambdas (using NestedSDFG)

edopao:
This PR proposes an alternative design for the lowering of let-lambdas to SDFG, than the design provided in #1589. The previous PR added a separate state for the definition of let-symbols: the disadvantage of that solution is that the generated SDFG does not clearly represent the data dependency. The alternative design proposed in this PR uses a nested SDFG to represent some local symbols in the lambda scope. The lambda function is lowered inside the nested SDFG. We rely on the simplify pass to remove unnecessary nested SDFGs.

The new test case `test_gtir_let_lambda_with_connectivity` would generate the following SDFG in baseline:
image

and with this PR:
image


edopao:
cscs-ci run default

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -317,47 +321,34 @@ def visit_FunCall(
...
         elif cpm.is_call_to(node.fun, "as_fieldop"):
             return gtir_builtin_translators.translate_as_field_op(
-                node, sdfg, head_state, self, let_symbols
+                node, sdfg, head_state, self, reduce_identity
             )
         elif isinstance(node.fun, gtir.Lambda):
-            # We use a separate state to ensure that the lambda arguments are evaluated
-            # before the computation starts. This is required in case the let-symbols
-            # are used in conditional branch execution, which happens in different states.
-            lambda_state = sdfg.add_state_before(head_state, f"{head_state.label}_symbols")
```
edopao, line 335: This is the most important change of the PR. Previous design was to allocate the results of let-lambdas as global arrays (by global I mean with lifecycle equal to the sdfg scope). The arrays were however initialized in the symbols state. In this PR, the arrays for let-lambdas results only exist in the scope of the let-lambda, which corresponds to the nested SDFG. Besides, within the nested SDFG, the let-lambdas arrays have the same name as the GTIR symbols for let-lambdas arguments, and we do not need to add an extra state.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -227,6 +225,32 @@ def _visit_expression(
...
+        for offset, offset_provider in dace_fieldview_util.filter_connectivities(
+            self.offset_provider
+        ).items():
+            scalar_kind = tt.get_scalar_kind(offset_provider.index_type)
+            local_dim = gtx_common.Dimension(offset, kind=gtx_common.DimensionKind.LOCAL)
+            type_ = ts.FieldType(
+                [offset_provider.origin_axis, local_dim], ts.ScalarType(scalar_kind)
+            )
+            # We store all connectivity tables as transient arrays here; later, while building
+            # the field operator expressions, we change to non-transient (i.e. allocated extrenally)
```
havogt, line 248:
```suggestion
            # the field operator expressions, we change to non-transient (i.e. allocated externally)
```

havogt (approved):
Looks good to me (though I didn't follow all details of building the SDFG...).

PR #1603, opened 2024-08-06
feat[next]: better wrap decorated program and field_operator

havogt:
Just an idea to eventually get autodoc work for GT4Py programs.

Maybe the cleaner way would be to add in indirection in the decorators and wrap them.

We still seem to have this issue https://github.com/sphinx-doc/sphinx/issues/10221, which could be worked-around with something like https://docs.celeryq.dev/en/latest/_modules/celery/contrib/sphinx.html.

### DIFF
--- a/src/gt4py/next/ffront/decorator.py
+++ b/src/gt4py/next/ffront/decorator.py
@@ -60,6 +60,37 @@ from gt4py.next.type_system import type_info, type_specifications as ts, type_tr
 DEFAULT_BACKEND: Callable = None
 
 
+# copied from the standard library, but updating with `object.__setattr__` for frozen dataclasses
+def _frozen_update_wrapper(
+    wrapper, wrapped, assigned=functools.WRAPPER_ASSIGNMENTS, updated=functools.WRAPPER_UPDATES
+):
+    """Update a wrapper function to look like the wrapped function
+
+    wrapper is the function to be updated
+    wrapped is the original function
+    assigned is a tuple naming the attributes assigned directly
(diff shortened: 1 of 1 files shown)


src/gt4py/next/ffront/decorator.py
```diff
@@ -114,12 +145,9 @@ def past_stage(self):
     def __post_init__(self):
         if self.backend is not None and self.backend.transforms_prog is not None:
             self.backend.transforms_prog.past_lint(self.past_stage)
+        _frozen_update_wrapper(self, self.definition)
         return next_backend.DEFAULT_PROG_TRANSFORMS.past_lint(self.past_stage)
 
-    @property
-    def __name__(self) -> str:
-        return self.definition_stage.definition.__name__
-
```
DropD, line 122: This works because `__name__` is in the default `functools.WRAPPER_ASSIGNMENTS` value for the `assigned` arg of `_frozen_update_wrapper()`, right?

DropD:
The current failures on CI are because `FieldOperatorFromPast` inherits `__post_init__` from `FieldOperator`, but has `.definition` set to None. Either `_frozen_update_wrapper()` should be called inside `.from_function`, guaranteeing it is only called when there is actually a function or `FieldOperatorFromPast` will have to override `.__post_init__` to do everything the superclass does except the `_frozen_update_wrapper()` call.

havogt:
TODO check this https://stackoverflow.com/questions/78802585/abbreviating-dataclass-decorator-without-losing-intellisense

havogt:
not immediately needed, will need to find a different solution later.

PR #1604, opened 2024-08-07
test[next]: Enable multi-GPU execution with xdist

edopao:


edopao:
The GPU execution time is definitely not the bottleneck of unittest execution. The parallelization across multiple GPUs is therefore not the highest priority. It would be more interesting to split a GH200 into 4 nodes and be able to run 4 Gitlab jobs in parallel.

PR #1605, opened 2024-08-07
build: update dependencies, remove deprecated types-all

havogt:


src/gt4py/next/ffront/ast_passes/fix_missing_locations.py
```diff
@@ -42,15 +42,15 @@ def generic_visit(self, node: ast.AST) -> ast.AST:
             node = copy.copy(node)
             parent_node = self._parent_nodes[-1]
 
-            node.lineno = parent_node.lineno
-            node.col_offset = parent_node.col_offset
+            node.lineno = parent_node.lineno  # type: ignore[attr-defined] # this is monkey patching as not all `ast.AST` have `lineno` etc
```
tehrengruber, line 45:
```suggestion
            node.lineno = parent_node.lineno  # type: ignore[attr-defined] # setting the attr even if it is not defined is what we want here
```
Monkey patching is updating the behavior of code at run-time here we are only adding an attribute.

src/gt4py/next/ffront/dialect_parser.py
```diff
@@ -110,9 +110,14 @@ def get_location(self, node: ast.AST) -> SourceLocation:
         line_offset = self.source_definition.line_offset
         col_offset = self.source_definition.column_offset
 
+        # seems we monkey patched all AST nodes in `fix_missing_locations`, therefore we can assume all location attributes are defined
```
tehrengruber, line 113:
```suggestion
        # `FixMissingLocations` ensures that all AST nodes have lineno, etc. even for nodes that by default don't have the attribute
        assert hasattr(node, "lineno")
```

tehrengruber (approved):

PR #1607, opened 2024-08-08
Vertical connectivity and KhalfDim implementation

nfarabullini:
Edits for `KhalfDim` and to allow vertical connectivity


havogt:
As we disussed, gtfn doesn't allow connectivities in the vertical. It allows only for one dimension with connectivity and one dimension with Cartesian shifts. Therefore, I'd propose to close this PR.

PR #1609, opened 2024-08-08
feat[next][dace]: Fix GTIR-to-SDFG lowering for failing tests

edopao:
This PR addresses some problems encountered during GTIR/DaCe integration.
- Handle field size symbol as explicit scalar argument
- Support SDFG positional arguments (not only keyword arguments)
- Use explicit subset on source node for target write
- Add test case for copy from one field to itself using an explicit domain subset
- Extend test case with `cond` by using an expression on scalar arguments as if-statement


havogt (approved):
Looks good to me.

PR #1610, opened 2024-08-09
feat[next][dace]: GTIR-to-SDFG lowering of cast_ builtin

edopao:
Add support to lower GTIR like:
`y @ c⟨ IDimₕ: [0, size) ⟩ ← as_fieldop(λ(a) → cast_(·a, float64), c⟨ IDimₕ: [0, size) ⟩)(x);`

Test case added.

### DIFF
--- a/tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace_fieldview.py
+++ b/tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace_fieldview.py
@@ -102,35 +102,38 @@ def make_mesh_symbols(mesh: MeshDescriptor):
     )
 
 
-def test_gtir_copy():
+def test_gtir_cast():
     domain = im.call("cartesian_domain")(
         im.call("named_range")(gtir.AxisLiteral(value=IDim.value), 0, "size")
     )
+    IFTYPE_INT64 = ts.FieldType(IFTYPE.dims, dtype=ts.ScalarType(kind=ts.ScalarKind.INT64))
     testee = gtir.Program(
-        id="gtir_copy",
+        id="test_gtir_cast",
         function_definitions=[],
         params=[
-            gtir.Sym(id="x", type=IFTYPE),
+            gtir.Sym(id="x", type=IFTYPE_INT64),
             gtir.Sym(id="y", type=IFTYPE),
             gtir.Sym(id="size", type=SIZE_TYPE),
         ],
         declarations=[],
         body=[
             gtir.SetAt(
-                expr=im.as_fieldop(im.lambda_("a")(im.deref("a")), domain)("x"),
+                expr=im.as_fieldop(
+                    im.lambda_("a")(im.call("cast_")(im.deref("a"), "float64")), domain
+                )("x"),
                 domain=domain,
                 target=gtir.SymRef(id="y"),
             )
         ],
     )
 
-    a = np.random.rand(N)
-    b = np.empty_like(a)
+    a = np.random.randint(-1000, +1000, N)
+    b = np.empty_like(a, dtype=np.float64)
 
     sdfg = dace_backend.build_sdfg_from_gtir(testee, CARTESIAN_OFFSETS)
 
     sdfg(x=a, y=b, **FSYMBOLS)
-    assert np.allclose(a, b)
+    assert np.allclose(a.astype(np.float64), b)
 
 
 def test_gtir_update():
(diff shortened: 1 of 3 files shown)


tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace_fieldview.py
```diff
@@ -102,35 +102,38 @@ def make_mesh_symbols(mesh: MeshDescriptor):
...
-    a = np.random.rand(N)
-    b = np.empty_like(a)
+    a = np.random.randint(-1000, +1000, N)
+    b = np.empty_like(a, dtype=np.float64)
 
     sdfg = dace_backend.build_sdfg_from_gtir(testee, CARTESIAN_OFFSETS)
 
     sdfg(x=a, y=b, **FSYMBOLS)
-    assert np.allclose(a, b)
+    assert np.allclose(a.astype(np.float64), b)
```
havogt, line 136: I think this will not fail, even if you remove conversion.  Probably there needs to be an extra explicit `assert a.astype(np.float64).dtype == b.dtype`

havogt (changes requested):

tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace_fieldview.py
edopao: Isn't the reason why it does not fail that the integer values are implicitly casted to float, during the comparison? I don't want (and don't need to) verify the type of the output field, because `b.dtype` is defined by the test at line 131.
To provide better test coverage I could apply some operation (e.g. div) on the input array.

havogt: yeah, my proposal was stupid.
Unfortunately, the current test case still passes without the cast (after removing typing checks on GTIR). The best strategy is probably https://github.com/GridTools/gt4py/blob/646425427d4a608aef1c2ee7bb1a24e3b1e1e3f0/tests/next_tests/integration_tests/feature_tests/iterator_tests/test_builtins.py#L326.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace_fieldview.py
```diff
@@ -97,35 +97,41 @@ def make_mesh_symbols(mesh: MeshDescriptor):
     )
 
 
-def test_gtir_copy():
+def test_gtir_cast():
```
havogt, line 100: like 
```
def test_gtir_cast():
    domain = im.call("cartesian_domain")(
        im.call("named_range")(gtir.AxisLiteral(value=IDim.value), 0, "size")
    )
    IFTYPE_FLOAT32 = ts.FieldType(IFTYPE.dims, dtype=ts.ScalarType(kind=ts.ScalarKind.FLOAT32))
    IFTYPE_BOOL = ts.FieldType(IFTYPE.dims, dtype=ts.ScalarType(kind=ts.ScalarKind.BOOL))
    testee = gtir.Program(
        id="test_gtir_cast",
        function_definitions=[],
        params=[
            gtir.Sym(id="x", type=IFTYPE),
            gtir.Sym(id="y", type=IFTYPE_FLOAT32),
            gtir.Sym(id="z", type=IFTYPE_BOOL),
            gtir.Sym(id="size", type=SIZE_TYPE),
        ],
        declarations=[],
        body=[
            gtir.SetAt(
                expr=im.op_as_fieldop("eq", domain)(
                    im.as_fieldop(
                        im.lambda_("a")(im.call("cast_")(im.deref("a"), "float32")), domain
                    )("x"),
                    "y",
                ),
                domain=domain,
                target=gtir.SymRef(id="z"),
            )
        ],
    )

    a = np.ones(N, dtype=np.float64) * np.sqrt(2.0)
    b = a.astype(np.float32)
    c = np.empty_like(a, dtype=np.bool_)

    sdfg = dace_backend.build_sdfg_from_gtir(testee, CARTESIAN_OFFSETS)

    sdfg(x=a, y=b, z=c, **FSYMBOLS)
    np.testing.assert_array_equal(c, True)
```

havogt (changes requested):

tests/next_tests/unit_tests/program_processor_tests/runners_tests/test_dace_fieldview.py
edopao: Good idea, now the test will fail if `cast_` is not applied on `x`.

havogt (approved):
did you intentionally remove the copy test?

edopao:
> did you intentionally remove the copy test?

Yes, the test was not useful because it used `as_fieldop` to do element-wise copy of the field. Besides, in #1609 I will introduce another kind of copy test which covers an interesting corner case of lowering to SDFG.

PR #1611, opened 2024-08-09
feat[next][dace]: GTIR-to-SDFG lowering of tuple return

edopao:
Add support to lower GTIR like:
```
{x, y} @ c⟨ IDimₕ: [0, size) ⟩
       ← {as_fieldop(λ(__arg0, __arg1) → ·__arg0 + ·__arg1, c⟨ IDimₕ: [0, size) ⟩)(x, 1.0), x};
```
Test case added.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
@@ -21,7 +21,8 @@ from __future__ import annotations
 
 import abc
 import dataclasses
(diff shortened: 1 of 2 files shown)


edopao:
@havogt This is the current status of tuple support in the lowering to SDFG. I think it already covers most of the cases (if not all) we need for initial support of icon4py stencils. Adding support for tuples as temporary results on intermediate expressions requires a major refactoring, which I could deliver later.

edopao:
There will be soon a new PR that contains the full support for tuples, not limited to program arguments.

PR #1615, opened 2024-08-15
feat[next]: allow import of Dimensions from modules within gt4py code

nfarabullini:
Allow to import module attributes directly in gt4py code, e.g. in `domain`here:

```python
    @gtx.program
    def prog(f: gtx.Field[Dims[IDim], float], out: gtx.Field[Dims[IDim], float]):
        field_op(f, out=out, domain={dims.IDim: (0, 8)})
```


src/gt4py/next/ffront/func_to_past.py
```diff
@@ -162,6 +162,19 @@ def visit_BinOp(self, node: ast.BinOp, **kwargs: Any) -> past.BinOp:
...
         return past.Name(id=node.id, location=self.get_location(node))
 
+    def visit_Attribute(self, node: ast.Attribute) -> past.AttributeExpr:
+        attr_type = (
+            type_translation.from_value(self.closure_vars[node.attr])
+            if node.attr in self.closure_vars
+            else None
+        )
+        return past.AttributeExpr(
+            attr=node.attr,
```
havogt, line 172: probably
```suggestion
            attr=self.visit(node.attr),
```

src/gt4py/next/ffront/program_ast.py
```diff
@@ -72,12 +72,18 @@ class TupleExpr(Expr):
     elts: list[Expr]
 
 
+class AttributeExpr(Expr):
+    attr: str
+    value: Any
+    type: Optional[ts.TypeSpec] = None  # A003
```
havogt, line 78:
```suggestion
```
this is already in `Expr`

src/gt4py/next/ffront/program_ast.py
```diff
@@ -72,12 +72,18 @@ class TupleExpr(Expr):
...
+
+
 class Constant(Expr):
     value: Any  # TODO(tehrengruber): be more restrictive
 
 
 class Dict(Expr):
-    keys_: list[Name]
+    keys_: list[Union[Name, AttributeExpr]]
```
havogt, line 86: is this needed? I would expect dict is nested instead of the key being an AttributeExpr.

src/gt4py/next/ffront/func_to_past.py
```diff
@@ -162,6 +162,19 @@ def visit_BinOp(self, node: ast.BinOp, **kwargs: Any) -> past.BinOp:
...
+            value=self.visit(node.value),
+            type=attr_type,
```
havogt, line 174: Isn't that already
```suggestion
            type=self.closure_vars[node.attr].type,
```

src/gt4py/next/ffront/func_to_past.py
nfarabullini: `node.attr`is a string. Do we need a visit for that?

src/gt4py/next/ffront/program_ast.py
nfarabullini: in `func_to_past` Dict keys can come in the form of Name or Attribute, respectively:
`domain={CellDim: (0, 10)}``
`domain={dims.CellDim: (0, 10)}`

src/gt4py/next/ffront/func_to_past.py
nfarabullini: In this specific case `self.closure_vars[node.attr]` is:
`Dimension(value='IDim', kind=<DimensionKind.HORIZONTAL: 'horizontal'>)`
Hence it does not have a type attribute. Also `attr_type` is needed because of the lines above, i.e.
```
        attr_type = (
            type_translation.from_value(self.closure_vars[node.attr])
            if node.attr in self.closure_vars
            else None
        )
```

src/gt4py/next/type_system/type_translation.py
```diff
@@ -204,6 +204,16 @@ def from_value(value: Any) -> ts.TypeSpec:
         elems = [from_value(el) for el in value]
         assert all(isinstance(elem, ts.DataType) for elem in elems)
         return ts.TupleType(types=elems)  # type: ignore[arg-type] # checked in assert
+    elif isinstance(value, typing.types.ModuleType):  # type: ignore[attr-defined] # correct import
```
havogt, line 207:
```suggestion
    elif isinstance(value, types.ModuleType):
```

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_import_from_mod.py
```diff
@@ -0,0 +1,36 @@
+# GT4Py - GridTools Framework
+#
+# Copyright (c) 2014-2024, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+from tests.next_tests.integration_tests.feature_tests.ffront_tests import dimensions_mod as dims
+from tests.next_tests.integration_tests.cases import IDim, KDim
```
havogt, line 10:
```suggestion
```
then see comment at `domain=...`

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_import_from_mod.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def test_import_dims_module(cartesian_case):
+    @gtx.field_operator
+    def mod_op(f: cases.IKField) -> cases.IKField:
+        return f
+
+    @gtx.program
+    def mod_prog(f: cases.IKField, out: cases.IKField):
+        mod_op(f, out=out, domain={dims.IDim: (0, 8), dims.KDim: (0, 3)})
```
havogt, line 29:
```suggestion
        mod_op(f, out=out, domain={cases.IDim: (0, 8), cases.KDim: (0, 3)})
```
otherwise, the definitions in `cases` for `IKField` might not match the dimensions defined in `dimensions_mod`. If you want to isolate the dimensions in a module you could probably define `dimensions_mod` as

```
from tests.next_tests.integration_tests import cases

IDim = cases.IDim
KDim = cases.KDim
```

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_import_from_mod.py
havogt: Something is still wrong in the implementation (didn't investigate yet), because if I remove the second line `from tests.next_tests.integration_tests.cases import IDim, KDim` which is not used in this file, the test breaks.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_import_from_mod.py
```diff
@@ -0,0 +1,36 @@
...
+from tests.next_tests.integration_tests.cases import IDim, KDim
+import gt4py.next as gtx
+import numpy as np
+from tests.next_tests.integration_tests import cases
+from tests.next_tests.integration_tests.cases import cartesian_case
+
+from tests.next_tests.integration_tests.feature_tests.ffront_tests.ffront_test_utils import (
+    exec_alloc_descriptor,
+    mesh_descriptor,
+)
```
havogt, line 19:
```suggestion
from next_tests.integration_tests import cases
from next_tests.integration_tests.cases import cartesian_case

from next_tests.integration_tests.feature_tests.ffront_tests.ffront_test_utils import (
    exec_alloc_descriptor,
    mesh_descriptor,
)
```
the top-level tests doesn't have an `__init__.py` (not sure why, but maybe there was a reason)

src/gt4py/next/ffront/func_to_past.py
havogt: makes sense to not visit the `str`, I thought it's an `ast.Name`.

src/gt4py/next/ffront/func_to_past.py
```diff
@@ -162,6 +162,19 @@ def visit_BinOp(self, node: ast.BinOp, **kwargs: Any) -> past.BinOp:
     def visit_Name(self, node: ast.Name) -> past.Name:
         return past.Name(id=node.id, location=self.get_location(node))
 
+    def visit_Attribute(self, node: ast.Attribute) -> past.AttributeExpr:
```
havogt, line 165: For safety it could make sense to check that `ctx` is `Load` and give a proper error message in the other cases, see https://docs.python.org/3/library/ast.html#ast.Attribute for context.

src/gt4py/next/ffront/func_to_past.py
havogt: - Actually, I realize now, the type should be only deduced in type_deduction, so here it should be left empty.
- But, the type here is actually wrong, because `self.closure_vars[node.attr]` is looking for a top-level name `node.attr`, that's why the tests break if you remove the top-level import, see my comment in the test. Correct would be to do something like `getattr(self.closure_vars[node.value.id], node.attr)`.

src/gt4py/next/ffront/program_ast.py
havogt: I see!

src/gt4py/next/ffront/func_to_past.py
nfarabullini: Makes sense. Should this be down here on in `type_deduction` ?

havogt: I would do it here, then we don't need to carry around the `ctx` information.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_import_from_mod.py
```diff
@@ -0,0 +1,46 @@
...
+        f_i_k = broadcast(f, (cases.IDim, cases.KDim))
+        return f_i_k
+
+    @gtx.program
+    def mod_prog(f: cases.IField, out: cases.IKField):
+        mod_op(
+            f,
+            out=out,
+            domain={
+                ffront_tests.ffront_test_utils.IDim: (0, 8),
```
nfarabullini, line 34: Add a comment that this is in purpose, do not change

nfarabullini, line 34:
```suggestion
                integration_tests.cases.IDim: (0, 8),
```

src/gt4py/next/type_system/type_translation.py
```diff
@@ -204,9 +219,15 @@ def from_value(value: Any) -> ts.TypeSpec:
...
+    elif isinstance(value, types.ModuleType):
+        return UnknownPythonObject(_object=value)
     else:
         type_ = xtyping.infer_type(value, annotate_callable_kwargs=True)
-        symbol_type = from_type_hint(type_)
+        try:
+            symbol_type = from_type_hint(type_)
+        except exceptions.TypeError_:
+            # valid unknown objects could be ModuleType, classes etc that are used to extract typeable objects
+            symbol_type = UnknownPythonObject(_object=value)
```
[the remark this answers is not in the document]
nfarabullini: Yes I have this change locally already, I am not done

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_import_from_mod.py
```diff
@@ -0,0 +1,87 @@
...
+            return f
+
+    with pytest.raises(gtx.errors.DSLError):
+
+        @gtx.field_operator
+        def field_op(f: cases.IField):
+            f_new = dummy_module.field_op_sample(f)
+            return f_new
+
+    with pytest.raises(TypeError):
```
havogt, line 79: Let's make that a DSLError because this should always be an error.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_import_from_mod.py
```diff
@@ -0,0 +1,87 @@
...
+    out = cases.allocate(cartesian_case, mod_prog, "out")()
+    expected = np.zeros_like(out.asnumpy())
+    expected[0:8, 0:3] = np.reshape(np.repeat(f.asnumpy(), out.shape[1], axis=0), out.shape)[
+        0:8, 0:3
+    ]
+
+    cases.verify(cartesian_case, mod_prog, f, out=out, ref=expected)
+
+
+def test_import_module_errors(cartesian_case):
```
havogt, line 57: let's split into 2 tests, one category for tests that should eventually work; the other for things that should always be an error.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_import_from_mod.py
```diff
@@ -0,0 +1,87 @@
...
+    f = cases.allocate(cartesian_case, mod_prog, "f")()
+    out = cases.allocate(cartesian_case, mod_prog, "out")()
+    expected = np.zeros_like(out.asnumpy())
+    expected[0:8, 0:3] = np.reshape(np.repeat(f.asnumpy(), out.shape[1], axis=0), out.shape)[
+        0:8, 0:3
+    ]
+
+    cases.verify(cartesian_case, mod_prog, f, out=out, ref=expected)
+
+
```
havogt, line 56: `@pytest.mark.checks_specific_error` can be used to skip tests in embedded that don't fail with an error.

tests/next_tests/unit_tests/type_system_tests/dummy_package/__init__.py
```diff
@@ -0,0 +1,9 @@
+# GT4Py - GridTools Framework
```
havogt, line 1: let's maybe move this to the top level `test/next_tests` as it's now used in both integration_tests and unit_tests or duplicate with only the relevant stuff inside.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_import_from_mod.py
```diff
@@ -0,0 +1,117 @@
...
+        def field_op(f: cases.IField):
+            f_new = dummy_module.field_op_sample(f)
+            return f_new
+
+    with pytest.raises(gtx.errors.DSLError):
+
+        @gtx.field_operator
+        def field_op(f: cases.IField):
+            f_new = field_op_sample(f)
+            return f_new
```
nfarabullini, line 87:
```suggestion
            return f
```

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_import_from_mod.py
```diff
@@ -0,0 +1,114 @@
...
+            return f
+
+        @gtx.program
+        def field_op(f: cases.IField):
+            dummy_module.field_op_sample(f, out=f, offset_provider={})
+
+
+@pytest.mark.checks_specific_error
+def test_import_module_errors(cartesian_case):
+    with pytest.raises(gtx.errors.DSLError):
```
havogt, line 93: The error that I want to see as a user is something like "Cannot capture 'XXX'.", where `XXX` in this case should probably be the type, i.e. `FieldType`. The error "Module imports of Fields not accepted." is a bit too low level.
The following test case should report the same error 
```
        new_field = gtx.as_field([cases.IDim], np.ones((10,), dtype=gtx.int32))

        @gtx.field_operator(backend=cartesian_case.executor)
        def field_op():
            return new_field
```
One option would be to skip this test for now and open an issue as the solution is probably unrelated to the PR.

Let me know if you want to discuss...

havogt (approved):
Looks good to me! Thanks!

PR #1616, opened 2024-08-19
feat[next]: Prototype offsets as symbols

havogt:
This came up in a discussion with Till, the problem is that itir.shift doesn't make this straight-forward...


havogt:
Closing outdated PRs, if you think it's still relevant, feel free to reopen.

PR #1622, opened 2024-08-30
bug[next]: enable scalar math operations in embedded

havogt:


src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -226,10 +227,12 @@ def astype(
...
-        # TODO(havogt): enable once we have a failing test (see `test_math_builtin_execution.py`)
-        # assert core_defs.is_scalar_type(value) # default implementation for scalars, Fields are handled via dispatch # noqa: ERA001 [commented-out-code]
-        # return getattr(math, name)(value)# noqa: ERA001 [commented-out-code]
-        raise NotImplementedError()
+        # TODO(havogt): enable tests in `test_math_builtin_execution.py`
+        assert core_defs.is_scalar_type(
+            value
+        )  # default implementation for scalars, Fields are handled via dispatch
+
+        return getattr(math, name)(value)
```
egparedes, line 235: This will fail for all functions without names being part of the standard `math` module, which in the case of the gt4py unary builtins I think is all `arc*` functions (`arcsin`, `arccos`, ...) since the correct name is just `a*` (`asin`, `acos`,...). The wrong naming was probably inherited from NumPy, but Data Array API corrects this and uses the proper names, so I suggest to also change the gt4py builtins accordingly.

egparedes (changes requested):

egparedes:
cscs ci run

egparedes:
Closed in favor of #1622

PR #1628, opened 2024-09-04
feat[next][dace]: Build clean nestedSDFG without unused data connectors

edopao:
The main objective of this PR is to reduce the GT4Py dependency on DaCe (especially on DaCe transformations) to produce clean SDFGs. Instead of relying on the `PruneConnectors` pass to remove unused data connectors in nested SDFGs, these connectors can be avoided in the first place, when lowering let-lambdas to nested SDFGs.


philip-paul-mueller (approved):
LGTM

PR #1633, opened 2024-09-06
feat[next][dace]: generic scalar expressions as fieldop args

edopao:
This PR adds support for symbolic expressions as field-operator arguments. This is needed to support expressions like `(0 - 4)` which is used in GTIR to represent the negative value `-4`.


edopao:
I will update this PR with support for generic scalar arguments.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -369,6 +369,89 @@ def translate_literal(
...
+    connectors = []
+    scalar_expr_args = []
+
+    for arg_expr in node.args:
+        is_symbol = True
+        if isinstance(arg_expr, gtir.SymRef):
+            try:
+                sdfg_builder.get_symbol_type(arg_expr.id)
+            except KeyError:
+                is_symbol = False
```
philip-paul-mueller, line 392: A suggestion, feel free to ignore.
But Would it make sense to create a helper function to test this, i.e. `sdfg_builder.is_symbol()`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -369,6 +369,89 @@ def translate_literal(
...
+
+        if is_symbol:
```
philip-paul-mueller, line 394: With symbol do you mean a DaCe symbol?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -369,6 +369,89 @@ def translate_literal(
...
+                isinstance(arg_type, ts.ScalarType)
+                and isinstance(arg_node.desc(sdfg), dace.data.Scalar)
+            ):
+                raise ValueError(f"Invalid argument to scalar expression {arg_expr}.")
+            param = f"__in_{arg_node.data}"
+            args.append(arg_node)
+            connectors.append(param)
+            scalar_expr_args.append(gtir.SymRef(id=param))
+        else:
+            # this is the case of reverved symbols, e.g. 'float64' used as type argument to `cast_`
```
philip-paul-mueller, line 413:
```suggestion
            # this is the case of reserved symbols, e.g. 'float64', used as type argument to `cast_`
```

I think the distinction here is quite critical and important. I would rephrase it to something like "non variable argument, e.g. target type, such as `float64`, in a casting expression, `cast_(vaiable, float64)`."
To make it more clear what you do here.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -369,6 +369,89 @@ def translate_literal(
...
+
+    # we visit the scalar expression replacing the input arguments with the corresponding data connectors
+    scalar_node = gtir.FunCall(fun=node.fun, args=scalar_expr_args)
+    python_code = gtir_python_codegen.get_source(scalar_node)
+    tasklet_node = sdfg_builder.add_tasklet(
+        "scalar_expr",
+        state,
+        set(connectors),
+        {"__out"},
+        f"__out = {python_code}",
```
philip-paul-mueller, line 425: I would suggest the use of named arguments here.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -379,17 +462,20 @@ def translate_symbol_ref(
     """Generates the dataflow subgraph for a `ir.SymRef` node."""
     assert isinstance(node, gtir.SymRef)
 
-    sym_value = str(node.id)
-    sym_type = sdfg_builder.get_symbol_type(sym_value)
+    sym_name = str(node.id)
+    sym_type = sdfg_builder.get_symbol_type(sym_name)
```
philip-paul-mueller, line 466: The change from value to name is quite significant, can you elaborate why, I am just curious.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -405,11 +407,7 @@ def visit_Lambda(
         # that is only the symbols that are used in the context of the lambda node
```
philip-paul-mueller, line 407:
```suggestion
        # that are only the symbols that are used in the context of the lambda node
```
I am not that good at grammar but I think it should be plural here.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -405,11 +407,7 @@ def visit_Lambda(
...
         self._add_sdfg_params(
             nsdfg,
-            [
-                gtir.Sym(id=p_name, type=p_type)
-                for p_name, p_type in lambda_symbols.items()
-                if p_name in used_symbols
-            ],
+            [gtir.Sym(id=p_name, type=p_type) for p_name, p_type in lambda_symbols.items()],
```
philip-paul-mueller, line 410: I remember some discussions here, why did you change from the used to the lambda symbols.
For me the used symbols is smaller and thus more appropriate as it would not include the ones that are not used.

philip-paul-mueller (commented):
There are no big changes, just some small suggestions and some questions. I think we can merge after that round.

As requested I skipped the changes in `type_synthesizer`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: No, I mean a symbol defined in the GT4Py program. Do you have any name suggestion how to distinguish them? It is probably confusing everywhere in this code. Maybe `gt_symbol` and `dc_symbol`?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: Good point. On the other hand, I was considering to add this additional check that anyway requires `get_symbol_type`:
```suggestion
        is_symbol = True
        if isinstance(arg_expr, gtir.SymRef):
            try:
                symbol_type = sdfg_builder.get_symbol_type(arg_expr.id)
                assert isinstance(symbol_type, ts.ScalarType)
            except KeyError:
                is_symbol = False
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: Good point.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: Agreed

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: `node.id` is the symbol name, here meaning a symbol defined in the GT4Py program.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
edopao: Correct. I noticed that even filtering on `used_symbols` in combination with tuples in let-statements (coming PR) does not necessarily remove all redundant connectors, and the `PruneUnusedConnectors` transformation is anyway needed. So I thought I could just ignore the filtering here and let the `PruneUnusedConnectors` transformation do all job. And we agreed that it will be part of the `gt_simplify` function.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
philip-paul-mueller: I agree. I like `gt_symbol` very much, but `dc_symbol` not so much, however, since I have no better idea, I would go along with it.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -369,6 +369,93 @@ def translate_literal(
...
+                isinstance(arg_type, ts.ScalarType)
+                and isinstance(arg_node.desc(sdfg), dace.data.Scalar)
+            ):
+                raise ValueError(f"Invalid argument to scalar expression {arg_expr}.")
+            param = f"__in_{arg_node.data}"
+            args.append(arg_node)
+            connectors.append(param)
+            scalar_expr_args.append(gtir.SymRef(id=param))
+        else:
+            assert isinstance(arg_expr, gtir.SymRef)
```
philip-paul-mueller, line 418: Was the removal of this comment intentional?

philip-paul-mueller (approved):
LGTM.
There is one point with the comment.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: I moved the comment above, inside the `except` block.

PR #1634, opened 2024-09-08
feat[next]: Add missing implementation for cond-builtin function

edopao:
This PR adds support for `cond` in GTIR-embedded backend and in string pretty-parser.

### DIFF
--- a/src/gt4py/next/iterator/embedded.py
+++ b/src/gt4py/next/iterator/embedded.py
@@ -1664,7 +1664,7 @@ def as_fieldop(fun: Callable, domain: runtime.CartesianDomain | runtime.Unstruct
 
 @builtins.cond.register(EMBEDDED)
 def cond(cond_: common.Field | Scalar, t: tuple | TupleField, f: tuple | TupleField):
-    if isinstance(cond_, common.Field) and len(cond_.domain.dims) == 0:
+    if isinstance(cond_, common.Field) and cond_.domain.ndim == 0:
         # promote zero-dimensional field to scalar
         cond_ = cond_.as_scalar()
     return t if cond_ else f
--- a/src/gt4py/next/iterator/embedded.py
+++ b/src/gt4py/next/iterator/embedded.py
@@ -334,6 +334,13 @@ def can_deref(it):
 
 
 @builtins.if_.register(EMBEDDED)
+def cond(cond_, t, f):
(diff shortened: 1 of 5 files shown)


edopao:
Not ready yet.

src/gt4py/next/iterator/embedded.py
```diff
@@ -1662,6 +1662,13 @@ def impl(*args):
     return impl
 
 
+@builtins.cond.register(EMBEDDED)
+def cond(cond_, t, f):
+    if hasattr(cond_, "as_scalar"):
```
tehrengruber, line 1667: As discussed: In case this is for 0d fields something along the lines of 
```suggestion
    if isinstance(cond_, Field) and cond_.ndim == 0:
```
would greatly improve readability. Otherwise a comment explaining this special casing would be nice.

src/gt4py/next/iterator/embedded.py
```diff
@@ -1662,6 +1662,13 @@ def impl(*args):
     return impl
 
 
+@builtins.cond.register(EMBEDDED)
+def cond(cond_, t, f):
```
tehrengruber, line 1666: Many of the builtins miss this, but can we add type annotations easily?

tehrengruber (changes requested):
Small things

src/gt4py/next/iterator/embedded.py
edopao: I copied this code from previous cycle's integration branch. @havogt Can you confirm that this check was meant for 0-dim fields?

src/gt4py/next/iterator/embedded.py
```diff
@@ -1662,6 +1662,14 @@ def impl(*args):
     return impl
 
 
+@builtins.cond.register(EMBEDDED)
+def cond(cond_: common.Field | Scalar, t: tuple | TupleField, f: tuple | TupleField):
+    if isinstance(cond_, common.Field) and len(cond_.domain.dims) == 0:
```
tehrengruber, line 1667:
```suggestion
    if isinstance(cond_, common.Field) and cond_.domain.ndims == 0:
```

tehrengruber (approved):

edopao:
We will replace `cond` with `if_`, so `if_` will work on both scalar and field predicates. This PR is not needed.

PR #1635, opened 2024-09-09
test[dace]: Add skip rules for dace optional dependency

edopao:
Tests that depend on dace should check if dace is installed, because dace is an optional dependency. Error detected in spack-c2sm CI where dace is not installed.
Additionally, added `requires_dace` marker to filter the testcases.


philip-paul-mueller (approved):
LGTM.

Just a question, you also removed `gtir_ts` from `test_dace_fieldview` was it unused in the test before, since it passes?

edopao:
> LGTM.
> 
> Just a question, you also removed `gtir_ts` from `test_dace_fieldview` was it unused in the test before, since it passes?

Correct.

PR #1637, opened 2024-09-10
test[dace]: Fix skip rules for dace optional dependency 

edopao:
This PR fixes the solution to skip testcases dependent on dace provided in #1635. The problem was that `pytest.skip` function was called within `__init__` and/or `conftest` files in the test folder.

Additional changes:
- Moved all dace-related tests in `dace_tests` package folder.
- Introduced pytest marker `requires_dace` at package level, which needs to be included in all test files.
- Run ruff on test code in `dace_tests` package to cleanup and sort the import list.


edopao:
@philip-paul-mueller Tested with Spack-build.

philip-paul-mueller (approved):
LGTM

PR #1638, opened 2024-09-11
bug[next]: Fix implicit offset provider

tehrengruber:
#1484 introduced support for writing `field(I+1)`. The implicit offset providers we generate for this accidentally also included local dimensions, which broke e.g. Icon4Py with errors like:
```
                    if not dim.kind == common.DimensionKind.VERTICAL:
>                       raise ValueError(
                            "Mapping an offset to a horizontal dimension in unstructured is not allowed."
                        )
E                       ValueError: Mapping an offset to a horizontal dimension in unstructured is not allowed.
```
This PR fixes that by only including horizontal and vertical, but not local dimensions in the implicit offset providers.

### DIFF
--- a/src/gt4py/next/ffront/decorator.py
+++ b/src/gt4py/next/ffront/decorator.py
@@ -208,9 +208,10 @@ class Program:
         for param in params:
             if isinstance(param.type, ts.FieldType):
                 for dim in param.type.dims:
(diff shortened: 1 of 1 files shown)


SF-N (approved):
Thanks, this looks good to me.

PR #1641, opened 2024-09-12
chore: Add pre-commit hook for the tach tool to control dependencies …

egparedes:


tach.toml
```diff
@@ -0,0 +1,39 @@
...
+forbid_circular_dependencies = true
+source_roots = [
+  "src"
+]
+
+[[modules]]
+depends_on = [
+  {path = "gt4py.eve"}
+]
+path = "gt4py._core"
```
havogt, line 12: I guess the order is coming from the formatter?

tach.toml
```diff
@@ -0,0 +1,39 @@
...
+  {path = "gt4py._core"},
+  {path = "gt4py.eve"},
+  {path = "gt4py.storage"}
+]
+path = "gt4py.next"
+
+[[modules]]
+depends_on = [
+  {path = "gt4py._core"},
+  {path = "gt4py.cartesian"},
```
havogt, line 36:
```suggestion
  {path = "gt4py.cartesian"}, # for backward-compatibility the cartesian allocators are in `gt4py.storage`
```

havogt (approved):
2 optional comments

tach.toml
egparedes: Yes, exactly, but in this case I don't think it makes sense so I'll exclude this file from the toml autoformatter.

PR #1642, opened 2024-09-16
bug[next]: fix lowering of `astype` on tuples containing scalars

havogt:
The unpacked tuple will contain `as_fieldop`ed `cast_`s for fields, but direct `cast_`s for scalars.

Note: A similar change is not needed for `where` which also works on tuples in the `true` and `false` branches. The reason is that `mask` is required to be a `Field`, therefore `true` and `false` are promoted to `Field`s, too.

### DIFF
--- a/src/gt4py/next/ffront/lowering_utils.py
+++ b/src/gt4py/next/ffront/lowering_utils.py
@@ -113,6 +113,7 @@ def process_elements(
         objs: The object whose elements are to be transformed.
         current_el_type: A type with the same structure as the elements of `objs`. The leaf-types
             are not used and thus not relevant.
+        with_type: If True, the last argument passed to `process_func` will be its type.
     """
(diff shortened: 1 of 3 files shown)


src/gt4py/next/ffront/lowering_utils.py
```diff
@@ -102,6 +102,7 @@ def process_elements(
     process_func: Callable[..., itir.Expr],
     objs: itir.Expr | Iterable[itir.Expr],
     current_el_type: ts.TypeSpec,
+    with_type: bool = False,
```
tehrengruber, line 105: Update to the docstring is missing.

tehrengruber (approved):

src/gt4py/next/ffront/lowering_utils.py
havogt: done

PR #1647, opened 2024-09-18
feat[dace]: Modified `gt_simplify()`

philip-paul-mueller:
Before the pass was just calling the native DaCe version. Now it will not call the `PromoteScalarToSymbol` and `ConstantPropagation` pass. This is because the lowering sometimes has to change between a symbol and a scalar and back and back again and so on. Furthermore, it looks like these passes have problems with that, so we excluded them.

This is a temporary solution, at the end, it might be feasible or good to run the full simplify pass.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
@@ -8,7 +8,7 @@
 
 """Fast access to the auto optimization on DaCe."""
 
-from typing import Any, Final, Optional, Sequence
+from typing import Any, Final, Iterable, Optional, Sequence
 
 import dace
 from dace.transformation import dataflow as dace_dataflow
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -32,19 +43,18 @@ def gt_simplify(
...
-
-    Note:
-        The reason for this function is that we can influence how simplify works.
-        Since some parts in simplify might break things in the SDFG.
-        However, currently nothing is customized yet, and the function just calls
-        the simplification pass directly.
     """
-
+    if skip is None:
+        skip = GT_SIMPLIFY_DEFAULT_SKIP_SET
```
edopao, line 57: We could use `GT_SIMPLIFY_DEFAULT_SKIP_SET` as default instead of `None` on line 39, or is it style-thing?

philip-paul-mueller: Actually we could do that.

edopao (approved):

PR #1649, opened 2024-09-19
feat[dace]: Custom SDFG inline pass

philip-paul-mueller:
Added a custom pass for inlining SDFG.
The function builds upon the traditional inline pass (`InlineSDFG`), however, before it is run some cleaning steps are preformed (`PruneConectors` and `PruneSymbols`) which increase the likelihood that the inlining can be done.
The cost is that state fusing is performed.

Also the `gt_simplify()` function is modified, instead of the build in behaviour it will only run the GT4Py specific one.
This behaviour can not be changed.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
@@ -35,19 +35,19 @@ def gt_simplify(
     sdfg: dace.SDFG,
     validate: bool = True,
     validate_all: bool = False,
-    skip: Iterable[str] = GT_SIMPLIFY_DEFAULT_SKIP_SET,
+    skip: Optional[Iterable[str]] = None,
 ) -> Any:
     """Performs simplifications on the SDFG in place.
 
     Instead of calling `sdfg.simplify()` directly, you should use this function,
     as it is specially tuned for GridTool based SDFGs.
 
-    By default this function will run the normal DaCe simplify pass, but skip
-    passes listed in `GT_SIMPLIFY_DEFAULT_SKIP_SET`. If `skip` is given it will
-    not be modified, i.e. `GT_SIMPLIFY_DEFAULT_SKIP_SET` is not added by default.
+    This function runs the DaCe simplification pass, but the following passes are
+    replaced:
+    - `InlineSDFGs`: Instead `gt_inline_nested_sdfg()` will be called.
 
-    Passes that are replaced:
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -48,18 +47,33 @@ def gt_simplify(
...
+
     Args:
         sdfg: The SDFG to optimize.
         validate: Perform validation after the pass has run.
         validate_all: Perform extensive validation.
         skip: List of simplify passes that should not be applied, defaults
             to `GT_SIMPLIFY_DEFAULT_SKIP_SET`.
     """
-    return dace_passes_simplify.SimplifyPass(
+    if skip is None:
```
edopao, line 60: Probably we should remove the `Optional` type, because anyway we have a default skip set. If instead we have to keep `Optional`, and I see then the need for this if-statement, then we can also move back the default set here.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
```diff
@@ -91,6 +105,58 @@ def gt_set_iteration_order(
...
 
 
+def gt_inline_nested_sdfg(
+    sdfg: dace.SDFG,
+    multistate: bool = True,
+    permissive: bool = False,
+    validate: bool = True,
+    validate_all: bool = False,
+) -> dace.SDFG:
+    """Perform inlining of nested SDFG into their parent SDFG.
```
edopao, line 115: nested SDFG -> nested SDFGs

edopao (commented):
Nice! Just one minor comment.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
philip-paul-mueller: I see your point, but keeping optional would allow for `skip=None` to tell the transformation to not skip anything.
Otherwise we would have to write `skip=[]` or `skip=set()` which is more cryptic. But I removed it, what do you think?
@edopao

edopao (approved):

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_opt.py
philip-paul-mueller: @edopao I made the changes we discussed.

edopao:
LGTM

PR #1656, opened 2024-09-20
fix[next]: Fix usage of DaCe fast-call to SDFG

edopao:
This PR addresses some flaky test failures observed in GT4Py CI. The root cause was that the dace backend did not check the connectivity arrays, which are passed as keyword-arguments to the SDFG. It did only check the positional arguments. The connectivity arrays do not have to be allocated on the device memory: for gpu execution, the backend ensures that the connectivity arrays are copied to device memory just before passing them to the SDFG call. Previous implementation worked sometimes, when by chance cupy was reusing the same array on gpu memory, hence the flaky behavior of the tests.
New test is added for the connectivity case. The previous test case is cleaned up and improved, by invalidating all scalar positional arguments at each SDFG call: this allows to test that they are overridden before fast_call.

Additionally, this PR reduces the overhead of regular SDFG call: previous implementation was copying all the connectivity arrays to gpu memory, with this PR we only allocate cupy arrays for the connectivities used in the SDFG.


src/gt4py/next/program_processors/runners/dace_common/workflow.py
```diff
@@ -25,22 +26,20 @@
 
 class CompiledDaceProgram(stages.CompiledProgram):
     sdfg_program: dace.CompiledSDFG
-    # Map SDFG argument to its position in program ABI; scalar arguments that are not used in the SDFG will not be present.
+    # Map SDFG positional arguments to the position in program ABI;
+    # scalar arguments that are not used in the SDFG will not be present.
     sdfg_arg_position: list[Optional[int]]
+    # Map arguments for connectivity tables to the position in program ABI;
+    # consider only the connectivity arrays, skip shape and stride symbols.
+    sdfg_conn_position: dict[str, int]
```
philip-paul-mueller, line 34: Are you still needing these?
According to my understanding everything is inside `inp.sdfg_arglist`.

src/gt4py/next/program_processors/runners/dace_common/workflow.py
```diff
@@ -25,22 +26,20 @@
...
-        sdfg_arg_pos_mapping = {param: pos for pos, param in enumerate(sdfg_arglist)}
-        sdfg_used_symbols = program.sdfg.used_symbols(all_symbols=False)
-
-        self.sdfg_program = program
-        self.sdfg_arg_position = [
-            sdfg_arg_pos_mapping[param]
-            if param in program.sdfg.arrays or param in sdfg_used_symbols
-            else None
-            for param in program.sdfg.arg_names
+        # method `signature_arglist` returns an ordered dictionary
```
philip-paul-mueller, line 37: You should add what this mapping is? and it is `SDFG.arglist()`
Something like "`SDFG.arglist()` returns a mapping from the argument name to its type, the order is given by the ABI and the same as in `lastargs[0]`.

src/gt4py/next/program_processors/runners/dace_common/workflow.py
```diff
@@ -119,28 +118,34 @@ def decorated_program(
...
-                    sdfg_program._lastargs[0][pos] = _get_ctype_value(arg, desc.dtype)
-                elif pos:
-                    sym_dtype = sdfg.symbols[param]
-                    sdfg_program._lastargs[0][pos] = _get_ctype_value(arg, sym_dtype)
+                elif isinstance(arg_type, dace.data.Scalar):
+                    if arg_name in kwargs:
+                        last_call_args[i] = _get_ctype_value(kwargs[arg_name], arg_type.dtype)
+                    else:
+                        # shape and strides of arrays are supposed not to change, and can therefore be omitted
+                        assert re.match(
```
philip-paul-mueller, line 144: Consider using a precompiled statement.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace.py
```diff
@@ -8,54 +8,81 @@
...
+        for i, (arg_name, arg_type) in enumerate(sdfg_arglist.items()):
+            if arg_name in positional_args and isinstance(arg_type, dace.data.Scalar):
+                self._lastargs[0][i] = None
+        return fast_call_result
+
+    monkeypatch.setattr(dace.codegen.compiled_sdfg.CompiledSDFG, "fast_call", mocked_fast_call)
+
+    # Wrap `compiled_sdfg.CompiledSDFG._construct_args` with mock object
+    mock_construct_args = unittest.mock.MagicMock()
+    mock_construct_args_attr = dace.codegen.compiled_sdfg.CompiledSDFG._construct_args
```
philip-paul-mueller, line 58: This is fully optional, but I think the name is a bit misleading, I would suggest `constrcut_args_impl` or something, because the prefix `mock` implies that it is also in someway decorated or so.
But feel free to ignore.

philip-paul-mueller (commented):
In general it looks good but I have some suggestions.

philip-paul-mueller (approved):
LGTM

PR #1657, opened 2024-09-23
fix: GitHub License Detection

fthaler:
GitHub fails to detect the license of GT4Py correctly (thus we get the ugly banner license). GitHub uses [licensee](https://github.com/licensee/licensee/) to detect the license as documented [here](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#detecting-a-license). This PR fixes license detection using licensee, which previously was broken due to an additional line above the typical BSD-3 copyright notice and the file `LICENSE_HEADER.txt` which confused licensee, too.


egparedes (approved):
LGTM.

PR #1660, opened 2024-09-24
refactor[next]: Cleanup toolchain

tehrengruber:
Over time a significant amount of technical debt in the toolchain workflows accumulated resulting in hardly readable code, slowed progress in refactoring efforts and long review cycles. This PR streamlines some of the interfaces and makes existing technical debt more visible (either by TODOs or refactorings).

- Bugfix to make `gt4py.next.backend.Transforms` actually work from `itir.Program | itir.FencilDefinition` (by skipping `past_to_itir` step in that case)
- Rename `AOTProgram` type alias to `CompilableProgram` (was inconsistent before)
- Fix backend type annotations. In many cases `ppi.ProgramExecutor` was used even though a backend was required.
- Remove entire `ProgramExecutor` protocol. This protocol was not properly used anywhere anymore.
- Remove left-over argument canonicalization in `Program` with embedded execution and instead only validate the arguments.
- For some reason there was a lot of code duplication in `decorator.py` in the form of
  ```python
  if self.backend is not None and self.backend.transforms is not None:
      return self.backend.transforms.<SOME CODE>
  return next_backend.DEFAULT_TRANSFORMS.<SOME CODE>
  ```
  which has been cleaned up into
  ```python
  return next_backend._frontend_transforms.<SOME CODE>
  ```
  The `self.backend.transforms is not None` condition has been dropped. According to the type annotation this could never occur, but `NoBackend` in the tests required this behavior. Instead `NoBackend` now initializes the backend properly using the default transforms and the technical debt is explicitly documented with a TODO at the place it occurs.
- Remove `CompileTimeArg` class. This class looked like an actual value, but wasn't. In many places it was used it was immediately converted to a `TypeSpec`. This was quite confusing as it appeared like values were passed around, while actually it was types only. By using `TypeSpec` directly it is clear what information is available & actually used. Similarly `SizeArg` was also removed and just replaced by a `ScalarType`.
- Removal of `ModularExecutor`. This class served two purposes:
  - The `name` attribute was used to distinguish backends. The `name` is now part of the backends directly avoiding an unnecessary indirection.
  - `__call__` This method duplicated `backend.Backend.__call__` and additionally dropped all keyword arguments (e.g. `column_axis`) except for the offset provider. The method was only used in the iterator tests via `FendefDispatcher`. To make this explicit this was moved to `FendefDispatcher` which uses the backend directly now.
  Removal of the class allowed to remove additional code-duplication in `roundtrip.py`.

Open issues:
- `gt4py.next.backend.Backend.executor`: The class and `executor` & `transforms` attribute should be renamed to better reflect what they are doing.
  - `Backend` -> `Toolchain`
  - `transforms` -> `frontend_transforms`
  - `executor` -> `backend_transforms`
While this is a relatively simple change it touches many lines and would make review of the PR unnecessarily hard. Additionally it would break Icon4Py. Will be tackled in a new PR.


src/gt4py/next/program_processors/runners/roundtrip.py
```diff
@@ -184,13 +181,13 @@ def fencil_generator(
 
 
 @dataclasses.dataclass(frozen=True)
-class Roundtrip(workflow.Workflow[stages.AOTProgram, stages.CompiledProgram]):
-    debug: Optional[bool] = None
+class Roundtrip(workflow.Workflow[stages.CompilableProgram, stages.CompiledProgram]):
+    debug: Optional[bool] = True  # TODO
```
havogt, line 185:
```suggestion
    debug: Optional[bool] = None
```

src/gt4py/next/iterator/runtime.py
```diff
@@ -49,75 +50,81 @@ class UnstructuredDomain(dict): ...
...
+        offset_provider = offset_provider or self.offset_provider
+        column_axis = column_axis or self.column_axis
 
         if backend is not None:
-            ensure_processor_kind(backend, ProgramExecutor)
-            backend(self.itir(*args, **kwargs), *args, **kwargs)
+            itir_node = self.itir(*args)
+
+            # TODO(tehrengruber): remove cirular dependency and place import at the top of the file
+            from gt4py.next import backend as next_backend
```
havogt, line 85: where is it coming from?

tehrengruber: `runtime` -> `backend` -> `ffront` passes -> `ffront.fbuiltins` -> `runtime.Offset` used in `FieldOffset`
`FieldOffset` already has an open that touches on this.

tests/next_tests/integration_tests/feature_tests/iterator_tests/test_builtins.py
```diff
@@ -180,13 +178,14 @@ def test_arithmetic_and_logical_functors_gtfn(builtin, inputs, expected):
...
-    gtfn_executor = run_gtfn.executor
     gtfn_without_transforms = dataclasses.replace(
-        gtfn_executor,
-        otf_workflow=gtfn_executor.otf_workflow.replace(
-            translation=gtfn_executor.otf_workflow.translation.replace(enable_itir_transforms=False)
-        ),
-    )  # avoid inlining the function
+        run_gtfn,
+        executor=dataclasses.replace(
+            run_gtfn.executor,
```
DropD, line 184:
```suggestion
        executor=run_gtfn.executor.replace(
```

Just a nitpick: The `dataclasses.replace` was only necessary for the `ModularExecutor` instance, because that was not a replaceable workflow. (Admittedly small) benefits to using `.replace` would be brevity and consistency.

DropD (changes requested):
Some small things that need to be addressed.

DropD:
> Open issues:
> 
> * `gt4py.next.backend.Backend.executor` should be renamed. This attribute contains already only the frontend transformations, but is strangely still called `executor`, while not doing any execution.

Slight correction: contains the compilation steps (AKA the backend of the backend), while the frontend transformations remain in `Backend.transformations`. This PR makes the transition away from Backends, which contain Executors towards Backends which are executors complete (finally), so it would be the perfect place to rename the attribute if you feel like it!

src/gt4py/next/ffront/past_process_args.py
```diff
@@ -94,7 +96,9 @@ def _process_args(
                         "Constituents of composite arguments (e.g. the elements of a"
                         " tuple) need to have the same shape and dimensions."
                     )
-                size_args.extend(shape if shape else [type_translation.SizeArg()] * len(dims))  # type: ignore[arg-type] ##(ricoh) mypy is unable to correctly defer the type of the ternary expression
+                size_args.extend(
+                    shape if shape else [ts.ScalarType(kind=ts.ScalarKind.INT32)] * len(dims)  # type: ignore[arg-type]  # shape is always empty
+                )
```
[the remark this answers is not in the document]
tehrengruber: Yes, but the shape is always empty so no need to complicate things with an `if` that always evaluates the false branch.

tehrengruber:
> This PR makes the transition away from Backends-which-contain-Executors, towards Backends-which-are-executors complete (finally)

Indeed this is was one of the goals of this PR. In the end I didn't do the final rename as the many changes would have been very distracting in review. Let's do it in another PR.

src/gt4py/next/ffront/past_process_args.py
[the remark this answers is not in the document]
DropD: My point exactly. What stands in the way of removing the wider `if`?

DropD (approved):

PR #1655, opened 2024-09-20
Fix storage module for ROCm compatiblity

egparedes:
## Requirements

- [ ] All fixes and/or new features come with corresponding tests.
- [ ] Important design decisions have been documented in the approriate ADR inside the [docs/development/ADRs/](docs/development/ADRs/Index.md) folder.

If this PR contains code authored by new contributors please make sure:

- [ ] The PR contains an updated version of the `AUTHORS.md` file adding the names of all the new contributors.

PR #1662, opened 2024-09-25
fix[cartesian]: fix bugs in CuPy-ROCm storage allocation

stubbiali:
Fix multiple bugs in the allocation of CuPy-ROCm storages introduced in previous PR #1655 (https://github.com/GridTools/gt4py/pull/1655), including an infinite recursion call in the main allocation function.


src/gt4py/storage/cartesian/utils.py
```diff
@@ -308,17 +308,21 @@ def allocate_gpu(
...
 
 
-if CUPY_DEVICE == core_defs.DeviceType.ROCM:
+if CUPY_DEVICE == core_defs.DeviceType.CUDA:
+    allocate_gpu = _allocate_gpu
+elif CUPY_DEVICE == core_defs.DeviceType.ROCM:
 
-    @functools.wraps(allocate_gpu)
-    def allocate_gpu_rocm(
+    @functools.wraps(_allocate_gpu)
```
egparedes, line 315: I think we could also move here from above the definition of the ndarray subclass adding the CUDA and HIP array interfaces to make the code more readable. Something like:

```suggestion
    class CUDAArrayInterfaceNDArray(cp.ndarray):
        def __new__(cls, input_array: "cp.ndarray") -> CUDAArrayInterfaceNDArray:
            return (
                input_array
                if isinstance(input_array, CUDAArrayInterfaceNDArray)
                else cp.asarray(input_array).view(cls)
            )

        @property
        def __cuda_array_interface__(self) -> dict:
            return {
                "shape": self.shape,
                "typestr": self.dtype.descr[0][1],
                "descr": self.dtype.descr,
                "stream": 1,
                "version": 3,
                "strides": self.strides,
                "data": (self.data.ptr, False),
            }
            
        __hip_array_interface__ = __cuda_array_interface__

    @functools.wraps(_allocate_gpu)
```

egparedes (commented):
Looks good. Just one comment to enhance the PR.

src/gt4py/storage/cartesian/utils.py
```diff
@@ -308,17 +308,21 @@ def allocate_gpu(
...
         alignment_bytes: int,
         aligned_index: Optional[Sequence[int]],
     ) -> Tuple["cp.ndarray", "cp.ndarray"]:
-        buffer, ndarray = allocate_gpu(shape, layout_map, dtype, alignment_bytes, aligned_index)
+        buffer, ndarray = _allocate_gpu(shape, layout_map, dtype, alignment_bytes, aligned_index)
         return buffer, CUDAArrayInterfaceNDArray(ndarray)
 
-    allocate_gpu = allocate_gpu_rocm
+    allocate_gpu = _allocate_gpu_rocm
+else:
```
egparedes, line 327:
```suggestion
elif CUPY_DEVICE is not None:
```

stubbiali:
Thanks for your quick review @egparedes. I did some tests with PMAP on LUMI, and everything looks good. No modifications are needed on the C++ side for now, and only a minor adaptation is needed on the PMAP side.

PR #1667, opened 2024-09-27
refactor[next]: ranaming of dace_utils module

edopao:
This is a pure refactoring PR, that applies the following changes in dace backends:
- renamed alias of `dace_common.utility` module: `dace_util` -> `dace_utils`
- renamed alias of `dace_common.utility` module: `dace_common_util` -> `dace_utils`
- renamed alias of `dace_fieldview.utility` module: `dace_fieldview_util` -> `dace_gtir_utils`
- moved method `as_dace_type` to `dace_common.utility`

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_common/dace_backend.py
+++ b/src/gt4py/next/program_processors/runners/dace_common/dace_backend.py
@@ -14,7 +14,7 @@ import numpy as np
 
 from gt4py.next import common as gtx_common
 
-from . import utility as dace_util
+from . import utility as dace_utils
 
 
 try:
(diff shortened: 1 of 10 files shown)


philip-paul-mueller (approved):
I have nothing per se against this change, I just want to raise the following concern:
The name `dace_utils` (actually even `dace_util`) could be understood as the `utils` module/subpacke from the `dace` package, however, it is not, it is a pure GT4Py thing.
However, I do not have a solution (aka better name).

edopao:
> I have nothing per se against this change, I just want to raise the following concern: The name `dace_utils` (actually even `dace_util`) could be understood as the `utils` module/subpacke from the `dace` package, however, it is not, it is a pure GT4Py thing. However, I do not have a solution (aka better name).

Good point. A more clear name would definitely be a longer name, like `dace_backend_utils`.

PR #1668, opened 2024-09-27
fix[next][dace]: enable subdomain in let-statements

edopao:
This PR fixes the lowering of let-statements with fields defined on sub-domain. The previous implementation was allocating dace array with the exact domain of the computation, eventually not starting from index 0. This required the use of offsets in dace arrays to compensate for the start index. There were two problems with this: 1) in the lowering from GTIR to SDFG because the array offsets were not correctly propagated across nested SDFGs for let-statements; and 2) the use of array offsets caused issue to SDFG-inlining pass. In order to reduce the overall complexity, this PR proposes to remove the usage of array offsets and therefore always allocate temporary fields starting from index 0.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
@@ -160,11 +160,7 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
         return shape, strides
 
     def _add_storage(
-        self,
-        sdfg: dace.SDFG,
-        name: str,
-        symbol_type: ts.DataType,
-        transient: bool = True,
+        self, sdfg: dace.SDFG, name: str, symbol_type: ts.DataType, transient: bool = True
     ) -> None:
         """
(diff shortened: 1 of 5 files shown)


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -399,9 +400,14 @@ def visit_Lambda(
...
-            pname: type_ for pname, (_, type_) in lambda_args_mapping.items()
-        }
+        referenced_symbols = set(
+            symbol_ref_utils.collect_symbol_refs(node.expr, self.global_symbols.keys())
+        )
+        lambda_symbols = {
+            pname: type_
+            for pname, type_ in self.global_symbols.items()
+            if pname in referenced_symbols
+        } | {pname: type_ for pname, (_, type_) in lambda_args_mapping.items()}
```
philip-paul-mueller, line 410: This last line is almost lost and might be overlooked.
I (personally) would put it on its own line, i.e. 
```python
lambda_symbol |= {pname: type_ for pname, (_, type_) in lambda_args_mapping.items()}
```
or using `dict.update`.

PR #1677, opened 2024-10-01
feat[next]: gtir lowering of broadcasted scalars

havogt:
TODO add test


src/gt4py/next/ffront/foast_to_gtir.py
```diff
@@ -373,7 +373,10 @@ def create_if(true_: itir.Expr, false_: itir.Expr) -> itir.FunCall:
     _visit_concat_where = _visit_where  # TODO(havogt): upgrade concat_where
 
     def _visit_broadcast(self, node: foast.Call, **kwargs: Any) -> itir.FunCall:
-        return self.visit(node.args[0], **kwargs)
+        expr = self.visit(node.args[0], **kwargs)
+        if type_info.is_type_or_tuple_of_type(node.args[0].type, ts.ScalarType):
```
havogt, line 377: this might actually be wrong for tuples, think about it...

tehrengruber: Two things that come to mind:
1. This case for example will not work properly: `(field, scalar)`
2. `deref` on a tuple is likely not a pattern we want to pass to dace. It makes sense to me if the backend only needs support for tuples on field level.
3. For the case where everything is actually a scalar gtfn and embedded will happily execute this.
Maybe `process_elements` like in `where` makes sense?

tehrengruber: `broadcast` actually does not accept tuples. I replaced it with a simple `isinstance` check.

PR #1679, opened 2024-10-04
feat[cartesian]: Absolute k indexation

FlorianDeconinck:
Syntax: `field[K_at(...)]
Backends:  `numpy` and `dace:cpu`

Can index with literals, parameters, externals and fields.

ToDo:
- change syntax to `field.at(K=...)`
- more protection & test for unwanted syntax (I, J in particular)


FlorianDeconinck:
Bad remote, sorry

PR #1682, opened 2024-10-08
fix[cartesian]: Protect K write against dynamic access 

FlorianDeconinck:
## Description

We allowed offset-writes in K in a previous PR e.g.
```python
out_field[0, 0, 1] = value
```
but we let open the possibility for the offset to be a dynamic value which would escape any extent analysis. E.g.
```python
index = value ** 29382938
out_field[0, 0, index] = other_value
```

The time might come where a legit case shows up and we could imagine introducing a mitigating strategy. Until that fateful day, we restrict access to literals.

## Requirements


FlorianDeconinck:
Won't do as we actually have the case listed above.

We logged a mitigation task which seems to us the best way to meet feature need and performance https://github.com/GridTools/gt4py/issues/1684

PR #1683, opened 2024-10-09
feat[next][dace]: GTIR-to-DaCe lowering of map-reduce (only full connectivity)

edopao:
This PR adds support for lowering of `map_` and `make_const_list` builtin functions. However, the current implementation only supports neighbor tables with full connectivity (no skip values). The support for skip values will be added in next PR.

To be noted:
- This PR generalizes the handling of tasklets without arguments inside a map scope. The return type for `input_connections` is extended to contain a `TaskletConnection` variant, which is lowered to an empty edge from map entry node to the tasklet node.
- The result of `make_const_list` is a scalar value to be broadcasted on a local field. However, in order to keep the lowering simple, this value is represented as a 1D 1-element array (`shape=(1,)`).


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -709,14 +816,17 @@ def visit_FunCall(self, node: gtir.FunCall) -> IteratorExpr | ValueExpr:
...
         elif cpm.is_applied_reduce(node):
             return self._visit_reduce(node)
 
         elif cpm.is_applied_shift(node):
             return self._visit_shift(node)
 
-        else:
-            assert isinstance(node.fun, gtir.SymRef)
+        elif not isinstance(node.fun, gtir.SymRef):
+            raise NotImplementedError
```
philip-paul-mueller, line 829: I think this function should be split.
In the beginning there is this, let's face it, switch that dispatches the various flavors of a `FunCall` to their respective implementations.
However, the function itself handles the case of `SymRef` that is absolutely unrelated to what the other functions do.
Yo I would suggest moving the code for `SymRef` into its own function and dispatch to it.
It would also make error handling more cleaner.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -761,17 +871,22 @@ def visit_FunCall(self, node: gtir.FunCall) -> IteratorExpr | ValueExpr:
...
 
-        assert isinstance(node.type, ts.ScalarType)
-        dtype = dace_utils.as_dace_type(node.type)
-
-        return self._get_tasklet_result(dtype, tasklet_node, "result")
+        if isinstance(node.type, itir_ts.ListType):
+            # make_const_list returns `ListType` but we represent it as a scalar value
+            assert isinstance(node.type.element_type, ts.ScalarType)
+            dtype = dace_utils.as_dace_type(node.type.element_type)
+            return self._get_tasklet_result(dtype, tasklet_node, "result", use_array=True)
```
philip-paul-mueller, line 878: I have nothing against having multiple `return` statements.
However, I think in this case it would increase the clarity if there is just one `return`.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -1323,63 +1323,86 @@ def test_gtir_reduce_with_skip_values():
...
+            im.op_as_fieldop(im.map_("multiplies"), vertex_domain)(
+                im.as_fieldop_neighbors("V2E", "edges", vertex_domain),
+                im.as_fieldop_neighbors("V2E", "edges", vertex_domain),
+            ),
+            im.op_as_fieldop("make_const_list", vertex_domain)(1.0),
+        )
     )
-    assert np.allclose(v, v_ref)
+
+    for i, stencil in enumerate([stencil_inlined, stencil_fieldview]):
```
philip-paul-mueller, line 1377: For me it looks like there are two different tests in this function.

philip-paul-mueller (commented):
There are some points that needs further clarifications but it does not look that bad.
There are also some suggestions that are not directly related to the PR per se.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -59,6 +59,8 @@ class SymbolExpr:
     dace.nodes.Node,
     Optional[str],
 ]
+# special connection from map entry node to a tasklet node without arguments
+TaskletConnection: TypeAlias = dace.nodes.Tasklet
```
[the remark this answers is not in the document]
edopao: Maybe you missed it, but `InputConnection` is a tuple, not a union. So the following point is wrong:

> The definition of TaskletConnection is redundant, since it is included in InputConnection

Both `InputConnection` and `TaskletConnection` are lowered to edges on the map entry node:

- `InputConnection` is a tuple containing the elements needed to "draw" an edge with a data memlet.
- `TaskletConnection` is used to "draw" an empty edge to tasklets that do not have inputs.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -186,16 +188,42 @@ def _add_tasklet(
...
+        code: str,
+        outputs: Union[Set[str], Dict[str, dace.dtypes.typeclass]],
+        **kwargs: Any,
+    ) -> tuple[dace.nodes.Tasklet, dace.nodes.MapEntry, dace.nodes.MapExit]:
+        """Helper method to add a mapped tasklet with unique name in current state."""
+        return self.subgraph_builder.add_mapped_tasklet(
+            name, self.state, map_ranges, inputs, code, outputs, **kwargs
+        )
 
     def _get_tasklet_result(
```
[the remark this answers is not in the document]
edopao: Agreed.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -428,9 +455,89 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+        assert isinstance(node.type, itir_ts.ListType)
+        assert isinstance(node.type.element_type, ts.ScalarType)
+        dtype = dace_utils.as_dace_type(node.type.element_type)
+
+        input_args = []
+        for arg in node.args:
+            input_args.append(self.visit(arg))
+
+        # TODO(edopao): extract offset_dim from the input arguments
+        offset_dim = gtx_common.Dimension("", gtx_common.DimensionKind.LOCAL)
```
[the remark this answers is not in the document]
edopao: There is a TODO comment, more work will be done on the local dimension.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -428,9 +455,89 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+        map_index = dace_gtir_utils.get_map_variable(offset_dim)
+        connectors = [f"__arg{i}" for i in range(len(input_args))]
+
+        input_memlets = {}
+        input_nodes = {}
+        local_size: Optional[int] = None
+        for conn, input_expr in zip(connectors, input_args, strict=True):
```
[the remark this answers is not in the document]
edopao: It loops through all map arguments and creates view nodes to slice the local dimension. For scalar argument, it just create memlets with subset 0.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -428,9 +455,89 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+            if isinstance(input_expr, MemletExpr):
+                if set(input_expr.subset.size()[:-1]) != {1}:
+                    raise ValueError(f"Invalid node {node}")
+                rstart, rstop, rstep = input_expr.subset[-1]
+                assert rstart == 0 and rstep == 1
+
+                desc = self.sdfg.arrays[input_expr.node.data]
+                view, _ = self.sdfg.add_view(
```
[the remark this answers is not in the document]
edopao: I am using a view because I want to expose a 1D array, along the local dimension.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -428,9 +455,89 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+            assert len(input_node.desc(self.sdfg).shape) == 1
+            input_size = input_node.desc(self.sdfg).shape[0]
+            if input_size == 1:
+                input_memlets[conn] = dace.Memlet(data=input_node.data, subset="0")
+            else:
+                input_memlets[conn] = dace.Memlet(data=input_node.data, subset=map_index)
+                if local_size and input_size != local_size:
+                    raise ValueError(f"Invalid node {node}")
+                else:
+                    local_size = input_size
```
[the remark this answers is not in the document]
edopao: No, it should be set only one time. If `input_size != local_size` on line 506, an exception will be thrown.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -428,9 +455,89 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+
+            input_nodes[input_node.data] = input_node
+
+        if local_size is None:
+            # corner case where map is applied to 1-element lists
+            assert len(input_nodes) >= 1
+            local_size = 1
+
+        temp, _ = self.sdfg.add_temp_transient((local_size,), dtype)
+        temp_node = self.state.add_access(temp)
```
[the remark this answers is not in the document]
edopao: Agreed.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -428,9 +455,89 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+
+        assert isinstance(node.fun, gtir.FunCall)
+        assert len(node.fun.args) == 1
+        fun_node = im.call(node.fun.args[0])(*connectors)
+        op_code = gtir_python_codegen.get_source(fun_node)
```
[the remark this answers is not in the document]
edopao: It is more complex than so.. those translators can be chained, actually. The important thing is that `gtir_python_codegen.get_source` just returns some python code in string format.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
edopao: Agreed.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
edopao: Agreed.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
edopao: Yes, correct. The same program written in two IR flavors. I used this pattern in all tests. I could open a separate PR to use a pytest fixture to templify the test case (and apply this fixture to all tests).

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -428,9 +455,89 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+        map_index = dace_gtir_utils.get_map_variable(offset_dim)
+        connectors = [f"__arg{i}" for i in range(len(input_args))]
+
+        input_memlets = {}
+        input_nodes = {}
+        local_size: Optional[int] = None
+        for conn, input_expr in zip(connectors, input_args, strict=True):
+            if isinstance(input_expr, MemletExpr):
+                if set(input_expr.subset.size()[:-1]) != {1}:
+                    raise ValueError(f"Invalid node {node}")
```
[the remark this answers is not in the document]
edopao: The condition being checked here is different. All dimensions should be indexed with single point access except the last dimension (in current IR the local dimension is always the last one). In other words, it should be possible to slice along the local dimension, which should be always the last dimension.

edopao:
Thank you @philip-paul-mueller for the review comments. I have pushed a commit that should clarify how `InputConnection` was used. I will address the remaining comments tomorrow.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -223,11 +230,13 @@ def translate_as_field_op(
     map_ranges = {dace_gtir_utils.get_map_variable(dim): f"{lb}:{ub}" for dim, lb, ub in domain}
     me, mx = sdfg_builder.add_map("field_op", state, map_ranges)
 
-    if len(input_connections) == 0:
-        # dace requires an empty edge from map entry node to tasklet node, in case there no input memlets
-        state.add_nedge(me, last_node, dace.Memlet())
-    else:
-        for data_node, data_subset, lambda_node, lambda_connector in input_connections:
+    # here we setup the edges from the map entry node
+    for x in input_connections:
```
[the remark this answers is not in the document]
edopao: Addressed by refactoring-PR #1685

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_sdfg.py
```diff
@@ -79,6 +79,21 @@ def add_tasklet(
...
+        map_ranges: Dict[str, str | dace.subsets.Subset]
+        | List[Tuple[str, str | dace.subsets.Subset]],
+        inputs: Union[Set[str], Dict[str, dace.dtypes.typeclass]],
+        code: str,
+        outputs: Union[Set[str], Dict[str, dace.dtypes.typeclass]],
+        **kwargs: Any,
+    ) -> tuple[dace.nodes.Tasklet, dace.nodes.MapEntry, dace.nodes.MapExit]:
+        """Wrapper of `dace.SDFGState.add_tasklet` that assigns unique name."""
+        unique_name = self.unique_tasklet_name(name)
+        return state.add_mapped_tasklet(unique_name, map_ranges, inputs, code, outputs, **kwargs)
```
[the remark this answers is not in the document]
edopao: I got a second thought on this. Even if I pass the arguments by name, the name would be the same as the variable.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -223,11 +230,13 @@ def translate_as_field_op(
...
+        if isinstance(x, dace.nodes.Tasklet):
+            # tasklet without arguments: simply add an edge from map entry node to the tasklet node
+            state.add_nedge(me, x, dace.Memlet())
```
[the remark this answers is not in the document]
edopao: Addressed by refactoring-PR #1685

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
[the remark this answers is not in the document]
edopao: Addressed by refactoring-PR #1685

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -186,16 +188,42 @@ def _add_tasklet(
         **kwargs: Any,
     ) -> dace.nodes.Tasklet:
         """Helper method to add a tasklet with unique name in current state."""
-        return self.subgraph_builder.add_tasklet(name, self.state, inputs, outputs, code, **kwargs)
+        tasklet_node = self.subgraph_builder.add_tasklet(
+            name, self.state, inputs, outputs, code, **kwargs
+        )
+        if len(inputs) == 0:
+            # tasklet nodes without arguments need an empty edge from map entry node
+            self.input_connections.append(tasklet_node)
```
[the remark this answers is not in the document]
edopao: Addressed by refactoring-PR #1685

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
```diff
@@ -428,9 +455,89 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+        self._add_mapped_tasklet(
+            name="map",
+            map_ranges={map_index: f"0:{local_size}"},
+            code=f"__out = {op_code}",
+            inputs=input_memlets,
+            input_nodes=input_nodes,
+            outputs={
+                "__out": dace.Memlet(data=temp, subset=map_index),
+            },
+            output_nodes={temp: temp_node},
```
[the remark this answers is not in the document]
edopao: Thanks, we will have to wait until next dace release.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_to_tasklet.py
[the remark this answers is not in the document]
edopao: I hope it is more clear now, after the refactoring-PR #1685. The dataflow we build in this class has some loose connections on input edges. These edges are described as set of nodes, that will have to be connected to external data source nodes passing through the map entry node. For `map_` and `neighbors` these edges will terminate on the view nodes, because it was much more complicated to terminate them directly on the map entry node.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -97,15 +120,16 @@ def _parse_fieldop_arg(
     if not isinstance(arg, Field):
         raise ValueError(f"Received {node} as argument to field operator, expected a field.")
 
+    def make_iterator_index(var: str) -> gtir_dataflow.SymbolExpr:
+        dtype = dace.dtype_to_typeclass(gtx_fbuiltins.IndexType)
+        return gtir_dataflow.SymbolExpr(var, dtype)
+
```
philip-paul-mueller, line 126: This is a private function, it can not be used everywhere else, however, you are only using it once in this function.
This looks a bit strange.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -188,18 +234,44 @@ def translate_as_field_op(
     assert isinstance(domain_expr, gtir.FunCall)
 
     # parse the domain of the field operator
-    domain = dace_gtir_utils.get_domain(domain_expr)
+    domain = translate_domain(domain_expr)
 
+    # The reduce identity value is used to fill the skip values in neighbors list
```
philip-paul-mueller, line 239: I thought the PR is only about full, aka. without skip values?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -188,18 +234,44 @@ def translate_as_field_op(
...
+    #
+    # A reduction expression can be either expressed in local view (itir):
+    # vertices @ u⟨ Vertexₕ: [0, nvertices) ⟩
```
philip-paul-mueller, line 242: What does the index `h` in `Vertexₕ` mean?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -188,18 +234,44 @@ def translate_as_field_op(
...
+    #        );
+    #
+    # In local view, the list of neighbors is (recursively) built while visiting
+    # the current expression.
+    # In field view, the list of neighbors is built as argument to the current
+    # expression. Therefore, the reduction identity value needs to be passed to
+    # the argument visitor (`reduce_identity_for_args = reduce_identity`).
+    # However, when the argument visitor hits the `neighbors` expression (see the
+    # second if-branch), we should stop carrying the reduce identity further
+    # (`reduce_identity_for_args = None`).
```
philip-paul-mueller, line 260: Okay, but why?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -566,7 +640,7 @@ def translate_symbol_ref(
 if TYPE_CHECKING:
```
philip-paul-mueller, line 640: Why is that thing at the end of the file?
According to the coding convention it should be at the top.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -157,7 +180,30 @@ def _create_temporary_field(
...
+            for arg in named_range.args[1:3]
+        ]
+        dim = gtx_common.Dimension(axis.value, axis.kind)
+        domain.append((dim, bounds[0], bounds[1]))
+
+    return domain
+
+
+def translate_as_fieldop(
     node: gtir.Node,
```
philip-paul-mueller, line 207: `FunCall` should be part of the type signature, but I would keep the assert.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -566,7 +640,7 @@ def translate_symbol_ref(
 if TYPE_CHECKING:
     # Use type-checking to assert that all translator functions implement the `PrimitiveTranslator` protocol
     __primitive_translators: list[PrimitiveTranslator] = [
```
philip-paul-mueller, line 642: Two underscore have a special meaning in Python, at least inside classes.
Why do you use them.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -295,14 +295,37 @@ def _add_tasklet(
...
         self,
         dtype: dace.typeclass,
         src_node: dace.nodes.Tasklet,
         src_connector: str,
+        use_array: bool = False,
     ) -> DataExpr:
         temp_name = self.sdfg.temp_data_name()
-        self.sdfg.add_scalar(temp_name, dtype, transient=True)
+        if use_array:
+            # Special case for local exprresions with list type annotation, that
```
philip-paul-mueller, line 322: I would start as "In some cases, such as list type annotation, we want ..." It might sound better.
The current version sounds more like the justification of a very gross hack.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -535,15 +557,99 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+        input_args = [self.visit(arg) for arg in node.args]
+
+        # TODO(edopao): extract offset_dim from the input arguments
+        offset_dim = gtx_common.Dimension("", gtx_common.DimensionKind.LOCAL)
+        map_index = dace_gtir_utils.get_map_variable(offset_dim)
+        connectors = [f"__arg{i}" for i in range(len(input_args))]
+
+        # The dataflow we build in this class has some loose connections on input edges.
+        # These edges are described as set of nodes, that will have to be connected to
+        # external data source nodes passing through the map entry node of the field map.
```
philip-paul-mueller, line 576: I still do not think that views are the best solution here.
Take a look at `SDFGState.add_memlet_path()` that can creates paths through multiple maps.

But I still does not see how the view helps, as you have to connect that thing still with the outside.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -535,15 +557,99 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+            if isinstance(input_expr, MemletExpr):
+                if set(input_expr.subset.size()[:-1]) != {1}:
+                    raise ValueError(f"Invalid node {node}")
+                rstart, rstop, rstep = input_expr.subset[-1]
+                assert rstart == 0 and rstep == 1
+
+                desc = self.sdfg.arrays[input_expr.node.data]
+                view, _ = self.sdfg.add_view(
+                    f"{input_expr.node.data}_view",
+                    (rstop + 1,),
```
philip-paul-mueller, line 593: why this shape and not `rstop - rstart`?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -535,15 +557,99 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
             memlet=dace.Memlet(data=neighbors_temp, subset=neighbor_idx),
         )
 
-        assert isinstance(node.type, itir_ts.ListType)
         return DataExpr(neighbors_node, node.type)
 
+    def _visit_map(self, node: gtir.FunCall) -> DataExpr:
```
philip-paul-mueller, line 562: I do not really understand what this function does.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -535,15 +557,99 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+
+        return DataExpr(out_node, dtype)
+
     def _visit_reduce(self, node: gtir.FunCall) -> DataExpr:
         op_name, reduce_init, reduce_identity = get_reduce_params(node)
         dtype = reduce_identity.dtype
 
         # We store the value of reduce identity in the visitor context while visiting
-        # the input to reduction; this value will be use by the `neighbors` visitor
+        # the input to reduction; this value will be used by the `neighbors` visitor
```
philip-paul-mueller, line 652: Shouldn't it be "might be used by"?
Also where does the `neighbors` built in comes from?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -863,7 +966,27 @@ def _visit_generic_builtin(self, node: gtir.FunCall) -> DataExpr:
...
 
-        return self._construct_tasklet_result(dtype, tasklet_node, "result")
+        if isinstance(node.type, itir_ts.ListType):
+            # The only builtin function (so far) handled here that returns a list
+            # is 'make_const_list'. There are other builtin functions (map_, neighbors)
+            # that return a list but they are handled in specialized visit methods.
+            # This method (the generic visitor for builtin functions) always returns
+            # scalars. This is also the case of 'make_const_list' expression: it simply
+            # broadcasts a scalar value on the local domain of another expression,
+            # for example 'map_(plus)(neighbors(V2Eₒ, it), make_const_list(1.0))'.
```
philip-paul-mueller, line 976: Just for my understanding, what does this expression do?
Does it add 1.0 to every first edge enighbor of the iterator `it`?
Furthermore, since this is the fieldview representation, why are you still using an iterator?

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
philip-paul-mueller: No you can keep it like that.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -157,7 +180,30 @@ def _create_temporary_field(
     return Field(field_node, field_type)
 
 
-def translate_as_field_op(
+def translate_domain(node: gtir.Node) -> FieldopDomain:
```
philip-paul-mueller, line 183: This is complete optional, but I do not like `translate` how about `process_domain()` or `extract_domain()`?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -295,14 +295,37 @@ def _add_tasklet(
...
+        self,
+        name: str,
+        map_ranges: Dict[str, str | dace.subsets.Subset]
+        | List[Tuple[str, str | dace.subsets.Subset]],
+        inputs: Union[Set[str], Dict[str, dace.dtypes.typeclass]],
+        code: str,
+        outputs: Union[Set[str], Dict[str, dace.dtypes.typeclass]],
+        **kwargs: Any,
+    ) -> tuple[dace.nodes.Tasklet, dace.nodes.MapEntry, dace.nodes.MapExit]:
+        """Helper method to add a mapped tasklet with unique name in current state."""
```
philip-paul-mueller, line 308: The doc string is a bit misleading as you still have to provide a unique name.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -295,14 +295,37 @@ def _add_tasklet(
...
-        self.sdfg.add_scalar(temp_name, dtype, transient=True)
+        if use_array:
+            # Special case for local exprresions with list type annotation, that
+            # are lowered to a tasklet with output scalar value: we still want to
+            # represent the output data as an array (single-element 1D array) in
+            # order to allow composition of array shape in external memlets.
+            self.sdfg.add_array(temp_name, (1,), dtype, transient=True)
+        else:
+            self.sdfg.add_scalar(temp_name, dtype, transient=True)
         data_type = dace_utils.as_scalar_type(str(dtype.as_numpy_dtype()))
```
philip-paul-mueller, line 329: Why do you have to convert it here into a string?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -535,15 +557,99 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+        if local_size is None:
+            # corner case where map is applied to 1-element lists
+            assert len(input_nodes) >= 1
+            local_size = 1
+
+        out, _ = self.sdfg.add_temp_transient((local_size,), dtype)
+        out_node = self.state.add_access(out)
+
+        assert isinstance(node.fun, gtir.FunCall)
+        assert len(node.fun.args) == 1
```
philip-paul-mueller, line 628: These checks should be at the top.

philip-paul-mueller (commented):
There is nothing serious, but some aspects need a bit more work.
But I expect that we can merge in the next round.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: The reason why we stop carrying the reduce identity further is that we do not need it, since reduce operates along a single axis. I have updated the comment.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -863,7 +998,27 @@ def _visit_generic_builtin(self, node: gtir.FunCall) -> DataExpr:
                     connector,
                 )
 
-        return self._construct_tasklet_result(dtype, tasklet_node, "result")
+        if isinstance(node.type, itir_ts.ListType):
+            # The only builtin function (so far) handled here that returns a list
+            # is 'make_const_list'. There are other builtin functions (map_, neighbors)
+            # that return a list but they are handled in specialized visit methods.
+            # This method (the generic visitor for builtin functions) always returns
+            # scalars. This is also the case of 'make_const_list' expression: it simply
```
philip-paul-mueller, line 1006: I think it is not super precise here.
The function does not _always_ return a scalar, in case of the `make_const_list()` the function returns an array of shape `(1,)` which is close to an array but not the same (even in DaCe you can use a `(1,)` arry only in ~90% of the cases were an array is needed (but this is probably a shortcoming in the implementation).
So you should make that a bit more precise.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -110,7 +110,7 @@ def _parse_fieldop_arg(
         dims = arg.data_type.dims + (
             # we add an extra anonymous dimension in the iterator definition to enable
             # dereferencing elements in `ListType`
-            [gtx_common.Dimension("")] if isinstance(arg.data_type.dtype, gtir_ts.ListType) else []
+            [gtx_common.Dimension("")] if isinstance(arg.data_type.dtype, itir_ts.ListType) else []
```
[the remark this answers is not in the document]
philip-paul-mueller: Okay it is a bit off topic, but [Rule 11](https://github.com/GridTools/gt4py/blob/main/docs/development/ADRs/0018-Canonical_SDFG_in_GT4Py_Transformations.md#maps) implies that the name of a dimensions and the iteration variable name is liked.

I think it is not a problem, but we should probably clarify the rule a bit.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -181,15 +181,22 @@ def translate_as_field_op(
     domain = dace_gtir_utils.get_domain(domain_expr)
 
     if cpm.is_applied_reduce(stencil_expr.expr):
+        # the reduce identity value is used to fill the skip values in neighbors list
         if reduce_identity is not None:
             raise NotImplementedError("nested reductions not supported.")
-
-        # the reduce identity value is used to fill the skip values in neighbors list
         _, _, reduce_identity = gtir_to_tasklet.get_reduce_params(stencil_expr.expr)
+        reduce_identity_for_args = reduce_identity
```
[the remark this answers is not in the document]
philip-paul-mueller: Okay, I would make this aspect more clearer.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_python_codegen.py
```diff
@@ -83,9 +83,14 @@ def builtin_if(*args: Any) -> str:
     return f"{true_val} if {cond} else {false_val}"
 
 
+def make_const_list(arg: str) -> str:
+    return arg
```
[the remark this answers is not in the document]
philip-paul-mueller: So you can say it is like a "lazy fill"?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -535,16 +571,118 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+        input_nodes = {}
+        local_size: Optional[int] = None
+        for conn, input_expr in zip(connectors, input_args, strict=True):
+            if isinstance(input_expr, MemletExpr):
+                if set(input_expr.subset.size()[:-1]) != {1}:
+                    raise ValueError(f"Invalid node {node}")
+                rstart, rstop, rstep = input_expr.subset[-1]
+                assert rstart == 0 and rstep == 1
+
+                desc = self.sdfg.arrays[input_expr.node.data]
```
philip-paul-mueller, line 623: I would add a note to inform the reader that in the current IR the local dimension is always at the end.

Please also add the TODO:
```
TODO(phimuell): Ensure that the local memory layout is correct/corrected during optimization
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
philip-paul-mueller: According to the [google style guide](https://google.github.io/styleguide/pyguide.html#31913-conditional-imports) such imports should be at the top.
But I am not sure if this is really needed at all.

philip-paul-mueller (commented):
There are some things that needs some discussion but they are rather minor points.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -11,56 +11,13 @@
...
-) -> dict[gtx_common.Dimension, tuple[dace.symbolic.SymbolicType, dace.symbolic.SymbolicType]]:
-    """
-    Returns domain represented in dictionary form.
-    """
-    domain = get_domain(node)
-
-    return {dim: (lb, ub) for dim, lb, ub in domain}
-
-
 def get_map_variable(dim: gtx_common.Dimension) -> str:
```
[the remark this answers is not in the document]
edopao: Added a TODO: in next PR I will raise an exception if dim string is not set.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -535,16 +571,118 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+        # external data source nodes passing through the map entry node of the field map.
+        # For `map_` and `neighbors` expressions, these edges will terminate on the view
+        # nodes (see the for-loop below), because it is simpler than representing
+        # map-to-map edges (which require memlets with 2 pass-nodes).
+        input_memlets = {}
+        input_nodes = {}
+        local_size: Optional[int] = None
+        for conn, input_expr in zip(connectors, input_args, strict=True):
+            if isinstance(input_expr, MemletExpr):
+                if set(input_expr.subset.size()[:-1]) != {1}:
```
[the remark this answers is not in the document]
edopao: Actually, it does not work. On the other hand, I think this condition is not strictly needed so I'll removed it.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
philip-paul-mueller: But a conditional declaration does not make sense to me.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
philip-paul-mueller: Then we keep it that way.
I do not see the point of it.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
philip-paul-mueller: I would add a note about that, because it is very confusing actually.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -101,11 +127,8 @@ def _parse_fieldop_arg(
         return gtir_dataflow.MemletExpr(arg.data_node, sbs.Indices([0]))
     elif isinstance(arg.data_type, ts.FieldType):
         indices: dict[gtx_common.Dimension, gtir_dataflow.ValueExpr] = {
-            dim: gtir_dataflow.SymbolExpr(
-                dace_gtir_utils.get_map_variable(dim),
-                IteratorIndexDType,
-            )
-            for dim, _, _ in domain
+            dim: gtir_dataflow.SymbolExpr(map_variable, INDEX_DTYPE)
+            for dim, map_variable, _ in _get_fieldop_ndrange(domain)
```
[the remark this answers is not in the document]
philip-paul-mueller: Sorry this was my bad, I was a bit confused.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -208,7 +281,8 @@ def translate_as_field_op(
     input_edges, output = taskgen.visit(stencil_expr, args=stencil_args)
     output_desc = output.expr.node.desc(sdfg)
 
-    domain_index = sbs.Indices([dace_gtir_utils.get_map_variable(dim) for dim, _, _ in domain])
+    fieldop_ndrange = _get_fieldop_ndrange(domain)
+    domain_index = sbs.Indices([var for _, var, _ in fieldop_ndrange])
```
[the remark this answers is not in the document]
philip-paul-mueller: Sorry, I was confused.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -412,6 +449,8 @@ def _visit_deref(self, node: gtir.FunCall) -> ValueExpr:
 
     def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
         assert len(node.args) == 2
+        assert isinstance(node.type, itir_ts.ListType)
+        assert self.reduce_identity is not None
```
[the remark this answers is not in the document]
philip-paul-mueller: My question was not that well phrased.
In `translate_as_fieldop()` you wrote "[...] when the argument visitor hits an expression returning a neighbors list (see the second if-branch), we stop carrying the reduce identity further [...]".
So my question was aimed at, why are you requiring that here the reduce expression is not `None`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -535,16 +571,118 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+        In the example below, the local dimension consists of a list of neighbor
+        values as the first argument, and a list of constant values `1.0`:
+        `map_(plus)(neighbors(V2Eₒ, it), make_const_list(1.0))`
+
+        The `plus` operation is lowered to a tasklet inside a map that computes
+        the domain of the local dimension (in this example, the number of neighbors).
+        """
+        assert isinstance(node.type, itir_ts.ListType)
+        assert isinstance(node.fun, gtir.FunCall)
+        assert len(node.fun.args) == 1  # the operation to be mapped on the arguments
```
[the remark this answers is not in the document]
philip-paul-mueller: Thanks for the explanation.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -535,16 +571,118 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+        assert isinstance(node.type.element_type, ts.ScalarType)
+        dtype = dace_utils.as_dace_type(node.type.element_type)
+
+        input_args = [self.visit(arg) for arg in node.args]
+        connectors = [f"__arg{i}" for i in range(len(input_args))]
+
+        # Convert the operation to be mapped into a function call on a list of
+        # internal parameters with the same names as the tasklet connectors.
+        fun_node = im.call(node.fun.args[0])(*connectors)
+        op_code = gtir_python_codegen.get_source(fun_node)
```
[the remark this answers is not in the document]
philip-paul-mueller: But then you should move this lines down or give them better names and mention what you are doing with them.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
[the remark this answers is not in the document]
philip-paul-mueller: This is strange.
Did it just give wrong results or did it produced a sytax error or so?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -188,18 +233,46 @@ def translate_as_field_op(
...
+    # In local view, the list of neighbors is (recursively) built while visiting
+    # the current expression.
+    # In field view, the list of neighbors is built as argument to the current
+    # expression. Therefore, the reduction identity value needs to be passed to
+    # the argument visitor (`reduce_identity_for_args = reduce_identity`).
+    # However, when the argument visitor hits an expression returning a neighbors
+    # list (see the second if-branch), we stop carrying the reduce identity further
+    # (`reduce_identity_for_args = None`) because it is not needed aymore;
+    # the reason being that reduce operates along a single axis, which in this case
+    # corresponds to the local dimension of the list of neighbor values.
```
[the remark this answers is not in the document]
philip-paul-mueller: I think the whole problem is that there are different reduce identities:
- `self.reduce_identity`
- `reduce_identity_for_args`

I do not get why they are split.
Also I do not fully understand the explanation.
I think it is that if you hit the neighbor built in, then you would reduce over the list of neighbor hood values of a point, and because you are _currently_ ignoring skip values, i.e. assume that all are defined, you do not have to fuill something.
So technically speaking it is not _needed_ but not wrong to do.
If there is no reason to threat is as special case then I would not treat it as special case.
Except you want to do some prep work for the skip value case, but then you have to clearly indicate this.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -157,7 +168,30 @@ def _create_temporary_field(
     return Field(field_node, field_type)
 
 
-def translate_as_field_op(
+def extract_domain(node: gtir.Node) -> FieldopDomain:
+    """
+    Visits the domain of a field operator and returns a list of dimensions and
+    the corresponding lower and upper bounds.
```
philip-paul-mueller, line 174: Since depending which range constructor in DaCe is used the upper bound is either inclusive (`tuple` based constructor) or exclusive (string based constructor) you should add which format the output follows.
I would also settle for one format (but I think this is already the case) and only deviate from it if necessary and make this clear.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: The `TYPE_CHECKING` flag is only `True` while we run mypy. In regular python execution, `TYPE_CHECKING` is `False` and the variable `__primitive_translators` is not declared.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -145,7 +145,7 @@ class DataflowOutputEdge:
     """
 
     state: dace.SDFGState
-    expr: DataExpr
```
philip-paul-mueller, line 148: `result` is probably a better name as `expr`, but I suggest you to make it more consistent with the other two classes especially `MemletInputEdge` and call it `destination`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: The point is to ensure that all `translate_` functions have the same signature.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -487,64 +523,156 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
         neighbor_idx = dace_gtir_utils.get_map_variable(offset_dim)
-        me, mx = self._add_map(
-            f"{offset}_neighbors",
-            {
-                neighbor_idx: f"0:{offset_provider.max_neighbors}",
-            },
-        )
+
         index_connector = "__index"
+        op_code = f"__field[{index_connector}]"
```
philip-paul-mueller, line 528: I do not like `op_code` but it is consistent with what is used elsewhere.
I would suggest you to call it `tasklet_expression` ([justification](https://en.cppreference.com/w/cpp/language/expressions)).
My main point is that "tasklet" should be included in the name.

This applies to all other occurrences of `op_code`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I have changed the implementation, I am not using anymore `rstart` and `rstop`. Now I am using:
`view_shape = tuple(desc.shape[i] for i in local_dim_indices)`

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -487,64 +523,156 @@ def _visit_neighbors(self, node: gtir.FunCall) -> DataExpr:
...
+
+        The map operation is applied on the local dimension of input fields.
+        In the example below, the local dimension consists of a list of neighbor
+        values as the first argument, and a list of constant values `1.0`:
+        `map_(plus)(neighbors(V2Eₒ, it), make_const_list(1.0))`
 
+        The `plus` operation is lowered to a tasklet inside a map that computes
+        the domain of the local dimension (in this example, the number of neighbors).
+
+        The result is a 1D local field, with same size as the input local dimension.
```
philip-paul-mueller, line 570: I would also mention what the value is.
Now that I think about it, what is the size of the local field anyway.
I would say it has size 1, because $V2E_{0}$ is only one element?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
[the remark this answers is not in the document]
edopao: It has a different meaning. The original check was that the range length on all dimensions except for last one was equal to one. The check you proposed was the opposite, only checking the last dimension.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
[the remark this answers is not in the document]
edopao: Because here we are visiting the neighbor expression itself. This is not the argument visitor (by that I mean we are not visiting the input arguments to the neighbor expression).

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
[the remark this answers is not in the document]
edopao: I will declare a variable called `output_connector` and use it instead of `__out` string.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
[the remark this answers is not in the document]
philip-paul-mueller: Now that you say it makes sense, but then I would suggest:
```python
if not all(dim_size == 1 for dim_size in input_expr.subset.size()[:-1]):
```

philip-paul-mueller (commented):
There are some smallish things.
But It generally looks good.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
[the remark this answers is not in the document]
edopao: `self.reduce_identity` is defined in `LambdaToDataflow`, not here. Here `reduce_identity` is an input to the function. By visiting the node arguments with `reduce_identity_for_args = None`, their `reduce_identity` will be `None`.

Setting `reduce_identity_for_args = None` will also work in case of skip values. The `reduce_identity` value is filled in in place of skip values in the context of `neighbors` itself, not in the arguments context.

You are right that it would not be wrong to set `reduce_identity_for_args` also in case of neighbors arguments. I do it because it enables a sanity check that the sequence `reduce(V2E) -> neighbors(V2E) -> reduce(C2E) -> neighbors(C2E)` is accepted, while the sequence `reduce(V2E) -> reduce(C2E) -> neighbors(V2E) -> neighbors(C2E)` is not. The latter sequence would raise the `NotImplementedError("nested reductions not supported.")` exception.

Side note. There is a simple alternative design where I can avoid passing `reduce_identity` around. It would be to write a dummy value for skip values instead of doing array access (for `neighbors`) or executing the map operation (for `map` - I cannot safely apply the map operation in case for example of division by 0). Then, in the lowering of reduce I would put -- before the reduce library node -- another local map that fills the skip values with the reduce identity. I really like this approach (separate PR) but it will result in one extra write - replacing the dummy value with the reduce identity.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: I have added the description of the returned bounds. Yes, the lowering follows the format with exclusive upper bound.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: The subscript is the letter 'O' and it stands for offset. I will remove it from the example in order to avoid confusion. In this example, the local size will be `V2E.max_neighbors`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: Good idea, agreed.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: It is not really a destination, it is rather the sink node of the dataflow. What about `sink`?

philip-paul-mueller: `sink` has a meaning in DaCe as "node with zero output degree", so it is a bit misleading, how about target?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
philip-paul-mueller: Again I would add what value the result has.

philip-paul-mueller (approved):
I have some strong suggestions to improve the documentation of the implementation and behaviour of internals.
But the PR kind of looks good, but some documenting aspects should be improved.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: In this case, `expr` is a sink node according to the dace definition. A data node the dataflow writes to.

philip-paul-mueller: but will it never be read from?

edopao: The sink node (typically a scalar or a local field) of the dataflow will actually be removed from the SDFG. The edge writing to the sink node will be replaced with an edge passing through the map exit node and writing to a new node, the transient field node (a multi-dimensional array with shape equal to the field operator domain).

edopao:
cscs-ci run default

edopao:
cscs-ci run default

PR #1685, opened 2024-10-10
feat[next][dace]: refactoring

edopao:
Implements some refactoring changes discussed during review of #1683.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_common/workflow.py
+++ b/src/gt4py/next/program_processors/runners/dace_common/workflow.py
@@ -43,6 +43,10 @@ class CompiledDaceProgram(stages.CompiledProgram):
         result = self.sdfg_program(*args, **kwargs)
         assert result is None
 
+    def fast_call(self) -> None:
+        result = self.sdfg_program.fast_call(*self.sdfg_program._lastargs)
+        assert result is None
+
 
 @dataclasses.dataclass(frozen=True)
 class DaCeCompiler(
@@ -142,7 +146,7 @@ def convert_args(
                         ), f"argument '{arg_name}' not found."
 
             if use_fast_call:
-                return sdfg_program.fast_call(*sdfg_program._lastargs)
+                return inp.fast_call()
 
         sdfg_args = dace_backend.get_sdfg_args(
             sdfg,
--- a/src/gt4py/next/program_processors/runners/dace_common/workflow.py
+++ b/src/gt4py/next/program_processors/runners/dace_common/workflow.py
@@ -40,7 +40,8 @@ class CompiledDaceProgram(stages.CompiledProgram):
         ]
 
     def __call__(self, *args: Any, **kwargs: Any) -> None:
-        self.sdfg_program(*args, **kwargs)
+        result = self.sdfg_program(*args, **kwargs)
+        assert result is None
 
 
 @dataclasses.dataclass(frozen=True)
@@ -104,7 +105,7 @@ def convert_args(
         *args: Any,
         offset_provider: common.OffsetProvider,
         out: Any = None,
-    ) -> Any:
+    ) -> None:
         if out is not None:
             args = (*args, out)
         if len(sdfg.arg_names) > len(args):
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
@@ -18,18 +18,18 @@ import dace.subsets as sbs
 from gt4py.next import common as gtx_common, utils as gtx_utils
 from gt4py.next.iterator import ir as gtir
 from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm
-from gt4py.next.iterator.type_system import type_specifications as gtir_ts
+from gt4py.next.iterator.type_system import type_specifications as itir_ts
 from gt4py.next.program_processors.runners.dace_common import utility as dace_utils
 from gt4py.next.program_processors.runners.dace_fieldview import (
+    gtir_dataflow,
     gtir_python_codegen,
-    gtir_to_tasklet,
     utility as dace_gtir_utils,
 )
 from gt4py.next.type_system import type_specifications as ts
 
 
 if TYPE_CHECKING:
-    from gt4py.next.program_processors.runners.dace_fieldview import gtir_to_sdfg
+    from gt4py.next.program_processors.runners.dace_fieldview import gtir_sdfg
 
 
 IteratorIndexDType: TypeAlias = dace.int32  # type of iterator indexes
@@ -51,8 +51,8 @@ class PrimitiveTranslator(Protocol):
         node: gtir.Node,
         sdfg: dace.SDFG,
         state: dace.SDFGState,
-        sdfg_builder: gtir_to_sdfg.SDFGBuilder,
-        reduce_identity: Optional[gtir_to_tasklet.SymbolExpr],
+        sdfg_builder: gtir_sdfg.SDFGBuilder,
+        reduce_identity: Optional[gtir_dataflow.SymbolExpr],
     ) -> FieldopResult:
         """Creates the dataflow subgraph representing a GTIR primitive function.
 
@@ -80,12 +80,12 @@ def _parse_fieldop_arg(
     node: gtir.Expr,
     sdfg: dace.SDFG,
     state: dace.SDFGState,
-    sdfg_builder: gtir_to_sdfg.SDFGBuilder,
+    sdfg_builder: gtir_sdfg.SDFGBuilder,
     domain: list[
         tuple[gtx_common.Dimension, dace.symbolic.SymbolicType, dace.symbolic.SymbolicType]
     ],
-    reduce_identity: Optional[gtir_to_tasklet.SymbolExpr],
-) -> gtir_to_tasklet.IteratorExpr | gtir_to_tasklet.MemletExpr:
+    reduce_identity: Optional[gtir_dataflow.SymbolExpr],
+) -> gtir_dataflow.IteratorExpr | gtir_dataflow.MemletExpr:
     arg = sdfg_builder.visit(
         node,
         sdfg=sdfg,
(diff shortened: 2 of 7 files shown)


src/gt4py/next/program_processors/runners/dace_common/workflow.py
```diff
@@ -40,7 +40,8 @@ def __init__(self, program: dace.CompiledSDFG):
         ]
 
     def __call__(self, *args: Any, **kwargs: Any) -> None:
```
philip-paul-mueller, line 42:
```suggestion
    def __call__(self, *args: Any, **kwargs: Any) -> bool:
```

src/gt4py/next/program_processors/runners/dace_common/workflow.py
```diff
@@ -104,7 +105,7 @@ def decorated_program(
         *args: Any,
         offset_provider: common.OffsetProvider,
         out: Any = None,
-    ) -> Any:
```
philip-paul-mueller, line 107: See comment above, must be bool.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -161,8 +161,8 @@ def translate_as_field_op(
     node: gtir.Node,
     sdfg: dace.SDFG,
     state: dace.SDFGState,
-    sdfg_builder: gtir_to_sdfg.SDFGBuilder,
-    reduce_identity: Optional[gtir_to_tasklet.SymbolExpr],
+    sdfg_builder: gtir_sdfg.SDFGBuilder,
+    reduce_identity: Optional[gtir_dataflow.SymbolExpr],
 ) -> FieldopResult:
     """Generates the dataflow subgraph for the `as_fieldop` builtin function."""
```
philip-paul-mueller, line 167: You describethe format of the argument below in a comment.
I would put them into the doc string, because this is the central place you start looking for.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -83,6 +76,83 @@ class IteratorExpr:
     indices: dict[gtx_common.Dimension, ValueExpr]
 
 
+class DataflowInputEdge(Protocol):
+    """Allows to setup an input edge in a map scope."""
+
+    @abc.abstractmethod
+    def connect(self, me: dace.nodes.MapEntry) -> None:
+        pass
```
philip-paul-mueller, line 84: I think the coding guide says that you have to use `...` here, but I am not sure.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -83,6 +76,83 @@ class IteratorExpr:
     indices: dict[gtx_common.Dimension, ValueExpr]
 
 
+class DataflowInputEdge(Protocol):
+    """Allows to setup an input edge in a map scope."""
```
philip-paul-mueller, line 80: Maybe the doc string of the class could be a bit more detailed.
I get the concept, like it and I think it is probably the right way to go.
But I would also elaborate how this class is intended to be used, in which situations.
And for completes you should also mention the `connect` function and its relation to the outflow edge.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -83,6 +76,83 @@ class IteratorExpr:
...
+    state: dace.SDFGState
+    node: dace.nodes.Tasklet
+
+    def connect(self, me: dace.nodes.MapEntry) -> None:
+        self.state.add_nedge(me, self.node, dace.Memlet())
+
+
+@dataclasses.dataclass(frozen=True)
+class DataflowOutputEdge:
+    """Allows to setup a data edge to write the result through a map exit node."""
```
philip-paul-mueller, line 126: Also a bit more explanation needed how it works and how it is intended to be used.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -116,7 +186,7 @@ def get_reduce_params(node: gtir.FunCall) -> tuple[str, SymbolExpr, SymbolExpr]:
     return op_name, reduce_init, reduce_identity
 
 
-class LambdaToTasklet(eve.NodeVisitor):
+class LambdaToDataflow(eve.NodeVisitor):
```
philip-paul-mueller, line 189: Just for my understanding and probably as additional comment.
Does this function builds everything inside a single map or multiple maps?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -186,9 +257,16 @@ def _add_tasklet(
         **kwargs: Any,
     ) -> dace.nodes.Tasklet:
         """Helper method to add a tasklet with unique name in current state."""
-        return self.subgraph_builder.add_tasklet(name, self.state, inputs, outputs, code, **kwargs)
+        tasklet_node = self.subgraph_builder.add_tasklet(
```
philip-paul-mueller, line 260: Why is there no dataflow input edge thingy constructed in this case?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -83,6 +76,83 @@ class IteratorExpr:
...
+class DataflowInputEdge(Protocol):
+    """Allows to setup an input edge in a map scope."""
+
+    @abc.abstractmethod
+    def connect(self, me: dace.nodes.MapEntry) -> None:
+        pass
+
+
+@dataclasses.dataclass(frozen=True)
+class DataflowMemletEdge(DataflowInputEdge):
```
philip-paul-mueller, line 88: I would add "input" to the name of the class to make it explicit that it is only for input.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -702,21 +782,8 @@ def _visit_shift(self, node: gtir.FunCall) -> IteratorExpr:
...
-
-        elif cpm.is_applied_reduce(node):
-            return self._visit_reduce(node)
-
-        elif cpm.is_applied_shift(node):
-            return self._visit_shift(node)
-
-        else:
-            assert isinstance(node.fun, gtir.SymRef)
+    def _visit_builtin(self, node: gtir.FunCall) -> DataExpr:
```
philip-paul-mueller, line 785: A reference to the switch is missing here, the name is to general.
Something along the line "This function is called if `visit_FunCall()` encounters a builtin function".
This is mostly because of its signature that begs the question "why is this function not called `visit_FunCall()`?!?".
And is the name actually correct?
I mean `dref()` is also a build in, but not handled by this function.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -764,32 +831,53 @@ def visit_FunCall(self, node: gtir.FunCall) -> IteratorExpr | ValueExpr:
...
+        elif cpm.is_applied_reduce(node):
+            return self._visit_reduce(node)
+
+        elif cpm.is_applied_shift(node):
+            return self._visit_shift(node)
+
+        elif isinstance(node.fun, gtir.SymRef):
+            return self._visit_builtin(node)
+        else:
+            raise NotImplementedError
```
philip-paul-mueller, line 852: There are two things here, you should add a bit more specific, what have you encountered.

My second point is more off topic.
For my understanding `raise NotImplementedError` throws the class _object `NotImplementedError` and not an instance of this class.
So would you not have to write `raise NotImplementedError()` to create an instance?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -194,55 +194,33 @@ def translate_as_field_op(
...
-    output_subset = [dace_gtir_utils.get_map_variable(dim) for dim, _, _ in domain]
-    if isinstance(output_desc, dace.data.Array):
-        # additional local dimension for neighbors
+    domain_index = sbs.Indices([dace_gtir_utils.get_map_variable(dim) for dim, _, _ in domain])
+    if isinstance(node.type.dtype, itir_ts.ListType):
+        assert isinstance(output_desc, dace.data.Array)
         assert set(output_desc.offset) == {0}
-        output_subset.extend(f"0:{size}" for size in output_desc.shape)
+        # additional local dimension for neighbors
+        output_subset = sbs.Range.from_indices(domain_index) + sbs.Range.from_array(output_desc)
```
philip-paul-mueller, line 206: Could you add the following todo here:
```
# TODO(phimuell): Investigate if we should swap the two.
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -194,55 +194,33 @@ def translate_as_field_op(
...
-    )
+    # allocate local temporary storage for the result field
+    result_field = _create_temporary_field(sdfg, state, domain, node.type, output_desc)
+
+    # here we setup the edges from the map entry node
+    for edge in input_edges:
+        edge.connect(me)
+
+    # and here the edge writing the result data through the map exit node
+    output.connect(mx, result_field.data_node, output_subset)
```
philip-paul-mueller, line 223: I would add an overload to the connect function such that you could write `output.connect(mx, result_field, output_subset)` i.e. that the function figures this out on its own.
It is a bit more high level.

Put this is probably the C++ programmer speaking that misses true function overloading.

philip-paul-mueller (commented):
I very much like the new Dataflow input concept.
There are some changes, but it is generally good.

src/gt4py/next/program_processors/runners/dace_common/workflow.py
philip-paul-mueller: Sorry, for some reason I read the `assert` as a `return`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
philip-paul-mueller: I looked it up, according to the first point of [this section](https://github.com/GridTools/gt4py/blob/main/CODING_GUIDELINES.md#common-questions) it needs to be `...`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -83,6 +76,103 @@ class IteratorExpr:
...
+
+    It provides the `connect` method to setup an input edge from an external data source.
+    Since the dataflow represents a stencil, we instantiate the dataflow inside a map scope
+    and connect its inputs and outputs to external data nodes by means of memlets that
+    traverse the map entry and exit nodes.
+    """
+
+    @abc.abstractmethod
+    def connect(self, me: dace.nodes.MapEntry) -> None:
+        pass
```
philip-paul-mueller, line 91:
```suggestion
        ...
```

philip-paul-mueller (approved):
There is one `pass` that is not changed into `...` I made a note there.
Otherwise, LGTM.

PR #1687, opened 2024-10-10
[next]: Fix inline lambda pass opcount preserving option

tehrengruber:
In #1531 the `itir.Node` class got a `type` attribute, that until now contributed to the hash computation of all nodes. As such two `itir.SymRef` with the same `id`, but one with a type inferred and one without (i.e. `None`) got a different hash value. Consequently the `inline_lambda` pass did not recognize them as a reference to the same symbol and erroneously inlined the expression even with `opcount_preserving=True`. This PR fixes the hash computation, such that again `node1 == node2` implies `hash(node1) == hash(node2)`.

### DIFF
--- a/src/gt4py/next/iterator/ir.py
+++ b/src/gt4py/next/iterator/ir.py
@@ -40,7 +40,8 @@ class Node(eve.Node):
         return hash(type(self)) ^ hash(
             tuple(
                 hash(tuple(v)) if isinstance(v, list) else hash(v)
-                for v in self.iter_children_values()
+                for (k, v) in self.iter_children_items()
+                if k not in ["location", "type"]
             )
         )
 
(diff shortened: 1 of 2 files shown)


src/gt4py/next/iterator/ir.py
```diff
@@ -40,7 +40,8 @@ def __hash__(self) -> int:
         return hash(type(self)) ^ hash(
             tuple(
                 hash(tuple(v)) if isinstance(v, list) else hash(v)
-                for v in self.iter_children_values()
+                for (k, v) in self.iter_children_items()
+                if k not in ["location", "type"]
             )
         )
```
egparedes, line 46: I think this can be refactored to only call `hash()` once...
```suggestion
        return hash(
            (
                type(self),
                *(
                    tuple(v) if isinstance(v, list) else v
                    for (k, v) in self.iter_children_items()
                    if k not in ["location", "type"]
                ),
            )
        )
```

PR #1693, opened 2024-10-17
feat[next][dace]: GTIR-to-SDFG broadcast tuple scalar arg + cleanup

edopao:
This PR implements the lowering of expressions like:
`as_fieldop(deref, domain)(x[0] + (x[1][0] × 2.0 + x[1][1] × 3.0))`
where the elements of tuple `x` being accessed are all scalar values, so `x[0] + (x[1][0] × 2.0 + x[1][1] × 3.0)` is a scalar expression.

Therefore, this PR allows to lower expressions where the argument to `as_field_op` is a scalar expression rather than a lambda. Note that:
- in case of a lambda, the `as_fieldop` function computes the lambda over the field domain;
- in case of scalar expression, it broadcasts the scalar value over the field domain.

Additionally, some cleanup:
- Removed call to `gt_simplify` so that dace backend returns a plain unoptimized SDFG.
- Removed `patch_gtir`, since the functionality of this IR pass is provided by similar changes delivered in #1677 and #1689.


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -219,11 +219,49 @@ def translate_as_fieldop(
     fun_node = node.fun
     assert len(fun_node.args) == 2
     stencil_expr, domain_expr = fun_node.args
-    assert isinstance(stencil_expr, gtir.Lambda)
-    assert isinstance(domain_expr, gtir.FunCall)
 
     # parse the domain of the field operator
     domain = extract_domain(domain_expr)
+    domain_index = sbs.Indices([dace_gtir_utils.get_map_variable(dim) for dim, _, _ in domain])
```
philip-paul-mueller, line 225: There are (potentially) multiple indices here, so I would name the variable `domain_indices`.
This would also allow you to write loops such as
```python
for domain_index in domain_indices:
        ...
```

philip-paul-mueller (commented):
There is a small thing I would suggest to you, but it looks good.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -219,11 +219,20 @@ def translate_as_fieldop(
     fun_node = node.fun
     assert len(fun_node.args) == 2
     stencil_expr, domain_expr = fun_node.args
+
+    if cpm.is_ref_to(stencil_expr, "deref"):
+        # Special usage of 'deref' as argument to fieldop expression, to pass a scalar
+        # value to 'as_fieldop' function. It results in broadcasting the scalar value
+        # over the field domain.
+        return translate_broadcast_scalar(node, sdfg, state, sdfg_builder, reduce_identity)
+
```
philip-paul-mueller, line 228: This is completely optional.
But moving the check to here:
```
elif isinstance(stencil_expr, gtir.Lambda):
        pass  # Handled below
else:
        raise NotImmplemtedError(...)
```
It would make it more obvious what is handled below.
Alternatively you could move the default case also into its own function.

philip-paul-mueller (approved):
I have one optional recommendation, but it looks good to me.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: Much nicer, agreed.

PR #1694, opened 2024-10-17
feat[next][dace]: GTIR-to-DaCe lowering of map-reduce with skip values

edopao:
This PR extends the solution for map-reduce provided in #1683 with the support for connectivity tables with skip values.

The field definition is extended with a `local_offset` attribute that stores the offset provider used to build the values in the local dimension. In case the local dimension is built by the `neighbors` expression, the `local_offset` corresponds to the offset provider used to access the neighbor dimension. Since this information is carried along the data itself, whenever the data is accessed it is also possible to access the corresponding offset provider and check whether the neighbor index is valid or if there is a skip value. For local dimensions already present in the program argument, this information is retreived from the field domain (enabled in new test case).

The data is accessed in the `map_` and `reduce` expressions. Here it is now possible to check for skip values. Therefore, the main objective of this PR is the lowering of map-reduce with skip values.

A secondary objective is to pave the road to simplify the lowering logic, by getting rid of the `reduce_identity` value. The current approach is propagate the `reduce_identity` value while visiting the arguments to `reduce` expressions. By introducing `local_offset`, the argument visitor will return the information needed to implement `reduce` of local values in presence of skip values. This solution is available as a draft PR, see [this diff](https://github.com/edopao/gt4py/compare/dace-gtir-map_reduce_skip...edopao:gt4py:dace-gtir-reduce_identity).


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -31,18 +31,42 @@
 
 @dataclasses.dataclass(frozen=True)
 class DataExpr:
```
philip-paul-mueller, line 33: What is the difference between this class and `Field`, they have the same members.
I am not against that per se, if they are used in different contexts and the name provides more hints about what is going on.
But, in that case there _must_ be a detailed description what differentiates this class from the other.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -64,16 +88,20 @@ class IteratorExpr:
     Args:
         field: The field this iterator operates on.
         dimensions: Field domain represented as a sorted list of dimensions.
-                    In order to dereference an element in the field, we need index values
```
philip-paul-mueller, line 67: I would rephrase this and combine the sentences.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -64,16 +88,20 @@ class IteratorExpr:
...
-                    for all the dimensions in the right order.
+            In order to dereference an element in the field, we need index values
+            for all the dimensions in the right order.
         indices: Maps each dimension to an index value, which could be either a symbolic value
-                 or the result of a tasklet computation like neighbors connectivity or dynamic offset.
+            or the result of a tasklet computation like neighbors connectivity or dynamic offset.
+        local_offset: Must be set for fields containing `ListType` data generated
+            from neighbors access in unstructured domain, in which case it indicates
+            the name of the offset provider used to generate the list of neighbor values.
```
philip-paul-mueller, line 97: See my comment about `DataExpr`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -152,7 +180,7 @@ def connect(
         mx: dace.nodes.MapExit,
         result_node: dace.nodes.AccessNode,
         subset: sbs.Range,
-    ) -> None:
+    ) -> str | None:
```
philip-paul-mueller, line 183: You should add a description that tells the user what it returns and why.
For my part, I do not really see why this function should return the local offset, it does not really make sense to me, at least at this point of the review.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -467,7 +498,9 @@ def _visit_deref(self, node: gtir.FunCall) -> ValueExpr:
                         assert isinstance(index_expr, SymbolExpr)
 
                 dtype = arg_expr.field.desc(self.sdfg).dtype
-                return self._construct_tasklet_result(dtype, deref_node, "val")
+                return self._construct_tasklet_result(
+                    dtype, deref_node, "val", local_offset=arg_expr.local_offset
```
philip-paul-mueller, line 502: It looks a bit strange that only the last argument is passed by its name.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -599,8 +632,17 @@ def _visit_map(self, node: gtir.FunCall) -> DataExpr:
         fun_python_code = gtir_python_codegen.get_source(fun_node)
         tasklet_expression = f"{output_connector} = {fun_python_code}"
 
-        # TODO(edopao): extract offset_dim from the input arguments
-        offset_dim = gtx_common.Dimension("", gtx_common.DimensionKind.LOCAL)
+        input_local_offsets = [
+            input_arg.local_offset for input_arg in input_args if input_arg.local_offset is not None
+        ]
+        if len(input_local_offsets) == 0:
+            raise ValueError(f"Missing information on local dimension for map node {node}.")
```
philip-paul-mueller, line 639: The description indicate that `local_offset` is only set for unstructured.
So what about reductions in fully structured meshes?

Furthermore, as far as I know there is one global reduction, I am not fully sure, but I guess they are over the edges. So how do you handle such cases (where you do not reduce over a local dimension but a vertical/horizontal one?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -599,8 +632,17 @@ def _visit_map(self, node: gtir.FunCall) -> DataExpr:
...
+
+        local_offset = input_local_offsets[0]
```
philip-paul-mueller, line 641: You tested if there is one, but what if there are multiple?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -599,8 +632,17 @@ def _visit_map(self, node: gtir.FunCall) -> DataExpr:
...
+        offset_provider = self.subgraph_builder.get_offset_provider(local_offset)
+        assert isinstance(offset_provider, gtx_common.Connectivity)
```
philip-paul-mueller, line 643:
```suggestion
        assert isinstance(self.subgraph_builder.get_offset_provider(local_offset), gtx_common.Connectivity)
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -599,8 +632,17 @@ def _visit_map(self, node: gtir.FunCall) -> DataExpr:
...
+        local_size = offset_provider.max_neighbors
```
philip-paul-mueller, line 644: This might be a very naïve question, but assume you have a grid were every point has 6 neighbours, then 10 is a valid value of `_max_neighbors`, although it is not a very good one.
Before you were inspecting the results, so it might be possible that you found that you only need 6.
So I would have have kept the old inspection, and if I did not found a result at the end used the max value.

However, I am not sure how realistic this situation is and if I explained it well enough (I am bad at that).

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -599,8 +632,17 @@ def _visit_map(self, node: gtir.FunCall) -> DataExpr:
...
+        offset_dim = gtx_common.Dimension(local_offset, gtx_common.DimensionKind.LOCAL)
```
philip-paul-mueller, line 645: Next dump question, but is the offset provider also always the name of local dimension it produces.
Shouldn't it have something like a `Dim` at the end?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -611,7 +653,7 @@ def _visit_map(self, node: gtir.FunCall) -> DataExpr:
         # than representing map-to-map edges (which require memlets with 2 pass-nodes).
         input_memlets = {}
         input_nodes = {}
-        local_size: Optional[int] = None
+        skip_value_connectivities = []
```
philip-paul-mueller, line 656:
```suggestion
        skip_value_connectivities: list[gtx_common.Connectivity] = []
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -621,22 +663,53 @@ def _visit_map(self, node: gtir.FunCall) -> DataExpr:
...
-            assert len(input_nodes) >= 1
-            local_size = 1
-
         out, _ = self.sdfg.add_temp_transient((local_size,), dtype)
         out_node = self.state.add_access(out)
 
+        if len(skip_value_connectivities) != 0:
+            # In case one or more of input expressions contain skip values, we use
+            # the connectivity-based offset provider as mask for map computation.
+            local_offset, offset_provider = skip_value_connectivities[0]
```
philip-paul-mueller, line 690: What if you have multiple?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -621,22 +663,53 @@ def _visit_map(self, node: gtir.FunCall) -> DataExpr:
...
+            origin_map_index = dace_gtir_utils.get_map_variable(offset_provider.origin_axis)
+
+            connectivity_slice = self._construct_local_view(
+                MemletExpr(
+                    self.state.add_access(connectivity),
+                    sbs.Range.from_string(f"{origin_map_index}, 0:{offset_provider.max_neighbors}"),
+                )
+            )
+
+            assert self.reduce_identity is not None
```
philip-paul-mueller, line 705: Optionally: I would promote this to an exception.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -1386,13 +1391,46 @@ def test_gtir_reduce_dot_product():
...
+        [
+            [x if i != skip_idx else gtx_common._DEFAULT_SKIP_VALUE for i, x in enumerate(row)]
+            for skip_idx, row in zip(
+                np.random.randint(0, connectivity_V2E.max_neighbors, size=SIMPLE_MESH.num_vertices),
+                connectivity_V2E.table,
+                strict=True,
+            )
+        ],
+        dtype=connectivity_V2E.table.dtype,
+    )
```
philip-paul-mueller, line 1407: Are you sure that this generates skip values?
Could you add an assert that checks if there are skip values, just to put my mind at ease.

philip-paul-mueller (commented):
There are some points that needs improvement but it generally looks good.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I don't know how that looks like in the IR.. all gt4py unit tests and all icon4py tests pass.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: True, it is not needed. Some old code.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -31,18 +31,42 @@
...
+    Local storage for the computation result returned by a tasklet node.
+
+    Arguments:
+        node: Access node to the data storage, can be either a scalar or a local list.
+        dtype: GT4Py type definition, which includes the field domain information.
+        local_offset: Must be set for `FieldType` with a local dimension generated
+            from neighbors access in unstructured domain, and indicates the name
+            of the offset provider used to generate the list of neighbor values.
+            It is always 'None' for scalar data. `FieldType` with 'local_offset=None'
+            contain only global (horizontal or vertical) dimensions.
```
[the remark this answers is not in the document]
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -31,18 +31,42 @@
...
+    Scalar or array data access through a memlet.
+
+    Arguments:
+        node: Access node to the data storage, can be either a scalar or a local list.
+        subset: Represents the subset to use in memlet to access the above data.
+        local_offset: Must be set for `FieldType` with a local dimension generated
+            from neighbors access in unstructured domain, and indicates the name
+            of the offset provider used to generate the list of neighbor values.
+            It is always 'None' for scalar data. `FieldType` with 'local_offset=None'
+            contain only global (horizontal or vertical) dimensions.
```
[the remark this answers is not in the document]
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: `DataExpr` is used in the context in a dataflow, a.k.a. a stencil expression. Therefore, it either represent scalar values (elements in the fields) or lists of local values. The other type (now renamed to `FieldopData`) represents the result of field operators, basically the data storage outside global maps.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
philip-paul-mueller: Then I would add a description about this, because it looks strange.
An alternative would also be to modify `get_map_variable()` to accept different arguments.
I mean it is Python and Python has ducktyping.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
philip-paul-mueller: But stuff is added to `skip_value_connectivities` if there are skip values.
The concrete value is irrelevant, only the fact that they are present matters.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -638,6 +637,14 @@ def _visit_map(self, node: gtir.FunCall) -> DataExpr:
...
             raise ValueError(f"Missing information on local dimension for map node {node}.")
 
+        # GT4Py guarantees that all connectivities used to generate lists of neighbors
+        # have the same length, that is the same value of 'max_neighbors'.
+        assert 1 == len(
+            {
+                self.subgraph_builder.get_offset_provider(offset).max_neighbors  # type: ignore[union-attr]
+                for offset in input_local_offsets
+            }
+        )
```
philip-paul-mueller, line 647:
```suggestion
        assert len(input_local_offsets) == 1
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -675,19 +682,18 @@ def _visit_map(self, node: gtir.FunCall) -> DataExpr:
...
 
-        out, _ = self.sdfg.add_temp_transient((local_size,), dtype)
+        out, _ = self.sdfg.add_temp_transient((local_size,), dc_dtype)
         out_node = self.state.add_access(out)
 
         if len(skip_value_connectivities) != 0:
             # In case one or more of input expressions contain skip values, we use
             # the connectivity-based offset provider as mask for map computation.
-            local_offset, offset_provider = skip_value_connectivities[0]
+            # GT4Py guarantees that all input expressions have the same skip values.
```
philip-paul-mueller, line 695: It is not about the value, it is about having them or not.

philip-paul-mueller (commented):
Some small questions.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: It is not equivalent.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I will change the signature of the function.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: Yes, it is also about the values. The skip values should appear at the same neighbor index for all arguments to map.

edopao:
cscs-ci run default

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -639,18 +638,19 @@ def _visit_map(self, node: gtir.FunCall) -> ValueExpr:
...
             }
         )
-        local_offset = input_local_offsets[0]
-        offset_provider = self.subgraph_builder.get_offset_provider(local_offset)
-        assert isinstance(offset_provider, gtx_common.Connectivity)
+        if len(set(table.max_neighbors for table in local_offset_providers.values())) != 1:
+            raise ValueError(
+                "Unexpected arguments to map expression with different local dimensions."
+            )
+        local_offset, offset_provider = next(iter(local_offset_providers.items()))
```
philip-paul-mueller, line 651: Bellow on line 679 you override this values.
I think it is confusing to create and define a variable outside a loop and then overwrite (not update or modify) it inside a loop.
You should rewrite it such that you do not have this.
The same is true for `local_offset`.

This is actually a feature I like in rust that everything is constant except you explicitly mark it as mutual.
And that you are forced to write curly brackets in `if` expressions.

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -15,12 +15,17 @@
 from gt4py.next.type_system import type_specifications as ts
 
 
-def get_map_variable(dim: gtx_common.Dimension) -> str:
+def get_map_variable(axis: gtx_common.Dimension | str) -> str:
```
philip-paul-mueller, line 18:
```suggestion
def get_map_variable(dim: gtx_common.Dimension | str) -> str:
```

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -15,12 +15,17 @@
...
     """
     Format map variable name based on the naming convention for application-specific SDFG transformations.
     """
+    if isinstance(axis, gtx_common.Dimension):
+        dim = axis
+    elif len(axis) != 0:
+        dim = gtx_common.Dimension(axis, gtx_common.DimensionKind.LOCAL)
```
philip-paul-mueller, line 25:
```suggestion
    if isinstance(axis, str) and len(axis) != 0:
        dim = gtx_common.Dimension(axis, gtx_common.DimensionKind.LOCAL)
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -34,9 +34,25 @@
...
+    Args:
+        dc_node: DaCe access node to the data storage.
+        gt_dtype: GT4Py type definition, which includes the field domain information.
+        local_offset: Provides information about the local dimension in`FieldType` data.
+            Set to 'None' for scalar data. Can be 'None' for `FieldType` data with
+            only global (horizontal or vertical) dimensions.
+    """
+
+    dc_node: dace.nodes.AccessNode
+    gt_dtype: ts.FieldType | ts.ScalarType
```
philip-paul-mueller, line 54: I really much like the prefix.
They make the code a bit longer but it is always clear what kind of thing you are looking at.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -337,36 +363,37 @@ def _construct_local_view(self, field: MemletExpr | DataExpr) -> DataExpr:
...
         else:
             return field
 
     def _construct_tasklet_result(
         self,
-        dtype: dace.typeclass,
+        dc_dtype: dace.typeclass,
         src_node: dace.nodes.Tasklet,
         src_connector: str,
+        local_offset: Optional[str] = None,
```
philip-paul-mueller, line 385: You have these classes that combines certain things like `FieldOpsData`, why are you passing the data individually here and not encapsulated inside them?

Or is this function too low level?

philip-paul-mueller (commented):
Some suggestions, but I think we are very close now.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I cannot use `FieldopData` in this case. `FieldopData` is supposed to contain fields with global domain. The tasklet result constructed here is in the local domain: by that I mean it returns just a scalar that represents an intermediate result of an element-wise operation.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: You will easily convince me by just mentioning Rust 🦀 

I agree, overwriting a variable at some point in a loop is hard to follow, in any programming language. In this case, it is even not needed.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -30,30 +30,54 @@
...
+    Args:
+        dc_node: Access node to the data storage, can be either a scalar or a local list.
+        gt_dtype: GT4Py type definition, which includes the field domain information.
+        local_offset: Provides information about the local dimension in`FieldType` data.
+            For a more detailed explanation see `gtir_builtin_translators.FieldopData`.
+    """
+
+    dc_node: dace.nodes.AccessNode
+    gt_dtype: itir_ts.ListType | ts.ScalarType
+    local_offset: Optional[str] = None
```
edopao, line 52: I've just seen that @havogt is implementing a similar feature in the IR, in #1703. I will merge my PR but probably later, when #1703 is merged, I will do some cleanup in the dace lowering because `local_offset` (a local dimension corresponding to the neighbors offset provider) can be stored inside `itir_ts.ListType`.

edopao:
cscs-ci run default

philip-paul-mueller (approved):
LGTM

PR #1695, opened 2024-10-20
feat[next][dace]: Cleanup scalar args

edopao:
This PR simplifies the representation of program scalar arguments in the SDFG: instead of promoting them to symbols, we represent them as scalar data. Scalar to symbol promotion is a transformation pass available in DaCe that should be applied after lowering to SDFG, eventually.

Additional changes:
- Fixes a problem in naming of input connectors for scalar expression tasklets: the connector name cannot match the argument name, otherwise we cannot pass the same value to two connectors in expressions like `out = tmp * tmp`.
- Only propagate to lambda scope the symbols that are referenced.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
@@ -261,10 +261,6 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
 
         Returns:
             A list of array nodes containing the result fields.
-
-        TODO: Do we need to return the GT4Py `FieldType`/`ScalarType`? It is needed
-        in case the transient arrays containing the expression result are not guaranteed
-        to have the same memory layout as the target array.
         """
         result = self.visit(node, sdfg=sdfg, head_state=head_state, reduce_identity=None)
 
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
@@ -37,7 +37,7 @@ from gt4py.next.program_processors.runners.dace_fieldview import (
 from gt4py.next.type_system import type_specifications as ts, type_translation as tt
 
 
-GTIR_DOMAIN_DEFAULT_SYMBOLS: Final[set[str]] = {
+GTIR_DOMAIN_SYMBOLS: Final[set[str]] = {
     "horizontal_start",
     "horizontal_end",
     "vertical_start",
@@ -222,13 +222,13 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
             dtype = dace_utils.as_dace_type(symbol_type)
             # Scalar program arguments are represented as scalar data in the SDFG;
             # the exception are program arguments used for symbolic domain expressions,
-            # listed in `GTIR_DOMAIN_DEFAULT_SYMBOLS`, and those used for symbolic
-            # array shape and strides.
+            # listed in `GTIR_DOMAIN_SYMBOLS`, and those used for symbolic array
+            # shape and strides.
             # The field size is sometimes passed as a scalar argument to the program,
             # so we have to check if the shape symbol was already allocated by
             # `_make_array_shape_and_strides`. We assume that the scalar argument
             # for field size always follows the field argument.
-            if name in GTIR_DOMAIN_DEFAULT_SYMBOLS:
+            if name in GTIR_DOMAIN_SYMBOLS:
                 sdfg.add_symbol(name, dtype)
             elif dace_utils.is_field_symbol(name):
                 if name in sdfg.symbols:
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
@@ -113,6 +113,10 @@ def _collect_symbols_in_domain_expressions(
     """
     Collect symbols accessed in domain expressions that also appear in the paremeter list.
 
+    This function is used to identify all parameters that are accessed in domain
+    expressions. They have to be passed to the SDFG call as DaCe symbols (instead
+    of scalars) such that they can be used as bounds in map ranges.
+
     Args:
         ir: GTIR node to be traversed and where to search for domain expressions.
         ir_params: List of parameters to search for in domain expressions.
@@ -201,7 +205,7 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
     def _add_storage(
         self,
         sdfg: dace.SDFG,
-        known_symbols: set[str],
+        symbol_arguments: set[str],
         name: str,
         gt_type: ts.DataType,
         transient: bool = True,
@@ -209,9 +213,9 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
         """
         Add storage in the SDFG for a given GT4Py data symbol.
 
-        GT4Py fields are allocated as dace arrays. GT4Py scalars are represented
-        as scalar objects in the SDFG; the exception are the symbols passed as
-        `known_symbols`, e.g. symbols used in domain expressions, and those used
+        GT4Py fields are allocated as DaCe arrays. GT4Py scalars are represented
+        as DaCe scalar objects in the SDFG; the exception are the symbols passed as
+        `symbol_arguments`, e.g. symbols used in domain expressions, and those used
         for symbolic array shape and strides.
 
         The fields used as temporary arrays, when `transient = True`, are allocated
@@ -220,7 +224,7 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
 
         Args:
             sdfg: The SDFG where storage needs to be allocated.
-            known_symbols: Set of GT4Py scalars that must be represented as SDFG symbols.
+            symbol_arguments: Set of GT4Py scalars that must be represented as SDFG symbols.
             name: Symbol Name to be allocated.
             gt_type: GT4Py symbol type.
             transient: True when the data symbol has to be allocated as internal storage.
@@ -236,7 +240,7 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
                 name, gt_type, flatten=True
             ):
                 tuple_fields.extend(
-                    self._add_storage(sdfg, known_symbols, tname, tsymbol_type, transient)
+                    self._add_storage(sdfg, symbol_arguments, tname, tsymbol_type, transient)
                 )
             return tuple_fields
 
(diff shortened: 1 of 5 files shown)


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -490,7 +501,10 @@ def visit_Lambda(
         ]
 
         # inherit symbols from parent scope but eventually override with local symbols
-        lambda_symbols = self.global_symbols | {
+        lambda_symbols = {
+            sym: self.global_symbols[sym]
+            for sym in symbol_ref_utils.collect_symbol_refs(node.expr, self.global_symbols.keys())
+        } | {
             pname: dace_gtir_utils.get_tuple_type(arg) if isinstance(arg, tuple) else arg.data_type
```
philip-paul-mueller, line 508: Just a question.
As far as I can tell this is strictly speaking not necessary.
It just filters out the symbols that are not used.
Is this correct?

philip-paul-mueller (commented):
There are some small things, but nothing major.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -37,6 +37,18 @@
...
 
+GTIR_DOMAIN_SYMBOLS: Final[set[str]] = {
+    "horizontal_start",
+    "horizontal_end",
+    "vertical_start",
+    "vertical_end",
+}
+"""
+Set of IR symbols that are passed as gt4py scalar arguments but we have to represent
+in the SDFG as dace symbols, because they are used in domain symbolic expressions.
```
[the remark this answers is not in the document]
edopao: Valid point. I can make it generic by implementing an IR visitor that extracts all symbols used in domain expressions.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -106,6 +107,34 @@ def visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
         ...
 
 
+def _collect_symbols_in_domain_expressions(
+    ir: gtir.Node, ir_params: Sequence[gtir.Sym]
+) -> set[str]:
+    """
+    Collect symbols accessed in domain expressions that also appear in the paremeter list.
```
philip-paul-mueller, line 114: I would also add for what is it used.
Something like:
This function is used to identify all program arguments that are used in domain expressions.
They have to be passed as symbols (instead of scalars) such that they can be used as bounds.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -172,60 +201,73 @@ def _make_array_shape_and_strides(
...
         """
-        Add storage for data containers used in the SDFG. For fields, it allocates dace arrays,
-        while scalars are stored as SDFG symbols.
+        Add storage in the SDFG for a given GT4Py data symbol.
 
-        The fields used as temporary arrays, when `transient = True`, are allocated and exist
-        only within the SDFG; when `transient = False`, the fields have to be allocated outside
-        and have to be passed as array arguments to the SDFG.
+        GT4Py fields are allocated as dace arrays. GT4Py scalars are represented
+        as scalar objects in the SDFG; the exception are the symbols passed as
```
philip-paul-mueller, line 213: I would write "DaCe scalar objects" to make it clear because we have to distinguish between different scalars (that are completely different) here.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -172,60 +201,73 @@ def _make_array_shape_and_strides(
...
+        `known_symbols`, e.g. symbols used in domain expressions, and those used
+        for symbolic array shape and strides.
+
+        The fields used as temporary arrays, when `transient = True`, are allocated
+        and exist only within the SDFG; when `transient = False`, the fields have
+        to be allocated outside and have to be passed as arguments to the SDFG call.
+
+        Args:
+            sdfg: The SDFG where storage needs to be allocated.
+            known_symbols: Set of GT4Py scalars that must be represented as SDFG symbols.
```
philip-paul-mueller, line 223: I am thinking that `known_symbols` is not a very good name and could be misleading.
I would suggest a name such as "symbol_arguments" or similar.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -172,60 +201,73 @@ def _make_array_shape_and_strides(
...
-            if is_tuple_member:
-                sdfg.add_scalar(name, dtype, transient=transient)
-            elif name in sdfg.symbols:
-                assert sdfg.symbols[name].dtype == dtype
-            else:
+        elif isinstance(gt_type, ts.ScalarType):
+            dtype = dace_utils.as_dace_type(gt_type)
+            if name in known_symbols:
                 sdfg.add_symbol(name, dtype)
+            elif dace_utils.is_field_symbol(name):
```
philip-paul-mueller, line 256: What is the difference with the `if` branch just above.
From the descriptions (comment below and the definition of `known_symbols`) I would say that both branches address the same situation.

philip-paul-mueller (commented):
Some smallish things.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: Agreed.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: They address the same situation but in a slightly different context.

If you take the gt4py unit tests, the range of cartesian domain is typically defined as [0, size].
Sometimes, when the field domain is implicitly derived from the field domain, the gt4py lowering adds the field size. Suppose a field `__sym`, then gt4py will add `__sym_size`. It never does this for the strides.

`program_sdfg(__sym1, __sym1_size_0, size, __sym1_stride_0)`

We are using `known_symbols` for the symbols used in domain expressions. The array symbols are derived from the field name (`__sym1` -> `__sym1_size_0, __sym1_stride_0`)

This if handles the specific case where gt4py adds the scalar argument for `__sym1_size_0`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -172,60 +205,73 @@ def _make_array_shape_and_strides(
     def _add_storage(
         self,
         sdfg: dace.SDFG,
+        symbol_arguments: set[str],
```
philip-paul-mueller, line 208: This is optional, but I would say `symbolic_arguments` would be a bit better from a grammar point of view.

-- Note whenever I argue on this level (grammar) be extremely wary.

Do not implement this suggestion:
You have a lot of argument that you pass around now (essentially the reduction identity is the same).
I would suggest that at some point you collect them all in one object that you then pass around.
Thus if you then notices that you have to add something new (in case of `symbl_arguments`) you do not have to change it everywhere.
In my view this is what encapsulation is.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -172,60 +205,73 @@ def _make_array_shape_and_strides(
...
+            symbol_arguments: Set of GT4Py scalars that must be represented as SDFG symbols.
+            name: Symbol Name to be allocated.
+            gt_type: GT4Py symbol type.
+            transient: True when the data symbol has to be allocated as internal storage.
 
         Returns:
-            List of data containers or symbols allocated as storage. This is a list, not a single value,
-            because in case of tuples we flat the tuple fields (eventually nested) and allocate storage
-            for each tuple element.
+            List of data containers or symbols allocated as storage. This is a list,
```
philip-paul-mueller, line 233: The description could be interpreted that you return the container itself, however, you return only the name.
Furthermore, you return a list of tuples, and the second element is not described.

Optional: The second sentence sounds a little bit odd, it reads like an apology.
I would write something like "In case of a tuple the list contains a flattened version, otherwise the list will only contain a single entry."

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
philip-paul-mueller: Thanks for the explanation.
I know that behaviour but I did not associate it with it.
However, I think it does this only in very specific cases, Till once explained this to me.

Optionally: Ask Till what the conditions are that this happens and note them down here and also make a link to it.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -274,14 +316,28 @@ def make_temps(field: gtir_builtin_translators.Field) -> gtir_builtin_translator
         temp_result = gtx_utils.tree_map(make_temps)(result)
         return list(gtx_utils.flatten_nested_tuple((temp_result,)))
 
-    def _add_sdfg_params(self, sdfg: dace.SDFG, node_params: Sequence[gtir.Sym]) -> list[str]:
-        """Helper function to add storage for node parameters and connectivity tables."""
+    def _add_sdfg_params(
```
philip-paul-mueller, line 319: Why is this function not using `_collect_symbols_in_domain_expressions()` which has a very similar description purpose?

Probably you should rephrase the description to something like:
"GT4Py `Fields` arguments will be translated to `dace.data.Array` (We should make a PR that makes `Array` accessible as `dace.Array`). By default GT4Py Scalars arguments will be translated to `dace.data.Scalar` arguments, except the they are listed in `symbol_arguments`/`symbolic_arguments`, in which case they will be symbols inside the SDFG."

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -501,11 +563,12 @@ def visit_Lambda(
         nstate = nsdfg.add_state("lambda")
 
         # add sdfg storage for the symbols that need to be passed as input parameters
+        lambda_params = [
+            gtir.Sym(id=p_name, type=p_type) for p_name, p_type in lambda_symbols.items()
+        ]
+        lambda_domain_symbols = _collect_symbols_in_domain_expressions(node, lambda_params)
```
philip-paul-mueller, line 569: This is just a question.
Why are you scanning here `node` but above you are scanning `node.expr`?

philip-paul-mueller (approved):
I have some optional recommendations.
However, nothing that would warrant another round of review.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: Done.

edopao:
cscs-ci run default

PR #1696, opened 2024-10-20
feat[next][dace]: Add helper method to replace unicode symbols

edopao:
DaCe requires C-compatible strings for the names of data containers, such as arrays and scalars.
GT4Py uses a unicode symbols (`ᐞ`) as name separator in the SSA pass, which generates invalid symbols for DaCe.
This PR introduces a helper method to find new names for invalid symbols present in the IR.


havogt (commented):
I would propose to remove DaCe from this pass, to be able to use it in other backends too. We might be too agressive, but I don't think it matters if we do e.g. a unicode->ascii trafo.

edopao:
> I would propose to remove DaCe from this pass, to be able to use it in other backends too. We might be too agressive, but I don't think it matters if we do e.g. a unicode->ascii trafo.

I can make a generic pass to replace a dictionary of symbols passed as argument. However, the detection of invalid symbols and the choice of replacement symbols should be backend-specific, no?

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -66,3 +70,40 @@ def get_tuple_type(data: tuple[Any, ...]) -> ts.TupleType:
...
+
+        def visit_SymRef(self, node: gtir.SymRef, *, symtable: Dict[str, str]) -> gtir.SymRef:
+            return self._replace_sym(node, symtable)
+
+    # program arguments are checked separetely, because they cannot be replaced
+    if not all(dace.dtypes.validate_name(str(sym.id)) for sym in ir.params):
+        raise ValueError("Invalid symbol in program parameters.")
+
+    invalid_symbols_mapping = {
+        sym_id: sdfg.temp_data_name()
```
edopao, line 102: @havogt Note here that I am using a dace-specific way to generate new valid symbols: the name string will be valid (because generated by dace itself) and the function will also assure that that symbol is not yet used in the SDFG.

havogt (approved):
let's take it for now

src/gt4py/next/program_processors/runners/dace_fieldview/utility.py
```diff
@@ -66,3 +70,40 @@ def get_tuple_type(data: tuple[Any, ...]) -> ts.TupleType:
...
+        for sym in eve.walk_values(ir).if_isinstance(gtir.SymRef).to_set()
```
edopao, line 103: Is it maybe more logical to check for `gtir.Sym` instead of `gtir.SymRef` here? I mean searching for symbol definitions rather than references to symbols?

PR #1698, opened 2024-10-21
bug[next][dace]: Fix lowering of broadcast

edopao:
Fix lowering of `as_fieldop` with broadcast expression after changes in #1701.

Additional change:
- Add support for GT4Py zero-dimensional fields, equivalent of numpy zero-dimensional arrays. Test case added.


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -346,15 +346,26 @@ def translate_broadcast_scalar(
...
+        arg_node = scalar_expr.node
+        arg_type = node.args[0].type
+    elif isinstance(node.args[0].type, ts.FieldType):
+        # zero-dimensional field
+        assert len(node.args[0].type.dims) == 0
+        assert isinstance(scalar_expr, gtir_dataflow.IteratorExpr)
+        arg_node = scalar_expr.field
+        arg_type = node.args[0].type.dtype
+    else:
+        raise ValueError(f"Unexpected argument {node.args[0]} in broadcast expression.")
```
edopao, line 365: This should probably be a `NotImplementedError`, in case in future we are going to support broadcasting of field slice.

src/gt4py/next/program_processors/runners/dace_common/dace_backend.py
```diff
@@ -26,6 +26,9 @@
 def _convert_arg(arg: Any, sdfg_param: str, use_field_canonical_representation: bool) -> Any:
     if not isinstance(arg, gtx_common.Field):
         return arg
+    if len(arg.domain.dims) == 0:
+        # pass zero-dimensional fields as scalars
+        return arg.asnumpy().item()
```
havogt, line 31: @egparedes What was the correct pattern for avoiding type conversions?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -144,6 +144,31 @@ def _parse_fieldop_arg(
...
+    accesses with offset -2 (which corresponds to -lower_bound). Instead, we choose
+    to allocate (N-2), leaving positions [0:2] unused. The reason is that array offset
+    is known to cause issues to SDFG inlining. Besides, map fusion will in any case
+    eliminate most of transient arrays.
+
+    Args:
+        domain: The field operator domain.
+
+    Returns:
+        A list of dimensions and a list of array sizes in each dimension.
```
philip-paul-mueller, line 166: It does not return _a_ (in the sense of a single one) but a tuple of them.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -359,19 +374,48 @@ def translate_broadcast_scalar(
...
+        input_subset = (
+            str(scalar_expr.subset) if isinstance(scalar_expr, gtir_dataflow.MemletExpr) else "0"
+        )
+        input_node = scalar_expr.dc_node
+        gt_dtype = node.args[0].type
+    elif isinstance(node.args[0].type, ts.FieldType):
+        assert isinstance(scalar_expr, gtir_dataflow.IteratorExpr)
+        if len(node.args[0].type.dims) == 0:  # zero-dimensional field
+            input_subset = "0"
+        elif not all(
```
philip-paul-mueller, line 398: I am not a friend of negative logic, so I would check if all are symbolic expressions and then handle the chase and in the else branch handle the error.
I.e. flipping the two branches.

philip-paul-mueller (approved):
There is a small change ion a description and a suggestion about negative logic.
However, it LGTM.
But you should ping @egparedes again to address what Hannes has pointed out.

src/gt4py/next/program_processors/runners/dace_common/dace_backend.py
edopao: This type conversion is not affecting GT4Py. The GT4Py argument has still type `ts.FieldType` with empty `dims` list. This type conversion (zero-dim array to scalar) only affects the argument passed to the SDFG call, because in the SDFG I am representing the zero-dim array as a dace scalar object.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: Fixed.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: Good point.

src/gt4py/next/program_processors/runners/dace_common/dace_backend.py
egparedes: I haven't reviewed the PR so I don't have a lot of context here, but I think Hannes' comment is about extracting the scalar value from the 0d numpy array without changing its type, because `ndarray.item()` always transforms the scalar value to a python scalar, which may change its precision. To avoid this, you should using the empty tuple as index for `.__getitem__()` like this.
```suggestion
        return arg.asnumpy()[()]
```

edopao: Interesting, thank you.

PR #1699, opened 2024-10-22
feat[next]: Index builtin

SF-N:
Adds index builtin for embedded and gtfn backends.


src/gt4py/next/iterator/pretty_parser.py
```diff
@@ -31,9 +31,9 @@
     INT_LITERAL: SIGNED_INT
     FLOAT_LITERAL: SIGNED_FLOAT
     OFFSET_LITERAL: ( INT_LITERAL | CNAME ) "ₒ"
-    _literal: INT_LITERAL | FLOAT_LITERAL | OFFSET_LITERAL
+    AXIS_LITERAL: CNAME ("ᵥ" | "ₕ")
```
SF-N, line 34: Renamed for consistency.

src/gt4py/next/iterator/pretty_parser.py
```diff
@@ -31,9 +31,9 @@
...
+    _literal: INT_LITERAL | FLOAT_LITERAL | OFFSET_LITERAL | AXIS_LITERAL
```
SF-N, line 35: Otherwise `index(OFFSET_LITERAL)` isn't parsed properly.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -611,7 +611,6 @@ def convert_el_to_sid(el_expr: Expr, el_type: ts.ScalarType | ts.FieldType) -> E
                 tuple_constructor=lambda *elements: SidComposite(values=list(elements)),
             )
 
-            assert isinstance(lowered_input_as_sid, (SidComposite, SidFromScalar, SymRef))
```
SF-N, line 614: This check was superfluous from the beginning since the format is checked anyway when the `SetAt` is constructed. We just removed it instead of allowing an `index` FunCall.

havogt (commented):
a few comments

src/gt4py/next/iterator/pretty_parser.py
SF-N: Yes

src/gt4py/next/iterator/embedded.py
```diff
@@ -1156,6 +1156,7 @@ def ndarray(self) -> core_defs.NDArrayObject:
     def asnumpy(self) -> np.ndarray:
         raise NotImplementedError()
 
+    # TODO(tehrengruber): Use a regular zero dimensional field instead.
```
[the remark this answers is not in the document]
tehrengruber: We moved the comment to a better place. With respect to the comment itself: The `IndexField` class is rather strange. It has two modes of operation: Either it is a field with (conceptually) `field(domain) == domain` or it is a zero-dimensional field. Both modes don't share any implementation similarities, but are mushed into the same class. The way the class behaves is then controlled using `_cur_index`. It would be much simpler to just make `field[index]` return a zero-dimensional field which is exactly what we want instead of re-implementing it here.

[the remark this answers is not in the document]
havogt: I don't understand what you are saying...

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -260,6 +260,17 @@ def visit_Program(self, node: gtfn_ir.Program, **kwargs: Any) -> Union[str, Coll
     #include <gridtools/fn/${grid_type_str}.hpp>
     #include <gridtools/fn/sid_neighbor_table.hpp>
     #include <gridtools/stencil/global_parameter.hpp>
+    #include <gridtools/stencil/positional.hpp>
+    
+    // TODO(havogt): move to gtfn?
```
havogt, line 265: I'd propose to wait until https://github.com/GridTools/gridtools/pull/1806 is merged next week.

havogt: Once #1720 is in, we can remove this.

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -260,6 +260,17 @@ def visit_Program(self, node: gtfn_ir.Program, **kwargs: Any) -> Union[str, Coll
...
+    namespace gridtools{
+    namespace fn{
+    template <class T>
+    auto index(T){
+    return stencil::positional<std::decay_t<T>>();}
+    }
+    }
+
```
havogt, line 273:
```suggestion
```

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
SF-N: Thanks, I removed that code.

havogt (approved):
lgtm

PR #1701, opened 2024-10-22
bug[next]: foast2gtir lowering of broadcasted field

havogt:
Wrap every broadcast in an `as_fieldop` (not only scalars). The materialization of intermediate broadcasted fields need to be optimized by transformations.


src/gt4py/next/ffront/foast_to_gtir.py
```diff
@@ -374,9 +374,7 @@ def create_if(true_: itir.Expr, false_: itir.Expr) -> itir.FunCall:
 
     def _visit_broadcast(self, node: foast.Call, **kwargs: Any) -> itir.FunCall:
         expr = self.visit(node.args[0], **kwargs)
-        if isinstance(node.args[0].type, ts.ScalarType):
-            return im.as_fieldop(im.ref("deref"))(expr)
-        return expr
+        return im.as_fieldop(im.ref("deref"))(expr)
```
tehrengruber, line 377:
```suggestion
        return im.as_fieldop("deref")(expr)
```

PR #1704, opened 2024-10-23
feat[next][dace]: Stop carrying reduce identity in args context

edopao:
This PR implements an alternative design to local expressions on lists containing skip values. The previous design was based on the assumption that such operations only exist in the context of a reduction (which is the most common case, but necessarily the only one) and was carrying the reduce identity value from a reduce node to the visiting context of the arguments and all their child nodes. The reduce identity value was used till fill the skip values when building lists of neighbors or when applying map expressions. Although quite effective for reduce expressions, this design led to a very complicated code. Besides, this design was mixing lowering with optimization, by implicitly optimizing the SDFG.

With this PR, we stop carrying the reduce identity to the arguments context. Instead, in presence of skip values we just write a dummy value. When it is time to reduce, we override the dummy value with the reduce identity. In this way, the reduce identity is only used in the context of the reduction expression.


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -739,66 +733,149 @@ def _visit_map(self, node: gtir.FunCall) -> ValueExpr:
...
+                {"__val"},
+                f"__val = {reduce_init.dc_dtype}({reduce_init.value})",
+            ),
+            "__val",
+            st_init.add_access("acc"),
+            None,
+            dace.Memlet(data="acc", subset="0"),
+        )
+        st_reduce = nsdfg.add_state_after(st_init, "reduce")
+        st_reduce.add_mapped_tasklet(
```
philip-paul-mueller, line 804: Your map has default schedule, which depending on the situation might be turned into a parallel schedule, which requires atomic reduction.
You should explicitly set it to sequential, which is a reasonable assumption for GT4Py (see later).
In that case you should also set `wcr_nonatomic` (this is a Memlet property) to `True`.
If you do that please add the following todo as well
```
TODO(phimuell): decide if auto optimizer should reset `wcr_nonatomic` properties, as DaCe does it.
```

Note if you follow my suggestions above, reductions along any dimension will be sequential, which is according to my understanding of Gt4Py the majority of cases, as you reduce along a local dimension.
However, this also applies to reduction along every dimensions not only locals but horizontal or vertical dimensions too!
This is certainly okay for local and probably to some degree and in some special situations for vertical, but certainly not for horizontal dimensions.
For that reason I would add a switch that checks (for now) if the reduction is along a local dimension and then do what you are doing now or otherwise puts a reduction library node there.

Furthermore, because you collect them in a local array, the memory loading is not combined, i.e. the address thread 2 loads is not adjacent to thread 1.
For that you would need to collect the data in a fundamentally different way, but for now this should be fine. 
However, this has no priority (please add some todo if you agree with me) and we should talk with Felix/Hannes/Yannis before we do that.
In the long run we will have to add this into a library node, or at least make a custom expansion.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -739,66 +733,149 @@ def _visit_map(self, node: gtir.FunCall) -> ValueExpr:
...
         else:
-            reduce_axes = None
+            local_dim_indices = list(range(len(desc.shape)))
+
+        assert len(local_dim_indices) == 1
+        local_dim_index = local_dim_indices[0]
+        assert desc.shape[local_dim_index] == offset_provider.max_neighbors
+
+        # we lower the reduction map with WCR out memlet in a nested SDFG
+        nsdfg = dace.SDFG("reduce_with_skip_values")
```
philip-paul-mueller, line 776: You should make the name unique.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -739,66 +733,149 @@ def _visit_map(self, node: gtir.FunCall) -> ValueExpr:
...
+            strides=(desc.strides[local_dim_index],),
+        )
+        nsdfg.add_array(
+            "neighbor_indices",
+            (connectivity_desc.shape[1],),
+            connectivity_desc.dtype,
+            strides=(connectivity_desc.strides[1],),
+        )
+        nsdfg.add_scalar("acc", desc.dtype)
+        st_init = nsdfg.add_state("init")
```
philip-paul-mueller, line 790: Give it a name that relates it to the nested SDFG, it is easier to find this way.
This also applies to the reduction state.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -739,66 +733,149 @@ def _visit_map(self, node: gtir.FunCall) -> ValueExpr:
...
+            ),
+            "__val",
+            st_init.add_access("acc"),
+            None,
+            dace.Memlet(data="acc", subset="0"),
+        )
+        st_reduce = nsdfg.add_state_after(st_init, "reduce")
+        st_reduce.add_mapped_tasklet(
+            name="reduce_with_skip_values",
+            map_ranges=dict(i=f"0:{offset_provider.max_neighbors}"),
```
philip-paul-mueller, line 806: Everywhere else you are using `{}` except here.
```suggestion
            map_ranges={"i": f"0:{offset_provider.max_neighbors}"},
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -739,66 +733,149 @@ def _visit_map(self, node: gtir.FunCall) -> ValueExpr:
...
+            code=f"__out = __val if __neighbor_idx != {gtx_common._DEFAULT_SKIP_VALUE} else {reduce_identity.dc_dtype}({reduce_identity.value})",
+            outputs={
+                "__out": dace.Memlet(data="acc", subset="0", wcr=reduce_wcr),
+            },
+            external_edges=True,
+        )
 
-        reduce_wcr = "lambda x, y: " + gtir_python_codegen.format_builtin(op_name, "x", "y")
-        reduce_node = self.state.add_reduce(reduce_wcr, reduce_axes, reduce_init.value)
+        nsdfg_node = self.state.add_nested_sdfg(
```
philip-paul-mueller, line 818: What is about the strides that you use, I would say you have to pass them through the symbol mapping.

philip-paul-mueller (commented):
There are no major problems but some aspects I would like to make you aware of

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I don't think we need to set the map schedule and the `wcr_nonatomic`. It is up to DaCe to decide the schedule for this loop and, based on that decision, whether the WCR should use atomics or not.


Note that we will never encounter skip values on global dimensions. Currently, the lowering supports only reduction along a single dimension (corresponding to the neighbors local dimension) and there is already a check for that: `assert len(local_dim_indices) == 1`
I could turn the assert into an exception.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: The strides are exactly the same as the parent SDFG, so they are automatically mapped.

philip-paul-mueller: Are you sure. I have never seen that feature.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -713,16 +708,15 @@ def _visit_map(self, node: gtir.FunCall) -> ValueExpr:
...
-                )
-            assert self.reduce_identity.dc_dtype == dc_dtype
             input_memlets["__neighbor_idx"] = dace.Memlet(
                 data=connectivity_slice.dc_node.data, subset=map_index
             )
             input_nodes[connectivity_slice.dc_node.data] = connectivity_slice.dc_node
-            tasklet_expression += f" if __neighbor_idx != {gtx_common._DEFAULT_SKIP_VALUE} else {dc_dtype}({self.reduce_identity.value})"
+
+            tasklet_expression += (
+                # in case of skip value we can write any dummy value
```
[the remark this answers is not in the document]
philip-paul-mueller: This was one of the first PRs I did in DaCe.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
philip-paul-mueller: This are exactly the points that are hard to find out on the SDFG level and a lot can go wrong.
Why spend compute time to figuring out something that can be set correctly from construction.
Here GT4Py has to help the optimization, because recovering the information is super hard and it would be much simpler, just to emit the right/smart constructs.
On the SDFG there is no fool prove way to figuring out what a local dimension is, on which you can make certain assumptions.
This is the whole point of having a DSL to have more information what you want to do, we have to put that knowledge to work.

philip-paul-mueller (commented):
There is some discussion needed about reduction.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -573,9 +577,13 @@ def _visit_neighbors(self, node: gtir.FunCall) -> ValueExpr:
         }
 
         if offset_provider.has_skip_values:
+            # in case of skip value we can write any dummy value
+            if ti.is_floating_point(node.type.element_type):
+                skip_value = "math.nan"
+            else:
+                skip_value = str(dace.dtypes.max_value(field_desc.dtype))
```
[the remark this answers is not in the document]
edopao: Agreed.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I think DaCe just looks at the hierarchy of the maps and applies the parallel schedule (GPU global or CPU OpenMP) on the outerst map. If the schedule is not set, the map will be sequential. This is case of all inner maps with the exception of GPU tiling transformation, probably. The setting of the map schedule is done at the very end, I think in code generation for CPU (but I would have to check). So this kind of reduce maps, that operate on a local dimension, will by definition always be inner maps and therefore be left with sequential schedule.

Anyway, I also get your point and I don't have a strong opinion against setting the schedule to sequential, because anyway that is what we want to achieve in the end.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: For both shape and strides, I allocate the inner arrays using the same symbols as in parent data descriptors:
```
        nsdfg.add_array(
            "values",
            (desc.shape[local_dim_index],),
            desc.dtype,
            strides=(desc.strides[local_dim_index],),
        )
        nsdfg.add_array(
            "neighbor_indices",
            (connectivity_desc.shape[1],),
            connectivity_desc.dtype,
            strides=(connectivity_desc.strides[1],),
        )
```
The symbols that are not present in the symbol mapping are automatically inherited from the parent SDFG. I know that if they were missing there, `add_nested_sdfg` would throw an error because of undefined symbol.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
philip-paul-mueller: Thanks.


This morning I realized something but I am not sure if it is actually done.
But you do reduction in one swoop, i.e. the replacement of skip values and the reduction is done in the same Tasklet?
It is not that first an array is build filled and then reduced.
If so the order must be changed, because then the initialization between the two loops blocks their fusing.

I am still thinking that this should be a library node, at least in the long run.
Because then we could at expansion time, put the initialization at the beginning of the scope, and avoid putting the reduction loop inside a nested SDFG.
This would allow to fuse it, because so the node can not be fused.

philip-paul-mueller (approved):
LGTM.

Please see my comment.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I totally agree about the usage of a library node, in the long-term, not only for reduce but also for neighbors and map. In this way we could transform and eventually fuse different expressions on local dimensions.

About the current implementation. The tasklet just writes a scalar value, either the input value or the reduce identity, into the accumulator scalar node. This write memlet uses a WCR to accumulate the result, rather than overwriting it.

> It is not that first an array is build filled and then reduced.

Unfortunately not. This was the original implementation, before I found a bug in the MapReduceFusion transformation which is used in the auto-optimize pipeline (you should maybe consider to remove it, it contains a bug). The bug is discussed here: https://chat.spcl.inf.ethz.ch/spcl/pl/8mtgtqjb378hfy7h9a96sy3nhc

> If so the order must be changed, because then the initialization between the two loops blocks their fusing.

Which two loops?

philip-paul-mueller: > you should maybe consider to remove it

I actually wanted to do that. I will make a PR right away.

> Which two loops?

This was a reference to the original implementation.
Where we had a collecting loop and a reduction loop.
Because, the nested SDFG was in between the two could not be fused together.

PR #1706, opened 2024-10-24
feat[next]: Allow partial type inference on ITIR

tehrengruber:


havogt (changes requested):
Let's add a few unittest, than it should be good to go. With these changes all gtir-fieldview-embedded tests pass.

havogt (approved):
lgtm

PR #1707, opened 2024-10-24
fix[next][dace]: Avoid bool cast in branch condition expressions

edopao:
Casting the condition expression on an inter-state edge to `bool` prevented dead-state elimination in the SDFG in case the expression was specialized to True/False.


philip-paul-mueller (approved):
LGTM.

PR #1708, opened 2024-10-24
[draft][next]: Add support for float16 and bfloat16

SF-N:
Added support for `float16` and `bfloat16`.
Tyeh are not yet supported in dace.


havogt:
Should we close this and re-start the work when needed. Custom nanobind seems to be not needed anymore, but there are still other questions, e.g. c++23?

PR #1717, opened 2024-10-29
ci: use actions/checkout@v4 and actions/setup-python@v5

romanc:
## Description

This PR updates the GitHub Actions (GHA) workflows to use `actions/checkout@v4` instead of `v3` (or `v2`  in the cartesian case) and `actions/setup-python@v5`. Development for `actions/checkout@v3` and `actions/setup-python@v4` stopped ~1 year ago and GH is currently enforcing newer node versions than the one that these actions were designed with, leading to the following warnings

image
_warnings in next workflows_

image
_warnings in cartesian workflows_

`deploy_release` action was following `actions/checkout@master`. Was this on purpose? Happy to revert if so. Unless there's a good reason, I suggest to keep all actions pinned at ideally the same major version.

`pre-commit/action` was updated to keep its dependencies up to date and avoid transitive warnings similar to the ones above.

No changes made to currently disabled workflows, i.e. the ones under `.github/workflows/_disabled/`.

Parent: https://github.com/GEOS-ESM/SMT-Nebulae/issues/89

## Requirements

  CI is running as before - just with fewer warnings.
  N/A

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: "3.10"
         cache: 'pip'
@@ -24,4 +24,4 @@ jobs:
           **/pyproject.toml
           **/constraints.txt
           **/requirements-dev.txt
-    - uses: pre-commit/action@v3.0.0
+    - uses: pre-commit/action@v3.0.1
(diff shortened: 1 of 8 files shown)


romanc:
@twicki / @FlorianDeconinck continuing the quest to reduce warnings ...

@egparedes / @havogt please review or re-distribute to whoever is knowledgeable / in charge of CI for the next version of gt4py

As far as I can tell, we should be good again for now. We might want to look into automatic reporting of outdated GHA versions in the future.

romanc:
We should also be aware the `pre-commit/action` is in [maintenance mode only](https://github.com/pre-commit/action) (see top of their README file). I wouldn't be very concerned. If we are happy with what we have now, it's super straight forward to just implement ourselves (instead of using their composite action) in case it ever drops support completely.

egparedes:
cscs-ci run

egparedes (approved):
LGTM

havogt (approved):

PR #1720, opened 2024-11-01
build: update dependencies

havogt:
updates minimum gridtools_cpp version for #1699

### DIFF
--- a/src/gt4py/next/ffront/dialect_parser.py
+++ b/src/gt4py/next/ffront/dialect_parser.py
@@ -106,11 +106,11 @@ class DialectParser(ast.NodeVisitor, Generic[DialectRootT]):
 
         # `FixMissingLocations` ensures that all nodes have the location attributes
         assert hasattr(node, "lineno")
-        line = node.lineno + line_offset if node.lineno is not None else None
+        line = node.lineno + line_offset
         assert hasattr(node, "end_lineno")
         end_line = node.end_lineno + line_offset if node.end_lineno is not None else None
         assert hasattr(node, "col_offset")
-        column = 1 + node.col_offset + col_offset if node.col_offset is not None else None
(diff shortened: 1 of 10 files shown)


src/gt4py/next/ffront/dialect_parser.py
```diff
@@ -106,11 +106,11 @@ def get_location(self, node: ast.AST) -> SourceLocation:
 
         # `FixMissingLocations` ensures that all nodes have the location attributes
         assert hasattr(node, "lineno")
-        line = node.lineno + line_offset if node.lineno is not None else None
+        line = node.lineno + line_offset
```
havogt, line 109: let's see if that works (mypy complain because line, colum are not Optional)

tehrengruber: Ah great they fixed this in python.

src/gt4py/next/utils.py
```diff
@@ -140,7 +140,7 @@ def impl(*args: Any | tuple[Any | tuple, ...]) -> _R | tuple[_R | tuple, ...]:
                 assert result_collection_constructor is not None
                 return result_collection_constructor(impl(*arg) for arg in zip(*args))
 
-            return fun(  # type: ignore[misc] # mypy not smart enough
+            return fun(  # type: ignore[call-arg, misc] # mypy not smart enough
```
tehrengruber, line 143: The cast is probably superfluous together with `call-arg`, but whatever.

tehrengruber (approved):

src/gt4py/next/utils.py
havogt: I double checked. The cast actually fixes a `arg-type` error. The `call-arg` is for "too few arguments", not motivated to investigate further.

PR #1725, opened 2024-11-05
build: add devcontainer setup

havogt:
Add devcontainer configuration with special customizations for VS Code.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,4 @@
+FROM mcr.microsoft.com/devcontainers/python:1-3.10-bookworm
+RUN apt-get update \
+    && export DEBIAN_FRONTEND=noninteractive && apt-get install -y libboost-dev \
+    && apt-get clean && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/* && rm -rf /tmp/*
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,3 +2,4 @@ FROM mcr.microsoft.com/devcontainers/python:1-3.10-bookworm
 RUN apt-get update \
     && export DEBIAN_FRONTEND=noninteractive && apt-get install -y libboost-dev \
     && apt-get clean && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/* && rm -rf /tmp/*
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh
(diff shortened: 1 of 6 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM mcr.microsoft.com/devcontainers/python:1-3.10-bookworm
+RUN apt-get update \
+    && export DEBIAN_FRONTEND=noninteractive && apt-get install -y libboost-dev \
+    && apt-get clean && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/* && rm -rf /tmp/*
```
egparedes, line 4: Can we also add some basic python tooling like `pipx` & `uv`? For example:
```suggestion
    && export DEBIAN_FRONTEND=noninteractive && apt-get install -y libboost-dev \
    && apt-get install pipx && ensurepath && pipx install uv \
    && apt-get clean && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/* && rm -rf /tmp/*
```

.devcontainer/setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+ln -sfn /workspaces/gt4py/.devcontainer/.vscode /workspaces/gt4py/.vscode
+python -m venv .venv
+source .venv/bin/activate
+pip install --upgrade pip setuptools wheel
+pip install -e .
+pip install -r requirements-dev.txt
+pip install -i https://test.pypi.org/simple/ atlas4py
```
egparedes, line 9: If we install `uv` in the dockerfile then we can use it here which should be nicer and much faster:
```suggestion
uv venv .venv
source .venv/bin/activate
uv pip install -r requirements-dev.txt
uv pip install -e .
uv pip install -i https://test.pypi.org/simple/ atlas4py
```

egparedes (commented):
Just a couple of suggestions to use `uv`:

havogt:
Should we just `--user` or `--system` install instead of venv? Since the devcontainer is anyway an isolated environment...

egparedes:
> Should we just `--user` or `--system` install instead of venv? Since the devcontainer is anyway an isolated environment...

Yes, I agree. I don't think we need a venv inside the container

egparedes:
@havogt Is there any blocker to get this updated and merged? I'd like to start using it soon.

egparedes:
I've added some vscode extensions and settings to the proper customization point in devcontainer.json. It looks good to me so I'm approving the PR but you should also take a look to my changes and modify then as you wish.

egparedes (approved):
LGTM

PR #1726, opened 2024-11-05
fix[next][dace]: Fix for nested SDFG outer data descriptor

edopao:
Fixes a bug in lowering of let-statements. The lambda expression of a let-statement is lowered to a nested SDFG. The result data produced in the nested SDFG is written to temporary data allocated in the parent SDFG. The previous lowering was directly using a copy of the inner data descriptor for the outer data. The bug is that some symbols for array shape and strides might not be available in the parent SDFG, so we have to apply the symbol mapping on the outer data descriptor.

The test case `test_gtir_let_lambda_with_cond` was modified to trigger this bug and verify the fix.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
@@ -710,46 +710,51 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
             head_state.add_edge(src_node, None, nsdfg_node, connector, memlet)
 
         def make_temps(
-            output_data: gtir_builtin_translators.FieldopData,
+            inner_data: gtir_builtin_translators.FieldopData,
         ) -> gtir_builtin_translators.FieldopData:
             """
             This function will be called while traversing the result of the lambda
             dataflow to setup the intermediate data nodes in the parent SDFG and
             the data edges from the nested-SDFG output connectors.
             """
-            desc = output_data.dc_node.desc(nsdfg)
-            if desc.transient:
+            inner_desc = inner_data.dc_node.desc(nsdfg)
+            if inner_desc.transient:
                 # Transient nodes actually contain some result produced by the dataflow
                 # itself, therefore these nodes are changed to non-transient and an output
                 # edge will write the result from the nested-SDFG to a new intermediate
                 # data node in the parent context.
-                desc.transient = False
-                temp, _ = sdfg.add_temp_transient_like(desc)
-                connector = output_data.dc_node.data
-                dst_node = head_state.add_access(temp)
+                inner_desc.transient = False
+                outer, outer_desc = sdfg.add_temp_transient_like(inner_desc)
+                # We cannot use a copy of the inner data descriptor directly, we have to apply the symbol mapping.
+                dace.symbolic.safe_replace(
+                    nsdfg_symbols_mapping,
+                    lambda m: dace.sdfg.replace_properties_dict(outer_desc, m),
+                )
+                connector = inner_data.dc_node.data
+                outer_node = head_state.add_access(outer)
                 head_state.add_edge(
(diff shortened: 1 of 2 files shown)


tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -1756,11 +1756,7 @@ def test_gtir_let_lambda_with_cond():
             gtir.SetAt(
                 expr=im.let("x1", "x")(
                     im.let("x2", im.op_as_fieldop("multiplies", domain)(2.0, "x"))(
-                        im.if_(
-                            gtir.SymRef(id="pred"),
-                            im.as_fieldop(im.lambda_("a")(im.deref("a")), domain)("x1"),
-                            im.as_fieldop(im.lambda_("a")(im.deref("a")), domain)("x2"),
-                        )
+                        im.if_("pred", "x1", "x2")
```
edopao, line 1759: The change to `test_gtir_let_lambda_with_cond` covers the bugfix, all other edits in this file are just refactoring.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -710,46 +710,51 @@ def _flatten_tuples(
...
             the data edges from the nested-SDFG output connectors.
             """
-            desc = output_data.dc_node.desc(nsdfg)
-            if desc.transient:
+            inner_desc = inner_data.dc_node.desc(nsdfg)
+            if inner_desc.transient:
                 # Transient nodes actually contain some result produced by the dataflow
                 # itself, therefore these nodes are changed to non-transient and an output
                 # edge will write the result from the nested-SDFG to a new intermediate
                 # data node in the parent context.
```
philip-paul-mueller, line 725: Not part of the PR, but more correct would be saying something like "Although the data only exists within the SDFG, because it serves as an output of the nested SDFG it must be global" or something like that.
But feel free to ignore.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -710,46 +710,51 @@ def _flatten_tuples(
             head_state.add_edge(src_node, None, nsdfg_node, connector, memlet)
 
         def make_temps(
-            output_data: gtir_builtin_translators.FieldopData,
+            inner_data: gtir_builtin_translators.FieldopData,
         ) -> gtir_builtin_translators.FieldopData:
             """
             This function will be called while traversing the result of the lambda
             dataflow to setup the intermediate data nodes in the parent SDFG and
             the data edges from the nested-SDFG output connectors.
```
philip-paul-mueller, line 718: I do not like the description:
- It is a bit cryptic
- It does not impose any restriction on `inner_data` (it must be part of a nested SDFG).
- What happens if the translation is not needed, i.e. `inner_data` is already global.

Also the name sounds a bit strange, it is super general, but the function is actually super specific as it only works on nested SDFGs.
How about renaming it to `create_output_for_nested_sdfg()` or `map_nested_data_container_to_parent_sdfg()` or `map_data_from_nested_sdfg()`?

An alternative description would be:
This function makes a data container, denoted by `inner_data`, that lives inside a nested SDFG available in the parent SDFG.
For this the data container inside the nested SDFG will be marked as global (a requirement of the SDFG IR) and create a new data container within the parent SDFG, with the same properties (shape, stride, etc.) appropriate remapped using the symbol mapping table.
If `inner_data` was already mapped into the parent SDFG then [I do not know what happens then, but I guess that thing is returned?}

philip-paul-mueller (approved):
The changes looks good.
However, I would propose to change the description and the name of the `make_temps()` function, see my comments.

PR #1727, opened 2024-11-06
ci: Bump gitlab ci on todi to ubuntu 22.04, cuda 12.6.2, cupy 13.3.0

tehrengruber:
We were using ubuntu 22.04 which shipped with gcc 9.x.x. In order to get something more recent with proper utf-8 support I bumped to 22.04 on todi. On daint strange hangs occured so I kept everything as is there.


ci/install_cuda_driver.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+CUDA_DRIVER_VERSION=$1
+
+echo "Installing CUDA driver version $CUDA_DRIVER_VERSION"
+apt-get -yqq install --no-install-recommends kmod wget
+wget -q https://us.download.nvidia.com/XFree86/Linux-x86_64/${CUDA_DRIVER_VERSION}/NVIDIA-Linux-x86_64-${CUDA_DRIVER_VERSION}.run
```
edopao, line 6: This script seems to be specific for x86 arch, but the Dockerfile file is common to x86 and aarm64. You could pass an additional argument `ARCH` to the container build.
On  the other hand, I would really try to avoid this custom driver installation.

tehrengruber: Absolutely, this was just a test to see if this was the issue.

ci/cscs-ci.yml
```diff
@@ -48,11 +48,14 @@ stages:
   variables:
     CUDA_VERSION: 11.2.2
     CUPY_PACKAGE: cupy-cuda11x
+    CUPY_VERSION: 12.3.0 # latest version that supports cuda 11
+    UBUNTU_VERSION: 20.04  # 22.04 hangs on daint in some tests for unknown reasons.
```
edopao, line 52: You need to pass UBUNTU_VERSION as docker build argument at line 45.

edopao: and CUPY_VERSION as well (btw, you need to use this variable in Dockerfile).

edopao:
Shall we also trigger the ARM jobs on Todi?

tehrengruber:
Oh, yes absolutely.

edopao (approved):

PR #1728, opened 2024-11-06
feat[next]: Extend the IR pass for pruning of unnecessary casts

edopao:
Extend the IR pass delivered in #1688 

Pruning of cast expressions may appear as a `as_fieldop` expression with form  `(⇑(λ(__val) → cast_(·__val, float64)))(a)`, where `a` is already a field with data type `float64` in this example. This PR adds pruning of such trivial expressions.


edopao:
We need to keep `as_fieldop` expressions with `deref` lambda inside because this is the lowered form of broadcast.

src/gt4py/next/iterator/transforms/prune_casts.py
```diff
@@ -21,24 +35,27 @@ class PruneCasts(PreserveLocationVisitor, NodeTranslator):
...
+            assert isinstance(lambda_node, ir.Lambda)
+            cast_expr = lambda_node.expr
+            assert isinstance(cast_expr, ir.FunCall)
+            cast_value = cast_expr.args[0]
+            if (
+                _is_cast_redundant(cast_expr)
+                and cpm.is_call_to(cast_value, "deref")
+                and cpm.is_ref_to(cast_value.args[0], lambda_node.params[0].id)
+            ):
+                return self.visit(node.args[0])
```
tehrengruber, line 52:
```suggestion
        if cpm.is_identity_as_fieldop(node):
            return node.args[0]
```

```python
def is_identity_as_fieldop(node: itir.Expr):
  if not is_applied_as_fieldop(node):
    return False
  stencil = node.fun.args[0]
  if is_ref_to(stencil, "deref"):
    return True
  if isinstance(stencil, itir.Lambda) and len(stencil.params) == 1 and stencil == im.lambda(stencil.params[0])(im.deref(stencil.params[0].id)):
    return True
  return False
```

Just a suggestion to avoid code duplication. The if clause here does the same `_is_cast_redundant(node)` check as below while the suggested version simply removes `as_fieldop(deref)(a)`.

tehrengruber (approved):
As discussed I think the approach here is wrong. The extension of the pass is purely driven by the fact that we want to avoid an additional materialization / memory copy of a field (i.e. remove the `as_fieldop(deref)(field)`) in the dace backend. All data-flow related optimization should happen on SDFG level not on ITIR. I approved nonetheless, but I believe we should pick up the discussion at some point again.

PR #1730, opened 2024-11-07
fix[next]: Bugfix in dace-ITIR backend: use canonical name for field shape symbol

edopao:
Shape and stride symbols are expected to match a canonical string pattern. This PR adopts the canonical pattern in ITIR dace backend.

Note: it solves an issue encountered in dace orchestration tests.


DropD (approved):
LGTM

PR #1731, opened 2024-11-08
feat[next][dace]: Skip lowering of trivial field operators that just copy the field argument

edopao:
It was observed the `prune_casts` IR pass might leave trivial `as_fieldop` expressions that where lowered to copy-maps in the SDFG. With this PR, we prune such expressions in the lowering phase (not in the IR, nor in the SDFG transformations) because the lowering still has domain information.

Additionally, a helper `ir_maker` function is introduced: `cast_as_fieldop`. Tests are updated to use it.

### DIFF
--- a/src/gt4py/next/iterator/ir_utils/ir_makers.py
+++ b/src/gt4py/next/iterator/ir_utils/ir_makers.py
@@ -511,11 +511,11 @@ def cast_as_fieldop(type_: str, domain: Optional[itir.FunCall] = None):
 
     Examples:
         >>> str(cast_as_fieldop("float32")("a"))
-        '(⇑(λ(__arg) → cast_(·__arg, float32)))(a)'
+        '(⇑(λ(__val) → cast_(__val, float32)))(a)'
     """
 
     def _impl(it: itir.Expr) -> itir.FunCall:
-        return as_fieldop(lambda_("__arg")(call("cast_")(deref("__arg"), type_)), domain)(it)
+        return as_fieldop(lambda_("__val")(call("cast_")(deref("__val"), type_)), domain)(it)
 
     return _impl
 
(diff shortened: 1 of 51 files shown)


src/gt4py/next/iterator/ir_utils/ir_makers.py
```diff
@@ -498,6 +498,28 @@ def _impl(*its: itir.Expr) -> itir.FunCall:
     return _impl
 
 
+def cast_as_fieldop(type_: str, domain: Optional[itir.FunCall] = None):
```
philip-paul-mueller, line 501: I am not 100% sure if this is the right place.
If it is the right place (in the sense of lack of alternatives) I would suggest that you change the word "Promote" from the description (the name of the functionname is okay).
You see for me a promotion is not a maker but something that changes something else that is already there, which is as far as I can tell from the example, not the case.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -246,9 +246,16 @@ def translate_as_fieldop(
...
 
     if isinstance(stencil_expr, gtir.Lambda):
-        # Default case, handled below: the argument expression is a lambda function
-        # representing the stencil operation to be computed over the field domain.
-        pass
+        if cpm.is_call_to(stencil_expr.expr, "deref") and cpm.is_ref_to(
+            stencil_expr.expr.args[0], stencil_expr.params[0].id
+        ):
+            # Prune trivial field operators with the form '(⇑(λ(__arg) → ·__arg))(a)'
+            # that are doing element-wise copy of the input field.
```
philip-paul-mueller, line 253: As far as I understood our discussion, this will translate the Map that does the copy into a Memlet?
I would add to the description above to what this case is translated to.

philip-paul-mueller (approved):
LGTM.
There are two suggestions regarding the doc strings, but if you do not agree, feel free to ignore.

edopao:
Solution in #1728 is preferable.

PR #1690, opened 2024-10-15
feat[next]: Add memory and disk-based caching to more workflow steps

SF-N:
Add memory and disk-based caching to other workflow steps and, therefore, removing unnecessary overhead of Program calls and significantly improving time to first computed value.

Changes:

- setting `cached = True` for `func_to_past_factory`
- wrapping the GTFN code generation into a `CachedStep` (using `Diskcache`) which is activated when setting `otf_workflow__cached_translation=True`, similar as in [PR#1474](https://github.com/GridTools/gt4py/pull/1474) (without CachedStep)
- Fixing hash function of `ProgramDefinition`
- New tests for added functionality

This leads to a runtime decrease of about 25% for PMAP-G in the advect-uniform testcase (5 hours) after caches are populated.

TODOs: 
- [x] improving hash functions of `fingerprint_stage`

PR #1733, opened 2024-11-12
bug[next]: Fix ITIR program hash stability

tehrengruber:
#1690 included a change to make the hash of an `itir.FencilDefinition` stable across multiple runs. This PR adopts the same change to an `itir.Program`,


egparedes (approved):
LGTM

PR #1648, opened 2024-09-18
feat[next]: GTIR embedded and GTFN temporaries with new lowering

tehrengruber:
Use new lowering for GTIR embedded, and GTFN. Only the dace iterator backend continues to use the old lowering.

Changes:
- Use GTIR lowering for all backends except for dace
- Old lowering and transformations only used in dace backend
  - workflows defined in [`gt4py.next.backend.LEGACY_TRANSFORMS`](https://github.com/GridTools/gt4py/pull/1648/files#diff-cf4385d02cbeacc310d4326350903b4cb6f9a61c7cd36dda162a5077ab8b8e86). Variable can be removed in a cleanup PR.
  - old `apply_common_transforms` in [pass_manager_legacy.py](https://github.com/GridTools/gt4py/pull/1648/files#diff-db17bff48ac16ee75ff974a1b9af98e3cf0c850971ce9898aa55b635bb046b72). Just a straight copy of the old function. No need to review, this is just to avoid deleting until gtir based dace backend is ready.
  - Re-add `symbolic_sizes` param. Was in temporary extraction, is now part of the domain inference. In preparation of icon-exclaim tests

PR #1738, opened 2024-11-14
feat[next]: Inline dynamic shifts

tehrengruber:
Dynamic shifts are not supported in the domain inference. In order to make them work nonetheless this PR aggressively inlines all arguments to `as_fieldop` until they contain only references to `itir.Program` params. Additionally the domain inference is extended to tolerate such `as_fieldop` by introducing a special domain marker that signifies a domain is unknown.


tests/next_tests/definitions.py
```diff
@@ -148,7 +151,7 @@ class ProgramFormatterId(_PythonObjectIdMixin, str, enum.Enum):
     (USES_ZERO_DIMENSIONAL_FIELDS, XFAIL, UNSUPPORTED_MESSAGE),
     (STARTS_FROM_GTIR_PROGRAM, SKIP, UNSUPPORTED_MESSAGE),
 ]
-GTIR_DACE_SKIP_TEST_LIST = [
+GTIR_DACE_SKIP_TEST_LIST = DOMAIN_INFERENCE_SKIP_LIST + [
     (USES_DYNAMIC_OFFSETS, XFAIL, UNSUPPORTED_MESSAGE),
```
edopao, line 155: In order to see if DaCe GTIR backend supports dynamic offsets, we would have to delete line 155 (or I can wait for next merge commit, after #1648 is merged.

edopao: I tested it locally and it works.
image

edopao:
I run icon4py stencils and I found that one small change is needed in dace backend. It makes sense to include it in this PR:
```
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
@@ -1057,7 +1057,6 @@ class LambdaToDataflow(eve.NodeVisitor):
         """Implements shift in unstructured domain by means of a neighbor table."""
         assert connectivity.neighbor_axis in it.dimensions
         neighbor_dim = connectivity.neighbor_axis
-        assert neighbor_dim not in it.indices
 
         origin_dim = connectivity.origin_axis
         assert origin_dim in it.indices
```

tehrengruber:
Feel free to push your Dace changes (as long as they are isolated to dace parts otherwise it'll be to confusing).

src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
```diff
@@ -67,6 +67,107 @@ def _is_tuple_expr_of_literals(expr: itir.Expr):
     return isinstance(expr, itir.Literal)
```
tehrengruber, line 67: All changes in this file are merely a small refactoring that extracts the field operator fusion for a single `as_fieldop` from the `FuseAsFieldOp` pass into a dedicated function `fuse_as_fieldop` used by the `InlineDynamicShifts` pass.

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -10,7 +10,7 @@
```
tehrengruber, line 10: @sf-n could you review the changes in this file?

SF-N (commented):
Thanks, the changes make sense to me. The only thing one could reconsider is the name `allowed_unknown_domains` as it somehow implies to be a collection of domains rather than references.

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -423,5 +476,10 @@ def infer_program(
         function_definitions=program.function_definitions,
         params=program.params,
         declarations=program.declarations,
-        body=[_infer_stmt(stmt, offset_provider, symbolic_domain_sizes) for stmt in program.body],
+        body=[
+            _infer_stmt(
+                stmt, offset_provider, symbolic_domain_sizes, [param.id for param in program.params]
```
havogt, line 481:
```suggestion
                stmt, offset_provider, symbolic_domain_sizes, allowed_unknown_domains=[param.id for param in program.params]
```
for readability

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -362,30 +398,45 @@ def infer_expr(
...
+        raise ValueError("Some accessed domains are unknown, e.g. because of a dynamic shift.")
+
     return expr, accessed_domains
 
 
 def _infer_stmt(
     stmt: itir.Stmt,
     offset_provider: common.OffsetProvider,
     symbolic_domain_sizes: Optional[dict[str, str]],
+    allowed_unknown_domains: list[str],
```
havogt, line 429: In general, I am confused why this is introduced... 
Are all program parameters allowed to be unknown, but the rest isn't?

havogt: If I understand the logic, then I would remove it from here and rather write a separate path for checking this and avoid uglifying the this interface.

src/gt4py/next/iterator/transforms/inline_dynamic_shifts.py
```diff
@@ -0,0 +1,76 @@
...
+    params_shifts = trace_shifts.trace_stencil(
+        node.fun.args[0],  # type: ignore[attr-defined]  # ensured by is_applied_as_fieldop
+        num_args=len(node.args),
+        save_to_annex=True,
+    )
+    dynamic_shifts = []
+    for param_shifts in params_shifts:
+        has_dynamic_shift = False
+        for shifts in param_shifts:
+            for _, offset in zip(shifts[::2], shifts[1::2], strict=True):
```
havogt, line 32: either 
```suggestion
            for offset in shifts[1::2]:
```
or use a function if you insist that does proper error checking.

src/gt4py/next/iterator/transforms/inline_dynamic_shifts.py
```diff
@@ -0,0 +1,76 @@
...
+                has_dynamic_shift |= offset == trace_shifts.Sentinel.VALUE
+        dynamic_shifts.append(has_dynamic_shift)
+    return dynamic_shifts
```
havogt, line 35:
```suggestion
        dynamic_shifts = [any(trace_shifts.Sentinel.VALUE in shifts for shifts in param_shifts) for param_shifts in params_shifts]
```

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -157,5 +161,8 @@ def apply_fieldview_transforms(
     ir = inline_fundefs.prune_unreferenced_fundefs(ir)
     ir = InlineLambdas.apply(ir, opcount_preserving=True, force_inline_lambda_args=True)
     ir = CollapseTuple.apply(ir, offset_provider=offset_provider)  # type: ignore[assignment] # type is still `itir.Program`
+    ir = inline_dynamic_shifts.InlineDynamicShifts.apply(
+        ir
+    )  # domain inference does not support dynamic offsets yet
     ir = infer_domain.infer_program(ir, offset_provider=offset_provider)
```
havogt, line 167: Would it make sense to support unknown domain in the type inference for `as_offset`?

havogt (changes requested):

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -58,9 +58,12 @@ def _split_dict_by_key(pred: Callable, d: dict):
...
 def _domain_union_with_none(
-    *domains: domain_utils.SymbolicDomain | None,
-) -> domain_utils.SymbolicDomain | None:
-    filtered_domains: list[domain_utils.SymbolicDomain] = [d for d in domains if d is not None]
+    *domains: domain_utils.SymbolicDomain | None | Literal["UNKNOWN"],
+) -> domain_utils.SymbolicDomain | None | Literal["UNKNOWN"]:
+    if any(d == "UNKNOWN" for d in domains):
+        return "UNKNOWN"
+
+    filtered_domains: list[domain_utils.SymbolicDomain] = [d for d in domains if d is not None]  # type: ignore[misc]  # domain can never be none because as these cases are filtered above
```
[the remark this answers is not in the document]
tehrengruber: I renamed it `allow_uninferred` which now also allows if the domain is never accessed (previously indicated by `None` now `DomainAccessDescriptor.NEVER`)

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -25,7 +25,7 @@
 from gt4py.next.utils import flatten_nested_tuple, tree_map
 
 
-DOMAIN: TypeAlias = domain_utils.SymbolicDomain | None | tuple["DOMAIN", ...]
+DOMAIN: TypeAlias = domain_utils.SymbolicDomain | None | Literal["UNKNOWN"] | tuple["DOMAIN", ...]
```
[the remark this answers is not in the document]
tehrengruber: I introduced an enum `DomainAccessDescriptor` which indicates `UNKNOWN` or `NEVER` (previously `None`).

src/gt4py/next/iterator/transforms/infer_domain.py
tehrengruber: I introduced a typed dict `InferenceOptions` that handles forwarding the options in all places where they are not used.

src/gt4py/next/iterator/transforms/pass_manager.py
tehrengruber: I think this is a useful option for the domain inference and it naturally emerged after I addressed your review comments so I added it. I do however prefer working with an IR that has all domains inferred, which is the case after `InlineDynamicShifts` so I would leave it like that for now.

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -25,10 +25,29 @@
 from gt4py.next.utils import flatten_nested_tuple, tree_map
 
 
-DOMAIN: TypeAlias = domain_utils.SymbolicDomain | None | tuple["DOMAIN", ...]
+class DomainAccessDescriptor(eve.StrEnum):
+    """
+    Descriptor for domains that could not be inferred.
+    """
+
+    #: The access if unknown because of a dynamic shift.whose extent is not known.
```
havogt, line 33:
```suggestion
    #: The access is unknown because of a dynamic shift whose extent is not known.
```

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -57,12 +76,19 @@ def _split_dict_by_key(pred: Callable, d: dict):
...
-) -> domain_utils.SymbolicDomain | None:
-    filtered_domains: list[domain_utils.SymbolicDomain] = [d for d in domains if d is not None]
+def _domain_union(
+    *domains: domain_utils.SymbolicDomain | DomainAccessDescriptor,
+) -> domain_utils.SymbolicDomain | DomainAccessDescriptor:
+    if any(d == DomainAccessDescriptor.UNKNOWN for d in domains):
+        return DomainAccessDescriptor.UNKNOWN
+
+    filtered_domains: list[domain_utils.SymbolicDomain] = [
+        d  # type: ignore[misc]  # domain can never be unknown because as these cases are filtered above
```
havogt, line 86:
```suggestion
        d  # type: ignore[misc]  # domain can never be unknown as these cases are filtered

File: src/gt4py/next/iterator/transforms/infer_domain.py [hunk 13]
```diff
@@ -25,10 +25,29 @@
...
+
+    #: The access if unknown because of a dynamic shift.whose extent is not known.
+    #: E.g.: `(⇑(λ(arg0, arg1) → ·⟪Ioffₒ, ·arg1⟫(arg0)))(in_field1, in_field2)`
+    UNKNOWN = "unknown"
+    #: The domain is never accessed.
+    #: E.g.: `{in_field1, in_field2}[0]`
+    NEVER = "never"
+
+
+DOMAIN: TypeAlias = domain_utils.SymbolicDomain | DomainAccessDescriptor | tuple["DOMAIN", ...]
```
havogt, line 41: please add a comment for the `tuple[DOMAIN]` case.

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -108,34 +140,42 @@ def _merge_domains(
...
     stencil: itir.Expr,
     input_ids: list[str],
-    target_domain: domain_utils.SymbolicDomain,
+    target_domain: domain_utils.SymbolicDomain | DomainAccessDescriptor,
     offset_provider: common.OffsetProvider,
     symbolic_domain_sizes: Optional[dict[str, str]],
+    allow_uninferred: bool,
 ) -> ACCESSED_DOMAINS:
-    accessed_domains: dict[str, domain_utils.SymbolicDomain | None] = {}
+    accessed_domains: dict[str, domain_utils.SymbolicDomain | DomainAccessDescriptor] = {}
```
havogt, line 158:
```suggestion
    accessed_domains: ACCESSED_DOMAINS = {}
```

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -108,34 +140,42 @@ def _merge_domains(
...
             for shift in shifts_list
         ]
-        # `None` means field is never accessed
-        accessed_domains[in_field_id] = _domain_union_with_none(
-            accessed_domains.get(in_field_id, None), *new_domains
+        accessed_domains[in_field_id] = _domain_union(
+            accessed_domains.get(in_field_id, DomainAccessDescriptor.NEVER), *new_domains
         )
 
     return typing.cast(ACCESSED_DOMAINS, accessed_domains)
```
havogt, line 181:
```suggestion
    return accessed_domains
```

havogt: please double check typing. It's unclear to me what it should be. Especially if `tuple[DOMAIN]` is included or not in the different cases.

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -144,18 +184,22 @@ def _extract_accessed_domains(
...
-        raise ValueError("'target_domain' cannot be 'None'.")
+    if not allow_uninferred and target_domain is DomainAccessDescriptor.NEVER:
+        raise ValueError("'target_domain' cannot be 'NEVER' unless `allow_uninferred=True`.")
     # FIXME[#1582](tehrengruber): Temporary solution for `tuple_get` on scan result. See `test_solve_triag`.
     if isinstance(target_domain, tuple):
-        target_domain = _domain_union_with_none(*flatten_nested_tuple(target_domain))
-    if not isinstance(target_domain, domain_utils.SymbolicDomain):
-        raise ValueError("'target_domain' needs to be a 'domain_utils.SymbolicDomain'.")
+        target_domain = _domain_union(*flatten_nested_tuple(target_domain))  # type: ignore[arg-type]  # mypy not smart enough
+    if not isinstance(target_domain, (domain_utils.SymbolicDomain, DomainAccessDescriptor)):
```
havogt, line 199: What else can it be at this point?

src/gt4py/next/iterator/transforms/infer_domain.py
havogt: or introduce a new alias for the one that includes tuple

havogt, line 41:
```suggestion
_Domain: TypeAlias = domain_utils.SymbolicDomain | DomainAccessDescriptor | tuple["Domain", ...]
```

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -25,10 +25,29 @@
...
+    #: The access if unknown because of a dynamic shift.whose extent is not known.
+    #: E.g.: `(⇑(λ(arg0, arg1) → ·⟪Ioffₒ, ·arg1⟫(arg0)))(in_field1, in_field2)`
+    UNKNOWN = "unknown"
+    #: The domain is never accessed.
+    #: E.g.: `{in_field1, in_field2}[0]`
+    NEVER = "never"
+
+
+DOMAIN: TypeAlias = domain_utils.SymbolicDomain | DomainAccessDescriptor | tuple["DOMAIN", ...]
 ACCESSED_DOMAINS: TypeAlias = dict[str, DOMAIN]
```
havogt, line 42:
```suggestion
_AccessedDomains: TypeAlias = dict[str, _Domain]
```
etc

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -25,10 +25,29 @@
 from gt4py.next.utils import flatten_nested_tuple, tree_map
 
 
-DOMAIN: TypeAlias = domain_utils.SymbolicDomain | None | tuple["DOMAIN", ...]
+class DomainAccessDescriptor(eve.StrEnum):
```
havogt, line 28:
```suggestion
class _DomainAccessDescriptor(eve.StrEnum):
```
this is internal to the pass, correct?

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -362,30 +405,35 @@ def infer_expr(
     - domain: The domain `expr` is read at.
     - symbolic_domain_sizes: A dictionary mapping axes names, e.g., `I`, `Vertex`, to a symbol
       name that evaluates to the length of that axis.
+    - allow_uninferred: Allow `as_fieldop` expressions whose domain is either unknown (e.g.
+      because of a dynamic shift) or empty.
```
havogt, line 409: "empty" is used only here, probably use a formulation with "never accessed" here

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -25,10 +25,29 @@
 from gt4py.next.utils import flatten_nested_tuple, tree_map
 
 
-DOMAIN: TypeAlias = domain_utils.SymbolicDomain | None | tuple["DOMAIN", ...]
+class DomainAccessDescriptor(eve.StrEnum):
+    """
+    Descriptor for domains that could not be inferred.
```
havogt, line 30: Add a note that this is only used internally for checking, then replaced by `None` (if I understand correctly).

havogt (changes requested):
Mostly about clarifying types.

src/gt4py/next/iterator/transforms/infer_domain.py
tehrengruber: Removed the cast, since it is not needed anymore.

src/gt4py/next/iterator/transforms/infer_domain.py
tehrengruber: They can be used from the outside and this is also desired if you use `infer_expr` directly, so I left them public.

src/gt4py/next/iterator/transforms/infer_domain.py
tehrengruber: See above.

src/gt4py/next/iterator/transforms/infer_domain.py
tehrengruber: No, this can be used from the outside, it will for example appear in the `domain` annex attribute (used in `extract_tmp`) and could be used (I actually do this in the unused-return-from-scan fix I use in PMAP).

src/gt4py/next/iterator/transforms/infer_domain.py
tehrengruber: This is not replaced by None, it appears in the `domain` annex attribute, but is not placed in the IR.

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -25,8 +25,32 @@
...
+    NEVER = "never"
+
+
+NonTupleDomain: TypeAlias = domain_utils.SymbolicDomain | DomainAccessDescriptor
+#: The domain can also be a tuple of domains, usually this only occurs for scan operators returning
+#: a tuple since other occurrences for tuples are removed before domain inference. This is
+#: however not a requirement of the pass and `make_tuple(vertex_field, edge_field)` infers just
+#: fine to a tuple of a vertexn and an edge domain.
+Domain: TypeAlias = NonTupleDomain | tuple["Domain", ...]
+AccessedDomains: TypeAlias = dict[str, Domain]
```
havogt, line 47: Can we make the TypeAliases private? Especially to avoid confusion with this `Domain` and `common.Domain`.

src/gt4py/next/iterator/transforms/infer_domain.py
havogt: Do you need this information (the info what kind of "no domain" it is)? I don't really like that this ad-hoc concept leaks into other passes. Especially, since it is not clear how this Descriptor should be extended. If we call it `NoDomainDescriptor` at least it is clear there is no domain info.

src/gt4py/next/iterator/transforms/infer_domain.py
havogt: Where do you do that?

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -25,8 +25,32 @@
...
+    #: E.g.: `{in_field1, in_field2}[0]`
+    NEVER = "never"
+
+
+NonTupleDomain: TypeAlias = domain_utils.SymbolicDomain | DomainAccessDescriptor
+#: The domain can also be a tuple of domains, usually this only occurs for scan operators returning
+#: a tuple since other occurrences for tuples are removed before domain inference. This is
+#: however not a requirement of the pass and `make_tuple(vertex_field, edge_field)` infers just
+#: fine to a tuple of a vertexn and an edge domain.
+Domain: TypeAlias = NonTupleDomain | tuple["Domain", ...]
```
havogt, line 46: I don't like `Domain`  here as it might create confusion with `common.Domain`.

src/gt4py/next/iterator/transforms/infer_domain.py
tehrengruber: Here: https://github.com/GridTools/gt4py/pull/1738/files#diff-36c97ae1e7197fbdd81532e533a2e6294febc1c23bbb1a6ef2968130e15368fbR77
I added an annotation there so that we can reverse search.

src/gt4py/next/iterator/transforms/infer_domain.py
tehrengruber: TODO: write that DomainAccessDescriptor.KNOWN is missing, but skipped since we don't need it for technical reasons.

src/gt4py/next/iterator/transforms/infer_domain.py
tehrengruber: Renamed to `DomainAccess` as discussed.

havogt (approved):
lgtm

PR #1740, opened 2024-11-17
feat[next]: Upgrade dace dependency to v1.0.0

edopao:
DaCe version upgraded to `1.0.0`. It is also constrained to `< 1.1.0 `because the plan for DaCe v1.x is to introduce some breaking changes.

GPU tests still fail with GTIR DaCe backend (`test_double_use_scalar`) so they will be enabled in a separate PR.

Additional changes:
- Removed limitation on SymPy version since DaCe is now compatible with SymPy v1.13
- CUDA version upgraded from 11.2 to 11.4 to avoid this compile error in gpu build:
`dace/codegen/../runtime/include/dace/math.h(499): error: A __device__ variable cannot be marked constexpr`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,7 @@ all-cuda12 = ['gt4py[cuda12,dace,formatting,jax-cuda12,performance,testing]']
 # Other extras
 cuda11 = ['cupy-cuda11x>=12.0']
 cuda12 = ['cupy-cuda12x>=12.0']
-dace = ['dace>=1.0.0']
+dace = ['dace>=1.0.0,<=1.0.0']
 formatting = ['clang-format>=9.0']
 gpu = ['cupy>=12.0']
 jax-cpu = ['jax[cpu]>=0.4.18; python_version>="3.10"']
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -76,7 +76,7 @@ all-cuda12 = ['gt4py[cuda12,dace,formatting,jax-cuda12,performance,testing]']
 # Other extras
 cuda11 = ['cupy-cuda11x>=12.0']
 cuda12 = ['cupy-cuda12x>=12.0']
-dace = ['dace>=0.16.1', 'sympy>=1.9,<1.13']  # see https://github.com/spcl/dace/pull/1620
+dace = ['dace>=1.0.0,<=1.0.0']
```
havogt, line 79:
```suggestion
dace = ['dace==1.0.0']
```
but not constraining the upper version should be better, why did you add the upper bound?

edopao: Updated in latest commit with explanation.

havogt (approved):

PR #1741, opened 2024-11-18
feat[next]: Enable GPU tests on GTIR DaCe backend

edopao:
DaCe v1.0.0 allows to enable GPU tests on the GTIR backend.

An issue was found in `test_double_use_scalar`. The dace gpu transformations have a bug that produces invalid code for SDFGs containing scalar expressions outside the field operator. A workaround is to run the simplify pass in order to bring the SDFG to a canonical form.

The changes in test code (`test_execution.py`) are pure cleanup.


philip-paul-mueller (approved):
LGTM

PR #1742, opened 2024-11-19
feat[next]: SDFGConvertible Program for dace_fieldview backend

DropD:
## Description

Add a `decrator.Program` subclass, which implements `SDFGConvertible` to `dace_fieldview` backend, analogous to the one in `dace_iterator`. Conditionally shadow `decorator.Program` with it and reactivate the orchestration tests by using `dace_fieldview` instead of `dace_iterator`. 

One caveat: The toolchain is not ready for pure `CompileTimeConnectivities` in all cases yet, so the `test_sdfgConvertible_connectivities` had to be adjusted for the moment.
## Requirements

- [x] The PR contains an updated version of the `AUTHORS.md` file adding the names of all the new contributors.


src/gt4py/next/program_processors/runners/dace_fieldview/program.py
```diff
@@ -0,0 +1,150 @@
...
+            self.backend.allocator, core_defs.DeviceType.CPU
+        ) and not allocators.is_field_allocator_for(
+            self.backend.allocator, core_defs.DeviceType.CPU_PINNED
+        )
+
+        sdfg = self.backend.executor.step.translation.generate_sdfg(  # type: ignore[attr-defined] # we can assume to get a DaCeTranslationStep here
+            gtir_stage.data,
+            offset_provider=gtir_stage.args.offset_provider,
+            column_axis=kwargs.get("column_axis", gtir_stage.args.column_axis),
+            auto_opt=True,
```
edopao, line 65:
```suggestion
            auto_opt=False,
```

I think we would like to run auto-optimization after the SDFGs are fused into one big SDFG.

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
```diff
@@ -0,0 +1,150 @@
...
+        on_gpu = not allocators.is_field_allocator_for(
+            self.backend.allocator, core_defs.DeviceType.CPU
+        ) and not allocators.is_field_allocator_for(
+            self.backend.allocator, core_defs.DeviceType.CPU_PINNED
+        )
+
+        sdfg = self.backend.executor.step.translation.generate_sdfg(  # type: ignore[attr-defined] # we can assume to get a DaCeTranslationStep here
+            gtir_stage.data,
+            offset_provider=gtir_stage.args.offset_provider,
+            column_axis=kwargs.get("column_axis", gtir_stage.args.column_axis),
```
edopao, line 64:
```suggestion
            column_axis=gtir_stage.args.column_axis,
```

I guess at this point we should just use `gtir_stage.args.column_axis` because we passed the keyword argument value when constructing the `gtir_stage`, see line 49..

src/gt4py/next/ffront/decorator.py
```diff
@@ -310,7 +310,7 @@ def __call__(
 
 
 try:
-    from gt4py.next.program_processors.runners.dace_iterator import Program
+    from gt4py.next.program_processors.runners.dace_fieldview.program import Program
```
edopao, line 313: Since `dace_iterator.Program` is no longer used nor tested, you could delete that code.

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
```diff
@@ -0,0 +1,150 @@
...
+                        "arrays"
+                    ][conn_id].storage
+                    break
+
+            # Build the closure dictionary
+            for name, conn_id in in_arrays_with_id:
+                if conn_id not in self.connectivity_tables_data_descriptors:
+                    conn = self.connectivities[name]
+                    self.connectivity_tables_data_descriptors[conn_id] = dace.data.Array(
+                        dtype=dace.int64 if conn.index_type == np.int64 else dace.int32,
```
edopao, line 134: I think that dace `dtypes` has a utility to parse numpy types:
`dace.dtypes.typeclass(conn.index_type)`

src/gt4py/next/ffront/decorator.py
DropD: 👍

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
DropD: and trying `dace.dtypes.typeclass(conn.index_type.type)` leads to wrong stencil results.

DropD: edit: after merging main, it passes. However according to type hints of `NeighborTableOffsetProvider`, your version should work and mine should fail. Yet, in the tests, the connectivities are set up in a way so that my version works and yours fails. The current version on the other hand is safe in both cases.

If I can get the test to produce the correct types so the official version works, I will change it. Otherwise I would prefer to leave this one until the neighbor table construction is anyway changed by what @havogt is working on.

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
DropD: I reverted the change because it made the tests fail.

edopao: I want to look into this because it is not expected.

DropD: For some reason it passes again, i might have changed something else at the same time accidentally?

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
```diff
@@ -0,0 +1,150 @@
...
+                args=arguments.CompileTimeArgs(
+                    args=tuple(p.type for p in self.past_stage.past_node.params),
+                    kwargs={},
+                    column_axis=column_axis,
+                    offset_provider=offset_provider,
+                ),
+            )
+        )
+
+        on_gpu = not allocators.is_field_allocator_for(
```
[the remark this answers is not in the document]
DropD: We could, if we do not support the other gpu device types (and don't plan to).

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -155,214 +147,3 @@ def build_sdfg_from_itir(
...
-            for closure in self.itir.closures
-            for in_field in closure.inputs
-        )  # backend only supports SymRef inputs, not `index` calls
-        input_fields = [
-            str(in_field.id)  # type: ignore[union-attr]  # ensured by assert
-            for closure in self.itir.closures
-            for in_field in closure.inputs
-            if str(in_field.id) in fields  # type: ignore[union-attr]  # ensured by assert
-        ]
-        sdfg.gt4py_program_input_fields = {
```
kotsaloscv, line 215: Could you please port this dynamic property to GTIR as well.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -155,214 +147,3 @@ def build_sdfg_from_itir(
...
-        for closure in self.itir.closures:
-            output = closure.output
-            if isinstance(output, itir.SymRef):
-                if str(output.id) in fields:
-                    output_fields.append(str(output.id))
-            else:
-                for arg in output.args:
-                    if str(arg.id) in fields:  # type: ignore[attr-defined]
-                        output_fields.append(str(arg.id))  # type: ignore[attr-defined]
-        sdfg.gt4py_program_output_fields = {
```
kotsaloscv, line 232: Same for this property.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -155,214 +147,3 @@ def build_sdfg_from_itir(
...
-            )
-            for param, shifts in zip(closure.inputs, params_shifts):
-                assert isinstance(
-                    param, SymRef
-                )  # backend only supports SymRef inputs, not `index` calls
-                if not isinstance(param.id, str):
-                    continue
-                if param.id not in sdfg.gt4py_program_input_fields:
-                    continue
-                sdfg.offset_providers_per_input_field.setdefault(param.id, []).extend(list(shifts))
```
kotsaloscv, line 256: Same here.

All these properties are needed for the automatic halo exchange placement.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -155,214 +147,3 @@ def build_sdfg_from_itir(
...
-        return closure_dict
-
-    def __sdfg_signature__(self) -> tuple[Sequence[str], Sequence[str]]:
-        args = []
-        for arg in self.past_stage.past_node.params:
-            args.append(arg.id)
-        return (args, [])
-
-
-def _crosscheck_dace_parsing(dace_parsed_args: list[Any], gt4py_program_args: list[Any]) -> bool:
```
kotsaloscv, line 341: This function also needs to be ported to GTIR as well.

kotsaloscv (changes requested):

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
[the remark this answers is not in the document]
edopao: This should be more generic:
`on_gpu = allocators.is_field_allocator_for(self.backed_allocator, allocators.CUPY)`

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
```diff
@@ -0,0 +1,148 @@
...
+                    offset_provider=offset_provider,
+                ),
+            )
+        )
+
+        on_gpu = allocators.is_field_allocator_for(
+            self.backend.allocator, core_defs.DeviceType.CUDA
+        )
+
+        sdfg = self.backend.executor.step.translation.generate_sdfg(  # type: ignore[attr-defined] # we can assume to get a DaCeTranslationStep here
```
edopao, line 59: We should replace this code with the following, in order to avoid duplication of parameters such as `offset_provider`, `column_axis`, etc.:
```
    program_source = self.backend.executor.step.translation(gtir_stage)
    sdfg: dace.SDFG = dace.SDFG.from_json(program_source.source_code)
```

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
DropD: Working on it, Program nodes will require a different approach.

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
```diff
@@ -0,0 +1,251 @@
...
+        )  # We know which backend we are using, but we don't know if the compile workflow is cached.
+        sdfg = dace.SDFG.from_json(compile_workflow.translation(gtir_stage).source_code)
+
+        self.sdfg_closure_cache["arrays"] = sdfg.arrays
+
+        # Halo exchange related metadata, i.e. gt4py_program_input_fields, gt4py_program_output_fields, offset_providers_per_input_field
+        # Add them as dynamic properties to the SDFG
+        program = typing.cast(
+            itir.Program, gtir_stage.data
+        )  # we already checked that our backend uses GTIR
```
edopao, line 71: > we already checked that our backend uses GTIR

where? otherwise we could make an `assert isinstance()`.

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
```diff
@@ -0,0 +1,251 @@
...
+        field_params = {
+            str(param.id): param for param in program.params if isinstance(param.type, ts.FieldType)
+        }
+
+        input_fields = (field_params[name] for name in InputNamesExtractor.only_fields(program))
+        # TODO (ricoh): This will associate the last horizontal dimension of each field with it's name
```
edopao, line 77: My suspect is that this works under the assumption that all fields have a single horizontal dimension.
@kotsaloscv If this is the case, should we make a check and raise an exception if this assumption is not met?

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
```diff
@@ -155,214 +147,3 @@ def build_sdfg_from_itir(
...
-                    output_fields.append(str(output.id))
-            else:
-                for arg in output.args:
-                    if str(arg.id) in fields:  # type: ignore[attr-defined]
-                        output_fields.append(str(arg.id))  # type: ignore[attr-defined]
-        sdfg.gt4py_program_output_fields = {
-            output: dim
-            for output in output_fields
-            for dim in fields[output].dims  # type: ignore[union-attr]
-            if dim.kind == common.DimensionKind.HORIZONTAL
```
DropD, line 236: Unless I am very much mistaken, this overwrites the value stored in "output" in a loop "for dim in fiels[output].dims if ...". So you end up with a dict containing `{<output_field_name>: <last horizontal dim in field type>}`. Is this intended? Not knowing how it is used later on, I would have expected a collection of horizontal dims per output field name.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
DropD: Results from discussion in a separate channel:
- it is correct, as only used in `icon4py` which uses "unstructured" fields. There is max one horizontal dimension per field.
- trace_shifts should work™️ partially on GTIR (but not accross `as_fieldop`).
- currently not actively used in `icon4py`, so defer implementation and put a TODO instead.

src/gt4py/next/program_processors/runners/dace_iterator/__init__.py
DropD: Results from discussion in a separate channel for "offset_providers_per_input_field":
- it is correct, as only used in `icon4py` which uses "unstructured" fields. There is max one horizontal dimension per field.
- trace_shifts should work™️ partially on GTIR (but not accross `as_fieldop`).
- currently not actively used in `icon4py`, so defer implementation and put a TODO instead.

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
DropD: We checked that `self.backend` is a DaCe backend. Probably should check that it is specifically the `dace_fieldview` backend.

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
DropD: We had discussed this and I refactored it to explicitly only add fields with a single horizontal dimension.

src/gt4py/next/program_processors/runners/dace_fieldview/program.py
```diff
@@ -0,0 +1,298 @@
...
+
+    @classmethod
+    def only_fields(cls, program: itir.Program) -> set[str]:
+        field_param_names = [
+            str(param.id) for param in program.params if isinstance(param.type, ts.FieldType)
+        ]
+        return {name for name in cls().visit(program) if name in field_param_names}
+
+
+class InputNamesExtractor(SymbolNameSetExtractor):
```
havogt, line 211: Could you move these utilities to some module in `iterator/transforms`. I would like to use them in other places.

edopao:
If it can help, I got the tests to pass locally with this diff:
```
diff --git a/src/gt4py/next/program_processors/runners/dace_fieldview/program.py b/src/gt4py/next/program_processors/runners/dace_fieldview/program.py
index 803ae866..d65f4fac 100644
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/program.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/program.py
@@ -176,7 +176,7 @@ class Program(decorator.Program, dace.frontend.python.common.SDFGConvertible):
                     conn = self.connectivities[name]
                     assert common.is_neighbor_table(conn)
                     self.connectivity_tables_data_descriptors[conn_id] = dace.data.Array(
-                        dtype=dace.int64 if conn.dtype == np.int64 else dace.int32,
+                        dtype=dace.dtypes.typeclass(conn.dtype.scalar_type),
                         shape=[
                             symbols[dace_utils.field_size_symbol_name(conn_id, 0)],
                             symbols[dace_utils.field_size_symbol_name(conn_id, 1)],
diff --git a/tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py b/tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py
index 9bde9aca..80cd9eac 100644
--- a/tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py
+++ b/tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py
@@ -133,7 +133,7 @@ def test_sdfgConvertible_connectivities(unstructured_case):
         data=xp.asarray([[0, 1], [1, 2], [2, 0]]),
         allocator=allocator,
     )
-    connectivities = {"E2V": e2v.__gt_type__()}
+    connectivities = {"E2V": e2v}
     offset_provider = OffsetProvider_t.dtype._typeclass.as_ctypes()(E2V=e2v.data_ptr())
 
     SDFG = sdfg.to_sdfg(connectivities=connectivities)
```

DropD:
> If it can help, I got the tests to pass locally with this diff:
> 
> ```
> diff --git a/src/gt4py/next/program_processors/runners/dace_fieldview/program.py b/src/gt4py/next/program_processors/runners/dace_fieldview/program.py
> index 803ae866..d65f4fac 100644
> --- a/src/gt4py/next/program_processors/runners/dace_fieldview/program.py
> +++ b/src/gt4py/next/program_processors/runners/dace_fieldview/program.py
> @@ -176,7 +176,7 @@ class Program(decorator.Program, dace.frontend.python.common.SDFGConvertible):
>                      conn = self.connectivities[name]
>                      assert common.is_neighbor_table(conn)
>                      self.connectivity_tables_data_descriptors[conn_id] = dace.data.Array(
> -                        dtype=dace.int64 if conn.dtype == np.int64 else dace.int32,
> +                        dtype=dace.dtypes.typeclass(conn.dtype.scalar_type),
>                          shape=[
>                              symbols[dace_utils.field_size_symbol_name(conn_id, 0)],
>                              symbols[dace_utils.field_size_symbol_name(conn_id, 1)],
> diff --git a/tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py b/tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py
> index 9bde9aca..80cd9eac 100644
> --- a/tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py
> +++ b/tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py
> @@ -133,7 +133,7 @@ def test_sdfgConvertible_connectivities(unstructured_case):
>          data=xp.asarray([[0, 1], [1, 2], [2, 0]]),
>          allocator=allocator,
>      )
> -    connectivities = {"E2V": e2v.__gt_type__()}
> +    connectivities = {"E2V": e2v}
>      offset_provider = OffsetProvider_t.dtype._typeclass.as_ctypes()(E2V=e2v.data_ptr())
>  
>      SDFG = sdfg.to_sdfg(connectivities=connectivities)
> ```

This is part of what I ended up doing but I wanted to honor the original purpose of the test, which was to compile the SDFG without knowledge of the connectivity tables and then call it with two different ones.

src/gt4py/next/iterator/transforms/extractors.py
```diff
@@ -0,0 +1,87 @@
...
+        field_param_names = [
+            str(param.id) for param in program.params if isinstance(param.type, ts.FieldType)
+        ]
+        return {name for name in cls().visit(program) if name in field_param_names}
+
+
+class InputNamesExtractor(SymbolNameSetExtractor):
+    """Extract the set of symbol names passed into field operators within a program."""
+
+    def visit_Program(self, node: itir.Program) -> set[str]:
```
edopao, line 37: `visit_Program`, `visit_IfStmt`, `visit_Temporary` and `visit_SymRef` are common to input and output visitors. I would move them to the common base class, but please do as you prefer.

tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py
```diff
@@ -37,16 +37,13 @@
...
         pytest.skip("DaCe-related test: Test SDFGConvertible interface for GT4Py programs")
 
-    # TODO(ricoh): enable test after adding GTIR support
-    pytest.skip("DaCe SDFGConvertible interface does not support GTIR program.")
+    allocator, backend = cartesian_case.allocator, cartesian_case.backend
 
-    allocator, backend = unstructured_case.allocator, unstructured_case.backend
-
-    if gtx_allocators.is_field_allocator_factory_for(allocator, gtx_allocators.CUPY_DEVICE):
+    if gtx_allocators.is_field_allocator_for(allocator, gtx_allocators.CUPY_DEVICE):
```
edopao, line 46: I suggest the following cleanup.
```
-    allocator, backend = cartesian_case.allocator, cartesian_case.backend
-
-    if gtx_allocators.is_field_allocator_for(allocator, gtx_allocators.CUPY_DEVICE):
-        import cupy as xp
-    else:
-        import numpy as xp
-
     in_field = cases.allocate(cartesian_case, laplap_program, "in_field")()
     out_field = cases.allocate(cartesian_case, laplap_program, "out_field")()
 
+    xp = in_field.array_ns
+    backend = cartesian_case.backend
+
```

tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py
```diff
@@ -113,48 +107,51 @@ def test_sdfgConvertible_connectivities(unstructured_case):
         name="OffsetProvider",
     )
 
+    e2v = gtx.as_connectivity(
+        [Edge, E2VDim],
+        codomain=Vertex,
+        data=xp.asarray([[0, 1], [1, 2], [2, 0]]),
+        allocator=allocator,
+    )
+    e2v_ndarray_copy = (
```
edopao, line 116: My understanding is that `e2v_ndarray_copy` can now be removed.

edopao (commented):
Some suggestions for cleanup, otherwise it looks good to me.

kotsaloscv (approved):
LGTM, although we need to additionally check it with ICON4Py.

PR #1744, opened 2024-11-19
fix[next]: change default cache folder in the config

egparedes:
## Requirements

### DIFF
--- a/src/gt4py/next/config.py
+++ b/src/gt4py/next/config.py
@@ -11,7 +11,6 @@ from __future__ import annotations
 import enum
 import os
 import pathlib
-import tempfile
 from typing import Final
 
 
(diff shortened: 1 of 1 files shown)


tehrengruber (commented):
I was always fine with the existing behavior as every project needs to configure this anyway and it shipped around still existing caching issues. Just a couple of notes / observations I thought are worth mentioning:
- `DEBUG=True` now means we will easily use out-dated cache entries indefinitely since the cache is not cleared anymore on reboot like in `tmp` and our cache key does not contain e.g. the GT4Py version.
- The caching behavior is now path dependent, if you run a driver script from a different path you will recompile completely.

havogt:
>     * `DEBUG=True` now means we will easily use out-dated cache entries indefinitely since the cache is not cleared anymore on reboot like in `tmp` and our cache key does not contain e.g. the GT4Py version.

I see this as an improvement over the previous behavior. Now it is consistent, before it was depending on the stability of your computer.

havogt (approved):

egparedes:
GT4Py cartesian, and I  believe also DaCe, have always used the current directory as the root for the cache folder, so `gt4py.next` using the global temp was actually a compatibility-breaking behavior in some way. cPython has a slightly different behavior (it places the bytecode cache folder where the sources are defined) but also doesn't use the global temp.

We should definitely open an issue about gt4py not using its version (or even the python version) for the caching key.

tehrengruber (approved):

PR #1745, opened 2024-11-21
fix[next]: Guard diskcache creation by file lock

tehrengruber:
The disk cache used to cache compilation in the gtfn backend has a race condition manifesting itself in `sqlite3.OperationalError: database is locked` errors if multiple python processes try to initialize the `diskcache.Cache` object concurrently. This PR fixes this by guarding the object creation by a file-based lock in the same directory as the database.

While this issue occurred frequently and was observed to be fixed on distributed file systems, the lock does not guarantee correct behavior in particular for accesses to the cache (beyond opening) since the underlying SQLite database is unreliable when stored on an NFS based file system. It does however ensure correctness of concurrent cache accesses on a local file system. See more information here:

https://grantjenks.com/docs/diskcache/tutorial.html#settings
https://www.sqlite.org/faq.html#q5
https://github.com/tox-dev/filelock/issues/73
NFS safe locking:
https://gitlab.com/warsaw/flufl.lock
[Barry Warsaw / FLUFL Lock · GitLab](https://gitlab.com/warsaw/flufl.lock)


src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -141,11 +144,24 @@ def fingerprint_compilable_program(inp: stages.CompilableProgram) -> str:
 
 class FileCache(diskcache.Cache):
     """
-    This class extends `diskcache.Cache` to ensure the cache is closed upon deletion,
-    i.e. it ensures that any resources associated with the cache are properly
-    released when the instance is garbage collected.
+    This class extends `diskcache.Cache` to ensure the cache is properly
+    - opened on a distributed file system by using a file lock. This guards the creating of the
```
havogt, line 148: Is writing not a problem? Except that it's much harder to make all processes write at the same time and there less likely to trigger.

tehrengruber: No, writing to the cache is actually fine. SQLite which is used internally by `diskcache.Cache` handles concurrency very well in most cases, except for changing settings of the database itself, and that is the problem here. On creation of the `diskcache.Cache` object, some sql `PRAGMA` statements are executed and they lead to the observed issues. It is not really clear why, this should work, but it doesn't.

egparedes (approved):
LGTM

PR #1750, opened 2024-11-26
refactor[next]: use eve.datamodel for types

havogt:


src/gt4py/next/ffront/past_passes/type_deduction.py
```diff
@@ -108,7 +108,7 @@ def visit_Subscript(self, node: past.Subscript, **kwargs: Any) -> past.Subscript
         value = self.visit(node.value, **kwargs)
         return past.Subscript(
             value=value,
-            slice_=self.visit(node.slice_, **kwargs),
+            slice_=node.slice_,  # TODO: I don't think we are using this type
```
havogt, line 111:
```suggestion
            slice_=node.slice_,
```
check and remove

src/gt4py/next/type_system/type_specifications.py
```diff
@@ -92,31 +79,37 @@ def __str__(self) -> str:
...
+    def __str__(self) -> str:
+        dims = "..." if self.dims is Ellipsis else f"[{', '.join(dim.value for dim in self.dims)}]"
+        return f"Field[{dims}, {self.dtype}]"
+
+
 class TupleType(DataType):
-    types: list[DataType]
+    # TODO: or DimensionType is a DataType
+    types: list[DataType | DimensionType | DeferredType]
+    # TODO validate DeferredType constraints
```
havogt, line 101: check this

src/gt4py/next/type_system/type_specifications.py
```diff
@@ -92,31 +79,37 @@ def __str__(self) -> str:
         return f"{kind_str}{self.shape}"
 
 
-@dataclass(frozen=True)
+class ListType(DataType):
+    element_type: DataType
+    # TODO(havogt): the `offset_type` is not yet used in type_inference,
+    # it is meant to describe the neighborhood (via the local dimension)
+    offset_type: Optional[common.Dimension] = None
```
tehrengruber, line 86: Why has this type been moved here / promoted to something shared between FOAST and ITIR type system?

src/gt4py/next/ffront/past_passes/type_deduction.py
tehrengruber: We don't have a type for `past.Slice`, but the visit here also types & validates its children (e.g `1+1.` in a slice bound would error here). For that reason I consider unconditionally visiting all children a good approach. The role of `ts.DeferredType` and `None` in the type system is however not well specified. I would postpone that discussion to the point where we also have time to think about generics / "templated" / partial types, because the discussion is related.

src/gt4py/next/iterator/type_system/inference.py
```diff
@@ -272,7 +272,7 @@ def _get_dimensions(obj: Any):
         if isinstance(obj, common.Dimension):
             yield obj
         elif isinstance(obj, ts.TypeSpec):
-            for field in dataclasses.fields(obj.__class__):
```
tehrengruber, line 275: I can only express my sincere apologies for this abomination of a function...

src/gt4py/next/type_system/type_specifications.py
```diff
@@ -92,31 +79,37 @@ def __str__(self) -> str:
...
+
+    def __str__(self) -> str:
+        dims = "..." if self.dims is Ellipsis else f"[{', '.join(dim.value for dim in self.dims)}]"
+        return f"Field[{dims}, {self.dtype}]"
+
+
 class TupleType(DataType):
-    types: list[DataType]
+    # TODO: or DimensionType is a DataType
+    types: list[DataType | DimensionType | DeferredType]
```
tehrengruber, line 100:
```suggestion
    # TODO(tehrengruber): Remove `DeferredType` again. This was erroneously
    #  introduced before we checked the annotations at runtime. All attributes of 
    #  a type that are types themselves must be concrete.
    types: list[DataType | DimensionType | DeferredType]
```
The `DataType | DimensionType` is somewhat ugly, what is actually meant here is `TypeSpec` that is not `DeferredType`, i.e. a concrete type (see `is_concrete`). Promoting `is_concrete` to something that could be expressed in an annotation, e.g. by inheriting from a new class or mixin `ConcreteType`, would be an idea.

src/gt4py/next/type_system/type_specifications.py
```diff
@@ -92,31 +79,37 @@ def __str__(self) -> str:
...
+class ListType(DataType):
+    element_type: DataType
+    # TODO(havogt): the `offset_type` is not yet used in type_inference,
+    # it is meant to describe the neighborhood (via the local dimension)
+    offset_type: Optional[common.Dimension] = None
+
+
+class FieldType(DataType, CallableType):
+    dims: list[common.Dimension]
+    dtype: ScalarType | ListType
```
tehrengruber, line 91: Unrelated to this PR which just exposes this, but this looks inconsistent. Our conclusion with respect to list types was that the frontend has the local dimension in `dims` while ITIR uses `ListType`. Consequently frontend and ITIR have different concepts of what a Field is should therefore also use different `FieldType`s.

src/gt4py/next/ffront/past_passes/type_deduction.py
havogt: What are you proposing to do? Accept `None` in TupleType elements, this proposed solution or add `visit_Slice` and return `DeferredType`.

src/gt4py/next/type_system/type_specifications.py
havogt: moved here because it's a possible dtype of FieldType, see your other comment...

src/gt4py/next/type_system/type_specifications.py
havogt: I would propose to document this here, because ultimately we should have one FieldType so refactoring in two different is probably quite some work that will disappear later.

src/gt4py/next/ffront/past_passes/type_deduction.py
tehrengruber: Ah I overlooked that `slice_` is a tuple of slices (maybe the attr name is not so great?). `visit_Slice` with `DeferredType` makes sense to me even though this would add another violation of all-type-attrs-of-types-are-concrete rule of a `ts.TypeSpec`. Otherwise introduction of a slice type makes sense to me too.

src/gt4py/next/type_system/type_specifications.py
tehrengruber: Makes sense, maybe a comment that this is not used in the frontend and only here because it can be in a field makes sense.

src/gt4py/next/type_system/type_specifications.py
tehrengruber: Makes sense.

egparedes (commented):
LGTM. Just a couple of questions.

src/gt4py/next/type_system/type_info.py
```diff
@@ -78,15 +78,15 @@ def type_class(symbol_type: ts.TypeSpec) -> Type[ts.TypeSpec]:
...
-    match symbol_type:
-        case ts.DeferredType(constraint):
-            if constraint is None:
-                raise ValueError(f"No type information available for '{symbol_type}'.")
-            elif isinstance(constraint, tuple):
-                raise ValueError(f"Not sufficient type information available for '{symbol_type}'.")
-            return constraint
-        case ts.TypeSpec() as concrete_type:
-            return concrete_type.__class__
+    if isinstance(symbol_type, ts.DeferredType):
```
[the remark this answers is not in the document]
havogt: as discussed, datamodels don't implement `__match_args__`

src/gt4py/next/ffront/past_passes/type_deduction.py
havogt: done

src/gt4py/next/type_system/type_specifications.py
```diff
@@ -6,21 +6,13 @@
...
 
 
-@dataclass(frozen=True)
-class TypeSpec:
-    def __hash__(self) -> int:
-        return hash(content_hash(self))
-
-    def __init_subclass__(cls) -> None:
-        cls.__hash__ = TypeSpec.__hash__  # type: ignore[method-assign]
+class TypeSpec(eve_datamodels.DataModel, kw_only=False, frozen=True): ...  # type: ignore[call-arg]
```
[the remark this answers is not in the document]
havogt: Since there is no issue (in our tests), I assume it's fine.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -290,6 +290,7 @@ def _add_storage(
                 # represent zero-dimensional fields as scalar arguments
                 return self._add_storage(sdfg, symbolic_arguments, name, gt_type.dtype, transient)
             # handle default case: field with one or more dimensions
+            assert isinstance(gt_type.dtype, ts.ScalarType)
```
tehrengruber, line 293:
```suggestion
            # list dtype can not occur since ...
            assert isinstance(gt_type.dtype, ts.ScalarType)
```

src/gt4py/next/type_system/type_specifications.py
```diff
@@ -92,31 +79,43 @@ def __str__(self) -> str:
...
-    types: list[DataType]
-
-    def __str__(self) -> str:
-        return f"tuple[{', '.join(map(str, self.types))}]"
+class ListType(DataType):
+    """Represents a neighbor list in the ITIR representation.
 
-    def __iter__(self) -> Iterator[DataType]:
-        yield from self.types
+    Note: not used in the frontend.
```
tehrengruber, line 85:
```suggestion
    Note: not used in the frontend. Neighbor fields are represented by local dimensions instead.
```
There was a good reason we put the type here instead of `itir_ts`, but I forgot again. Can we also add comment for the reasoning here?

tehrengruber (approved):
Just a couple of requests to add comments.

src/gt4py/next/otf/binding/nanobind.py
```diff
@@ -86,6 +86,7 @@ def _type_string(type_: ts.TypeSpec) -> str:
         return f"std::tuple<{','.join(_type_string(t) for t in type_.types)}>"
     elif isinstance(type_, ts.FieldType):
         ndims = len(type_.dims)
+        assert isinstance(type_.dtype, ts.ScalarType)
```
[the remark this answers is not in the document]
havogt: at this point the local dimension is just a local dimension, later they are transformed to field of tuple.

src/gt4py/next/type_system/type_specifications.py
havogt: Yes, it appears in FieldType which is the next type below.

PR #1751, opened 2024-11-27
feat[next][dace]: Lowering to SDFG of index builtin

edopao:
Implements the lowering to SDFG of the GTIR index builtin.


philip-paul-mueller (approved):
LGTM

PR #1756, opened 2024-11-28
bug[next]: ConstantFolding after create_global_tmps

SF-N:
`create_global_tmps` calls domain inference which creates domain unions, i.e., there were many nested minima and maxima.
Now `ConstantFolding` is done within `domain_union` to avoid this.

### DIFF
--- a/src/gt4py/next/iterator/ir_utils/domain_utils.py
+++ b/src/gt4py/next/iterator/ir_utils/domain_utils.py
@@ -16,6 +16,7 @@ from gt4py.next import common
 from gt4py.next.iterator import ir as itir
 from gt4py.next.iterator.ir_utils import ir_makers as im
 from gt4py.next.iterator.transforms import trace_shifts
+from gt4py.next.iterator.transforms.constant_folding import ConstantFolding
 
 
 def _max_domain_sizes_by_location_type(offset_provider: Mapping[str, Any]) -> dict[str, int]:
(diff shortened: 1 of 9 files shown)


src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -418,10 +419,14 @@ def infer_program(
...
+        itir.Program(
+            id=program.id,
+            function_definitions=program.function_definitions,
+            params=program.params,
+            declarations=program.declarations,
+            body=[
+                _infer_stmt(stmt, offset_provider, symbolic_domain_sizes) for stmt in program.body
+            ],
+        )
     )
```
tehrengruber, line 432:
```suggestion
    inferred_program = itir.Program(
        id=program.id,
        function_definitions=program.function_definitions,
        params=program.params,
        declarations=program.declarations,
        body=[
            _infer_stmt(stmt, offset_provider, symbolic_domain_sizes) for stmt in program.body
        ],
    )
    # simplify domain expressions again
    return ConstantFolding.apply(inferred_program)  # type: ignore[return-value]  # returns same type as input
```

tehrengruber (approved):

tehrengruber (changes requested):
Have to revert my approval as this likely doesn't work because the domain is stored in the annex which will not be constant folded.

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -168,6 +169,8 @@ def domain_union(*domains: SymbolicDomain) -> SymbolicDomain:
             lambda current_expr, el_expr: im.call("maximum")(current_expr, el_expr),
             [domain.ranges[dim].stop for domain in domains],
         )
+        # constant fold expression to keep the tree small
+        start, stop = ConstantFolding.apply(start), ConstantFolding.apply(stop)  # type: ignore[assignment]  # always an itir.FunCall
```
tehrengruber, line 173:
```suggestion
        start, stop = ConstantFolding.apply(start), ConstantFolding.apply(stop)  # type: ignore[assignment]  # always an itir.Expr
```
e.g. for `minimum(0, 0)` -> `0` which is an `itir.Literal`

tehrengruber (approved):

PR #1758, opened 2024-11-29
refact[next][dace]: Initialize GTIRToSDFG data class with SDFG

edopao:
Add SDFG attribute to the GTIRToSDFG data class, which implies adding a `get_sdfg(self)` method on the SDFGBuilder protocol. This refactoring allows to avoid passing around the SDFG being built by the visitor.


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -100,22 +100,40 @@ class SDFGBuilder(DataflowBuilder, Protocol):
...
+    ) -> Tuple[str, dace.dtypes.Array]:
+        """Wrapper of `dace.SDFG.add_temp_transient()`."""
+        return self.get_sdfg().add_temp_transient(shape, dtype, **kwargs)
+
+    def add_temp_scalar(
+        self, dtype: dace.dtypes.typeclass, name: Optional[str] = None, **kwargs: Any
+    ) -> Tuple[str, dace.dtypes.Scalar]:
+        """Helper function analogue of `dace.SDFG.add_temp_transient()`, but for scalars."""
+        name = name or self.get_sdfg().temp_data_name()
+        return self.get_sdfg().add_scalar(name, dtype, **kwargs)
```
philip-paul-mueller, line 135: Currently you can not do that, but I think it should be possible to write:
```python
sdfg.add_scalar(name=None, ...)
```
and this implies `find_new_name=True`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -196,12 +232,11 @@ def make_field(
...
-
-    def unique_nsdfg_name(self, sdfg: dace.SDFG, prefix: str) -> str:
+    def unique_nsdfg_name(self, prefix: str) -> str:
         nsdfg_list = [
-            nsdfg.label for nsdfg in sdfg.all_sdfgs_recursive() if nsdfg.label.startswith(prefix)
+            nsdfg.label
+            for nsdfg in self.sdfg.all_sdfgs_recursive()
+            if nsdfg.label.startswith(prefix)
         ]
         return f"{prefix}_{len(nsdfg_list)}"
```
philip-paul-mueller, line 241: I do not really like that the get-unique-name-for-a-nested-SDFG step and the add-nested-sdfg are separated.
Is there a way to make this in one step?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -413,73 +448,35 @@ def _add_sdfg_params(
...
-        First, it will allocate field and scalar storage for global data. The storage
-        represents global data, available everywhere in the SDFG, either containing
-        external data (aka non-transient data) or temporary data (aka transient data).
-        The temporary data is global, therefore available everywhere in the SDFG
-        but not outside. Then, all statements are translated, one after the other.
-        """
-        if node.function_definitions:
-            raise NotImplementedError("Functions expected to be inlined as lambda calls.")
+    def visit_Program(self, node: gtir.Program) -> None:
+        """Translates the body of `ir.Program` inside the SDFG that belongs to this builder."""
```
philip-paul-mueller, line 458: OPTIONAL: the description sounds a bit strange, how does:
Translates the program contained in `gtir.Program` into an SDFG.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -413,73 +448,35 @@ def _add_sdfg_params(
...
-        # DaCe requires C-compatible strings for the names of data containers,
-        # such as arrays and scalars. GT4Py uses a unicode symbols ('ᐞ') as name
-        # separator in the SSA pass, which generates invalid symbols for DaCe.
-        # Here we find new names for invalid symbols present in the IR.
-        node = dace_gtir_utils.replace_invalid_symbols(sdfg, node)
-
-        # start block of the stateful graph
-        entry_state = sdfg.add_state("program_entry", is_start_block=True)
+        # create start block of the stateful graph
+        assert len(self.sdfg.nodes()) == 0
```
philip-paul-mueller, line 465: I would make this a hard test not just an assert, or at least add a description.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -671,30 +654,25 @@ def get_field_domain_offset(
...
-            nsdfg, node_params=lambda_params, symbolic_arguments=lambda_domain_symbols
-        )
-
-        lambda_result = lambda_translator.visit(
-            node.expr,
-            sdfg=nsdfg,
-            head_state=nstate,
+        lambda_translator = GTIRToSDFG(
+            nsdfg,
+            self.offset_provider_type,
```
philip-paul-mueller, line 670: Why do you no longer have to pass `lambda_params` to the function.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -703,7 +681,7 @@ def get_field_domain_offset(
         #
         connectivity_arrays = {
             dace_utils.connectivity_identifier(offset)
-            for offset in dace_utils.filter_connectivity_types(self.offset_provider_type)
+            for offset in dace_utils.filter_connectivity_types(self.offset_provider_type).keys()
```
philip-paul-mueller, line 684: The `keys()` here is technically unnecessary as it is the default when you iterate.

I also prefer doing it, but I was told to avoid it.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -864,13 +840,41 @@ def build_sdfg_from_gtir(
         offset_provider_type: The definitions of offset providers used by the program node
 
     Returns:
-        An SDFG in the DaCe canonical form (simplified)
+        The SDFG as a result of the GTIR lowering, without applying any transformation.
```
philip-paul-mueller, line 843: You should add to the documentation:
- That `PruneCasts` is run.
- The description of the `Returns` is a bit confusing as you apply transformation on the GTIR, but not on the SDFG.
    I would reformulate it to something like: "The SDFG generated through the lowering, without applying any post processing to it."

For further documentation you should also add that the SDFG follows the rules that are outlined in ADR18.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -163,7 +163,6 @@ class PrimitiveTranslator(Protocol):
     def __call__(
         self,
         node: gtir.Node,
-        sdfg: dace.SDFG,
         state: dace.SDFGState,
```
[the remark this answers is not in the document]
edopao: Yes, I could remove it. That would require a `get_current_state()` on the `SDFGBuilder` interface because a new state is added by each `Stmt`:
```
# visit one statement at a time and expand the SDFG from the current head state
for i, stmt in enumerate(node.body):
     # include `debuginfo` only for `ir.Program` and `ir.Stmt` nodes: finer granularity would be too messy
    head_state = self.sdfg.add_state_after(head_state, f"stmt_{i}")
    head_state._debuginfo = dace_utils.debug_info(stmt, default=self.sdfg.debuginfo)
    head_state = self.visit(stmt, state=head_state)
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: Yes, that would be simpler.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -172,21 +190,39 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
...
+                [gtir.Sym(id=k, type=v) for k, v in self.global_symbols.items()],
+                symbolic_arguments=self.domain_symbols,
+            )
+        )
+
     def get_offset_provider_type(self, offset: str) -> gtx_common.OffsetProviderTypeElem:
         return self.offset_provider_type[offset]
 
+    def get_sdfg(self) -> dace.SDFG:
+        return self.sdfg
```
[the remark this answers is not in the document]
edopao: I also do not like it, indeed. I could turn `SDFGBuilder` into a class that inherits from `dace.SDFG`, what do you think?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: I could create a wrapper of `add_nested_sdfg()` that changes the name of the nested SDFG, but I don't know if that is a better solution because it would silently change the name of the nested SDFG.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: I agree.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: Yes, I had similar thoughts.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: Because `lambda_params` corresponds to the set of keys in`lambda_global_symbols`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: That explains why it worked in baseline.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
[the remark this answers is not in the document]
edopao: @egparedes Do you have any advice how to include `sdfg` as an attribute of the `SDFGBuilder`?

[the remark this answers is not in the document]
egparedes: I don't fully understand what is the question here, @edopao, nor the comment about properties and data encapsulation by @philip-paul-mueller ...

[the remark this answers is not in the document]
philip-paul-mueller: Essentially it is
```python
class A:
    def get_sdfg(self) -> dace.SDFG:
        return self.sdfg
```
vs.
```python
class A:
    @property
    def sdfg(self) -> dace.SDFG:
        return self._sdfg
```

edopao:
Out-of-date and not on priority list.

PR #1762, opened 2024-12-01
feature[next]: Non-tree-size-increasing collapse tuple on ifs

tehrengruber:
Removing the tuple expressions across `if_` calls on ITIR has been a pain point in the past. While the `PROPAGATE_TO_IF_ON_TUPLES` option of the `CollapseTuplePass` works very reliably, the resulting increase in the tree size has been prohibitive. With the refactoring to GTIR this problem became much less pronounced, as we could restrict the propagation to field-level, i.e., outside of stencils, but the tree still grew exponentially in the number of references to boolean arguments used inside `if_` conditions. This PR adds an additional option `PROPAGATE_TO_IF_ON_TUPLES_CPS` to the `CollapseTuplePass`, which is similar to the existing `PROPAGATE_TO_IF_ON_TUPLES`, but propagates in the opposite direction, i.e. into the tree. This allows removal of tuple expressions across `if_` calls without increasing the size of the tree. This is particularly important for `if` statements in the frontend, where outwards propagation can have devastating effects on the tree size, without any gained optimization potential. For example
```
complex_lambda(if cond1
  if cond2
    {...}
  else:
    {...}
else
  {...})
```
is problematic, since `PROPAGATE_TO_IF_ON_TUPLES` would propagate, and hence duplicate, `complex_lambda` three times, while we only want to get rid of the tuple expressions inside of the `if_`s. Note that this transformation is not mutually exclusive to `PROPAGATE_TO_IF_ON_TUPLES`.


src/gt4py/next/iterator/type_system/inference.py
```diff
@@ -561,19 +594,22 @@ def visit_StencilClosure(self, node: itir.StencilClosure, *, ctx) -> it_ts.Stenc
...
-        return ts.DimensionType(dim=self.dimensions[node.value])
+        return ts.DimensionType(dim=common.Dimension(value=node.value, kind=node.kind))
 
     # TODO: revisit what we want to do with OffsetLiterals as we already have an Offset type in
     #  the frontend.
-    def visit_OffsetLiteral(self, node: itir.OffsetLiteral, **kwargs) -> it_ts.OffsetLiteralType:
+    def visit_OffsetLiteral(
+        self, node: itir.OffsetLiteral, **kwargs
+    ) -> it_ts.OffsetLiteralType | ts.DeferredType:
+        if self.reinfer:
```
tehrengruber, line 604: Write explanation

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -312,6 +358,73 @@ def transform_propagate_to_if_on_tuples(self, node: ir.FunCall, **kwargs) -> Opt
...
+                        returns=node.type,
+                    )
+                    f_params = [
+                        im.sym(self.uids.sequential_id(prefix="__ct_el_cps"), type_)
+                        for type_ in tuple_type.types
+                    ]
+                    f_args = [im.ref(param.id, param.type) for param in f_params]
+                    f_body = _with_altered_arg(node, i, im.make_tuple(*f_args))
+                    # simplify, e.g., inline trivial make_tuple args
+                    new_f_body = self.fp_transform(f_body, **kwargs)
```
havogt, line 391: In this recursion, we handle all the other args after the current one? I am confused...

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -312,6 +358,73 @@ def transform_propagate_to_if_on_tuples(self, node: ir.FunCall, **kwargs) -> Opt
...
+                    if new_f_body is f_body:
+                        continue
+                    # if the function is not trivial the transformation would still work, but
+                    # inlining would result in a larger tree again and we didn't didn't gain
+                    # anything compared to regular `propagate_to_if_on_tuples`. Not inling also
+                    # works, but we don't want bound lambda functions in our tree (at least right
+                    # now).
+                    if not _is_trivial_or_tuple_thereof_expr(new_f_body):
+                        continue
+                    f = im.lambda_(*f_params)(new_f_body)
```
havogt, line 403: basically here it's decided that we actually do something.

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -312,6 +358,73 @@ def transform_propagate_to_if_on_tuples(self, node: ir.FunCall, **kwargs) -> Opt
                     return im.if_(cond, new_true_branch, new_false_branch)
         return None
 
+    def transform_propagate_to_if_on_tuples_cps(
+        self, node: ir.FunCall, **kwargs
+    ) -> Optional[ir.Node]:
+        if not cpm.is_call_to(node, "if_"):
+            for i, arg in enumerate(node.args):
```
havogt, line 365: Here we iterate over any functions args or do we know more?

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -312,6 +358,73 @@ def transform_propagate_to_if_on_tuples(self, node: ir.FunCall, **kwargs) -> Opt
                     return im.if_(cond, new_true_branch, new_false_branch)
         return None
 
+    def transform_propagate_to_if_on_tuples_cps(
```
havogt, line 361: Can you break this function into smaller pieces, I am completely lost...

src/gt4py/next/iterator/type_system/inference.py
```diff
@@ -417,24 +430,44 @@ def apply(
...
+
+    @classmethod
+    def apply_reinfer(cls, node: T) -> T:
+        """
+        Given a partially typed node infer the type of ``node`` and its sub-nodes.
+
+        Contrary to the regular inference, this method does not descend into already typed sub-nodes
+        and can be used as a lightweight way to restore type information during a pass.
+
+        Note that this function is stateful, which is usually desired, and more performant.
```
havogt, line 448: Where is the statefulness?

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_collapse_tuple.py
```diff
@@ -239,3 +240,48 @@ def test_tuple_get_on_untyped_ref():
 
     actual = CollapseTuple.apply(testee, allow_undeclared_symbols=True, within_stencil=False)
     assert actual == testee
+
+
+def test_if_make_tuple_reorder_cps():
```
havogt, line 245: Can you add a test with ifs in more than one arg?

src/gt4py/next/iterator/transforms/collapse_tuple.py
tehrengruber: The expression is split into a continuation (which includes all the args following the current one) and the currently transformed arg. Since the currently used arg will use the newly build continuation it is natural to first build & transform the continuation and only then tackle the arg. In the new node (continuation + arg)  the continuation appears first. So in the expression the continuation appears first, but evaluation is still first arg then continuation. This is somewhat what continuations are all about (https://en.wikipedia.org/wiki/Continuation-passing_style):
> This has the effect of turning expressions "inside-out" because the innermost parts of the expression must be evaluated first

src/gt4py/next/iterator/transforms/collapse_tuple.py
tehrengruber: No we don't know more. We need to build and simplify the continuation first to know if we want to transform. It might be possible to handle multiple ifs, but the code would become more complex and error prone, and there is no benefit. It is just important to remember that the entire node is split into the continuation and the first matching `if` argument.

src/gt4py/next/iterator/type_system/inference.py
tehrengruber: Reformulated:
> Note that this function alters the input node, which is usually desired, and more performant.

src/gt4py/next/iterator/type_system/inference.py
```diff
@@ -417,24 +430,44 @@ def apply(
...
+            offset_provider_type=None, dimensions=None, allow_undeclared_symbols=True, reinfer=True
         )
+        instance.visit(node, ctx=_INITIAL_CONTEXT)
         return node
 
     def visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
+        # we found a node that is typed, do not descend into children
+        if self.reinfer and isinstance(node, itir.Node) and node.type:
+            if isinstance(node.type, ts.FunctionType):
+                return _type_synthesizer_from_function_type(node.type)
```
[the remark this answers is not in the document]
tehrengruber: Functions are represented as `TypeSynthesizer` during inference. See their docstring:

> Callable that given the type of the arguments to a function derives its return type.
> 
> In case the function is a higher-order function the returned value is not a type, but another
> function type-synthesizer.
> 
> In addition to the derivation of the return type a function type-synthesizer can perform checks
> on the argument types.
> 
> The motivation for this class instead of a simple callable is to allow
>  - isinstance checks to determine if an object is actually (meant to be) a type
>    synthesizer and not just any callable.
>  - writing simple type synthesizers without cluttering the signature with the additional
>    offset_provider_type argument that is only needed by some.


Since the function definition might be out-of-scope during reinference we rebuild the synthesizer from the `ts.FunctionType` for example present in a `SymRef` to that function.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_collapse_tuple.py
tehrengruber: Done: `test_nested_if_make_tuple_reorder_cps`

src/gt4py/next/iterator/transforms/collapse_tuple.py
tehrengruber: The construction of the continuation would be the only candidate, but it also doesn't fit well.

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -312,6 +358,73 @@ def transform_propagate_to_if_on_tuples(self, node: ir.FunCall, **kwargs) -> Opt
...
+            for i, arg in enumerate(node.args):
+                if cpm.is_call_to(arg, "if_"):
+                    itir_type_inference.reinfer(arg)
+                    if not any(isinstance(branch.type, ts.TupleType) for branch in arg.args[1:]):
+                        continue
+
+                    cond, true_branch, false_branch = arg.args
+                    tuple_type: ts.TupleType = true_branch.type  # type: ignore[assignment]  # type ensured above
+                    tuple_len = len(tuple_type.types)
+                    itir_type_inference.reinfer(node)
```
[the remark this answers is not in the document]
tehrengruber: Which part?

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -312,6 +367,100 @@ def transform_propagate_to_if_on_tuples(self, node: ir.FunCall, **kwargs) -> Opt
...
+                ]
+                f_args = [im.ref(param.id, param.type) for param in f_params]
+                f_body = _with_altered_arg(node, i, im.make_tuple(*f_args))
+                # simplify, e.g., inline trivial make_tuple args
+                new_f_body = self.fp_transform(f_body, **kwargs)
+                # if the continuation did not simplify there is nothing to gain. Skip
+                # transformation of this argument.
+                if new_f_body is f_body:
+                    continue
+                # if the function is not trivial the transformation we would get a larger tree
```
havogt, line 430:
```suggestion
                # if the function is not trivial the transformation would create a larger tree
```

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -312,6 +367,100 @@ def transform_propagate_to_if_on_tuples(self, node: ir.FunCall, **kwargs) -> Opt
...
+                    continue
+                f = im.lambda_(*f_params)(new_f_body)
+
+                # this is the symbol refering to the tuple value inside the two branches of the
+                # if, e.g. a symbol refering to `{1, 2}` and `{3, 4}` respectively
+                tuple_var = self.uids.sequential_id(prefix="__ct_tuple_cps")
+                # this is the symbol refering to our continuation, e.g. `cont` in our example.
+                f_var = self.uids.sequential_id(prefix="__ct_cont")
+                new_branches = []
+                for branch in arg.args[1:]:
```
havogt, line 442:
```suggestion
                for branch in [true_branch, false_branch]:
```

havogt (approved):
I only have a few proposals to improve readability.

PR #1764, opened 2024-12-02
feat[next]: Improve fieldop fusion

tehrengruber:
- Add support for inlining into scans.
- Fuse `make_tuple(as_fieldop(...), as_fieldop(...))` calls into `as_fieldop(λ(...) → make_tuple(...))(...)`.
- Refactor pass such that inlining decision is expressed in a dedicated function `_arg_inline_predicate`.
- Inline all let vars with dtype list.
- Performance improvement: Stop visiting when reaching a stencil.
- Bugfix for inlining of `as_fieldop` args that use the same arg twice, e.g. `as_fieldop(...)(a, b)`.
- Bugfix such that only expressions inside the expr of an `itir.SetAt` are considered.


src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
```diff
@@ -83,7 +87,12 @@ def _inline_as_fieldop_arg(
 
     for inner_param, inner_arg in zip(stencil.params, inner_args, strict=True):
         if isinstance(inner_arg, itir.SymRef):
-            stencil_params.append(inner_param)
+            if inner_arg.id in extracted_args:
+                assert extracted_args[inner_arg.id] == inner_arg
+                alias = stencil_params[list(extracted_args.keys()).index(inner_arg.id)]
+                stencil_body = im.let(inner_param, im.ref(alias.id))(stencil_body)
+            else:
+                stencil_params.append(inner_param)
```
tehrengruber, line 95: This is a fix for the case covered in [test_trivial_same_arg_twice](https://github.com/GridTools/gt4py/pull/1764/files#diff-b27471e57df42d67a9d9c781f66431cdcd867d6d0cccb5de6b372466c8136147R58)

src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
```diff
@@ -119,9 +175,7 @@ def fuse_as_fieldop(
                 pass
             elif cpm.is_call_to(arg, "if_"):
```
havogt, line 176: I was going through the whole file again and got stuck here. Can we briefly discuss
- if it makes the code cleaner, if instead of building an applied field_op we extract the stencil_body and args
- instead of special casing for `if` we support any function (because only `if` should be marked as eligible, but if we change the algorithm to mark eligible, we should not have to touch this line)

src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
```diff
@@ -204,48 +305,141 @@ def apply(
...
+        return new_node
+
+    def visit_SetAt(self, node: itir.SetAt, **kwargs):
+        return itir.SetAt(
+            expr=self.visit(node.expr, **kwargs | {"within_set_at_expr": True}),
+            domain=node.domain,
+            target=node.target,
+        )
+
+    def visit_FunCall(self, node: itir.FunCall, **kwargs):
```
havogt, line 337: Let's rewrite the whole thing together for 2 reasons:
- knowledge sharing
- after it is now (maybe) correct, focus on readability.

havogt:
Can we generate a complicated IR (which we also can easily maintain) as a performance regression test?

havogt (approved):
looks good for now. as discussed, it would be good to spend time later to improve interoperability of transformations.

PR #1766, opened 2024-12-03
feat[next]: Change interval syntax in ITIR pretty printer

tehrengruber:
We currently use `)` in the pretty printer to express an open interval. This is quite cumbersome when debugging the IR because it breaks matching parenthesis in the editor of functions and calls, e.g. when does a function start and end.  This PR simply uses `[` instead.

### DIFF
--- a/src/gt4py/next/iterator/ir_utils/ir_makers.py
+++ b/src/gt4py/next/iterator/ir_utils/ir_makers.py
@@ -423,11 +423,11 @@ def domain(
     ...         },
     ...     )
     ... )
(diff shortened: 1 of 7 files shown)


havogt:
Actually, what about using `(a,b)` and say it's Python range semantics?

egparedes:
@tehrengruber I know this PR has been already merged but in my opinion this looks really terrible, just like a bug.  Why not using a different right bracket unicode symbol for this? For example:
```
[⟧  [⧽  [〗[⁆  [】[⦊  [⦈
```
I found a longer list of options here: https://stackoverflow.com/a/13535289

PR #1771, opened 2024-12-04
feat[next]: allocator in field

havogt:
sketch of the idea

### DIFF
--- a/src/gt4py/next/constructors.py
+++ b/src/gt4py/next/constructors.py
@@ -11,6 +11,8 @@ from __future__ import annotations
 from collections.abc import Mapping, Sequence
 from typing import Optional, cast
 
+from typing_extensions import NotRequired, TypedDict
+
 import gt4py._core.definitions as core_defs
 import gt4py.eve as eve
 import gt4py.eve.extended_typing as xtyping
@@ -77,10 +79,11 @@ def empty(
     dtype = core_defs.dtype(dtype)
     if allocator is None and device is None:
         device = core_defs.Device(core_defs.DeviceType.CPU, device_id=0)
-    buffer = next_allocators.allocate(
+    allocate = next_allocators.make_concrete_allocator(
         domain, dtype, aligned_index=aligned_index, allocator=allocator, device=device
     )
-    res = common._field(buffer.ndarray, domain=domain)
+    buffer = allocate()
+    res = common._field(buffer, domain=domain, allocator=allocate)
     assert isinstance(res, common.MutableField)
     assert isinstance(res, nd_array_field.NdArrayField)
     return res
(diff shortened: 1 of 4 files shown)


src/gt4py/next/constructors.py
```diff
@@ -349,12 +352,51 @@ def as_connectivity(
...
+    if domain is not None:
+        kwargs["domain"] = domain
+    if dtype is not None:
+        kwargs["dtype"] = core_defs.dtype(dtype)
+    if aligned_index is not None:
+        kwargs["aligned_index"] = aligned_index
+    if allocator is not eve.NOTHING:
+        kwargs["allocator"] = allocator
+    if device is not eve.NOTHING:
+        kwargs["device"] = device
```
havogt, line 399:
```suggestion
    if allocator is not None:
        kwargs["allocator"] = allocator
    if device is not None:
        kwargs["device"] = device
```

havogt:
- [ ] think about  long names -> maybe rename to `NDArray...`
- [ ] `__copy__`, `__deepcopy__`
- [ ] `as_ndarray(allocator: ConcreteAllocator, copy: Optional[bool])`
- [ ] can `TensorBuffer` be removed?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -428,6 +434,24 @@ def _slice(
...
+        allocation_ns = (
+            self._allocation_ns
+            if self._allocation_ns is not None
+            else _allocators.get_array_allocation_namespace(self.array_ns)
+        )
+        ndarray_copy = allocation_ns.asarray(
+            self.ndarray,
+            domain=self.domain,
+            dtype=self.dtype,
+            copy=True,  # aligned_index???
```
havogt, line 448: Actually this is a missing piece in the current implementation. We should probably provide this as a default in all functions.

src/gt4py/next/_allocators.py
```diff
@@ -323,20 +383,215 @@ def allocate(
...
+        device = core_defs.Device(actual_allocator.__gt_device_type__, 0)
+    elif device.device_type != actual_allocator.__gt_device_type__:
+        raise ValueError(f"Device '{device}' and allocator '{actual_allocator}' are incompatible.")
+    return actual_allocator, device
+
+
+def get_array_allocation_namespace(
+    allocator: Optional[FieldBufferAllocationUtil | core_defs.ArrayApiNamespace],
+    device: Optional[core_defs.Device] = None,
+) -> GTArrayAllocationNamespace:
```
havogt, line 423: Maybe put aligned_index here and add it as a default to the construction functions.
Same for device.

tests/next_tests/unit_tests/test_constructors.py
```diff
@@ -21,16 +37,49 @@
...
+            v, id=_pretty_print_allocator_device_namespace(v), marks=pytest.mark.requires_jax
+        )
+
+
+@pytest.fixture(params=allocator_device_refnamespace_params())
+def allocator_device_refnamespace(request):
+    return request.param
+
+
+def test_empty(allocator_device_refnamespace):
```
egparedes, line 80: Just a suggestion: we could avoid the intermediate fixture here and just parametrize the test explicitly: 
```suggestion
@pytest.mark.parametrize("allocator, device, xp", [*allocator_device_refnamespace_params()])
def test_empty(allocator, device, xp):
```

tests/next_tests/unit_tests/test_constructors.py
```diff
@@ -57,50 +99,48 @@ def test_zeros(allocator, device):
...
         device=device,
     )
-    ref = np.full((sizes[I] - 2, sizes[J] - 2), 42.0).astype(gtx.float32)
+    ref = xp.full((sizes[I] - 2, sizes[J] - 2), 42.0).astype(gtx.float32)
+
+    assert xp.array_equal(a.ndarray, ref)
 
-    assert np.array_equal(a.ndarray, ref)
+
+def test_copy():
```
egparedes, line 134: Isn't this test the same than the one in the embedded `test_nd_array_field` tests?

egparedes (commented):
I added some basic comments, but it's not a detailed review. I agree with should use this strategy short-term to be able to use `jax` and similar immutable data array api libraries, but I'm not sure yet if the current implementation strategy is good enough or it should be refactored. I'd propose to discuss it further offline.

PR #1774, opened 2024-12-06
refactor[next]: remove FencilDefinition definition

havogt:
After this commit `FencilDefinition`s are completely removed. Next step could be to cleanup `itir` -> `gtir` everywhere.


edopao (approved):
LGTM

PR #1775, opened 2024-12-06
feat[dace]: Restirct Loop Blocking

philip-paul-mueller:
Made it possible to disable loop blocking if there are no independent nodes.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_optimize.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_optimize.py
@@ -32,7 +32,7 @@ def gt_auto_optimize(
     gpu_block_size: Optional[Sequence[int | str] | str] = None,
     blocking_dim: Optional[gtx_common.Dimension] = None,
     blocking_size: int = 10,
-    require_independent_nodes: Optional[bool] = None,
+    blocking_only_if_independent_nodes: Optional[bool] = None,
     reuse_transients: bool = False,
     gpu_launch_bounds: Optional[int | str] = None,
     gpu_launch_factor: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_optimize.py
```diff
@@ -32,6 +32,7 @@ def gt_auto_optimize(
     gpu_block_size: Optional[Sequence[int | str] | str] = None,
     blocking_dim: Optional[gtx_common.Dimension] = None,
     blocking_size: int = 10,
+    require_independent_nodes: Optional[bool] = None,
```
edopao:
```suggestion
    blocking_only_if_independent_nodes: Optional[bool] = None,
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_optimize.py
```diff
@@ -90,6 +91,8 @@ def gt_auto_optimize(
             one for all.
         blocking_dim: On which dimension blocking should be applied.
         blocking_size: How many elements each block should process.
+        require_independent_nodes: If `True` only apply loop blocking if there
```
edopao:
```suggestion
        blocking_only_if_independent_nodes: If `True` only apply loop blocking in a dimension X if the map contains nodes that do not depend on X.
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/loop_blocking.py
```diff
@@ -42,7 +42,8 @@ class LoopBlocking(dace_transformation.SingleStateTransformation):
     Args:
         blocking_size: The size of the block, denoted as `B` above.
         blocking_parameter: On which parameter should we block.
-
+        require_independent_nodes: If `True` then the transformation will only
```
edopao, line 45:
```suggestion
        blocking_only_if_independent_nodes: If `True` only apply loop blocking in a dimension X if the map contains nodes that do not depend on X.
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/auto_optimize.py
philip-paul-mueller: The whole documentation and naming uses the word dependent and independent so we should stuck to it too.
I updated the description of the `LoopBlocking` transformation to be a bit more clearer.
I also added a reference to the the `LoopBlocking` transformation for that parameter.

edopao (approved):

PR #1776, opened 2024-12-06
feat[next][dace]: lowering of scan to SDFG

edopao:
This PR contains the lowering of the scan builtin function.


ci/cscs-ci.yml
```diff
@@ -156,7 +156,7 @@ build_py38_image_x86_64:
   variables:
     CRAY_CUDA_MPS: 1
     SLURM_JOB_NUM_NODES: 1
-    SLURM_TIMELIMIT: 15
+    SLURM_TIMELIMIT: 20
```
edopao: @philip-paul-mueller The execution of gpu tests without optimization is probably very slow.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/gpu_utils.py
```diff
@@ -76,6 +76,15 @@ def gt_gpu_transformation(
         len(kwargs) == 0
     ), f"gt_gpu_transformation(): found unknown arguments: {', '.join(arg for arg in kwargs.keys())}"
 
+    # We run simplify to bring the SDFG into a canonical form that the gpu transformations
+    # can handle. This is a workaround for an issue with scalar expressions that are
+    # promoted to symbolic expressions and accessed on the host (CPU) as arguments of
+    # interstate edge conditions, but the scalar data is stored as a GPU global variable.
+    # For details, see the dace issue https://github.com/spcl/dace/issues/1773
+    dace_transformation.passes.SimplifyPass(
```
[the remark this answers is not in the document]
edopao: An alternative, unless we find the proper fix, is to exclude the testcase `test_ternary_scan` for now.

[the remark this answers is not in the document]
edopao: I also tried to use `gtx_transformations.gt_simplify(sdfg, skip={"ConstantPropagation"})` but it resulted in some deadlock.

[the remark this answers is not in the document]
edopao: There is still some deadlock happening. I will revert this change and disable the test case.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -756,6 +858,298 @@ def translate_scalar_expr(
...
+            temp, temp_desc = nsdfg.add_temp_transient_like(data_desc)
+            temp_node = compute_state.add_access(temp)
+            compute_state.add_nedge(data_node, temp_node, dace.Memlet.from_array(temp, temp_desc))
+
+    # build the mapping of symbols from nested SDFG to parent SDFG
+    nsdfg_symbols_mapping: dict[str, dace.symbolic.SymExpr] = {}
+    for dim, _, _ in domain:
+        if dim != scan_dim:
+            dim_map_variable = dace_gtir_utils.get_map_variable(dim)
+            nsdfg_symbols_mapping[dim_map_variable] = dim_map_variable
```
philip-paul-mueller, line 1094: You have a type error here, because `dim_map_variable` must be a `str`, but at the same time it must also be a `SymExpr`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -546,6 +562,226 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
 
         return self._construct_tasklet_result(field_desc.dtype, deref_node, "val")
 
+    def _visit_if(self, node: gtir.FunCall) -> ValueExpr | tuple[ValueExpr | tuple[Any, ...], ...]:
```
philip-paul-mueller, line 565: Why you are also needing ifs in scans?

I also did not fully look at this function as it is very complex, with its several levels of nesting.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -546,6 +562,226 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+        nsdfg = dace.SDFG(self.unique_nsdfg_name(prefix="if_stmt"))
+        nsdfg.debuginfo = dace_utils.debug_info(node, default=self.sdfg.debuginfo)
+
+        if use_conditional_block:
+            if_region = dace.sdfg.state.ConditionalBlock("if")
+            nsdfg.add_node(if_region)
+            entry_state = nsdfg.add_state("entry", is_start_block=True)
+            nsdfg.add_edge(entry_state, if_region, dace.InterstateEdge())
+
+            if_body = dace.sdfg.state.ControlFlowRegion("if_body", sdfg=nsdfg)
```
philip-paul-mueller, line 590: Suggestion, call it `then_body` because it matches then good with `if-then-else`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -546,6 +562,226 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+                    dace_subsets.Range.from_string("0"),
+                )
+            else:
+                assert isinstance(condition_value, MemletExpr)
+                input_memlets["__cond"] = (condition_value.dc_node, condition_value.subset)
+                nsdfg_symbol_mapping.update(
+                    {sym: sym for sym in condition_value.subset.free_symbols}
+                )
+
+        def visit_branch(
```
philip-paul-mueller, line 625: I would call it `_visit_if_branch`.
- It is used by the if translator.
- You should make it a free function.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -546,6 +562,226 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+            state: dace.SDFGState, expr: gtir.Expr
+        ) -> tuple[
+            list[DataflowInputEdge],
+            DataflowOutputEdge | tuple[DataflowOutputEdge | tuple[Any, ...], ...],
+        ]:
+            assert state in nsdfg.states()
+
+            def visit_arg(arg: IteratorExpr | DataExpr) -> IteratorExpr | ValueExpr:
```
philip-paul-mueller, line 633: This is an inner function in an inner function.
This is too much nesting in my opinion.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -546,6 +562,226 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+                        arg_subset = arg.subset
+                    else:
+                        arg_desc = arg_node.desc(self.sdfg)
+                        arg_subset = None
+
+                arg_data = arg_node.data
+                # SDFG data containers with name prefix '__tmp' are expected to be transients
+                inner_data = (
+                    arg_data.replace("__tmp", "__input")
+                    if arg_data.startswith("__tmp")
```
philip-paul-mueller, line 654: Just a very picky comment.
You explicitly check if the variable starts with `__tmp`, but replace will replace every occurence of `__tmp`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -546,6 +562,226 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+                )
+
+                try:
+                    inner_desc = nsdfg.data(inner_data)
+                    assert not inner_desc.transient
+                except KeyError:
+                    inner_desc = arg_desc.clone()
+                    inner_desc.transient = False
+                    nsdfg.add_datadesc(inner_data, inner_desc)
+                    input_memlets[inner_data] = (arg_node, arg_subset)
```
philip-paul-mueller, line 665: This looks strange, you should rewrite it with an `if`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -546,6 +562,226 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+
+                if arg_subset:
+                    # symbols used in memlet subset are not automatically mapped to the parent SDFG
+                    nsdfg_symbol_mapping.update({sym: sym for sym in arg_subset.free_symbols})
```
philip-paul-mueller, line 669:
```suggestion
                    nsdfg_symbol_mapping.update({str(sym): sym for sym in arg_subset.free_symbols})
```
As the signature of `nsdfg_symbol_mapping` says that the key is `str`.
Btw, if I am not mistaken, this variable was defined two levels above.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -546,6 +562,226 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+                edge.connect(map_entry=None)
+
+            def construct_output(
+                output_state: dace.SDFGState, edge: DataflowOutputEdge, sym: gtir.Sym
+            ) -> ValueExpr:
+                output_data = str(sym.id)
+                try:
+                    output_desc = nsdfg.data(output_data)
+                    assert not output_desc.transient
+                except KeyError:
```
philip-paul-mueller, line 710: Using an `if` is clearer.

philip-paul-mueller (commented):
The code is quite complex, so I think a refactoring is needed to make it simpler.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -756,6 +858,298 @@ def translate_scalar_expr(
...
+            update_expr=f"{scan_loop_var} = {scan_loop_var} - 1",
+            inverted=False,
+        )
+
+    nsdfg.add_node(scan_loop)
+    compute_state = scan_loop.add_state("scan_compute", is_start_block=True)
+    update_state = scan_loop.add_state("scan_update")
+    scan_loop.add_edge(compute_state, update_state, dace.InterstateEdge())
+
+    init_state = nsdfg.add_state("scan_init", is_start_block=True)
```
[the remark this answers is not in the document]
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -756,6 +858,298 @@ def translate_scalar_expr(
...
+        scan_output_edge: gtir_dataflow.DataflowOutputEdge, sym: gtir.Sym
+    ) -> FieldopData:
+        scan_result = scan_output_edge.result
+        assert isinstance(scan_result.gt_dtype, ts.ScalarType)
+        assert scan_result.gt_dtype == sym.type
+        scan_result_data = scan_result.dc_node.data
+        scan_result_desc = scan_result.dc_node.desc(nsdfg)
+
+        output, _ = nsdfg.add_array(
+            scan_output_name(sym.id), scan_output_shape, scan_result_desc.dtype, find_new_name=True
```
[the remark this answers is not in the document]
edopao: You are right.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -199,6 +207,25 @@ def make_field(
...
+    def is_column_dimension(self, dim: gtx_common.Dimension) -> bool:
+        assert self.column_dim
+        return dim == self.column_dim
+
+    def nested_context(
+        self,
+        sdfg: dace.SDFG,
+        global_symbols: dict[str, ts.DataType],
+        field_offsets: dict[str, Optional[list[dace.symbolic.SymExpr]]],
+    ) -> SDFGBuilder:
```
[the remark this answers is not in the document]
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -379,7 +407,7 @@ def _add_sdfg_params(
         self,
         sdfg: dace.SDFG,
         node_params: Sequence[gtir.Sym],
-        symbolic_arguments: set[str],
+        symbolic_arguments: Optional[set[str]],
```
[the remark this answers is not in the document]
edopao: `Optional` is not needed anymore.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -111,6 +112,21 @@ def get_symbol_type(self, symbol_name: str) -> ts.DataType:
...
+        ...
+
+    @abc.abstractmethod
+    def nested_context(
+        self,
+        sdfg: dace.SDFG,
+        global_symbols: dict[str, ts.DataType],
+        field_offsets: dict[str, Optional[list[dace.symbolic.SymExpr]]],
+    ) -> SDFGBuilder:
+        """Create a new empty context, useful to build a nested SDFG."""
```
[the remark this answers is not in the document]
edopao: Done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -830,6 +930,288 @@ def translate_scalar_expr(
...
+    )
+
+    # now initialize the scan carry
+    scan_carry_input = (
+        dace_gtir_utils.make_symbol_tree(scan_carry, scan_carry_type)
+        if isinstance(scan_carry_type, ts.TupleType)
+        else im.sym(scan_carry, scan_carry_type)
+    )
+
+    def init_scan_carry(sym: gtir.Sym) -> None:
```
philip-paul-mueller, line 1084: I would turn this function as well as `connect_scan_output()` and `construct_output_edge()` into free independent function.
I also think that the file has become a bit too large, so I would recommend to you to split the file into several, and put the scan implementation into one.
I would recommend the following layout:
- `utils.py`
- `scan.py`
- `misc.py` (contains all translators that do not have dedicated file)

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -830,6 +930,288 @@ def translate_scalar_expr(
...
+    # connect the dataflow result nodes to the carry variables
+    def connect_scan_output(
+        scan_output_edge: gtir_dataflow.DataflowOutputEdge, sym: gtir.Sym
+    ) -> FieldopData:
+        scan_result = scan_output_edge.result
+        assert isinstance(scan_result.gt_dtype, ts.ScalarType)
+        scan_result_data = scan_result.dc_node.data
+        scan_result_desc = scan_result.dc_node.desc(nsdfg)
+
+        # `sym` represents the global output data, that is the lambda output connector
```
philip-paul-mueller, line 1116: Sorry what do you mean with "lambda output connector".

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -830,6 +930,288 @@ def translate_scalar_expr(
...
+        lambda_output = str(sym.id)
+        output = scan_output_name(lambda_output)
+        assert scan_result.gt_dtype == sym.type
```
philip-paul-mueller, line 1119: You should move that up after the other `assert` it is out of place here.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -830,6 +930,288 @@ def translate_scalar_expr(
...
+    # the reason is that the map for horizontal domain is outside the scan loop region
+    for edge in input_edges:
+        edge.connect(map_entry=None)
+
+    # connect the dataflow result nodes to the carry variables
+    def connect_scan_output(
+        scan_output_edge: gtir_dataflow.DataflowOutputEdge, sym: gtir.Sym
+    ) -> FieldopData:
+        scan_result = scan_output_edge.result
+        assert isinstance(scan_result.gt_dtype, ts.ScalarType)
```
philip-paul-mueller, line 1112: I am a bit lost, but could you tell me why the scan result is always a scalar?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -830,6 +930,288 @@ def translate_scalar_expr(
...
+            (not data_desc.transient)
+            or data_node.data.startswith(
+                scan_carry
+            )  # exceptional case where the carry variable is not used, not a scan indeed
+        ):
+            compute_state.remove_node(data_node)
+
+    # build the mapping of symbols from nested SDFG to parent SDFG
+    nsdfg_symbols_mapping = {str(sym): sym for sym in nsdfg.free_symbols}
+    for inner, arg in lambda_flat_args.items():
```
philip-paul-mueller, line 1157: I would call it `inner_arg` and `outer_arg`.
My main point is that `arg` is in my view not screaming enough "I am on the outside".

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -830,6 +930,288 @@ def translate_scalar_expr(
     return FieldopData(temp_node, node.type, offset=None)
 
 
+def translate_scan(
```
philip-paul-mueller, line 933: I think you should split this function into three sub functions.
One for handling the input another for translating the body and one for integrating the nested sdfg node into the parent SDFG node.
Because currently this function is way too big to understand and to reason about it.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -130,6 +131,37 @@ def get_symbol_type(self, symbol_name: str) -> ts.DataType:
...
+    @abc.abstractmethod
+    def nested_context(
+        self,
+        expr: gtir.Expr,
+        sdfg: dace.SDFG,
+        global_symbols: dict[str, ts.DataType],
+        field_offsets: dict[str, Optional[list[dace.symbolic.SymExpr]]],
+    ) -> SDFGBuilder:
+        """
+        Create a new context for lowering of an expression in a nested SDFG.
```
philip-paul-mueller, line 148: I like this description, however to make it clearer (especially to people like me) I would suggest to you to somehow add something like "to translate a nested expression in an independent way from the currently translated expression".

philip-paul-mueller (commented):
It looks generally good but I would strongly suggest to split the `translate_scan` function into several smaller functions that each handle one particular aspect of the translation.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -830,6 +930,288 @@ def translate_scalar_expr(
...
+
+    # make naming consistent throughut this function scope
+    def scan_input_name(input_name: str) -> str:
+        return f"__gtir_scan_input_{input_name}"
+
+    def scan_output_name(input_name: str) -> str:
+        return f"__gtir_scan_output_{input_name}"
+
+    # visit the initialization value of the scan expression
+    init_data = sdfg_builder.visit(init_value, sdfg=sdfg, head_state=state)
```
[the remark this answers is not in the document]
edopao: Good idea.

edopao (commented):
I have addressed all comments except splitting the lowering of the scan into multiple functions / separate module. I will push another commit on this PR for that, so it is easier to review.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -357,37 +426,65 @@ def _create_field_operator(
...
-            for dim, lower_bound, upper_bound in domain
-        },
-    )
+    domain_dims, _, _ = _get_field_layout(domain)
+
+    assert scan_dim is None or scan_dim in domain_dims
+    if scan_dim and len(domain_dims) == 1:
+        # We construct the scan field operator only on the horizontal domain.
+        # If the field operator computes only the scan dimension,
+        # there is no horizontal domain, therefore the map scope is not needed.
```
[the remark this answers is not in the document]
edopao: Yes, exactly. I can explain it a bit better.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -332,6 +344,62 @@ def _create_field_operator_impl(
...
+        raise NotImplementedError("List of values not supported in scan field operators.")
+
+    # allocate local temporary storage
+    assert dataflow_output_desc.dtype == dace_utils.as_dace_type(field_dtype)
+    field_name, field_desc = sdfg_builder.add_temp_array(
+        sdfg, field_shape, dataflow_output_desc.dtype
+    )
+    scan_output_stride = field_desc.strides[scan_dim_index]
+    dataflow_output_desc.strides = (scan_output_stride,)
+    field_node = state.add_access(field_name)
```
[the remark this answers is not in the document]
edopao: I forgot about this comment, I am going to fix it in next commit.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: For now I will move the scan translate function to a separate module.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: Done.

edopao:
Thanks for the review comments. I have moved the scan translator to a separate module and split it into small functions. At some point I stopped, in order to avoid too much fragmentation. I agree, it looks much better now!

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_scan_translator.py
```diff
@@ -0,0 +1,657 @@
...
+        inner_desc = nsdfg.data(inner_dataname)
+        outer_desc = outer_arg.dc_node.desc(sdfg)
+        nsdfg_symbols_mapping |= {
+            str(nested_symbol): parent_symbol
+            for nested_symbol, parent_symbol in zip(
+                [*inner_desc.shape, *inner_desc.strides],
+                [*outer_desc.shape, *outer_desc.strides],
+                strict=True,
+            )
+            if isinstance(nested_symbol, dace.symbol)
```
philip-paul-mueller, line 625: You should explain why you only use `dace.symbol` here and not also `SymExpr`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_scan_translator.py
```diff
@@ -0,0 +1,657 @@
...
+    """
+    domain_dims, _, _ = gtir_translators.get_field_layout(domain)
+
+    if len(domain_dims) == 1:
+        # We construct the scan field operator on the horizontal domain, while the
+        # vertical dimension (the column axis) is computed by the loop region.
+        # If the field operator computes only the column axis (a 1d scan field operator),
+        # there is no horizontal domain, therefore the map scope is not needed.
+        # This case currently triggers a DaCe issue and produces wrong CUDA code,
+        # thus the corresponding test is disabled (see pytest marker `uses_scan_1d_field`).
```
philip-paul-mueller, line 203: I would be more specific and say that it refers to a problem in the GPU transformation.
Also if possible you should refer to a DaCe issue.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_scan_translator.py
```diff
@@ -0,0 +1,657 @@
...
+        map_entry, map_exit = (None, None)
+    else:
+        # create map range corresponding to the field operator domain
+        map_entry, map_exit = sdfg_builder.add_map(
```
philip-paul-mueller, line 207: I am a bit confused.
Where do these maps life?
From my understanding they must be located inside the loop region, and must wrap the dataflow generated from scan expression, which implies that it is located inside the nested SDFG.
Is this correct, because it is not obvious from how it is called.

philip-paul-mueller (commented):
There are some smaller issues, but also one that is a bit bigger.
I think my main confusion comes from how it is translated.
As far as I can remember a lowered scan should look something like:
```
for(int i = scan_start; i != scan_end; ++i)
{
        map_over_horizonal(); //Parallel
}
```
But is it the other way around?

edopao:
> There are some smaller issues, but also one that is a bit bigger. I think my main confusion comes from how it is translated. As far as I can remember a lowered scan should look something like:
> 
> ```
> for(int i = scan_start; i != scan_end; ++i)
> {
>         map_over_horizonal(); //Parallel
> }
> ```
> 
> But is it the other way around?

It is the other way around. It has always been like that, even in the ITIR backend. It is not like it should or should not, both ways will produce the same result. I am using the external map because it follows the GTIR representation, where `scan` is an operator called inside a field operator.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_scan_translator.py
```diff
@@ -0,0 +1,678 @@
...
+        field_shape = [*domain_shape, dataflow_output_desc.shape[1]]
+        field_offset = [*domain_offset, dataflow_output_desc.offset[1]]
+        field_subset = field_subset + dace_subsets.Range.from_string(
+            f"0:{dataflow_output_desc.shape[1]}"
+        )
+
+    # allocate local temporary storage
+    assert dataflow_output_desc.dtype == dace_utils.as_dace_type(field_dtype)
+    field_name, field_desc = sdfg_builder.add_temp_array(
+        sdfg, field_shape, dataflow_output_desc.dtype
```
philip-paul-mueller, line 157: Just to be sure, this array life outside a map?
It is because we have to avoid that `new` is called inside it.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_scan_translator.py
```diff
@@ -0,0 +1,678 @@
...
+        if isinstance(scan_carry_symbol.type, ts.TupleType)
+        else scan_carry_symbol
+    )
+
+    def init_scan_carry(sym: gtir.Sym) -> None:
+        scan_carry_dataname = str(sym.id)
+        scan_carry_desc = nsdfg.data(scan_carry_dataname)
+        input_scan_carry_dataname = _scan_input_name(scan_carry_dataname)
+        input_scan_carry_desc = scan_carry_desc.clone()
+        nsdfg.add_datadesc(input_scan_carry_dataname, input_scan_carry_desc)
```
philip-paul-mueller, line 406: I do not understand this.
You get a data descriptor from the nested SDFG, clone it and add it then to the nested SDFG again, finally you make the original data descriptor a transient.
And then you connect those twos, why not keep using the original one?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_scan_translator.py
```diff
@@ -0,0 +1,678 @@
...
+
+        # `sym` represents the global output data, that is the nested-SDFG output connector
+        lambda_output = str(sym.id)
+        output = _scan_output_name(lambda_output)
+        nsdfg.add_array(output, scan_output_shape, scan_result_desc.dtype)
+        output_node = compute_state.add_access(output)
+
+        compute_state.add_nedge(
+            scan_result.dc_node, output_node, dace.Memlet(data=output, subset=output_subset)
+        )
```
philip-paul-mueller, line 457: Here you write `scan_result.dc_node` into `output_node`.
However, below you write scan result into lambda output.
For what do you need the `output_node`?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_scan_translator.py
```diff
@@ -0,0 +1,678 @@
...
+    # for output connections, we create temporary arrays that contain the computation
+    # results of a column slice for each point in the horizontal domain
+    lambda_output_tree = gtx_utils.tree_map(
+        lambda lambda_output_data: _connect_nested_sdfg_output_to_temporaries(
+            sdfg, nsdfg, nsdfg_node, state, lambda_output_data
+        )
+    )(lambda_output)
+
+    # we call a helper method to create a map scope that will compute the entire field
+    return _create_scan_field_operator(
```
philip-paul-mueller, line 676: How does this function knows where the body of the scan is.

Furthermore, for me it looks like everything is done at this place.
I mean the nested SDFG containing the body has been created, the outputs are connected, everything that is left is the connection of the inputs, which `_create_scan_field_operator()` does, but then that function also calls `_create_scan_field_operator_impl()` which does much more.
This confuses me.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_scan_translator.py
```diff
@@ -0,0 +1,678 @@
...
+        )
+        lambda_input_edges.append(input_edge)
+
+    # for output connections, we create temporary arrays that contain the computation
+    # results of a column slice for each point in the horizontal domain
+    lambda_output_tree = gtx_utils.tree_map(
+        lambda lambda_output_data: _connect_nested_sdfg_output_to_temporaries(
+            sdfg, nsdfg, nsdfg_node, state, lambda_output_data
+        )
+    )(lambda_output)
```
philip-paul-mueller, line 673: There is an asymmetry between inputs and outputs.
You connect the outputs here but the inputs are on the return statement.

I know it is quite late, but I would think it is more logical if the function would do:
- Creating/setting up the inputs,
- Creating the body,
- Setting up the outputs.

Is there a particular reason why currently it is handled as:
- Creating body,
- Handle output,
- Handle inputs.

To me the function appears a bit unstructured.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_scan_translator.py
```diff
@@ -0,0 +1,678 @@
...
+    sdfg: dace.SDFG,
+    state: dace.SDFGState,
+    sdfg_builder: gtir_sdfg.SDFGBuilder,
+) -> gtir_translators.FieldopResult:
+    """
+    Generates the dataflow subgraph for the `as_fieldop` builtin with a scan operator.
+
+    It differs from `translate_as_fieldop()` in that the horizontal domain is lowered
+    to a map scope while the vertical dimension where the scan column is computed
+    is lowered to a `LoopRegion`. The current design choice is to keep the map scope
```
philip-paul-mueller, line 541:
```suggestion
    is lowered to a `LoopRegion`, that is inside the horizontal map. The current design choice is to keep the map scope
```
I think this information is important as it confused me for a long time.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_scan_translator.py
```diff
@@ -0,0 +1,678 @@
...
+        assert isinstance(scan_init_data.gt_type, ts.ListType)
+        assert scan_init_data.gt_type.offset_type
+        offset_type = scan_init_data.gt_type.offset_type
+        offset_provider_type = sdfg_builder.get_offset_provider_type(offset_type.value)
+        assert isinstance(offset_provider_type, gtx_common.NeighborConnectivityType)
+        list_size = offset_provider_type.max_neighbors
+        return [scan_column_size, dace.symbolic.SymExpr(list_size)]
+
+    if isinstance(init_data, tuple):
+        lambda_result_shape = gtx_utils.tree_map(get_scan_output_shape)(init_data)
```
philip-paul-mueller, line 353: Okay I get that part of the output shape can be inferred from the shape of the initial data.
But what is with the horizontal isn't it missing?

Furthermore, this value, `lambda_result_shape`, is defined here and then used 100 lines later.
Because of this I have the feeling that its definition and its usage is fully disconnected.

philip-paul-mueller (commented):
I think that everything that is needed is here, but it is organized in a way that I not fully understand.

PR #1779, opened 2024-12-10
feat[next]: Support for direct field operator call with domain arg

tehrengruber:
Adds support for directly calling a field operator with a domain argument, which was previously only supported inside of a program. Many field operators in icon4py use the domain argument resulting in excessive amounts of boilerplate programs that can be removed now.
```python
@field_operator
def testee(inp: IField) -> IField:
    return inp

testee(inp, domain={IDim: (0, 10)})
```

Support in the dace backend is missing and will be added in a seperate PR.


src/gt4py/next/ffront/past_process_args.py
```diff
@@ -83,40 +83,46 @@ def _process_args(
...
+                index_type = ts.ScalarType(kind=ts.ScalarKind.INT32)
                 size_args.extend(
-                    shape if shape else [ts.ScalarType(kind=ts.ScalarKind.INT32)] * len(dims)  # type: ignore[arg-type]  # shape is always empty
+                    range_ if range_ else [ts.TupleType(types=[index_type, index_type])] * len(dims)  # type: ignore[arg-type]  # shape is always empty
                 )
     return tuple(rewritten_args), tuple(size_args), kwargs
 
 
-def _field_constituents_shape_and_dims(
+def _field_constituents_range_and_dims(
```
tehrengruber, line 107: This part of the code-base is an extremely poor state, let's not prettify a blobfish and invest time on improving this here.

havogt (changes requested):
Please split into 2 PRs.

havogt:
Is #1780 required for this to work, I still don't understand that part...

havogt (approved):

PR #1781, opened 2024-12-11
refact[next][dace]: split handling of let-statement lambdas from stencil body

edopao:
This is a refactoring of the code to lower lambda nodes: it splits the lowering of let-statements from the lowering of stencil expressions.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
@@ -30,7 +30,7 @@ from gt4py.next.program_processors.runners.dace_fieldview import (
     gtir_python_codegen,
     utility as dace_gtir_utils,
 )
-from gt4py.next.type_system import type_specifications as ts
+from gt4py.next.type_system import type_info as ti, type_specifications as ts
 
 
 if TYPE_CHECKING:
@@ -366,36 +366,36 @@ def translate_as_fieldop(
     """
     assert isinstance(node, gtir.FunCall)
     assert cpm.is_call_to(node.fun, "as_fieldop")
-    assert isinstance(node.type, ts.FieldType)
 
     fun_node = node.fun
     assert len(fun_node.args) == 2
-    stencil_expr, domain_expr = fun_node.args
+    fieldop_expr, domain_expr = fun_node.args
 
-    if isinstance(stencil_expr, gtir.Lambda):
-        # Default case, handled below: the argument expression is a lambda function
-        # representing the stencil operation to be computed over the field domain.
-        pass
-    elif cpm.is_ref_to(stencil_expr, "deref"):
+    assert isinstance(node.type, ts.FieldType)
+    if cpm.is_ref_to(fieldop_expr, "deref"):
         # Special usage of 'deref' as argument to fieldop expression, to pass a scalar
         # value to 'as_fieldop' function. It results in broadcasting the scalar value
         # over the field domain.
         stencil_expr = im.lambda_("a")(im.deref("a"))
-        stencil_expr.expr.type = node.type.dtype  # type: ignore[attr-defined]
+        stencil_expr.expr.type = node.type.dtype
+    elif isinstance(fieldop_expr, gtir.Lambda):
+        # Default case, handled below: the argument expression is a lambda function
+        # representing the stencil operation to be computed over the field domain.
+        stencil_expr = fieldop_expr
     else:
         raise NotImplementedError(
-            f"Expression type '{type(stencil_expr)}' not supported as argument to 'as_fieldop' node."
+            f"Expression type '{type(fieldop_expr)}' not supported as argument to 'as_fieldop' node."
         )
 
     # parse the domain of the field operator
     domain = extract_domain(domain_expr)
 
     # visit the list of arguments to be passed to the lambda expression
-    stencil_args = [_parse_fieldop_arg(arg, sdfg, state, sdfg_builder, domain) for arg in node.args]
+    fieldop_args = [_parse_fieldop_arg(arg, sdfg, state, sdfg_builder, domain) for arg in node.args]
 
     # represent the field operator as a mapped tasklet graph, which will range over the field domain
     taskgen = gtir_dataflow.LambdaToDataflow(sdfg, state, sdfg_builder)
-    input_edges, output_edge = taskgen.visit(stencil_expr, args=stencil_args)
+    input_edges, output_edge = taskgen.apply(stencil_expr, args=fieldop_args)
 
     return _create_field_operator(
         sdfg, state, domain, node.type, sdfg_builder, input_edges, output_edge
@@ -654,7 +654,7 @@ def translate_tuple_get(
 
     if not isinstance(node.args[0], gtir.Literal):
         raise ValueError("Tuple can only be subscripted with compile-time constants.")
-    assert node.args[0].type == dace_utils.as_itir_type(INDEX_DTYPE)
+    assert ti.is_integral(node.args[0].type)
     index = int(node.args[0].value)
 
     data_nodes = sdfg_builder.visit(
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
@@ -10,10 +10,22 @@ from __future__ import annotations
 
 import abc
 import dataclasses
-from typing import Any, Dict, Final, List, Optional, Protocol, Set, Tuple, TypeAlias, Union
+from typing import (
+    Any,
+    Dict,
+    Final,
+    List,
+    Optional,
+    Protocol,
+    Sequence,
+    Set,
+    Tuple,
+    TypeAlias,
+    Union,
+)
 
 import dace
-import dace.subsets as sbs
+from dace import subsets as sbs
 
 from gt4py import eve
 from gt4py.next import common as gtx_common
@@ -68,7 +80,7 @@ class MemletExpr:
 
     dc_node: dace.nodes.AccessNode
     gt_dtype: itir_ts.ListType | ts.ScalarType
-    subset: sbs.Indices | sbs.Range
+    subset: sbs.Range
 
 
 @dataclasses.dataclass(frozen=True)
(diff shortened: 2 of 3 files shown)


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -366,36 +366,36 @@ def translate_as_fieldop(
     """
     assert isinstance(node, gtir.FunCall)
     assert cpm.is_call_to(node.fun, "as_fieldop")
-    assert isinstance(node.type, ts.FieldType)
```
philip-paul-mueller, line 369: Why did you remove this `assert`?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -366,36 +366,36 @@ def translate_as_fieldop(
...
 
     fun_node = node.fun
     assert len(fun_node.args) == 2
-    stencil_expr, domain_expr = fun_node.args
+    fieldop_expr, domain_expr = fun_node.args
```
philip-paul-mueller, line 372: The only thing that you changed here is that you renamed `stencil_expr` to `fieldop_expr` and then later you essentially alias it back to to `stencil_expr`, but handle a specific case differently.

I could imagine, that it would look better and cleaner, if you would move the unpacking into the `if`.
I know it will leads to a little bit of code duplication, but I think that it is worth it, since you get rid of `fieldop_expr` which does not serve any purpose.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -10,10 +10,22 @@
...
+    Sequence,
+    Set,
+    Tuple,
+    TypeAlias,
+    Union,
+)
 
 import dace
-import dace.subsets as sbs
+from dace import subsets as sbs
```
philip-paul-mueller, line 28: I know in DaCe they use this expression.
However, in the transformation code I would import this as:
```python
from dace import subsets as dace_subsets
```
This makes it clear where it comes from, this is the same as used in [JaCe](https://github.com/GridTools/jace/blob/e1e46ee2cd91aaec6d8092eac7f91d6b405c25ef/CODING_GUIDELINES.md?plain=1#L32).
I would suggest you to also adopt that style for the following reasons:
- It would it make compatible with the style used in the transformations.
- It makes it clear where it comes from, I mean you import `gt4py.next.common` as `gtx_common` and this is from the same package.
- Except some DaCe experts nobody knows what `sbs` means.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -1264,39 +1276,38 @@ def visit_FunCall(self, node: gtir.FunCall) -> IteratorExpr | DataExpr:
         elif cpm.is_applied_shift(node):
             return self._visit_shift(node)
 
+        elif isinstance(node.fun, gtir.Lambda):
+            raise AssertionError("Lambda node should be visited with 'apply()' method.")
```
philip-paul-mueller, line 1280: I am actually not sure if you should do that, raise an `AssertionError`.
One reason is that the `-O` flag does not work, which is supposed to disable assertions.
However, this is more an input error, so I would either raise a `ValueError` or a `NotImplementedError`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -1309,3 +1320,38 @@ def visit_SymRef(self, node: gtir.SymRef) -> IteratorExpr | MemletExpr | SymbolE
         # if not in the lambda symbol map, this must be a symref to a builtin function
         assert param in gtir_python_codegen.MATH_BUILTINS_MAPPING
         return SymbolExpr(param, dace.string)
+
+    def apply(
```
philip-paul-mueller, line 1324: I don't remember that there is a call to `apply()`.
As far as I can tell `apply` is only called in this function, so it is a recursive function.
However, it does some highly non trivial pre and post processing.
So I think that:
- This function needs a better name, `apply` is just too generic.
- This function needs a doc string.
- This function needs comment that explain why the pre and post processing is needed and how it is done.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -1309,3 +1320,38 @@ def visit_SymRef(self, node: gtir.SymRef) -> IteratorExpr | MemletExpr | SymbolE
...
+        self,
+        node: gtir.Lambda,
+        args: Sequence[IteratorExpr | MemletExpr | SymbolExpr],
+    ) -> tuple[list[DataflowInputEdge], DataflowOutputEdge]:
+        # lambda arguments are mapped to symbols defined in lambda scope
+        prev_symbols: dict[
+            str,
+            Optional[IteratorExpr | MemletExpr | SymbolExpr],
```
philip-paul-mueller, line 1332: Why do you use `None` to indicate not present (I guess). I would think it is much more logical they would simply be not present.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -1309,3 +1320,38 @@ def visit_SymRef(self, node: gtir.SymRef) -> IteratorExpr | MemletExpr | SymbolE
...
+        ] = {}
+        for p, arg in zip(node.params, args, strict=True):
+            symbol_name = str(p.id)
+            prev_symbols[symbol_name] = self.symbol_map.get(symbol_name, None)
+            self.symbol_map[symbol_name] = arg
```
philip-paul-mueller, line 1337: This needs motivation, why you do this and why you ignore some parts.
I would think it should be simpler to just make a copy of `symbol_map` and then restore it.
I also would have guessed that you could turn it into a dict comprehension, which could be more efficient (maybe).

philip-paul-mueller (commented):
There are some aspects that needs some changes but it looks not bad.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I agree, it is simpler to make a copy of `symbol_map` and then restore it.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I changed based on the suggestion in next comment.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I would like to keep the name `apply` for the entry point of this visitor class, since it is consistent with other visitor classes in GT4Py. However, I agree on the rest so I will write some documentation.

philip-paul-mueller: I still think that `apply()` is the wrong name. If it is for compatibility, then why was it there before?
`_vistit_let()` should be the much better name.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -1309,3 +1324,45 @@ def visit_SymRef(self, node: gtir.SymRef) -> IteratorExpr | MemletExpr | SymbolE
...
+        before the stencil expression. In other words, the `Lambda` node representing
+        the stencil expression is always the innermost node.
+        Therefore, the lowering of let-statements results in recursive calls to
+        `apply()` until the stencil expression is found. At that point, it falls
+        back to the `visit()` function.
+        """
+
+        # lambda arguments are mapped to symbols defined in lambda scope.
+        prev_symbol_map = self.symbol_map
+        self.symbol_map = self.symbol_map.copy()
```
philip-paul-mueller, line 1352:
```suggestion
        prev_symbol_map = self.symbol_map.copy()
```

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -1326,32 +1330,39 @@ def apply(
...
             input_edges, output_edge = self.apply(let_node.fun, args=let_args)
-
         else:
-            # this lambda is not a let-statement, but a stencil expression
-            output_edge = self.visit_Lambda(node)
+            # this lambda node is not a let-statement, but a stencil expression
+            output_edge = self.visit(node)
             input_edges = self.input_edges
 
         # remove locally defined lambda symbols and restore previous symbols
```
philip-paul-mueller, line 1365:
```suggestion
        # restore previous symbols, thus removing locally defined lambda symbols in the let-scope
```

philip-paul-mueller (commented):
There are some small points, but there is nothing serious.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I was thinking to save the `self.symbol_map` object as `prev_symbol_map`, just to be on the safe side.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: I have now made a code change in this direction.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
philip-paul-mueller: In my view this is indicating that you have some big problems.

philip-paul-mueller (approved):
I still think that `apply()` is not needed.
It is _not_ an entry point because it was not needed before (as `visit()` was used).

edopao:
> I still think that `apply()` is not needed. It is _not_ an entry point because it was not needed before (as `visit()` was used).

That is true. I will make another change in this direction.

PR #1782, opened 2024-12-17
feat[dace][next]: Fixing strides in optimization

philip-paul-mueller:
Fixing some strides.
Mostly the work of Edoardo.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
@@ -460,7 +460,7 @@ def _gt_map_strides_into_nested_sdfg(
     inner_data: str,
     outer_subset: dace.subsets.Subset,
     outer_desc: dace_data.Data,
-    ignore_symbol_mapping: bool = True,
+    ignore_symbol_mapping: bool,
 ) -> None:
     """Modify the strides of `inner_data` inside `nsdfg_node` to match `outer_desc`.
 
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
@@ -223,7 +223,7 @@ def gt_map_strides_to_dst_nested_sdfg(
     state: dace.SDFGState,
     edge: dace.sdfg.graph.Edge,
     outer_node: dace.nodes.AccessNode,
-    ignore_symbol_mapping: bool = False,
+    ignore_symbol_mapping: bool = True,
     processed_nsdfgs: Optional[set[PropagatedStrideRecord]] = None,
 ) -> None:
     """Propagates the strides of `outer_node` along `edge` in the dataflow direction.
@@ -460,7 +460,7 @@ def _gt_map_strides_into_nested_sdfg(
     inner_data: str,
     outer_subset: dace.subsets.Subset,
     outer_desc: dace_data.Data,
-    ignore_symbol_mapping: bool = False,
+    ignore_symbol_mapping: bool = True,
 ) -> None:
     """Modify the strides of `inner_data` inside `nsdfg_node` to match `outer_desc`.
 
@@ -479,9 +479,12 @@ def _gt_map_strides_into_nested_sdfg(
         ignore_symbol_mapping: If possible the function will perform the renaming
             through the `symbol_mapping` of the nested SDFG. If `True` then
             the function will always perform the renaming.
+            Note that setting this value to `False` might have negative side effects.
 
     Todo:
         - Handle explicit dimensions of size 1.
+        - What should we do if the stride symbol is used somewhere else, creating an
+            alias is probably not the right thing?
     """
     # We need to compute the new strides. In the following we assume that the
     #  relative order of the dimensions does not change, but we support the case
@@ -526,7 +529,9 @@ def _gt_map_strides_into_nested_sdfg(
     #  The first is to create an alias in the `symbol_mapping`, however,
     #  this is only possible if the current strides are singular symbols,
     #  like `__a_strides_1`, but not expressions such as `horizontal_end - horizontal_start`
-    #  or literal values.
+    #  or literal values. Furthermore, this would change the meaning of the
+    #  old stride symbol in any context and not only in the one of the stride
+    #  of a single and isolated data descriptor.
     #  The second way would be to replace `strides` attribute of the
     #  inner data descriptor. In case the new stride consists of expressions
     #  such as `value1 - value2` we have to make them available inside the
@@ -579,11 +584,6 @@ def _gt_map_strides_into_nested_sdfg(
             assert sym.name in sdfg.symbols, f"Expected that '{sym}' is defined in the parent SDFG."
             nsdfg_node.sdfg.add_symbol(sym.name, sdfg.symbols[sym.name])
             nsdfg_node.symbol_mapping[sym.name] = sym
-
-        # Now create aliases for the old symbols that were used as strides.
-        for old_sym, new_sym in zip(inner_strides_init, new_strides):
-            if dace.symbolic.issymbolic(old_sym) and old_sym.is_symbol:
-                nsdfg_node.symbol_mapping[str(old_sym)] = dace.symbolic.pystr_to_symbolic(new_sym)
 
 
 def _gt_find_toplevel_data_accesses(
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
@@ -484,16 +484,7 @@ def _gt_map_strides_into_nested_sdfg(
 
     new_strides: list = []
     for dim_ostride, dim_oinflow in zip(outer_strides, outer_inflow, strict=True):
-        current_inner_dim = len(new_strides)
-
-        if inner_shape[current_inner_dim] == 1 and dim_oinflow == 1:
-            # There is an explicit size 1 dimension. Because the only valid
-            #  index for this dimension is `0` we can use any value here.
-            #  To give the compiler more information we explicitly use `0`,
-            #  instead of the outer value.
-            new_strides.append(0)
-
-        elif dim_oinflow == 1:
+        if dim_oinflow == 1:
             # Only something flows in, thus there is no stride in this dimension.
             pass
 
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
@@ -41,6 +41,11 @@ def gt_change_transient_strides(
     transients in the optimal way.
     The function should run after all maps have been created.
 
+    After the strides have been adjusted the function will also propagate
+    the strides into nested SDFG. This propagation will happen with
+    `ignore_symbol_mapping` set to `True`, see `gt_propagate_strides_of()`
+    for more.
+
     Args:
         sdfg: The SDFG to process.
         gpu: If the SDFG is supposed to run on the GPU.
@@ -123,13 +128,14 @@ def _gt_change_transient_strides_non_recursive_impl(
                 state=state,
                 outer_node=access_node,
                 processed_nsdfgs=processed_nsdfgs,
+                ignore_symbol_mapping=True,
             )
 
 
 def gt_propagate_strides_of(
     sdfg: dace.SDFG,
     data_name: str,
-    ignore_symbol_mapping: bool = False,
+    ignore_symbol_mapping: bool = True,
 ) -> None:
     """Propagates the strides of `data_name` within the whole SDFG.
 
@@ -140,7 +146,7 @@ def gt_propagate_strides_of(
     Args:
         sdfg: The SDFG on which we operate.
         data_name: Name of the data descriptor that should be handled.
-        ignore_symbol_mapping: If `False`, the default, try to modify the `symbol_mapping`
+        ignore_symbol_mapping: If `False` (default is `True`) try to modify the `symbol_mapping`
             of NestedSDFGs instead of manipulating the data descriptor.
     """
 
@@ -164,7 +170,7 @@ def gt_propagate_strides_from_access_node(
     sdfg: dace.SDFG,
     state: dace.SDFGState,
     outer_node: dace_nodes.AccessNode,
-    ignore_symbol_mapping: bool = False,
+    ignore_symbol_mapping: bool = True,
     processed_nsdfgs: Optional[set[PropagatedStrideRecord]] = None,
 ) -> None:
     """Propagates the stride of `outer_node` to any adjacent reachable through its edges.
@@ -184,7 +190,7 @@ def gt_propagate_strides_from_access_node(
         state: The state where the data node is used.
         edge: The edge that reads from the data node, the nested SDFG is expected as the destination.
         outer_node: The data node whose strides should be propagated.
-        ignore_symbol_mapping: If `False`, the default, try to modify the `symbol_mapping`
+        ignore_symbol_mapping: If `False` (default is `True`), try to modify the `symbol_mapping`
             of NestedSDFGs instead of manipulating the data descriptor.
         processed_nsdfgs: Set of NestedSDFG that were already processed and will be ignored.
             Only specify when you know what your are doing.
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
@@ -6,6 +6,7 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
+import warnings
 from typing import Optional, TypeAlias
 
 import dace
@@ -408,6 +409,7 @@ def _gt_map_strides_to_nested_sdfg_src_dst(
         # Now set the stride of the data descriptor inside the nested SDFG to
         #  the ones it has outside.
         _gt_map_strides_into_nested_sdfg(
+            sdfg=sdfg,
             nsdfg_node=nsdfg_node,
             inner_data=inner_data,
             outer_subset=get_subset(state, edge),
@@ -440,6 +442,7 @@ def _gt_map_strides_to_nested_sdfg_src_dst(
 
 
 def _gt_map_strides_into_nested_sdfg(
+    sdfg: dace.SDFG,
     nsdfg_node: dace.nodes.NestedSDFG,
     inner_data: str,
     outer_subset: dace.subsets.Subset,
@@ -453,6 +456,7 @@ def _gt_map_strides_into_nested_sdfg(
     match the ones of `outer_desc`.
 
     Args:
+        sdfg: The SDFG containing the NestedSDFG.
         nsdfg_node: The node in the parent SDFG that contains the NestedSDFG.
         inner_data: The name of the data descriptor that should be processed
             inside the NestedSDFG (by construction also a connector name).
@@ -539,7 +543,9 @@ def _gt_map_strides_into_nested_sdfg(
             if dace.symbolic.issymbolic(new_stride_dim):
                 new_strides_symbols.append(str(new_stride_dim))
             else:
-                new_strides_symbols.extend(sym for sym in new_stride_dim.free_symbols)
+                # NOTE: In DaCe `free_symbols` is `set[str]` but in `sympy` it
+                #   returns `set[symbol]`. We need `str` so we have to cast them.
+                new_strides_symbols.extend(str(sym) for sym in new_stride_dim.free_symbols)
 
         # Now we determine the set of symbols that should be mapped inside the NestedSDFG.
         #  We will exclude all that are already inside the `symbol_mapping` (we do not
@@ -551,9 +557,19 @@ def _gt_map_strides_into_nested_sdfg(
             for sym in new_strides_symbols
             if not (sym in nsdfg_node.sdfg.symbols or sym in nsdfg_node.symbol_mapping)
         }
+        # Now create the symbol we in the NestedSDFG.
         for sym in missing_symbol_mappings:
-            # We can not create symbols in the nested SDFG, because we do not have
-            #  the type of the symbols.
+            if sym in sdfg.symbols:
+                # TODO(phimuell): Handle the case the symbol is already defined.
+                nsdfg_node.sdfg.add_symbol(sym, sdfg.symbols[sym])
+            else:
+                # The symbol is not known in the parent SDFG, but we need a symbol
+                #  for it. So we use the default.
+                nsdfg_node.sdfg.add_symbol(sym, dace.symbol("__INVALID_SYMBOL__").dtype)
+                warnings.warn(
+                    f"Could not find the symbol '{sym}' in the parent SDFG while modifying the strides.",
+                    stacklevel=1,
+                )
             nsdfg_node.symbol_mapping[sym] = dace.symbolic.pystr_to_symbolic(sym)
 
         # Now create aliases for the old symbols that were used as strides.
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
@@ -575,8 +575,6 @@ def _gt_map_strides_into_nested_sdfg(
         # Now propagate the symbols from the parent SDFG to the NestedSDFG.
         for sym in missing_symbol_mappings:
             if sym in sdfg.symbols:
-                # TODO(phimuell): Handle the case the symbol is already defined in
-                #   the nested SDFG.
                 nsdfg_node.sdfg.add_symbol(sym.name, sdfg.symbols[sym.name])
             else:
                 # The symbol is not known in the parent SDFG, so we add it
@@ -589,7 +587,7 @@ def _gt_map_strides_into_nested_sdfg(
 
         # Now create aliases for the old symbols that were used as strides.
         for old_sym, new_sym in zip(inner_strides_init, new_strides):
-            if dace.symbolic.issymbolic(old_sym):
+            if dace.symbolic.issymbolic(old_sym) and old_sym.is_symbol:
                 nsdfg_node.symbol_mapping[str(old_sym)] = dace.symbolic.pystr_to_symbolic(new_sym)
 
 
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
@@ -39,8 +39,6 @@ def gt_change_transient_strides(
 
     Todo:
         - Implement the estimation correctly.
-        - Handle the case of nested SDFGs correctly; on the outside a transient,
-            but on the inside a non transient.
     """
     # TODO(phimeull): Implement this function correctly.
 
@@ -50,22 +48,32 @@ def gt_change_transient_strides(
         return sdfg
 
     for nsdfg in sdfg.all_sdfgs_recursive():
-        # TODO(phimuell): Handle the case when transient goes into nested SDFG
-        #   on the inside it is a non transient, so it is ignored.
         _gt_change_transient_strides_non_recursive_impl(nsdfg)
 
 
 def _gt_change_transient_strides_non_recursive_impl(
     sdfg: dace.SDFG,
 ) -> None:
-    """Essentially this function just changes the stride to FORTRAN order.
+    """Set optimal strides of all transients in the SDFG.
+
+    The function will look for all top level transients, see `_gt_find_toplevel_data_accesses()`
+    and set their strides such that the access is optimal, see Note. The function
+    will also run `gt_propagate_strides_of()` to propagate the strides into nested SDFGs.
+
+    This function should never be called directly but always through
+    `gt_change_transient_strides()`!
+
+    Note:
+        Currently the function just reverses the strides of the data descriptor
+        it processes. Since DaCe generates `C` order by default this lead to
+        FORTRAN order, which is (for now) sufficient to optimize the memory
+        layout to GPU.
 
     Todo:
         Make this function more intelligent to analyse the access pattern and then
         figuring out the best order.
     """
-
-    # NOTE: processing the transient here is enough. If we are inside a
+    # NOTE: Processing the transient here is enough. If we are inside a
     #   NestedSDFG then they were handled before on the level above us.
     top_level_transients_and_their_accesses = _gt_find_toplevel_data_accesses(
         sdfg=sdfg,
--- a/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
+++ b/src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
@@ -475,7 +475,9 @@ def _gt_find_toplevel_data_accesses(
                 # We also check if it was ever found on the top level, this should
                 #  not happen, as everything should go through Maps. But some strange
                 #  DaCe transformation might do it.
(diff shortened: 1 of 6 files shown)


src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -97,3 +107,125 @@ def _find_toplevel_transients(
...
+        return
+
+    outer_strides = outer_node.desc(sdfg).strides
+    _gt_map_strides_to_nested_sdfg(edge.dst, edge.dst_conn, edge.data, outer_strides)
+
+    for inner_state in edge.dst.sdfg.states():
+        for inner_node in inner_state.data_nodes():
+            if inner_node.data == edge.dst:
+                for inner_edge in inner_state.out_edges(inner_node):
+                    gt_map_strides_to_dst_nested_sdfg(sdfg, state, inner_edge, inner_node)
```
philip-paul-mueller, line 146: This might be a bit inefficient.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -97,3 +107,125 @@ def _find_toplevel_transients(
...
+
+
+def gt_map_strides_to_src_nested_sdfg(
```
philip-paul-mueller, line 149: Should  be merged with `dst` version.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -97,3 +107,125 @@ def _find_toplevel_transients(
...
+) -> None:
+    # We need to propagate the strides inside the nested SDFG on the global arrays
+    new_strides = tuple(
+        stride
+        for stride, to_map_size in zip(
+            outer_strides,
+            edge_data.subset.size(),
+            strict=True,
+        )
+        if to_map_size != 1
```
philip-paul-mueller, line 203: @edopao why do we exclude these dimensions?

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -97,3 +107,125 @@ def _find_toplevel_transients(
...
+        new_strides_symbols: list[dace.symbol] = functools.reduce(
+            lambda acc, itm: (acc + list(itm.free_symbols))  # type: ignore[union-attr]
+            if dace.symbolic.issymbolic(itm)
+            else acc,
+            new_strides,
+            [],
+        )
+        new_strides_free_symbols = {
+            sym for sym in new_strides_symbols if sym.name not in nsdfg_node.sdfg.symbols
+        }
```
philip-paul-mueller, line 228: Look at this more.

philip-paul-mueller (commented):
First round.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
edopao: No, actually there was a reason. The reason is that currently the scan is writing into a 1D array along the vertical dimension, so the memlet from the nested SDFG implementing the scan is implicitly un-slicing the data into the outer field (which is usually multidimensional). So that code is there to handle array slicing through memlet subset. Question is whether we want to support it, or not since it will be deprecated in next dace release.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
-    edge_data: dace.Memlet,
-    outer_strides: Iterable[int | dace.symbolic.SymExpr],
+    outer_subset: dace.subsets.Subset,
+    outer_desc: dace_data.Data,
     ignore_symbol_mapping: bool = False,
 ) -> None:
-    """
+    """Modify the strides of `inner_data` inside `nsdfg_node` to match `outer_desc`.
+
+    `inner_data` is the name of of a data descriptor inside the NestedSDFG.
```
edopao, line 447:
```suggestion
    `inner_data` is the name of a data descriptor inside the NestedSDFG.
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+    The function will then modify the modify the strides of `inner_data` to
```
edopao, line 448:
```suggestion
    The function will then modify the strides of `inner_data`, assuming this is an array, to
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
-        for stride, to_map_size in zip(
-            outer_strides,
-            edge_data.subset.size(),
-            strict=True,
-        )
-        if to_map_size != 1
-    )
-    inner_desc = nsdfg_node.sdfg.arrays[inner_data]
+    # We need to compute the new strides. In the following we assume that the
+    #  relative order of the dimension does not change, but some dimensions
```
edopao, line 466:
```suggestion
    #  relative order of the dimensions does not change, but we support the case where some dimensions
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+    #  that are present on the outside are not present on the inside. For
```
edopao, line 467:
```suggestion
    #  of the outer data descriptor are not present on the inside. For
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+    #  example this happens for the Memlet `a[__i0, 0:__a_size1]`.
+    #  We detect this case by checking if that dimension has size 1.
```
edopao, line 469:
```suggestion
    #  We detect this case by checking if the memlet subset in that dimension has size 1.
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+    # TODO(phimuell): Handle the case were some additional size 1 dimensions are added.
+    inner_desc: dace_data.Data = nsdfg_node.sdfg.arrays[inner_data]
+    inner_shape = inner_desc.shape
+    inner_strides_init = inner_desc.strides
+
+    # TODO(phimuell): For now this is fine, but it should be possisble to allow it.
```
edopao, line 475: You are right, we do not need to impose this limitation. You can remove the assert.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+
+        if inner_shape[current_inner_dim] == 1 and dim_oinflow == 1:
+            # There is an explicit size 1 dimension. Because the only valid
+            #  index for this dimension is `0` we can use any value here.
+            #  To give the compiler more information we explicitly use `0`,
+            #  instead of the outer value.
+            new_strides.append(0)
+
+        elif dim_oinflow == 1:
+            # Only something flows in, thus there is no stride in this dimension.
```
edopao, line 493: I do not understand the comment. Suggestion:
```suggestion
            # This is the case of implicit slicing along one dimension. The inner array descriptor has shape != 1 in `current_inner_dim`, which has to map to a subsequent dimension of `outer_inflow`.
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+        assert len(new_strides) <= len(inner_shape)
+
+    if len(new_strides) != len(inner_shape):
+        raise ValueError("Failed to compute the inner strides.")
+
+    # If we have a scalar on the inside, then there is nothing to adjust.
+    #  We could have performed the test above, but doing it here, gives us
+    #  the chance of validating it.
+    if isinstance(inner_desc, dace_data.Scalar):
+        if len(new_strides) != 0:
```
edopao, line 509: I have second thoughts on this check. How can it happen that `len(new_strides) == 0` ?
By looking at the code above, and knowing that a scalar has shape `(1,)`, I would expect that the access `inner_shape[current_inner_dim]` result on OOB access, because new_strides is `[0]`.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+        raise TypeError(
+            f"Expected that '{inner_data}' is an 'Array' but it is '{type(inner_desc).__name__}'."
+        )
+
+    # Now we actually replace the strides, there are two ways of doing it.
+    #  The first is to create an alias in the `symbol_mapping`, however,
+    #  this is only possible if the current strides are singular symbols,
+    #  like `__a_strides_1`, but not expressions such as `horizontal_end - horizontal_start`
+    #  or literal values.
+    #  The second way would be to replace `strides` attributer of the
```
edopao, line 523:
```suggestion
    #  The second way would be to replace `strides` attribute of the
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
-            if dace.symbolic.issymbolic(itm)
-            else acc,
-            new_strides,
-            [],
-        )
-        new_strides_free_symbols = {
-            sym for sym in new_strides_symbols if sym.name not in nsdfg_node.sdfg.symbols
+        # Now find the free symbols that the new strides need.
+        new_strides_symbols: list[str] = []
+        for new_stride_dim in new_strides:
```
edopao, line 544: Shouldn't it be like this?
```
        for new_stride_dim in new_strides:
            if dace.symbolic.issymbolic(new_stride_dim):
                new_strides_symbols.extend(sym for sym in new_stride_dim.free_symbols)
```

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+            if dace.symbolic.issymbolic(new_stride_dim):
+                new_strides_symbols.append(str(new_stride_dim))
+            else:
+                new_strides_symbols.extend(sym for sym in new_stride_dim.free_symbols)
+
+        # Now we determine the set of symbols that should be mapped inside the NestedSDFG.
+        #  We will exclude all that are already inside the `symbol_mapping` (we do not
```
edopao, line 551: I would prefer to have a separate check to ensure that the symbols already inside `symbol_mapping` map to equivalent expressions.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+        #  check if they map to the same value, we just hope it). Furthermore,
+        #  we will exclude all symbols that are listed in the `symbols` property
+        #  of the SDFG that is nested, and hope that it has the same meaning.
+        missing_symbol_mappings: set[str] = {
+            sym
```
edopao, line 556: `new_stride_dim.free_symbols` returns symbols, not strings: something is strange here, since you no longer need to call `sym.name`.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+            for sym in new_strides_symbols
+            if not (sym in nsdfg_node.sdfg.symbols or sym in nsdfg_node.symbol_mapping)
         }
-        for sym in new_strides_free_symbols:
-            nsdfg_node.sdfg.add_symbol(sym.name, sym.dtype)
-            nsdfg_node.symbol_mapping[sym.name] = sym
+        for sym in missing_symbol_mappings:
+            # We can not create symbols in the nested SDFG, because we do not have
+            #  the type of the symbols.
+            nsdfg_node.symbol_mapping[sym] = dace.symbolic.pystr_to_symbolic(sym)
```
edopao, line 563: I think it is better (more clear and safer) to store symbols as symbols, not as strings. That way, you can also avoid to call `pystr_to_symbolic()` here and below.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -426,59 +435,137 @@ def next_edges_by_connector(
...
+
+        # Now create aliases for the old symbols that were used as strides.
```
edopao, line 565: It should be safe to remove them. There is a `used_symbols()` function, if I remember right. So we could assert that they are not used.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +61,507 @@ def gt_change_transient_strides(
...
 
+        # Now we have to propagate the changed strides. Because we already have
+        #  collected all the AccessNodes we are using the
+        #  `gt_propagate_strides_from_access_node()` function, but we have to
+        #  create `processed_nsdfg` set already outside here.
+        #  Furthermore, the same comment as above apply, we do not have to
+        #  propagate the non-transients, because they either come from outside,
+        #  or they were already handled in the levels above, where they were
+        #  defined and then propagated down.
+        # TODO(phimuell): Updated the functions such that only once scan is needed.
```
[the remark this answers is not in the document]
philip-paul-mueller: I agree it is a bit strangely written, hope the new version is a bit better.

The main point I wanted to make is, that currently we scan the SDFG multiple times.
What we should do instead is to scan the SDFG once, including the nested ones and then reuse this information.
This will be a problem when we have bigger SDFGs.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +61,507 @@ def gt_change_transient_strides(
...
+
+def gt_map_strides_to_dst_nested_sdfg(
+    sdfg: dace.SDFG,
+    state: dace.SDFGState,
+    edge: dace.sdfg.graph.Edge,
+    outer_node: dace.nodes.AccessNode,
+    ignore_symbol_mapping: bool = False,
+    processed_nsdfgs: Optional[set[PropagatedStrideRecord]] = None,
+) -> None:
+    """Propagates the strides of `outer_node` along `edge` along the dataflow.
```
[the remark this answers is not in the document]
philip-paul-mueller: I think your proposal is a bit too long for the one liner.
But I see that we have to explain this a bit better, so I updated the description, what are you thinking.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +61,507 @@ def gt_change_transient_strides(
...
+
+    For more information see the description of `_gt_map_strides_to_nested_sdfg_src_dst().
+    However it is recommended to use `gt_propagate_strides_of()` directly.
```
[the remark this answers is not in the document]
philip-paul-mueller: I updated the description.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +61,507 @@ def gt_change_transient_strides(
...
+
+def gt_map_strides_to_src_nested_sdfg(
+    sdfg: dace.SDFG,
+    state: dace.SDFGState,
+    edge: dace.sdfg.graph.Edge,
+    outer_node: dace.nodes.AccessNode,
+    ignore_symbol_mapping: bool = False,
+    processed_nsdfgs: Optional[set[PropagatedStrideRecord]] = None,
+) -> None:
+    """Propagates the strides of `outer_node` along `edge` against the dataflow.
```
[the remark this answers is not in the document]
philip-paul-mueller: See the `dst` version.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +61,507 @@ def gt_change_transient_strides(
...
+        #  more uniform processing.
+        # TODO(phimuell): Instead of scanning every level for every data we modify
+        #   we should scan the whole SDFG once and then reuse this information.
+        accesses_in_nested_sdfg = _gt_find_toplevel_data_accesses(
+            sdfg=nsdfg_node.sdfg,
+            only_transients=False,  # Because on the nested levels they are globals.
+            only_arrays=True,
+        )
+        for nested_state, nested_access in accesses_in_nested_sdfg.get(inner_data, list()):
+            # We have to use `gt_propagate_strides_of()` here because we have to
```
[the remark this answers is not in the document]
philip-paul-mueller: You see `gt_propagate_strides_from_access_node()` was once called that.
But then I realized that the name was not matching the functionality well, so I changed it, but forgot to update the name.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +61,507 @@ def gt_change_transient_strides(
...
+
+            elif data in top_level_data:
+                # The data is already known to be in top level data, so we must add the
+                #  AccessNode to the list of known nodes. But nothing else.
+                top_level_data[data].append((state, dnode))
                 continue
+
             elif gtx_transformations.util.is_view(dnode, sdfg):
+                # The AccessNode refers to a View so we ignore it anyway
+                # TODO(phimuell/edopao): Should the function return them?
```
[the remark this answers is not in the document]
philip-paul-mueller: You are right.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
philip-paul-mueller: You are right, the reason is that this is buggy.
The scan tests were failing locally, I have essentially rewritten this function.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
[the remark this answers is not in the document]
edopao: Thanks, it makes sense now. I was thinking at the GTIR scan, that is why it didn't make sense :)

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
philip-paul-mueller: You are completely right.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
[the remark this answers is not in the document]
edopao: A view node has its own "virtual" data. By that, I mean that the view has a name, like an array. So I do not expect to encounter view nodes for a given data name that corresponds to a nestedSDFG connector, or?

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
philip-paul-mueller: I cast them to `str` when I created the `new_strides_symbols`.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
philip-paul-mueller: I agree.
However, you could apply this check only to the things that is defined in the symbol mapping and there it might be hard to do in general.
Furthermore, you could have interstate assignments.

I added a Todo, if you have concrete ideas let me know.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
philip-paul-mueller: Actually, this is a big mess in general.
However, if you look at `dace.sdfg.nodes.NestedSDFG::symbol_mapping` you will see that this is a `dict` that maps `str` to `dace.symbolic.simbol`.

The same holds for `dace.sdfg.sdfg.SDFG::symbols` which is `dict[str, typeclass]`.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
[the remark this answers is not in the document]
philip-paul-mueller: I mean this situation:
```
(A[10, 10]) ----- [:, 5] -----> (v[10])
```
i.e. `v` is a view of the fifth column.
The strides of `A` are `[__A_stride0, __A_stride1]` and the strides of `v` is `[__v_stride0]`, which has to be the same as `__A_stride0`, to make sense.
Therefore, `v[i]` is the same as `*(v + i * __A_stride0)`.

When I now change the strides of `A` from `[__A_stride0, __A_stride1]` to `[__A_stride1, __A_stride0]` then this change has to be reflected in the strides of `v`.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
philip-paul-mueller: I think I will keep it.
Using `used_symbols()` on the SDFG, will be very expensive as the whole thing has to be scanned.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -224,8 +223,13 @@ def gt_map_strides_to_dst_nested_sdfg(
...
     """Propagates the strides of `outer_node` along `edge` along the dataflow.
 
-    For more information see the description of `_gt_map_strides_to_nested_sdfg_src_dst().
-    However it is recommended to use `gt_propagate_strides_of()` directly.
+    In this context "along the dataflow" means that `edge` is an outgoing
+    edge of `outer_node` and the strides are into all NestedSDFGs that
+    are downstream of `outer_node`.
+
+    Except in certain cases this function should not be used directly. It is
+    instead recommended to use `gt_propagate_strides_of()`, which propagates
```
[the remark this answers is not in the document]
philip-paul-mueller: It is not the access node version, because that thing does not handle all AN in the SDFG.
For a (stupid, aka. myself) I want that `gt_propagate_strides_of()` is used that acts on the whole SDFG.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -258,8 +263,13 @@ def gt_map_strides_to_src_nested_sdfg(
...
     """Propagates the strides of `outer_node` along `edge` against the dataflow.
 
-    For more information see the description of `_gt_map_strides_to_nested_sdfg_src_dst().
-    However it is recommended to use `gt_propagate_strides_of()` directly.
+    In this context "along the dataflow" means that `edge` is an incoming
+    edge of `outer_node` and the strides are into all NestedSDFGs that
+    are upstream of `outer_node`.
+
+    Except in certain cases this function should not be used directly. It is
+    instead recommended to use `gt_propagate_strides_of()`, which propagates
```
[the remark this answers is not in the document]
philip-paul-mueller: see above.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -540,31 +545,49 @@ def _gt_map_strides_into_nested_sdfg(
...
+        #  Note that usually `free_symbols` returns `set[str]`, but here, because
+        #  we fall back on SymPy, we get back symbols. We will keep them, because
+        #  then we can use them to extract the type form them, which we need later.
+        new_strides_symbols: list[dace.symbol] = []
         for new_stride_dim in new_strides:
             if dace.symbolic.issymbolic(new_stride_dim):
-                new_strides_symbols.append(str(new_stride_dim))
-            else:
                 new_strides_symbols.extend(sym for sym in new_stride_dim.free_symbols)
+            else:
```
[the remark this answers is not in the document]
philip-paul-mueller: After some thinking probably not.
It could be that because some things are handled less strict, it is allowed (if we have multiple levels of nesting), but I will put an assert there.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -24,6 +40,11 @@ def gt_change_transient_strides(
     transients in the optimal way.
     The function should run after all maps have been created.
 
+    After the strides have been adjusted the function will also propagate
+    the strides into nested SDFG. This propagation will happen with
+    `ignore_symbol_mapping` set to `True`, see `gt_propagate_strides_of()`
```
edopao, line 45: I am of the opinion that `ignore_symbol_mapping` should be `False`. Actually, I would even prefer that this parameter was not available as configuration option. If the original SDFG was using symbol `A` for the stride of the inner array, the transformed SDFG should still use `A`, only the symbol mapping should change. Otherwise, if we replace symbols in nested SDFGs, we are doing two things in one transformation: stride propagation and symbol propagation.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +65,606 @@ def gt_change_transient_strides(
...
+        #  defined and then propagated down.
+        # TODO(phimuell): Updated the functions such that only one scan is needed.
+        processed_nsdfgs: set[dace_nodes.NestedSDFG] = set()
+        for state, access_node in accesses:
+            gt_propagate_strides_from_access_node(
+                sdfg=sdfg,
+                state=state,
+                outer_node=access_node,
+                processed_nsdfgs=processed_nsdfgs,
+                ignore_symbol_mapping=True,
```
edopao, line 130: See comment above.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +65,606 @@ def gt_change_transient_strides(
...
+            )
+
+
+def gt_propagate_strides_of(
+    sdfg: dace.SDFG,
+    data_name: str,
+    ignore_symbol_mapping: bool = True,
```
edopao, line 137: See comment above.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +65,606 @@ def gt_change_transient_strides(
...
+        )
+
 
-def _find_toplevel_transients(
+def gt_map_strides_to_dst_nested_sdfg(
     sdfg: dace.SDFG,
+    state: dace.SDFGState,
+    edge: dace.sdfg.graph.Edge,
+    outer_node: dace.nodes.AccessNode,
+    ignore_symbol_mapping: bool = False,
```
edopao, line 226: Strange to see different default values for `ignore_symbol_mapping`. In any case, as commented above, I am of the opinion that this should not be configurable and symbol mapping should always be used.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +65,606 @@ def gt_change_transient_strides(
...
+        ignore_symbol_mapping=ignore_symbol_mapping,
+    )
+
+
+def gt_map_strides_to_src_nested_sdfg(
+    sdfg: dace.SDFG,
+    state: dace.SDFGState,
+    edge: dace.sdfg.graph.Edge,
+    outer_node: dace.nodes.AccessNode,
+    ignore_symbol_mapping: bool = False,
```
edopao, line 266: See comment above.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
```diff
@@ -46,54 +65,606 @@ def gt_change_transient_strides(
...
+            )
+
+
+def _gt_map_strides_into_nested_sdfg(
+    sdfg: dace.SDFG,
+    nsdfg_node: dace.nodes.NestedSDFG,
+    inner_data: str,
+    outer_subset: dace.subsets.Subset,
+    outer_desc: dace_data.Data,
+    ignore_symbol_mapping: bool = False,
```
edopao, line 463: Here it is OK to leave the parameter, since this is a low-level private method. However, I would prefer to remove the default value.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
philip-paul-mueller: I have now thought a bit more about it and came to the conclusion that in certain situation `ignore_symbol_mapping=False` is actually wrong and will produce invalid results.
The reason is that using the symbol mapping actually affects the whole (nested) SDFG and not only the data descriptor.
The problem only occurs if different data descriptor use the same symbols as stride, which happens for example for temporaries, I mean for a long time they had the size `(horizontal_end, vertical_end)` which which would result in the strides `(horizontal_end, 1)`, which all transient shares.
This is the situation that `tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_strides.py::test_strides_propagation_shared_symbols_sdfg` demonstrates.
For some reasons the strides of only one descriptors (`a1`) are propagated, the other remains the same. If the symbol mapping would be used, then we would also change, as a side effect, the strides of `b2` (the inner view).

Earlier I have removed the symbol aliasing code, `2b03bb4`, in the `ignore_symbol_mapping=True` branch, exactly of that reason.
I also do not think that this is a big problem, as strides are not directly accessed, the shape is much more likely to be used, which does not change.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
philip-paul-mueller: I also saw them, it was an error on my side.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
philip-paul-mueller: I agree.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
edopao: You are right. Based on your explanation, I now agree on using `ignore_symbol_mapping=True` in its current meaning.

However, as a suggestion, we could change the meaning of `ignore_symbol_mapping`:
- `ignore_symbol_mapping=True` would behave as of current implementation: takes the outer strides and just apply it on the inner arrays.
- `ignore_symbol_mapping=False` would create new symbols in the nested SDFG (using the utility function to generate stride symbols based on the data name, and this should be enough to ensure that they are unique) and apply symbol mapping to the outer strides. The benefit of this is to avoid the risk of using a symbol that is already defined in the nested SDFG, with a different value than in parent SDFG.

edopao (approved):
The PR looks good, thank you for improving the fix. You can do as you prefer with my suggestion.

src/gt4py/next/program_processors/runners/dace_fieldview/transformations/strides.py
philip-paul-mueller: This is an issue indeed, I added a todo.

PR #1783, opened 2024-12-18
build: test python 3.12, 3.13

havogt:


havogt:
Will try again after we dropped 3.8, 3.9 as the frozen dependencies are too old for newer Python versions (no available binary distributions). Also for python 3.13, no numpy < 2 binary is available, therefore we should probably wait for https://github.com/spcl/dace/pull/1613.

PR #1786, opened 2024-12-30
feat[next]: Add support for more datatypes

SF-N:
This builds on [PR#1708](https://github.com/GridTools/gt4py/pull/1708) without the `float16` and `bfloat16` changes.

Add support for `int8, uin8, int16, uint16, uint32` and `uint64`.

Move builtin definitions from `src/gt4py/next/iterator/ir.py` to `src/gt4py/next/iterator/builtins.py`.

Use ascending integer values in `ScalarKind`-Enum and modify tests respectively.

Set `start: int = 1` in `tests/next_tests/integration_tests/cases.py` to not start initialization from zero as this has the same value as zero-initialized memory and modify tests respectively.

### DIFF
--- a/src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
+++ b/src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
@@ -417,7 +417,6 @@ class DaCeIRBuilder(eve.NodeTranslator):
         global_ctx: DaCeIRBuilder.GlobalContext,
         iteration_ctx: DaCeIRBuilder.IterationContext,
         symbol_collector: DaCeIRBuilder.SymbolCollector,
-        loop_order,
         k_interval,
         **kwargs: Any,
     ):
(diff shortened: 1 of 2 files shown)


tests/next_tests/integration_tests/cases.py
```diff
@@ -192,13 +192,13 @@ class UniqueInitializer(DataInitializer):
     data containers.
     """
 
-    start: int = 0
+    start: int = 1
```
havogt, line 195: I might have added a test that relies on start being zero, which is a mistake. Let me know if you see a problem.

tehrengruber (approved):
Please double check the changes I pushed.

SF-N:
> Please double check the changes I pushed.

I checked them and they look good to me. `test_arithmetic_and_logical_builtins` is failing for `(not_, [[True, False]], [False, True])` and I am not sure why.

PR #1787, opened 2025-01-06
ci: fix boost install in cartesian and daily ci plan

havogt:
Boost download link expired, but actually no custom boost (header) installation is required.


havogt:
daily plan is tested here https://github.com/GridTools/gt4py/actions/runs/12628593244

tehrengruber (approved):

PR #1788, opened 2025-01-06
bug[next]: Fix propagated symbols order stability across runs

tehrengruber:
If statements in FOAST have a special `propagated_symbols` annex attributed that contains all symbols that are defined and hence available outside of the if. The compuation of these symbols is based on a set operation whose order is not stable across runs. This PR fixes that by sorting them. Since behavior changes across runs are hard to test and we don't have any facilities for that no test is provided.

### DIFF
--- a/src/gt4py/next/ffront/field_operator_ast.py
+++ b/src/gt4py/next/ffront/field_operator_ast.py
@@ -180,7 +180,7 @@ class IfStmt(Stmt):
     @datamodels.root_validator
     @classmethod
     def _collect_common_symbols(cls: type[IfStmt], instance: IfStmt) -> None:
-        common_symbol_names = (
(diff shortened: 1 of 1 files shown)


src/gt4py/next/ffront/field_operator_ast.py
```diff
@@ -180,7 +180,7 @@ class IfStmt(Stmt):
     @datamodels.root_validator
     @classmethod
     def _collect_common_symbols(cls: type[IfStmt], instance: IfStmt) -> None:
-        common_symbol_names = (
+        common_symbol_names = sorted(  # sort is required to get stable results across runs
```
havogt, line 183: why are they not always in the same order?

havogt (approved):

PR #1790, opened 2025-01-08
feat[next][dace]: iterator-view support to DaCe backend

edopao:
The lowering of scan to SDFG requires the support for iterator view. This PR introduces a subset of iterator features:
- Local `if_` with exclusive branch execution
- Lowering of `list_get`, `make_tuple` and `tuple_get` in iterator view
- Field operators returning a tuple of fields
- Tuple of fields with different size

Iterator tests are enabled on dace CPU backend without SDFG transformations (`auto_optimize=False`).


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -190,8 +196,10 @@ class EmptyInputEdge(DataflowInputEdge):
     state: dace.SDFGState
     node: dace.nodes.Tasklet
 
-    def connect(self, me: dace.nodes.MapEntry) -> None:
-        self.state.add_nedge(me, self.node, dace.Memlet())
+    def connect(self, map_entry: Optional[dace.nodes.MapEntry]) -> None:
+        # the empty edge is not created if the dataflow is instatiated without a map
```
philip-paul-mueller, line 200: I am curious why.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -200,38 +208,49 @@ class DataflowOutputEdge:
...
 
     def connect(
         self,
-        mx: dace.nodes.MapExit,
+        map_exit: Optional[dace.nodes.MapExit],
         dest: dace.nodes.AccessNode,
         subset: dace_subsets.Range,
     ) -> None:
         # retrieve the node which writes the result
         last_node = self.state.in_edges(self.result.dc_node)[0].src
```
philip-paul-mueller, line 229: You should move this definition of `last_node` into the `if` branch, because the `else` branch redefines it.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -545,6 +568,257 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+            input_memlets["__cond"] = condition_value
+
+        for if_branch_state, arg in zip([tstate, fstate], node.args[1:3]):
+            # visit each if-branch in the corresponding state of the nested SDFG
+            in_edges, out_edge = self._visit_if_branch(nsdfg, if_branch_state, arg, input_memlets)
+            for edge in in_edges:
+                edge.connect(map_entry=None)
+
+            # the result of each branch needs to be moved to the parent SDFG
+            def construct_output(
```
philip-paul-mueller, line 728: I would not define this function _inside_ a loop, you should declare it at the beginning of this function.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -545,6 +568,257 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+            # because not all tuple fields are necessarily used in the lambda scope
+            for data_node in if_branch_state.data_nodes():
+                data_desc = data_node.desc(nsdfg)
+                if (not data_desc.transient) and (if_branch_state.degree(data_node) == 0):
+                    # isolated node, connect it to a transient to avoid SDFG validation errors
+                    temp, temp_desc = nsdfg.add_temp_transient_like(data_desc)
+                    temp_node = if_branch_state.add_access(temp)
+                    if_branch_state.add_nedge(
+                        data_node, temp_node, dace.Memlet.from_array(temp, temp_desc)
+                    )
```
philip-paul-mueller, line 773: - Why does this only affect transients?
- Why not simply delete the input nodes, I mean they do not serve any purpose?

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -545,6 +568,257 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+            else:
+                self._add_edge(
+                    input_expr.dc_node,
+                    None,
+                    nsdfg_node,
+                    inner,
+                    self.sdfg.make_array_memlet(input_expr.dc_node.data),
+                )
+
+        def connect_output(inner_value: ValueExpr) -> ValueExpr:
```
philip-paul-mueller, line 800: You should move this function declaration to the beginning of this function.

philip-paul-mueller (commented):
It looks good, there are some issues I see, but it looks good.

edopao (commented):
Thank you for the review comments, I tried to address most of them.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -545,6 +568,257 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
+                arg_node = arg.field
+                arg_desc = arg_node.desc(self.sdfg)
+                arg_expr = MemletExpr(
+                    arg_node, arg.gt_dtype, dace_subsets.Range.from_array(arg_desc)
+                )
+
+            arg_data = arg_node.data
+            # SDFG data containers with name prefix '__tmp' are expected to be transients
+            inner_data = (
+                arg_data.replace("__tmp", "__input", 1)
```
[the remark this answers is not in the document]
philip-paul-mueller: This requirement does not come from the SDFG parts of DaCe, it is because we use the frontend in very special ways and the frontend makes this assumption.
In my view it is either:
- A bug in the frontend (because it does not make sure that they are actually temporaries).
- It is undefined behaviour, because the documentation _clearly_ states that every identifier starting with two underscore might have a special meaning to DaCe (this is also the case in the `C` standard), so we have to use different names.

In addition you should also add a comment that explains why you do it.
However, I still thinks that the best and most clean solution would be to not use `sdfg.temp_data_name()` at all but use a wrapper, such that we do not have to it, because, we do not know when the same thing happens again.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
philip-paul-mueller: Yes you are right, I did not saw that.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -652,6 +926,55 @@ def _visit_neighbors(self, node: gtir.FunCall) -> ValueExpr:
...
+        list_arg = self.visit(node.args[1])
+        assert isinstance(list_arg, ValueExpr)
+        assert isinstance(list_arg.gt_dtype, ts.ListType)
+        assert isinstance(list_arg.gt_dtype.element_type, ts.ScalarType)
+
+        list_desc = list_arg.dc_node.desc(self.sdfg)
+        assert len(list_desc.shape) == 1
+
+        result_dtype = dace_utils.as_dace_type(list_arg.gt_dtype.element_type)
+        result, _ = self.sdfg.add_scalar(self.sdfg.temp_data_name(), result_dtype, transient=True)
```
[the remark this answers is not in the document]
philip-paul-mueller: See my comment above.
In summary: it is centralized, but we have, in certain case at certain places, to modify the names to avoid some specific behaviour in DaCe.
What we should do instead is generate names from the beginning that we will not run into that issue.
Because now, every time you essentially call `clone()` on a data descriptor, and changes its `transient` state, you have to remember and do the change.
Furthermore, currently it is only one pattern `__tmp` but there is no guarantee that it will remain that way.

philip-paul-mueller (commented):
There is the aspect of naming variables, i.e. the centralized naming stuff, I am not really happy.
The result looks good.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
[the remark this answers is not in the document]
edopao: Fixed. You were right, the `__tmp` prefix in the parameter name can be avoided in the first place.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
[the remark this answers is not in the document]
philip-paul-mueller: You should also add an `add_temp_array()` method, it is wired if it is missing.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -53,6 +53,13 @@ def unique_map_name(self, name: str) -> str: ...
     @abc.abstractmethod
     def unique_tasklet_name(self, name: str) -> str: ...
 
+    def add_temp_scalar(
+        self, sdfg: dace.SDFG, dtype: dace.dtypes.typeclass
+    ) -> tuple[str, dace.data.Scalar]:
+        """Add a temporary scalar to the SDFG."""
+        temp_name = sdfg.temp_data_name()
```
philip-paul-mueller, line 60: - To fully avoid the renaming problem, you should perform here the renaming, i.e. `__tmp` -> `__gtit_tmp`, also add a note to why this is done.
- It looks rather strange that there is no `add_temp_array()` because they suffer from the same problem. Furthermore, it is just strange if that method is missing.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -594,72 +658,88 @@ def _visit_if_branch(
...
+        return input_edges, output_edges
 
     def _visit_if(self, node: gtir.FunCall) -> ValueExpr | tuple[ValueExpr | tuple[Any, ...], ...]:
         """
         Lowers an if-expression with exclusive branch execution into a nested SDFG, in which
         each branch is lowered into a dataflow in a separate state and the if-condition is represented
         as the inter-state edge condtion.
         """
+
+        def visit_if_branch_result(
```
philip-paul-mueller, line 702: It is a bit strange that `visit_if_branch_result()` is an inner function but `_visit_if_branch_arg()` is not.
This is a bit inconsistent.

I see it captures `nsdfg`, I am against this capturing so I would add it as a note or add it as argument, as it is done for `state`.

philip-paul-mueller (commented):
I have two final remarks about, most importantly I think that an `add_temp_array()` function is missing and we should no longer generate `__tmp` names, but turn them differently.
Then there is something that I think is a bit inconsistent.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
```diff
@@ -296,7 +299,14 @@ def get_tuple_type(
...
     return ts.TupleType(
-        types=[get_tuple_type(d) if isinstance(d, tuple) else d.gt_dtype for d in data]
+        types=[
+            get_tuple_type(d)
+            if isinstance(d, tuple)
+            else d.get_field_type()
+            if isinstance(d, IteratorExpr)
+            else d.gt_dtype
+            for d in data
+        ]
```
philip-paul-mueller, line 309: I highly recommend to turn it into a `if-elif-else` into a `for` loop.

philip-paul-mueller (approved):
LGTM

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_dataflow.py
edopao: Fixed, thank you.

Issue #1754, opened 2024-11-27
[cartesian] K-offset writes fail on GPU


FlorianDeconinck:
See failure on `gt:gpu` here; https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4455690602105886/4525297225819146/-/jobs/8381258008

Translate tests is `test_K_offset_write_conditional`

Temporary fix to deactivate the feature TBD

FlorianDeconinck:
PR for temp fix: https://github.com/GridTools/gt4py/pull/1755

edopao:
I suspect that the same issue appears in test_K_offset_write, see this CI job on a commit on top of #1755: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4455690602105886/4525297225819146/-/jobs/8498719906

FlorianDeconinck:
Sorry @edopao that's a mistake on my end fixing that asap

FlorianDeconinck:
Fixing the `utest` in https://github.com/GridTools/gt4py/pull/1791

Underlyign problem is that the `inteval(1, None)` should have failed hard on argument validation

romanc:
Note: this is missing the [gt4py.cartesian](https://github.com/GridTools/gt4py/issues?q=is%3Aissue%20state%3Aopen%20%20-label%3Agt4py.next%20AND%20NOT%20%22next%22%20label%3Agt4py.cartesian) label.

Issue #1684, opened 2024-10-09
[cartesian]: Mitigation for K offset write


FlorianDeconinck:
The new write offset in K  means we can go out of bounds without the system being able to warn about it since extent calculation won't be able to cover those access.

Think about the degenerate case of

```python
index = some_bad_math_that_gives_back_2349082093840
field[0, 0, value] = value
```

This was an iron rule for GT4Py that we now allow to be broken so we can match patterns seen regularly in physics.

To mitigate the issue, we should introduce on `DEBUG` a bound check, drop before every K offset write which should error out cleanly.

----

- [ ] On `DEBUG` generate a check for bounds _before_ K offset write

romanc:
Note: this is missing the [gt4py.cartesian](https://github.com/GridTools/gt4py/issues?q=is%3Aissue%20state%3Aopen%20%20-label%3Agt4py.next%20AND%20NOT%20%22next%22%20label%3Agt4py.cartesian) label.

PR #1791, opened 2025-01-08
fix[cartesian]: Race condition in unit test for K write

FlorianDeconinck:
The `interval` analysis in the unit test `test_K_offset_write_conditional` fails to catch a mistake in the code that leads to a race condition.

Work:
- Fix the bad interval
- Remove not needed restriction on CUDA version

Further work to fix the underlying problem and the larger issue of bound check on variable indexing is covered  [here](https://github.com/GridTools/gt4py/issues/1684) and [there](https://github.com/GridTools/gt4py/issues/1754)

### DIFF
--- a/tests/cartesian_tests/integration_tests/multi_feature_tests/test_code_generation.py
+++ b/tests/cartesian_tests/integration_tests/multi_feature_tests/test_code_generation.py
@@ -582,17 +582,6 @@ def test_K_offset_write(backend):
     # Cuda generates bad code for the K offset
     if backend == "cuda":
         pytest.skip("cuda K-offset write generates bad code")
-    if backend in ["dace:gpu"]:
-        import cupy as cp
-
-        if cp.cuda.runtime.runtimeGetVersion() < 12000:
-            pytest.skip(
-                f"{backend} backend with CUDA 11 and/or GCC 10.3 is not capable of K offset write, update CUDA/GCC if possible"
-            )
-    if backend in ["gt:gpu"]:
-        pytest.skip(
-            f"{backend} backend is not capable of K offset write, bug remains unsolved: https://github.com/GridTools/gt4py/issues/1754"
-        )
 
     arraylib = get_array_library(backend)
     array_shape = (1, 1, 4)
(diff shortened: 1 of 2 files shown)


havogt:
I am not any more familiar enough with cartesian. What's the expected behavior? Does the user need to protect the oob in the vertical interval or should gt4py take care but it's buggy?
Was the problem that we are reading `B` oob?

tests/cartesian_tests/integration_tests/multi_feature_tests/test_code_generation.py
```diff
@@ -700,6 +678,42 @@ def column_physics_conditional(A: Field[np.float64], B: Field[np.float64], scala
...
+    # - lev = 1
+    # - A[2] == 42 && B[2] == 1 => True
+    # - A[3] = -1
+    # - B[2] = -1
+    # - lev = 2
+    # - A[2] == 42 && B[2] == -1 => False
+    # End of iteration state
+    # - A[...] = A[40, 2.0, 2.0, -1]
+    # - B[...] = A[1, 1, -1, 42]
+    # ITERATION k = 1 of [3:1]
```
havogt, line 701:
```suggestion
    # ITERATION k = 1 of [2:1]
```

FlorianDeconinck:
> I am not any more familiar enough with cartesian. What's the expected behavior? Does the user need to protect the oob in the vertical interval or should gt4py take care but it's buggy?

Sorry I put this PR in a rush I should have explained a bit better.

This PR is aimed at stabilizing the CI first and foremost. The issue is dual:
- On vertical access with scalar offset, gt4py should catch the wrong interval and yell. This is buggy and/or not extended to writes. 
- On vertical access with variable, there's no way we can check, so a system of bound check on debug should be put in place to mitigate (logged as [ticket 1683](https://github.com/GridTools/gt4py/issues/1684))

Now of course this is part of the larger discussion on physics and there's a definite need for us to sit down and probably talk it through as I am very well aware that we are stepping beyond the bounds of the original design of cartesian.

But we are really responding to the need of the algorithm and our current wider strategy is to _not_ rewrite the algorithm

havogt (approved):

PR #1794, opened 2025-01-10
feat[next]: Only inline scalars outside of stencils

tehrengruber:
The `InlineScalar` pass previously inlined all scalar expressions which is not needed and unnecessarily increases the tree size. Instead we only inline them in field view now.


tehrengruber:
Failing tests are just due to too old compiler version in the CSCS CI. Problem disappears as soon as we switch to alps. On hold until then.

havogt:
cscs-ci run

havogt (approved):

PR #1797, opened 2025-01-14
fix[next][dace]: remove unused connectivities

edopao:
By design, the arrays for connectivity tables are initially created as transient and marked as non-transient during the lowering when they are used. At the end of lowering to SDFG, the unused connectivities (still transient arrays) should be removed.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace_common/utility.py
+++ b/src/gt4py/next/program_processors/runners/dace_common/utility.py
@@ -18,12 +18,13 @@ from gt4py.next.iterator import ir as gtir
 from gt4py.next.type_system import type_specifications as ts
 
 
-# regex to match the symbols for field shape and strides
-FIELD_SYMBOL_RE: Final[re.Pattern] = re.compile(r"__.+_(size|stride)_\d+")
-
-
 # arrays for connectivity tables use the following prefix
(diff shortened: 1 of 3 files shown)


src/gt4py/next/program_processors/runners/dace_common/utility.py
```diff
@@ -48,7 +52,14 @@ def as_itir_type(dtype: dace.typeclass) -> ts.ScalarType:
 
 
 def connectivity_identifier(name: str) -> str:
-    return f"connectivity_{name}"
+    return f"{CONNECTIVITY_INDENTIFIER_PREFIX}{name}"
+
+
+def is_connectivity(name: str, datadesc: dace.data.Data) -> bool:
+    is_array = isinstance(datadesc, dace.data.Array) and (
+        not isinstance(datadesc, (dace.data.ArrayReference, dace.data.ArrayView))
```
philip-paul-mueller, line 60: It should be suffice to check if it is an array only, because:
- only the true connectives, i.e. the arrays, will have the correct name; I would even say that creating a view or a reference with the name of a connectivity should be considered an error.
- I think we should not use reference at all, so we should not check for them either.

philip-paul-mueller (approved):
I have one remark concerning the test `is_array`, which is too broad in my view.
But this is not a blocking issue, just something you should consider again.

PR #1801, opened 2025-01-16
ci: disable x86-targets (DaintXC) in CSCS CI-Ext

edopao:
Disable x86-targets (DaintXC) in CSCS CI-Ext, since DaintXC has been decommissioned.
We only disable but still keep the x86 configuration in case in future we need to set it up on a Alps vCluster.

### DIFF
--- a/ci/cscs-ci.yml
+++ b/ci/cscs-ci.yml
@@ -51,6 +51,7 @@ stages:
     CUPY_VERSION: 13.3.0
     UBUNTU_VERSION: 22.04
 
+# prefix the x86_64 config with dot in order to not execute it
 .build_py311_baseimage_x86_64:
   extends: .build_baseimage_x86_64
   variables:
@@ -60,6 +61,7 @@ build_py311_baseimage_aarch64:
   variables:
     <<: *py311
 
+# prefix the x86_64 config with dot in order to not execute it
 .build_py310_baseimage_x86_64:
   extends: .build_baseimage_x86_64
   variables:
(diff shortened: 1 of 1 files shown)


ci/cscs-ci.yml
```diff
@@ -51,7 +51,7 @@ stages:
     CUPY_VERSION: 13.3.0
     UBUNTU_VERSION: 22.04
 
-build_py311_baseimage_x86_64:
+.build_py311_baseimage_x86_64:
   extends: .build_baseimage_x86_64
   variables:
     <<: *py311
```
havogt, line 57: Does the `.` mean, it is not executed, but you could derive from it? Maybe for clarity, commenting it out would be better.

edopao: Yes, correct. I'll add a comment.

ci/cscs-ci.yml
```diff
@@ -51,7 +51,8 @@ stages:
     CUPY_VERSION: 13.3.0
     UBUNTU_VERSION: 22.04
 
-build_py311_baseimage_x86_64:
+# prefix the x86_64 config with dot in order to not execute it
+.build_py311_baseimage_x86_64:
   extends: .build_baseimage_x86_64
   variables:
     <<: *py311
```
havogt, line 58:
```suggestion
#build_py311_baseimage_x86_64:
#  extends: .build_baseimage_x86_64
#  variables:
#    <<: *py311
```
Actually, I meant that. Any, reason not to do like that? But not super important.

edopao: I agree, more clear. Fixed.

havogt (approved):

PR #1802, opened 2025-01-16
fix[next][dace]: Fixing Strides Reconstruction During Propagation

philip-paul-mueller:
NestedSDFG essentially allows to perform some slices, there are technically three chases:
- The data container on the inside has a smaller rank than the one on the outside, thus some dimensions were removed.
- The data container on the inside has the same rank than the one on the outside.
- The data container on the inside has a larger rank than the one on the outside, thus some dimensions were added.

The last case is not handled, as it does not happens in GT4Py. Before, the first and second case were handled together, but it was realized that the second case was not implemented properly and it was added explicitly.

This PR fixes the issues with `TestFusedVelocityAdvectionStencil1To7` and `TestFusedVelocityAdvectionStencil8To13` in [ICON4Py#638](https://github.com/C2SM/icon4py/pull/638), however, the later test now fails with a segmentation fault.


edopao (approved):
LGTM

PR #1805, opened 2025-01-20
style[cartesian]: Remove unused optional keyword arguments

romanc:
## Description

These two keyword arguments aren't used in the function. Since the two functions have a "catch-all" `*kwargs` at the end of the argument list, we can safely delete the unused keyword arguments.

## Requirements

`mypy` and/or existing tests would complain if these changes weren't appropriate
N/A

### DIFF
--- a/src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
+++ b/src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
@@ -417,7 +417,6 @@ class DaCeIRBuilder(eve.NodeTranslator):
         global_ctx: DaCeIRBuilder.GlobalContext,
         iteration_ctx: DaCeIRBuilder.IterationContext,
         symbol_collector: DaCeIRBuilder.SymbolCollector,
-        loop_order,
         k_interval,
         **kwargs: Any,
     ):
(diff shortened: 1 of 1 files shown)


FlorianDeconinck (approved):
LGTM.

I really don't like the `kwargs`, but it's used throughout

twicki (approved):
LGTM

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -417,7 +417,6 @@ def visit_HorizontalExecution(
         global_ctx: DaCeIRBuilder.GlobalContext,
         iteration_ctx: DaCeIRBuilder.IterationContext,
         symbol_collector: DaCeIRBuilder.SymbolCollector,
-        loop_order,
```
havogt, line 420: Shouldn't we remove it on the call site as well? https://github.com/GridTools/gt4py/blob/efad7267bd8f6005e1a657145d3b8c8e77903d69/src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py#L548

romanc: Indeed, no need to pass it down (from `visit_VerticalLoop`). It's never used below. Thanks for pointing this out.

PR #1713, opened 2024-10-25
[Draft] feat[next]: Gtir concat where

SF-N:

PR #1806, opened 2025-01-20
Gtir concat where 2

havogt:
Replaces #1713


tehrengruber:
Moved back to #1713 for simplicity.

PR #1808, opened 2025-01-20
refactor[next][dace]: normalize SDFG field type with local dimension

edopao:
Fields with a local dimension can be passed as program arguments. The corresponding `FieldType` parameter type in GTIR contains the local dimension in the list of field domain dimensions, while the data type of the field elements is `ScalarType`. For example:

`ts.FieldType(dims=[Vertex, V2EDim], dtype=FLOAT_TYPE)`

where:
```
V2EDim = gtx.Dimension("V2E", kind=gtx.DimensionKind.LOCAL)
FLOAT_TYPE = ts.ScalarType(kind=ts.ScalarKind.FLOAT64)
```

Except for the program arguments, the internal representation in the SDFG lowering should only contain global dimensions in the field domain, and use `ListType` for the element type in case of a list of values:

`ts.FieldType(dims=[Vertex], dtype=ts.ListType(element_type=FLOAT_TYPE, offset_type=V2EDim))`

With this PR, the normalized form is used across the SDFG lowering. The `make_field` helper method is modified to convert the type definition of field arguments to the normalized form.


src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
```diff
@@ -234,11 +234,50 @@ def get_offset_provider_type(self, offset: str) -> gtx_common.OffsetProviderType
...
+            # do nothing: the field domain consists of all global dimensions
+            field_type = data_type
+        elif len(local_dims) == 1:
+            local_dim = local_dims[0]
+            local_dim_index = data_type.dims.index(local_dim)
+            # the local dimension is converted into `ListType` data element
+            if not isinstance(data_type.dtype, ts.ScalarType):
+                raise ValueError(f"Invalid field type {data_type}.")
+            if local_dim_index != (len(data_type.dims) - 1):
+                raise ValueError(f"Invalid field domaoin with local dimension {data_type.dims}.")
```
philip-paul-mueller, line 269: There is a type in "domain", however, I think that the error message is not accurate as it could be.
I would write: "Invalid field domain: expected the local dimension at the end, but it was {local_dim_index}."

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
```diff
@@ -111,31 +111,10 @@ def get_local_view(
...
-                )
-
-            else:
-                raise ValueError(
-                    f"Unexpected data field {self.dc_node.data} with more than one local dimension."
-                )
+            assert all(dim != gtx_common.DimensionKind.LOCAL for dim in self.gt_type.dims)
+            return gtir_dataflow.IteratorExpr(
+                self.dc_node, self.gt_type.dtype, field_domain, it_indices
+            )
```
philip-paul-mueller, line 117: Why could you delete this code?
I guess that the current logic guarantees that code path will also hit `make_field` and the functionality is now there, is this correct?
Because from the changes it is not obvious to me.

philip-paul-mueller (commented):
LGTM.
I have one suggestion and one question but it looks good to me.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_builtin_translators.py
edopao: Correct. I can add a comment to explain that the assert condition is ensured by `make_field`.

src/gt4py/next/program_processors/runners/dace_fieldview/gtir_sdfg.py
edopao: Good idea.

philip-paul-mueller (approved):
LGTM

PR #1813, opened 2025-01-21
build: update infrastructure to use the uv tool

egparedes:
Modernize development infrastructure to use newer python tools. The `uv` (https://docs.astral.sh/uv/) tool is used as project management tool, `nox` (https://nox.thea.codes/) replaces `tox` as the project test runner, custom scripts and `cog` snippets are replaced by a central `dev-tasks.py` tool with multiple commands.

Additionally, documentation and CI workflows are updated accordingly and some tests which were not properly executed properly are fixed.

Other related tasks to enhance the development infrastructure which could be done after this PR is merged are being tracked in issue #1829


.github/workflows/daily-ci.yml
```diff
@@ -5,116 +5,92 @@ on:
...
   #   branches:
   #   - main
-  ## END
+  # # END
 
 jobs:
   daily-ci:
     strategy:
       matrix:
+        dependencies-factor: ["lowest-direct", "highest"]
```
havogt, line 19: these are uv builtin names? maybe add a comment, to make it easy to look up their meaning

.github/workflows/test-cartesian.yml
```diff
@@ -20,35 +20,36 @@ concurrency:
...
 jobs:
   test-cartesian:
-    runs-on: ubuntu-latest
     strategy:
       matrix:
+        codegen-factor: [internal, dace]
+        os: ["ubuntu-latest"]
         python-version: ["3.10", "3.11"]
-        tox-factor: [internal, dace]
+      fail-fast: false
```
havogt, line 28: What is it doing in this context and what it the reason? maybe document.

.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,44 @@
+name: "Test Jupyter Notebooks"
```
havogt, line 1: Maybe the filename could be `test-example-notebooks`? Will check what test_docs does...

noxfile.py
```diff
@@ -0,0 +1,247 @@
...
+        *session.posargs,
+    )
+    session.run(
+        *"pytest --doctest-modules --doctest-ignore-import-errors -sv".split(),
+        str(pathlib.Path("src") / "gt4py" / "cartesian"),
+    )
+
+
+@nox.session(python=["3.10", "3.11"])
+def test_docs(session: nox.Session) -> None:
```
havogt, line 105: I would not call it docs, because I connect `docs` with sphinx documentation, more than with executable notebooks.

.pre-commit-config.yaml
```diff
@@ -45,86 +48,26 @@ repos:
...
   rev: v0.8.6
-  ##[[[end]]]
   hooks:
-    # Run the linter.
-    # TODO: include tests here
   - id: ruff
     files: ^src/
     args: [--fix]
-    # Run the formatter.
+    # TODO: include `tests` here
```
havogt, line 56: highlighting a TODO

noxfile.py
```diff
@@ -0,0 +1,247 @@
...
+from typing import Final, Literal, TypeAlias
+
+import nox
+
+try:
+    import pytest
+
+    NO_TESTS_COLLECTED_EXIT_CODE = pytest.ExitCode.NO_TESTS_COLLECTED
+except ImportError:
+    NO_TESTS_COLLECTED_EXIT_CODE = 5
```
havogt, line 24: please document this

noxfile.py
```diff
@@ -0,0 +1,247 @@
...
+)
+CodeGenTestSettings: Final[dict[str, dict[str, Sequence]]] = {
+    "internal": {"extras": [], "markers": ["not requires_dace"]},
+    "dace": {"extras": ["dace"], "markers": ["requires_dace"]},
+}
+
+
+# -- nox sessions --
+@nox.session(python=["3.10", "3.11"], tags=["cartesian"])
+@nox.parametrize("device", [DeviceNoxParam.cpu, DeviceNoxParam.cuda12])
```
havogt, line 71: why only cuda12 for gpu backends? it's the only option that we use, but if someone would like to try rocm, it would have to be added here?

pyproject.toml
```diff
@@ -1,9 +1,64 @@
...
+  'sphinx-rtd-theme>=3.0.1',
+  'sphinx-toolbox>=3.8.1'
+]
+lint = [
+  'pre-commit>=4.0.1',
+  'ruff>=0.8.0',
+  'tach>=0.16.0'
+]
+test = [
+  'atlas4py>=0.35',
```
havogt, line 39: does it belong here?

pyproject.toml
```diff
@@ -1,9 +1,64 @@
...
+  'nbmake>=1.4.6',
+  'nox>=2024.10.9',
+  'pytest>=8.0.1',
+  'pytest-benchmark>=5.0.0',
+  'pytest-cache>=1.0',
+  'pytest-cov>=5.0.0',
+  'pytest-factoryboy>=2.6.1',
+  'pytest-instafail>=0.5.0',
+  'pytest-xdist[psutil]>=3.5.0',
+  'tox>=4.4.0'
```
havogt, line 50: tox still needed?

havogt (commented):
Looks good, just some questions for documentation.

.github/workflows/daily-ci.yml
egparedes: Done.

.pre-commit-config.yaml
egparedes: I would postpone it for a subsequent PR.

noxfile.py
egparedes: Simplified and documented.

pyproject.toml
egparedes: Nope, thanks.

.github/workflows/test-cartesian.yml
egparedes: `fail-fast: true`, which is the default, makes the action to stop all matrix runs immediately after the first failure, which is a sensible default to save CI minutes (for [reference](https://www.edwardthomson.com/blog/github_actions_6_fail_fast_matrix_workflows)). 

In this particular case, I added the `fail: fast` here just is also present in the other subpackages, although I can't remember why we set it up like this. Probably because since gt4py is open source and has plenty of free CI minutes, the cost is not a critical factor and it might be still useful to see if all combinations of parameters in the matrix fail or only some of them. But we could also remove this settings from all the workflows if it makes more sense....

.github/workflows/test-docs.yml
egparedes: Renamed to `test-examples`

noxfile.py
egparedes: Renamed to `test_examples`

havogt:
cscs-ci run

tach.toml
```diff
@@ -3,7 +3,7 @@ source_roots = [
     "src",
 ]
 exact = true
-forbid_circular_dependencies = true
+# forbid_circular_dependencies = true
```
havogt, line 6: why?

tests/conftest.py
```diff
@@ -8,5 +8,55 @@
...
+    session: pytest.Session, config: pytest.Config, items: list[pytest.Item]
+) -> None:
+    """Pytest hook to modify the collected test items.
+
+    See: https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_collection_modifyitems
+    """
+    for item in items:
+        # Visit the test item parents chain in reverse order, until we get to
+        # the module object where the test function (or class) has been defined.
+        # At that point, processthe custom package-level marks attribute if it
```
havogt, line 52:
```suggestion
        # At that point, process the custom package-level marks attribute if it
```

pyproject.toml
egparedes: It is not really a gt4py dependency but an external package which can be used by gt4py users. In gt4py, it is only used in tests.

noxfile.py
egparedes: The actual session function works for any `DeviceOption` value, including `rocm` devices. We can add any other `DeviceOption` value to the parametrization at any time, the only problem is that the full matrix of parametrization will be run if someone forgets to specify the correct device parameter in the test session, and we don't usually run the tests on machines with ROCm devices...

egparedes:
cscs-ci run

havogt (approved):
lgtm

egparedes:
cscs-ci run

noxfile.py
```diff
@@ -0,0 +1,247 @@
...
+) -> None:
+    """Install session packages using uv."""
+    session.run_install(
+        "uv",
+        "sync",
+        "--no-dev",
+        *(f"--extra={e}" for e in extras),
+        *(f"--group={g}" for g in groups),
+        env={key: value for key, value in os.environ.items()}
+        | {"UV_PROJECT_ENVIRONMENT": session.virtualenv.location},
```
havogt, line 238:
```suggestion
        | {"UV_PROJECT_ENVIRONMENT": session.virtualenv.location, "UV_PYTHON": session.python},
```

egparedes: Done in a different way with a command line argument.

PR #1799, opened 2025-01-15
fix[dace][next]: Fix for DistributedBufferRelocator

philip-paul-mueller:
This PR fixes an error that was reported by Edoardo (@edopao).
The bug was because the `DistributedBufferRelocator` transformation did not check if its insertion would create a read-write conflict. This commit adds such a check, that is, however, not very sophisticated and needs some improvements. However, the example /`model/atmosphere/dycore/tests/dycore_stencil_tests/test_compute_exner_from_rhotheta.py`) where it surfaced, does hold more challenges. The main purpose of this PR is to unblock further development in ICON4Py.

Link to ICON4Py PR: https://github.com/C2SM/icon4py/pull/638

PR #1814, opened 2025-01-22
fix[next][dace]: Fix for `DistributedBufferRelocator`

philip-paul-mueller:
This PR fixes a bug in `DistributedBufferRelocator` that was observed in ICON4Py's `TestUpdateThetaAndExner` test.

In essence there was an `assert` that assumed that checked if this temporary was a sink node.
but, the code that finds all write backs was never excluding such cases, i.e. the temporaries that were selected might not be sink nodes in the state where they are defined.
The `assert` was not part of the original implementation and is not a requirement of the transformation, instead it was introduced by [PR#1799](https://github.com/GridTools/gt4py/pull/1799), that fixed some issues in the analysis of read write dependencies.

There are two solutions for this, either removing the `assert` or prune these kinds of temporaries.
After some consideration, it was realized that handling such cases will not lead to invalid SDFG, as long as the other restrictions on the global data are respected. For that reason the `assert` was removed.
However, we should thinking of doing something more intelligent in that case.


edopao (approved):
I have tested it, it works.

philip-paul-mueller:
@edopao let's hope it stays that way.

PR #1818, opened 2025-01-22
feat[next][dace]: support for field origin in lowering to SDFG

edopao:
This PR adds support for GT4Py field arguments with non-zero start index, for example:

`inp = constructors.empty(common.domain({IDim: (1, 9)}), ...)`

which was supported in baseline only for temporary fields, by means of a data structure called `field_offsets`. This data structure is removed for two reasons:
1. the name "offset" is a left-over from previous design based on dace array offset
3. offset has a different meaning in GT4Py

We introduce the GT4Py concept of field origin and use it for both temporary fields and program arguments. The field origin corresponds to the start of the field domain range.

This PR also changes the symbolic definition of array shape. Before, the array shape was defined as `[data_size_0, data_size_1, ...]`, now the size corresponds to the range extent `stop - start` as `[(data_0_range_1 - data_0_range_0), (data_1_range_1 - data_1_range_0), ...]`.

The translation stage of the dace workflow is extended with an option `disable_field_origin_on_program_arguments` to set the field range start symbols to constant value zero. This is needed for the dace orchestration, because the signature of a dace-orchestrated program does not provide the domain origin.


src/gt4py/next/program_processors/runners/dace/sdfg_callable.py
```diff
@@ -142,18 +150,16 @@ def get_sdfg_args(
 
     dace_args = _get_args(sdfg, args)
     dace_field_args = {n: v for n, v in dace_args.items() if not np.isscalar(v)}
+    dace_field_strides = _get_stride_args(sdfg.arrays, dace_field_args)
     dace_conn_args = get_sdfg_conn_args(sdfg, offset_provider, on_gpu)
-    dace_shapes = _get_shape_args(sdfg.arrays, dace_field_args)
```
philip-paul-mueller, line 146: Just for my understanding.
The shapes are now included in the `dace_args` value?

philip-paul-mueller (commented):
It looks generally good, but there are a few things.

The most important thing is, that are sometimes a not defined origin is converted to all zero, sometimes it is kept as `None`.
I made several comments about that using one form, i.e. implicitly converting to all zero, would be more consistent and could simplify the code.
Although it would no longer be possible to distinguish between the case of the case "explicitly set to zero" and "not defined", but I am not sure if this would be an issue.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -279,7 +291,8 @@ def make_field(
             raise NotImplementedError(
                 "Fields with more than one local dimension are not supported."
             )
-        return gtir_builtin_translators.FieldopData(data_node, field_type, domain_offset)
+        field_origin = gtx_dace_utils.get_symbolic_origin(data_node.data, field_type)
```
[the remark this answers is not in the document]
edopao: `make_field` is actually called for `SymRefs` to global symbols, that is non-transient data.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -201,6 +199,24 @@ def _collect_symbols_in_domain_expressions(
     )
 
 
+def _get_field_doman_subset(
+    domain: gtir_builtin_translators.FieldopDomain, data: gtir_builtin_translators.FieldopData
+) -> dace.subsets.Range:
+    """Helper method to build a memlet subset of a field over the given domain."""
+    # convert domain expression to dictionary to ease access to dimension boundaries
+    domain_ranges = {dim: (lb, ub) for dim, lb, ub in domain}
+    if isinstance(data.gt_type, ts.FieldType) and len(data.gt_type.dims) != 0:
```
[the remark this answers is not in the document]
edopao: Correct.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -201,6 +199,24 @@ def _collect_symbols_in_domain_expressions(
...
+        origin = ([0] * len(domain)) if data.origin is None else data.origin
+        subset = ",".join(
+            f"{domain_ranges[dim][0] - origin}:{domain_ranges[dim][1] - origin}"
+            for dim, origin in zip(data.gt_type.dims, origin, strict=True)
+        )
```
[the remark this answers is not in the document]
edopao: Done.

src/gt4py/next/program_processors/runners/dace/utils.py
```diff
@@ -73,6 +73,27 @@ def field_stride_symbol_name(field_name: str, axis: int) -> str:
     return field_symbol_name(field_name, axis, "stride")
 
 
+def range_start_symbol(field_name: str, axis: int) -> str:
```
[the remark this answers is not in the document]
edopao: Done.

edopao (commented):
Thanks for the review.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -107,20 +130,64 @@ def get_local_view(
...
+            in the parent SDFG. This includes the range start and stop symbols (used
+            to calculate the array shape as range 'stop - start') and the strides.
+        """
+        if isinstance(self.gt_type, ts.ScalarType):
+            return {}
+        ndims = len(self.gt_type.dims)
+        outer_desc = self.dc_node.desc(sdfg)
+        assert isinstance(outer_desc, dace.data.Array)
+        if self.origin is None:
+            outer_field_origin = [0] * ndims
```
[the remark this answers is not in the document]
philip-paul-mueller: This is right, I did not consider this case.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -941,10 +901,47 @@ def visit_SymRef(
         return gtir_builtin_translators.translate_symbol_ref(node, sdfg, head_state, self)
 
 
+def _remove_field_origin_symbols(ir: gtir.Program, sdfg: dace.SDFG) -> None:
```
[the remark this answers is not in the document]
philip-paul-mueller: The second part is also wrong.
I wanted to write was "why do you _not_ do it for transients".
However, in another comment you wrote that only non transient data descriptor have this symbolic sizes.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -248,9 +248,10 @@ def make_field(
         self, data_node: dace.nodes.AccessNode, data_type: ts.FieldType | ts.ScalarType
     ) -> gtir_builtin_translators.FieldopData:
         """
-        Helper method to build the field data type associated with an access node in the SDFG.
+        Helper method to build the field data type associated with an access node
+        to non-transient data in the SDFG.
 
-        In case of `ScalarType` data, the descriptor is constructed with `offset=None`.
+        In case of `ScalarType` data, the descriptor is constructed with `origin=None`.
```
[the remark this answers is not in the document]
philip-paul-mueller: Could you also add a test that `data_node` is a transient.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -331,18 +341,25 @@ def _make_array_shape_and_strides(
...
+                shape.append(
+                    dace.symbolic.pystr_to_symbolic(neighbor_table_types[dim.value].max_neighbors)
+                )
+            elif gtx_dace_utils.is_connectivity_identifier(name, self.offset_provider_type):
+                # we use symbolic size for the global dimension of a connectivity
+                shape.append(
+                    dace.symbolic.pystr_to_symbolic(gtx_dace_utils.field_size_symbol_name(name, i))
+                )
+            else:
+                # the size of global dimensions for a regular field is the symbolic
```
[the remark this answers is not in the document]
philip-paul-mueller: You should add "[...] used for temporary field, the default of DaCe, which is row major, is used and might be changed during optimization".
I think this makes it a bit clearer that the lowering does not really care about the strides beside being that they are correct (but not necessarily optimal).

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -39,31 +39,28 @@
 
 
 def get_domain_indices(
-    dims: Sequence[gtx_common.Dimension], offsets: Optional[Sequence[dace.symbolic.SymExpr]] = None
+    dims: Sequence[gtx_common.Dimension], origin: Optional[Sequence[dace.symbolic.SymExpr]] = None
```
philip-paul-mueller, line 42: This is just a suggestion, but I would not give `origin` a default.
Having a default is great for stuff that is not so important, however, forcing the user (I am sorry that is you) to explicitly specify the `origin` every time will:
- Make aware of this important setting.
- It will not be forgotten.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -78,26 +75,59 @@ class FieldopData:
...
+        else:
+            outer, outer_desc = sdfg_builder.add_temp_array_like(outer_sdfg, inner_desc)
+            # We cannot use a copy of the inner data descriptor directly, we have to apply the symbol mapping.
+            dace.symbolic.safe_replace(
+                symbol_mapping,
+                lambda m: dace.sdfg.replace_properties_dict(outer_desc, m),
+            )
+            # Same applies to the symbols used as field origin (the domain range start)
+            assert self.origin is not None
+            outer_origin = [val.subs(symbol_mapping) for val in self.origin]
```
philip-paul-mueller, line 117: You are using `safe_replace` above, since it solves cases like `repl = {'a': 'b', 'b': 'a'}`, since you use here the same repl dict, you should either use `safe_replace` here too or you do not use `safe_replace()` above.
And I am not thinking that the second option is good.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -107,20 +137,58 @@ def get_local_view(
...
+                gtx_dace_utils.range_stop_symbol(dataname, i): (
+                    self.origin[i] + outer_desc.shape[i]
+                )
+                for i in range(ndims)
+            }
+            | {
+                gtx_dace_utils.field_stride_symbol_name(dataname, i): stride
+                for i, stride in enumerate(outer_desc.strides)
+            }
+        )
```
philip-paul-mueller, line 190: I would combine this as a series of calls to `update()`, i.e. something like
```python
symbol_mapping = {gtx_dace_utils.range_start_symbol(dataname, i): self.origin[i] for i in range(ndims)}
symbol_mapping.update(...)
symbol_mapping.update(...)
return symbol_mapping
```
it is in my view much cleaner.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -835,12 +837,19 @@ def write_output_of_nested_sdfg_to_temporary(inner_value: ValueExpr) -> ValueExp
 
         outputs = {outval.dc_node.data for outval in gtx_utils.flatten_nested_tuple((result,))}
 
+        if nsdfg_symbols_mapping is None:
+            # `None` means that all free symbols are mapped to the symbols available in parent SDFG
```
philip-paul-mueller, line 841:
```suggestion
            # `None` means that all free symbols are mapped to the symbols available in parent SDFG by the `add_nested_sdfg()` function.
```
This case also means that we never need to do a remapping where the names inside and outside are different.

I would actually kill this `if` because as far as I can see `nsdfg_symbol_mapping` is either `None` or `{"__cond": ...}`.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -835,12 +837,19 @@ def write_output_of_nested_sdfg_to_temporary(inner_value: ValueExpr) -> ValueExp
...
+            pass
+        else:
```
philip-paul-mueller, line 843: I do not really understand this case, is it really needed?

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
```diff
@@ -626,46 +631,23 @@ def translate_scan(
...
-            str(nested_symbol): parent_symbol
-            for nested_symbol, parent_symbol in zip(
-                [*inner_desc.shape, *inner_desc.strides],
-                [*outer_desc.shape, *outer_desc.strides],
-                strict=True,
-            )
-            if dace.symbolic.issymbolic(nested_symbol)
-        }
+    for psym, arg in lambda_args_mapping:
+        nsdfg_symbols_mapping |= gtir_translators.get_arg_symbol_mapping(psym.id, arg, sdfg)
```
philip-paul-mueller, line 637: This is just my paranoia, for me there is a potential sever error.
Could it be that one argument needs the mapping `{'x': 'y'}` while another argument needs `{'x': 'z'}`, or is this not possible or is implementing the check not worth it?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -321,28 +335,46 @@ def unique_tasklet_name(self, name: str) -> str:
 
     def _make_array_shape_and_strides(
         self, name: str, dims: Sequence[gtx_common.Dimension]
-    ) -> tuple[list[dace.symbol], list[dace.symbol]]:
+    ) -> tuple[list[dace.symbolic.SymExpr], list[dace.symbolic.SymExpr]]:
```
philip-paul-mueller, line 338:
```suggestion
    ) -> tuple[list[dace.symbolic.SymbolicType], list[dace.symbolic.SymbolicType]]:
```

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -888,33 +862,34 @@ def construct_output_for_nested_sdfg(
...
                 # Non-transient nodes are just input nodes that are immediately returned
                 # by the lambda expression. Therefore, these nodes are already available
                 # in the parent context and can be directly accessed there.
-                outer_data = lambda_arg_nodes[inner_data.dc_node.data]
+                outer_data = lambda_arg_nodes[inner_dataname]
             else:
-                outer_node = head_state.add_access(inner_data.dc_node.data)
-                outer_data = inner_data.make_copy(outer_node)
+                # This must be a symbol captured from the lambda parent scope.
+                outer_node = head_state.add_access(inner_dataname)
```
philip-paul-mueller, line 889: If I am not mistaken, then this implicitly assumes that the data container on the inside and the outside always have the same name.
If the above is true, then I am not sure if this correct all the time.
If the above statement is false, then what does it means.

Furthermore, what do you mean with the symbol?
How can a symbol be an output?
I am sure I miss something here.

philip-paul-mueller (commented):
There are some points that needs some further work, but nothing serious.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -78,8 +75,8 @@ class FieldopData:
     Args:
         dc_node: DaCe access node to the data storage.
         gt_type: GT4Py type definition, which includes the field domain information.
-        origin: List of start indices, in each dimension, when the dimension range
```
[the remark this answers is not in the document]
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -78,8 +75,8 @@ class FieldopData:
...
-            does not start from zero; assume zero, if origin is not set.
+        origin: List of start indices, in each dimension, for `FieldType` data.
+            Set to `None` only for `ScalarType` data.
```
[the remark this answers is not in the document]
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
edopao: Good point.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
edopao: However, `safe_replace()` seems to work with in-place replacement.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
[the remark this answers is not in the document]
philip-paul-mueller: This sounds like a good idea.
If not yet done, could you add a note to keep track of this.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -76,13 +76,22 @@ class FieldopData:
         dc_node: DaCe access node to the data storage.
         gt_type: GT4Py type definition, which includes the field domain information.
         origin: List of start indices, in each dimension, for `FieldType` data.
-            Set to `None` only for `ScalarType` data.
+            Has to be `None` only for `ScalarType` data. For fields it is assumed
+            to be all zeros if not given.
```
philip-paul-mueller, line 80: This is not fully correct with the check below.
Assume you have a field type and pass `None` as `origin`.
Then `self.origin` will be set to `None` and the test in the `__post_init__()` function will fail.
But the doc string implies that it is silently converted to all zero.

Furthermore, something that I have not yet realized.
The class is frozen, so code such as:
```python
self.origin = 1
```
is invalid, but
```
self.origin.clear()
```
is valid.
Thus you should change the type to `Optional[tuple[dace.symbolic.SymbolicType, ...]]`.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -113,8 +122,7 @@ def map_to_parent_sdfg(
                 lambda m: dace.sdfg.replace_properties_dict(outer_desc, m),
             )
             # Same applies to the symbols used as field origin (the domain range start)
-            assert self.origin is not None
-            outer_origin = [val.subs(symbol_mapping) for val in self.origin]
+            outer_origin = [val.subs(symbol_mapping) for val in self.origin]  # type: ignore[union-attr]  # origin is checked at construction
```
philip-paul-mueller, line 125: Please either remove the usage of `safe_replace` above (which I do not recommend) or use it here too.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
philip-paul-mueller: Bump.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
philip-paul-mueller: Bump

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
philip-paul-mueller: Bump

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
philip-paul-mueller: Bump

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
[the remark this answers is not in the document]
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
edopao: Makes sense, I will simplify this code. However, I have to deal with the mapping of the domain symbols:
`ValueError: Missing symbols on nested SDFG "if_stmt_0": ['i_KDim_gtx_vertical', 'i_IDim_gtx_horizontal']`

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
edopao: That check should in the IR. I assume the IR is correct and the parameter list contains unique names.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: This is the case where a lambda returns a tuple of fields, and one of the field is just a data container that exists in the parent SDFG.

I am thinking if the lowering logic would be simpler if a dataflow, implementing an IR expression like `make_tuple`, would never return an input global symbol (this case). This could be achieved by making a copy to a transient, whenever a global container is referenced. This approach would also move the double buffering on the inputs, rather then the outputs.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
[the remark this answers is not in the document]
edopao: Done

edopao (commented):
I apologize that I didn't address some of your last comments, I simply missed it.

edopao:
This PR now includes #1836 to avoid test failures with `ConstantPropagation` in `gt_substitute_compiletime_symbols()`. PR #1836 is to be merged first, indeed.

src/gt4py/next/program_processors/runners/dace/utils.py
```diff
@@ -100,3 +100,26 @@ def filter_connectivity_types(
...
+    in the original exoression.
+
+    Args:
+        val: The symbolic expression where to apply the replacement.
+        symbol_mapping: The mapping table for symbol replacement.
+
+    Returns:
+        A new symbolic expression as result of symbol replacement.
+    """
+    x = [val]
```
philip-paul-mueller, line 123:
```suggestion
    # This is needed because `subs()` returns a new object and can not handle replacement dicts of the form `{'x': 'y', 'y': 'x'}`.
    #  Better approaches are welcomed.  
    x = [val]
```

philip-paul-mueller (approved):

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -75,21 +75,20 @@ class FieldopData:
...
-            Has to be `None` only for `ScalarType` data. For fields it is assumed
-            to be all zeros if not given.
+        origin: Tuple of start indices, in each dimension, for `FieldType` data.
+            Default to empty tuple for `ScalarType` data.
     """
 
     dc_node: dace.nodes.AccessNode
     gt_type: ts.FieldType | ts.ScalarType
-    origin: Optional[list[dace.symbolic.SymExpr]]
+    origin: tuple[dace.symbolic.SymbolicType, ...] = tuple()
```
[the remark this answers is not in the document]
edopao: No. The empty tuple means empty list of dimensions. This is only the case of `ScalarType` (or zero-dimensional fields, but this is a very corner case).

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -122,10 +121,12 @@ def map_to_parent_sdfg(
                 lambda m: dace.sdfg.replace_properties_dict(outer_desc, m),
             )
             # Same applies to the symbols used as field origin (the domain range start)
-            outer_origin = [val.subs(symbol_mapping) for val in self.origin]  # type: ignore[union-attr]  # origin is checked at construction
+            outer_origin = [
+                gtx_dace_utils.safe_replace_symbolic(val, symbol_mapping) for val in self.origin
+            ]
```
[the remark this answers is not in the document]
edopao: This solution was your idea, and I really like it too. Thank you.

src/gt4py/next/program_processors/runners/dace/utils.py
edopao: Right, I forgot to explain this evil code.

PR #1819, opened 2025-01-23
refactor[next]: neg builtin for unary minus

SF-N:
Add `neg` builtin to GTIR and use it for unary minus, e.g. `-val`, instead of `0-val` as before.


src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -227,6 +227,7 @@ def astype(
     "floor": math.floor,
     "ceil": math.ceil,
     "trunc": math.trunc,
+    "neg": np.negative,
```
SF-N, line 230: There seems to be no unary minus in `math`.

tehrengruber: It's in the operator module `operator`. But it also belongs to `UNARY_MATH_NUMBER_BUILTIN`s is it accepts int and not just booleans.

src/gt4py/next/iterator/builtins.py
```diff
@@ -420,6 +425,7 @@ def bool(*args):  # noqa: A001 [builtin-variable-shadowing]
     "floor",
     "ceil",
     "trunc",
```
tehrengruber, line 427: As above belongs to UNARY_MATH_NUMBER_BUILTINS.

src/gt4py/next/iterator/embedded.py
```diff
@@ -541,7 +548,7 @@ def promote_scalars(val: CompositeOfScalarOrField):
     }
     decorator = getattr(builtins, math_builtin_name).register(EMBEDDED)
     impl: Callable
-    if math_builtin_name in ["gamma", "not_"]:
```
tehrengruber, line 544: Can be put in the list above. The operator impl also works for columns out-of-the-box.

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -76,6 +76,7 @@ class GTFNCodegen(codegen.TemplatedGenerator):
         "xor_": "std::bit_xor{}",
         "mod": "std::modulus{}",
         "not_": "std::logical_not{}",
+        "neg": "std::negate{}",
```
tehrengruber, line 79: I would place this side-by-side to abs.

src/gt4py/next/program_processors/runners/dace/gtir_python_codegen.py
```diff
@@ -70,6 +70,7 @@
     "xor_": "({} ^ {})",
     "mod": "({} % {})",
     "not_": "(not {})",  # ~ is not bitwise in numpy
+    "neg": "(- {})",
```
tehrengruber, line 73: Same as above.

tehrengruber (changes requested):

tehrengruber (approved):

PR #1825, opened 2025-01-24
feat[next][dace]: extend support for sparse fields

edopao:
Add lowering support for deref and cartesian shift of a sparse field. Unstructured shift of a sparse field (that is calling `neighbors` on a sparse field) is not supported yet.


edopao:
The IR for expressions on sparse fields might change in near future. Besides, these expressions are not reachable from lowering of user code.

PR #1826, opened 2025-01-27
refactor[next]: fixed point transformation pass infrastructure

SF-N:
Executing transformations until a fixed point is reached, i.e., not transformation is applicable anymore is a common pattern encountered in many passes. The `CollapseTuple` pass already contained some infrastructure / utilities for this purpose. Since we now want to use the same approach in `ConstantFolding` and `FieldOpFusion`, they are extracted into a common base class here.

This is essentially a step into the direction of a more general pass manager that allows composition of transformations. It is very hard to design something on the scratch board that fulfills all needs we have in the various transformations. The idea is to extend the capabilities of this class step-by-step to cover more and more transformation use-cases such that more passes can use this infrastructure until we can eventually evaluate how dissimilar passes can be composed, e.g. CollapseTuple and ConstantFolding.

Edit @tehrengruber

### DIFF
--- a/src/gt4py/next/iterator/transforms/collapse_tuple.py
+++ b/src/gt4py/next/iterator/transforms/collapse_tuple.py
@@ -14,7 +14,6 @@ import functools
 import operator
 from typing import Optional
 
-from gt4py import eve
 from gt4py.eve import utils as eve_utils
 from gt4py.next import common
 from gt4py.next.iterator import ir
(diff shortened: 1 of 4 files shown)


havogt (approved):

tehrengruber:
Merging already. An ADR will follow soon that describes the current state.

PR #1828, opened 2025-01-28
feat[next][dace]: keep transients on the output of a mapped nested SDFG

edopao:
Small change to the lowering from GTIR, that keeps a transient buffer on the output of a nested SDFG, before the write memlet through the `MapExit` node. Removal of the transient, which was the baseline behavior, prevents map fusion in the optimization workflow.


philip-paul-mueller (commented):
LGTM

philip-paul-mueller (approved):
LGTM

PR #1830, opened 2025-01-28
bug[next]: Fix accept args

tehrengruber:
`type_info.accept_args` is used in the iterator type inference to check if a node with type `ts.FunctionType` is callable for a given set of arguments [here](https://github.com/GridTools/gt4py/blob/764ef5098078b327d086c4e06db357b45b046669/src/gt4py/next/iterator/type_system/inference.py#L237). This failed when the function type had an iterator argument with `position_dims="unknown"`. This PR promotes the `_is_compatible_type` function from the iterator type inference to a shared function available in all type systems and switches `accept_args` (or more precisely `function_signature_incompatibilities_func`) to use that function instead of `is_concretizable`.

Additionally this PR contains a small cleanup of the type deduction / type system tests:
- `test_type_deduction.py` was moved from integration to unit tests (as it only contains unit tests).
- All type system tests are moved from `test_type_deduction.py` as they are not specific to the frontend type deduction.

### DIFF
--- a/src/gt4py/next/type_system/type_info.py
+++ b/src/gt4py/next/type_system/type_info.py
@@ -26,6 +26,7 @@ import numpy as np
 
 from gt4py.eve.utils import XIterable, xiter
 from gt4py.next import common
+from gt4py.next.iterator.type_system import type_specifications as it_ts
 from gt4py.next.type_system import type_specifications as ts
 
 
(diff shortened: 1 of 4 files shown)


src/gt4py/next/type_system/type_info.py
```diff
@@ -725,11 +789,7 @@ def function_signature_incompatibilities_func(
     for i, (a_arg, b_arg) in enumerate(
         zip(list(func_type.pos_only_args) + list(func_type.pos_or_kw_args.values()), args)
     ):
-        if (
-            b_arg is not UNDEFINED_ARG
-            and a_arg != b_arg
-            and not is_concretizable(a_arg, to_type=b_arg)
-        ):
+        if b_arg is not UNDEFINED_ARG and a_arg != b_arg and not is_compatible_type(a_arg, b_arg):
```
tehrengruber, line 792: This and the change below is only real change in this PR.

havogt (approved):

PR #1833, opened 2025-01-29
refactor[cartesian]: Minor cleanup in backends

romanc:
## Description

I was reading a lot of code around DaCe/gt-codegen when debugging the new DaCe/gt4py bridge. This PR combines three cleanup commits:

- Always get stencil_ir from builder in GTBaseBackends. I've found no usage of `stencil_ir` being anything else than `self.build.gtir` if it was explicitly passed as an argument at all. There's thus no need to pass around `self.build.gtir` as long as we stay in the same class hierarchy.
- Avoid unnecessary indenting in generated code. Generated code is optionally formatted, but even if not, we can make sure the code doesn't look too ugly.
- Avoid double formatting of source code (if gt4py/dace is configured to do so). No need for formatting intermediate code parts because it's formatted anyway at the end.

## Requirements

  Updated test accordingly.
  N/A

### DIFF
--- a/src/gt4py/cartesian/backend/gtc_common.py
+++ b/src/gt4py/cartesian/backend/gtc_common.py
@@ -236,19 +236,15 @@ class BaseGTBackend(gt_backend.BasePyExtBackend, gt_backend.CLIBackendMixin):
 
     def generate_computation(self) -> Dict[str, Union[str, Dict]]:
         dir_name = f"{self.builder.options.name}_src"
-        src_files = self.make_extension_sources(stencil_ir=self.builder.gtir)
+        src_files = self._make_extension_sources()
         return {dir_name: src_files["computation"]}
 
-    def generate_bindings(
-        self, language_name: str, *, stencil_ir: Optional[gtir.Stencil] = None
-    ) -> Dict[str, Union[str, Dict]]:
-        if not stencil_ir:
-            stencil_ir = self.builder.gtir
-        assert stencil_ir is not None
+    def generate_bindings(self, language_name: str) -> Dict[str, Union[str, Dict]]:
         if language_name != "python":
             return super().generate_bindings(language_name)
+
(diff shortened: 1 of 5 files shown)


src/gt4py/cartesian/backend/gtc_common.py
```diff
@@ -260,32 +256,26 @@ def generate_extension(self, **kwargs: Any) -> Tuple[str, str]:
...
-        self, *, stencil_ir: Optional[gtir.Stencil] = None, uses_cuda: bool = False
-    ) -> Tuple[str, str]:
+    def make_extension(self, *, uses_cuda: bool = False) -> Tuple[str, str]:
         build_info = self.builder.options.build_info
         if build_info is not None:
             start_time = time.perf_counter()
 
-        if not stencil_ir:
-            stencil_ir = self.builder.gtir
-        assert stencil_ir is not None
```
FlorianDeconinck, line 272: That assert could still be technically true? (if `self.build.gtir` is None) Is is conceivable we would reach make extension with no `gtir` ?

FlorianDeconinck (approved):
One doubt about a removed assert but I would hope we catch no `gtir` much earlier that this code

src/gt4py/cartesian/backend/gtc_common.py
romanc: Or no, thinking again, I don't think we need the assert anymore. The `assert` was only there because `stencil_ir` was of type `gtir.Stencil | None` and I guess if the code above as written as
```python
if stencil_ir is None:
    stencil_ir = self.builder.gtir
```
then it wouldn't have been necessary in the first place.

That is, of course, depending on our trust in the type system and `mypy`. There are special exceptions for `mypy` to not check too carefully in the `cartesian/` folder. In that case, however, I think we should be safe.

PR #1841, opened 2025-01-31
build: use frozen dependencies when running mypy in pre-commit

havogt:
... otherwise it might update `uv.lock` which will create a pre-commit error because of changed files. See also https://github.com/astral-sh/uv/issues/10845


tehrengruber (approved):

havogt:
CI failure is due to a problem in slurm on daint. Will merge anyway.

PR #1843, opened 2025-02-01
refactor[cartesian]: unexpanded sdfg cleanups

romanc:
## Description

Refactors from recent debugging sessions around transient arrays in the "unexpanded SDFG" (the one with the library nodes):

- Remove unused `**kwargs` from `OirSDFGBuilder`
- Forward debugging information about transient arrays to DaCe
- Use a (constant) variable for connector prefixes of data going in to/out of the library nodes
- Configure the lifetime of transient arrays directly in `OirSDFGBuilder`
- Configure storage type of transient arrays directly in `OirSDFGBuilder` (for GPU targets)
- Configure the `LibraryNode`'s device type directly in `OirSDFGBuilder` (for GPU targets)
- Enum type for `StorageDevice` in `LayoutInfo`

Sidenote on the allocation lifetime: In the orchestrated code path, we reset the allocation lifetime of transients to `SDFG` when we freeze the stencil with origin/domain

https://github.com/GridTools/gt4py/blob/ac253b6b10a8adf87313a48b62328debaf39f07f/src/gt4py/cartesian/backend/dace_backend.py#L270-L317

This might be relevant when tracking down orchestration performance. Seems odd at least.

## Requirements

  Covered by existing tests
  N/A

### DIFF
--- a/src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
+++ b/src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
@@ -20,9 +20,8 @@ import dace.subsets
 from gt4py import eve
 from gt4py.cartesian.gtc.dace import daceir as dcir
 from gt4py.cartesian.gtc.dace.expansion.tasklet_codegen import TaskletCodegen
-from gt4py.cartesian.gtc.dace.expansion.utils import get_dace_debuginfo
 from gt4py.cartesian.gtc.dace.symbol_utils import data_type_to_dace_typeclass
-from gt4py.cartesian.gtc.dace.utils import make_dace_subset
+from gt4py.cartesian.gtc.dace.utils import get_dace_debuginfo, make_dace_subset
 
 
 class StencilComputationSDFGBuilder(eve.VisitorWithSymbolTableTrait):
@@ -268,13 +267,13 @@ class StencilComputationSDFGBuilder(eve.VisitorWithSymbolTableTrait):
             for memlet in computation.read_memlets:
                 if memlet.field not in read_acc_and_conn:
                     read_acc_and_conn[memlet.field] = (
-                        sdfg_ctx.state.add_access(memlet.field, debuginfo=dace.DebugInfo(0)),
+                        sdfg_ctx.state.add_access(memlet.field),
                         None,
                     )
             for memlet in computation.write_memlets:
                 if memlet.field not in write_acc_and_conn:
                     write_acc_and_conn[memlet.field] = (
-                        sdfg_ctx.state.add_access(memlet.field, debuginfo=dace.DebugInfo(0)),
+                        sdfg_ctx.state.add_access(memlet.field),
                         None,
                     )
             node_ctx = StencilComputationSDFGBuilder.NodeContext(
@@ -298,7 +297,7 @@ class StencilComputationSDFGBuilder(eve.VisitorWithSymbolTableTrait):
             dtype=data_type_to_dace_typeclass(node.dtype),
             storage=node.storage.to_dace_storage(),
             transient=node.name not in non_transients,
-            debuginfo=dace.DebugInfo(0),
+            debuginfo=get_dace_debuginfo(node),
         )
 
     def visit_SymbolDecl(
@@ -343,7 +342,6 @@ class StencilComputationSDFGBuilder(eve.VisitorWithSymbolTableTrait):
                 inputs=node.input_connectors,
                 outputs=node.output_connectors,
                 symbol_mapping=symbol_mapping,
-                debuginfo=dace.DebugInfo(0),
             )
             self.visit(
                 node.read_memlets,
--- a/src/gt4py/cartesian/gtc/dace/oir_to_dace.py
+++ b/src/gt4py/cartesian/gtc/dace/oir_to_dace.py
@@ -9,7 +9,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Dict
+from typing import Dict, Literal
 
 import dace
 import dace.properties
@@ -31,6 +31,17 @@ from gt4py.cartesian.gtc.passes.oir_optimizations.utils import (
     AccessCollector,
     compute_horizontal_block_extents,
 )
+
+
+transient_storage_per_device: Dict[Literal["cpu", "gpu"], dace.StorageType] = {
+    "cpu": dace.StorageType.Default,
+    "gpu": dace.StorageType.GPU_Global,
+}
+
+device_type_per_device: Dict[Literal["cpu", "gpu"], dace.DeviceType] = {
+    "cpu": dace.DeviceType.CPU,
+    "gpu": dace.DeviceType.GPU,
+}
 
 
 class OirSDFGBuilder(eve.NodeVisitor):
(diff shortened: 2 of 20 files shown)


romanc:
(rebased onto `main` branch because there was a hickup on the CSCS-CI)

src/gt4py/cartesian/gtc/dace/oir_to_dace.py
```diff
@@ -18,16 +18,32 @@
...
 
+transient_storage_per_device: Dict[Literal["cpu", "gpu"], dace.StorageType] = {
+    "cpu": dace.StorageType.Default,
+    "gpu": dace.StorageType.GPU_Global,
+}
+
+device_type_per_device: Dict[Literal["cpu", "gpu"], dace.DeviceType] = {
+    "cpu": dace.DeviceType.CPU,
+    "gpu": dace.DeviceType.GPU,
+}
```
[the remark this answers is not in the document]
FlorianDeconinck: Yeah we need an enum for this

FlorianDeconinck:
Re: `freeze_origin_domain_sdfg`

My cursory look seems to point that this is only called at top level for `DaceLazyStencil` et al. e.g. it's not called when we are doing orchestration _but_ for the top level arguments (and then it's fine).

We should log a task to double check _and_ we need a methodological/code way to differentiate code path we expect is in stencil mode and what is in orchestrated mode

src/gt4py/cartesian/gtc/dace/oir_to_dace.py
```diff
@@ -117,23 +138,25 @@ def visit_VerticalLoop(
         access_collection = AccessCollector.apply(node)
 
         for field in access_collection.read_fields():
-            access_node = state.add_access(field, debuginfo=dace.DebugInfo(0))
-            library_node.add_in_connector("__in_" + field)
+            access_node = state.add_access(field, debuginfo=get_dace_debuginfo(declarations[field]))
+            connector_name = CONNECTOR_PREFIX_IN + field
```
FlorianDeconinck, line 142: Nitpick: should we try to always use `f-string` when we have string operation to have a visual easy way to pick on the type of things?

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -268,13 +267,13 @@ def visit_ComputationState(
...
                 if memlet.field not in read_acc_and_conn:
                     read_acc_and_conn[memlet.field] = (
-                        sdfg_ctx.state.add_access(memlet.field, debuginfo=dace.DebugInfo(0)),
+                        sdfg_ctx.state.add_access(memlet.field),
                         None,
                     )
             for memlet in computation.write_memlets:
                 if memlet.field not in write_acc_and_conn:
                     write_acc_and_conn[memlet.field] = (
-                        sdfg_ctx.state.add_access(memlet.field, debuginfo=dace.DebugInfo(0)),
```
FlorianDeconinck, line 277: State gets no debug info? Do you expect the bottom to be enough? (I have no idea how this works inside DaCe)

FlorianDeconinck (changes requested):
Looking good. Couple of question.

Do the `enum` for the device type, the string is making me sad.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
romanc: Yes, in the DaCeIR, computation states don't derive from `LocNode` and thus don't have a `SourceLocation` attached. On the other side, `debuginfo` is optional in DaCe (as it is in gt4py) and imo passing `DebugInfo(0)` (which is short for `start_line: 0`, `start_column: 0`, `end_line: -1`, `end_column: 0`, `filename: None`) doesn't add meaningful information. I thus dropped it for simplicity.

Imo the whole debug info thing needs to be revisited at a bigger scale before we ship to atmospheric scientists. In this PR, I just plug easy holes in the current (wobbly) system.

src/gt4py/cartesian/gtc/dace/oir_to_dace.py
romanc: Sure, we can. No strong opinion on my side, so let's do it.

FlorianDeconinck (approved):
LGTM

FlorianDeconinck:
@havogt Care to review or ok to push?

src/gt4py/storage/cartesian/layout.py
```diff
@@ -30,9 +30,14 @@
         cp = None
 
 
+class StorageDevice(Enum):
+    CPU = 1
+    GPU = 2
```
havogt, line 35: It would be nice to use the `DeviceType` from `_core.definitions`. I don't have time to check how `ROCM` would fit here.
Maybe we can do
```
CurrentGPU = core_defs.DeviceType.ROCM if cp.cuda.runtime.is_hip else core_defs.DeviceType.CUDA
```
somewhere. cupy currently has a limitation that it is either installed for cuda or for rocm, so it would be fine.

havogt (commented):
I would prefer to use the DeviceType enum from `_core.definitions` instead of introducing a new enum. Can you check if that works?

src/gt4py/storage/cartesian/layout.py
romanc: Sure, let's reuse `DeviceType` from the core definitions. I honestly don't know what our support level for AMD graphic cards is. From what I read in the code, it looks like that's gonna be an item on the tasklist at some point. So I personally wouldn't worry too much about `ROCM` for now

FlorianDeconinck: ECMWF runs AMD hardware - so we need to keep this alive

romanc:
Putting this back to draft mode.

- Re-using `DeviceType` from `_core.definitions` is possible and I think it's worth doing.
- It's a bit of work and I might need to re-shuffle some things because `tach` complains that `gt4py.cartesian` should not depend on `gt4py._core` directly. I guess everything related to cpu/gpu device(s) should go through the storage module. In the end, there are many checks like "if this is a gpu device, then do this ..." and they should be centralized (in the `gt_storage` package or wherever) such that we can deal with `CUDA` / `ROCM` in a single place.
- Mixing the original changes in this PR with changing the "cpu" / "gpu" strings re-use `DeviceType` is going to look ugly in the git history (especially when squashed).
- I'll push a WIP state here and then probably separate the two things into separate (consecutive) PRs tomorrow.

src/gt4py/storage/cartesian/layout.py
romanc: For me to understand, which backend is compatible with AMD hardware? From what I read, the `CudaBackend` as well as the `DaceGPUBackend`, generate cuda code.

FlorianDeconinck: So `DaCe` can HIP - I believe that's where they are getting there AMD goodies. @stubbiali, can you talk to the AMD hardware use? Am I misremembering ?

stubbiali: Both GridTools and DaCe GPU backends support AMD graphics cards. We have been running GT4Py codes on LUMI for at least 1.5 years.

romanc:
Closing as superseded by https://github.com/GridTools/gt4py/pull/1860 and https://github.com/GridTools/gt4py/pull/1867.

PR #1846, opened 2025-02-05
fix[next][dace]: make if_ always execute branch exclusively

edopao:
This PR reverts the change previously made in #1824. Lowering `if_` expressions to tasklet is semantically wrong, from a dataflow perspective. It causes segmentation faults in several stencils, that rely on exclusive branch execution.

The source problem was that full array shape was passed into the nested SDFG scope, which prevented map fusion in most cases. This PR extends the lowering with the detection of simple iterator dereferencing, without shifts: for this type of data access, only the local element is moved into the nested SDFG. However, when shift is applied on the iterator input (which typically happens in iterator view), the full array shape is still passed.

This approach increases the optimization opportunities by enabling more map fusion. At the same time, it keeps the `if_` semantics of exclusive branch execution.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
@@ -533,14 +533,17 @@ class LambdaToDataflow(eve.NodeVisitor):
             assert isinstance(node.type, ts.ScalarType)
             return MemletExpr(arg_expr.field, arg_expr.gt_dtype, subset="0")
 
-        # default case: deref a field with one or more dimensions
+        # handle default case below: deref a field with one or more dimensions
+
+        # when the indices are all dace symbolic expressions, the deref is lowered
+        # to a memlet, where the index is the memlet subset
         if all(isinstance(index, SymbolExpr) for index in arg_expr.indices.values()):
             # when all indices are symbolic expressions, we can perform direct field access through a memlet
             field_subset = arg_expr.get_memlet_subset(self.sdfg)
             return MemletExpr(arg_expr.field, arg_expr.gt_dtype, field_subset)
 
-        # we use a tasklet to dereference an iterator when one or more indices are the result of some computation,
-        # either indirection through connectivity table or dynamic cartesian offset.
+        # when any of the indices is a runtime value (either a dynamic cartesian
+        # offset or a connectivity offset), the deref is lowered to a tasklet
         assert all(dim in arg_expr.indices for dim, _ in arg_expr.field_domain)
         assert len(field_desc.shape) == len(arg_expr.field_domain)
         field_indices = [(dim, arg_expr.indices[dim]) for dim, _ in arg_expr.field_domain]
@@ -559,7 +562,7 @@ class LambdaToDataflow(eve.NodeVisitor):
             for dim, index in field_indices
         )
         deref_node = self._add_tasklet(
-            "runtime_deref",
+            "deref",
             {"field"} | set(index_connectors),
             {"val"},
             code=f"val = field[{index_internals}]",
--- a/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
@@ -29,7 +29,7 @@ from dace import subsets as dace_subsets
 
 from gt4py import eve
 from gt4py.next import common as gtx_common, utils as gtx_utils
-from gt4py.next.iterator import builtins, ir as gtir
+from gt4py.next.iterator import builtins as gtir_builtins, ir as gtir
 from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm, ir_makers as im
 from gt4py.next.iterator.transforms import symbol_ref_utils
 from gt4py.next.program_processors.runners.dace import (
@@ -602,6 +602,7 @@ class LambdaToDataflow(eve.NodeVisitor):
         if_branch_state: dace.SDFGState,
         param_name: str,
         arg: IteratorExpr | DataExpr,
+        is_shifted_iterator: bool,
         if_sdfg_input_memlets: dict[str, MemletExpr | ValueExpr],
     ) -> IteratorExpr | ValueExpr:
         """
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -613,35 +613,53 @@ def _visit_if_branch_arg(
             if_branch_state: The state inside the nested SDFG where the if branch is lowered.
             param_name: The parameter name of the input argument.
             arg: The input argument expression.
+            is_shifted_iterator: When True, the given argument is an iterator and it is shifted inside the branch expression.
```
philip-paul-mueller, line 616: I do not really like the name.
Furthermore, in my opinion, the description does not fully reflects its meaning.

I actually think that `use_full_shape` would be a better name it also looks to me as if one `use_full_shape` or `is_shifted_iterator` are redundant, if  you also know the type of `arg`.
So I would remove (the inner) `use_full_shape` and rename `is_shifted_iterator` to `use_full_shape` also a name `try_to_use_full_shape` could be appropriate, because the decision if the full shape is used also depends on the type of `arg`.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -613,35 +613,53 @@ def _visit_if_branch_arg(
...
+                # In order to shift the iterator inside the branch dataflow,
+                # we have to pass the full array shape.
+                arg_expr = MemletExpr(
+                    arg.field, arg.gt_dtype, dace_subsets.Range.from_array(arg_desc)
+                )
+                use_full_shape = True
+            else:
+                # If the iterator is just dereferenced, inside the branch state,
+                # we can access the array data outside and just pass the local data.
+                # This approach increases the chances of applying map fusion.
```
philip-paul-mueller, line 636:
```suggestion
                # This approach makes the data dependencies of nested structures more explicit and thus makes it easier for MapFusion to correctly infer the data dependencies.  
```

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -613,35 +613,53 @@ def _visit_if_branch_arg(
...
+            elif isinstance(arg.gt_dtype, ts.ScalarType):
+                inner_desc = dace.data.Scalar(arg_desc.dtype)
+            else:
+                # for list of values, we retrieve the local size from the corresponding offset
+                assert arg.gt_dtype.offset_type is not None
+                offset_provider_type = self.subgraph_builder.get_offset_provider_type(
+                    arg.gt_dtype.offset_type.value
+                )
+                assert isinstance(offset_provider_type, gtx_common.NeighborConnectivityType)
+                inner_desc = dace.data.Array(arg_desc.dtype, [offset_provider_type.max_neighbors])
```
philip-paul-mueller, line 657: Is my understanding of this code correct.
As far as I understand this effectively limits this "lets dereference outside" scheme to scalars (okay that is above) and local dimensions.
So it effectively prevents that slices of non local dimensions, such as all edges are created?
Is this correct?
If so I would suggest to add a comment to make this clear.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -674,6 +692,17 @@ def _visit_if_branch(
         """
         assert if_branch_state in if_sdfg.states()
 
+        # collect all field iterators that are shifted inside the branch expression
```
philip-paul-mueller, line 695: Please add an explanation why you do this.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -674,6 +692,17 @@ def _visit_if_branch(
...
+        shifted_iterators = set()
+        for shift_node in eve.walk_values(expr).filter(lambda x: cpm.is_applied_shift(x)):
+            shifted_iterators |= (
+                eve.walk_values(shift_node)
+                .if_isinstance(gtir.SymRef)
+                .map(lambda x: str(x.id))
+                .filter(lambda x: isinstance(self.symbol_map.get(x, None), IteratorExpr))
+                .to_set()
+            )
```
philip-paul-mueller, line 704: I am not fully remembering, but we discussed something with Till and there you explained some heuristic and (as far as I can remember) his assessment of it was that it is most likely fine but not in general.
So now, is this a new heuristic?
I am aware that the problem is now different, as we always generate exclusive `if`s and to me it looks fine, I just want to ask how the following cases are handled:
- Reduction, especially with skip values?
- dymref?

philip-paul-mueller (commented):
It looks good, there are some small issues that have to be clarified.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
edopao: Correct. Although I would not use "prevent". In current implementation, the `deref` operator does not allow to slice a field along a global dimension, not here nor in the generic `deref`.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
edopao: I prefer to keep the two concepts separate, although I have changed the bool variable names to make the logic more clear (at least I hope so).

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
philip-paul-mueller: Okay thanks for the explanation.
With `dymref` I meant the runtime deref (you create tasklets with the name `tlet_{NUMBER}_runtime_deref`) are they detected or are they just a shift combined with a deref?

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -613,35 +614,56 @@ def _visit_if_branch_arg(
             if_branch_state: The state inside the nested SDFG where the if branch is lowered.
             param_name: The parameter name of the input argument.
             arg: The input argument expression.
+            deref_on_input_memlet: When True, the given iterator argument can be dereferenced on the input memlet.
```
philip-paul-mueller, line 617:
```suggestion
            deref_on_input_memlet: When True, the given iterator argument can be dereferenced on the input memlet. This means that the values are copied into a temporary storage which is passed into the nested SDFG.
```
I think the description of this option should have a bit more information.

philip-paul-mueller (commented):
It generally looks good to me, there is however style of question of the runtime deref.
Furthermore, if I run it locally the access nodes are not created, so we are still at square one (maps can not be fused).

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
edopao: The change you propose is not correct. We are not allocating temporary storage, we are just narrowing the memlet subset.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
edopao: They are just shifts. I will change the name of those tasklets.

edopao: Sorry, I should have written this is just a `deref` tasklet. When the indices are all dace symbolic expressions, I lower the `deref` as a memlet, where the index is the memlet subset. When any of the indices is a runtime value (either a dynamic cartesian offset or a connectivity offset), the `deref` is lowered to a tasklet.

philip-paul-mueller (commented):
LGTM

The issue I observed is somehow related to `substitute_compiletime_symbols()`, however, I have no idea how.

philip-paul-mueller (approved):
Too fast.

PR #1850, opened 2025-02-06
fix[dace]: Updating MapFusion

philip-paul-mueller:
The [MapFusion PR](https://github.com/spcl/dace/pull/1629) in DaCe is still under review.
However, the MapFusion in that PR has evolved, i.e. some bugs were fixed and now GT4Py is also these bugs.
This PR essentially back ports some of the fixes to GT4Py.

Note that this is a temporary solution and as soon as the MapFusion PR has been merged (and parallel map fusion has been introduced) the GT4Py version will go away.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_serial.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_serial.py
@@ -17,7 +17,7 @@ import copy
 from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
 import dace
-from dace import data, dtypes, properties, subsets, symbolic, transformation
+from dace import data, properties, subsets, symbolic, transformation
 from dace.sdfg import SDFG, SDFGState, graph, nodes
 
 from . import map_fusion_helper as mfh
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_serial.py
```diff
@@ -967,41 +973,88 @@ def handle_intermediate_set(
...
+        # Over approximation will leave us with some unneeded size one dimensions.
+        #  If they are removed some dace transformations (especially auto optimization)
+        #  will have problems.
+        new_inter_shape_raw = symbolic.overapproximate(producer_subset.size())
+        inter_shape = inter_desc.shape
+        if not self.strict_dataflow:
+            squeezed_dims: List[int] = []  # These are the dimensions we removed.
+            new_inter_shape: List[int] = []  # This is the final shape of the new intermediate.
+            for dim, (proposed_dim_size, full_dim_size) in enumerate(
+                zip(new_inter_shape_raw, inter_shape)
```
edopao, line 1048: Since `inter_shape` is cloned from a function input argument and `producer_subset` is also an input argument, I would suggest:
```suggestion
                zip(new_inter_shape_raw, inter_shape, strict=True)
```

edopao (approved):
Only one minor comment.

Issue #720, opened 2022-04-08
Add additional integration tests for horizontal regions


jdahm:
There is one test suite test, but there are a number of tests that @gronerl used for some initial testing that I think should be integrated after we merge the initial PR.

PR #1851, opened 2025-02-06
tests[cartesian]: Increase horizontal region test coverage

romanc:
## Description

A a test case where horizontal regions are used to write to a subset of the field. Note that this is different from
`TestHorizontalRegionsCorners` because we have specialized optimizations in place for corners of the cube.

Related issue: https://github.com/GridTools/gt4py/issues/720

## Requirements

  New tests added
  N/A


romanc:
The new bridge has [one case](https://github.com/GridTools/gt4py/commit/27bc0b2fd5074beb67734a8016ca83c5d79d3261) where this new test fails. That bug is partly why I've seen nan's in the computation of e.g. `DelnFlux` where we use horizontal regions to only write to parts of a field.

FlorianDeconinck (approved):
LGTM

PR #1854, opened 2025-02-10
test[next]: enable gpu test for 1d scan on dace backend

edopao:
The dace backend did not support the case of scan on a 1D vertical array, when lowered to GPU code. After upgrading the dace package to latest main and adopting the `LoopRegion` construct for the lowering of scan, this case is no longer an issue. This PR just re-enables the scan test case.


philip-paul-mueller (commented):
I technically have no problem with this changes.
However, there is this link to DaCe issue 1773, that it is still open.
So my question is how did the issue got resolved?

edopao:
> I technically have no problem with this changes. However, there is this link to DaCe issue 1773, that it is still open. So my question is how did the issue got resolved?

I should have explained it better "this case is no longer an issue". The dace issue is still open. The issue is actually reported on a different test case (`test_execution.py::test_double_use_scalar`). The 1d scan was a sub-issue, that shared the same root cause as the test case `test_execution.py::test_double_use_scalar`. However, the 1d scan is no longer an issue because after upgrading to dace main the scan code generation starts from a `LoopRegion`, while before (on dace v1) the `LoopRegion` was converted to a state machine. In other words, we do not hit the issue in code generation.

philip-paul-mueller (approved):
LGTM.

PR #1856, opened 2025-02-10
fix[next][dace]: fix map fusion and loop blocking

edopao:
Improve optimization for icon4py stencil `apply_diffusion_to_vn` by means of two changes:
- Ignore check of dynamic volume property on memlet that prevented serial map fusion.
- Add support for NestedSDFG nodes in loop blocking transformation.


src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_serial.py
```diff
@@ -589,9 +589,6 @@ def partition_first_outputs(
                 ):
                     if inner_consumer_edge.data.src_subset is None:
                         return None
-                    if inner_consumer_edge.data.dynamic:
```
philip-paul-mueller, line 592: Last week when I started debugging I also found this issue, however, I also found a different one that indicates that the list of control flow region got messed up.
So I am not sure if it is safe because it will just prevent that we run into another error.

Never the less I think that this change is good as this case is also allowed in the MapFusion PR.
However, I suggest to you that you copy the content of the `partition_first_outputs()` from the MapFusion PR into this file.
I think I will change my plan and update the MapFusion in this PR with the one in DaCe.
However, for full compatibility I will have to make some changes, i.e. add parallel map fusion.
But this should not concern you at this point.

Note that ignoring the dynamic properties for the consumer is okay but not for the producer.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -380,6 +380,14 @@ def _classify_node(
             ):
                 return False
 
+        elif isinstance(node_to_classify, dace.nodes.NestedSDFG):
+            # Similar checks apply to a nested SDFG node
+            if not all(
+                isinstance(out_edge.dst, dace_nodes.AccessNode)
+                for out_edge in state.out_edges(node_to_classify)
```
philip-paul-mueller, line 387:
```suggestion
                for out_edge in state.out_edges(node_to_classify)
                if not out_edge.data.is_empty()
```

In addition you have to inspect if the nested SDFG uses the blocking parameter.
I am not sure if checking `free_symbols` is enough.
Consider the following (partial) symbol mappings; where `k` is the block parameter:
- `{'k': 'k'}` in this case it is certainly enough to check `free_symbol`.
- `{'inner_value': 'k'}` as far as I understand `k` will not be a free symbol of the nested SDFG.

So I think to be fully sure you will have to do something along:
```python
any('k' in ss for ss in [symbol_mapping.keys(), symbol_mapping.values(), nsdfg.free_symbol])
```
Maybe the check on free symbol is overkill.
Also be aware the `symbol_mapping.keys()` returns strings while `symbol_mapping.values()` returns symbols (they imposed some type casting in their properties but I am not sure how good it is).
Also be aware of cases such as `k + 10` or `k_othersymbol` which makes matching harder.

philip-paul-mueller (commented):
There are some issues, but it looks generally good.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_serial.py
edopao: I have now copied function `partition_first_outputs()` from the dace PR, with exclusion of input argument `param_repl`, which is still computed inside.

philip-paul-mueller: No worry this file has gone beyond saving and we will purge it soon^{(TM)}.

philip-paul-mueller (approved):
LGTM.

There are some small things:
- Please move the check of the `symbol_mapping` of the `NestedSDFG` to the other `NestedSDFG` specific cases, see the comment for more detail. 
- In test I would suggest the `default:` case of the match to throw a `NotImplementedError` instead of an `assert`, I think it is a bit more specific. But this is probably a matter of taste.
- I have fully ignored the changes to `map_fusion_serial.py` as it will be superseded by https://github.com/GridTools/gt4py/pull/1857

PR #1866, opened 2025-02-14
fix[next][dace]: Fixed a bug in the map fusion's default parameter

philip-paul-mueller:
`MapFusion` was behaving as the one from DaCe, i.e. it only did serial fusion by default, now it allows both by default. There are still `MapFusionSerial` and `MapFusionParallel` with the respective settings. This commit also changes auto optimizer to explicitly request them.


src/gt4py/next/program_processors/runners/dace/transformations/map_fusion.py
```diff
@@ -81,11 +83,18 @@ class MapFusion(dace_map_fusion.MapFusion):
     def __init__(
         self,
         strict_dataflow: bool = False,
+        allow_serial_map_fusion: bool = True,
```
edopao, line 86: It is a matter of taste, but I would make a simpler interface and not expose these 2 new parameters.

edopao (approved):

PR #1867, opened 2025-02-14
refactor[cartesian]: use DeviceType from core_defn

romanc:
## Requirements

### DIFF
--- a/src/gt4py/cartesian/backend/base.py
+++ b/src/gt4py/cartesian/backend/base.py
@@ -45,9 +45,9 @@ if TYPE_CHECKING:
 REGISTRY = gt_utils.Registry()
 
 
-def from_name(name: str) -> Optional[Type[Backend]]:
+def from_name(name: str) -> Type[Backend]:
     backend = REGISTRY.get(name, None)
-    if not backend:
+    if backend is None:
         raise NotImplementedError(
             f"Backend {name} is not implemented, options are: {REGISTRY.names}"
         )
@@ -84,7 +84,7 @@ class Backend(abc.ABC):
     #: Backend-specific storage parametrization:
     #:
     #:  - "alignment": in bytes
-    #:  - "device": "cpu" | "gpu"
+    #:  - "device": core_defs.DeviceType | None
     #:  - "layout_map": callback converting a mask to a layout
     #:  - "is_optimal_layout": callback checking if a storage has compatible layout
     storage_info: ClassVar[gt_storage.layout.LayoutInfo]
@@ -435,21 +435,20 @@ def disabled(message: str, *, enabled_env_var: str) -> Callable[[Type[Backend]],
     enabled = bool(int(os.environ.get(enabled_env_var, "0")))
     if enabled:
         return deprecated(message)
-    else:
 
-        def _decorator(cls: Type[Backend]) -> Type[Backend]:
-            def _no_generate(obj) -> Type[StencilObject]:
-                raise NotImplementedError(
-                    f"Disabled '{cls.name}' backend: 'f{message}'\n",
-                    f"You can still enable the backend by hand using the environment variable '{enabled_env_var}=1'",
-                )
+    def _decorator(cls: Type[Backend]) -> Type[Backend]:
+        def _no_generate(obj) -> Type[StencilObject]:
+            raise NotImplementedError(
+                f"Disabled '{cls.name}' backend: 'f{message}'\n",
+                f"You can still enable the backend by hand using the environment variable '{enabled_env_var}=1'",
+            )
 
-            # Replace generate method with raise
-            if not hasattr(cls, "generate"):
-                raise ValueError(f"Coding error. Expected a generate method on {cls}")
-            # Flag that it got disabled for register lookup
-            cls.disabled = True  # type: ignore
-            cls.generate = _no_generate  # type: ignore
-            return cls
+        # Replace generate method with raise
+        if not hasattr(cls, "generate"):
+            raise ValueError(f"Coding error. Expected a generate method on {cls}")
+        # Flag that it got disabled for register lookup
+        cls.disabled = True  # type: ignore
+        cls.generate = _no_generate  # type: ignore
+        return cls
 
-        return _decorator
+    return _decorator
--- a/src/gt4py/cartesian/stencil_object.py
+++ b/src/gt4py/cartesian/stencil_object.py
@@ -15,7 +15,7 @@ import time
 from dataclasses import dataclass
 from numbers import Number
 from pickle import dumps
-from typing import Any, Callable, ClassVar, Dict, Literal, Optional, Tuple, Union, cast
+from typing import Any, Callable, ClassVar, Dict, Optional, Tuple, Union, cast
 
 import numpy as np
 
@@ -24,6 +24,7 @@ import gt4py.storage.cartesian.utils as storage_utils
 from gt4py import cartesian as gt4pyc
 from gt4py.cartesian.definitions import AccessKind, DomainInfo, FieldInfo, ParameterInfo
 from gt4py.cartesian.gtc.definitions import Index, Shape
+from gt4py.storage.cartesian.layout import LayoutInfo
 
 
 try:
@@ -51,7 +52,6 @@ def _compute_domain_origin_cache_key(
 
 @dataclass
 class ArgsInfo:
-    device: str
     array: FieldType
     original_object: Optional[Any] = None
     origin: Optional[Tuple[int, ...]] = None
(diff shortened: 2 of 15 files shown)


src/gt4py/cartesian/backend/base.py
```diff
@@ -58,15 +58,13 @@ def register(backend_cls: Type[Backend]) -> Type[Backend]:
     assert issubclass(backend_cls, Backend) and backend_cls.name is not None
 
     if isinstance(backend_cls.name, str):
-        gt_storage.register(backend_cls.name, backend_cls.storage_info)
```
romanc, line 61: Layouts were registered twice. Once here and once in the `storage/cartesian/layout.py`. Most layouts were registered under the same name (i.e. the backend name), however some were duplicated. I choose to remove the layout registration here and changed all the names in `storage/cartesian/layout.py` to match backend names. This keeps layout tests separate.

src/gt4py/cartesian/backend/base.py
```diff
@@ -58,15 +58,13 @@ def register(backend_cls: Type[Backend]) -> Type[Backend]:
...
         return REGISTRY.register(backend_cls.name, backend_cls)
 
-    else:
-        raise ValueError(
-            "Invalid 'name' attribute ('{name}') in backend class '{cls}'".format(
-                name=backend_cls.name, cls=backend_cls
-            )
```
romanc, line 68: Unrelated: no `else` needed after `return` in `if`

src/gt4py/cartesian/stencil_object.py
```diff
@@ -51,22 +52,21 @@ def _compute_domain_origin_cache_key(
 
 @dataclass
 class ArgsInfo:
-    device: str
```
romanc, line 54: Never used, so I removed it

src/gt4py/storage/__init__.py
```diff
@@ -11,7 +11,7 @@
 from . import cartesian
 from .cartesian import layout
 from .cartesian.interface import empty, from_array, full, ones, zeros
-from .cartesian.layout import from_name, register
```
romanc, line 14: layout registration is now private. Not sure if this is the idea. Happy to revert and register layouts in the base backend instead.

romanc (commented):
This turned out to be a bit bigger than expected. Main pain point was the test generation (which needs a static way of knowing which backend/layout needs a GPU).

I left a couple of questions & remarks inline. Let's have a discussion ...

romanc:
Digging a bit more, I found that the only reason (left) for `storage` to depend on `cartesian` is `GT4PY_USE_HIP` (see PR https://github.com/GridTools/gt4py/pull/1278) which was added in `gt4py/cartesian/config.py` instead of next to the aloocators in the storage package. If we are doing this, it might be worth to get rid of that dependency cycle too.

To me, it feels like this AMD/NVIDIA device handling should all be centralized and re-used in next and cartesian. Like, the `CUPY_DEVICE` should live in `_core` and a potential `GT4PY_USE_HIP` override too.

Question for @stubbiali: Do you guys actually use the `GT4PY_USE_HIP` env variable (when running on LUMI)?

stubbiali:
IIRC, we introduced `GT4PY_USE_HIP` following cupy's `CUPY_INSTALL_USE_HIP` (see [here](https://docs.cupy.dev/en/stable/install.html#using-cupy-on-amd-gpu-experimental)), but the automatic detection using `cupy.cuda.get_hipcc_path()` has been working well, so there's actually no reason to set `GT4PY_USE_HIP` via an env variable.

tests/cartesian_tests/definitions.py
```diff
@@ -14,42 +14,47 @@
...
         marks.append(pytest.mark.requires_dace)
-    return pytest.param(name, marks=marks)
-
-
-_ALL_BACKEND_NAMES = list(gt4pyc.backend.REGISTRY.keys())
+    return pytest.param(backend_name, marks=marks)
 
 
-def _get_backends_with_storage_info(storage_info_kind: str):
+def _filter_backends(filter: Callable[[Backend], bool]) -> List[str]:
```
FlorianDeconinck, line 44: Let's make sure the test generated are the same as before. `pytest --collect-only` will collect all combinations of tests and not run them

src/gt4py/cartesian/testing/suites.py
```diff
@@ -206,7 +207,7 @@ def hyp_wrapper(test_hyp, hypothesis_data):
                 )
 
                 marks = test["marks"].copy()
-                if gt4pyc.backend.from_name(test["backend"]).storage_info["device"] == "gpu":
+                if test["backend"] in GPU_BACKEND_NAMES:
```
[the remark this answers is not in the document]
FlorianDeconinck: And we could extend it even further by re-routing allocation through, etc. I am no fan of massive class that does it all, but in this case `Backend` seems underused a single source of truth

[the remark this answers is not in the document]
FlorianDeconinck: Probably out of scope for this PR

FlorianDeconinck (changes requested):
Looking good. Test failing, some musing about where layout metadata should live

src/gt4py/storage/cartesian/layout.py
```diff
@@ -136,37 +159,53 @@ def make_cuda_layout_map(dimensions: Tuple[str, ...]) -> Tuple[Optional[int], ..
 
 NaiveCPULayout: Final[LayoutInfo] = {
     "alignment": 1,
-    "device": "cpu",
+    "device": core_defs.DeviceType.CPU,
     "layout_map": lambda axes: tuple(i for i in range(len(axes))),
     "is_optimal_layout": lambda *_: True,
 }
-register("naive_cpu", NaiveCPULayout)
+_register("numpy", NaiveCPULayout)
```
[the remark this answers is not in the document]
egparedes: I haven't looked at this part of the cartesian codebase in years but I think the root of the problem is that there is no actual separation in the code between storage layouts and the backend, although they are orthogonal. I mean, the layout definition inside a backend encodes the optimal layout for that particular backend, but most of the backends will work anyway with other layouts, and several backends can have the same optimal layout, so they are clearly different concepts.

The registry in this file was probably meant to encode the mapping between layout names and its specification and it would then make sense that the backend object contained the layout name, which should be looked up in this registry. But since the backend object contains the whole `LayoutInfo` instance, this registry doesn't make any sense and it should be then deleted, since the backend registry is enough to recover the same information:
```python
backend.REGISTRY[backend_name].storage_info == layout.REGISTRY[backend_name]
```
So I would either add a new indirection layer by storing only the layout name in the backend or just remove the layout registry and only use the backend registry for everything (at this point I think it only makes sense to go with the second option to make the codebase simpler and delete redundant code).

src/gt4py/storage/cartesian/layout.py
```diff
@@ -136,37 +159,53 @@ def make_cuda_layout_map(dimensions: Tuple[str, ...]) -> Tuple[Optional[int], ..
...
     "is_optimal_layout": layout_checker_factory(make_cuda_layout_map),
 }
-register("cuda", CUDALayout)
+_register("cuda", CUDALayout)
 
 GPULayout: Final[LayoutInfo] = CUDALayout
-register("gpu", GPULayout)
+_register("gt:gpu", GPULayout)
+
+DaceCPULayout: Final[LayoutInfo] = {
```
egparedes, line 197: The dace layout is useless if the dace backend is not installed...

egparedes (commented):
Just a question and my two cents about the design of the backend-layout relationship.

PS: I'm sorry for the late reply, I was on vacation.

src/gt4py/cartesian/backend/base.py
```diff
@@ -84,7 +82,7 @@ class Backend(abc.ABC):
     #: Backend-specific storage parametrization:
     #:
     #:  - "alignment": in bytes
-    #:  - "device": "cpu" | "gpu"
+    #:  - "device": core_defs.DeviceType | None
```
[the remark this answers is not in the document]
romanc: The old documentation is correct for the old value. The old was was a static `"cpu" | "gpu"` string defined in the layout. In this PR I changed `layout.device` to be the actual device (not just a descriptor anymore) and because the GPU devices are auto-detected as

```python
CUPY_DEVICE: Final[Literal[DeviceType.CUDA, DeviceType.ROCM] | None] = (
    None if not cp else (DeviceType.ROCM if cp.cuda.get_hipcc_path() else DeviceType.CUDA)
)
```

The `None` type then slips in for situations where `cupy` is unable to find a graphics card. I'm not quite sure yet where we'll end up with this PR (probably a bunch of follow up PR and then close this one). You are right that it feels a bit awkward.

PR #1868, opened 2025-02-15
feat[next]: ITIR type inference: store param types or `itir.Lambda`

tehrengruber:


tests/next_tests/unit_tests/iterator_tests/test_type_inference.py
```diff
@@ -245,6 +245,7 @@ def test_aliased_function():
     assert result.args[0].type == ts.FunctionType(
         pos_only_args=[int_type], pos_or_kw_args={}, kw_only_args={}, returns=int_type
     )
+    assert result.args[0].params[0].type == int_type
```
egparedes, line 248: Just a question: is there any test case using lambdas with more than one parameter?

src/gt4py/next/iterator/type_system/inference.py
```diff
@@ -37,6 +37,10 @@ def _set_node_type(node: itir.Node, type_: ts.TypeSpec) -> None:
         assert type_info.is_compatible_type(
             node.type, type_
         ), "Node already has a type which differs."
+    if isinstance(node, itir.Lambda):
```
egparedes, line 40: Would it make sense to add an explanation comment here?

egparedes (approved):
LGTM

PR #1872, opened 2025-02-17
fix[next][dace]: use logical and/or/xor operators, not bitwise

edopao:
The current mapping from GTIR logical operators to python code was inconsistent. The mapping was using bitwise operators instead of logical ones. This still resulted in functionally correct code because the boolean type has an integer representation in python language. This PR introduces the correct mapping, and leaves the dace toolchain and target compiler the possibility to generate optimized code.


philip-paul-mueller (approved):
LGTM.

PR #1874, opened 2025-02-19
feat[eve]: Preserve annex in custom visitors 

tehrengruber:
Preserve annex attributes in the `NodeTranslator` for nodes returned from custom `visit_` methods, not only from nodes which are recreated in `generic_visit`. This is often desired and with this PR doesn't need to be done manually in child classes of `NodeTranslator`. The `visit` method can still decide if the node it returns should have a different annex attribute as `_preserve_annex` does not overwrite in case the attribute already exists.


src/gt4py/eve/visitors.py
```diff
@@ -158,13 +173,7 @@ def generic_visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
...
                 }
             )
-            if self.PRESERVED_ANNEX_ATTRS and (old_annex := getattr(node, "__node_annex__", None)):
-                # access to `new_node.annex` implicitly creates the `__node_annex__` attribute in the property getter
-                new_annex_dict = new_node.annex.__dict__
-                for key in self.PRESERVED_ANNEX_ATTRS:
-                    if (value := getattr(old_annex, key, NOTHING)) is not NOTHING:
-                        assert key not in new_annex_dict
-                        new_annex_dict[key] = value
+            _preserve_annex(node, new_node, self.PRESERVED_ANNEX_ATTRS)
```
egparedes, line 176: What do you think about leaving the condition outside of the function? This is probably the hottest loop in the toolchain so I'd rather try to avoid a unnecessary function call if possible. Additionally, it would also make the function name fully correct :smile: 
```suggestion
            if preserved_annex_attrs and hasattr(node, "__node_annex__"):
                _preserve_annex(node, new_node, self.PRESERVED_ANNEX_ATTRS)
```

src/gt4py/eve/visitors.py
```diff
@@ -189,3 +198,11 @@ def generic_visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
             )
 
         return copy.deepcopy(node, memo=memo)
+
+    def visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
+        new_node = super().visit(node, **kwargs)
+
+        if isinstance(node, concepts.Node) and isinstance(new_node, concepts.Node):
+            _preserve_annex(node, new_node, self.PRESERVED_ANNEX_ATTRS)
```
egparedes, line 206: Same comment as above

src/gt4py/next/iterator/transforms/infer_domain.py
```diff
@@ -436,8 +436,12 @@ def _infer_stmt(
     **kwargs: Unpack[InferenceOptions],
 ):
     if isinstance(stmt, itir.SetAt):
+        # constant fold once otherwise constant folding after domain inference might create (syntactic) differences
```
egparedes, line 439: Does this change really belong to this PR?

src/gt4py/eve/visitors.py
tehrengruber: I thought about this, but went for not duplicating the condition. Makes sense to me though, so I reverted that.

src/gt4py/next/iterator/transforms/infer_domain.py
tehrengruber: I initially checked that the new annex attribute is equal to the old one. That would be a desirable property of the IR and this change here is a step in that direction. This also motivated the comment above
```
# Note: The annex value of the new node might not be equal
# (in the sense that an equality comparison returns false),
# but in the context of the pass, they are equivalent.
# Therefore, we don't assert equality here.
```
However there are annex values that are equivalent, but not equal e.g. the `domain` annex with a value of
`(domain, domain)` (tuple of domains) is equivalent to `domain`. I therefore gave up on the assert, but kept this change here. Does that make sense?

src/gt4py/eve/visitors.py
```diff
@@ -121,6 +121,21 @@ def generic_visit(self, node: concepts.RootNode, **kwargs: Any) -> Any:
...
+    # access to `new_node.annex` implicitly creates the `__node_annex__` attribute in the property getter
+    old_annex = node.annex
+    new_annex_dict = new_node.annex.__dict__
+    for key in preserved_annex_attrs:
+        if (value := getattr(old_annex, key, NOTHING)) is not NOTHING:
+            # Note: The annex value of the new node might not be equal
+            # (in the sense that an equality comparison returns false),
+            # but in the context of the pass, they are equivalent.
+            # Therefore, we don't assert equality here.
+            new_annex_dict[key] = value
```
egparedes, line 136: It just occurred to me that the code could be simplified further (or am I missing something?):
```suggestion
    # access to `new_node.annex` implicitly creates the `__node_annex__` attribute in the property getter
    new_annex_dict = new_node.annex.__dict__
    old_annex_dict = node.annex.__dict
    for key in old_annex_dict.keys() & preserved_annex_attrs:
        # Note: The annex value of the new node might not be equal
        # (in the sense that an equality comparison returns false),
        # but in the context of the pass, they are equivalent.
        # Therefore, we don't assert equality here.
        new_annex_dict[key] = old_annex_dict[key]
```
Also after reading your reply  to the other issue below, I don't really understand this comment... Which two values should be semantically equivalent even if not compare true for equality? If you refer to the value in the old and new annex, I don't see how could that happen, given that the value is copied literally here...

src/gt4py/next/iterator/transforms/infer_domain.py
egparedes: I'm now fine with the change here, but don't really understand the source comment above in `_preserve_annex()` (see above) :sweat_smile:

src/gt4py/eve/visitors.py
tehrengruber: The value of the annex of the new node and the old node should be equivalent. I omitted the "old node" in the comment to keep the sentence focused, but apparently it became unclear that way. I clarified it. I took your first suggestion, but adopted it such that the new annex value is taken when it already exists. This had been a shortcoming in my original version too, which I overlooked. I've added a test `test_annex_preservation_translated_node_overwritten`.

egparedes (approved):
LGTM

PR #1876, opened 2025-02-19
Revert "feat[next][dace]: keep transients on the output of a mapped nested SDFG"

edopao:
Reverts GridTools/gt4py#1828


edopao:
Only a test

PR #1877, opened 2025-02-19
fix[next][dace]: remove temporary arrays with runtime shape on the output of a mapped nested SDFG

edopao:
This PR provides a better fix than the one delivered earlier in https://github.com/GridTools/gt4py/pull/1828. It adds a check to detect whether the temporary output data has compile-time or runtime size. In case of runtime size, the transient array on the output connector of a mapped nested SDFG is removed. This is needed in order to avoid dynamic memory allocation inside the cuda kernel that represents a parallel map scope.


philip-paul-mueller (approved):
LGTM

PR #1883, opened 2025-02-20
bug[next]: Respect bc in `InlineCenterDerefLiftVars`

tehrengruber:
Changes the `InlineCenterDerefLiftVars` such that it respect boundary condition by lazy evaluating the inlined values.

Consider the following case:
```
let(var, (↑deref)(it2))(if ·on_bc then 0 else ·var)
```
Then var should only be dereferenced in case `·on_bc` evalutes to False. Previously we just evaluated all values unconditionally:
```
let(_icdlv_1, ·it)(if ·on_bc then 0 else _icdlv_1)
```
Now we instead create a 0-ary lambda function for `_icdlv_1` and evaluate it when the value is needed.

```
let(var, (↑deref)(it2))(if ·on_bc then 0 else ·var)
```
Note that as a result we do evaluate the function multiple times. To avoid redundant recomputes usage of the common subexpression is therefor required.


havogt (approved):
Can you replace bc by something like: conditionally evaluate?

PR #1884, opened 2025-02-24
<feat>[cartesian] Debug Backend

twicki:
- PR Description:

    Description of the main changes with links to appropriate issues/documents/references/...
-->

## Description

This PR implements a plain python backend for debugging and rapid prototyping of features. 

The main difference between the debug backend and the numpy backend is that this backend is ment to be easier to step through with a debugger  to understand the generated code or to visually inspect the generated code because model developers are more familiar with with loop based implementations. 

The second big difference is that this backend directly represents the OIR without lowering to an additional lower level IR to visualize the resulting OIR a bit more directly.

This backend is not optimized for speed but rather for readabiltiy. 

## Requirements


FlorianDeconinck:
cscs-ci run

havogt (changes requested):
Just had a quick look and found some merge conflicts.

src/gt4py/cartesian/gtc/common.py
```diff
@@ -317,6 +317,18 @@ def zero(cls) -> CartesianOffset:
     def to_dict(self) -> Dict[str, int]:
         return {"i": self.i, "j": self.j, "k": self.k}
 
+    def to_str(self, dimensions: tuple[bool, bool, bool]) -> str:
```
romanc, line 320: Imo 

```python
offset.to_str((True, False, True))
```

feels a bit magical.

What is the goal of this function? For usage in the debug backend, see below. For debugging, `__str__` or `__repr__` could be useful to override.

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+            return f"np.{data_type.name.lower()}"
+        else:
+            return data_type.name.lower()
+
+    def visit_VariableKOffset(self, variable_k_offset: VariableKOffset, **_) -> str:
+        return f"i,j,k+int({self.visit(variable_k_offset.k)})"
+
+    def visit_CartesianOffset(self, cartesian_offset: CartesianOffset, **kwargs) -> str:
+        dimensions = kwargs["dimensions"]
+        return cartesian_offset.to_str(dimensions)
```
romanc, line 226: Imo `CartesianOffset` shouldn't need to provide a function such that we can use a one-liner here. If `offset.to_str()` was only written to be used here, I think we shouldn't add `to_str()` to the class and instead do the work manually here (based on the existing `offset.to_dict()` in combination with a dimension filter).

tests/cartesian_tests/integration_tests/multi_feature_tests/test_debug_backend.py
```diff
@@ -0,0 +1,366 @@
...
+# All rights reserved.
+#
+# This file is part of the GT4Py project and the GridTools framework.
+# GT4Py is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or any later
+# version. See the LICENSE.txt file at the top-level directory of this
+# distribution for a copy of the license or check <https://www.gnu.org/licenses/>.
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
```
romanc, line 21: This license block seems different than the other ones. That's probably not right

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -1013,11 +1014,20 @@ def visit_Constant(
...
+                    if isinstance(value, int):
+                        value_type = np.dtype(
+                            f"i{int(int(build_settings['literal_floating_point_precision'])/8)}"
+                        )
+                    else:
+                        value_type = np.dtype(
+                            f"f{int(int(build_settings['literal_floating_point_precision'])/8)}"
+                        )
+                else:
+                    value_type = np.dtype(type(value))
```
romanc, line 1030: Do we need frontend changes to add the debug backend?

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+    @staticmethod
+    def compute_extents(node: Stencil, **_) -> tuple[dict[str, Extent], dict[int, Extent]]:
+        ctx: StencilExtentComputer.Context = StencilExtentComputer().visit(node)
+        return ctx.fields, ctx.blocks
+
+    def initial_declarations(
+        self, stencil: Stencil, field_extents: dict[str, Extent]
+    ) -> dict[str, FieldDecl]:
+        self.body.append("# ===== Domain Description ===== #")
+        self.body.append("i_0, j_0, k_0 = 0,0,0")
```
romanc, line 90:
```suggestion
        self.body.append("i_0, j_0, k_0 = 0, 0, 0")
```
just to be consistent with whitespace?

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+                symbol_table[str(declaration.name)] = declaration
+        return symbol_table
+
+    def generate_run_function(self, stencil: Stencil):
+        function_signature = "def run(*"
+        args = []
+        for param in stencil.params:
+            args.append(self.visit(param))
+        function_signature = ",".join([function_signature, *args])
+        function_signature += ",_domain_, _origin_):"
```
romanc, line 127:
```suggestion
        function_signature += ",".join(args)
        function_signature += ",_domain_, _origin_):"
```
would this be more clear to read since we are relying on `+=` to add origin and domain?

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+            for section in loop.sections:
+                with self.create_k_loop_code(section, loop):
+                    for execution in section.horizontal_executions:
+                        with self.generate_ij_loop(block_extents, execution):
+                            self.visit(execution)
+
+    @contextmanager
+    def create_k_loop_code(self, section: VerticalLoopSection, loop: VerticalLoop) -> Generator:
+        loop_bounds: str = self.visit(section.interval, var="k", direction=loop.loop_order)
+        iterator = "1" if loop.loop_order != LoopOrder.BACKWARD else "-1"
```
romanc, line 146:
```suggestion
        increment = "1" if loop.loop_order == LoopOrder.FORWARD else "-1"
```
avoiding negations in conditions generally reduces cognitive load (at least in cases where both, if & else, branches are used)

(name change to be propagated within the function scope)

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+        loop_code = "for k in range(" + loop_bounds + "," + iterator + "):"
```
romanc, line 147: f-string?

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+            return ",".join(
+                [
+                    self.visit(interval.end, **kwargs) + "- 1",
+                    self.visit(interval.start, **kwargs) + "- 1",
+                ]
+            )
+        else:
+            return ",".join(
+                [self.visit(interval.start, **kwargs), self.visit(interval.end, **kwargs)]
+            )
```
romanc, line 197:
```suggestion
    def visit_Interval(self, interval: Interval, **kwargs):
        if kwargs["direction"] == LoopOrder.BACKWARD:
            return ",".join(
                [
                    self.visit(interval.end, **kwargs) + "- 1",
                    self.visit(interval.start, **kwargs) + "- 1",
                ]
            )

        return ",".join(
            [self.visit(interval.start, **kwargs), self.visit(interval.end, **kwargs)]
        )
```

No explicit `else` needed after `return` in `if`. Reduces the amount of indent needed. It might even make sense to change to "if direction == FORWARD:" because that case is simpler and the you get less indent levels for the "more complicated" case of `BACKWARD` looping.

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+        offset = [str(off) for off in field_offset] + ["0"] * (
+            1 + len(temporary_declaration.data_dims)
+        )
+        return f"{temporary_declaration.name} = Field.empty(({shape_decl}), {dtype}, ({', '.join(offset)}))"
+
+    def visit_DataType(self, data_type: DataType, **_) -> str:
+        if data_type not in {DataType.BOOL}:
+            return f"np.{data_type.name.lower()}"
+        else:
+            return data_type.name.lower()
```
romanc, line 219:
```suggestion
    def visit_DataType(self, data_type: DataType, **_) -> str:
        if data_type in {DataType.BOOL}:
            return data_type.name.lower()

        return f"np.{data_type.name.lower()}"
```

Handle special case first. Avoid `else` after `return` in `if` branch.

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+        return cartesian_offset.to_str(dimensions)
+
+    def visit_SymbolRef(self, symbol_ref: SymbolRef) -> str:
+        return symbol_ref
+
+    def visit_FieldAccess(self, field_access: FieldAccess, **_) -> str:
+        if str(field_access.name) in self.symbol_table:
+            dimensions = self.symbol_table[str(field_access.name)].dimensions
+        else:
+            dimensions = (True, True, True)
```
romanc, line 235: I think it would be more readable if we had a better type here. `(bool, bool, bool)` reads like voodoo magic in my eyes :mage:

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+            data_index_access = ",".join(
+                [self.visit(data_index) for data_index in field_access.data_index]
+            )
+            if offset_str == "":
+                full_string = field_access.name + f"[{data_index_access}]"
+            else:
+                full_string = field_access.name + "[" + offset_str + "," + data_index_access + "]"
+        else:
+            full_string = field_access.name + "[" + offset_str + "]"
+        return full_string
```
romanc, line 249: how about using early return statements instead?

```suggestion
        if field_access.data_index:
            data_index_access = ",".join(
                [self.visit(data_index) for data_index in field_access.data_index]
            )
            if offset_str == "":
                return field_access.name + f"[{data_index_access}]"

            return field_access.name + "[" + offset_str + "," + data_index_access + "]"

        return field_access.name + "[" + offset_str + "]"
```

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def visit_AssignStmt(self, assignment_statement: AssignStmt, **_):
+        self.body.append(
+            self.visit(assignment_statement.left) + "=" + self.visit(assignment_statement.right)
```
romanc, line 253: f-string?

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+            else:
+                literal_value = "False"
+        else:
+            literal_value = str(literal.value)
+
+        if literal.dtype.bit_count() != 4:
+            literal_code = f"{self.visit(literal.dtype)}({literal_value})"
+        else:
+            literal_code = literal_value
+        return literal_code
```
romanc, line 272: early returns seem like a good fit here

```suggestion
    def visit_Literal(self, literal: Literal, **_) -> str:
        if literal.dtype == DataType.BOOL:
            return "True" if literal.value == BuiltInLiteral.TRUE else "False"

        literal_value = str(literal.value)
        if literal.dtype.bit_count() != 4:
            return f"{self.visit(literal.dtype)}({literal_value})"

        return literal_value
```

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def visit_Cast(self, cast: Cast, **_) -> str:
+        return f"{self.visit(cast.dtype)}({self.visit(cast.expr)})"
+
+    def visit_HorizontalExecution(self, horizontal_execution: HorizontalExecution, **_):
+        for stmt in horizontal_execution.body:
+            self.visit(stmt)
```
romanc, line 279: nitpick
```suggestion
        for statement in horizontal_execution.body:
            self.visit(statement)
```

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+        )
+
+    def visit_CacheDesc(self, cache_descriptor: CacheDesc, **_):
+        raise NotImplementedError("Caches should never be visited in the debug backend")
+
+    def visit_IJCache(self, ij_cache: IJCache, **_):
+        raise NotImplementedError("Caches should never be visited in the debug backend")
+
+    def visit_KCache(self, k_cache: KCache, **_):
+        raise NotImplementedError("Caches should never be visited in the debug backend")
```
romanc, line 348: make these differentiable such  that - in case of error - it's clear which one triggered.

```suggestion
    def visit_CacheDesc(self, cache_descriptor: CacheDesc, **_):
        raise NotImplementedError("CacheDescriptors should never be visited in the debug backends")

    def visit_IJCache(self, ij_cache: IJCache, **_):
        raise NotImplementedError("IJCaches should never be visited in the debug backend.")

    def visit_KCache(self, k_cache: KCache, **_):
        raise NotImplementedError("KCaches should never be visited in the debug backend.")
```

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+from gt4py.eve import codegen
+from gt4py.eve.concepts import SymbolRef
+
+
+class DebugCodeGen(codegen.TemplatedGenerator, eve.VisitorWithSymbolTableTrait):
+    def __init__(self) -> None:
+        self.body = utils.text.TextBlock()
+        self.symbol_table: dict[str, FieldDecl] = {}
+
+    def visit_Stencil(self, stencil: Stencil, **_):
```
romanc, line 65: Some methods have return type annotations, others (like this one) don't. We currently don't enforce this through `mypy`

https://github.com/GridTools/gt4py/blob/3249aa5eb20980e3a6f4d0810577ecc30e177776/pyproject.toml#L227-L233

but maybe you can do another pass to add them consistently.

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,354 @@
...
+    def visit_ScalarDecl(self, scalar_declaration: ScalarDecl, **_) -> str:
+        return scalar_declaration.name
+
+    def visit_NativeFuncCall(self, native_function_call: NativeFuncCall, **_) -> str:
+        arglist = [self.visit(arg) for arg in native_function_call.args]
+        arguments = ",".join(arglist)
+        return f"ufuncs.{native_function_call.func.value}({arguments})"
+
+    def visit_UnaryOp(self, unary_operator: UnaryOp, **_) -> str:
+        return unary_operator.op.value + " " + self.visit(unary_operator.expr)
```
romanc, line 325: f-string?

src/gt4py/cartesian/utils/field.py
```diff
@@ -0,0 +1,75 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+
+import numbers
+from typing import Tuple
+
+import numpy as np
+
+
+class Field:
```
romanc, line 16: what's the intent of this class? Can it be re-used in other backends? If not, I'd suggest to move it into the debug backend folder.

tests/cartesian_tests/definitions.py
```diff
@@ -48,7 +48,9 @@ def _get_backends_with_storage_info(storage_info_kind: str):
 GPU_BACKENDS = _get_backends_with_storage_info("gpu")
 ALL_BACKENDS = CPU_BACKENDS + GPU_BACKENDS
 
-_PERFORMANCE_BACKEND_NAMES = [name for name in _ALL_BACKEND_NAMES if name not in ("numpy", "cuda")]
+_PERFORMANCE_BACKEND_NAMES = [
+    name for name in _ALL_BACKEND_NAMES if name not in ("numpy", "cuda", "debug")
+]
```
romanc, line 53: just a heads-up, this will generate merge conflicts with changes in PR https://github.com/GridTools/gt4py/pull/1885 (where I inlined the performance backend names).

tests/cartesian_tests/integration_tests/multi_feature_tests/test_debug_backend.py
```diff
@@ -0,0 +1,366 @@
...
+
+def test_simple_stencil():
+    field_in = gt_storage.ones(
+        dtype=np.float64, backend="debug", shape=(6, 6, 6), aligned_index=(0, 0, 0)
+    )
+    field_out = gt_storage.zeros(
+        dtype=np.float64, backend="debug", shape=(6, 6, 6), aligned_index=(0, 0, 0)
+    )
+
+    @gtscript.stencil(backend="debug")
```
romanc, line 38: General question (for the whole file): are these tests really specific to the debug backend? If not, would it make sense to run them parametrized on `ALL_BACKENDS` (and rename the file)?

romanc (commented):
Looking good in general. I left a bunch of inline comments. Most of them are simple code transformation that try to simplify and enhance readability.

src/gt4py/cartesian/backend/debug_backend.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def recursive_write(root_path: Path, tree: dict[str, Union[str, dict]]):
+    root_path.mkdir(parents=True, exist_ok=True)
+    for key, value in tree.items():
+        if isinstance(value, dict):
+            recursive_write(root_path / key, value)
+        else:
+            src_path = root_path / key
+            src_path.write_text(value)
```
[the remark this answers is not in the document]
twicki: :+1:

src/gt4py/cartesian/backend/debug_backend.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@register
+class DebugBackend(BaseBackend, CLIBackendMixin):
+    """Debug backend using plain python loops."""
+
+    name = "debug"
+    options: ClassVar[dict[str, Any]] = {
+        "oir_pipeline": {"versioning": True, "type": OirPipeline},
+        # TODO: Implement this option in source code
```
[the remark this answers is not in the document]
twicki: no, sorry. My bad

src/gt4py/cartesian/backend/debug_backend.py
```diff
@@ -0,0 +1,81 @@
...
+        "ignore_np_errstate": {"versioning": True, "type": bool},
+    }
+    storage_info = storage.layout.NaiveCPULayout
+    languages = {"computation": "python", "bindings": ["python"]}
+    MODULE_GENERATOR_CLASS = ModuleGenerator
+
+    def generate_computation(self) -> dict[str, Union[str, dict]]:
```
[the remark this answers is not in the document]
twicki: could be done

src/gt4py/cartesian/config.py
```diff
@@ -65,6 +65,7 @@
     "extra_link_args": extra_link_args,
     "parallel_jobs": multiprocessing.cpu_count(),
     "cpp_template_depth": os.environ.get("GT_CPP_TEMPLATE_DEPTH", GT_CPP_TEMPLATE_DEPTH),
+    "literal_floating_point_precision": os.environ.get("GT4PY_LITERAL_PRECISION", None),
```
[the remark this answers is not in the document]
twicki: This could be extracted out.

src/gt4py/cartesian/gtc/common.py
```diff
@@ -317,6 +317,18 @@ def zero(cls) -> CartesianOffset:
...
+        dimension_strings = []
+
+        if dimensions[0]:
+            dimension_strings.append(f"i + {self.i}")
+        if dimensions[1]:
+            dimension_strings.append(f"j + {self.j}")
+        if dimensions[2]:
+            dimension_strings.append(f"k + {self.k}")
```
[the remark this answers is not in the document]
twicki: there is no need. If you write `a = b + -c` the minus will be a unary operator on c and the result is as expected. we do sacrifice a bit of readability in the generated code for simplicity here, which I think is ok

src/gt4py/cartesian/gtc/debug/debug_codegen.py
twicki: we can do the work here. This is not used anywhere else immediately. I figured an `i+a, j+b, k+c` representation might be useful, but we can extract it if it is actually needed anywhere else

tests/cartesian_tests/integration_tests/multi_feature_tests/test_debug_backend.py
twicki: good catch

src/gt4py/cartesian/frontend/gtscript_frontend.py
twicki: same point as above, could be refactored out

src/gt4py/cartesian/gtc/debug/debug_codegen.py
twicki: There is the one awkwardness of it generating `def run(*,_domain_, _origin_)` in a non arg world with this approach

src/gt4py/cartesian/gtc/debug/debug_codegen.py
twicki: Does not work here, because parallel exists and also needs +1. But we can change the if logic if you want

src/gt4py/cartesian/gtc/debug/debug_codegen.py
twicki: done

src/gt4py/cartesian/gtc/debug/debug_codegen.py
twicki: Well, the dimension is stored in the OIR as such here: https://github.com/GridTools/gt4py/blob/70569bc5e917bf4000f6543573aadc5d31c86c13/src/gt4py/cartesian/gtc/oir.py#L114

The problem is that we need to carry the information of the sizing of the field upon initialization to the field-accesses.

What we could do is make the offset visiting default to a 3d field, does that sound ok?

src/gt4py/cartesian/gtc/debug/debug_codegen.py
twicki: does not quite work because we need to visit the dtype on bools also, but I pushed an adapted version

src/gt4py/cartesian/gtc/debug/debug_codegen.py
twicki: That's what we pay you for. 

Expanding it here makes sense, but stmt itself if very wide-spead in gt4py: 
image
but we can start cleaning up where it makes sense

src/gt4py/cartesian/gtc/debug/debug_codegen.py
twicki: done

src/gt4py/cartesian/utils/field.py
twicki: It can be re-used. For now it is hard-coded as code-generation in the numpy backend: https://github.com/GridTools/gt4py/blob/70569bc5e917bf4000f6543573aadc5d31c86c13/src/gt4py/cartesian/gtc/numpy/npir_codegen.py#L80

The goal would be to move this at some point in the future

src/gt4py/cartesian/gtc/ufuncs.py
```diff
@@ -37,6 +37,9 @@
 abs: np.ufunc = np.abs  # noqa: A001 [builtin-variable-shadowing]
 minimum: np.ufunc = np.minimum
 maximum: np.ufunc = np.maximum
+max: np.ufunc = np.maximum  # noqa: A001
```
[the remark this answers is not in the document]
twicki: this is so we can uniformly generate `ufunc.function_name`

tests/cartesian_tests/integration_tests/multi_feature_tests/test_debug_backend.py
twicki: We could think of generalizing this in the future, yes

src/gt4py/cartesian/gtc/debug/debug_codegen.py
romanc: oh well ... In that case, I'd suggest to log a task for OIR cleanup and just leave it as you wrote it.

src/gt4py/cartesian/utils/field.py
romanc: I guess it's out of scope for this PR. Can we log a task instead to do this in the future and maybe add a comment in code that this will be done?

src/gt4py/cartesian/gtc/debug/debug_codegen.py
romanc: hmm okay - then I guess this is as good as it gets. all good.

src/gt4py/cartesian/backend/debug_backend.py
```diff
@@ -0,0 +1,81 @@
...
+        computation_name = (
+            self.builder.caching.module_prefix
+            + "computation"
+            + self.builder.caching.module_postfix
+            + ".py"
+        )
```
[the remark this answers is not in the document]
romanc: Updated the knowledge base accordingly: [coding guidelines / string concatenation](https://geos-esm.github.io/SMT-Nebulae/backend/coding-guidelines/#string-concatenation)

src/gt4py/cartesian/backend/base.py
```diff
@@ -317,6 +317,15 @@ def make_module_source(self, *, args_data: Optional[ModuleData] = None, **kwargs
...
         return source
 
+    def recursive_write(self, root_path: pathlib.Path, tree: dict[str, Union[str, dict]]):
+        root_path.mkdir(parents=True, exist_ok=True)
+        for key, value in tree.items():
+            if isinstance(value, dict):
+                self.recursive_write(root_path / key, value)
+            else:
+                src_path = root_path / key
+                src_path.write_text(value)
```
romanc, line 327: This function used to be a standalone function. Now it is part of a base class without accessing any property or method of that class. Have you considered leaving it a standalone function? We could keep it in the same file or  create `src/gt4py/cartesian/backend/utils.py` and put it there. Another prime candidate to move to a potential `backend/utils.py` (in a follow-up PR) would be 

https://github.com/GridTools/gt4py/blob/b6a316285ad5825b9a9f2be5b3a8df36fa2450e1/src/gt4py/cartesian/backend/base.py#L433-L455

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,355 @@
...
+        self.body.dedent()
+        self.body.dedent()
+
+    def visit_While(self, while_node: While, **_) -> None:
+        while_condition = self.visit(while_node.cond)
+        while_code = f"while {while_condition}:"
+        self.body.append(while_code)
+        self.body.indent()
+        for statement in while_node.body:
+            self.visit(statement)
```
romanc, line 174: passing on/down the received `kwargs` is a common pattern in node visitors, e.g. `visit_Interval` below does that. Here `kwargs` aren't forwarded to the body of the while loop. Is that okay to do?

src/gt4py/cartesian/utils/field.py
romanc: Is there already an issue for this cleanup? I fear we might forget otherwise

src/gt4py/cartesian/backend/debug_backend.py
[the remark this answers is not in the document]
romanc: As we talked about getting rid of the `CLIBackendMixin` all together, let's leave this as-is and not put in extra work.

romanc (approved):
Looking nice and clean now in my eyes :) Just two small questions inline.

src/gt4py/cartesian/gtc/debug/debug_codegen.py
twicki: We could opt for generically forwarding all the kwargs all the time. I chose to, for the sake of clarity, only do the forwarding where it is neccesasry to pass in pre-compted information.

src/gt4py/cartesian/backend/base.py
twicki: I think since it is so tied to the backends and this is the only place this is used, adding it to the class gives it a more direct link to where the work is used.  I prefer this way to avoid generic `utils` files that have a tendency to grow big. If there are utilities that are used across larger modules, I see value that way.

We could make it a `staticmethod` though

src/gt4py/cartesian/utils/field.py
twicki: https://github.com/GridTools/gt4py/issues/1902

romanc:
cscs-ci run

romanc:
cscs-ci run

FlorianDeconinck:
cscs-ci run

romanc:
cscs-ci run

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,355 @@
...
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+
+from collections.abc import Generator
+from contextlib import contextmanager
+
+from gt4py import eve
+from gt4py.cartesian import utils
+from gt4py.cartesian.gtc.common import (
```
havogt, line 15: Since this is a new file, let's impose strictly our coding guidelines. Only import modules, see https://google.github.io/styleguide/pyguide.html#22-imports.

Here I would do
```suggestion
from gt4py.cartesian.gtc import common as gtc_common
```
to distinguish from other `common` modules.

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,355 @@
...
+    CartesianOffset,
+    DataType,
+    FieldAccess,
+    HorizontalInterval,
+    HorizontalMask,
+    LevelMarker,
+    LoopOrder,
+    While,
+)
+from gt4py.cartesian.gtc.definitions import Extent
```
havogt, line 27:
```suggestion
from gt4py.cartesian.gtc import definitions as gtc_definitions
```

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,355 @@
...
+    ScalarDecl,
+    Stencil,
+    Temporary,
+    TernaryOp,
+    UnaryOp,
+    UnboundedInterval,
+    VariableKOffset,
+    VerticalLoop,
+    VerticalLoopSection,
+)
```
havogt, line 54:
```suggestion
from gt4py.cartesian.gtc import oir
```
etc

src/gt4py/cartesian/utils/field.py
```diff
@@ -0,0 +1,75 @@
...
+#
+# Copyright (c) 2014-2024, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+
+import numbers
+from typing import Tuple
```
havogt, line 11: fyi, this import is ok, as pure typing imports are excluded from the rule, see https://google.github.io/styleguide/pyguide.html#2241-exemptions

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -0,0 +1,355 @@
...
+from gt4py.cartesian.gtc.passes.oir_optimizations.utils import StencilExtentComputer
+from gt4py.eve import codegen
+from gt4py.eve.concepts import SymbolRef
+
+
+class DebugCodeGen(codegen.TemplatedGenerator, eve.VisitorWithSymbolTableTrait):
```
havogt, line 60: - It seems you are not using the SymbolTableTrait, but instead of managing symbols manually. What's the reason?
- You are also not using the TemplatedGenerator, only a normal NodeVisitor as far as I can see.

I could give an intro to both in case there are questions.

havogt (changes requested):
I only had a closer look now. Main points:
- import style
- how to use the helpers on top of eve.NodeVisitor.

havogt:
cscs-ci run

havogt (approved):
lgtm

romanc:
cscs-ci run

PR #1886, opened 2025-02-25
test[next]: initialize output array in dace-gtir subdomain tests

edopao:
Some GTIR DaCe tests verify that only a subdomain is computed, and the remaining points in the output field should remain unchanged. These tests were not properly initializing the output array. In some rare case, when the uninitialized data was `NaN`, the result validation failed.


havogt (commented):
Why did the tests pass sometimes?

edopao:
> Why did the tests pass sometimes?

I do not have the answer. It worked in 99.9% of the runs. Probably `NaN` is just an attribute (like a flag bit) on a regular data object, and in our runs the uninitialized value stored in the data object happened to satisfy the comparison, but the value happened to be too small or too large for the default tolerance.

havogt (approved):
lgtm

PR #1889, opened 2025-02-26
fix[dace][next]: Updated `gt_auto_optimize()`

philip-paul-mueller:
The problem was that in the auto optimizer the "set stride correctly" function was called after the "go to GPU transformation". The GPU transformation will call `CopyToMap` transformation, this is needed such that we can set the GPU block size and their order. However, the decision if we call `CopyToMap` also depends on the strides, so we need to set them first.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
@@ -66,10 +66,11 @@ def gt_auto_optimize(
         one with stride one.
     5. If requested the function will now apply loop blocking, on the dimension
         indicated by `leading_dim`.
-    6. If requested the SDFG will be transformed to GPU. For this the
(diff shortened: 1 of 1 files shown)


edopao (approved):

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -66,10 +66,11 @@ def gt_auto_optimize(
         one with stride one.
     5. If requested the function will now apply loop blocking, on the dimension
         indicated by `leading_dim`.
-    6. If requested the SDFG will be transformed to GPU. For this the
+    6. The strides of temporaries are set to match the compute order..
```
edopao, line 69:
```suggestion
    6. The strides of temporaries are set to match the compute order.
```
A suggestion for a future PR, you do not need to fix it in this PR.

PR #1890, opened 2025-02-28
feat[next][dace]: Added `DistributedGlobalSelfCopyElimination`

philip-paul-mueller:
The transformation `DistributedGlobalSelfCopyElimination` is very similar to the `GT4PyGlobalSelfCopyElimination` but they target slightly different cases.
The transformation `GT4PyGlobalSelfCopyElimination`, which is already there, handles the pattern `(G) -> (T) -> (G)`, i.e. the global data `G` is copied into the transient `T` is then immediately copied back into `G`.
Because of ADR-18 we know that this has no effect, because `G` is used as input and output and must therefore be point wise, so `G[i, j]` in the output can only be `G[i, j]` at the beginning.

The new transformation `GT4PyGlobalSelfCopyElimination` handles a different case, it looks for patterns `(G) -> (T)` and `(T) -> (G)`, which is essentially the same, but this time the definition of `T` and the write back of `T` into `G` does not need to be in the same state.

In the long run, the two transformation should be combined.


src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+            aname
+            for aname, desc in sdfg.arrays.items()
+            if not desc.transient
+            and isinstance(desc, dace_data.Array)
+            and not isinstance(desc, dace_data.View)
+        ]
+
+        candidates: dict[str, set[str]] = dict()
+        for gname in global_data:
+            possible_ts = self._classify_candidate(sdfg, gname, access_states)
```
edopao, line 150:
```suggestion
            candidate_tnames = self._classify_candidate(sdfg, gname, access_states)
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+
+    def _classify_candidate(
+        self,
+        sdfg: dace.SDFG,
+        gname: str,
+        access_states: dict[str, set[dace.SDFGState]],
+    ) -> Optional[set[str]]:
+        """The function tests if the global data `gname` can be handled.
+
+        It essentially checks if all conditions above, which is that the only
```
edopao, line 165:
```suggestion
        It essentially checks all conditions above, which is that the global is
        only written through transients that are fully defined by the data itself.
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+        writes to it are through transients that are fully defined by the data
+        itself.
+
+        The function returns `None` if `gname` can not be handled by the function.
+        If `gname` can be handled the function returns a set of all data descriptors
+        that acts as distributed buffers.
```
edopao, line 171:
```suggestion
        that act as distributed buffers.
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+                    if not possible_t_desc.transient:
+                        continue  # we must write into a transient.
+                    if isinstance(possible_t_desc, dace_data.View):
+                        continue  # We must write into an array and not a view.
+                    if not isinstance(possible_t_desc, dace_data.Array):
+                        continue
+
+                    # Currently we do not handle the pattern `(T) -> (G) -> (T)`,
+                    #  see `_remove_writes_to_global()` for more, thus we filter
+                    #  these pattern here.
```
edopao, line 234:
```suggestion
                    #  this pattern here.
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+                    ):
+                        return None
+
+                    # Now add the data to the list of data that reads from `G`.
+                    reads_from_g.add(possible_t.data)
+
+        if len(writes_to_g) == 0:
+            return None
+
+        # Now every write to `G` must come from an access node that was created by
```
edopao, line 248:
```suggestion
        # Now every write to `G` necessarily comes from an access node that was created by
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+        #  direct a read from `G`. We can do this by ensuring that `writes_to_g` is a
```
edopao, line 249:
```suggestion
        # a direct read from `G`. We ensure this by checking that `writes_to_g` is a
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+        #  subset of `reads_to_g`.
+        #  Note that all `T`, which might not be unique, which happens in case
```
edopao, line 251:
```suggestion
        # Note that the `T` nodes might not be unique, which happens in case
        # of separate memlets for different subsets.
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+        #  of different subsets, are contained in `
+        if not writes_to_g.issubset(reads_from_g):
+            return None
+
+        # If we have branches, it might be that an `T` is defined differently depending
```
edopao, line 256:
```suggestion
        # If we have branches, it might be that different data is written to `T` depending
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+        #  on which branch is selected, i.e. `T = G if cond else foo(A)`. For that
+        #  reason we must now check that `G` is the only data source of `T`, but this
+        #  time we must do the check `T`. Note we only have to remove that particular
```
edopao, line 259:
```suggestion
        #  time we must do the check on `T`. Note we only have to remove the particular access node
        #  to `T` where `G` is the only data source, while we keep the other access nodes.
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+            # Now we look for all writes to `tname` and remove them, since there
+            #  are no reads.
+            for state in access_states[tname]:
+                for dnode in list(state.data_nodes()):
+                    if dnode.data == tname:
+                        # We have to store potential sources nodes, which is `G`.
+                        #  This is because the local `G` node could become isolated.
+                        #  We do not need to consider the outgoing edges, because
+                        #  they are reads which we have handled above.
+                        neigbourhood.update((state, iedge.src) for iedge in state.in_edges(dnode))
```
edopao, line 395: Am I wrong or here we could assert `isinstance(iedge.src, dace_nodes.AccessNode) and iedge.src.data == gname` ?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -0,0 +1,559 @@
...
+                        state.remove_node(dnode)
+                        removed_t.add(dnode.data)
+
+        # Check if an isolated node was created.
+        removed_isolated_nodes: set[dace_nodes.Node] = set()
+        for state, nh_node in neigbourhood:
+            assert isinstance(nh_node, dace_nodes.AccessNode) and nh_node.data == gname
+            if (nh_node not in removed_isolated_nodes) and (state.degree(nh_node) == 0):
```
edopao, line 403: and we could `assert state.in_degree(nh_node) == 0` and only check `state.out_degree(nh_node) == 0`

edopao (commented):
LGTM

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: This is a very good idea.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: You are right.
The assert is there, but in the loop that actually removes the isolated nodes.
My problem with the assert here was that I would either have to use a `for` loop or go through the set of edges twice.
The `for` loop I did not like because it adds to much indentation and the going through the neighbours is potentially expensive.
So I originally put the assert later on line 405.
However, I agree that having the assert where the problem originates is better.
I settled with the `for` loop.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: I do not really understand that comment.

I am not sure if I understand you correctly, but if you propose to remove `removed_isilated_nodes`?
In the current code we can not do that, in the current version.
But your suggestion made me thinking and I rewrite that part.

However, I am not fully sure if that is better, because in Python the scope of constructs is so clear, curly brackets would help, what do you think?

src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
```diff
@@ -58,7 +58,11 @@ def gt_simplify(
 
     Further, the function will run the following passes in addition to DaCe simplify:
     - `GT4PyGlobalSelfCopyElimination`: Special copy pattern that in the context
-        of GT4Py based SDFG behaves as a no op.
+        of GT4Py based SDFG behaves as a no op, i.e. `(G) -> (T) -> (G)`.
+    - `DistributedGlobalSelfCopyElimination`: Very similar to
```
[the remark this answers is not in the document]
philip-paul-mueller: I agree the names are very bad and your names are very good.
They are actually so good that I now want to keep `SingleStateGlobalSelfCopyElimination` such that I can keep `MultiStateGlobalSelfCopyElimination` and do not have to rename it to `GlobalSelfCopyElimination`.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
edopao: I mean: there should never be a write to the global, since this is a candidate. Instead, there might be a read, that is why we should check the output degree.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
edopao: I see, please do as you prefer.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: You can not assert on degree, because it might be that the neighbour node `nh_node` has not become isolated.
It might be that it is read somewhere else.
Consider this example, when `w_concorr_c` is `G` and `__tmp18` is `T`, then by deleting the `T` node will not make the `G` node isolated.

2025-02-28-135103_1073x644_scrot

edopao: I wrote **input** degree above, but maybe it is too restrictive:

```
if (nh_node not in removed_isolated_nodes) and (state.out_degree(nh_node) == 0):
    assert state.in_degree(nh_node) == 0
```

philip-paul-mueller: Not fully sure.
I do not have a counter example, but in most cases you might be right.
However, I would keep the `degree == 0` test since we are interested in removing isolated nodes.

edopao: Agreed.

edopao (approved):

PR #1893, opened 2025-03-02
next: runtime field domains

tehrengruber:


src/gt4py/next/iterator/builtins.py
```diff
@@ -402,6 +402,11 @@ def bool(*args):  # noqa: A001 [builtin-variable-shadowing]
     raise BackendNotSelectedError()
 
 
+@builtin_dispatch
+def get_domain(*args):
```
havogt:
```suggestion
def get_range(*args):
```
probably?
or 
```suggestion
def get_domain_range(*args):
```

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -37,11 +37,7 @@ def decorated_program(
         if out is not None:
             args = (*args, out)
 
-        if fun.implicit_domain:
-            # Generate implicit domain size arguments only if necessary
-            size_args = arguments.iter_size_args(args)
-            args = (*args, *size_args)
-
+        # TODO: this doesn't belong here and should by done in the dace backend
```
havogt: we are in the dace backend

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -271,6 +273,17 @@ def visit_Program(self, node: gtfn_ir.Program, **kwargs: Any) -> Union[str, Coll
     #include <gridtools/fn/sid_neighbor_table.hpp>
     #include <gridtools/stencil/global_parameter.hpp>
     
+    // TODO(tehrengruber): This should disappear as soon as we introduce a proper builtin.
+    namespace gridtools::fn {
+        // TODO(tehrengruber): `typename gridtools::sid::lower_bounds_type<S>, typename gridtools::sid::upper_bounds_type<S>`
+        // fails as type used for index calculations in gtfn differs
```
havogt: let's discuss this quickly
(and let's place the TODOs outside of the generated code)

havogt:
One more thing: can you add an ADR summarizing our discussion regarding the builtin and the gtfn domain etc. I.e. document why the current state is like it as and the direction we would like to go.

PR #1807, opened 2025-01-20
test[cartesian]: Increased coverage for horizontal regions

romanc:
## Description

Working on the gt4py/dace bridge showed that code coverage for horizontal regions is low. In particular (code generation) tests for conditionals inside horizontal regions were missing.

## Requirements

- [x] All fixes and/or new features come with corresponding tests.
Added test cases.
- [ ] Important design decisions have been documented in the appropriate ADR inside the [docs/development/ADRs/](docs/development/ADRs/Index.md) folder.
N/A

Issue #1829, opened 2025-01-28
[build] Pending enhancements to development infrastructure


egparedes:
Track possible enhancements to  the development infrastructure after the  #1813 PR bringing new tools (e.g. `uv`, `nox`) is merged.

- [ ] Bump `ruff` to the newest version and tackle the format changes it brings
- [ ] Add `tests` folder to pre-commit `ruff` hook
- [ ] Add more linting rules to `ruff` config. Pointers:
    -  [ruff config in gridtools/jace](https://github.com/GridTools/jace/blob/main/pyproject.toml)
- [ ] Bump `tach` version and try to get rid of the circular dependency issue by creating finer grain modules
- [ ] Add versioning infrastructure relying on git tags as the source of truth for versions (e.g. `versioningit`) 
- [ ] Review the Dockerfiles used in CI and clean them up to leverage `uv` when possible. Some pointers:
    - Use `.python-version` file in the repo as python version selector for the container image 
    - https://hynek.me/articles/docker-uv/
    - https://pythonspeed.com/articles/uv-python-production/
 - [ ] Fix doc generation settings and add CI action to test it
 - [ ] Add `macos` system to the daily CI action
 - [ ] ReDdiscuss the strategy to use for tests with optional dependencies like `dace`
     - use only `importorskip()`
     - use only custom marks
     - use both combined
     - other options

romanc:
@egparedes in PR https://github.com/GridTools/gt4py/pull/1894#discussion_r1987548296 the question came up if we could enforce consistent usage of either `Union[X, Y]` or `X | Y` with `ruff` (same for `list[T]` vs `List[T]` and friends). This wasn't an option before dropping `3.8` and `3.9` support. I half-way remember that `ruff` has this feature to auto-upgrade code and do such code transformations, but I don't seem to find it anymore.

In any case, would it be okay to add that discussion to the list here? Or does it warrant a separate issue?

egparedes:
@romanc Yes, I agree, we should upgrade the codebase to drop the old type syntax, since it's not longer needed. I added a new point to the topic of adding more strict checks to the `ruff` configuration.

PR #1894, opened 2025-03-03
refactor[cartesian]: Dace backend: expose control flow

romanc:
## Description

This PR refactors the GT4Py/DaCe bridge to expose control flow elements (`if` statements and `while` loops) to DaCe. Previously, the whole contents of a vertical loop was put in one big Tasklet. With this PR, that Tasklet is broken apart in case control flow is found such that control flow is visible in the SDFG. This allows DaCe to better analyze code and will be crucial in future (within the current milestone) performance optimization work.

The main ideas in this PR are the following

1. Introduce `oir.CodeBlock` to recursively break down `oir.HorizontalExecution`s into smaller pieces that are either code flow or evaluated in (smaller) Tasklets.
2. Introduce `dcir.Condition`and `dcir.WhileLoop` to represent if statements and while loops that are translated into SDFG states. We keep the current `dcir.MaskStmt` / `dcir.While` for if statements / while loops inside horizontal regions, which aren't yet exposed to DaCe (see https://github.com/GridTools/gt4py/issues/1900).
3. Add support for `if` statements and `while` loops in the state machine of `sdfg_builder.py`
4. We are breaking up vertical loops inside stencils in multiple Tasklets. It might thus happen that we write a "local" scalar in one Tasklet and read it in another Tasklet (downstream). We thus create output connectors for all scalar writes in a Tasklet and input connectors for all reads (unless previously written in the same Tasklet). 
5. Memlets can't be generated per horizontal execution anymore and need to be more fine grained. `TaskletAccessInfoCollector` does this work for us, duplicating some logic in `AccessInfoCollector`. A refactor task has been logged to fix/re-evaluate this later.

This PR depends on the following (downstream) DaCe fixes

- https://github.com/spcl/dace/pull/1954
- https://github.com/spcl/dace/pull/1955

which have been merged by now.

Follow-up issues

- unrelated changes have been moved to https://github.com/GridTools/gt4py/pull/1895
- https://github.com/GridTools/gt4py/issues/1896
- https://github.com/GridTools/gt4py/issues/1898
- https://github.com/GridTools/gt4py/issues/1900

Related issue: https://github.com/GEOS-ESM/NDSL/issues/53

## Requirements

  Added new tests and increased coverage of horizontal regions with PRs #1807 and #1851.
  Docs are [in our knowledge base](https://geos-esm.github.io/SMT-Nebulae/technical/backend/dace-bridge/) for now. Will be ported.


src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -145,16 +351,108 @@ def visit_Tasklet(
...
+            sdfg=sdfg_ctx.sdfg,
         )
 
+        # We are breaking up vertical loops inside stencils in multiple Tasklets
+        # It might thus happen that we write a "local" scalar in one Tasklet and
+        # read it in another Tasklet (downstream).
+        # We thus create output connectors for all writes to scalar variables
+        # inside Tasklets. And input connectors for all scalar reads unless
+        # previously written in the same Tasklet. DaCe's simplify pipeline will get
+        # rid of any dead dataflow introduced with this general approach.
```
FlorianDeconinck, line 363: We need a unit test on this bad boy

src/gt4py/cartesian/gtc/dace/expansion/tasklet_codegen.py
```diff
@@ -62,17 +56,30 @@ def _visit_offset(
         res = dace.subsets.Range([r for i, r in enumerate(ranges.ranges) if int_sizes[i] != 1])
         return str(res)
 
-    def visit_CartesianOffset(self, node: common.CartesianOffset, **kwargs: Any) -> str:
+    def visit_CartesianOffset(
+        self, node: common.CartesianOffset, explicit=False, **kwargs: Any
+    ) -> str:
+        # If called from the explicit pass we need to be add manually the relative indexing
+        if explicit:
```
FlorianDeconinck, line 63: Fold in `explicit_indexing` new function

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -260,6 +555,9 @@ def visit_ComputationState(
         **kwargs: Any,
     ) -> None:
         sdfg_ctx.add_state()
+
+        # Remove node_ctx from **kwargs in case it exists. We are building a new one.
+        kwargs.pop("node_ctx", None)
```
FlorianDeconinck, line 560: That sounds... scary? We loose that context for the downstream?

FlorianDeconinck (commented):
Few comments on a first read

src/gt4py/cartesian/gtc/dace/daceir.py
```diff
@@ -730,7 +727,8 @@ class Literal(common.Literal, Expr):
 
 
 class ScalarAccess(common.ScalarAccess, Expr):
-    pass
+    is_target: bool
```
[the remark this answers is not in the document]
romanc: It seems tempting. I would argue against since there's no precedence in `gtc.common`. `ScalarAccess` and `IndexAccess` are defined as separate entities

https://github.com/GridTools/gt4py/blob/c610561e918a0edc3fe5d0fb411ac4eae08c0cf0/src/gt4py/cartesian/gtc/common.py#L334-L343

without a base class to derive `name` and `kind` from. If we want to looks at that class hierarchy, I suggest we do it in a separate PR and not just for the DaCe backend.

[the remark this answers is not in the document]
FlorianDeconinck: Fair. LLVM 13 had split their `verb` and `location` in their IR and I liked that. You had `Access` `Write` `Read` `Store` and `Local` `Global` `Register` that then would combine up. I don't know if they carried it (it's LLVM IR so probably not)

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -171,9 +469,6 @@ def visit_Tasklet(
             symtable=symtable,
             **kwargs,
         )
-        StencilComputationSDFGBuilder._add_empty_edges(
```
twicki, line 174: why is this no longer needed?

src/gt4py/cartesian/gtc/dace/expansion/tasklet_codegen.py
```diff
@@ -62,17 +56,34 @@ def _visit_offset(
...
+            return self._explicit_indexing(node, **kwargs)
+
         return self._visit_offset(node, **kwargs)
 
-    def visit_VariableKOffset(self, node: dcir.VariableKOffset, **kwargs: Any) -> str:
+    def visit_VariableKOffset(
+        self, node: dcir.VariableKOffset, explicit=False, **kwargs: Any
+    ) -> str:
+        if explicit:
```
twicki, line 76: @FlorianDeconinck are we planning on supporting explicit, variable offsets?

src/gt4py/cartesian/gtc/dace/utils.py
```diff
@@ -347,6 +347,165 @@ def compute_dcir_access_infos(
...
+class TaskletAccessInfoCollector(eve.NodeVisitor):
+    @dataclass
+    class Context:
+        axes: dict[str, list[dcir.Axis]]
+        access_infos: dict[str, dcir.FieldAccessInfo] = field(default_factory=dict)
+
+    def __init__(
+        self, collect_read: bool, collect_write: bool, *, horizontal_extent, k_interval, grid_subset
+    ):
+        self.collect_read: bool = collect_read
```
twicki, line 359: I think these type annotations are not needed if the args coming in are typed. But if you like it for clarity, we can keep them

src/gt4py/cartesian/gtc/dace/expansion/tasklet_codegen.py
FlorianDeconinck: Hmm, this might be aggregating more than Variable "offset", but variable access altogether. 

So we shouldn't, but I wonder if the implementation is not aggregating too much use case here

src/gt4py/cartesian/gtc/dace/daceir.py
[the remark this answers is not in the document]
romanc: I made a note in https://github.com/GridTools/gt4py/issues/1898 to revisit this in the future.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
romanc: udpated

src/gt4py/cartesian/gtc/dace/daceir.py
```diff
@@ -730,7 +727,8 @@ class Literal(common.Literal, Expr):
...
+    original_name: Optional[str] = None
```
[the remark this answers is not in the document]
romanc: We use `original_name` as part of handling "local"[^1] temporaries that are potentially written in one Tasklet and later read in another one. Not every scalar access is a local temporary. For example, scalar stencil parameters "globally" (as in through the stencil) available and don't need/have an `original_name`.

[^1]: "local" as in local within the horizontal execution, which might be split (with this PR) in more than one Tasklet.

src/gt4py/cartesian/gtc/dace/daceir.py
```diff
@@ -744,7 +742,12 @@ def no_casts_in_offset_expression(self, _: datamodels.Attribute, expression: Exp
 
 
 class IndexAccess(common.FieldAccess, Expr):
-    offset: Optional[Union[common.CartesianOffset, VariableKOffset]]
+    # ScalarAccess used for field index
+    offset: Optional[common.CartesianOffset | Literal | ScalarAccess | VariableKOffset]
+    is_target: bool
+
+    # Use to access as a full field with explicit indices
+    explicit_indices: Optional[list[Literal | ScalarAccess | VariableKOffset]] = None
```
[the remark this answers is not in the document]
romanc: :+1:  added  as doc-string with with the proper `"""` triple quotes.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -39,54 +40,96 @@
...
+            # Find `tasklet_symbol` in dcir_statements because we can't know (from the oir statements)
+            # where the tasklet boundaries will be. Consider
+            #
+            # with computation(PARALLEL), interval(...):
+            #   statement1
+            #   if condition:
+            #     statement2
+            #   statement3
+            #
+            # statements 1 and 3 will end up in the same block but aren't in the same tasklet.
```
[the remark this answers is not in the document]
romanc: Added as "... will end up in the same CodeBlock"

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
romanc: This was used in situations were there was

- only one Tasklet inside one (or multiple) maps
- all inputs of the Tasklet were directly connected to the outputs of the map entry  and all outputs of the Tasklet were directly connected to the map exit

With this PR we don't generate that kind of SDFGs anymore. We always wrap Tasklets (inside maps) in a state and add access nodes for Tasklet inputs & outputs. We don't loose anything by doing so - just one less special case.

src/gt4py/cartesian/gtc/dace/utils.py
romanc: I like them for declaring & defining class variables (like in this case). If I were to declare `self.collect_read: bool` outside the `__init__` function, I would also repeat the type.

src/gt4py/cartesian/gtc/dace/daceir.py
```diff
@@ -842,33 +845,145 @@ class IterationNode(eve.Node):
...
+    condition: Tasklet
+    body: list[ComputationState | Condition | WhileLoop]
+
+    @datamodels.validator("condition")
+    def condition_has_boolean_expression(
+        self, attribute: datamodels.Attribute, tasklet: Tasklet
+    ) -> None:
+        assert isinstance(tasklet, Tasklet)
+        assert len(tasklet.stmts) == 1
+        assert isinstance(tasklet.stmts[0], AssignStmt)
```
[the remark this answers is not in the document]
romanc: As discussed, this Tasklet is used to evaluate the condition, which is then used in the guard state.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -318,167 +367,348 @@ def visit_FieldAccess(
...
-        to make sure all side effects are being properly resolved. Frontend checks ensure that no
-        parallel code issues sips here.
-        """
+    ) -> dcir.IndexAccess | dcir.ScalarAccess:
+        """Generate the relevant accessor to match the memlet that was previously setup."""
+
+        # Distinguish between writing to a variable and reading a previously written variable.
+        # In the latter case (read after write), we need to read from the "gtOUT__" symbol.
+        is_write = is_target
+        is_target = is_target or (
```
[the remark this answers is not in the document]
romanc: As discussed, the case where this is true is when we are on the RHS reading variable that was previously written win the same Tasklet, e.g.

The code

```python
A *= A
B = A + 1
```

will be translated to the following Tasklet code

```python
gtOUT__A = gtIN__A * 2
gtOUT__B = gtOUT_A + 1
```

where the Tasklet gets an input connector `A` -> `gtIN__A` and two output connectors `gtOUT__A` -> `A` and `gtOUT__B` -> `B`. When we visit assignment statements, we always evaluate the RHS first not to have the targets from the LHS already in the list of targets. So when `A` is visited in the first statement, we are on the RHS and `A` is not a target. Then, we are on the LHS and `A` is a target. When we visit the second assignment statement, `A` is on the RHS and it's not a target. However, we need to read from the previously written `gtOUT__A`. We thus set `is_target=True` while `is_write` remains `False`.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -318,167 +367,348 @@ def visit_FieldAccess(
...
+    ) -> None:
+        for memlet in memlets:
             """
             This loop handles the special case of a tasklet performing array access.
             The memlet should pass the full array shape (no tiling) and
             the tasklet expression for array access should use all explicit indexes.
             """
-            array_ndims = len(global_ctx.arrays[memlet.field].shape)
-            field_decl = global_ctx.library_node.field_decls[memlet.field]
+            array_ndims = len(global_context.arrays[memlet.field].shape)
```
FlorianDeconinck, line 707: This is clearly symptomatic of a bigger issue around `memlet` size and `indexing` in general. Not to be changed here, but let it be written in the larger documentation that we should not have to fix memlets - this should be dealt with before.

FlorianDeconinck (commented):
LGTM - will require a re-read with the documentation you are writing for a go-ahead

romanc:
Two things for reviewers

1. There's technical docs available in https://geos-esm.github.io/SMT-Nebulae/technical/backend/dace-bridge/. Feedback welcome if you have any.
2. I pushed a small PR today to re-organize all the prefix strings that we use in the bridge.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -490,22 +720,171 @@ def visit_HorizontalExecution(
...
+
+        # Gather all statements that aren't control flow (e.g. everything except Condition and WhileLoop),
+        # put them in a tasklet, and call "to_state" on it.
+        # Then, return a new list with types that are either ComputationState, Condition, or WhileLoop.
+        dace_nodes: list[dcir.ComputationState | dcir.Condition | dcir.WhileLoop] = []
+        current_block: list[dcir.Stmt] = []
+        for index, statement in enumerate(statements):
+            is_control_flow = isinstance(statement, (dcir.Condition, dcir.WhileLoop))
+            if not is_control_flow:
+                current_block.append(statement)
```
philip-paul-mueller, line 797: I think `current_block` is not a good name, first to me it indicates only one block but not multiples (which you did for the `dace_nodes`).
Also "current" is too general in my view okay, I have to say that `no_control_blocks` is also not the best, but I would say it is much more informative.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -490,22 +720,171 @@ def visit_HorizontalExecution(
...
+                self._fix_memlet_array_access(
+                    tasklet=tasklet,
+                    memlets=[*read_memlets, *write_memlets],
+                    global_context=global_ctx,
+                    symbol_collector=symbol_collector,
+                    targets=targets,
+                    **kwargs,
+                )
+
+                dace_nodes.append(*self.to_state(tasklet, grid_subset=iteration_ctx.grid_subset))
```
philip-paul-mueller, line 835: I also do not understand the logic of this append.
I know this is most likely wrong, but I have the feeling that you add the same thing, in some cases, twice to `dace_nodes`.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -490,22 +720,171 @@ def visit_HorizontalExecution(
...
+        expansion_items = global_ctx.library_node.expansion_specification[stages_idx + 1 :]
+
+        iteration_ctx = iteration_ctx.push_axes_extents(
+            {k: v for k, v in zip(dcir.Axis.dims_horizontal(), extent)}
+        )
+        iteration_ctx = iteration_ctx.push_expansion_items(expansion_items)
+        assert iteration_ctx.grid_subset == dcir.GridSubset.single_gridpoint()
+
+        code_block = oir.CodeBlock(body=node.body, loc=node.loc, label=f"he_{id(node)}")
+        targets: list[oir.FieldAccess | oir.ScalarAccess] = list()
```
philip-paul-mueller, line 872:
```suggestion
        targets: list[oir.FieldAccess | oir.ScalarAccess] = []
```

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -490,22 +720,171 @@ def visit_HorizontalExecution(
...
+            targets=targets,
+            **kwargs,
+        )
 
         for item in reversed(expansion_items):
             iteration_ctx = iteration_ctx.pop()
-            dcir_node = self._process_iteration_item(
-                [dcir_node],
+            dcir_nodes = self._process_iteration_item(
+                dcir_nodes,
```
philip-paul-mueller, line 887: In my opinion something deserves some kind of explanation, especially if you perform the iteration in reverse.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -36,18 +40,17 @@ class SDFGContext:
...
         state_stack: List[dace.SDFGState] = dataclasses.field(default_factory=list)
 
-        def add_state(self):
-            new_state = self.sdfg.add_state()
+        def add_state(self, label: Optional[str] = None) -> None:
+            new_state = self.sdfg.add_state(label=label)
             for edge in self.sdfg.out_edges(self.state):
                 self.sdfg.remove_edge(edge)
                 self.sdfg.add_edge(new_state, edge.dst, edge.data)
             self.sdfg.add_edge(self.state, new_state, dace.InterstateEdge())
```
philip-paul-mueller, line 48: There is an `SDFGState.add_state_before()` call.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -36,18 +40,17 @@ class SDFGContext:
         state: dace.SDFGState
         state_stack: List[dace.SDFGState] = dataclasses.field(default_factory=list)
 
-        def add_state(self):
-            new_state = self.sdfg.add_state()
+        def add_state(self, label: Optional[str] = None) -> None:
```
philip-paul-mueller, line 43: This might be outside of the scope of the PR, but the name of this function.
As it does not add a state, it adds a state before the current had state (this behavior is not documented anywhere you have to read the source for that).
It would not be that bad, however, the `SDFG.add_state()` function is also called directly, whcih might lead to confusion.
So I would recommend to rename this function to something like `push_state()`.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -36,18 +40,17 @@ class SDFGContext:
...
             self.sdfg.add_edge(self.state, new_state, dace.InterstateEdge())
             self.state = new_state
-            return self
 
-        def add_loop(self, index_range: dcir.Range):
-            loop_state = self.sdfg.add_state()
-            after_state = self.sdfg.add_state()
+        def add_loop(self, index_range: dcir.Range) -> None:
+            loop_state = self.sdfg.add_state("loop_state")
+            after_state = self.sdfg.add_state("loop_after")
```
philip-paul-mueller, line 53: In an SDFG all states needs to have a unique name.
This means inside one SDFG there can only be one loop, which might be okay for you but you should add some name mangling scheme.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -75,9 +78,126 @@ def add_loop(self, index_range: dcir.Range):
...
+
+            self.state is set to init_state and the other states are pushed on the stack to be
+            popped with `pop_condition_*()` methods.
+            """
+            # Data model validators enforce this to exist
+            assert isinstance(node.condition.stmts[0], dcir.AssignStmt)
+            assert isinstance(node.condition.stmts[0].left, dcir.ScalarAccess)
+            condition_name = node.condition.stmts[0].left.original_name
+
+            merge_state = self.sdfg.add_state("condition_after")
```
philip-paul-mueller, line 100: See note above about names of states in an SDFG and the note about the `add_state_before()`.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -75,9 +78,126 @@ def add_loop(self, index_range: dcir.Range):
...
+            # Evaluate node condition
+            init_state = self.sdfg.add_state("condition_init")
+            self.sdfg.add_edge(self.state, init_state, dace.InterstateEdge())
+
+            # Promote condition (from init_state) to symbol
+            condition_state = self.sdfg.add_state("condition_guard")
+            self.sdfg.add_edge(
+                init_state,
+                condition_state,
+                dace.InterstateEdge(assignments=dict(if_condition=condition_name)),
```
philip-paul-mueller, line 114: This means that `if_condition` is a global variable (a symbol) that is used everywhere as condition variable.
This is kind of okay, but I would recommend to use a different name for every condition.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -131,6 +251,91 @@ def _add_empty_edges(
...
+        sdfg_ctx: StencilComputationSDFGBuilder.SDFGContext,
+        node_ctx: StencilComputationSDFGBuilder.NodeContext,
+        symtable: ChainMap[eve.SymbolRef, dcir.Decl],
+        **kwargs: Any,
+    ) -> None:
+        sdfg_ctx.add_while(node)
+        assert sdfg_ctx.state.label.startswith("while_init")
+
+        read_acc_and_conn: dict[Optional[str], tuple[dace.nodes.Node, Optional[str]]] = {}
+        write_acc_and_conn: dict[Optional[str], tuple[dace.nodes.Node, Optional[str]]] = {}
```
philip-paul-mueller, line 267: As far as I can tell, the second element of the `tuple` will always be `None`, why do you need it.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -145,16 +350,108 @@ def visit_Tasklet(
...
             symtable=symtable,
+            sdfg=sdfg_ctx.sdfg,
         )
 
+        # We are breaking up vertical loops inside stencils in multiple Tasklets
+        # It might thus happen that we write a "local" scalar in one Tasklet and
+        # read it in another Tasklet (downstream).
+        # We thus create output connectors for all writes to scalar variables
+        # inside Tasklets. And input connectors for all scalar reads unless
+        # previously written in the same Tasklet. DaCe's simplify pipeline will get
```
philip-paul-mueller, line 361: Are you sure that this works.
For this to work DaCe would have to manipulate Tasklets, which it will not do.
So you will still have writes to transients on the SDFG level.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -145,16 +350,108 @@ def visit_Tasklet(
...
+            target_name = access_node.left.name
+
+            field_access = (
+                len(
+                    set(
+                        [
+                            memlet.connector
+                            for memlet in [*node.write_memlets]
+                            if memlet.connector == target_name
+                        ]
```
philip-paul-mueller, line 377:
```suggestion
                        memlet.connector
                        for memlet in [*node.write_memlets]
                        if memlet.connector == target_name
```

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -145,16 +350,108 @@ def visit_Tasklet(
...
+                    )
+                )
+                > 0
+            )
+            defined_symbol = False
+            for symbol_map in symtable.maps:
+                for symbol in symbol_map.keys():
+                    if symbol == read_name:
+                        defined_symbol = True
```
philip-paul-mueller, line 418:
```suggestion
            defined_symbol = any(read_name in symbol_map for symbol_map in symtable.maps)
```

philip-paul-mueller (commented):
There are some general comment or suggestions.
But I have some suggestions regarding DaCe, especially the usage of newer API.

edopao (commented):
My 2 cents review.

src/gt4py/cartesian/gtc/dace/daceir.py
```diff
@@ -842,33 +845,145 @@ class IterationNode(eve.Node):
     grid_subset: GridSubset
 
 
+class Condition(eve.Node):
+    condition: Tasklet
+    true_state: list[ComputationState | Condition | WhileLoop]
+
+    # Currently unused due to how conditions are parsed higher up in the pipeline
+    false_state: list[ComputationState | Condition | WhileLoop] = eve.field(default_factory=list)
```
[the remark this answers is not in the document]
romanc: Renamed as proposed.

As to where the `else` branch gets lost: That happens in [gtir_to_oir](https://github.com/GridTools/gt4py/blob/2f4db7236c7d7e5b2f2daec382966bd1605f088f/src/gt4py/cartesian/gtc/gtir_to_oir.py#L142-L184) and looks like something that should move to the numpy backend. Issue https://github.com/GridTools/gt4py/issues/1898 tracks refactors that we want to do in that region. I added an explicit comment to that issue in code.

src/gt4py/cartesian/gtc/dace/daceir.py
```diff
@@ -842,33 +845,145 @@ class IterationNode(eve.Node):
     grid_subset: GridSubset
 
 
+class Condition(eve.Node):
+    condition: Tasklet
+    true_state: list[ComputationState | Condition | WhileLoop]
```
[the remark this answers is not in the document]
philip-paul-mueller: As I said I do not care that much about it.
In GT4Py it seems that it is mixed, but mostly (with exceptions) consistent within a file.
However, you have to discuss this with Enrique.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -39,54 +40,96 @@
...
+
+
+def _field_access_iterator(
+    code_block: oir.CodeBlock | oir.MaskStmt | oir.While, access_type: AccessType
+):
+    if access_type == AccessType.WRITE:
+        return filter(
             lambda node: isinstance(node, oir.FieldAccess),
-            node.walk_values().if_isinstance(oir.AssignStmt).getattr("left"),
+            code_block.walk_values().if_isinstance(oir.AssignStmt).getattr("left"),
```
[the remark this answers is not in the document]
philip-paul-mueller: I am not an eve expert, but you should be able to write something like:
```python
code_block.walk_values().if_isinstance(oir.AssignStmt).getattr("left").filter(lambda node: isinstance(node, oir.FieldAccess))
```
According to my understanding it should be even possible to write something like:
```python
code_block.walk_values().if_isinstance(oir.AssignStmt).getattr("left").if_isinstance(oir.FieldAccess)
```

src/gt4py/cartesian/gtc/dace/daceir.py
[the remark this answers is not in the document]
romanc: Alright - I'll coordinate with Enrique. I guess we talk about in the context of https://github.com/GridTools/gt4py/issues/1829.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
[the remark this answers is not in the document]
romanc: Ah true, I can just chain another `if_instance()` at the end. Nice!

src/gt4py/cartesian/gtc/dace/utils.py
```diff
@@ -347,6 +347,165 @@ def compute_dcir_access_infos(
...
+def compute_tasklet_access_infos(
+    node: oir.CodeBlock | oir.MaskStmt | oir.While,
+    *,
+    collect_read: bool = True,
+    collect_write: bool = True,
+    declarations: dict[str, oir.Decl],
+    horizontal_extent,
+    k_interval,
+    grid_subset,
+) -> dace.properties.DictProperty:
```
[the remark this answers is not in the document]
romanc: Removed the wrong type hint and added a doc string (to the best of my knowledge - that part is still a bit fuzzy, seems to complicated to me and will probably resolved as part of the IR refactor)

src/gt4py/cartesian/gtc/dace/utils.py
```diff
@@ -347,6 +347,165 @@ def compute_dcir_access_infos(
...
+    def visit_While(self, node: oir.While, **kwargs):
+        self.visit(node.cond, is_write=False, **kwargs)
+        self.visit(node.body, **kwargs)
+
+    def visit_HorizontalRestriction(self, node: oir.HorizontalRestriction, **kwargs):
+        self.visit(node.mask, is_write=False, **kwargs)
+        self.visit(node.body, region=node.mask, **kwargs)
+
+    @staticmethod
+    def _global_grid_subset(
```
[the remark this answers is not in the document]
romanc: Good catch. It's just a regular class method now :+1:

src/gt4py/cartesian/gtc/dace/utils.py
```diff
@@ -347,6 +347,165 @@ def compute_dcir_access_infos(
     return ctx.access_infos
 
 
+class TaskletAccessInfoCollector(eve.NodeVisitor):
```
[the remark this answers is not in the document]
romanc: You are correct. They share a lot of code because one is basically a modified copy of the other. A task is logged in https://github.com/GridTools/gt4py/issues/1898 to re-evaluate that decision.

src/gt4py/cartesian/gtc/dace/utils.py
```diff
@@ -347,6 +347,165 @@ def compute_dcir_access_infos(
...
+        axes,
+        region: Optional[common.HorizontalMask],
+    ) -> dcir.FieldAccessInfo:
+        # Check we have expression offsets in K
+        offset = [offset_node.to_dict()[k] for k in "ijk"]
+        variable_offset_axes = [dcir.Axis.K] if isinstance(offset_node, VariableKOffset) else []
+
+        global_subset = self._global_grid_subset(region, self.he_grid, offset)
+        intervals = {}
+        for axis in axes:
```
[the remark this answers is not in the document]
romanc: Nice, I like it

src/gt4py/cartesian/gtc/dace/utils.py
```diff
@@ -347,6 +347,165 @@ def compute_dcir_access_infos(
...
+        self.grid_subset = grid_subset
+
+    def visit_CodeBlock(self, _node: oir.CodeBlock, **_kwargs):
+        raise RuntimeError("We shouldn't reach code blocks anymore")
+
+    def visit_AssignStmt(self, node: oir.AssignStmt, **kwargs):
+        self.visit(node.right, is_write=False, **kwargs)
+        self.visit(node.left, is_write=True, **kwargs)
+
+    def visit_MaskStmt(self, node: oir.MaskStmt, **kwargs):
```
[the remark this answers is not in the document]
romanc: Indeed, if we ended up here from line 501, we'd crash. We don't because of what we visit in case of `MaskStmt` /`While`. For `MaskStmt`, we visit `node.mask` and for `While`, we visit `node.cond`. Both are simple expressions (i.e. not assignments and not masks). Most of the time, we just evaluate a boolean scalar. Since we don't come by `visit_AssignStmt`, we never set `is_write` from there and thus set that manually to `False` (since we are just evaluating a condition). Does that make sense? I will add a comment in line 501 since it took me a moment to understand again what is going on.

src/gt4py/cartesian/gtc/dace/expansion/expansion.py
```diff
@@ -120,6 +120,14 @@ def _fix_context(
             if key in nsdfg.symbol_mapping:
                 del nsdfg.symbol_mapping[key]
 
+        for edge in parent_state.in_edges(node):
+            if edge.dst_conn not in nsdfg.in_connectors:
+                # Drop connection if connector is not found in the expansion of the library node
+                parent_state.remove_edge(edge)
+                if parent_state.in_degree(edge.src) + parent_state.out_degree(edge.src) == 0:
```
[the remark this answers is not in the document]
romanc: Indeed, this is something that DaCe could, arguable should do. I'm doing manual cleanup here since - in some cases - we connect more access nodes than necessary when building the "unexpanded sdfg"[^1], see "Refactor opportunity: Transient read after write". I tried to fix this as part of the current PR and as part of a cleanup PR. Both times, it would get messy fast and I thus decided to push this refactor down the line and accept that we have to do manual clean up here.

For your concern about the "quite useless AccessNode". After dropping unused connections, I remove any isolated AccessNode that might linger. This is - as you mentioned - a form of DeadDataflowElimination that DaCe would support in case we'd build a nice SDFG in the first place. Lots of duct-tape / tech dept here. I increased it and we are fine with that for now. I've added an explicit entry in https://github.com/GridTools/gt4py/issues/1898 such that we don't forget to clean this up later.

[^1]: https://geos-esm.github.io/SMT-Nebulae/technical/backend/dace-bridge/#building-the-coarse-grained-sdfg

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -318,167 +367,348 @@ def visit_FieldAccess(
...
-    ):
-        extent = global_ctx.library_node.get_extents(node)
-        decls = [self.visit(decl, **kwargs) for decl in node.declarations]
-        targets: Set[str] = set()
-        stmts = [
-            self.visit(
-                stmt,
+    ) -> dcir.Tasklet:
+        condition_expression = node.mask if isinstance(node, oir.MaskStmt) else node.cond
+        prefix = "if" if isinstance(node, oir.MaskStmt) else "while"
```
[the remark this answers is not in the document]
romanc: What I like about the ternary statements is that it assigns in one place. I know in python this doesn't matter since `if` statements don't open a scope, and still ... habit I guess ...  I'll pull out `is_mask_statement = isinstance(node, oir.MaskStmt)` as a compromise.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -318,167 +367,348 @@ def visit_FieldAccess(
...
+        tmp_name = f"{prefix}_expression_{id(node)}"
+
+        # Reset the set of targets (used for detecting read after write inside a tasklet)
```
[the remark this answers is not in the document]
romanc: `targets` is passed along as we build the `daceir` to detect read after write situations (as the comment mentions). Since this part isn't straight forward, let me elaborate a bit: Essentially, whenever we write to an access node, it's name gets registered in `targets` (see `FieldAccess`  / `ScalarAccess` visitor). Then, when we read (see the same visitors), we can check if we already wrote to that same variable inside the same Tasklet before (and if so, we read from `gtOUT__{varname}` instead of `gtIN__{varname}`).

Now consider the image at the end of [this paragraph](https://geos-esm.github.io/SMT-Nebulae/technical/backend/dace-bridge/#building-dace-ir-from-oir). If you scroll down, you'll see a mock-up version of how the states (and Tasklets) will look in the generated SDFG. Notice how statements 0, 1, and 8 are in the same `CodeBlock`, but not in the same Tasklet. Also notice how there's a Tasklet that says "eval if condition".

Say we write to `x` in `statement1`, and read it in `statement8`, we'll have it explicitly written to and read from an access node (of `x`). We thus need to clear the set of targets before. Actually, every Tasklet should have its own set of `targets`. This goes back to the dual-usage of the DaCe-IR and might be nicer once we get to clean this out.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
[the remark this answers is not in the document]
romanc: scratch that, if I pull out the `isinstance`, `mypy` gets confused, so I'd need to go with the traditional `if` as you suggested and - as mentioned - I like the ternaries better. So I won't change things here.

src/gt4py/cartesian/gtc/dace/utils.py
```diff
@@ -347,6 +346,165 @@ def compute_dcir_access_infos(
...
+        for axis in axes:
+            if axis in variable_offset_axes:
+                intervals[axis] = dcir.IndexWithExtent(
+                    axis=axis, value=axis.iteration_symbol(), extent=(0, 0)
+                )
+            else:
+                intervals[axis] = dcir.IndexWithExtent(
+                    axis=axis,
+                    value=axis.iteration_symbol(),
+                    extent=(offset[axis.to_idx()], offset[axis.to_idx()]),
```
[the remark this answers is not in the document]
romanc: No idea, this is code I copy/pasted from `AccessInfoCollector` and never questioned.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -39,54 +40,96 @@
...
+            #     statement2
+            #   statement3
+            #
+            # statements 1 and 3 will end up in the same CodeBlock but aren't in the same tasklet.
+            names = []
+            for statement in dcir_statements:
+                for access in statement.walk_values().if_isinstance(
+                    dcir.ScalarAccess, dcir.IndexAccess
+                ):
+                    names.append(access.name)
```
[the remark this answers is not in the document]
romanc: No, `dcir_statements` is a list of statements. This is a double `for` loop over all statements in `dcir_statements` and then all access nodes in every statement.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
romanc: Naming might not be perfect, but what you are describing is not what the code is doing.

We are looping over statements, grouping them in blocks. Note that we only group one block at a time, the "current block". I thus think this naming makes sense and I am not willing to change that.

src/gt4py/cartesian/gtc/dace/utils.py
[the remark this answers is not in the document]
philip-paul-mueller: This makes sense, I was thinking that it would be something like that but I wanted to be sure.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
romanc: No I don't. The contents of the "current block" (see above) is written to a `Tasklet`. That `Tasklet` is put into a state and then this `ComputationState` is added to `dace_nodes`. After adding the `Tasklet`, the `current_block` is reset (in the next statement).

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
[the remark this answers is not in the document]
philip-paul-mueller: Thanks a lot.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
romanc: From what I understand, this function adds `new_state` _after_ `self.state`, so if anything, I think we could use `add_state_after()`.

In any case, that would be unrelated to the refactor in this PR. We just add a `label` here.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
romanc: As said above, we just add a label here. We can't fix everything at the same time. We might be using `add_state_after()` in a future refactor.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
romanc: This is not true. DaCe will make sure that the state will have a unique name by appending a counter at the end of the string.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
romanc: See notes above: the naming is not a problem. Regarding `add_state_{before, after}()`, I'm a bit reluctant to use it here since I'm adding a state machine with parallel (true / false) branches so I like the flexibility of adding edges manually. This will all change once we move to control flow graphs.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
romanc: I can live with `if_condition` being a global variable. This will all change once we move to control flow graphs.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
romanc: Yes, I'm sure this works.

DaCe will go and change code inside the Tasklet (at least for python), see [DeadDataflowElimination (https://github.com/spcl/dace/blob/11d0e330993c8406e0f9ad4d1801e8da0a17abae/dace/transformation/passes/dead_dataflow_elimination.py#L127-L174) pass.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
romanc: I am not 100% sure of all the cases that are covered (or not anymore) by this structure. This structure was written before me and I'd rather not change it for no good reason, especially not on a PR as big as this one. Worst-case we have a couple `None`s flying around ...

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
```diff
@@ -75,9 +79,126 @@ def add_loop(self, index_range: dcir.Range):
...
+            self._pop_last("condition_after")
+
+        def add_while(self, node: dcir.WhileLoop) -> None:
+            """Inserts a while loop after the current state."""
+            # Data model validators enforce this to exist
+            assert isinstance(node.condition.stmts[0], dcir.AssignStmt)
+            assert isinstance(node.condition.stmts[0].left, dcir.ScalarAccess)
+            condition_name = node.condition.stmts[0].left.original_name
+
+            after_state = self.sdfg.add_state("while_after")
```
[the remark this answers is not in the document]
romanc: Thanks, Eduardo. That is valuable feedback. I left a note in https://github.com/GridTools/gt4py/issues/1898 such that we don't forget it in the future.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
```diff
@@ -318,167 +367,348 @@ def visit_FieldAccess(
...
+        global_ctx: DaCeIRBuilder.GlobalContext,
+        iteration_ctx: DaCeIRBuilder.IterationContext,
+        symbol_collector: DaCeIRBuilder.SymbolCollector,
+        horizontal_extent,
+        k_interval,
+        targets: list[oir.FieldAccess | oir.ScalarAccess],
+        inside_horizontal_region: bool = False,
+        **kwargs: Any,
+    ) -> dcir.MaskStmt | dcir.Condition:
+        if inside_horizontal_region:
```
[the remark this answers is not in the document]
romanc: > I've learned from my mistakes that lowering if-statements inside tasklets does not ensure exclusive branch execution: all inputs of a tasklet are evaluated, no matter the value of the if-condition inside the tasklet node. Randomly, this can result in correct code or not.

I don't quite follow. Let's assume we have the following stencil

```python
def small_conditional(in_field: FloatField, out_field: FloatField):
        if in_field < 4:
            tmp = in_field
        else:
            tmp = in_field + 1
```


Are you saying both branches, `if` and `else`, are evaluated?  That is not what we are looking at the generated code, this translates to real `if` statements in cpp code that work as I would expect them to. Can you elaborate on this?

Background: Before this PR, gt4py-cartesian would would have all conditionals inside Tasklets. I'm surprised to learn that this might pose a problem.

> I guess the gt4py user code mostly uses dynamic K-offsets, which limits the vertical domain of intermediate fields. On the horizontal region, fields are always defined on the full domain. Is this the motivation behind treating the vertical region differently?

@FlorianDeconinck / @twicki  can you make sense of that? I'm just reading words here without understanding ...

> If what I wrote so far makes sense, my question is the following. Is it theoretically possible to write cartesian programs that use horizontal dynamic offsets, and would also require `dcir.Condition` in the horizontal region?

What I can say is that it is not possible to have `dcir.Condition`s (or `dcir.WhileLoop`s) inside horizontal regions. There might be `dcir.MaskStmt`s (or `dcir.While` loops) inside horizontal regions. In that case, we won't translate them to `dcir.Condition`s  / `dcir.WhileLoop`s and codegen `if`/`while` statements inside the Tasklet (without exposing them to DaCe) just as we'd do before this PR.

romanc:
Thanks for all the feedback. I have addressed and/or answered all review comments. Please have a second look if you have strong opinions.

@FlorianDeconinck you wanted to do a re-read once the docs are done. The docs are [here](https://geos-esm.github.io/SMT-Nebulae/technical/backend/dace-bridge/).

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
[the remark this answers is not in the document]
romanc: Ah - now I understand. True, I should be able to pull this one out and not re-create `names` for every loop iteration :+1:

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
philip-paul-mueller: I completely understand that, and it again proves that comments are good.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
philip-paul-mueller: I did not new that, thanks a lot.

src/gt4py/cartesian/gtc/dace/utils.py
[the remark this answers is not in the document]
philip-paul-mueller: Okay, let's leave it at that, but it is something to investigate.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
philip-paul-mueller: Now I get it, I have to see a block as a collection of statements.
Yes you are right this makes perfect sense.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
philip-paul-mueller: Actually I did not know that.

src/gt4py/cartesian/gtc/dace/expansion/sdfg_builder.py
philip-paul-mueller: This is fine, I just wanted to make aware of it, because that functionality was added recently.
In fact I did that because I got so annoyed by that manual stuff all the time.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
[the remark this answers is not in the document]
romanc: Took me two tries, but it's now out of the loop. Thanks for catching this.

src/gt4py/cartesian/gtc/dace/expansion/daceir_builder.py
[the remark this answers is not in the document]
edopao: No, only one of the if-banchesis is executed. What I meant is the following, consider:
```
def small_conditional(in_field1: FloatField, in_field2: FloatField, k: IntField):
        if k > 10:
            tmp = in_field1
        else:
            tmp = in_field2
```

Both inputs to the tasklet (`in_field1` and `in_field2`) have to be defined for all `k` values `[0:N]`. That because the two dataflows that compute `in_field1` and `in_field2` are executed before evaluating the if-expression.

However, this case maybe is handled differently in cartesian.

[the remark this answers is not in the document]
FlorianDeconinck: K-domain validation will indeed take care of that.

src/gt4py/cartesian/gtc/dace/daceir.py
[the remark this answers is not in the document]
FlorianDeconinck: Log it

FlorianDeconinck (approved):
LGTM - one ticket to log

src/gt4py/cartesian/gtc/dace/daceir.py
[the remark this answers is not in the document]
romanc: Left a note (linking here) in https://github.com/GridTools/gt4py/issues/1898.

PR #1899, opened 2025-03-04
style[cartesian]: Cleanup codegen tests

twicki:
## Description

This PR cleans up all the warnings that would be present in the test_code_generation.

### DIFF
--- a/tests/cartesian_tests/integration_tests/multi_feature_tests/test_code_generation.py
+++ b/tests/cartesian_tests/integration_tests/multi_feature_tests/test_code_generation.py
@@ -34,6 +34,9 @@ from cartesian_tests.integration_tests.multi_feature_tests.stencil_definitions i
 )
 
 
+rng = np.random.default_rng(1337)
+
+
 @pytest.mark.parametrize("name", stencil_definitions)
 @pytest.mark.parametrize("backend", ALL_BACKENDS)
 def test_generation(name, backend):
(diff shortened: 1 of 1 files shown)


romanc (approved):
:broom: :broom: :broom: Thanks for the cleanup

Personally, I'm not a fan of

```diff
- field_a = field_b
+ field_a[0, 0, 0] = field_b
```

To me this feels like a step backwards, i.e. undoing the "convenience feature" that you can just write `field_a` for center reads/writes.

twicki:
> 🧹 🧹 🧹 Thanks for the cleanup
> 
> Personally, I'm not a fan of
> 
> ```diff
> - field_a = field_b
> + field_a[0, 0, 0] = field_b
> ```
> 
> To me this feels like a step backwards, i.e. undoing the "convenience feature" that you can just write `field_a` for center reads/writes.

The advantages that I see is:
1.  No checker is unhappy about unused args.
2. (and more importantly) Especially in physics we are getting to a point where stencils are becoming really large, eg 
image and we've seen people miss argument removal because they don't trust the unused arg checker anymore. In these small cases I agree that it should not matter, but I wanted to take this opportuntiy to highlight that this is an option to work around it if we so desire.
3. Especially when mixing lower and higher dimenional fields I see it as valuable to do this from time to time to clarify what is going on:
image
In here for example we update zfix, an IJ-only field though a forward computation while also working on 3d fields next to it

romanc:
cscs-ci run

PR #1901, opened 2025-03-04
feat[dace][next]: Added `CopyChainRemover`

philip-paul-mueller:
This PR adds the `CopyChainRemover` transformation.

The introduction of `concat_where` introduced a new pattern, which is essentially a chain of copies.
As an example imagine the case that a domain is split into 3 subdomain.
The result on the first subdomain is stored in `T1`, the one of the second in `T2` and the one for the third domain in `T3`.
`T1` and `T2` are then copied into `T4`, finally `T4` together with `T3` are then copied into `T5`.
This transformation will remove `T1`, `T2`, `T3` and `T4` thus the results will be written into `T5` directly.

There are some limitation, if we have the pattern `(A1) -> (A2)`, then we eliminate `A1` only if:
- `A1` is fully read; this is to avoid some nasty adjustments of map bounds.
- There can only be one connection between `A1` and `A2`.

The transformation was added twice to the simplify pass, which allows us to mitigate DaCe [issue#1959](https://github.com/spcl/dace/issues/1959).


src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -125,6 +125,8 @@ def gt_auto_optimize(
     """
     device = dace.DeviceType.GPU if gpu else dace.DeviceType.CPU
 
+    # TODO(phimuell): Should we propagate all the strides just to be sure?
```
edopao, line 128: This TODO comment is not very clear. Besides it is a question mark. I suggest to remove it.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -562,3 +595,450 @@ def apply(
...
+@dace_properties.make_properties
+class CopyChainRemover(dace_transformation.SingleStateTransformation):
+    """Removes chain of redundant copies, mostly related to `concat_where`.
+
+    `concat_where`, especially when nested, will build "chains" of AccessNodes, this
+    this transformation will remove them. It should be called repeatedly until a
+    fix point is reached and should be seen as an addition to the array removal passes
+    that ship with DaCe.
+    The transformation will look for the pattern: `(A1) -> (A2)`, i.e. a data container
+    is copied into another one. The transformation adders to ADR-18 and imposes the
```
edopao, line 609: I do not understand, what is the additional constraint?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -562,3 +595,450 @@ def apply(
...
+
+        # We remove `a1` so it must be a transient and used only once.
+        if not a1_desc.transient:
+            return False
+        if not self.is_single_use_data(sdfg, a1):
+            return False
+
+        # For simplicity we assume that neither of `a1` nor `a2` are views. There
+        #  are some cases were we might be able to remove some of them, but it would
+        #  be very hard to analyse. Furthermore, the lowering currently uses views
```
edopao, line 678: > Furthermore, the lowering currently uses views with special strides for broadcasting.

Not true anymore, can be removed.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -562,3 +595,450 @@ def apply(
...
+            current_subset = current_memlet.get_src_subset(current_edge, state)
+            new_src = a2
+            new_src_conn = None
+            new_dst = current_edge.dst
+            new_dst_conn = current_edge.dst_conn
+            assert current_edge.src_conn is None
+
+        # If the subset we care about, which is always on the `a1` side, was not
+        #  specified we assume that the whole `a1` has been written.
+        # TODO(edopao): Fix lowering that this does not happens, it happens for example
```
edopao, line 917: Does it happen in docstring tests?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -562,3 +595,450 @@ def apply(
...
+        self,
+        is_producer_edge: bool,
+        new_edge: dace_graph.MultiConnectorEdge,
+        a2_offsets: Sequence[dace_sym.SymExpr],
+        state: dace.SDFGState,
+        sdfg: dace.SDFG,
+        a1: dace_nodes.AccessNode,
+        a2: dace_nodes.AccessNode,
+    ) -> None:
+        """Modify modify the data flow associated to `new_edge`.
```
edopao, line 962:
```suggestion
        """Modify the data flow associated to `new_edge`.
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -562,3 +595,450 @@ def apply(
...
+
+        The `_reroute_edge()` function creates a new edge, but it does not modify
+        the data flow at the other side, of the connection, this is done by this
+        function.
+
+        Depending on the value of `is_producer_edge` the function will either modify
+        the source of `new_edge` (`True`) or it will modify the data flow associated
+        to the destination pf `new_edge` (`False`).
```
edopao, line 970:
```suggestion
        to the destination of `new_edge` (`False`).
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -562,3 +595,450 @@ def apply(
...
+        a2: dace_nodes.AccessNode = self.node_a2
+        a1_to_a2_edge: dace_graph.MultiConnectorEdge = next(
+            oedge for oedge in graph.out_edges(a1) if oedge.dst is a2
+        )
+        a1_to_a2_memlet: dace.Memlet = a1_to_a2_edge.data
+        a1_to_a2_dst_subset: dace_sbs.Range = a1_to_a2_memlet.get_dst_subset(a1_to_a2_edge, graph)
+
+        # Note that it is possible that `a1` is connected to the same node multiple
+        #  times, although through different edges. We have to modify the data
+        #  flow there, since the offsets and that dat has changed. However, we must
```
edopao, line 770:
```suggestion
        #  flow there, since the offsets and the data have changed. However, we must
```

edopao (approved):
Great job! I imagined it would have been much simpler..

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
philip-paul-mueller: Agreed, I guess it ended there during some debugging.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: I agree it was written very bad, and in fact it was wrong.
I have updated the description.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: You are right.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: I have not really looked into it, I just know that in that file and that test there is a Memlet without `other_subset`.
It is not a high priority, but validation complains, and this case only has three access nodes, so it should be simple to fix, once we have time.

edopao: I've found the place, that's because I call `sdfg.make_array_memlet(field.dc_node.data)` which only sets `subset`, not `other_subset`.

edopao (approved):

PR #1905, opened 2025-03-06
refactor[next]: Partial cleanup OffsetLiteralType

SF-N:


src/gt4py/next/iterator/transforms/inline_lifts.py
```diff
@@ -154,6 +154,7 @@ def visit_FunCall(
             ir.FunCall(
                 fun=self.generic_visit(node.fun, is_scan_pass_context=_is_scan(node), **kwargs),
                 args=self.generic_visit(node.args, **kwargs),
+                type=node.type,
```
SF-N, line 157: This fix is unrelated to the PR itself but was coincidentally triggered here. Without the change, we lose all type information when doing an InlineLift.

tehrengruber (approved):

Issue #1880, opened 2025-02-20
refactor: storage, layout, and devices (mostly cartesian)


romanc:
## Description

PRs https://github.com/GridTools/gt4py/pull/1843 and https://github.com/GridTools/gt4py/pull/1867 raised a couple of questions regarding storage, layouts, and device support/handling on the cartesian side. This is a collector issue in which I'd like to group a couple small PRs and maybe have discussions around these refactors.

So far, we've identified a potential for improvement in the following areas:

- [x] limit `DeviceType` to what is actually used/supported.
  PR https://github.com/GridTools/gt4py/pull/1879
- [x] share a centralized `CUPY_DEVICE` (currently defined separately in `next` and `cartesian`) in `_core`
  PR https://github.com/GridTools/gt4py/pull/1908
- [x] break the dependency cycle between `storage` and `cartesian`, see [this comment](https://github.com/GridTools/gt4py/pull/1867#issuecomment-2662606213)
  PR https://github.com/GridTools/gt4py/pull/1916
- [x] get rid of `GT4PY_USE_HIP` environment variable since auto-detection of NVIDIA/AMD GPUs seems to work fine, see [this comment](https://github.com/GridTools/gt4py/pull/1867#issuecomment-2662742715)
  PR https://github.com/GridTools/gt4py/pull/1916
- [ ] [cartesian] Move `Layout` into `Backend`, see https://github.com/GridTools/gt4py/pull/1867#discussion_r1959892278

Consider this as a low-priority wish list for refactors on the side.

PR #1908, opened 2025-03-11
refactor: centralize `CUPY_DEVICE_TYPE` in `_core/definitions`

romanc:
## Description

This PR centralizes the definition of `CUPY_DEVICE_TYPE` in `_core/definitions`. This effectively de-duplicates the definition of `CUPY_DEVICE` in gt4py next and cartesian. Fixes a couple of typos along the way.

Related issue: https://github.com/GridTools/gt4py/issues/1880

## Requirements

  Should be covered by existing tests.
  N/A

### DIFF
--- a/src/gt4py/_core/definitions.py
+++ b/src/gt4py/_core/definitions.py
@@ -51,6 +51,9 @@ if TYPE_CHECKING:
 
     JaxNDArray: TypeAlias = jnp.ndarray
 
+# The actual assignment happens after the definition of `DeviceType` enum.
+CUPY_DEVICE_TYPE: Literal[None, DeviceType.CUDA, DeviceType.ROCM]
+"""Type of the GPU accelerator device, if present."""
 
 # -- Scalar types supported by GT4Py --
 bool_ = np.bool_
@@ -399,7 +402,7 @@ DeviceTypeT = TypeVar(
 )
 
 
-CUPY_DEVICE: Final[Literal[None, DeviceType.CUDA, DeviceType.ROCM]] = (
+CUPY_DEVICE_TYPE = (
     None if not cp else (DeviceType.ROCM if cp.cuda.runtime.is_hip else DeviceType.CUDA)
 )
 
--- a/src/gt4py/_core/definitions.py
+++ b/src/gt4py/_core/definitions.py
@@ -39,9 +39,12 @@ from gt4py.eve.extended_typing import (
 )
 
 
-if TYPE_CHECKING:
+try:
     import cupy as cp
+except ImportError:
+    cp = None
 
+if TYPE_CHECKING:
     CuPyNDArray: TypeAlias = cp.ndarray
 
     import jax.numpy as jnp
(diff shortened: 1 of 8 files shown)


romanc:
Friendly ping for review. Let me know if I should ping someone else instead.

src/gt4py/_core/definitions.py
```diff
@@ -396,6 +399,11 @@ class DeviceType(enum.IntEnum):
 )
 
 
+CUPY_DEVICE: Final[Literal[None, DeviceType.CUDA, DeviceType.ROCM]] = (
```
egparedes, line 402: I would also rename it to `CUPY_DEVICE_TYPE`. The place in the file is weird but I understand that it can't be defined until `DeviceType` is defined. Maybe would help to declare the type at the top and here only assign the value. For example, at `_definitions.py` line 54:

```python
#: Type of the GPU accelerator device, if present. The actual assignment
#: happens after the definition of `DeviceType` enum.
CUPY_DEVICE_TYPE: Final[Literal[None, DeviceType.CUDA, DeviceType.ROCM]]
```

and then here:
```python
CUPY_DEVICE_TYPE = None if not cp else (DeviceType.ROCM if cp.cuda.runtime.is_hip else DeviceType.CUDA)
```

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace.py
```diff
@@ -21,13 +20,12 @@
 from gt4py.next.ffront.fbuiltins import where
 
 from next_tests.integration_tests import cases
-from next_tests.integration_tests.cases import E2V, cartesian_case, unstructured_case
+from next_tests.integration_tests.cases import E2V, cartesian_case, unstructured_case  # noqa: F401
```
egparedes, line 23: Just a question: why are those new noqa ignores needed?

egparedes (commented):
Looks good. Just 2 suggestions for readability.

src/gt4py/_core/definitions.py
romanc: I like that. Let me change it accordingly.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace.py
romanc: Since we don't run `ruff` on `tests/` as mentioned in https://github.com/GridTools/gt4py/issues/1829 and configured in

https://github.com/GridTools/gt4py/blob/42850f0422a516c937100744810d338d7b99d0b8/.pre-commit-config.yaml#L53

they aren't strictly necessary. The language server sees them as unused, so I tried to remove them and test started failing. I found other places where explicit annotations were added to similar imports, e.g.

https://github.com/GridTools/gt4py/blob/42850f0422a516c937100744810d338d7b99d0b8/tests/next_tests/integration_tests/feature_tests/dace/test_orchestration.py#L15-L25

I'm happy to remove them if you feel like we should only do this once we (re-)activate ruff for the `tests/` folder

src/gt4py/_core/definitions.py
romanc: The only thing to consider with this approach is that `CUPY_DEVICE_TYPE` can't be `Final` anymore because it's not defined and declared in the same line anymore. For me personally, that's a compromise I'm willing to accept. No strong opinion though, feel free to argue the other way.

egparedes (approved):
LGTM

PR #1909, opened 2025-03-11
fix[next][dace]: assert on subset_to_adjust in CopyChainRemover

edopao:
The `subset_to_adjust` was `None` in icon4py CI, see for example:
https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504671/-/jobs/9378871351

This code change seems to retrieve it correctly.


edopao:
Included on #1910

PR #1911, opened 2025-03-12
fix[next][dace]: Mapping of dace symbol to nested SDFG

edopao:
It was found that certain IR that mixes `where` and `concat_where` expressions requires access to scalar values in domain expressions in a parent expression, but the child expression might access the same values outside of a domain expression. The lowering represents scalars used by domain expressions as dace symbols in the SDFG. When creating a nested SDFG, we need to keep the same representation (dace symbols) in order to make the correct mapping of values and avoid symbol-to-scalar conversion.


philip-paul-mueller (approved):
LGTM

edopao:
cscs-ci run default

PR #1912, opened 2025-03-13
fix[next][dace]: Fix array subset on the branches of if-expressions

edopao:
The issue was discovered by Philip (@philip-paul-mueller).

The lowering of if-expressions has to ensure that all branches (e.g. if/then states) write the same subset to the result array. This was not the case in baseline when one the branches contained a simple expression such as `{dwdx, dwdy}`. The bug was that the full shape of the global fields was copied into the result array, although the result array was allocated on a sub-domain. This caused an out-of-bound write and a segfault.


src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -618,57 +718,67 @@ def translate_if(
...
-            raise ValueError(
-                f"Different type of result fields on if-branches '{true_br.gt_type}' vs '{false_br.gt_type}'."
+    if isinstance(node.type, ts.TupleType):
+        symbol_tree = gtir_sdfg_utils.make_symbol_tree("x", node.type)
+        if isinstance(node.annex.domain, tuple):
+            domain_tree = node.annex.domain
+        else:
+            # TODO(edopao): this is a workaround for some IR nodes where the inferred
+            #   domain on a tuple of fields is not a tuple, see `test_execution.py::test_ternary_operator_tuple()`
+            domain_tree = gtx_utils.tree_map(lambda _: node.annex.domain)(symbol_tree)
```
edopao, line 739: @tehrengruber I am using a workaround because domain inference does not create a tuple, as expected, in `test_execution.py::test_ternary_operator_tuple()`

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -654,14 +662,15 @@ def test_gtir_cond():
...
-    sdfg = build_dace_sdfg(testee, CARTESIAN_OFFSETS)
+    sdfg = infer_domain_and_build_dace_sdfg(testee, CARTESIAN_OFFSETS)
 
     for s1, s2 in [(1, 2), (2, 1)]:
         d = np.empty_like(a)
         sdfg(a, b, c, d, s1, s2, scalar=1.0, **FSYMBOLS)
         assert np.allclose(d, (a + b + 1) if s1 > s2 else (a + c + 1))
 
 
+@pytest.mark.xfail(reason="error in domain inference")
```
edopao, line 673: @tehrengruber Domain inference fails in this test IR

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -1471,6 +1480,7 @@ def test_gtir_reduce_dot_product():
     assert np.allclose(v, v_ref)
 
 
+@pytest.mark.xfail(reason="error in domain inference")
```
edopao, line 1483: @tehrengruber ..and on this one too.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
philip-paul-mueller: In my opinion it would be good if there is a concrete issue.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -103,6 +103,14 @@ def build_dace_sdfg(
     )
 
 
+def infer_domain_and_build_dace_sdfg(
```
philip-paul-mueller, line 106: I am not an expert, but why is this function only used in some places?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -503,8 +501,11 @@ def make_temps(
             else:
                 temp, _ = self.add_temp_array_like(sdfg, desc)
```
philip-paul-mueller, line 502:
```suggestion
                temp, temp_desc = self.add_temp_array_like(sdfg, desc)
```

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -503,8 +501,11 @@ def make_temps(
...
                 temp_node = head_state.add_access(temp)
+                field_range = dace.subsets.Range.from_array(desc)
```
philip-paul-mueller, line 504:
```suggestion
                field_range = dace.subsets.Range.from_array(desc)
                temp_range = dace.subsets.Range.from_array(temp_desc)
```

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -503,8 +501,11 @@ def make_temps(
...
                 head_state.add_nedge(
-                    field.dc_node, temp_node, sdfg.make_array_memlet(field.dc_node.data)
+                    field.dc_node,
+                    temp_node,
+                    dace.Memlet(data=temp, subset=field_range, other_subset=field_range),
```
philip-paul-mueller, line 508:
```suggestion
                    dace.Memlet(data=temp, subset=temp_range, other_subset=field_range),
```

- There was some kind of sharing between the subsets, if you modify one it would also change the other, which is dangerous (it is essentially like calling `copy()` on a `list` of NumPy objects and then modify the objects in the copy).
- Furthermore this Memlet is highly non trivial as `data` refers to the destination, so giving better names to the argument is better.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -271,8 +271,6 @@ def make_field(
         Returns:
             The descriptor associated with the SDFG data storage, filled with field origin.
         """
-        if isinstance(data_type, ts.ScalarType):
-            return gtir_builtin_translators.FieldopData(data_node, data_type, origin=())
```
philip-paul-mueller, line 275: I know you plan to allow zero dim fields, but I do not fully understand why you could remove this check, as there are no other changes in the function?

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -576,6 +576,106 @@ def translate_as_fieldop(
...
+
+
+def _write_if_branch_output(
+    sdfg: dace.SDFG,
+    state: dace.SDFGState,
+    src: FieldopData,
+    dst: FieldopData,
+) -> None:
+    """
+    Helper function called by `translate_if()` to write the result on an if branch,
```
philip-paul-mueller, line 634:
```suggestion
    Helper function called by `translate_if()` to write the result of an `if` branch,
```

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -576,6 +576,106 @@ def translate_as_fieldop(
...
+    here `src` field, to the output 'dst' field. The write subset is based on the
```
philip-paul-mueller, line 635:
```suggestion
    here `src` field, to the output `dst` field. The write subset is based on the
```

I am not sure what the last sentences mean, does it mean that the whole `dst` field is written?

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -576,6 +576,106 @@ def translate_as_fieldop(
...
+
+    if isinstance(src.gt_type, ts.ScalarType):
+        state.add_nedge(
+            src.dc_node,
+            dst_node,
+            dace.Memlet(data=src.dc_node.data, subset="0"),
+        )
+    else:
+        if isinstance(src.gt_type.dtype, ts.ListType):
+            src_origin = [*src.origin, 0]
```
philip-paul-mueller, line 653: Why are you appending a zero here?

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -576,6 +576,106 @@ def translate_as_fieldop(
...
+            dst_origin = [*dst.origin, 0]
+        else:
+            src_origin = [*src.origin]
+            dst_origin = [*dst.origin]
+
+        data_subset = dace_subsets.Range(
+            (
+                f"{dst_start - src_start}",
+                f"{dst_start + size - src_start - 1}",
```
philip-paul-mueller, line 662:
```suggestion
                f"{dst_start - src_start + size - 1}",
```
This makes it more uniform and with the start.
I would also add a comment about why there is a `-1`.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -618,57 +718,67 @@ def translate_if(
     sdfg.add_edge(cond_state, false_state, dace.InterstateEdge(condition=(f"not ({if_stmt})")))
     sdfg.add_edge(false_state, state, dace.InterstateEdge())
 
-    true_br_args = sdfg_builder.visit(
+    true_br_result = sdfg_builder.visit(
```
philip-paul-mueller, line 721: This is just my curiosity, but why the rename (although that I have to say that I like `result` better).

philip-paul-mueller (commented):
There are some things that should be changed (especially the sharing of some variables used as subset) but there is nothing big.
However, we should wait for the feedback of Till.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
edopao: It does not really fail, I need an API to correctly retrieve the annex domain in case of tuples.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
edopao: As agreed, I am going to use an API to correctly retrieve the annex domain when it is available.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
edopao: It was only used in the places where it was strictly needed, that is the IR programs that used if-nodes. However, it is not wrong to call it always. I am going to do it.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: Good idea!

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
edopao: I like `result` better too, no special reason.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
edopao: That's for the local dimension, and it represents the start index of a list (always zero).

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
edopao: Yes, that's what happens. I can write it in the comment.

philip-paul-mueller (approved):
From my side it looks okay, but I would wait for Till's feedback.

edopao:
> From my side it looks okay, but I would wait for Till's feedback.

Thanks for the review. I have already discussed the domain inference problems with Till, there will be some utility functions to get the tuple domain (and the domain range in general, which might be useful in multiple places of the SDFG lowering).

PR #1913, opened 2025-03-17
feat[dace][next]: Adjusted the handling of non standard GPU Memlets.

philip-paul-mueller:
Due to the update to DaCe in [PR#1976](https://github.com/spcl/dace/pull/1976) the code generator is now able to handle more Memlets directly as Cuda `memcpy()` calls. Thus we have to change this function to reflect this case.
It even stands to reason that these functions can be removed altogether.

Note that the merge has become possible because of our switch to our DaCe fork in [PR#2012](https://github.com/GridTools/gt4py/pull/2012).


src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -269,16 +271,9 @@ def _gt_expand_non_standard_memlets_sdfg(
...
-                        is_dst_cont = False
-                    if is_src_cont and is_dst_cont:
-                        continue
-                else:
+                is_fortran_order = src_strides[0] == 1 and dst_strides[0] == 1
+                is_c_order = src_strides[-1] == 1 and dst_strides[-1] == 1
+                if is_c_order or is_fortran_order:
                     continue
             elif dims > 2:
                 if not (src_strides[-1] != 1 or dst_strides[-1] != 1):
```
edopao, line 279: I was wondering why we do not handle the case `dims > 2`. Is it because the generic `cudaMemcpy()` does not provide parameters `wOffset` and `hOffset`?

edopao: By handle, I mean why we do not consider the Fortran order.

philip-paul-mueller: This has to do with the code generator that can only C order at the moment for more than 3 dimensions.
This is something I want to fix in DaCe but was not pressing enough yet.

edopao (approved):
The GPU tests pass, although we are on a dace baseline that does not include the PR mentioned in the PR description. I suspect this is because gt4py (and icon4py as well) use symbolic field strides, by default. If this is the case, is it safe to merge this PR?

edopao:
This PR can merged, since the dace source now points to the GridTools fork repo.

PR #1915, opened 2025-03-17
feat[dace][next]: Updated Array Removal Transformations

philip-paul-mueller:
This PR updates how the redundant array removal transformations works.

First of all it updates the "self copy" transformation, i.e. where information is copied from global memory into global memory, either directly or indirectly.
Because of ADR-18 we know that we can simply remove such connections without doing an elaborated analysis.
There are two different transformations, that handles the case of a self copy in the same state (there is also one that handles them across multiple states):
- `SingleStateGlobalSelfCopyElimination`: This transformation was there before. It essentially handles the pattern `(G) -> (T) -> (G)`, i.e. the write is not direct but involves an intermediate. The transformation was updated to handle more cases.
- `SingleStateGlobalDirectSelfCopyElimination`: This transformation was newly added, it essentially handles the following pattern `(G) -> (G)`, i.e. there is no temporary data involved.

In the long run these two transformation should be combined.

In addition the `CopyChainRemover` was updated.
Before it was required that the whole array, that should be removed, henceforth called `A1`, was read, however, in some cases this could not be checked symbolically.
To handle these cases, a new heuristic was implemented, which assumes a certain behavior of the domain interference, i.e. that all arrays are as small as possible, i.e. nothing that is not needed is computed.


src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -604,6 +639,112 @@ def apply(
...
+        # NOTE: We do not check if `G` is read downstream.
+        if read_g.data != write_g.data:
+            return False
+        if g_desc.transient:
+            return False
+        if graph.scope_dict()[read_g] is not None:
+            return False
+
+        # Ensure that no cycles are created by the removal
+        for oedge in graph.out_edges(read_g):
```
edopao, line 696: Without thinking too much, my impression is that it should always be possible to remove `(G) -> (G)`. If there are cycles or strange implicit shifts, these problems were already an issue of the original SDFG. We can discuss it.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -604,6 +639,112 @@ def apply(
...
+        #  then also remove it.
+        graph.remove_node(read_g)
+
+        if graph.degree(write_g) == 0:
+            graph.remove_node(write_g)
+
+
 @dace_properties.make_properties
 class CopyChainRemover(dace_transformation.SingleStateTransformation):
     """Removes chain of redundant copies, mostly related to `concat_where`.
```
edopao, line 750: Is it correct to say that `(G) -> (G)` could be a bi-product of `CopyChainRemover`? in this case, we could add a not to recommend to run the `*SelfCopyElimination` passes after `CopyChainRemover`.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: An interesting idea.
If I understand it correctly what you say is, that it is nit needed to remove the nodes, you just have to remove the connection?
I will think about it and I think it can work.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: I am actually not sure where they are coming from, it is likely that they are indeed from the `CopyChainRemover`.
We can add this note, but those transformation are run in simplify and simplify is run several times, so we will still need it.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
edopao: No, that was not my idea.. we can discuss it.

philip-paul-mueller:
@edopao Can we merge this PR?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -477,14 +477,23 @@ class SingleStateGlobalSelfCopyElimination(dace_transformation.SingleStateTransf
...
+    second `G` node will redirected such that they use the first `G` node.
+    Furthermore, if `T` is not used downstream then also the `T` node will be
+    removed as well.
+
+    This transformation assumes that the SDFG follows rule 3 of ADR-18, which
+    guarantees that there is only a point wise dependency of the output on the
+    input.
+
+    Todo:
+        If `T` is only used in this state, then also redirect reads from `T` such
```
edopao, line 495: Is this not implemented by the `_is_read_downstream()` check?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -582,20 +595,42 @@ def apply(
...
-            graph.remove_node(read_g)
+        # If the `T` is no longer needed then remove it. If the first node has become
+        #  isolated, also remove it.
+        if not tmp_node_is_still_needed:
             graph.remove_node(tmp_node)
+
+            if graph.degree(read_g) == 0:
+                graph.remove_node(read_g)
+
             # It could still be used in a parallel branch.
```
edopao, line 634: This sounds like an undefined behavior, we should discuss it.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -604,6 +639,154 @@ def apply(
...
+    This transformation is extremely similar to `SingleStateGlobalSelfCopyElimination`,
+    however, this transformation does not need a buffer between the two global
+    AccessNodes. Therefore it matches the pattern `(G) -> (G)`, where `G` refers to
+    global data.
+    The transformation has two modes in how this is achieved, which one is chosen,
+    depends on if cycles are created or not. The first mode, which is selected if
+    cycles would be created, is to merge the two AccessNodes together. For this
+    the edges of the first AccessNode are reconnected to the second AccessNode.
+
+    In the second mode, which is selected if cycles would be created, is to simply
```
edopao, line 655: I do not understand in which case cycles would appear. We can discuss it.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -604,6 +639,154 @@ def apply(
...
+        read_g = self.node_read_g
+        write_g = self.node_write_g
+        g_desc = read_g.desc(sdfg)
+
+        # NOTE: We do not check if `G` is read downstream.
+        if read_g.data != write_g.data:
+            return False
+        if g_desc.transient:
+            return False
+        if graph.scope_dict()[read_g] is not None:
```
edopao, line 692: What does this check mean?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -731,10 +919,25 @@ def can_be_applied(
...
         #  that we do not have to adjust maps.
         a1_range = dace_sbs.Range.from_array(a1_desc)
-        if not src_subset.covers(a1_range):
+        if src_subset.covers(a1_range):
+            pass
+        elif all(ss.is_constant() for ss in src_subset.size()):
+            # If the subset is fully known then we require that it is covered.
+            return False
+        elif (
+            (not a2_desc.transient)
```
edopao, line 935: Do we really need to check `(not a2_desc.transient)`? we should not have any implicit shift, right?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
edopao: If the pattern (G) -> (T) is found somewhere else in this state, then remove it.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
edopao: We can rewrite "if-branch"

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
edopao: Understood in our discussion, thank you.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
edopao: Understood in our discussion, maybe add a comment note.

edopao (approved):

PR #1920, opened 2025-03-19
fix[next][dace]: Make default make_persistent=False

edopao:
GT4Py programs use fields with symbolic shape and strides. This allows to invoke the same stencil multiple times with different argument fields, without needing to recompile it (no JIT). Therefore, the temporary fields used inside the program cannot be made persistent. That means that the lifescope of transient arrays is limited to the SDFG scope.


philip-paul-mueller (approved):
LGTM

PR #1925, opened 2025-03-21
fix[next]: get scalar value from 0d array in `_hyperslice` 

halungge:
This PR fixes an issue when running embedded on CUPY arrays using the connectivity tables on CUPY arrays: when using the slice later on it fails because it contains 0d arrays and not scalars.


halungge:
cscs-ci default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

PR #1931, opened 2025-03-26
fix[next]: upgrade ProgramWithBoundArgs to GTIR pipelines

havogt:
TODO:
- [ ] add test

### DIFF
--- a/src/gt4py/next/ffront/decorator.py
+++ b/src/gt4py/next/ffront/decorator.py
@@ -397,24 +397,10 @@ class ProgramWithBoundArgs(Program):
         return super().__call__(*tuple(full_args), offset_provider=offset_provider, **full_kwargs)
 
     @functools.cached_property
-    def itir(self):
-        new_itir = super().itir
-        for new_clos in new_itir.closures:
-            new_args = [ref(inp.id) for inp in new_clos.inputs]
-            for key, value in self.bound_args.items():
-                index = next(
-                    index
-                    for index, closure_input in enumerate(new_clos.inputs)
-                    if closure_input.id == key
-                )
-                new_args[new_args.index(new_clos.inputs[index])] = promote_to_const_iterator(
-                    literal_from_value(value)
-                )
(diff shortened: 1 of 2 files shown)


src/gt4py/next/iterator/transforms/bind_args.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any
+
+from gt4py.eve import NodeTranslator, PreserveLocationVisitor
+from gt4py.next.iterator import ir
+from gt4py.next.iterator.ir_utils import ir_makers as im
+
+
+class BindArgs(PreserveLocationVisitor, NodeTranslator):
```
tehrengruber, line 8: Another possibility would be to use the `RemapSymbolRefs`. It only needs a slight extension such that it is happy with `Program`s. That way the args could appear everywhere in the IR, e.g. in a domain bound.

src/gt4py/next/ffront/decorator.py
```diff
@@ -397,24 +397,10 @@ def __call__(self, *args, offset_provider: common.OffsetProvider, **kwargs):
...
-                )
-                new_args[new_args.index(new_clos.inputs[index])] = promote_to_const_iterator(
-                    literal_from_value(value)
-                )
-                new_clos.inputs.pop(index)
-            params = [sym(inp.id) for inp in new_clos.inputs]
-            expr = itir.FunCall(fun=new_clos.stencil, args=new_args)
-            new_clos.stencil = itir.Lambda(params=params, expr=expr)
-        return new_itir
+    def gtir(self):
```
tehrengruber, line 400: Are you sure this works? This function is not called anywhere. As far as I understood the strategy in the toolchain refactorings was to preserve all public API. The result was a lot of boilerplate code that is either never called or called from strange places where the code path is very hard to follow. I think `Program.gtir` is one of the former examples.

havogt:
Indeed not solving the issue, as the itir/gtir functions are not used in the pipeline...

PR #1938, opened 2025-03-28
fix-CI: upgrade cmake in base image

edopao:
This addresses an issue observed in CI: `Compatibility with CMake < 3.5 has been removed from CMake.`

The `cmake` version in ubuntu base image did not satisfy the minimum requirement `>= 3.5`. By adding cmake to the list of packages to be installed, the latest package version `3.22.1` is installed, which satisfies the requirement.


edopao:
cscs-ci run

edopao:
cscs-ci run default

edopao:
This PR doesn't solve the problem.

PR #1939, opened 2025-03-28
feat[next][dace]: Dead Dataflow Elimination

philip-paul-mueller:
Added the `DeadMemletElimination` and `DeadMapElimination` transformation.
Thy process the dataflow graph and remove the Memlets and Maps that are no ops, i.e. transfer zero data.
This is not handled by the DaCe implementation at the moment.


src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
```diff
@@ -110,6 +112,22 @@ def gt_simplify(
                 result = result or {}
                 result.update(inline_res)
 
+        # TODO(phimuell): Is this the correct location for this, should it be even
```
edopao, line 115: I do not have any objection.. I think you can remove the TODO comment unless there is a concrete alternative.

edopao (approved):

src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
edopao: I have had some second thoughts about this. The lowering to SDFG applies some similar pruning, for the trivial empty memlets already present in the IR. So the only case when new trivial empty memlets can occur is after we substitute constant symbols. If we keep the pruning in the lowering, maybe the right place to run this pass is in the `gt_substitute_compiletime_symbols()` method. We can discuss it.

philip-paul-mueller: I kind of agree.
However, this transformation does not only remove dead dataflow, in doing so it also enables further transformation to apply.
But to be effective the transformations also implicitly require that other transformation have run, mostly inline SDFG.

After some consideration I also think that the transformation should run after state fusion, because it can not eliminate dead data flow across states.
Thus I would say that we should move the call after DaCe's simplify.

What I could do is write a more efficient scanning implementation.

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -158,101 +161,102 @@ def apply(
                     sdfg.remove_data(removed_data, validate=False)
 
 
-@dace_properties.make_properties
-class DeadMemletElimination(dace_transformation.SingleStateTransformation):
+def gt_dead_memlet_elimination(
+    sdfg: dace.SDFG,
+    single_use_data: Optional[dict[dace.SDFG, set[str]]] = None,
+) -> int:
     """Removes Melets that does not carry any dataflow.
```
edopao, line 168:
```suggestion
    """Removes Melets that do not carry any dataflow.
```

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -158,101 +161,102 @@ def apply(
...
 
-    The transformation matches an AccessNode and then processes all of its outgoing
-    edges. An edge is removed if:
-    - The Memlet carries no data.
-    - The Memlet is connected to another AccessNode.
+    The function will recursively examine the outgoing Memlets of all AccessNodes.
+    If it finds a Memlet that does not carrying any data it will be removed. However
```
edopao, line 171:
```suggestion
    If it finds a Memlet that does not carry any data, the Memlet will be removed. However
```

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -158,101 +161,102 @@ def apply(
...
+    the function will only consider connection between two AccessNodes.
```
edopao, line 172:
```suggestion
    the function will only consider connections between two AccessNodes.
```

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -158,101 +161,102 @@ def apply(
...
+    If the removing of the Memlets will lead to isolated nodes, these will be removed.
```
edopao, line 173:
```suggestion
    If the removing of the Memlet leads to isolated nodes, these will be removed.
```

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -158,101 +161,102 @@ def apply(
...
+    Furthermore, if `single_use_data` was passed the function will also removed
```
edopao, line 174:
```suggestion
    Furthermore, if `single_use_data` was passed the function will also remove
```

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -158,101 +161,102 @@ def apply(
...
-    ) -> None:
-        self._single_use_data = None
-        if single_use_data is not None:
-            self._single_use_data = single_use_data
-        super().__init__(*args, **kwargs)
+def _gt_dead_memlet_elimination_sdfg(
+    sdfg: dace.SDFG,
+    single_use_data: Optional[dict[dace.SDFG, set[str]]] = None,
+) -> int:
+    """Removes Melets that does not carry any dataflow.
```
edopao, line 195:
```suggestion
    """Removes Melets that do not carry any dataflow.
```

src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
```diff
@@ -154,6 +138,23 @@ def gt_simplify(
                     result["CopyChainRemover"] = 0
                 result["CopyChainRemover"] += copy_chain_remover_result
 
+        # TODO(phimuell): Is this the correct location? I think it is best calling
+        #   it after we have created dataflow as large as possible, but should we run
```
edopao, line 142: From what I remember, `CopyChainRemover` does not modify the memlet subsets. I mean, it does not merge different memlet terminated on the same access node. So it should be correct to call `DeadDataflowElimination` before `CopyChainRemover`. The advantage would be that `CopyChainRemover` has potentially less edges to analyze.

PR #1941, opened 2025-03-31
feat[next][dace]: setup cached and non-cached DaCe backends

edopao:
The `run_dace_cpu` and `run_dace_gpu` backends were using the gt4py cache. Following the GTFN pattern, this PR introduces separate backends (`run_dace_cpu_cached` and `run_dace_gpu_cached`) to enable the cache. The former backends are configured without cache, and are used in unit tests.
At the same time, this PR improves the cache configuration, by adding caching of the translation stage. The former cache configuration was only caching the compiled programs. With this fix, it is possible to load translated programs (SDFGs) from disk cache.

Note that this PR does not address the limitation on test parallelism (slow down observed when we increase the number of pytest workers).


src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -54,8 +54,14 @@ class Params:
     transforms = backend.DEFAULT_TRANSFORMS
 
 
-run_dace_cpu = DaCeBackendFactory(cached=True, auto_optimize=True)
-run_dace_cpu_noopt = DaCeBackendFactory(cached=True, auto_optimize=False)
+run_dace_cpu = DaCeBackendFactory(auto_optimize=True)
+run_dace_cpu_noopt = DaCeBackendFactory(auto_optimize=False)
+run_dace_cpu_cached = DaCeBackendFactory(
+    auto_optimize=True, cached=True, otf_workflow__cached_translation=True
```
philip-paul-mueller, line 60: Is there a reason why there are two underscores here?

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -61,7 +61,11 @@ def generate_sdfg(
         )
 
         if auto_opt:
-            gtx_transformations.gt_auto_optimize(sdfg, gpu=on_gpu)
+            gtx_transformations.gt_auto_optimize(
+                sdfg,
+                gpu=on_gpu,
+                make_persistent=False,
+            )
```
philip-paul-mueller, line 68: Not for this PR, but we should find a way to pass configuration to the auto optimizer.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace.py
```diff
@@ -17,17 +17,65 @@
...
+        backend=cached_dace_backend,
+        offset_provider={"Ioff": IDim},
+        default_sizes={IDim: 10},
+        grid_type=gtx_common.GridType.CARTESIAN,
+        allocator=cached_dace_backend.allocator,
+    )
+
+
+@pytest.fixture
+def unstructured_case(request, cached_dace_backend, mesh_descriptor):  # noqa: F811
```
philip-paul-mueller, line 67:
```suggestion
def unstructured_case(request, cached_dace_backend, mesh_descriptor):  # noqa: F811 [redefined-while-unused]
```
According to the code style you should also add this short descriptive name.
But why is it actually needed?

philip-paul-mueller (approved):
I have some remarks, but nothing blocking.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
edopao: It comes from the [factory_boy](https://factoryboy.readthedocs.io/en/stable/index.html) module, the two underscores allow to access attribute `cached_translation` of the subfactory `otf_workflow`.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
edopao: Yes, we have a gt4py cycle project for that: toolchain instrumentation.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace.py
edopao: You are right, it is not needed.

edopao (commented):
Thanks for the review!

PR #1943, opened 2025-04-01
feat[next]: Allow setting compile-time stride to 1 for horizontal dimension

havogt:
Enable to propagate stride 1 information.

For now, it can only be enable for *all* fields and only for the horizontal dimension. The 2 lines enabling this are a quick hacky implementation, the proper implementation would allow to pass this as meta data per field argument.

### DIFF
--- a/src/gt4py/next/config.py
+++ b/src/gt4py/next/config.py
@@ -81,3 +81,8 @@ BUILD_CACHE_LIFETIME: BuildCacheLifetime = BuildCacheLifetime[
 CMAKE_BUILD_TYPE: CMakeBuildType = CMakeBuildType[
     os.environ.get("GT4PY_CMAKE_BUILD_TYPE", "debug" if DEBUG else "release").upper()
 ]
+
+#: Experimental, use at your own risk: assume horizontal dimension has stride 1
+UNSTRUCTURED_HORIZONTAL_STRIDE_1: bool = env_flag_to_bool(
+    "GT4PY_UNSTRUCTURED_HORIZONTAL_STRIDE_1", default=False
+)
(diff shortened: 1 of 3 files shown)


src/gt4py/next/config.py
```diff
@@ -81,3 +81,8 @@ def env_flag_to_bool(name: str, default: bool) -> bool:
 CMAKE_BUILD_TYPE: CMakeBuildType = CMakeBuildType[
     os.environ.get("GT4PY_CMAKE_BUILD_TYPE", "debug" if DEBUG else "release").upper()
 ]
+
+#: Experimental, use at your own risk: assume horizontal dimension has stride 1
+UNSTRUCTURED_HORIZONTAL_STRIDE_1: bool = env_flag_to_bool(
```
tehrengruber, line 86: I would rename this to
```suggestion
UNSTRUCTURED_HORIZONTAL_UNIT_STRIDE: bool = env_flag_to_bool(
```
and read this as: "The horizontal dimension is the one with unit stride".

Probably closer to the final solution and also works for cartesian and cpu memory layout where K has unit stride (right?):
```suggestion
UNIT_STRIDE_DIMS: list[str] = ...
```
I'm fine with the initial proposal though.

src/gt4py/next/otf/binding/nanobind.py
```diff
@@ -28,6 +29,7 @@ class Expr(eve.Node):
 
 class DimensionType(Expr):
```
tehrengruber, line 30:
```suggestion
class DimensionSpec(Expr):
```
Unrelated, but feels to me like this would blend more with the rest of the file.

tehrengruber (approved):
Only thing I would change is the naming of the env var, rest is optional / for the future.

src/gt4py/next/config.py
havogt: I added a "HAS" to the name. I think it's fine to be very verbose as you should not use that accidentally. ;)

havogt (commented):
.

PR #1946, opened 2025-04-01
fix[next][dace]: fix orchestration for cached translation stage

edopao:
This PR solves a runtime issue encountered in dace orchestration tests in icon4py, which use the cached backend. The issue follows the change in #1941 that makes the translation stage a `CachedStep` object in the cached backend.


philip-paul-mueller (approved):
LGTM

PR #1949, opened 2025-04-02
fix[dace][next]: Fixes GT4Py State Fusion

philip-paul-mueller:
Before the subgraphs were computed by finding the connected component, however, it turned out that in some cases this was too string and and some read write conflict could not be resolved. The solution was to go back to the `concurrent_subgraph()` function provided by DaCe, which takes into account how the data flows inside the generated code.


edopao (approved):
LGTM, also tested in blue-line granule.

PR #1942, opened 2025-04-01
feat[next]: add flufl lock to compiledb and gtfn program compilation

havogt:
Introduce flufl.lock (which is supposed to be NFS-safe) for locking the compiledb and the gtfn program compilation directory. This allows compiling in multi-rank mpi setups (one rank compiles, the other wait).

Additionally, replace the filelock with flufl.lock in the Filecache.

PR #1951, opened 2025-04-02
feat[next]: Asynchronous pre-compilation

havogt:
Sketch as discussed this morning.

Contains #1942.

TODO:
- [ ] tests
- [ ] check all variants of program do something meaningful with `compile()`


src/gt4py/next/ffront/decorator.py
```diff
@@ -54,6 +57,9 @@
 
 DEFAULT_BACKEND: Optional[next_backend.Backend] = None
 
+# TODO(havogt): We would lke this to be a ProcessPoolExecutor, which requires (to decide what) to pickle.
```
havogt, line 60:
```suggestion
# TODO(havogt): We would like this to be a ProcessPoolExecutor, which requires (to decide what) to pickle.
```

src/gt4py/next/ffront/decorator.py
```diff
@@ -245,10 +258,50 @@ def __call__(self, *args: Any, offset_provider: common.OffsetProvider, **kwargs:
                 ctx.run(self.definition_stage.definition, *args, **kwargs)
             return
 
-        self.backend(
-            self.definition_stage,
-            *args,
-            **(kwargs | {"offset_provider": offset_provider}),
+        if self._compiled_program is not None:
+            # TODO(havogt): we should check that offset_provider_type of compiled program
```
havogt, line 262: I would like to discuss how to implement this.

src/gt4py/next/ffront/decorator.py
```diff
@@ -245,10 +258,50 @@ def __call__(self, *args: Any, offset_provider: common.OffsetProvider, **kwargs:
...
+        past_node_type = self.past_stage.past_node.type
+        arg_types_dict = past_node_type.definition.pos_or_kw_args
+
+        # TODO(havogt): we currently don't support pos_only or kw_only args at the program level
+        assert not past_node_type.definition.kw_only_args
+        assert not past_node_type.definition.pos_only_args
+
+        args = tuple(arg_types_dict.values())
+        compile_time_args = arguments.CompileTimeArgs(
+            offset_provider=offset_provider_type, column_axis=None, args=args, kwargs={}
```
havogt, line 297: TODO check column_axis

havogt: added a test

src/gt4py/next/ffront/decorator.py
```diff
@@ -244,11 +258,59 @@ def __call__(self, *args: Any, offset_provider: common.OffsetProvider, **kwargs:
...
-            self.definition_stage,
-            *args,
-            **(kwargs | {"offset_provider": offset_provider}),
+        if self._compiled_program is not None:
+            # TODO(havogt): make offset_provider_type part of the compiled program hash once we have variants
+            # TODO(havogt): measure overhead of canonicalize_arguments
+            program_type = self.past_stage.past_node.type
+            assert isinstance(program_type, ts_ffront.ProgramType)
+            args, kwargs = type_info.canonicalize_arguments(program_type, args, kwargs)
+            self._compiled_program.result()(*args, **kwargs, offset_provider=offset_provider)
```
egparedes, line 267: After the first call to `result()` we could save the value directly instead of calling `result()` multiple times. I don't know if it is really worth it just to save one (probably cheap) function call, but if it is, one approach could be something like:
```suggestion
            try:
                self._compiled_program(*args, **kwargs, offset_provider=offset_provider)
            except NotCallableSomethiongError:
                self._compiled_program = self._compiled_program.result()
                self._compiled_program(*args, **kwargs, offset_provider=offset_provider)
```

egparedes (commented):
It looks good, just some simple comments.

src/gt4py/next/ffront/decorator.py
```diff
@@ -244,11 +258,59 @@ def __call__(self, *args: Any, offset_provider: common.OffsetProvider, **kwargs:
...
+        else:
+            self.backend(
+                self.definition_stage,
+                *args,
+                **(kwargs | {"offset_provider": offset_provider}),
+            )
+
+    def compile(
```
egparedes, line 275: What is the intended behavior if `compile()` is called multiple times in the same program?

src/gt4py/next/ffront/decorator.py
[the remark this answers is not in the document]
havogt: That's what we briefly discussed. Currently, I would have to store somewhere the `offset_provider_type` that was used for compilation to be able to match it here. Then here in the hot path, we would have to extract `offset_provider_type` from `offset_provider` and compare to the stored one.

In the follow-up PR with variants, this would be slightly easier: we make offset_provider_type part of the lookup of the program, if we don't find it, we will error. But then there is still the problem of a non-trivial conversion in the hot path. I would propose we discuss it there and leave this unsafe for now. There is the TODO at the top.

havogt: Should we do it? In the profile it wasn't visible.

src/gt4py/next/ffront/decorator.py
havogt: I think it should be an error, I'll add a check.

src/gt4py/next/ffront/decorator.py
```diff
@@ -75,8 +81,12 @@ class Program:
...
     backend: Optional[next_backend.Backend]
     connectivities: Optional[
-        common.OffsetProvider  # TODO(ricoh): replace with common.OffsetProviderType once the temporary pass doesn't require the runtime information
-    ]
+        common.OffsetProvider
+    ]  # TODO(ricoh): replace with common.OffsetProviderType once the temporary pass doesn't require the runtime information
+
+    _compiled_program: concurrent.futures.Future[stages.CompiledProgram] | None = dataclasses.field(
+        init=False, default=None
+    )
```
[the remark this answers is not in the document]
havogt: I forgot why I did it, but mypy told me again: we are inheriting from this class and add a new member which doesn't have a default, therefore this pattern. We need to do some cleanups...

src/gt4py/next/ffront/decorator.py
havogt: I did it.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_decorator.py
```diff
@@ -64,3 +64,97 @@ def testee(a: cases.IField, out: cases.IField):
...
+        pytest.skip("Embedded compiled program doesn't make sense.")
+
+    assert compile_testee._compiled_program is None
+    compile_testee.compile(offset_provider_type=cartesian_case.offset_provider)
+    assert compile_testee._compiled_program is not None
+
+    args, kwargs = cases.get_default_data(cartesian_case, compile_testee)
+
+    # make sure the backend is never called
+    object.__setattr__(compile_testee, "backend", _always_raise_callable)
```
egparedes, line 110: Maybe `unittest.mock.patch()` is a better fit here (and other similar cases below), since it's just a modification for testing.

egparedes (approved):
LGTM

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_decorator.py
havogt: I'll merge it and try that in the follow-up PR.

PR #1952, opened 2025-04-02
fix[dace][next]: Added a second way to specify the leading dimension.

philip-paul-mueller:
Before it was only possible to specify which dimension should be considered as "leading dimensions", i.e. associated to the stride 1. This PR adds a second method, which can be selected through the `leading_kind` argument, which is a GT4Py `DimensionKind` enum value. This allows to generally specify if horizontal or vertical dimensions are leading dimensions.


philip-paul-mueller:
I have no issue with the proposed change.

edopao:
I think `leading_dim` is used in two contexts in dace auto-optimize: for stride-1 constant substitution (where `unit_stride` is a valid replacement) and for sorting of the iteration dimension. In the latter case, it is not necessary that the memory layout of fields has stride 1 in one dimension (although it is the most common case). In this sense, `leading_dim` is more generic (although confusing, I misunderstood its meaning when I first tried to use it).

philip-paul-mueller:
Regarding the name I think that telling the auto optimizer where the stride 1 is (at least for global data) is the better way.
It can then figuring out on its own what the correct order of the loop iterations is.
However, this is kind of broken and I do not have time to fix it, because until now it was not so important, because for our cases it just happened to just work.
For temporaries it should determine them on its own.

Furthermore, the _only_ thing that the auto optimizer does regarding the stride 1 constant substitution is that it calls the substitution function, which strides should be replaced with one has to be done on the outside, the auto optimizer just does it.
I want to change that in the long term but I simply lack the time and we would also have to do it for temporaries, which is hard.
The auto optimizer currently does some stride reordering for transients, but it is quite brittle (it assumes the lowering gives us C order which is okay for CPU and just swaps it to FROTRAN order on the GPU).
The main reason for this is, that in the auto optimizer it is very hard to figuring out which dimension of an Array (especially for transients) is associated to which dimension.

src/gt4py/next/program_processors/runners/dace/transformations/map_orderer.py
```diff
@@ -56,60 +63,80 @@ class MapIterationOrder(dace_transformation.SingleStateTransformation):
...
     - `Map[VertexDim, KDim]` -> `Map[KDim, VertexDim]`.
     - `Map[EdgeDim, KDim]` -> `Map[KDim, EdgeDim]`.
     - `Map[CellDim, KDim]` -> `Map[CellDim, KDim]` (no modification).
 
+    The second way is to specify `unit_strides_kind`, which is a GT4Py `DimensionKind`
+    enum value. In that case the all dimensions that are associated to this kind
+    will be ordered such that they are the inner most. If a Map has multiple
+    dimensions of the same kind, then their relative order is unspecific.
+
+    In any case the order of the non matches dimensions is unspecific.
```
edopao, line 82:
```suggestion
    In any case the order of the unmatched dimensions is undefined.
```

src/gt4py/next/program_processors/runners/dace/transformations/map_orderer.py
```diff
@@ -56,60 +63,80 @@ class MapIterationOrder(dace_transformation.SingleStateTransformation):
...
     following:
     - `Map[EdgeDim, KDim, VertexDim]` -> `Map[KDim, VertexDim, EdgeDim]`.
     - `Map[VertexDim, KDim]` -> `Map[KDim, VertexDim]`.
     - `Map[EdgeDim, KDim]` -> `Map[KDim, EdgeDim]`.
     - `Map[CellDim, KDim]` -> `Map[CellDim, KDim]` (no modification).
 
+    The second way is to specify `unit_strides_kind`, which is a GT4Py `DimensionKind`
+    enum value. In that case the all dimensions that are associated to this kind
+    will be ordered such that they are the inner most. If a Map has multiple
+    dimensions of the same kind, then their relative order is unspecific.
```
edopao, line 80:
```suggestion
    dimensions of the same kind, then their relative order is undefined.
```

edopao (approved):

PR #1953, opened 2025-04-03
Fix[dace][next]: Fixed Local Buffer Transformation

philip-paul-mueller:
Modified how and when local buffers are added.

There were some issues with this functionality, the primary bug was that the dataflow downstream was not updated, for example if the data was consumed by a Map, then the dataflow there was not updated, i.e. the Memlets would still refer to the old (outer) memory.
Then there was a second bug.
The bug was discovered because the transformation was accidentally run multiple times, once before `LoopBlocking` (as it was intended) and once after it.
This lead to the creation of buffers that were more than one scalar.
While this was a behaviour that the implementation allowed it was realized that it was wrong, because we assume pointwise and need this buffers only to ensure some global order of execution.
Thus the transformation was updated to not allow this case any further.

However, the whole transformation needs to be reworked as it is most likely fundamentally flawed.


src/gt4py/next/program_processors/runners/dace/transformations/utils.py
philip-paul-mueller, line 1: You can ignore the changes to this file, since they are actually included in [PR#1954](https://github.com/GridTools/gt4py/pull/1954), I just needed theme here, but I did not bother to properly put them here.

PR #1954, opened 2025-04-03
feat[dace][next]: Added the `SplitAccessNode` Transformation

philip-paul-mueller:
The `SplitAccessNode` searches for AccessNodes with multiple producers and consumers.
For every consumer it then identifies which (single) producer covers its dependencies, i.e. if everything the consumer reads from the AccessNode is generated by a single producer (incoming edge).
If this is the case then the AccessNode is split, thus instead of writing into a single AccessNode, each producer writes into a dedicated AccessNode and the consumer directly reads from there.
Note that this is a first version that might need more extensions depending on the situation.


edopao (approved):
LGTM

PR #1958, opened 2025-04-04
feat: set the CUDA architecture automatically in gtfn compilation

egparedes:
Set the CUDA architecture flag from the device information exposed by cupy. The user can still override it by defining the standard CMake `CUDAARGS` env variable.

Additional changes:
- Minor refactoring of the compiledb factory to reuse code from cmake factory
- Add configure and build scripts to the generated project folders for letting developers to reproduce and tinker with the build process during debugging

### DIFF
--- a/src/gt4py/next/otf/compilation/build_systems/compiledb.py
+++ b/src/gt4py/next/otf/compilation/build_systems/compiledb.py
@@ -176,7 +176,7 @@ class CompiledbProject(
         (self.root_path / "compile_commands.json").write_text(json.dumps(compile_db))
 
         (build_script_path := self.root_path / "build.sh").write_text(
-            "\n".join(["#!/bin/sh", "cd build"] + [entry["command"] for entry in compile_db])
+            "\n".join(["#!/bin/sh", "cd build", *(entry["command"] for entry in compile_db)])
         )
         try:
             build_script_path.chmod(0o755)


src/gt4py/next/otf/compilation/build_systems/cmake.py
```diff
@@ -9,16 +9,45 @@
...
+def get_cmake_device_arch_option() -> str:
+    cmake_flag = ""
+
+    device_archs = os.environ.get("CUDAARCHS", get_device_arch() or "").strip()
+    if device_archs:
+        match core_defs.CUPY_DEVICE_TYPE:
+            case core_defs.DeviceType.CUDA:
+                cmake_flag = f"DCMAKE_CUDA_ARCHITECTURES={device_archs}"
+            case core_defs.DeviceType.ROCM:
+                # `CMAKE_HIP_ARCHITECTURES` is not officially supported yet
```
havogt, line 45:
```suggestion
```
this is officially supported, but there is no environment variable similar to `CUDAARCHS`. You chose to use CUDAARCHS also for HIP, which is ok for now, I think. Maybe add a comment.

src/gt4py/next/otf/compilation/build_systems/cmake.py
```diff
@@ -9,16 +9,47 @@
...
+        # TODO(egparedes): Implement this properly, either parsing the output of `$ rocminfo`
+        # or using the HIP low level bindings.
+        # Check: https://rocm.docs.amd.com/projects/hip-python/en/latest/user_guide/1_usage.html
+        return "gfx90a"
+
+    return None
+
+
+def get_cmake_device_arch_option() -> str:
+    cmake_flag: str
```
havogt, line 37:
```suggestion
    cmake_flag = ""
```
and remove the `case None`?

src/gt4py/next/otf/compilation/build_systems/cmake.py
```diff
@@ -9,16 +9,47 @@
...
+
+    match core_defs.CUPY_DEVICE_TYPE:
+        case core_defs.DeviceType.CUDA:
+            device_archs = os.environ.get("CUDAARCHS", "").strip() or get_device_arch()
+            cmake_flag = f"DCMAKE_CUDA_ARCHITECTURES={device_archs}"
```
tehrengruber, line 42:
```suggestion
            cmake_flag = f"-DCMAKE_CUDA_ARCHITECTURES={device_archs}"
```

src/gt4py/next/otf/compilation/build_systems/cmake.py
egparedes: Done

src/gt4py/next/otf/compilation/build_systems/cmake_lists.py
```diff
@@ -45,12 +45,6 @@ class CMakeListsGenerator(eve.codegen.TemplatedGenerator):
         project({{project_name}})
 
         # Languages
-        if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
```
havogt, line 48: Maybe instead of reverting the changes, we could generate a script that contains the cmake invocation command (maybe including the `env` when the command was executed). This leaves the cleaner version as implemented here, but allows to re-run from scratch.

havogt: I came to this idea, because initially I wanted to propose to write down: our strategy is to put all the information in the cmakelists.txt such that it can be re-run, then I realized that this would also mean putting the compilers etc into the file, which might or might not be what the person wants that tries to re-configure.

src/gt4py/next/otf/compilation/build_systems/compiledb.py
```diff
@@ -331,6 +336,17 @@ def _cc_create_compiledb(
...
 
+    build_script_path = path / "build.sh"
+    try:
+        build_script_path.chmod(0o755)
+    except Exception:
+        pass
+    with build_script_path.open("w") as build_script_pointer:
+        build_script_pointer.write("#!/bin/sh\n")
+        for entry in compile_db:
+            build_script_pointer.write(f"{entry['command']}\n")
```
havogt, line 347: This is not working, it contains the non expanded parameters. Anyway the build command is already included in the log, the thing that's missing is the configure script (but not too many use-cases for that, I guess). Anyway, I opened a PR in your fork with a proposal.

havogt (changes requested):

src/gt4py/next/otf/compilation/build_systems/compiledb.py
egparedes: Fixed

DropD (approved):
Nice, good improvements, no obvious problems that I can see.

PR #1962, opened 2025-04-08
bug[next]: fix constant folding of expression with boolean literals

havogt:
`_value_from_literal` for booleans did `bool("False")` which evaluates to `True`.

### DIFF
--- a/src/gt4py/next/iterator/embedded.py
+++ b/src/gt4py/next/iterator/embedded.py
@@ -523,38 +523,33 @@ def promote_scalars(val: CompositeOfScalarOrField):
         )
 
 
+_python_builtins: dict[str, Callable] = {
+    "bool": bool,
+    "str": str,
+    "plus": operator.add,
+    "minus": operator.sub,
+    "multiplies": operator.mul,
+    "divides": operator.truediv,
+    "mod": operator.mod,
+    "floordiv": operator.floordiv,
+    "eq": operator.eq,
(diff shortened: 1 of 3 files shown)


src/gt4py/next/iterator/transforms/constant_folding.py
```diff
@@ -15,12 +15,16 @@
...
+from gt4py._core import definitions as core_defs
 from gt4py.next.iterator import builtins, embedded, ir
 from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm, ir_makers as im
 from gt4py.next.iterator.transforms import fixed_point_transformation
+from gt4py.next.type_system import type_specifications as ts
 
 
-def _value_from_literal(literal: ir.Literal):
+def _value_from_literal(literal: ir.Literal) -> core_defs.Scalar:
+    if literal.type.kind == ts.ScalarKind.BOOL:
```
havogt, line 26: here is the bugfix

src/gt4py/next/iterator/embedded.py
```diff
@@ -523,38 +523,33 @@ def promote_scalars(val: CompositeOfScalarOrField):
         )
 
 
+_python_builtins: dict[str, Callable] = {
```
havogt, line 526: this change is unrelated (note I removed float and int from the list) and consider the TODO resolved. Will role back if this breaks something...

PR #1963, opened 2025-04-08
bug[next]: GTFN reduce python overhead

havogt:
Reduces Python overhead in the GTFN call. Replaces the implicit copy of connectivities to device by a check that it's on device if needed.

Similar changes in DaCe will be in a next PR.

### DIFF
--- a/src/gt4py/next/program_processors/runners/gtfn.py
+++ b/src/gt4py/next/program_processors/runners/gtfn.py
@@ -29,12 +29,12 @@ from gt4py.next.program_processors.codegens.gtfn import gtfn_module
 
 # TODO(ricoh): Add support for the whole range of arguments that can be passed to a fencil.
 def convert_arg(arg: Any) -> Any:
+    if origin := getattr(
+        arg, "__gt_origin__", None
+    ):  # TODO: double-check all fields have __gt_origin__
+        return arg.ndarray, origin
     if isinstance(arg, tuple):
         return tuple(convert_arg(a) for a in arg)
-    if isinstance(arg, common.Field):
-        arr = arg.ndarray
-        origin = getattr(arg, "__gt_origin__", tuple([0] * len(arg.domain)))
-        return arr, origin
     if isinstance(arg, np.bool_):
         # nanobind does not support implicit conversion of `np.bool` to `bool`
         return bool(arg)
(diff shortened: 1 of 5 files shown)


src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -64,42 +61,30 @@ def decorated_program(
     return decorated_program
 
 
-def _ensure_is_on_device(
+def _verify_connectivity_array_type(
```
edopao, line 64: Can you make the same change in dace backend? we have similar function:
https://github.com/GridTools/gt4py/blob/76a076dd62bd82107ccfff8d39c602705bb6f917/src/gt4py/next/program_processors/runners/dace/sdfg_callable.py#L63

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -29,17 +28,14 @@
 
 # TODO(ricoh): Add support for the whole range of arguments that can be passed to a fencil.
 def convert_arg(arg: Any) -> Any:
+    if (origin := getattr(arg, "__gt_origin__", None)) is not None:
```
edopao, line 31: We have a similar function in dace backend:
https://github.com/GridTools/gt4py/blob/76a076dd62bd82107ccfff8d39c602705bb6f917/src/gt4py/next/program_processors/runners/dace/sdfg_callable.py#L27

src/gt4py/next/program_processors/runners/gtfn.py
havogt: Will do, once we agree that we want to do this change.

src/gt4py/next/program_processors/runners/gtfn.py
havogt: I briefly looked at the decoration_stage of DaCe, I guess we have to do a bit of work to get that fast.

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -29,17 +28,14 @@
...
+        return arg.ndarray, origin
```
tehrengruber, line 32:
```suggestion
        assert isinstance(arg, common.Field)
        return arg.ndarray, origin
```
For correctness & readability when executing in debug mode.

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -64,42 +61,30 @@ def decorated_program(
...
-            conn_arg = _ensure_is_on_device(conn.ndarray, device)
-            args.append((conn_arg, tuple([0] * 2)))
-        elif isinstance(conn, common.Dimension):
-            pass
-        else:
-            raise AssertionError(
-                f"Expected offset provider '{name}' to be a 'Connectivity' or 'Dimension', "
-                f"but got '{type(conn).__name__}'."
-            )
+    # Note: the order here needs to agree with the order of the generated bindings
```
tehrengruber, line 80: Maybe cast this into an assert

src/gt4py/next/program_processors/runners/gtfn.py
havogt: Let's do that in a next PR. We can do that together as I have the profiling setup ready.

egparedes (approved):
LGTM

PR #1964, opened 2025-04-09
fix[next]: Properly show compilation errors if compilation fails

havogt:
Before this PR, the actual compilation error can only be found in the build directory (and only if build with persistent cache). Now, the error is attached to the exception.

### DIFF
--- a/src/gt4py/next/otf/compilation/build_systems/compiledb.py
+++ b/src/gt4py/next/otf/compilation/build_systems/compiledb.py
@@ -194,8 +194,8 @@ class CompiledbProject(
                     )
         except subprocess.CalledProcessError as e:
             with logfile.open(mode="r") as log_file_pointer:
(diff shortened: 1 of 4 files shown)


src/gt4py/next/otf/compilation/build_systems/compiledb.py
```diff
@@ -193,8 +194,8 @@ def _run_build(self) -> None:
                     )
         except subprocess.CalledProcessError as e:
             with logfile.open(mode="r") as log_file_pointer:
-                print(log_file_pointer.read())
-            raise e
+                msg = log_file_pointer.read()
+            raise errors.CompilationError(msg) from e
```
egparedes, line 198:
```suggestion
                log = log_file_pointer.read()
            raise errors.CompilationError(log) from e
```

egparedes (approved):
LGTM

PR #1965, opened 2025-04-10
fix[next]: use gridtools::nanobind::dynamic_stride for  compatibility

havogt:
In #1943 we introduced stride 1 in a way that is not backwards compatible with nanobind < 2.0. Instead of `-1` we should use the `gridtools::nanobind::dyamic_size` to indicate runtime strides.

### DIFF
--- a/.github/workflows/daily-ci.yml
+++ b/.github/workflows/daily-ci.yml
@@ -7,9 +7,9 @@ on:
 
   ## COMMENTED OUT: only for testing CI action changes.
   ## It only works for PRs to `main` branch from branches in the upstream gt4py repo.
-  pull_request:
-    branches:
-    - main
+  # pull_request:
+  #   branches:
+  #   - main
   ## END
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/daily-ci.yml
```diff
@@ -7,9 +7,9 @@ on:
 
   ## COMMENTED OUT: only for testing CI action changes.
   ## It only works for PRs to `main` branch from branches in the upstream gt4py repo.
-  # pull_request:
-  #   branches:
-  #   - main
+  pull_request:
+    branches:
+    - main
```
havogt, line 12:
```suggestion
  # pull_request:
  #   branches:
  #   - main
```

egparedes (approved):
LGTM. Other option would be to just raise the minimum required version for `nanobind` to `>=2`. I'm fine with both solutions

havogt:
Agree, I think this is cleaner anyway (more descriptive in the generated code), but I was considering doing both.

PR #1972, opened 2025-04-14
bug[next]: Fix gtfn backend error when cupy installed but no gpu present

tehrengruber:
Small fix for a bug introduced in #1958

### DIFF
--- a/src/gt4py/next/otf/compilation/build_systems/cmake.py
+++ b/src/gt4py/next/otf/compilation/build_systems/cmake.py
@@ -12,6 +12,7 @@ import dataclasses
 import os
 import pathlib
 import subprocess
+import warnings
 
 from gt4py._core import definitions as core_defs
 from gt4py.next import config
@@ -25,9 +26,11 @@ def get_device_arch() -> str | None:
         # use `cp` from core_defs to avoid trying to re-import cupy
         try:
             return core_defs.cp.cuda.Device(0).compute_capability  # type: ignore[attr-defined]
-        except core_defs.cp.cuda.runtime.CUDARuntimeError as e:
-            if "no CUDA-capable device is detected" not in str(e):
-                raise e
+        except core_defs.cp.cuda.runtime.CUDARuntimeError as e:  # type: ignore[attr-defined]
+            warnings.warn(
+                UserWarning(f"Could not determine the CUDA compute capability: {e}"), stacklevel=2
+            )
+            return None
     elif core_defs.CUPY_DEVICE_TYPE == core_defs.DeviceType.ROCM:
         # TODO(egparedes): Implement this properly, either parsing the output of `$ rocminfo`
         # or using the HIP low level bindings.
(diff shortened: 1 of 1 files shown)


havogt (commented):
~~Or should we provide some default?~~

Didn't carfully read the code...

src/gt4py/next/otf/compilation/build_systems/cmake.py
```diff
@@ -23,7 +23,11 @@
 def get_device_arch() -> str | None:
     if core_defs.CUPY_DEVICE_TYPE == core_defs.DeviceType.CUDA:
         # use `cp` from core_defs to avoid trying to re-import cupy
-        return core_defs.cp.cuda.Device(0).compute_capability  # type: ignore[attr-defined]
+        try:
+            return core_defs.cp.cuda.Device(0).compute_capability  # type: ignore[attr-defined]
+        except core_defs.cp.cuda.runtime.CUDARuntimeError as e:
+            if "no CUDA-capable device is detected" not in str(e):
+                raise e
```
egparedes, line 30: After discussing the issue with @havogt, the best solution seems to be to remove this snippet and add a check for non-empty `device_archs` value inside `get_cmake_device_arch_options()`:  

```python
    match core_defs.CUPY_DEVICE_TYPE:
        case core_defs.DeviceType.CUDA:
            device_archs = os.environ.get("CUDAARCHS", "").strip() or get_device_arch()
            cmake_flag_template = "-DCMAKE_CUDA_ARCHITECTURES={device_archs}"
        case core_defs.DeviceType.ROCM:
            # `HIPARCHS` is not officially supported by CMake yet, but it might be in the future
            device_archs = os.environ.get("HIPARCHS", "").strip() or get_device_arch()
            cmake_flag_template = "-DCMAKE_HIP_ARCHITECTURES={device_archs}"

    cmake_flag = cmake_flag_template.format(device_archs=device_archs) if device_archs else ""

    return cmake_flag
```

egparedes (changes requested):
I think there is a better fix for this.

src/gt4py/next/otf/compilation/build_systems/cmake.py
havogt: Not entirely remove, but I would remove the special case for no CUDA-capable device" and on any error just return None. If cuda runtime fails, I would not error here in the building stage.

src/gt4py/next/otf/compilation/build_systems/cmake.py
```diff
@@ -23,7 +24,13 @@
 def get_device_arch() -> str | None:
     if core_defs.CUPY_DEVICE_TYPE == core_defs.DeviceType.CUDA:
         # use `cp` from core_defs to avoid trying to re-import cupy
-        return core_defs.cp.cuda.Device(0).compute_capability  # type: ignore[attr-defined]
+        try:
+            return core_defs.cp.cuda.Device(0).compute_capability  # type: ignore[attr-defined]
+        except core_defs.cp.cuda.runtime.CUDARuntimeError as e:  # type: ignore[attr-defined]
+            warnings.warn(
```
havogt, line 30: Warning looks like

```
/workspace/gt4py/src/gt4py/next/otf/compilation/build_systems/cmake.py:42: UserWarning: Could not determine the CUDA compute capability: cudaErrorInsufficientDriver: CUDA driver version is insufficient for CUDA runtime version
```

tehrengruber: I would rather not get this warning all the time^^ I think there is nothing to warn in case you don't have a gpu.

havogt: What's your use-case? I tried this on gitpod (which is my only setup without gpu), and for the tests I didn't even get to the point where this warning was triggered. There was a hard-error before.

havogt: I pushed one more change maybe that fixes your use-case?

havogt (approved):

PR #1973, opened 2025-04-14
feat[dace][next]: Updated `LoopBlocking`

philip-paul-mueller:
This PR updates the classification function of the `LoopBlocking` transformation.
The transformation is now able to detect is a nested Map depends on the blocking parameter or not, thus the map can be moved outside of the inner map.
Furthermore, the classification function is now also able to handle reduction nodes.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
@@ -282,7 +282,8 @@ class LoopBlocking(dace_transformation.SingleStateTransformation):
 
         while True:
             # Find all the nodes that we have to classify in this iteration.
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -343,9 +344,12 @@ def _classify_node(
         - All incoming edges must start either at `self.outer_entry` or at dependent nodes.
         - All output Memlets are non empty.
 
+        It is important that to realize that the function will add the node to
```
edopao, line 347:
```suggestion
        The function will first add the node to classify to `self._independent_nodes`on its own. Then, the function might add other nodes beside `node_to_classify`.
```

edopao (approved):

PR #1832, opened 2025-01-29
build: migrate from bump-my-version to versioningit

egparedes:
Use `versioningit` as a `setuptools` plugin to use the information in git tags as source of truth for the distribution package versioning. This should allow for a much finer version specification, since essentially every commit defines a new version, even when installed in editable mode. Some other important packages are following similar approaches (e.g. jax).  

Changes:
- Add versioningit configuration and custom logic for getting the precise version at run-time in different scenarios
- Add nox session and github workflow to run the new package-level tests on CI
- Update CI and release documentation
- Add `pyproject.toml export-subst` to `.gitattributes` as explained in the `git-archive` method [documentation](https://versioningit.readthedocs.io/en/v1.1.0/configuration.html#git-archive)
- Fix typos in README and dev-tasks.py

PR #1975, opened 2025-04-14
bug[cartesian]: fix setuptools for CUDA compilation

havogt:
In migration to versioningit (#1832 ), a `setuptools.cmdclass` is set
```python
[tool.setuptools.cmdclass]
# This is required for the `onbuild` versioningit hook
build_py = "versioningit.cmdclass.build_py"
sdist = "versioningit.cmdclass.sdist"
```
This overrides the `cmdclass` for CUDA that is passed to the `pyext_builder` (instead of extending it).

Now, we set extend the `cmdclass` after the `Distribution` is created by using internal hooks to first create the Distribution and then execute it.

The code is not thread-safe!

---------

Co-authored-by: Enrique González Paredes <enriqueg@cscs.ch>

### DIFF
--- a/src/gt4py/cartesian/backend/pyext_builder.py
+++ b/src/gt4py/cartesian/backend/pyext_builder.py
@@ -11,7 +11,7 @@ import copy
 import io
 import os
 import shutil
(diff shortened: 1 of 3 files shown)


egparedes (approved):
LGTM

PR #1976, opened 2025-04-14
build[next]: update dace version from main

edopao:
We update the dace version form main development branch in order to include https://github.com/spcl/dace/pull/1984

This PR solves an issue in cpu code generation that caused a compilation error for MacOS users, on the dace cpu backend, because a warning in the reduce function template was treated as an error by clang.


edopao:
We use #1977, I'll close my PR.

PR #1978, opened 2025-04-15
feat[next]: Performance timers

havogt:
Adds a config option `COLLECT_METRICS_LEVEL` to collect performance metrics with different levels of granularity. If enabled it will collect for each program some of the following metrics:
- total runtime: if programs are not pre-compiled this contains everything from call to result.
- compute time: time spend on the C++ side in gtfn, for CUDA this adds a `deviceSynchronize` after the kernel which may impact total runtime of an application.

Additional timers of the transformation/compilation pipeline could be added like the following:
- create a new `TimedStep` that wraps any workflow step
- if metrics collection is enabled, we could wrap the steps in `OTFCompileWorkflow` with the `TimedStep`.


src/gt4py/next/ffront/decorator.py
```diff
@@ -259,6 +264,9 @@ def __call__(self, *args: Any, offset_provider: common.OffsetProvider, **kwargs:
                 ctx.run(self.definition_stage.definition, *args, **kwargs)
             return
 
+        if config.COLLECT_METRICS:
+            offset_provider["exec_info"] = metrics.global_metric_container[self.__name__]
```
havogt, line 268: the only reason why I didn't get rid of this hack is that I don't have the name available in the gtfn run...

src/gt4py/next/otf/binding/nanobind.py
```diff
@@ -117,16 +123,31 @@ class BindingCodeGenerator(TemplatedGenerator):
...
         {
+            if (exec_info.has_value()) {                
+                exec_info->operator[]("run_cpp_start_time") = static_cast<double>(
+                        std::chrono::duration_cast<std::chrono::nanoseconds>(
+                            std::chrono::high_resolution_clock::now().time_since_epoch()).count())/1e9;
+            }
             {{body}}
+            {% if sync %}
+            cudaDeviceSynchronize();
+            {% endif %}
```
havogt, line 137: not sure why I added this here, probably copied from cartesian, should be just unconditionally in the next if

src/gt4py/next/metrics.py
```diff
@@ -0,0 +1,48 @@
...
+@dataclasses.dataclass
+class RuntimeMetric:
+    cpp_time: MetricAccumulator = dataclasses.field(default_factory=MetricAccumulator)
+    total_time: MetricAccumulator = dataclasses.field(default_factory=MetricAccumulator)
+
+
+global_metric_container: dict[str, RuntimeMetric] = defaultdict(RuntimeMetric)
+
+
+def summary() -> None:
```
havogt, line 41:
```suggestion
def print_stats() -> None:
```

src/gt4py/next/metrics.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@dataclasses.dataclass
+class RuntimeMetric:
+    cpp_time: MetricAccumulator = dataclasses.field(default_factory=MetricAccumulator)
+    total_time: MetricAccumulator = dataclasses.field(default_factory=MetricAccumulator)
+
+
+global_metric_container: dict[str, RuntimeMetric] = defaultdict(RuntimeMetric)
+
```
havogt, line 39: add `dump_stats()`

src/gt4py/next/otf/binding/nanobind.py
```diff
@@ -117,16 +123,30 @@ class BindingCodeGenerator(TemplatedGenerator):
...
+                            std::chrono::high_resolution_clock::now().time_since_epoch()).count())/1e9;
+            }
+            auto gt4py_res_ = {{body}}
+            if (exec_info.has_value()) {
+                {% if _this_node.device %}cudaDeviceSynchronize();{% endif %}
+                exec_info->operator[]("run_cpp_end_time") = static_cast<double>(
+                        std::chrono::duration_cast<std::chrono::nanoseconds>(
+                            std::chrono::high_resolution_clock::now().time_since_epoch()).count())/1e9;
+            }
+            return gt4py_res_;
```
havogt, line 141: Is it actually needed to measure this on the cpp side?

src/gt4py/next/otf/binding/nanobind.py
```diff
@@ -114,20 +121,40 @@ class BindingCodeGenerator(TemplatedGenerator):
...
+        self, node: WrapperFunction, **kwargs: Any
+    ) -> Union[str, Collection[str]]:
+        return_stmt = "return _gt4py_return;" if isinstance(node.body, ReturnStmt) else ""
+        return self.generic_visit(node, return_stmt=return_stmt)
+
     WrapperFunction = as_jinja(
         """\
         decltype(auto) {{name}}(
-            {{"\n,".join(parameters)}}
+            {{"\n,".join(parameters + ["std::optional<nanobind::dict> exec_info"])}}
```
havogt, line 133: try once how the cpp timers compare to timers on python side

src/gt4py/next/metrics.py
```diff
@@ -0,0 +1,87 @@
...
+    def __str__(self) -> str:
+        return f"{self.mean():.5e} +/- {self.std():.5e}"
+
+    def append(self, value: float) -> None:
+        self.values.append(value)
+
+
+@dataclasses.dataclass
+class RuntimeMetric:
```
havogt, line 36: default_dict

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -56,6 +57,8 @@ def apply_common_transforms(
     #: more details.
     symbolic_domain_sizes: Optional[dict[str, str]] = None,
 ) -> itir.Program:
+    if config.COLLECT_METRICS:
```
havogt, line 60: remove "transforms", sketch idea of TimedStep

src/gt4py/next/metrics.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+
+    name: str | None = None
+    samples: MutableSequence[float] = dataclasses.field(default_factory=list)
+
+    def __post_init__(self) -> None:
+        if self.name:
+            object.__setattr__(self, "name", sys.intern(self.name))
+
+    @property
```
edopao, line 58: A Python question. What happens if I make multiple calls to `mean` attribute, which is cached, and in between the calls the `samples` list is updated? Do I get an updated `mean` value or the cached one?

src/gt4py/eve/utils.py
```diff
@@ -256,6 +258,75 @@ def __getitem__(self, key: _S | Tuple[_S, ...]) -> _T:
...
+
+    def __getitem__(self, key: _K) -> _V:
+        try:
+            return super().__getitem__(key)
+        except KeyError:
+            self.data[key] = (value := self.value_factory(key))
+            return value
+
+
+class custom_defaultdict(dict[_K, _V]):
```
edopao, line 298: This class is not used. Do you want to keep it?

src/gt4py/next/metrics.py
```diff
@@ -0,0 +1,205 @@
...
+#: Global metric collection store for the entire program.
+program_metrics: MetricCollectionStore[str] = MetricCollectionStore()
+
+#: Context variable to store the active metric collection of the current program.
+active_metric_collection: contextvars.ContextVar[MetricCollection | None] = contextvars.ContextVar(
+    "active_metric_collection", default=None
+)
+
+
+def dumps(metric_cs: MetricCollectionStore | None = None) -> str:
```
edopao, line 147: The function name here is not much different from the next one (`dump`). I don't know if this is a naming convention, where the `s` at the end stands for string. I see that the same pattern is used for json (`dumps_json` and `dump_json`).

edopao (commented):
Looks very good, I just have some minor comments.

tests/next_tests/unit_tests/test_metrics.py
```diff
@@ -0,0 +1,119 @@
...
+
+@pytest.fixture
+def empty_metrics():
+    return metrics.MetricCollectionStore({"program1": {}, "program2": {}})
+
+
+def test_dumps(sample_metrics):
+    """Test that dumps correctly formats the metrics as a string table."""
+    result = metrics.dumps(sample_metrics)
+    assert isinstance(result, str)
```
edopao, line 40: Should we maybe go line by line and assert as below (the split is remove duplicated spaces)?

`' '.join(line.split()) == "prorgam1 2.00000e+00 1.00000e+00 5.00000e+00 1.00000e+00"`

Do you think that this over-testing?

src/gt4py/next/metrics.py
egparedes: `@property` is always executed on access so here there is not problem. `@functools.cached_property` only runs the first time is accessed, and from that point on, the cached value is used.

src/gt4py/next/otf/binding/nanobind.py
```diff
@@ -59,13 +64,15 @@ class FunctionParameter(eve.Node):
 class WrapperFunction(eve.Node):
     name: str
     parameters: Sequence[FunctionParameter]
-    body: ReturnStmt
+    body: Union[ExprStmt, ReturnStmt]
+    device: bool = False
```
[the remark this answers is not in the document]
egparedes: Changed to `.on_device`

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -58,13 +58,25 @@ def decorated_program(
...
             *converted_args,
             *(arguments.iter_size_args(args) if inp.implicit_domain else ()),
             *conn_args,
+            **opt_kwargs,
         )
 
+        if collect_metrics:
+            assert metrics.active_metric_collection.get() is not None
+            value = exec_info["run_cpp_end_time"] - exec_info["run_cpp_start_time"]
+            metrics.active_metric_collection.get().add_sample(metrics.CC_METRIC, value)  # type: ignore[union-attr]  # guarded by the if
```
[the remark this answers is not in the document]
egparedes: Thanks, it makes sense.

tests/next_tests/unit_tests/test_metrics.py
egparedes: I don't know if it's worth but anyway I changed it the way you suggested.

edopao (approved):
It looks good to me. I am not sure about the metric name, but I am also fine with the current `STENCIL_METRIC`.

src/gt4py/next/metrics.py
```diff
@@ -0,0 +1,205 @@
...
+from collections.abc import MutableSequence, Sequence
+
+import numpy as np
+
+from gt4py.eve import utils
+from gt4py.eve.extended_typing import Final, TypeVar
+
+
+# Common metric names
+CC_METRIC: Final = sys.intern("compiled_code")
```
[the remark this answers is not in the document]
egparedes: Renamed to `compute` as agreed offline.

src/gt4py/next/ffront/decorator.py
```diff
@@ -274,46 +274,47 @@ def _compiled_programs(self) -> compiled_program.CompiledProgramsPool:
...
-        if collect_metrics_level := config.COLLECT_METRICS_LEVEL:
-            assert metrics.active_metric_collection.get() is None
-            metrics.active_metric_collection.set(
-                program_metrics := metrics.program_metrics[self.__name__]
+        program_name = (
+            f"{self.__name__}[{getattr(self.backend, 'name', '<embedded>')}]"
+            if config.COLLECT_METRICS_LEVEL
+            else ""
+        )
+        with metrics.metrics_collection(program_name) as metrics_collection:
```
edopao, line 282: I am a bit lost. Here we check `config.COLLECT_METRICS_LEVEL`. In which case could `metrics_collection` be `None` given that `config.COLLECT_METRICS_LEVEL >= metrics.INFO`? Could we write as follows:

```
with metrics.metrics_collection(program_name) as metrics_collection:
  collect_metrics = config.COLLECT_METRICS_LEVEL >= metrics.INFO
  if collect_metrics:
    assert metrics_collection is not None
    start = time.time()
 ....
  if collect_metrics:
    metrics_collection[metrics.TOTAL_METRIC].add_sample(time.time() - start)
```

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -59,10 +59,17 @@ def decorated_program(
         converted_args = (convert_arg(arg) for arg in args)
         conn_args = extract_connectivity_args(offset_provider, device)
 
-        opt_kwargs: dict[str, Any] = {}
-        if collect_metrics := (config.COLLECT_METRICS_LEVEL >= metrics.PERFORMANCE):
-            exec_info: dict[str, float]
-            opt_kwargs["exec_info"] = (exec_info := {})
+        opt_kwargs: dict[str, Any]
+        metric_collection = metrics.get_active_metric_collection()
+        if metric_collection is not None and (
```
edopao: Similar comment here.

src/gt4py/next/ffront/decorator.py
egparedes: You're right, the semantics are exactly as you say. Checking if `metrics_collection is None` is just a small performance optimization to avoid checking the metric level in case the metrics are fully disabled.

PR #1981, opened 2025-04-15
build[next]: increase number of pytest workers

edopao:


edopao:
Not ready, the root issue is not clear.

PR #1982, opened 2025-04-15
ci: add nox utilities for conditional session execution instead of using github actions

egparedes:
Add extra utilities so now `nox` sessions can:
- read the supported python versions from the `.python-versions` file
- skip sessions conditionally based on the changes from a specific target commit
- run nox sessions using the PEP 723 noxfiles way supported since 2025.02.09
- use the standard `PYTEST_XDIST_AUTO_NUM_WORKERS` variable (used by `pytest-xdist`) instead of custom `NUM_PROCESSES` variable, to select the number of workers in a pytest session

The not-longer needed "fallback" GitHub Actions workflows have been also removed and the development documentation updated to reflect these changes. 

<details>
<summary>Show a summary per file (GitHub Copilot)</summary>

| File                                | Description                                                        |
| ----------------------------------- | ------------------------------------------------------------------ |
| pyproject.toml                      | Updated the nox dependency to version >=2025.02.09                  |
| noxfile_utils.py                    | Added new utility functions and decorators to manage nox sessions    |
| noxfile.py                          | Refactored sessions to use the new noxfile_utils and adjusted sys.path |
| docs/development/tools/ci-infrastructure.md | Updated documentation to reflect the removal of fallback workflows and integration of conditional nox sessions |
| dev-tasks.py                        | Modified version extraction logic for improved robustness            |
| .pre-commit-config.yaml             | Added a local hook for checking noxfile_utils functionality             |
| .github/workflows/*.yml             | Updated workflows to use the new script execution pattern and proper cloning depth for branch target PRs |
</details>


noxfile.py
```diff
@@ -10,12 +10,17 @@
...
 import types
-from collections.abc import Sequence
-from typing import Final, Literal, TypeAlias
+from collections.abc import Callable, Sequence
+from unittest import mock
+from typing import Any, Final, Literal, TypeAlias, TypeVar
 
 import nox
 
+with mock.patch("sys.path", [".", *sys.path]):
```
havogt, line 21: ?

noxfile.py
```diff
@@ -145,7 +171,18 @@ def test_examples(session: nox.Session) -> None:
         )
 
 
-@nox.session(python=["3.10", "3.11"], tags=["next"])
+@nox_utils.customize_session(
+    python=PYTHON_VERSIONS,
+    tags=["next"],
+    env_vars=["NUM_PROCESSES"],
+    ignore_paths=[  # Skip when only gt4py.cartesian or doc files have been updated
```
havogt, line 178: I guess you are experimenting with ignore_paths vs paths? Or is there any reason why you use ignore_paths for next and paths for cartesian?

noxfile_utils.py
```diff
@@ -0,0 +1,238 @@
+# GT4Py - GridTools Framework
```
havogt, line 1: Should we unittest this?

havogt (commented):
Looks good!

noxfile_utils.py
egparedes: Yes, it's in the to-do list

noxfile.py
egparedes: This is mimicking what we currently have on CI: `paths-ignore` for _cartesian_ and _next_, and `paths` for _eve_ and _storage_

noxfile.py
egparedes: Temporarily hack the `sys.path` to load the companion file `noxfile_utils.py`, since this is run in script mode without a venv. I added a comment.

noxfile_utils.py
egparedes: Done

noxfile.py
```diff
@@ -228,11 +296,11 @@ def test_storage(
 
     device_settings = DeviceTestSettings[device]
 
-    _install_session_venv(
+    nox_utils.install_session_venv(
         session, extras=["performance", "testing", *device_settings["extras"]], groups=["test"]
     )
 
-    num_processes = os.environ.get("NUM_PROCESSES", "auto")
+    num_processes = session.env.get("NUM_PROCESSES", "auto")
```
edopao, line 303: We changed `session.env.get("NUM_PROCESSES", "auto")` to `os.environ.get("NUM_PROCESSES", "auto")` because we saw problems in the CI pipelines, in this #1869
It seems like the problem is back with this change, see the CI log:
`288 workers [168 items]`

noxfile_utils.py
```diff
@@ -0,0 +1,305 @@
...
+        Only one of the `paths` / `ignore_paths` options can be used at the same time for now.
+
+    Example:
+        ```python
+        @customize_session(paths=["src/", "tests/"], name="test")
+        def run_tests(session):
+            session.run("pytest")
+        ```
+    """
+    if paths and ignore_paths:
```
halungge, line 88: Question: I think it is more clear to only have `paths` or `ignore_paths`, because it is easy to configure something inconsistent, but it would actually work even if you have both of them.

noxfile_utils.py
```diff
@@ -0,0 +1,305 @@
...
+ENV_VAR_VERBOSE: Final = f"{ENV_VAR_PREFIX}CI_NOX_VERBOSE"
+
+COMMIT_SPEC_FOR_CHANGES: Final = os.environ.get(ENV_VAR_RUN_ONLY_IF_CHANGED_FROM, "")
+VERBOSE: Final = os.environ.get(ENV_VAR_VERBOSE, "").lower() in [
+    "1",
+    "on",
+    "true",
+]
+
+_changed_files_from_commit: dict[str, list[str]] = {}
```
halungge, line 53: Another question: what is the live time of these caches? One pipeline? one call to nox?

.github/workflows/test-cartesian.yml
```diff
@@ -53,8 +49,11 @@ jobs:
...
         NUM_PROCESSES: auto
+        CI_NOX_PREFIX: GT4PY
+        GT4PY_CI_NOX_RUN_ONLY_IF_CHANGED_FROM: "origin/${{ github.base_ref }}"
+        GT4PY_CI_NOX_VERBOSE: 1
         # Force a specific compiler for setuptools/distutils
         CXX: g++
         CC: gcc
       shell: bash
-      run: uv run nox -s 'test_cartesian-${{ matrix.python-version }}(${{ matrix.codegen-factor }}, cpu)'
+      run: uv run --script noxfile.py -s 'test_cartesian-${{ matrix.python-version }}(${{ matrix.codegen-factor }}, cpu)'
```
halungge, line 59: Do you need to change to running nox as script in order to be able to customize the session? Or is this barely because of the sys.path issue?

noxfile_utils.py
egparedes: It is one call to nox, but since `_changed_files_from_commit` is only populated in the session run, then is effectively per session, which doesn't make any sense. Good catch, I 'll remove it.

noxfile.py
egparedes: Yes, you are right, reading from `session.env` at this point was always a bug from our side. I was fooled by the `include_outer_env` option of `session.run()` but it doesn't apply here because we are not running a command in the session, only reading the environment.

The solution I've implemented now is to get rid of our custom `NUM_PROCESSES` variable and use instead `PYTEST_XDIST_AUTO_NUM_WORKERS` as documented in pytest-xdist docs: https://pytest-xdist.readthedocs.io/en/stable/distribution.html#running-tests-across-multiple-cpus

Do you agree @edopao ?

halungge (approved):
LGTM

noxfile.py
edopao: I agree with this solution. However, I see that you also added `--dist loadgroup` and I don't understand the need for that. We do not use the `xdist_group` mark in our tests, right?

> Tests are grouped by the xdist_group mark. Groups are distributed to available workers as whole units. This guarantees that all tests with same xdist_group name run in the same worker.

```
@pytest.mark.xdist_group(name="group1")
def test1():
    pass
```

edopao: I see, this is a way to allow to serialize some tests. We maybe need this for the tests with `SDFGConvertible`. I see that one test is failing in https://github.com/GridTools/gt4py/actions/runs/14888031268/job/41812791488 but the log does not show the reason.

egparedes: Yes, this option allow to group together tests that should be executed by the same process. It was already used in `cartesian` and I adopted it now for all other sessions, since when groups are not used in the tests, it just falls back to the default behavior.

romanc: FWIW, we added the option in cartesian to mitigate an issue (test dependencies) in our test setup. See https://github.com/GridTools/gt4py/pull/1849 for more details.

If your tests don't have implicit test dependencies, the default way of scheduling should work out just fine. I'd stick with the default if you can.

ci/cscs-ci.yml
```diff
@@ -110,15 +110,20 @@ build_py310_image_aarch64:
...
+  - echo "CI_NOX_RUN_ONLY_IF_CHANGED_FROM=${CI_NOX_RUN_ONLY_IF_CHANGED_FROM}"
   - NOX_SESSION_ARGS="${VARIANT:+($VARIANT}${SUBVARIANT:+, $SUBVARIANT}${DETAIL:+, $DETAIL}${VARIANT:+)}"
-  - nox -e "test_$SUBPACKAGE-${PYVERSION:0:4}$NOX_SESSION_ARGS"
+  - nox -s "test_$SUBPACKAGE-${PYVERSION:0:4}$NOX_SESSION_ARGS"
   variables:
     CRAY_CUDA_MPS: 1
     SLURM_JOB_NUM_NODES: 1
-    SLURM_TIMELIMIT: 20
-    NUM_PROCESSES: auto
+    SLURM_TIMELIMIT: 25
```
edopao, line 123: Please revert the `SLURM_TIMELIMIT`.

edopao (commented):
LGTM

noxfile.py
egparedes: > If your tests don't have implicit test dependencies, the default way of scheduling should work out just fine. I'd stick with the default if you can.

The `loadgroup` setting uses the default `load` policy for unmarked tests, which I think it is exactly the behavior we want:
```
Tests without the xdist_group mark are distributed normally as in the --dist=load mode.
```

romanc: Ah right - I forgot about that part. All good then. I agree, that is a good default behavior.

PR #1983, opened 2025-04-15
feat[next][dace]: cleanup fastcall

edopao:
Some cleanup in the dace fast_call API, although the performance gain still needs to be assessed.
- remove bad import of private method `_array_interface_ptr` from `dace.codegen.compiled_sdfg`
- remove several runtime calls to `isinstance()`
- use fast_call API also when the SDFG is called with different arrays; this is enabled by generating the shape and stride symbols for all arrays


src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -48,43 +47,42 @@ def decorated_program(
...
-                        break
+                    assert field_utils.verify_device_field_type(arg, device)
+                    last_call_args[i].value = (
+                        ndarray.__cuda_array_interface__["data"][0]
+                        if on_gpu
+                        else ndarray.__array_interface__["data"][0]
+                    )
+                    # backfill dictionary of arguments with field shape and stride symbols
+                    kwargs.update(sdfg_callable.get_field_domain_symbols(arg_name, arg.domain))
+                    kwargs.update(sdfg_callable.get_field_stride_symbols(arg_desc, ndarray))
```
philip-paul-mueller, line 72: Are you sure that it is save to do that, for sure it is not if we have persistent memory, then it is for sure dangerous.
It could also kill us if we do some jitting (but I guess that might be handled some levels above).

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -48,43 +47,42 @@ def decorated_program(
             flat_args = (*flat_args, *flat_size_args)
 
         if sdfg_program._lastargs:
```
philip-paul-mueller, line 49:
```suggestion
        if sdfg_program._lastargs:
        # Initialization was done by the `CompiledSDFG` object.
```

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -48,43 +47,42 @@ def decorated_program(
...
-                            arg_name
-                        ), f"argument '{arg_name}' not found."
+                    actype = arg_desc.dtype.as_ctypes()
+                    last_call_args[i] = actype(arg)
 
-            if use_fast_call:
-                return inp.fast_call()
+            return inp.fast_call()
 
         sdfg_args = sdfg_callable.get_sdfg_args(
```
philip-paul-mueller, line 81: I would actually put this inside an `else` branch and add some note saying something like: "First call, the sdfg is not intitalized, so forward the call to `CompiledSDFG` to proper initilize it, later calls will be handled through the `fast_call()` API.

Furthermore, I would also change the order of the branches, i.e. handling this (the first/genesis call) case first.
It is more logical, because it then looks as:
```python
if never_called:
    do_proper_initialization()
else:
    # Already called, we will manipulate the internal `_lastargs` object
    magic()
```

philip-paul-mueller (commented):
It generally looks good I have some suggestions.
However, my main point of concern is, that this new code essentially skips some tests.
I think we should add a note that clearly says why it is okay to do it in that way.

src/gt4py/next/program_processors/runners/dace/sdfg_callable.py
```diff
@@ -59,40 +79,24 @@ def _get_args(sdfg: dace.SDFG, args: Sequence[Any]) -> dict[str, Any]:
     return sdfg_arguments | range_symbols
 
 
-def _get_shape_args(
-    arrays: Mapping[str, dace.data.Array], args: Mapping[str, core_defs.NDArrayObject]
-) -> dict[str, int]:
+def _get_shape_args(sdfg: dace.SDFG, args: Mapping[str, core_defs.NDArrayObject]) -> dict[str, int]:
```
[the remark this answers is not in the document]
edopao: Good idea.

src/gt4py/next/program_processors/runners/dace/sdfg_callable.py
```diff
@@ -59,40 +79,24 @@ def _get_args(sdfg: dace.SDFG, args: Sequence[Any]) -> dict[str, Any]:
...
+                assert isinstance(sym, dace.symbol)
                 shape_args[sym.name] = size
-            elif sym != size:
-                raise ValueError(
-                    f"Expected shape {arrays[name].shape} for arg {name}, got {value.shape}."
-                )
     return shape_args
 
 
 def _get_stride_args(
```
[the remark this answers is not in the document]
edopao: I could finally remove these functions, it looks cleaner now.

edopao:
Adding @havogt and @egparedes to address two specific questions, thank you.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -46,19 +46,53 @@ def decorated_program(
...
+                *flat_args,
+                filter_args=False,
+                on_gpu=on_gpu,
+            )
+            with dace.config.temporary_config():
+                dace.config.Config.set("compiler", "allow_view_arguments", value=True)
+                return inp(**this_call_args)
+
+        else:
+            # Initialization of `_lastargs` was done by the `CompiledSDFG` object.
```
philip-paul-mueller, line 65:
```suggestion
            # Initialization of `_lastargs` was done by the `CompiledSDFG` object,
            #  so we just update it.
```

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -46,19 +46,53 @@ def decorated_program(
...
+                )
+            # The loop below will traverse the `sdfg_arglist`, whose order reflects
+            # the DaCe calling convention: first the array arguments, then the scalar
+            # arguments. We exploit this knowledge to update `this_call_args` with
+            # the shape and stride symbols of array arguments, when they are visited.
+            # The corresponding scalar arguments are visited later in `sdfg_arglist`.
             for i, (arg_name, arg_desc) in enumerate(inp.sdfg_arglist):
-                arg = kwargs.get(arg_name, None)
+                arg = this_call_args.get(arg_name, None)
                 if arg is None:
```
philip-paul-mueller, line 79:
```suggestion
                if arg is None and not assume_immutable_offset_provider:
```

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -46,19 +46,53 @@ def decorated_program(
...
-                    # connectivities are supposed not to change, and can therefore be omitted
-                    assert gtx_dace_utils.is_connectivity_identifier(
-                        arg_name, common.offset_provider_to_type(offset_provider)
-                    ) or gtx_dace_utils.is_connectivity_symbol(
-                        arg_name, common.offset_provider_to_type(offset_provider)
+                    # In case of immutable offset provider, the connectivity arrays
+                    # and the associate shape and stride symbols can be omitted.
+                    assert assume_immutable_offset_provider and (
```
philip-paul-mueller, line 82:
```suggestion
                    assert (
```

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -46,19 +46,53 @@ def decorated_program(
...
+                            arg_name, common.offset_provider_to_type(offset_provider)
+                        )
                     ), f"argument '{arg_name}' not found."
-                elif (ndarray := getattr(arg, "ndarray", None)) is not None:
+                elif (ndarray := getattr(arg, "ndarray", None)) is None:
+                    assert isinstance(arg_desc, dace.data.Scalar)
+                    assert isinstance(last_call_args[i], ctypes._SimpleCData)
+                    actype = arg_desc.dtype.as_ctypes()
+                    last_call_args[i] = actype(arg)
+                else:
```
philip-paul-mueller, line 95: I would put that branch before the scalar branch, because it is visited _first_, so the order should reflect that.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -46,19 +46,53 @@ def decorated_program(
...
+                    assert assume_immutable_offset_provider and (
+                        gtx_dace_utils.is_connectivity_identifier(
+                            arg_name, common.offset_provider_to_type(offset_provider)
+                        )
+                        or gtx_dace_utils.is_connectivity_symbol(
+                            arg_name, common.offset_provider_to_type(offset_provider)
+                        )
                     ), f"argument '{arg_name}' not found."
-                elif (ndarray := getattr(arg, "ndarray", None)) is not None:
+                elif (ndarray := getattr(arg, "ndarray", None)) is None:
```
philip-paul-mueller, line 90: Add something like this:
As outlined above, because of how `sdfg_arglist` is constructed, all arrays will be processed first, thus all scalars that are associated with the shape or strides have been updated in `this_call_args` and it is thus safe to use them.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -67,27 +101,19 @@ def decorated_program(
...
-            dace.config.Config.set("compiler", "allow_view_arguments", value=True)
-            return inp(**sdfg_args)
+                    # When we find an array we update the `this_call_args` map
+                    # with the shape and stride symbols that are associated to it.
+                    # Note that `inp.sdfg_arglist` was constructed from an ordered
+                    # dictionary whose order reflects the DaCe calling convention:
+                    # first the array arguments, then the scalar arguments.
+                    # Thus, when we enter the branch above for scalar arguments
+                    # we know that all arrays have been processed and all their
+                    # associated symbols have been added to `this_call_args`.
```
philip-paul-mueller, line 111:
```suggestion
                    # Because of the order `sdfg_arglist` is constructed (see above) the
                    #  arrays are found first. We will now update the values that are
                    #  associated to their shape and strides.  
```

Maybe it is possible to make it smoother.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -48,43 +47,42 @@ def decorated_program(
...
                     assert isinstance(last_call_args[i], ctypes.c_void_p)
-                    if last_call_args[i].value != data_ptr:
-                        use_fast_call = False
-                        break
+                    assert field_utils.verify_device_field_type(arg, device)
+                    last_call_args[i].value = (
+                        ndarray.__cuda_array_interface__["data"][0]
+                        if on_gpu
+                        else ndarray.__array_interface__["data"][0]
+                    )
```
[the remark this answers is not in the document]
philip-paul-mueller: I am not thinking that it is "more" generic than what we need, we need to read from AMD.
But I agree we should wait for @egparedes feedback.

philip-paul-mueller (commented):
I have only minimal suggestions I think that the main question is how re replace `_array_interface_ptr()`.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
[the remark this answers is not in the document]
egparedes: I guess here you're trying to get the memory pointer of the device buffer, right? The most generic way to do this would be by calling `__dlpack__()`, but decoding the DLPack Tensor Capsule (https://dmlc.github.io/dlpack/latest/python_spec.html) in pure python is a bit cumbersome and requires using `cytpes` (check any of the examples using `ctypes` in the previous link).

A much simpler way to get the pointer in a way that works for CPU/CUDA/ROCm is using the `byte_bounds()` function from numpy (and cupy), which returns pointers to the start and end position of the array (https://devdocs.io/numpy~2.0/reference/generated/numpy.lib.array_utils.byte_bounds). We are already using this in the `storage` subpackage and it works perfectly. The only annoying part is that since the function is not part of the Data Array API, it has been moved from their original location in the root numpy package to the `.lib.array_utils` subpackage in 2.0, so if has to be read from different places depending on the numpy version installed. Inside `gt4py.storage` we deal with this issue with an ugly (temporary) workaround here: https://github.com/GridTools/gt4py/blob/748f33e2ab6956e5bd97ad6554ddffd229f27b98/src/gt4py/storage/allocators.py#L308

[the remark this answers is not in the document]
egparedes: But can't you directly use `byte_bounds` from cupy?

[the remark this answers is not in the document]
edopao: You are right, I have applied this change. Now I am using `byte_bounds` from cupy and it works, thanks.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -77,42 +78,52 @@ def decorated_program(
...
+                    # As outlined above, because of how `sdfg_arglist` is constructed,
+                    # all arrays will be visited first, thus all scalars that are
+                    # associated with the shape and stride symbols have been updated
+                    # in `this_call_args` and it is thus safe to use them.
+                    assert isinstance(arg_desc, dace.data.Scalar)
+                    assert isinstance(last_call_args[i], ctypes._SimpleCData)
+                    actype = arg_desc.dtype.as_ctypes()
+                    last_call_args[i] = actype(arg)
+
             else:
```
philip-paul-mueller, line 127: Why is this `else` here, it is associated to the `for` isn't it?

philip-paul-mueller (approved):
LGTM.

See comment about `else`.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -42,54 +41,90 @@ def decorated_program(
...
+                        raise ValueError("argument '{arg_name}' not found.")
+
+                elif (ndarray := getattr(arg, "ndarray", None)) is not None:
+                    assert isinstance(arg_desc, dace.data.Array)
                     assert isinstance(last_call_args[i], ctypes.c_void_p)
-                    if last_call_args[i].value != data_ptr:
-                        use_fast_call = False
-                        break
+                    assert field_utils.verify_device_field_type(arg, device)
+                    last_call_args[i].value = (  # different location of `byte_bounds` in cupy/numpy
```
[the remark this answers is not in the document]
edopao: I've found someone (probably you and @kotsaloscv) already implemented this feature, I can just use it:
```
src/gt4py/next/embedded/nd_array_field.py

        # Extension of NdArrayField adding SDFGConvertible support in GT4Py Programs
        def _dace_data_ptr(self) -> int:
            array_ns = self.array_ns
            array_byte_bounds = (  # TODO(egparedes): make this part of some Array namespace protocol
                array_ns.byte_bounds
                if hasattr(array_ns, "byte_bounds")
                else array_ns.lib.array_utils.byte_bounds
            )
            return array_byte_bounds(self.ndarray)[0]

```

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
edopao: Yes, it is. I would prefer to remove it, do you agree?

PR #1985, opened 2025-04-15
fix[next]: Temporary extraction for fields defined on different domains

tehrengruber:
Splits tuple values that are defined on different domains into separate SetAts.

During temporary extraction we might extract a SetAt that contains a tuple with elements defined on different domains (e.g. Vertex and Edge). SetAts are required to have a single domain, therefore these SetAts are split into one SetAt per domain at the end of temporary extraction.

Note: it seems many of the cases in icon4py are actually cases where the domain is semantically equal, but the domain expression is different (alla `1+1` != `2`). This can result in many unnecessary kernel calls.


havogt:
>it seems many of the cases in icon4py are actually cases where the domain is semantically equal, but the domain expression is different (alla 1+1 != 2). This can result in many unnecessary kernel calls.

Is this horizontal or vertical? With the variant precompile the domain expressions should get simpler, also I would like to do another round of removing superfluous concat_wheres on the horizontal. So the problem might disappear.

tehrengruber:
The case I saw was horizontal only.

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -19,11 +19,73 @@
...
+    # can be reproduced in icon4py `dycore_tests/test_solve_nonhydro.py::test_nonhydro_predictor_step`
+    # compute_theta_rho_face_values_and_pressure_gradient_and_update_vn
+    assert isinstance(stmt.expr.annex.domain, tuple)
+    assert domain in stmt.expr.annex.domain
+    for i, (el, el_domain) in enumerate(zip(stmt.expr.args, stmt.expr.annex.domain)):
+        assert not isinstance(el_domain, tuple), "Nested tuples not supported."
+        assert isinstance(el_domain, SymbolicDomain)
+        if el_domain == domain:
+            new_targets.append(im.tuple_get(i, stmt.target))
+            new_els.append(el)
```
egparedes, line 44: Just two suggestions: 
- I would move the assertions about the inputs to the beginning of the function
- I would move the asserts per element to the call site outside this function, since I think it's more clear and also avoids asserting multiple times the same conditions.
```python 
            assert not isinstance(el_domain, tuple), "Nested tuples not supported."
            assert isinstance(el_domain, SymbolicDomain)
```

```suggestion
    assert cpm.is_call_to(stmt.expr, "make_tuple")
    assert isinstance(stmt.expr.annex.domain, tuple)
    assert domain in stmt.expr.annex.domain

    new_targets = []
    new_els = []
    # we cannot use `el.annex.domain` since while it might be semantically equivalent, it does
    # not necessarily be equal.
    # can be reproduced in icon4py `dycore_tests/test_solve_nonhydro.py::test_nonhydro_predictor_step`
    # compute_theta_rho_face_values_and_pressure_gradient_and_update_vn
    for i, (el, el_domain) in enumerate(zip(stmt.expr.args, stmt.expr.annex.domain)):
        if el_domain == domain:
            new_targets.append(im.tuple_get(i, stmt.target))
            new_els.append(el)
```

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -19,11 +19,73 @@
...
+                )
+            )
+            if len(distinct_domains) > 1:
+                for domain in distinct_domains:
+                    new_stmts.append(_filter_setat_by_domain(stmt, domain))
+            else:
+                new_stmts.append(
+                    itir.SetAt(
+                        expr=stmt.expr,
+                        domain=next(iter(distinct_domains)).as_expr(),
```
egparedes, line 71:
```suggestion
                        domain=distinct_domains[0].as_expr(),
```

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -106,17 +150,47 @@ def _transform_by_pattern(
...
-                declarations.append(itir.Temporary(id=tmp_name, domain=domain_expr, dtype=dtype))  # noqa: B023 # function only used inside loop
+            def allocate_temporary(
+                tmp_name: str, domain: domain_utils.SymbolicDomain, dtype: ts.ScalarType
+            ):
+                declarations.append(
+                    itir.Temporary(id=tmp_name, domain=domain.as_expr(), dtype=dtype)
+                )  # function only used inside loop
 
-            next_utils.tree_map(allocate_temporary)(tmp_names, tmp_dtypes)
+            next_utils.tree_map(allocate_temporary)(tmp_names, tmp_domains, tmp_dtypes)
```
egparedes, line 181: Not important, but this can be simplified by deleting the  `allocate_temporary()` function and instead using the constructor directly in `tree_map` which I also find slightly more readabale.
```suggestion
        declarations.extend(
            next_utils.tree_map(itir.Temporary)(
                tmp_names, tuple(d.as_expr() for d in tmp_domains), tmp_dtypes
            )
        )
```

havogt: there is a flatten missing and the eve.Node only accepts kwargs, I propose
```
            declarations.extend(
                itir.Temporary(id=tmp_name, domain=domain.as_expr(), dtype=dtype)
                for tmp_name, domain, dtype in zip(
                    next_utils.flatten_nested_tuple(tmp_names),
                    next_utils.flatten_nested_tuple(tmp_domains),
                    next_utils.flatten_nested_tuple(tmp_dtypes),
                )
            )
```

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -106,17 +150,47 @@ def _transform_by_pattern(
...
+                assert isinstance(domain, domain_utils.SymbolicDomain)
+                return domain
+
+            tmp_domains: (
+                domain_utils.SymbolicDomain | tuple[domain_utils.SymbolicDomain | tuple, ...]
+            ) = type_info.apply_to_primitive_constituents(
+                get_domain,
+                tmp_expr.type,
+                with_path_arg=True,
+                tuple_constructor=lambda *elements: tuple(elements),
```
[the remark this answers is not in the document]
havogt: I deleted the whole thing because to my understanding it just reproduces `tmp_expr.annex.domain`

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -127,7 +199,13 @@ def allocate_temporary(tmp_name: str, dtype: ts.ScalarType):
             # TODO(tehrengruber): _transform_stmt not needed if deepest_expr_first=True
             tmp_stmts.extend(
                 _transform_stmt(
-                    itir.SetAt(target=target_expr, domain=domain_expr, expr=tmp_expr),
+                    itir.SetAt(
+                        target=target_expr,
+                        # The domain is populated later in `_populate_and_homogenize_domains`
+                        # at this point the `SetAt` contains possibly expressions no different domains.
```
havogt, line 205:
```suggestion
                        # at this point the `SetAt` contains possibly expressions on different domains.
```

src/gt4py/next/iterator/ir_utils/ir_makers.py
```diff
@@ -58,7 +60,11 @@ def ref(
     if isinstance(ref_or_name, itir.SymRef):
         assert not type_
         return ref_or_name
-    return itir.SymRef(id=ref_or_name, type=ensure_type(type_))
+    ref = itir.SymRef(id=ref_or_name, type=ensure_type(type_))
+    if annex is not None:
+        for key, value in annex.items():
+            setattr(ref.annex, key, value)
+    return ref
```
egparedes, line 67: What is the intended behavior if we pass a `SymRef` together with a non-empty annex? Shouldn't a new `SymRef` be created with the new annex?

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -19,11 +20,94 @@
...
+    `SetAt`s have a single domain therefore the target domains have to be the same.
+    We support nested tuples by flattening them and recombining them into non-nested tuples.
+    """
+    new_stmts: list[itir.Stmt] = []
+    for stmt in stmts:
+        if isinstance(stmt, itir.SetAt):
+            assert hasattr(stmt.expr.annex, "domain")
+            distinct_domains: list[domain_utils.SymbolicDomain] = list(
+                # ordered set for reproducibility
+                dict.fromkeys(next_utils.flatten_nested_tuple(stmt.expr.annex.domain))
```
egparedes, line 85: I'm a bit confused here: why is guaranteed that the list is always sorted in the same way? Doesn't the `dict` created with `fromkeys()` depend on the the iteration order of whatever is in the flattened version of the `annex.domain`? Are we sure then that the contents of `annex.domain` are always sorted consistently?

egparedes (commented):
A couple of questions.

src/gt4py/next/iterator/transforms/global_tmps.py
havogt: It's only important the for the same program the order is always the same (for test reproducibility).

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -103,20 +169,58 @@ def _transform_by_pattern(
...
-            def allocate_temporary(tmp_name: str, dtype: ts.ScalarType):
-                declarations.append(itir.Temporary(id=tmp_name, domain=domain_expr, dtype=dtype))  # noqa: B023 # function only used inside loop
+            tmp_domains: SymbolicDomain | tuple[SymbolicDomain | tuple, ...] = tmp_expr.annex.domain
+
+            if cpm.is_applied_as_fieldop(tmp_expr):
+                # In this case all tuple elements have the same size (or will be `NEVER`).
+                # Create the tuple structure with that domain.
+                domains = list(
+                    set(next_utils.flatten_nested_tuple((tmp_domains,)))
+                    - set([infer_domain.DomainAccessDescriptor.NEVER])  # type: ignore[list-item] # type should always be `SymbolicDomain`
```
egparedes, line 182: There is a simpler way to create sets with specific elements:
```suggestion
                    - {infer_domain.DomainAccessDescriptor.NEVER}  # type: ignore[list-item] # type should always be `SymbolicDomain`
```

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -103,20 +169,58 @@ def _transform_by_pattern(
...
                 type_info.extract_dtype,
                 tmp_expr.type,
-                tuple_constructor=lambda *elements: tuple(elements),
+                tuple_constructor=lambda *elements: elements,
             )
 
-            # allocate temporary for all tuple elements
-            def allocate_temporary(tmp_name: str, dtype: ts.ScalarType):
-                declarations.append(itir.Temporary(id=tmp_name, domain=domain_expr, dtype=dtype))  # noqa: B023 # function only used inside loop
+            tmp_domains: SymbolicDomain | tuple[SymbolicDomain | tuple, ...] = tmp_expr.annex.domain
```
egparedes, line 175: It's a bit weird that something called `tmp_domains` in plural might contain a single value. I would expect it to always be a collection, even if ti were empty or only had a single element. 

```suggestion
            tmp_domains: SymbolicDomain | tuple[SymbolicDomain | tuple, ...] = tmp_expr.annex.domain
```

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -103,20 +169,58 @@ def _transform_by_pattern(
...
+
+            if cpm.is_applied_as_fieldop(tmp_expr):
+                # In this case all tuple elements have the same size (or will be `NEVER`).
+                # Create the tuple structure with that domain.
+                domains = list(
```
egparedes, line 180: Kind of related to the previous point, but if `tmp_domains` were always a collection, it wouldn't be needed to defined here the new `domains` symbol (which is also a bit confusing)

egparedes (approved):
Minor style comments, not super important, so I'll approve.

src/gt4py/next/iterator/transforms/global_tmps.py
havogt: ups

PR #1986, opened 2025-04-17
fix[next][dace]: set gpu block size to avoid dace warnings

edopao:
The CI log is polluted with dace warnings about missing gpu block size. With this PR we set a default gpu block size, which seems good from preliminary benchmark results. However, the log-term solution is to make it configurable through the gt4py toolchain, so that the block size can be optimized for each program.


philip-paul-mueller (approved):
LGTM for a temporary solution until the final way to set it is in place.

PR #1990, opened 2025-04-18
fix[next][dace]: Revisit dace config and compile arguments

edopao:
The purpose of this PR is to review and update the dace configuration for SDFG lowering and compilation.

The following changes are made with respect to the default/previous configuration:
- disable storing of the transformation history in SDFG
- remove `use-fast-math` from compiler args

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
@@ -74,22 +74,18 @@ class DaCeCompiler(
         with dace.config.temporary_config():
             dace.config.Config.set("compiler", "build_type", value=self.cmake_build_type.value)
             dace.config.Config.set("compiler", "use_cache", value=False)  # we use the gt4py cache
+
+            # dace dafault setting use fast-math in both cpu and gpu compilation, we disable it
+            dace.config.Config.set("compiler", "cpu", "args", value="-std=c++14 -fPIC -Wall -Wextra -O3 -march=native -Wno-unused-parameter -Wno-unused-label")
+            dace.config.Config.set("compiler", "cuda", "args", value="-Xcompiler -march=native -Xcompiler -Wno-unused-parameter")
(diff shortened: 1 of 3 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -72,25 +72,29 @@ def __call__(
...
+            dace.config.Config.set("compiler.use_cache", value=False)  # we use the gt4py cache
+
+            # dace dafault setting use fast-math in both cpu and gpu compilation, don't use it here
+            dace.config.Config.set(
+                "compiler.cpu.args",
+                value="-std=c++14 -fPIC -Wall -Wextra -O3 -march=native -Wno-unused-parameter -Wno-unused-label",
+            )
+            dace.config.Config.set(
+                "compiler.cuda.args",
+                value="-Xcompiler -march=native -Xcompiler -Wno-unused-parameter",
```
havogt, line 85: Wondering if this should get a `-O3` for host code as well.

edopao: I was wondering that too, I just copied the dace default here. I will add it.

edopao:
Is there any other compiler setting we use in GridTools and I should apply here?

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -88,6 +88,11 @@ def generate_sdfg(
         if not self.disable_itir_transforms:
             ir = itir_transforms.apply_fieldview_transforms(ir, offset_provider=offset_provider)
         offset_provider_type = common.offset_provider_to_type(offset_provider)
+
+        # do not store transformation history in SDFG
+        dace_config_store_history = dace.Config.get("store_history")
+        dace.Config.set("store_history", value=False)
```
havogt, line 94:
```suggestion
with dace.config.set_temporary("store_history", value=False):
```
Can you use the context handler here?

havogt (approved):
lgtm. See comment about the context handler.

PR #1992, opened 2025-04-18
feat[next][dace]: Add SDFG transformation for map splitting

edopao:
This PR introduces 2 new transformations: `gt_horizontal_map_fusion` and `gt_vertical_map_fusion`

The objective of these transformations is to identify maps with partially overlapping range and extract the range intersection as a separate map, in order to enable map fusion. The horizontal and vertical transformations differ in the pattern they look for:
- horizontal looks for any two maps with a common ancestor that can be fused by parallel fusion
- vertical looks for serial maps (i.e. two maps connected by some data container in between) in order to promote serial map fusion

However, several pieces are common, in particular the utilities for splitting the map range and copying a map scope.

--------
Co-authored by @iomaganaris @philip-paul-mueller


src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -0,0 +1,342 @@
...
+        )
+        assert splitted_range is not None
+
+        first_map_splitted_range, second_map_splitted_range = splitted_range
+
+        # make copies of the first map with splitted ranges
+        for i, r in enumerate(first_map_splitted_range):
+            map_fusion_utils.copy_map_graph_with_new_range(
+                sdfg, graph, first_map_entry, first_map_exit, r, str(i)
+            )
```
philip-paul-mueller: You could put that inside a `for` loop, it would reduce code doublication.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -0,0 +1,342 @@
...
+
+    return ret
+
+
+@dace_properties.make_properties
+class SplitMapRange(dace_transformation.SingleStateTransformation):
+    """
+    Identify overlapping range between maps, and split the range in order to
+    promote map fusion.
+    """
```
philip-paul-mueller: I would add a bit a more elaborated description of what it does.
Then for the two cases you can simply write something like "`SplitMapRange` for the case where we have horizontal Maps" or something.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -0,0 +1,342 @@
...
+        super().__init__(*args, **kwargs)
+
+    @classmethod
+    def expressions(cls) -> Any:
+        """Get the match expressions.
+
+        The function generates two match expressions. The first match describes
+        the case where the top map must be promoted, while the second case is
+        the second/lower map must be promoted.
+        """
```
philip-paul-mueller:
```suggestion
```

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def copy_map_graph(
+    sdfg: dace.SDFG,
+    graph: dace.SDFGState,
+    map_entry: dace_nodes.MapEntry,
+    map_exit: dace_nodes.MapExit,
+    suffix: Optional[str] = None,
+) -> tuple[dace_nodes.MapEntry, dace_nodes.MapExit]:
+    """Performs a full copy of the map graph.
```
philip-paul-mueller:
```suggestion
    """Performs a full copy of the dataflow inside the Map.
```

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
```diff
@@ -0,0 +1,276 @@
...
+
+    subgraph = graph.scope_subgraph(map_entry, include_entry=True, include_exit=True)
+    map_nodes = subgraph.nodes()
+    map_edges = subgraph.edges()
+
+    new_map_entry = None
+    new_map_exit = None
+
+    for node in map_nodes:
+        if isinstance(node, dace_nodes.AccessNode):
```
philip-paul-mueller, line 58: It can happen that you have global data inside the Map (okay it is unlikely, but DaCe's `auto_optimizer()` does this).
So you must actually check what the scope of that data is, for which you would have to look at its lifetime.
I think that in a first iteration it would be enough to look at its `.lifetime` attribute and look if it is `Scope` or it refers to global data.
But I think for our cases, checking for global data is enough.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
```diff
@@ -0,0 +1,276 @@
...
+        new_nodes[node] = node_
+        if node == map_entry:
+            new_map_entry = node_
+        elif node == map_exit:
+            new_map_exit = node_
+
+    # we have to ensure that the exit node references the new map node
+    assert new_map_entry is not None
+    assert new_map_exit is not None
+    new_map_exit.map = new_map_entry.map
```
philip-paul-mueller, line 103: You set here the `map` property of the ExitMap to the one of the EntryMap, but where do you set the map attribute of the EntryMap?

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
```diff
@@ -0,0 +1,276 @@
...
+) -> None:
+    """Helper function to delete a map from a state graph.
+
+    Args:
+        graph: The SDFG state containing the map graph.
+        map_entry: The entry node of the map graph.
+        map_exit: The exit node of the map graph.
+    """
+    for node in graph.scope_subgraph(map_entry, include_entry=True, include_exit=True).nodes():
+        graph.remove_node(node)
```
philip-paul-mueller:
```suggestion
    graph.remove_nodes_from(list(graph.scope_subgraph(map_entry, include_entry=True, include_exit=True).nodes()))
```

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
```diff
@@ -0,0 +1,276 @@
...
+        [second_map_dict[param] for param in sorted(second_map_params)]
+    )
+
+    try:
+        if first_map_sorted_range.intersects(second_map_sorted_range) == False:  # noqa: E712 [true-false-comparison]  # SymPy fuzzy bools.
+            # in case of disjoint ranges, we cannot find an overlapping range
+            return None
+    except TypeError:
+        # cannot determine truth value of Relational
+        return None
```
philip-paul-mueller: You should add a note what this means.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
```diff
@@ -0,0 +1,276 @@
...
+                (overlap_range_start, overlap_range_stop, step),
+            ]
+            if first_map_range[0] < overlap_range_start:
+                first_map_splitted_dict[param].append(
+                    (first_map_range[0], overlap_range_start - step, step)
+                )
+            if overlap_range_stop < first_map_range[1]:
+                first_map_splitted_dict[param].append(
+                    (overlap_range_stop + step, first_map_range[1], step)
+                )
```
philip-paul-mueller: I had to stare at that thing for several minutes.
I would add an explanation to it.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_horizontal_map_fusion.py
```diff
@@ -0,0 +1,284 @@
...
+    state.add_edge(task4, "__out0", small_map_exit, "IN_tmp4", dace.Memlet("tmp4[0]"))
+    state.add_edge(small_map_exit, "OUT_tmp4", tmp4, None, dace.Memlet("tmp4[0]"))
+    state.add_edge(tmp4, None, mexit, "IN_out1", dace.Memlet("tmp4[0]"))
+
+    state.add_edge(a, None, mentry, "IN_a", sdfg.make_array_memlet("a"))
+    state.add_edge(b, None, mentry, "IN_b", sdfg.make_array_memlet("b"))
+    state.add_edge(mexit, "OUT_out1", out1, None, sdfg.make_array_memlet("out1"))
+    for name in ["a", "b"]:
+        mentry.add_in_connector("IN_" + name)
+        mentry.add_out_connector("OUT_" + name)
```
philip-paul-mueller:
```suggestion
    for name in ["a", "b"]:
        mentry.add_scope_connectors(name)
```

philip-paul-mueller (commented):
It looks good, there are some small improvements, but they are essentially details.
Given the fact that we use this code already for a long time.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
iomaganaris: Good point. Instead of a `for` loop I think it makes more sense to wrap the duplicate code inside a function and call it twice 👍

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
iomaganaris: It seems to me that the suggestion here is exactly the same. Do you have some specific suggestions regarding the docstring?

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
iomaganaris: Do you think that this will be a problem because of the restriction that an `AccessNode` with `Global` scope should only exist once in a state? If yes I'm still skeptical about this restriction since currently there are states generated with more than one `Global` `AccessNode` without any issue. I guess it would only be an issue if we we wrote inside the map this `Global` `AccessNode` and the new map had the same range as the previous but I don't think this case is ever reached.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
iomaganaris: This should be set by
```
new_map_entry = node_
```
in line 93. I think we just make sure here that the map between the entry and the exit are the same.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
philip-paul-mueller: The suggestion is to remove it, because there is only one pattern that is matched, so it does not make sense in my view to add them.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
philip-paul-mueller: The issue is when you clone the data descriptor you also clone its `not transient` data, so you would create an additional array parameter that has to be supplied from the outside.
However, I agree that it is very unlikely.
But I would add an `assert data_desc.transient` at line 62.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
philip-paul-mueller: You are right you also copy the `.map` attribute.

philip-paul-mueller (approved):
LGTM.
There are two small things that you should look at, but they do not justify another round.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
iomaganaris: Okay, this makes sense 👍

PR #1993, opened 2025-04-18
fix[next]: Remove unnecessary visit in prune broadcast

tehrengruber:


src/gt4py/next/iterator/transforms/prune_broadcast.py
```diff
@@ -28,7 +28,7 @@ def visit_FunCall(self, node: itir.FunCall) -> itir.FunCall:
         node = self.generic_visit(node)
 
         if cpm.is_call_to(node, "broadcast"):
-            expr = self.visit(node.args[0])
+            expr = node.args[0]
```
havogt, line 31: Why is it not necessary? What about nested broadcasts?

tehrengruber: All broadcasts in `node.args` are already removed by the `generic_visit` above. This just create a `O(broadcast_nesting_depth^2)` runtime. In general we only visit again when a transformation opens up new optimization potential in child nodes and even those cases are usually only there because information flow between transformations on different levels of tree is very difficult right now.

havogt: Ah, I missed the generic_visit.

havogt (approved):

PR #1994, opened 2025-04-22
refactor[next]: Improved typing for common_pattern_matcher

havogt:


src/gt4py/next/iterator/ir_utils/common_pattern_matcher.py
```diff
@@ -7,13 +7,52 @@
...
+    False
+    >>> is_call_to(node, ("plus", "minus"))
+    True
+    """
+    assert not isinstance(fun, itir.Node)  # to avoid accidentally passing the fun as first argument
+    if isinstance(fun, (list, tuple, set, Iterable)) and not isinstance(fun, str):
+        return any((is_call_to(node, f) for f in fun))
+    return (
+        isinstance(node, itir.FunCall) and isinstance(node.fun, itir.SymRef) and node.fun.id == fun
+    )
```
egparedes, line 49: Shouldn't be simpler to change the order of the conditions here?
```suggestion
    if isinstance(fun, str):
        return (
            isinstance(node, itir.FunCall) and isinstance(node.fun, itir.SymRef) and node.fun.id == fun
        )
    else: 
        return any((is_call_to(node, f) for f in fun))
```

src/gt4py/next/iterator/ir_utils/common_pattern_matcher.py
```diff
@@ -7,13 +7,52 @@
...
-def is_applied_lift(arg: itir.Node) -> TypeGuard[itir.FunCall]:
+_Fun = TypeVar("_Fun", bound=itir.Expr)
+
+
+class _IsCall(itir.FunCall, Generic[_Fun]):
+    fun: _Fun
+    args: List[itir.Expr]
+
+
+_IsCallToPattern: TypeAlias = _IsCall[itir.SymRef]
```
egparedes, line 24: It's not a blocker but I think these names can be improved. What about `_IsCall` -> `FunCallTo` , `_IsCallToPattern` -> `FunCallToSymRef`, and `_IsFunCallToFunCallToRef` -> `FunCallToFunCallToSymRef` ?

src/gt4py/next/iterator/ir_utils/common_pattern_matcher.py
```diff
@@ -53,46 +92,24 @@ def is_applied_shift(arg: itir.Node) -> TypeGuard[itir.FunCall]:
...
+def is_applied_as_fieldop(arg: itir.Node) -> TypeGuard[_IsFunCallToFunCallToRef]:
     """Match expressions of the form `as_fieldop(stencil)(*args)`."""
     return isinstance(arg, itir.FunCall) and is_call_to(arg.fun, "as_fieldop")
 
 
-def is_let(node: itir.Node) -> TypeGuard[itir.FunCall]:
-    """Match expression of the form `(λ(...) → ...)(...)`."""
-    return isinstance(node, itir.FunCall) and isinstance(node.fun, itir.Lambda)
-
+_LetPattern: TypeAlias = _IsCall[itir.Lambda]
```
egparedes, line 100: `FunCallToLambda` ?

src/gt4py/next/iterator/transforms/power_unrolling.py
```diff
@@ -53,18 +53,19 @@ def visit_FunCall(self, node: ir.FunCall):
...
 
-                    ret = im.multiplies_(ret, f"power_{2 ** pow_cur}")
+                    ret = im.multiplies_(ret, f"power_{2**pow_cur}")
 
                 # Nest target expression to avoid multiple redundant evaluations
                 for i in range(pow_max, 0, -1):
                     ret = im.let(
-                        f"power_{2 ** i}", im.multiplies_(f"power_{2**(i-1)}", f"power_{2**(i-1)}")
+                        f"power_{2**i}",
+                        im.multiplies_(f"power_{2 ** (i - 1)}", f"power_{2 ** (i - 1)}"),
```
egparedes, line 68: It looks like an unnecessary format change just because there is trailing comma...
after `im.multiplies_`

egparedes (commented):
It looks good. Only minor style comments.

src/gt4py/next/iterator/transforms/power_unrolling.py
```diff
@@ -64,8 +64,7 @@ def visit_FunCall(self, node: ir.FunCall) -> ir.Expr:
                 # Nest target expression to avoid multiple redundant evaluations
                 for i in range(pow_max, 0, -1):
                     ret = im.let(
-                        f"power_{2**i}",
-                        im.multiplies_(f"power_{2 ** (i - 1)}", f"power_{2 ** (i - 1)}"),
+                        f"power_{2 ** i}", im.multiplies_(f"power_{2**(i-1)}", f"power_{2**(i-1)}")
```
egparedes, line 67: Ah, now I understand the format change. The difference was not in the trailing comma, which seems to be already present in the previous version but in the space around the `-` in the exponent of the `power_` expression. Feel free to put back the trailing comma if you prefer.

src/gt4py/next/iterator/ir_utils/common_pattern_matcher.py
```diff
@@ -16,15 +16,15 @@
...
 
 
-class _IsCall(itir.FunCall, Generic[_Fun]):
+class FunCallTo(itir.FunCall, Generic[_Fun]):
     fun: _Fun
     args: List[itir.Expr]
 
 
-_IsCallToPattern: TypeAlias = _IsCall[itir.SymRef]
+_FunCallToSymRef: TypeAlias = FunCallTo[itir.SymRef]
```
egparedes, line 24: I would make all these symbols public, or private if you prefer, but I don't understand why would `FunCallTo` be public while `_FunCallToSymRef` and similar are private...

egparedes (approved):
LGTM. I still have two comments about style but they are so minor that don't really need another round of review, so I'm approving the PR now and you can decide what to do.

src/gt4py/next/iterator/transforms/power_unrolling.py
havogt: I left the original one...

PR #1995, opened 2025-04-22
ci: set `CC` and `CXX` to force compiler used in `setuptools` / `distutils`

romanc:
## Description

`gt4py.cartesian` uses `setuptools` / `distutils` to compiler sources. For some unknown reason, the default compiler changed in the GHA runners over the last weekend. Daily CI started to fail as a consequence. This PR forces the GNU compiler suite by setting `CC` and `CXX` environment variable to `gcc` and `g++` respectively.

While `setuptools` version 75 is happy with just `CXX`, the oldest supported `setuptools` version (70) also requires setting `CC` in order not to mix the (apparent new default) `clang` compiler with `g++` as `C` and `C++` compilers respectively. From reading

- https://github.com/pypa/setuptools/issues/1732
- https://github.com/pypa/distutils/pull/228 (scroll all the way down to get to comments as recent as 2024)

setting compilers is a bit fragile and subject to active discussions/development.

Tested by also running the "Daily CI" job temporarily on this PR.

## Requirements

- [ ] All fixes and/or new features come with corresponding tests. N/A

### DIFF
--- a/.github/workflows/daily-ci.yml
+++ b/.github/workflows/daily-ci.yml
@@ -7,9 +7,9 @@ on:
 
   ## COMMENTED OUT: only for testing CI action changes.
   ## It only works for PRs to `main` branch from branches in the upstream gt4py repo.
-  pull_request:
-    branches:
-    - main
+  # pull_request:
+  #   branches:
+  #   - main
   ## END
 
 jobs:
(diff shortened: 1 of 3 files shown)


havogt (approved):
Thanks for investigating!

romanc:
> Thanks for investigating!

Always happy to help :)

PR #1996, opened 2025-04-23
ci: Avoid pulling full git history

romanc:
## Description

While looking at GHA workflows for PRhttps://github.com/GridTools/gt4py/pull/1995, I noticed that we are pulling the full git history in some workflows. A comment mentions this is to "get the correct version from tags". In case of code quality I'm sure we don't need this. Others also seem to pass. The only one that I can't (or don't know how to test) is the release workflow.

To me, this looks like a leftover from an intermediate strategy in PR https://github.com/GridTools/gt4py/pull/1832. If so, I suggest, we clean this up. For me, it's more about the (possibly) outdated comment than about pulling the full history vs. sparse checkout. It's not like a sparse checkout would save a huge amount of time or something like that (kudos by the way for keeping the repo in such good shape that this doesn't make a difference).

This is not urgent at all. Feel free to take a look whenever you find time.

## Requirements

- [ ] All fixes and/or new features come with corresponding tests. N/A

### DIFF
--- a/.github/workflows/daily-ci.yml
+++ b/.github/workflows/daily-ci.yml
@@ -51,6 +51,9 @@ jobs:
       env:
         NUM_PROCESSES: auto
         UV_RESOLUTION: ${{ matrix.dependencies-strategy }}
+        # Force a specific compiler for setuptools/distutils (used in gt4py.cartesian)
+        CXX: g++
+        CC: gcc
       run: uv run nox -s 'test_${{ matrix.gt4py-module }}-${{ matrix.python-version }}' -t 'cpu'
 
     - name: Notify slack
--- a/.github/workflows/test-cartesian.yml
+++ b/.github/workflows/test-cartesian.yml
@@ -53,5 +53,8 @@ jobs:
     - name: Run CPU 'cartesian' tests with nox
       env:
         NUM_PROCESSES: auto
+        # Force a specific compiler for setuptools/distutils
+        CXX: g++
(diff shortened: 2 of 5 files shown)


egparedes:
You're right that the "code quality" and the "daily ci" workflows don't need the full checkout anymore (they needed it at some point during the PR development), but decided to keep it anyway to get a more accurate version if needed for debugging. It's probably pointless so we can remove the full checkout from these two. In the release workflow we should use it to make sure we get the right version number.

BTW, the (not yet finished) PR #1982 will actually require to add full checkouts in more workflows for other reasons, so I could also remove the full checkout in the two workflows you identified here...

egparedes (commented):
You're right that the "code quality" and the "daily ci" workflows don't need the full checkout anymore (they needed it at some point during the PR development), but decided to keep it anyway to get a more accurate version if needed for debugging. It's probably pointless so we can remove the full checkout from these two. In the release workflow we should use it to make sure we get the right version number.

BTW, the not yet finished PR #1982 will require full checkout in other workflows too for other reasons, so we could also remove the full checkout from these two workflows in that PR...

romanc:
> BTW, the not yet finished PR #1982 will require full checkout in other workflows too for other reasons, so we could also remove the full checkout from these two workflows in that PR...

Ah, I see. Yeah, let's do it there. I wasn't aware of any other work going on in the CI, but since you already have a PR open, we don't need to keep this one.

PR #1997, opened 2025-04-24
feat[next]: GTIR concat_where embedded

havogt:
TODO:
- [ ] probably needs more changes because `DomainType` is moved


havogt:
too big as first step

PR #1999, opened 2025-04-24
feat[next][dace]: Add cpp timer to dace backend

edopao:
This PR builds on #1978 and adds support for `COMPUTE` metric timers in the dace backend. This metric is measured by means of CPP instrumentation of the top-level SDFG, which inserts a chrono timer around the SDFG execution. Collection of this metric overrides the asynchronous execution mode.


src/gt4py/next/program_processors/runners/dace/sdfg_callable.py
```diff
@@ -123,15 +124,13 @@ def get_sdfg_args(
         args:               The list of arguments passed to the dace runner.
         filter_args:        If True, return only the arguments that are expected
                             according to the SDFG signature.
-        on_gpu:             If True, this method ensures that the arrays for the
-                            connectivity tables are allocated in GPU memory.
```
philip-paul-mueller, line 127: The `get_sdfg_args()` function still has this argument in its signature.

src/gt4py/next/program_processors/runners/dace/sdfg_callable.py
```diff
@@ -98,7 +97,9 @@ def get_sdfg_conn_args(
             assert gtx_common.is_neighbor_connectivity(connectivity)
             assert field_utils.verify_device_field_type(
                 connectivity,
-                core_defs.CUPY_DEVICE_TYPE if on_gpu else core_defs.DeviceType.CPU,  # type: ignore[arg-type] # if we are `on_gpu` we expect `CUPY_DEVICE_TYPE` to be not `None`
+                core_defs.CUPY_DEVICE_TYPE
+                if sdfg.arrays[name].storage == dace.dtypes.StorageType.GPU_Global
+                else core_defs.DeviceType.CPU,  # type: ignore[arg-type] # if we are `on_gpu` we expect `CUPY_DEVICE_TYPE` to be not `None`
```
philip-paul-mueller, line 102: I am not an expert in that regard, so maybe my question is stupid.
Here you essentially changed the "source of truth" of where the memory of the connectivities is located.
Before, it was an argument now it is inferred from the data descriptors.
This means that you have to optimize the SDFG before you can call this function.
However, in the benchmark repo this function is called first (okay, maybe I am using it wrongly).

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -125,11 +131,8 @@ def __call__(
             #   is translating from `CompilableProgram` (ITIR.Program + CompileTimeArgs)
             #   to `ProgramSource`, so this step is storing in cache only the result
             #   of the SDFG transformations, not the compiled program binary.
+            dace_worflow_utils.set_dace_cache_config()
```
philip-paul-mueller, line 134: This is the minimal set that is needed.
However, I would say you should move everything inside this function.
This way you are sure that you always have the full GT4Py DaCe environment.
And when you need to change you only have to change it at one single location.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -14,57 +14,73 @@
...
+                fun.sdfg_program.sdfg,
                 offset_provider,
                 *flat_args,
                 filter_args=False,
-                on_gpu=on_gpu,
             )
-            with dace.config.temporary_config():
-                dace.config.Config.set("compiler", "allow_view_arguments", value=True)
-                return inp(**this_call_args)
+            with dace.config.set_temporary("compiler", "allow_view_arguments", value=True):
```
philip-paul-mueller, line 59: Despite my comment from above, here you should use `set_temporary()` as it will not write to disc.

philip-paul-mueller (commented):
There are some small points but nothing big.

philip-paul-mueller (approved):
LGTM

PR #1612, opened 2024-08-09
release the assert to be warning message to allow relaxed checking

xyuan:
This changes allow the unreachable code to be released with non-monotonic increasing field extensions of loop, as it trimmed the function scalar arguments, and cause crash in our code. This is bug fix.

- PR Title: release the assert to be warning message to allow relaxed checking

    <type>:
        - build: Changes that affect the build system or external dependencies
        - ci: Changes to our CI configuration files and scripts
        - docs: Documentation only changes
        - feat: A new feature
        - fix: A bug fix
        - perf: A code change that improves performance
        - refactor: A code change that neither fixes a bug nor adds a feature
        - style: Changes that do not affect the meaning of the code
        - test: Adding missing tests or correcting existing tests

    <scope>: cartesian | eve | next | storage
    # ONLY if changes are limited to a specific subsytem

- PR Description:

    Description of the main changes with links to appropriate issues/documents/references/...
-->

## Requirements

- [y ] All fixes and/or new features come with corresponding tests.
- [ ] Important design decisions have been documented in the approriate ADR inside the [docs/development/ADRs/](docs/development/ADRs/Index.md) folder.

If this PR contains code authored by new contributors please make sure:

- [ ] All the authors are covered by a valid contributor assignment agreement provided to ETH Zurich and signed by the employer if needed.
- [ ] The PR contains an updated version of the `AUTHORS.md` file adding the names of all the new contributors.

PR #2000, opened 2025-04-25
fix[cartesian]: daceir DomainInterval intersection

romanc:
## Description

Fixes argument validation in intersection code of DaCeIR's DomainInterval. Adds missing test cases.

## Requirements


FlorianDeconinck (approved):
LGTM - yay for stronger checks!

romanc:
Linking https://github.com/GridTools/gt4py/pull/1612 here for future reference since both PRs arrived at the same `assert`. Back then, we were chasing something different and didn't realize that the `assert` was missing the case of "A is contained in B" (or B in A for what it's worth).

PR #2002, opened 2025-04-28
fix[dace][next]: Fixed `MoveDataflowIntoIfBody`

philip-paul-mueller:
This PR fixes how the if mover handling the "patching", i.e. how it makes the data that are needed inside available. The previous implementation did not considered the case when an external AccessNode had multiple edges to the enclosing Map. It detected correctly that it only had to create it once, however the clean up was wrong, as it would remove the inner edge but leave the outer edge, the one connecting the AccessNode with the enclosing Map alive.
This created dangling connectors.


src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -398,6 +407,36 @@ def _replicate_dataflow_into_branch(
...
+        #  node, that was relocated, has incoming connections we must remove them
+        #  and the parent dataflow.
+        for node_to_remove in all_relocatable_dataflow:
+            for iedge in list(state.in_edges(node_to_remove)):
+                if iedge.src in all_relocatable_dataflow:
+                    continue
+                dace_sutils.remove_edge_and_dangling_path(state, iedge)
+
+            if isinstance(node_to_remove, dace_nodes.AccessNode):
+                assert node_to_remove.desc(sdfg).transient
```
edopao, line 434: Maybe this assert could be put at the top of the loop body, at line 428.

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -398,6 +407,36 @@ def _replicate_dataflow_into_branch(
...
+                sdfg.remove_data(node_to_remove.data, validate=False)
+
+        # Remove the original nodes.
```
edopao, line 437: You first remove the array data, then the access node. That's fine, it works. Maybe just add to the line comment:
```suggestion
        # Remove the original nodes (the array data was already deleted in the above loop).
```

edopao (approved):

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
philip-paul-mueller: Then you would need to check two times if `node_to_remove` is an AccessNode unless you also move the call to `remove_data()` up.
However, I do not want to do that, because the `remove_edge_and_dangling_path()` might get confused if the data has been removed. so I would like to keep it there.

PR #2004, opened 2025-04-28
fix[dace][next]: Fixes expanding of Not Supported Memlets

philip-paul-mueller:
We have to expand some Memlets into copy Maps, because there is no suitable call to `cudaMemcpy*`, because we have to set their order right. But this is a gross hack, it should be solved once [PR#1913](https://github.com/GridTools/gt4py/pull/1913) has been merged, but for that DaCe [PR#1976](https://github.com/spcl/dace/pull/1976) has to be merged.

The error was that there was the assumption that at least one of these newly created Maps survived. However, this does not seem to be the case, depending in which order MapFusion is applied. Thus the assert was removed.
We now assume that if we do not find a newly created MapEntry that is was integrated into a Map that was already present, we further assume that the Map parameter of that Map were used and that they are correct.


edopao (approved):
Tested, it works.

PR #2005, opened 2025-04-28
feat[dace][next]: Updated `MapFusion`

philip-paul-mueller:
It is now possible to handle cases where the intermediate has multiple producer.
This will also be turned into a DaCe PR.


edopao (approved):
LGTM

PR #2007, opened 2025-04-28
fix[next]: iterator type inference: propagate a DeferredType

havogt:


tehrengruber:
As discussed obsolete after https://github.com/GridTools/gt4py/pull/2010 and https://github.com/GridTools/gt4py/pull/2008

PR #2008, opened 2025-04-29
fix[next]: Preserve annex in fixed point transformations

tehrengruber:


src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -328,7 +328,7 @@ def transform_letify_make_tuple_elements(
             for arg in node.args:
```
tehrengruber, line 328: The changes in this file are required to preserve the domain annex attribute in case the `_ct_el` or `__ct_flat_remapped_` are not removed, which happens occasionally.

egparedes (approved):
LGTM

PR #2010, opened 2025-04-29
fix[next]: Make type reinference optional in fixed point transform

tehrengruber:


tehrengruber:
Has been orally approved by @havogt. Merging.

PR #2013, opened 2025-04-30
fix[next]: Prevent extraction from lambdas in cse pass

tehrengruber:
The common subexpression elimination pass already has a mechanism to avoid extracting from `if_` calls, such that the evaluation of expressions can be guarded in order to to avoid out-of-bounds accesses or excessive pre-computation. The former protection can be circumvented by placing things inside a lambda outside of the protecting if:
```
if cond then
  let f = λ() → ·guarded_it in
    if cond2 then f() else 0
else
  ·guarded_it
```
This PR prevents that by not extracting from lambdas unless they are a let-form in which case it is clear that the function is unconditionally evaluated.

### DIFF
--- a/tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_cse.py
+++ b/tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_cse.py
@@ -304,8 +304,6 @@ def test_scalar_extraction_inside_as_fieldop():
 def test_no_extraction_from_unapplied_lambda():
     testee = im.if_(
         "cond",
-        # the `+0`, and `+1` respectively, are only needed to avoid that f itself is collected
-        # which is not what we are interested in here
         im.let("f", im.lambda_()(im.deref("guarded_it")))(im.if_("cond2", im.call("f")(), 0)),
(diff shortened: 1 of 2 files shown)


tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_cse.py
```diff
@@ -293,3 +299,27 @@ def test_scalar_extraction_inside_as_fieldop():
 
     actual = CSE.apply(testee, within_stencil=False)
     assert actual == expected
+
+
+def test_no_extraction_from_unapplied_lambda():
+    testee = im.if_(
+        "cond",
+        # the `+0`, and `+1` respectively, are only needed to avoid that f itself is collected
+        # which is not what we are interested in here
```
egparedes, line 308: Is this the comment in sync with the rest of the code? I don't understand what  `+0` and `+1` mean here. Maybe would help to copy here the IR/pseudocode example from the PR description...

egparedes (commented):
It looks good to me. Only have questions and comments about style.

egparedes (approved):
LGTM

PR #2014, opened 2025-04-30
refactor[next][dace]: Introduce context dataclass in SDFG lowering

edopao:
This PR is a code refactoring that introduces a `SubgraphContext` class to pass context information through the lowering of GTIR (it replaces sdfg and state).

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
@@ -129,6 +129,14 @@ class DataflowBuilder(Protocol):
         return state.add_mapped_tasklet(unique_name, map_ranges, inputs, code, outputs, **kwargs)
 
 
+@dataclasses.dataclass(frozen=True)
+class SubgraphContext:
+    """Represents the subgraph context in which to lower a GTIR expression to dataflow."""
+
+    sdfg: dace.SDFG
+    state: dace.SDFGState
+
+
 class SDFGBuilder(DataflowBuilder, Protocol):
     """Visitor interface available to GTIR-primitive translators."""
 
@@ -155,11 +163,11 @@ class SDFGBuilder(DataflowBuilder, Protocol):
     def setup_nested_context(
         self,
         expr: gtir.Lambda,
-        sdfg: dace.SDFG,
-        parent: dace.SDFG,
+        sdfg_name: str,
+        parent_ctx: SubgraphContext,
         scope_symbols: dict[str, ts.DataType],
         symbolic_inputs: set[str],
-    ) -> SDFGBuilder:
+    ) -> tuple[SDFGBuilder, SubgraphContext]:
         """
         Create an nested SDFG context to lower a lambda expression, indipendent
         from the current context where the parent expression is being translated.
@@ -171,8 +179,8 @@ class SDFGBuilder(DataflowBuilder, Protocol):
 
         Args:
             expr: The lambda expression to be lowered as a nested SDFG.
-            sdfg: The nested SDFG where to lower the nested expression.
-            parent: The parent SDFG.
+            sdfg_name: The name of the nested SDFG where to lower the given expression.
+            parent_ctx: The parent SDFG context.
             scope_symbols: Mapping from symbol name to data type for the GTIR symbols
                 forwarded to the nested context.
             symbolic_inputs: Arguments that have to be passed to the nested SDFG
@@ -302,12 +310,17 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
     def setup_nested_context(
         self,
         expr: gtir.Lambda,
-        sdfg: dace.SDFG,
-        parent: dace.SDFG,
+        sdfg_name: str,
+        parent_ctx: SubgraphContext,
         scope_symbols: dict[str, ts.DataType],
         symbolic_inputs: set[str],
-    ) -> SDFGBuilder:
+    ) -> tuple[SDFGBuilder, SubgraphContext]:
+        sdfg = dace.SDFG(name=self.unique_nsdfg_name(parent_ctx.sdfg, sdfg_name))
+        sdfg.debuginfo = gtir_to_sdfg_utils.debug_info(expr, default=parent_ctx.sdfg.debuginfo)
+        state = sdfg.add_state(f"{sdfg_name}_entry")
+        nested_ctx = SubgraphContext(sdfg, state)
         nsdfg_builder = GTIRToSDFG(self.offset_provider_type, self.column_axis, scope_symbols)
+
         # We pass to the nested SDFG all GTIR-symbols in scope, which includes the
         # values mapped to the parameters of the lambda expression (`node.params`)
         # and the GTIR-symbols defined in the current context.
@@ -331,7 +344,9 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
         # Scalar GTIR-symbols represented as dace symbols in parent SDFG are mapped
         # to dace symbols in the nested SDFG.
         parent_symbols = {
-            name for p in flatten_tuple_syms(params) if (name := str(p.id)) in parent.symbols
+            name
+            for p in flatten_tuple_syms(params)
+            if (name := str(p.id)) in parent_ctx.sdfg.symbols
         }
 
         # All GTIR-symbols accessed in domain expressions by the lambda need to be
@@ -354,7 +369,7 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
             params,
             symbolic_params=(domain_symbols | parent_symbols | symbolic_inputs),
         )
-        return nsdfg_builder
+        return nsdfg_builder, nested_ctx
 
     def unique_nsdfg_name(self, sdfg: dace.SDFG, prefix: str) -> str:
         nsdfg_list = [
@@ -524,7 +539,11 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
         raise NotImplementedError("Temporaries not supported yet by GTIR DaCe backend.")
 
     def _visit_expression(
-        self, node: gtir.Expr, sdfg: dace.SDFG, head_state: dace.SDFGState, use_temp: bool = True
+        self,
+        node: gtir.Expr,
+        sdfg: dace.SDFG,
+        head_state: dace.SDFGState,
+        use_temp: bool = True,
     ) -> list[gtir_to_sdfg_types.FieldopData]:
         """
         Specialized visit method for fieldview expressions.
@@ -535,7 +554,9 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
         Returns:
             A list of array nodes containing the result fields.
         """
-        result = self.visit(node, sdfg=sdfg, head_state=head_state)
+
+        ctx = SubgraphContext(sdfg, head_state)
+        result = self.visit(node, ctx=ctx)
 
         # sanity check: each statement should preserve the property of single exit state (aka head state),
         # i.e. eventually only introduce internal branches, and keep the same head state
@@ -744,22 +765,21 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
     def visit_FunCall(
         self,
         node: gtir.FunCall,
-        sdfg: dace.SDFG,
-        head_state: dace.SDFGState,
+        ctx: SubgraphContext,
     ) -> gtir_to_sdfg_types.FieldopResult:
         # use specialized dataflow builder classes for each builtin function
         if cpm.is_call_to(node, "concat_where"):
-            return gtir_to_sdfg_concat_where.translate_concat_where(node, sdfg, head_state, self)
+            return gtir_to_sdfg_concat_where.translate_concat_where(node, ctx, self)
         elif cpm.is_call_to(node, "if_"):
-            return gtir_to_sdfg_primitives.translate_if(node, sdfg, head_state, self)
+            return gtir_to_sdfg_primitives.translate_if(node, ctx, self)
         elif cpm.is_call_to(node, "index"):
-            return gtir_to_sdfg_primitives.translate_index(node, sdfg, head_state, self)
+            return gtir_to_sdfg_primitives.translate_index(node, ctx, self)
         elif cpm.is_call_to(node, "make_tuple"):
-            return gtir_to_sdfg_primitives.translate_make_tuple(node, sdfg, head_state, self)
+            return gtir_to_sdfg_primitives.translate_make_tuple(node, ctx, self)
         elif cpm.is_call_to(node, "tuple_get"):
-            return gtir_to_sdfg_primitives.translate_tuple_get(node, sdfg, head_state, self)
+            return gtir_to_sdfg_primitives.translate_tuple_get(node, ctx, self)
         elif cpm.is_applied_as_fieldop(node):
-            return gtir_to_sdfg_primitives.translate_as_fieldop(node, sdfg, head_state, self)
+            return gtir_to_sdfg_primitives.translate_as_fieldop(node, ctx, self)
         elif isinstance(node.fun, gtir.Lambda):
             # Special handling of scalar arguments of a let-lambda that can be lowered
             # as symbolic expressions: when all the GTIR-symbols the scalar expression
@@ -780,37 +800,27 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
                     ):
                         raise
                     continue
-                if all(str(s) in sdfg.symbols for s in symbolic_expr.free_symbols):
+                if all(str(s) in ctx.sdfg.symbols for s in symbolic_expr.free_symbols):
                     symbolic_args[str(p.id)] = symbolic_expr
             # All other lambda arguments are lowered to some dataflow that produces a data node.
             args = {
                 str(p.id): (
                     gtir_to_sdfg_types.SymbolicData(p.type, symbolic_args[param])  # type: ignore[arg-type]
                     if (param := str(p.id)) in symbolic_args
-                    else self.visit(
-                        arg,
-                        sdfg=sdfg,
-                        head_state=head_state,
-                    )
+                    else self.visit(arg, ctx=ctx)
                 )
                 for p, arg in zip(node.fun.params, node.args, strict=True)
             }
-            return self.visit(
-                node.fun,
-                sdfg=sdfg,
-                head_state=head_state,
-                args=args,
-            )
+            return self.visit(node.fun, ctx=ctx, args=args)
         elif isinstance(node.type, ts.ScalarType):
-            return gtir_to_sdfg_primitives.translate_scalar_expr(node, sdfg, head_state, self)
+            return gtir_to_sdfg_primitives.translate_scalar_expr(node, ctx, self)
         else:
             raise NotImplementedError(f"Unexpected 'FunCall' expression ({node}).")
 
     def visit_Lambda(
         self,
         node: gtir.Lambda,
-        sdfg: dace.SDFG,
-        head_state: dace.SDFGState,
+        ctx: SubgraphContext,
         args: Mapping[str, gtir_to_sdfg_types.FieldopResult | gtir_to_sdfg_types.SymbolicData],
     ) -> gtir_to_sdfg_types.FieldopResult:
         """
@@ -859,18 +869,11 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
         }
 
         # lower let-statement lambda node as a nested SDFG
-        nsdfg = dace.SDFG(name=self.unique_nsdfg_name(sdfg, "lambda"))
-        nsdfg.debuginfo = gtir_to_sdfg_utils.debug_info(node, default=sdfg.debuginfo)
-        lambda_translator = self.setup_nested_context(
-            node, nsdfg, sdfg, lambda_symbols, symbolic_inputs=set(symbolic_args.keys())
+        lamnda_translator, lambda_ctx = self.setup_nested_context(
+            node, "lambda", ctx, lambda_symbols, symbolic_inputs=set(symbolic_args.keys())
         )
 
-        nstate = nsdfg.add_state("lambda")
-        lambda_result = lambda_translator.visit(
-            node.expr,
-            sdfg=nsdfg,
-            head_state=nstate,
-        )
+        lambda_result = lamnda_translator.visit(node.expr, ctx=lambda_ctx)
 
         # Process lambda inputs
         #
@@ -883,23 +886,23 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
         }
 
         input_memlets = {}
-        for nsdfg_dataname, nsdfg_datadesc in nsdfg.arrays.items():
+        for nsdfg_dataname, nsdfg_datadesc in lambda_ctx.sdfg.arrays.items():
             if nsdfg_datadesc.transient:
                 continue
 
             if nsdfg_dataname in lambda_arg_nodes:
                 src_node = lambda_arg_nodes[nsdfg_dataname].dc_node
                 dataname = src_node.data
-                datadesc = src_node.desc(sdfg)
+                datadesc = src_node.desc(ctx.sdfg)
             else:
                 dataname = nsdfg_dataname
-                datadesc = sdfg.arrays[nsdfg_dataname]
+                datadesc = ctx.sdfg.arrays[nsdfg_dataname]
 
             # ensure that connectivity tables are non-transient arrays in parent SDFG
             if dataname in connectivity_arrays:
                 datadesc.transient = False
 
-            input_memlets[nsdfg_dataname] = sdfg.make_array_memlet(dataname)
+            input_memlets[nsdfg_dataname] = ctx.sdfg.make_array_memlet(dataname)
 
         # Process lambda outputs
         #
@@ -925,12 +928,12 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
         lambda_outputs = {
             output_data.dc_node.data
             for output_data in lambda_output_data
-            if output_data.dc_node.desc(nsdfg).transient
+            if output_data.dc_node.desc(lambda_ctx.sdfg).transient
         }
 
         # Map free symbols to parent SDFG
         nsdfg_symbols_mapping = {}
-        for sym in nsdfg.free_symbols:
+        for sym in lambda_ctx.sdfg.free_symbols:
             if (sym_id := str(sym)) in lambda_arg_nodes:
                 assert isinstance(lambda_arg_nodes[sym_id].gt_type, ts.ScalarType)
                 raise NotImplementedError(
@@ -941,24 +944,24 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
             else:
                 nsdfg_symbols_mapping[sym_id] = sym
         for param, arg in data_args.items():
-            nsdfg_symbols_mapping |= gtir_to_sdfg_utils.get_arg_symbol_mapping(param, arg, sdfg)
+            nsdfg_symbols_mapping |= gtir_to_sdfg_utils.get_arg_symbol_mapping(param, arg, ctx.sdfg)
 
-        nsdfg_node = head_state.add_nested_sdfg(
-            nsdfg,
-            parent=sdfg,
+        nsdfg_node = ctx.state.add_nested_sdfg(
+            lambda_ctx.sdfg,
+            parent=ctx.sdfg,
             inputs=set(input_memlets.keys()),
             outputs=lambda_outputs,
             symbol_mapping=nsdfg_symbols_mapping,
-            debuginfo=gtir_to_sdfg_utils.debug_info(node, default=sdfg.debuginfo),
+            debuginfo=gtir_to_sdfg_utils.debug_info(node, default=ctx.sdfg.debuginfo),
         )
 
         for connector, memlet in input_memlets.items():
             if connector in lambda_arg_nodes:
                 src_node = lambda_arg_nodes[connector].dc_node
             else:
-                src_node = head_state.add_access(memlet.data)
+                src_node = ctx.state.add_access(memlet.data)
 
-            head_state.add_edge(src_node, None, nsdfg_node, connector, memlet)
+            ctx.state.add_edge(src_node, None, nsdfg_node, connector, memlet)
 
         def construct_output_for_nested_sdfg(
             inner_data: gtir_to_sdfg_types.FieldopData,
@@ -975,7 +978,7 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
             The same happens to symbols available in the lambda context but not explicitly passed as lambda
             arguments, that are simply returned by the lambda: it can be directly accessed in the parent SDFG.
             """
-            inner_desc = inner_data.dc_node.desc(nsdfg)
+            inner_desc = inner_data.dc_node.desc(lambda_ctx.sdfg)
             inner_dataname = inner_data.dc_node.data
             if inner_desc.transient:
                 # Transient data nodes only exist within the nested SDFG. In order to return some result data,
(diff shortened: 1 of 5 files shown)


src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -0,0 +1,182 @@
...
+
+DomainRange: TypeAlias = list[tuple[gtx_common.Dimension, dace_subsets.Subset]]
+"""
+Domain of a field operator represented as a list of tuples with 2 elements:
+ - dimension definition
+ - dimension range represented as a dace subset
+"""
+
+
+class GTIRDomainParser:
```
philip-paul-mueller, line 31: I am not sure if a class is the right approach here.
As this class essentially only has one function, `simplify()`.
I am guessing that you choosed that approach because of the work you perform in `__init__()`, but I am not sure how much you safe there.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -0,0 +1,182 @@
...
+    Construct the memlet subset to access a point in the field domain.
+
+    Args:
+        domain: The field domain.
+
+    Returns:
+        Range to be used as memlet subset.
+    """
+    if len(domain) == 0:
+        return dace_subsets.Range([])
```
philip-paul-mueller, line 179: When this is the scalar case, I would say this should be `'0'`, or do I miss something here.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -115,6 +116,19 @@ def add_mapped_tasklet(
         return state.add_mapped_tasklet(unique_name, map_ranges, inputs, code, outputs, **kwargs)
 
 
+@dataclasses.dataclass(frozen=True)
+class SDFGContext:
+    """Represents the SDFG context in which to lower a GTIR node to dataflow."""
```
philip-paul-mueller, line 121: I like this combination of the state and the sdfg.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -115,6 +116,19 @@ def add_mapped_tasklet(
...
+
+    sdfg: dace.SDFG
+    state: dace.SDFGState
+    domain_parser: gtir_domain.GTIRDomainParser
```
philip-paul-mueller, line 125: This is more a conceptional point.
Why is the domain parser part of the context and not on its own.
Because for me an SDFG does not need to have a single "domain" (which might be a bit of an overloaded word), so an SDFG can have multiple domains.
Maybe the name is just a little bit of.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -311,21 +331,32 @@ def is_column_axis(self, dim: gtx_common.Dimension) -> bool:
...
         self,
-        sdfg: dace.SDFG,
-        parent: dace.SDFG,
+        node: gtir.Node,
+        sdfg_name: str,
+        parent_ctx: SDFGContext,
         scope_symbols: dict[str, ts.DataType],
-    ) -> SDFGBuilder:
-        nsdfg_builder = GTIRToSDFG(self.offset_provider_type, self.column_axis, scope_symbols)
+        state_name: Optional[str] = None,
```
philip-paul-mueller, line 338: A bit out of scope but I would add a description to this function.
Especially `node` confused me until I saw that it was a GTIR node and its only usage is to serve as the source for the debug information.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -311,21 +331,32 @@ def is_column_axis(self, dim: gtx_common.Dimension) -> bool:
...
+    ) -> tuple[SDFGBuilder, SDFGContext]:
+        sdfg = dace.SDFG(name=self.unique_nsdfg_name(parent_ctx.sdfg, sdfg_name))
```
philip-paul-mueller, line 340: For better understanding I would consider naming this `nested_sdfg` or maybe not.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -115,6 +116,19 @@ def add_mapped_tasklet(
         return state.add_mapped_tasklet(unique_name, map_ranges, inputs, code, outputs, **kwargs)
 
 
+@dataclasses.dataclass(frozen=True)
+class SDFGContext:
+    """Represents the SDFG context in which to lower a GTIR node to dataflow."""
+
+    sdfg: dace.SDFG
+    state: dace.SDFGState
```
philip-paul-mueller, line 124: Since the class is frozen, does this means than an SDFG can only have one state?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg_utils.py
```diff
@@ -127,3 +130,16 @@ def visit_SymRef(self, node: gtir.SymRef, *, symtable: Dict[str, str]) -> gtir.S
     # assert that the new symbol names are not used in the IR
     assert ir_sym_ids.isdisjoint(invalid_symbols_mapping.values())
     return ReplaceSymbols().visit(ir, symtable=invalid_symbols_mapping)
+
+
+def get_symbolic(expr: gtir.Expr) -> dace.symbolic.SymbolicType:
+    """
+    Specialized visit method to lower symbolic expressions.
```
philip-paul-mueller, line 137: Does transforming it into a symbolic expression limits what you can digest here.
Or could you get a GT4Py expression that is not a DaCe symbolic expression?

philip-paul-mueller (commented):
There are some aspects that needs clarifying.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -0,0 +1,182 @@
...
+        """Simplifies a symbolic domain expression by applying some constraints.
+
+        Args:
+            expr: The symbolic expression to simplify.
+
+        Returns:
+            A new symbolic expression.
+        """
+        for lb, ub, size in self.domain_constraints:
+            expr = expr.subs(lb, ub - size).subs(size, ub - lb)
```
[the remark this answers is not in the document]
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
philip-paul-mueller: I guess you gain something, but you are right we would need to have an experiment.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
philip-paul-mueller: Then it makes sense.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
philip-paul-mueller: Thanks for the explanation.
In that case I would say that `DataflowContext` is also inaccurate, because dataflow could technically extend over multiple state, I mean this is what the SDFG stands for.
However, I think that by adding the word "state" into the name, such as `StateDataflowContext` or `InStateDataflowContext` you could solve that.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -0,0 +1,146 @@
...
+    return domain_dims, domain_lbs, domain_sizes
+
+
+def get_field_subset(domain: DomainRange) -> dace_subsets.Range:
+    """
+    Construct the memlet subset to access a point in the field global domain.
+
+    The subset is limited to only the global dimensions. The remaining part of the
+    subset (being it a local list or the scalar representing a zero-dimensional
+    field) should be handled outside by the caller.
```
philip-paul-mueller, line 134:
```suggestion
    field) must be handled outside by the caller.
```

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
philip-paul-mueller: Then this means that this context is only used for a single state?
In that case you should add this to the documentation.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -115,6 +116,18 @@ def add_mapped_tasklet(
...
 
 
+@dataclasses.dataclass(frozen=True)
+class DataflowContext:
+    """Represents the SDFG context in which to lower a GTIR node to dataflow."""
+
+    sdfg: dace.SDFG
+    state: dace.SDFGState
+
+    def clone(self, state: dace.SDFGState) -> "DataflowContext":
```
philip-paul-mueller, line 126: When I see a method with the name `clone()` then I expect that it returns a copy of it.
However, here it is essentially the function is essentially a sloppy constructor as you do not need to specify the SDFG.
Thus either give give it a better name, such as `creat_new_state_context_for()` (the missing `e` is intended) or remove it and use the, in this case, much more expressive constructor directly.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -140,26 +153,34 @@ def is_column_axis(self, dim: gtx_common.Dimension) -> bool:
...
+
         This method will setup the global symbols, that correspond to the parameters
-        of the expression to be lowered, as well as the set of symbolic arguments,
-        that is scalar values represented as dace symbols in the parent SDFG.
+        of the expression to be lowered, as well as the set of symbolic arguments.
 
         Args:
-            sdfg: The SDFG where to lower the nested expression.
-            parent: The parent SDFG.
+            node: The expression to be lowered as a nested SDFG.
```
philip-paul-mueller, line 173:
```suggestion
            node: The GTIR expression to be lowered as a nested SDFG.
```

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -140,26 +153,34 @@ def is_column_axis(self, dim: gtx_common.Dimension) -> bool:
...
+        parent_ctx: DataflowContext,
         scope_symbols: dict[str, ts.DataType],
-    ) -> SDFGBuilder:
+        state_name: Optional[str] = None,
+    ) -> tuple[SDFGBuilder, DataflowContext]:
         """
         Create an SDFG context to translate a nested expression, indipendent
         from the current context where the parent expression is being translated.
 
+        The nested expression is a `Lambda` node, therefore it comes with a list
```
philip-paul-mueller, line 166:
```suggestion
        The nested expression, passed through the argument `node`, is a `Lambda` node, therefore it comes with a list
```

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -24,45 +23,10 @@
...
 Domain of a field operator represented as a list of tuples with 2 elements:
  - dimension definition
- - dimension range represented as a dace subset
+ - dimension range represented as a dace subset: start, stop (included), step
 """
 
 
-class GTIRDomainParser:
-    """Utility class to apply domain constraints on a dace symbolic expression.
-
```
philip-paul-mueller: Without this functionality, does the title of this PR and the PR still make sense?

philip-paul-mueller (commented):
There are some things but it looks good.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: Good point, I will use the constructor directly.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
edopao: Fixed

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: Renamed to `SDFGBuilderContext`.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -0,0 +1,140 @@
...
+from gt4py.next.iterator import ir as gtir
+from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm, domain_utils
+from gt4py.next.program_processors.runners.dace import gtir_python_codegen, gtir_sdfg_utils
+
+
+FieldopDomain: TypeAlias = list[tuple[gtx_common.Dimension, dace_subsets.Subset]]
+"""
+Domain of a field operator represented as a list of tuples with 2 elements:
+ - dimension definition
+ - dimension range represented as a dace subset: start, stop (included), step
```
philip-paul-mueller: The wording implies a `Range` but you wrote `Subset`.
This is problematic as an `Index` is a subset but does not have a step nor an end.
Since you in almost all cases have a `Range` I would suggest it to change the annotation to `Range`.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -0,0 +1,140 @@
...
+        for dim, drange in node.ranges.items():
+            lower_bound = parse_range_boundary(drange.start)
+            upper_bound = parse_range_boundary(drange.stop)
+            domain_dims.append(dim)
+            domain_range.append((lower_bound, upper_bound - 1, 1))
+
+    else:
+        raise ValueError(f"Invalid domain {node}.")
+
+    return list(zip(domain_dims, dace_subsets.Range(domain_range), strict=True))
```
philip-paul-mueller: In the optimization pipeline I use `dace_sbs` for the subsets.
I use it because in DaCe they abbreviate it that way.
To make everything a bit more uniform I would suggest that we settle on of the two.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def get_domain_indices(
+    dims: Sequence[gtx_common.Dimension], origin: Sequence[dace.symbolic.SymExpr] | None
+) -> dace_subsets.Indices:
+    """
+    Construct the list of indices for a field domain, applying an optional origin
```
philip-paul-mueller: I very much like that `origin` does not have a default argument, which forces the user to think about if it is used, so I would not change that.
The word "optional", however, indicates to me that it is possible to not specify the argument, so I would remove that line, I think that the description for the argument "If set to `None`, ..." is clear enough.

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
```diff
@@ -194,7 +192,8 @@ def _create_scan_field_operator(
     Refer to `gtir_builtin_translators._create_field_operator()` for the
     description of function arguments and return values.
     """
-    domain_dims, _, _ = gtir_translators.get_field_layout(domain)
+    sdfg, state = ctx.sdfg, ctx.state
```
philip-paul-mueller: This looks a bit strange to me.
It looks like the only purpose of this line looks like compatibility.
So I would replace every occurrence of `sdfg` with `ctx.sdfg` otherwise `ctx` does not make sense.

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
```diff
@@ -310,18 +309,19 @@ def _lower_lambda_to_nested_sdfg(
     """
 
     # the lambda expression, i.e. body of the scan, will be created inside a nested SDFG.
-    nsdfg = dace.SDFG(sdfg_builder.unique_nsdfg_name(sdfg, "scan"))
-    nsdfg.debuginfo = gtir_sdfg_utils.debug_info(lambda_node, default=sdfg.debuginfo)
+    lambda_translator, lambda_init_ctx = sdfg_builder.setup_nested_context(
+        lambda_node, "scan", ctx, lambda_symbols, state_name="scan_init"
+    )
+    nsdfg, init_state = lambda_init_ctx.sdfg, lambda_init_ctx.state
```
philip-paul-mueller: This also looks like a compatibility hack to avoid to fully go to the new context.

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
```diff
@@ -310,18 +309,19 @@ def _lower_lambda_to_nested_sdfg(
...
     # This property is used by pattern matching in SDFG transformation framework
     # to skip those transformations that do not yet support control flow blocks.
     nsdfg.using_explicit_control_flow = True
-    lambda_translator = sdfg_builder.setup_nested_context(lambda_node, nsdfg, sdfg, lambda_symbols)
 
     # use the vertical dimension in the domain as scan dimension
     scan_domain = [
-        (dim, lower_bound, upper_bound)
-        for dim, lower_bound, upper_bound in domain
+        (dim, dim_range[0], dim_range[1] + dim_range[2])
```
philip-paul-mueller: I do not understand what the `+ dim_range[2]` is for?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
philip-paul-mueller: Now that I think about it, I do not fully like the name, as it indicates to me, that I can use it to fully build the SDFG, with an unlimited number of states.
So I would suggest something like `InStateDafaflowBuilderContext` okay maybe a bit long.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -206,16 +224,16 @@ def _collect_symbols_in_domain_expressions(
...
     """Helper method to build a memlet subset of a field over the given domain."""
     # convert domain expression to dictionary to ease access to the dimensions,
     # since the access indices have to follow the order of dimensions in field domain
     if isinstance(data.gt_type, ts.FieldType) and len(data.gt_type.dims) != 0:
         assert data.origin is not None
-        domain_ranges = {dim: (lb, ub) for dim, lb, ub in domain}
+        domain_ranges = dict(domain)
         return dace.subsets.Range(
-            (domain_ranges[dim][0] - origin, domain_ranges[dim][1] - origin - 1, 1)
+            (domain_ranges[dim][0] - origin, domain_ranges[dim][1] - origin, domain_ranges[dim][2])
```
philip-paul-mueller: Also here, why do you include the step?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -550,7 +578,11 @@ def _add_storage_for_temporary(self, temp_decl: gtir.Temporary) -> dict[str, str
         raise NotImplementedError("Temporaries not supported yet by GTIR DaCe backend.")
 
     def _visit_expression(
-        self, node: gtir.Expr, sdfg: dace.SDFG, head_state: dace.SDFGState, use_temp: bool = True
+        self,
+        node: gtir.Expr,
+        sdfg: dace.SDFG,
+        head_state: dace.SDFGState,
```
philip-paul-mueller: Why does this function has to build the context itself and is not passed directly?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -862,23 +875,23 @@ def visit_Lambda(
...
 
             if nsdfg_dataname in lambda_arg_nodes:
                 src_node = lambda_arg_nodes[nsdfg_dataname].dc_node
                 dataname = src_node.data
-                datadesc = src_node.desc(sdfg)
+                datadesc = src_node.desc(ctx.sdfg)
             else:
                 dataname = nsdfg_dataname
-                datadesc = sdfg.arrays[nsdfg_dataname]
+                datadesc = ctx.sdfg.arrays[nsdfg_dataname]
```
philip-paul-mueller: It looks to me as you could remove the `if` and only keep the `else` branch or is there something I miss?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -956,22 +969,22 @@ def construct_output_for_nested_sdfg(
...
+            inner_desc = inner_data.dc_node.desc(lambda_ctx.sdfg)
             inner_dataname = inner_data.dc_node.data
             if inner_desc.transient:
                 # Transient data nodes only exist within the nested SDFG. In order to return some result data,
                 # the corresponding data container inside the nested SDFG has to be changed to non-transient,
                 # that is externally allocated, as required by the SDFG IR. An output edge will write the result
                 # from the nested-SDFG to a new intermediate data container allocated in the parent SDFG.
                 outer_data = inner_data.map_to_parent_sdfg(
-                    self, nsdfg, sdfg, head_state, nsdfg_symbols_mapping
+                    self, lambda_ctx.sdfg, ctx.sdfg, ctx.state, nsdfg_symbols_mapping
```
philip-paul-mueller: Would it be possible to only pass the context here?

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -387,17 +312,12 @@ def _create_field_operator_impl(
         The field data descriptor, which includes the field access node in the
         given `state` and the field domain offset.
     """
+    sdfg, state = ctx.sdfg, ctx.state
```
philip-paul-mueller: Also here you should either refer to `ctx.sdfg` or `ctx.state` or scrap the idea of the context.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -472,11 +390,8 @@ def _create_field_operator(
         }
     else:
         # create map range corresponding to the field operator domain
-        map_range = {
-            gtir_sdfg_utils.get_map_variable(dim): f"{lower_bound}:{upper_bound}"
-            for dim, lower_bound, upper_bound in domain
-        }
-    map_entry, map_exit = sdfg_builder.add_map("fieldop", state, map_range)
+        map_range = {gtir_sdfg_utils.get_map_variable(dim): dim_range for dim, dim_range in domain}
+    map_entry, map_exit = sdfg_builder.add_map("fieldop", ctx.state, map_range)
```
philip-paul-mueller: I would suggest to update `add_map()` such that it accepts contexts instead of states directly.
I think if you apply this idea to other functions, you could improve the readability.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -698,15 +577,16 @@ def _write_if_branch_output(
...
-    state: dace.SDFGState,
+    ctx: gtir_sdfg.SDFGBuilderContext,
     sdfg_builder: gtir_sdfg.SDFGBuilder,
 ) -> FieldopResult:
     """Generates the dataflow subgraph for the `if_` builtin function."""
     assert cpm.is_call_to(node, "if_")
     assert len(node.args) == 3
     cond_expr, true_expr, false_expr = node.args
 
+    sdfg, state = ctx.sdfg, ctx.state
```
philip-paul-mueller: Also here use `ctx.sdfg` or `ctx.state` directly.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -915,40 +791,30 @@ def _get_symbolic_value(
...
-    state: dace.SDFGState,
+    ctx: gtir_sdfg.SDFGBuilderContext,
     sdfg_builder: gtir_sdfg.SDFGBuilder,
 ) -> FieldopResult:
     """Generates the dataflow subgraph for a `ir.Literal` node."""
     assert isinstance(node, gtir.Literal)
 
     data_type = node.type
-    data_node = _get_symbolic_value(sdfg, state, sdfg_builder, node.value, data_type)
+    data_node = _get_symbolic_value(ctx.sdfg, ctx.state, sdfg_builder, node.value, data_type)
```
philip-paul-mueller: This function also looks like a candidate that should be changed to accept contexts directly.

philip-paul-mueller (commented):
It looks good, there are some improvements and some inconsistency that should be fixed.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
edopao: Agreed.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
edopao: Good idea.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
edopao: Agreed

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
edopao: Good catch. Yes, the domain range is indeed a `Range`.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
edopao: `dace_sbs` is shorter, so let's go for it.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
edopao: Good point.

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
edopao: `dim_range[2]` is the range step. I want to compute the not-included boundary of the range. I will add a comment.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: Well, here I have to. The `Range` type is a tuple of three elements, where the third one is the step.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: Good idea.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: The arg node (`lambda_arg_nodes[nsdfg_dataname].dc_node.data`) is a different data container than the inner container (`nsdfg_dataname`).

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
edopao: That is true. We can discuss it, I am not sure anymore about this code change.

edopao (commented):
Thanks for the review. We can discuss whether this is a good refactoring.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: What about a generic `LoweringContext`? Maybe `InStateLoweringContext` (though too long)?

philip-paul-mueller: How about `StateContext`, but it is only a little bit shorter though.

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
philip-paul-mueller: Okay, thanks.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
philip-paul-mueller: Sorry I missed the comma.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -243,7 +242,7 @@ class PrimitiveTranslator(Protocol):
     def __call__(
         self,
         node: gtir.Node,
-        ctx: gtir_sdfg.SDFGBuilderContext,
+        ctx: gtir_sdfg.InStateContext,
```
philip-paul-mueller: I like this name, I think you could even shorten it to `StateContext` without reducing readability.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -605,25 +598,23 @@ def translate_if(
     #           ==> |   head   | <==
     #               ------------
     #
-    cond_state = sdfg.add_state_before(state, state.label + "_cond")
-    sdfg.remove_edge(sdfg.out_edges(cond_state)[0])
+    cond_state = ctx.sdfg.add_state_before(ctx.state, ctx.state.label + "_cond")
+    ctx.sdfg.remove_edge(ctx.sdfg.out_edges(cond_state)[0])
```
philip-paul-mueller: First I was thinking "maybe add the `add_state_before()` method to the context" But then I realized that this would be violate the intention of a context that is centered around the state.
So I think it was a good call to nod add this method.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -605,25 +598,23 @@ def translate_if(
...
+    ctx.sdfg.add_edge(cond_state, false_state, dace.InterstateEdge(condition=(f"not ({if_stmt})")))
+    ctx.sdfg.add_edge(false_state, ctx.state, dace.InterstateEdge())
 
-    true_br_result = sdfg_builder.visit(
-        true_expr, ctx=gtir_sdfg.SDFGBuilderContext(sdfg=ctx.sdfg, state=true_state)
-    )
-    false_br_result = sdfg_builder.visit(
-        false_expr, ctx=gtir_sdfg.SDFGBuilderContext(sdfg=ctx.sdfg, state=false_state)
-    )
+    true_br_ctx = gtir_sdfg.InStateContext(sdfg=ctx.sdfg, state=true_state)
```
philip-paul-mueller: I am not sure if this idea is any good, but, here you explicitly construct the `{true, false}_state` and only then you construct the context.
Thus to promote the "conextualization" it might be worth changing the code to something like:
```python
true_br_ctx = gtir_sdfg.InStateContext(
    sdfg=ctx.sdfg,
    state=ctx.sdfg.add_state(
        ctx.state.label + "_true_branch"
    )
)
```
Because this way the state is only accessible through the context.
But as I said I am not sure if this is a good idea.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -691,28 +678,26 @@ def translate_index(
...
     index_write_tasklet = sdfg_builder.add_tasklet(
+        ctx,
         "index",
-        state,
         inputs={},
         outputs={"__val"},
         code=f"__val = {dim_index}",
     )
-    state.add_edge(
+    ctx.state.add_edge(
```
philip-paul-mueller: Maybe you should also add `add_edge()` to the methods of the context, I mean you can already add access nodes.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -721,65 +706,63 @@ def translate_index(
     )
 
     input_edges = [
-        gtir_dataflow.EmptyInputEdge(state, index_write_tasklet),
+        gtir_dataflow.EmptyInputEdge(ctx.state, index_write_tasklet),
```
philip-paul-mueller: Maybe for later, but I would suggest that these function also accepts the context directly.

src/gt4py/next/program_processors/runners/dace/gtir_builtin_translators.py
```diff
@@ -721,65 +706,63 @@ def translate_index(
...
         "get_value",
-        state,
         {},
         {"__out"},
         f"__out = {symbolic_expr}",
     )
-    temp_name, _ = sdfg.add_scalar(
-        temp_name or sdfg.temp_data_name(),
+    temp_name, _ = ctx.sdfg.add_scalar(
+        temp_name or ctx.sdfg.temp_data_name(),
```
philip-paul-mueller: Above I wrote that I think that `add_state_after()` should not be a method of the context, but for these two functions I am not fully sure, it might be worth, but maybe it is too much.

philip-paul-mueller (commented):
I think there are only a few small changes and we should be able to merge next time.

philip-paul-mueller (approved):
LGTM

PR #2016, opened 2025-05-01
feature[next]: field operator fusion argument deduplication

tehrengruber:
When a field operator is fused into another field operator some arguments may occur multiple times like `inp` in this example
```
as_fieldop(plus_stencil)(as_fieldop("deref")(inp), inp)
```
In order to create a readable IR and to allow common subexpression elimination inside the stencil these arguments should be deduplicated such that the resulting field operators reads
```
as_fieldop(λ(it) → ·it + ·it)(inp)
```
instead of 
```
as_fieldop(λ(it1, it2) → ·it1 + ·it2)(inp, inp)
```
While this has already been the case when the argument is a `itir.SymRef` the mechanism was complex and error prune. This PR simplifies the mechanism and also allows non `itir.SymRef` expressions to be deduplicated, e.g. something like `index(IDim)`. This greatly improves the readability of large field operator with many `concat_where` expressions after fusion.

### DIFF
--- a/src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
+++ b/src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
@@ -37,10 +37,10 @@ from gt4py.next.type_system import type_info, type_specifications as ts
 
 
 def _merge_arguments(
-    args1: dict[str, itir.Expr], arg2: dict[str, itir.Expr]
+    args1: dict[str, itir.Expr], args2: dict[str, itir.Expr]
 ) -> dict[str, itir.Expr]:
     new_args = {**args1}
-    for stencil_param, stencil_arg in arg2.items():
+    for stencil_param, stencil_arg in args2.items():
         assert stencil_param not in new_args
         new_args[stencil_param] = stencil_arg
     return new_args
--- a/src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
+++ b/src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
@@ -7,6 +7,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import annotations
 
+import collections
 import dataclasses
 import enum
 import functools
@@ -40,10 +41,8 @@ def _merge_arguments(
 ) -> dict[str, itir.Expr]:
     new_args = {**args1}
     for stencil_param, stencil_arg in arg2.items():
-        if stencil_param not in new_args:
-            new_args[stencil_param] = stencil_arg
-        else:
-            assert new_args[stencil_param] == stencil_arg
+        assert stencil_param not in new_args
+        new_args[stencil_param] = stencil_arg
     return new_args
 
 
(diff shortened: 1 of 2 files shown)


src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
```diff
@@ -40,10 +41,8 @@ def _merge_arguments(
 ) -> dict[str, itir.Expr]:
     new_args = {**args1}
     for stencil_param, stencil_arg in arg2.items():
-        if stencil_param not in new_args:
```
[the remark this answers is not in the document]
egparedes: Just a question about your comment: isn't `_prettify_as_fieldop_args` called after `merge_arguments`? How can then `_prettify_as_fieldop_args` handle pre-conditions for `merge_arguments`?

src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
```diff
@@ -40,10 +41,8 @@ def _merge_arguments(
 ) -> dict[str, itir.Expr]:
     new_args = {**args1}
```
egparedes, line 42: Probably unrelated to this PR, but why are they called `args1` and `arg2`? Shouldn't the second argument be called `args2`?

src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
```diff
@@ -95,6 +84,41 @@ def _inline_as_fieldop_arg(
...
+        new_args_inverted[arg].append(name)
+
+    new_args: dict[str, itir.Expr] = {}
+    new_stencil_body = stencil_body
+    for arg, names in new_args_inverted.items():
+        # put internal names at the end
+        names = sorted(names, key=lambda s: (s.startswith("_"), s))
+        new_args[names[0]] = arg
+        if len(names) > 1:
+            new_stencil_body = im.let(*((name, names[0]) for name in names[1:]))(new_stencil_body)
```
egparedes, line 101: I would unpack directly  on creation the head and tail of the sorted list (with better names if possible) 
```suggestion
        head, *tail = sorted(names, key=lambda s: (s.startswith("_"), s))
        new_args[head] = arg
        if tail:
            new_stencil_body = im.let(*((name, head) for name in tail))(new_stencil_body)
```

egparedes (commented):
Looks good to me. Just a couple of questions and suggestions.

src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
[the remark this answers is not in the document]
tehrengruber: The new mechanism works as follows:
- generate a unique `stencil_param`
- `_merge_arguments`
- `_deduplicate_as_fieldop_args`
- `_prettify_as_fieldop_args`
So here we still have unique params, whereas before the stencil_params were deduplicated before `_merge_arguments` and you had to do the special casing here.

src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
tehrengruber: Oh, absolutely.

tehrengruber (commented):
_

egparedes (approved):
LGTM

PR #1968, opened 2025-04-11
feat[next]: Precompile with static variants

havogt:
Adds the option to declare parameters as static (i.e. will be jitted) and the option to (pre-)compile for scalar arguments.

Additionally, add `enable_jit` to the program. If `True`, variants for the program will be compiled, if they don't exist yet. If `False` raises an error if the program doesn't exist, i.e. wasn't pre-compiled.

The static parameters are static for all program variants
E.g. if param `a` and `b` are declared static (either explicitly or implicitly by calling `compile`), then
- in the next call (with jit enabled) a new program will be compiled for the values of `a` and `b`
- compile can only be called with static arguments for `a` and `b` (not `a` or `b` only and not `c`)

PR #2017, opened 2025-05-05
fix[next]: Improve dead-code-elimination to avoid domain inference failure 

tehrengruber:
#1968 Introduced the possibility to specify (non-field) arguments at compile time. The resulting IR in some cases contained expressions similar to:
```
let field = as_fieldop(...)
  if_ True then
    val1
  else
    field
end
```
If constant folding is executed on such an expression
```
let field = as_fieldop(...)
  val1
end
```
`field` will never be used and the domain inference fails. This PR introduces a new transformation `dead_code_elimination` based upon `InlineLambdas`, `ConstantFolding`, `CollapseTuple` that properly handles such cases.

### DIFF
--- a/src/gt4py/next/iterator/transforms/dead_code_elimination.py
+++ b/src/gt4py/next/iterator/transforms/dead_code_elimination.py
@@ -22,15 +22,16 @@ def dead_code_elimination(
     collapse_tuple_uids: Optional[eve_utils.UIDGenerator] = None,
     offset_provider_type: common.OffsetProviderType,
 ) -> itir.Program:
-    # remove dead-code, e.g. `let var=True in if_(var, val1, val2) end`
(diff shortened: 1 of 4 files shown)


src/gt4py/next/iterator/transforms/dead_code_elimination.py
egparedes, line 1: What about unit tests for this?

egparedes (commented):
It looks good, except for the lack of unit tests and a couple of minor suggestions.

src/gt4py/next/iterator/transforms/dead_code_elimination.py
tehrengruber: I've added a small test.

egparedes (commented):
One last final question and it'd be ready to go.

PR #2018, opened 2025-05-07
fix[next]: Fix cse extraction of lambda functions

tehrengruber:
Fix a small bug introduced in #2013, where `f` in the following expression
```
let val = 1 in
  let f = λ() → val in
     opaque_fun(f, g)
  end
end
```
was erroneously extracted beyond the definition of `val`.


egparedes (approved):
LGTM

PR #2019, opened 2025-05-07
feat[next]: Cache type synthesizer invocation

tehrengruber:
The ITIR type inference is generally fast. In some narrow cases encountered in icon4py where a lambda function is called many times the runtime increases significantly since the return type is synthesized many types even though it never changes (and can never change as all lambda invocations need to have the same type by specification of the IR). This PR restores linear runtime in the length of the expression by caching the return type synthesis.

This was encountered in icon4py in [`prepare_ffsl_flux_area_patches_list`](https://github.com/C2SM/icon4py/blob/82dd7cf9af5aa8506e20cb8bbafb6880f1b55ca2/model/atmosphere/advection/src/icon4py/model/atmosphere/advection/stencils/prepare_ffsl_flux_area_patches_list.py#L600) after field operator fusion. The lambdas that are evaluated repeatedly are created by `InlineCenterDerefLiftVars`.

### DIFF
--- a/src/gt4py/eve/utils.py
+++ b/src/gt4py/eve/utils.py
@@ -408,7 +408,7 @@ hashable_by_id = hashable_by(id)
 def lru_cache(
     func: Optional[Callable[_P, _T]] = None,
     *,
-    key: Optional[Callable[_P, Any]] = None,
+    key: Optional[Callable[_P, int]] = None,
     maxsize: Optional[int] = 128,
     typed: bool = False,
 ) -> Union[Callable[_P, _T], Callable[[Callable[_P, _T]], Callable[_P, _T]]]:
--- a/src/gt4py/eve/utils.py
+++ b/src/gt4py/eve/utils.py
@@ -414,10 +414,7 @@ def lru_cache(  # redefinition of unused function
 
     Be careful: `key(obj1) == key(obj2)` must imply `obj1 == obj2`.
 
-    >>> def key_fn(x):
-    ...     return id(x)
-
-    >>> @lru_cache(key=key_fn)
+    >>> @lru_cache(key=id)
     ... def func(x):
     ...     print("called")
     ...     return x
--- a/src/gt4py/eve/utils.py
+++ b/src/gt4py/eve/utils.py
@@ -365,7 +365,7 @@ def optional_lru_cache(  # redefinition of unused function
 class HashableBy(Generic[_T]):
     __slots__ = ["func", "value"]
 
-    def __init__(self, func: Callable[[_T], int], value: _T):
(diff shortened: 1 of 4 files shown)


src/gt4py/eve/utils.py
```diff
@@ -402,6 +402,48 @@ def hashable_by(
...
+        def cached_func(*args: HashableBy, **kwargs: HashableBy) -> _T:
+            return func(*(arg.value for arg in args), **{k: v.value for k, v in kwargs.items()})
+
+        @functools.wraps(func)
+        def inner(*args, **kwargs):  # type: ignore[no-untyped-def]  # cast below restores type info
+            return cached_func(
+                *(hashable_by(key, arg) for arg in args),
+                **{k: hashable_by(key, arg) for k, arg in kwargs.items()},
+            )
+
```
egparedes, line 441: I think this is also needed to honor the `cache_parameters()` instrumentation function (https://devdocs.io/python~3.13/library/functools#functools.lru_cache).
```suggestion
        inner.cache_parameters = cached_func.cache_parameters
```

tests/eve_tests/unit_tests/test_utils.py
```diff
@@ -369,3 +369,21 @@ def test_xenumerate():
...
+
+    def func(x):
+        nonlocal call_count
+        call_count += 1
+        return x
+
+    cached = lru_cache(func, key=id)
+
+    obj = object()
+    assert cached(obj) is obj
```
egparedes, line 387: You could test the same using the `cache_parameters()` instrumentation function. Please check also the `__wrapped__` attribute is set properly.

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -20,6 +22,12 @@
 from gt4py.next.utils import tree_map
 
 
+def _type_synth_arg_cache_key(type_or_synth: TypeOrTypeSynthesizer) -> int:
+    if isinstance(type_or_synth, TypeSynthesizer):
+        return id(type_or_synth)
+    return hash(eve_utils.content_hash(type_or_synth))
```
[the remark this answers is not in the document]
tehrengruber: This is the default case and happens most of the time. All types support `content_hash`. They should also support `__hash__`, but currently don't since some of them contain `list`s and `dict`s.

tests/eve_tests/unit_tests/test_utils.py
tehrengruber: Testing both now.

egparedes (approved):
LGTM

PR #2020, opened 2025-05-07
feat[dace][next]: Modified How the GPU Block Size Is Set

philip-paul-mueller:
Before all GPU Maps had the same block size, although it was possible to select different ones depending on the number of dimensions of a Map, i.e. selecting `(128, 1, 1)` for Maps with just one parameter but `(32, 8, 1)` for Maps with 2 parameters.
This PR allows that the block size of Maps is adjusted downwards, if it is known at code generation time, that the number of iterations is smaller.

For an example assume that the block size is `(32, 8, 1)` and the Map, for which we want to set the block size, only performs 4 iterations in the `y` dimension, which is a common pattern with `concat_where` applied to the vertical dimension.
Before we would have set the block size of that map to `(32, 8, 1)`, which means that 50% of the threads would do nothing.
With this PR we will use a block size of `(32, 4, 1)`.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
@@ -63,6 +63,8 @@ def gt_gpu_transformation(
     Notes:
         - In addition it might fuse Maps together that should not be fused. To prevent
             that you should set `try_removing_trivial_maps` to `False`.
+        - The function assumes that the iteration order has been set correctly before
+            it is called.
 
     Todo:
         - Currently only one block size for all maps is given, add more options.
@@ -343,6 +345,10 @@ def gt_set_gpu_blocksize(
         launch_bounds: The value for the launch bound that should be used.
         launch_factor: If no `launch_bounds` was given use the number of threads
             in a block multiplied by this number.
+
+    Note:
+        If a Map is found whose range is smaller than the specified block size for
+        that dimension then the block size for that map is reduced.
     """
     for dim in [1, 2, 3]:
         for arg, val in {
@@ -448,6 +454,12 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
     three dimensional). If no value is specified then the block size `(32, 1, 1)`
     will be used an no launch bound will be be emitted.
 
+    If a Map is found whose iteration size in a particular dimension is smaller than
+    the block size, the block size for that dimension is reduced to that value. For
+    example, the block size for 2D Maps is `(32, 8, 1)` and a Map that only performs
+    four iteration in the second dimension, will get a block size of `(32, 4, 1)`.
+    Note that this modification will not influence the launch bound value.
+
     Args:
         block_size_Xd: The size of a thread block on the GPU for `X` dimensional maps.
         launch_bounds_Xd: The value for the launch bound that should be used for `X`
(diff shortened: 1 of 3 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -569,16 +582,37 @@ def apply(
     ) -> None:
         """Modify the map as requested."""
         gpu_map: dace_nodes.Map = self.map_entry.map
-        if len(gpu_map.params) == 1:
-            block_size = self.block_size_1d
+        map_params: list[str] = gpu_map.params
+        map_size = gpu_map.range.size()
+        map_dim = len(map_params)
```
edopao, line 587:
```suggestion
        num_map_params = len(gpu_map.params)
```

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -569,16 +582,37 @@ def apply(
...
+
+        # Because of a particularity of the DaCe code generator, the iteration
+        #  variable that is associated to the `x` dimension of the block is the
+        #  last parameter, i.e. `map_params[-1]`. The one for `y` the second last.
```
edopao, line 591:
```suggestion
        #  last parameter, i.e. `gpu_map.params[-1]`. The one for `y` the second last.
```

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -569,16 +582,37 @@ def apply(
...
+        if map_dim == 1:
+            block_size = list(self.block_size_1d)
             launch_bounds = self.launch_bounds_1d
-        elif len(gpu_map.params) == 2:
-            block_size = self.block_size_2d
+            inspect_dims = 1
```
edopao, line 595:
```suggestion
            num_block_dims = 1
```

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -569,16 +582,37 @@ def apply(
...
+            block_size = list(self.block_size_3d)
             launch_bounds = self.launch_bounds_3d
-        gpu_map.gpu_block_size = block_size
+            # If there are more than three dimensions DaCe will condense them into
+            #  the `z` dimension of the block, so we have to ignore the `z` dimension,
+            #  when we modify the block sizes.
+            inspect_dims = 3 if map_dim == 3 else 2
+
+        # Cut down the block size.
+        # TODO(phimuell): Think if it is useful to also modify the launch bounds.
```
edopao, line 609: I guess so, if `launch_bounds` is not `None` (for whatever reason) it should also be updated.

philip-paul-mueller: I am not sure what is the best way, should we apply some proportional scaling?
Since we do not use it, I think that the best approach for now is to just keep it and not modify it.

edopao: It took some time because I had to read the documentation. You are right, we probably should not change the `maxThreadsPerBlock` (the first element of the `__launch_bounds__` tuple) passed by the user.

edopao (approved):

PR #2022, opened 2025-05-08
fix[dace][next]: Updated How Dataflow Inside Maps is Optimized

philip-paul-mueller:
This PR changes how Maps inside Maps are handled. Before they were fused, however, this created dependency between otherwise unrelated dataflow. This might or will block other transformations such as `LoopBlocking` or `MoveDataflowIntoIfBody`.

What might be worth considering is that serial Map fusion should be enabled.


src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -328,27 +337,6 @@ def _gt_auto_process_dataflow_inside_maps(
         validate_all=validate_all,
     )
 
-    fusion_transformation = gtx_transformations.MapFusion(
```
edopao, line 331: Maybe I would keep serial fusion, but just calling `apply_transformations_repeated` without the while loop.

philip-paul-mueller: There are also some cases where it will be a blocker, they are rarer, but it is not really needed anyway.
I checked the assembly output.

edopao (approved):
LGTM

PR #2024, opened 2025-05-08
feat[dace][next]: Updated When `LoopBlocking` Applies

philip-paul-mueller:
Modifies when `LoopBlocking` is applied.
The transformation now checks if the number of iteration the Map performs is below the blocking size if this is the case the transformation does not apply.

Furthermore, the transformation will now apply only if there are independent nodes.
It was possible to select that mode before, but it was disabled by default and now it is enabled.
In addition this PR also changes what it means that "there are independent nodes".
Before all nodes were equal, but now Taskles, with empty Memlets (and their AccessNodes) are now longer counted when deciding if the transformation should apply or not.

NOTE:
The whole selection of the blocksize should be reworked, i.e. it should be possible to apply a blocking size that is less.
However, this is also highly related to how the thread block size.


src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -698,3 +710,43 @@ def _rewire_map_scope(
...
         # TODO(phimuell): Use a less expensive method.
         dace.sdfg.propagation.propagate_memlets_state(sdfg, state)
+
+    def _test_if_blocking_is_favourable(
+        self,
+        state: dace.SDFGState,
+    ) -> bool:
+        """Test if the nodes are really independent nodes.
+
+        After the classification the function will examine the set to see if some
```
edopao, line 720: I would suggest to test if we can achieve the same through `scalar_to_symbol` promotion: it should probably be able to remove such scalars and tasklets if the value is converted to a symbol. The `scalar_to_symbol` pass is supposed to work now, and we have talked about re-enabling it.

philip-paul-mueller: Yesterday's DaCe meeting was again, ScalarToSymbol has some problems, so I am against to use it.
We do not really use symbols, so I am also against opening a new bag of problems.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -698,3 +710,43 @@ def _rewire_map_scope(
         _ = sdfg.hash_sdfg()
         # TODO(phimuell): Use a less expensive method.
         dace.sdfg.propagation.propagate_memlets_state(sdfg, state)
+
+    def _test_if_blocking_is_favourable(
```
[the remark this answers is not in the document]
philip-paul-mueller: You are right the name is better.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
edopao: Agreed.

edopao (commented):
LGTM

edopao (approved):

PR #2025, opened 2025-05-08
fix[next][dace]: Use SDFG hash in dace cache for orchestration tests

edopao:
This PR fixes the setting for dace cache in the dace orchestration tests. These tests use the `dace.program` annotation and therefore use dace as a frontend, and rely on the dace cache (not the gt4py cache) for the build of the top-level SDFG. The default setting of the dace cache is to create a folder with the SDFG name (not a good option since we have multiple variants of the same test by means of a fixture). The option `unique` was partially used, and that works, but it was missing on the first call to SDFG compile.

Additional changes to avoid the timeout of the CI pipeline:
- Skip tests with `DACE_GPU_NO_OPT` backend
- Increases the number of pytest workers from 16 to 32


egparedes (approved):
LGTM. Thanks for the fix!

PR #2026, opened 2025-05-09
fix[dace][next]: Order of Map Fusion in `auto_optimizer()`

philip-paul-mueller:
The match for parallel Map fusion is very general as it matches any two maps in the SDFG. To reduce the search space this PR changes auto optimizer in such a way that first serial map fusion is performed on its own. After that has ended, serial and parallel Map fusion are performed as before.

The PR also updates the `MapFusion{Serial, Parallel}` wrappers. They now override the `expression()` method, thus serial Map fusion will now no longer implicitly match parallel Maps, which should speed up things.

The implementation is pretty hacky.
However, DaCe will move inside this direction anyway, so they are just a temporary fix.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_dace.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_dace.py
@@ -239,26 +239,34 @@ class MapFusion(transformation.SingleStateTransformation):
         assert expr_index == self.expr_index
         assert self.expr_index in [0, 1], f"Found invalid 'expr_index' {self.expr_index}"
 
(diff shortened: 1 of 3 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_dace.py
```diff
@@ -222,7 +222,7 @@ def expressions(cls) -> Any:
 
     def can_be_applied(
         self,
-        graph: Union[SDFGState, SDFG],
+        graph: Union[dace.SDFGState, SDFG],
```
edopao, line 225: Style: you can use the module prefix on both.

edopao (approved):

PR #2027, opened 2025-05-09
fix[dace][next]: Fixed a Bug in `LoopBlocking`

philip-paul-mueller:
This PR modifies the `LoopBlocking` in three different ways.

First it modifies how Maps are handled.
Maps are handled specially as they are not incrementally checked for independence but as a whole.
So if a MapEntry is encountered the whole Map scope is added.
Before it was checked that the incoming edges where independent and the loop body did not contains the blocking variable.
However, the output edges were not checked.
The problem was that not all Memlets in a Memlet path needs to refer to the same data, if they cross a scope node, which is what happened here.
Thus we have to explicitly check the that.
Because if we would not do that, the destination in which the Map writes to will be dependent (might be possible though).

The second changes adds a post processing to the set of independent nodes.
This is needed because in some cases it is possible that a Tasklet becomes a boundary node.
This happens if a tasklet writes into an AccessNode (see also later) that is directly connected to the MapExit, in which case the AccessNode is not independent.
If the Tasklet is independent then it will be a boundary, but this is not possible because there is nothing outside that could store/cache its output.
Thus we remove these kind of nodes.

The third modification is related to the second change.
If that AccessNode happens to be a scalar then it is handled.
Between the AccessNode and the MapExit a copy Tasklet is implicitly inserted, this prevent that the inner Map is empty, which is the whole reason for this "an independent node can not be connected to the outer MapExit node" anyway.
However, we only do it for scalars, the reason is, to ensure that no dynamic allocation happens and we do not copy too much around.

Notes:
- Only merge after Edoardo has confirmed that the bug has checked that the bug is really solved.
- Only merge after [PR#2024](https://github.com/GridTools/gt4py/pull/2024) has been merged.


src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -515,6 +543,57 @@ def _classify_node(
...
+
+        This function is mostly there to handle cases that are not implemented.
+        This function might remove, nodes from the set of independent nodes.
+        """
+        assert self._independent_nodes is not None  # silence MyPy
+        outer_entry: dace_nodes.MapEntry = self.outer_entry  # for caching.
+
+        for node in list(self._independent_nodes):
+            # The only nodes that are important here, are the independent nodes at
+            #  the boundaries, i.e. whose outgoing edges, will be split.
```
edopao, line 560:
```suggestion
            #  the boundaries, i.e. whose outgoing edges will be split.
```

edopao (approved):
The fix is working.

PR #2033, opened 2025-05-15
fix/build[dace][next]: Fixing Expanding of Empty Memlets

philip-paul-mueller:
Memlets that can not be expressed as a `cudaMemcpy*()` call are translated into Maps.
For that are Memlets are scanned, however, if there is an empty Memlet between two AccessNodes there is an error, because the subsets are not defined.
Thus we have to skip those Memlets.

Note that this is essentially the same as in [DaCe Pr#2006](https://github.com/spcl/dace/pull/2006) and therefore we also have to update the DaCe dependency.


edopao:
I see that the dace PR was approved. I propose we wait for it to be merged and then we update the dace version.

edopao (approved):

PR #2035, opened 2025-05-16
feat[next][dace]: Use async sdfg call

edopao:


src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -64,6 +64,32 @@ def _language_settings() -> languages.LanguageSettings:
...
+
+    has_gpu_code = any(getattr(node, "schedule", False) for node, _ in sdfg.all_nodes_recursive())
+
+    if not has_gpu_code:
+        # The async execution mode requires CUDA, therefore it is only available on GPU
+        return
+
+    # Emit cpp code for the SDFG to set the cuda streams
+    sdfg.append_global_code(f"""\
+DACE_EXPORTED bool __dace_gpu_set_stream({sdfg.name}_state_t *__state, int streamid, gpuStream_t stream);
```
philip-paul-mueller, line 78: The suffix `_state_t` is not necessarily correct, see [here](https://github.com/spcl/dace/pull/1413).

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -64,6 +64,32 @@ def _language_settings() -> languages.LanguageSettings:
     return languages.LanguageSettings(formatter_key="", formatter_style="", file_extension="sdfg")
 
 
+def _make_sdfg_async(sdfg: dace.SDFG) -> None:
+    """Sets all cuda stream to the default stream."""
```
philip-paul-mueller, line 68: You should also describe what this function is doing.
For example you should mention that this will turn off all cuda syncronization, also between stage transitions, but this is not that bad because everything is running on one stream, however, see also note below.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -64,6 +64,32 @@ def _language_settings() -> languages.LanguageSettings:
...
+    __dace_gpu_set_stream(__state, i, stream);
+}}\
+""")
+    sdfg.append_init_code(f"""\
+__set_stream_{sdfg.name}(__state, cudaStreamDefault);
+""")
+
+    # Loop over all state in the top-level SDFG
+    for state in sdfg.states():
+        state.nosync = True
```
philip-paul-mueller, line 90: There is something that worries me:
Consider the following case a kernel is computing something, that runs asynchronous.
The result it computes, lets say a scalar otherwise some random array index (maybe `highest_horizontal_velocity[nflatlev]`) is accessed, because it is needed for an interstage edge.
So you will launch the kernel and then initiate the copy to host.
The copy will not start because they are scheduled on the same stream.
However, who will stop the CPU to make sure that the copy has concluded and it can actually check and access the value it copied from device to host?

Normally, I would say that the the sync will take care, because inter state edge transition is after the state?
We have to check that.

Looking at our SDFG (and DaCe's inability to return scalars) I think it should be fine to set this flag only at the very last state, but it must be a true terminal state without any outgoing edges.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -64,6 +64,32 @@ def _language_settings() -> languages.LanguageSettings:
...
+    # Emit cpp code for the SDFG to set the cuda streams
+    sdfg.append_global_code(f"""\
+DACE_EXPORTED bool __dace_gpu_set_stream({sdfg.name}_state_t *__state, int streamid, gpuStream_t stream);
+DACE_EXPORTED void __set_stream_{sdfg.name}({sdfg.name}_state_t *__state, gpuStream_t stream) {{
+for (int i = 0; i < __state->gpu_context->num_streams; i++)
+    __dace_gpu_set_stream(__state, i, stream);
+}}\
+""")
+    sdfg.append_init_code(f"""\
+__set_stream_{sdfg.name}(__state, cudaStreamDefault);
```
philip-paul-mueller, line 85: I do not see the point in generating the `__set_stream_()` function, instead I would generate the `for` loop directly inside the init function.

philip-paul-mueller (commented):
I think that the general approach is okay, but there is some point that worries me, especially setting `no_sync` on all states.

philip-paul-mueller (commented):
There are some smallish points.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -60,8 +61,61 @@ def _find_constant_symbols(
...
+
+    has_gpu_code = any(getattr(node, "schedule", False) for node, _ in sdfg.all_nodes_recursive())
+
+    if not has_gpu_code:
+        # The async execution mode requires CUDA, therefore it is only available on GPU
+        return
+
+    # Emit cpp code for the SDFG to set the cuda streams
+    dace_state_type_name = dace_codegen_cpp.mangle_dace_state_struct_name(sdfg)
+    sdfg.append_global_code(f"""\
```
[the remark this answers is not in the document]
philip-paul-mueller: Here is the [DaCe PR](https://github.com/spcl/dace/pull/2013) that addresses that issue.
I will update the fork as well.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -60,8 +60,63 @@ def _find_constant_symbols(
...
+
+    if not has_gpu_code:
+        # The async execution mode requires CUDA, therefore it is only available on GPU
+        return
+
+    # Loop over all states in the top-level SDFG
+    is_nosync_used = False
+    for state in sdfg.states():
+        if state.parent_graph is not sdfg:
+            # We ignore states that are used inside 'ControlFlowRegion' nodes
```
philip-paul-mueller, line 95: Why can we do that?
The rest looks good.

edopao: We actually have to do that. The states that are located inside a `ConditionalBlock` or a `LoopRegion` cannot be inspected by `sdfg.out_edges(state)`. You would get an error:
```
/icon4py/.nox/test_model-3-10-stencils-diffusion/lib/python3.10/site-packages/gt4py/next/program_processors/runners/dace/workflow/translation.py:94: in make_sdfg_async
    for oedge in sdfg.out_edges(state):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = SDFG (apply_diffusion_to_w_and_compute_horizontal_gradients_for_turbulence)
node = SDFGState (stmt_0_true_branch)
    def out_edges(self, node: NodeT) -> List[Edge[EdgeT]]:
>       return list(self._nodes[node][1].values())
E       KeyError: SDFGState (stmt_0_true_branch)
```

Also, I think it is safer to not use async inside `ControlFlow` regions.

philip-paul-mueller: I am aware of that and it caused much headache in the beginning of the year when we did the switch.
However, in that case I would say that we have to inspect the outgoing edges of the `ControlFlow` regions.
I think in general we do not use this nesting feature in a very extreme case, but it might happens for `if` blocks.

philip-paul-mueller (approved):
LGTM

PR #2038, opened 2025-05-20
feat[next]: Faster inline lambda

tehrengruber:
So far just a spontaneous idea I tried. Tried this with the large muphys field operator where it was roughly 10% faster. Real benefit would only be there if we cache the `collect_symbol_refs` in the annex (which was not possible with the previous approach).

### DIFF
--- a/src/gt4py/next/iterator/transforms/inline_lambdas.py
+++ b/src/gt4py/next/iterator/transforms/inline_lambdas.py
@@ -11,9 +11,10 @@ from typing import Optional
 
 from gt4py.eve import NodeTranslator, PreserveLocationVisitor
 from gt4py.next.iterator import ir
+from gt4py.next.iterator.ir_utils import ir_makers as im
 from gt4py.next.iterator.ir_utils.common_pattern_matcher import is_applied_lift
-from gt4py.next.iterator.transforms.remap_symbols import RemapSymbolRefs, RenameSymbols
-from gt4py.next.iterator.transforms.symbol_ref_utils import CountSymbolRefs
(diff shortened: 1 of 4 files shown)


src/gt4py/next/iterator/transforms/inline_lambdas.py
```diff
@@ -63,53 +66,24 @@ def inline_lambda(  # see todo above
...
-            while name in refs or name in syms or name in name_map.values():
-                name += "_"
-            return name
-
-        for sym in clashes:
-            name_map[sym] = new_name(sym)
-
-        expr = RenameSymbols().visit(expr, name_map=name_map)
-
     symbol_map = {
```
havogt, line 69: Just leaving a remark here, that there was #2134. Maybe it's already resolved here.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_inline_lambdas.py
```diff
@@ -40,6 +40,19 @@
...
         im.multiplies_(im.plus(2, 1), im.plus("x", "x")),
     ),
+    (
+        "name_shadowing_external",
+        im.call(im.lambda_("x")(im.lambda_("y")(im.plus("x", "y"))))(im.plus("x", "y")),
+        im.lambda_("y_")(im.plus(im.plus("x", "y"), "y_")),
+    ),
+    (
+        "renaming_collision",
+        # the `y` param of the lambda may not be renamed to `y_` as this name is already referenced
```
havogt, line 50: Hmm. Is this the same problem as in #2134

PR #2044, opened 2025-05-22
feat[next]: Hash function for offset provider

edopao:


tehrengruber (approved):

src/gt4py/next/ffront/decorator.py
```diff
@@ -274,6 +279,17 @@ def _compiled_programs(self) -> compiled_program.CompiledProgramsPool:
...
         )
 
+    def _offset_provider_extended(
+        self,
+        offset_provider: common.OffsetProvider,
+    ) -> common.OffsetProvider:
+        try:
+            op_extended = self._offset_provider_extended_cache[offset_provider]
+        except KeyError:
+            op_extended = {**self._implicit_offset_provider, **offset_provider}
```
egparedes:
```suggestion
            op_extended = self._implicit_offset_provider | 
            offset_provider
```

src/gt4py/next/common.py
```diff
@@ -1035,6 +1035,16 @@ def offset_provider_to_type(
     }
 
 
+def offset_provider_hash(offset_provider: OffsetProvider) -> int:
```
egparedes: Just a suggestion, but in my opinion a more specific name could help here, since this is more like a workaround than the proper way of hashing offset providers 
```suggestion
def key_value_id_hash(mapping: Mapping[Hashable, Any]) -> int:
```
or 
```suggestion
def hash_by_key_and_value_id(mapping: Mapping[Hashable, Any]) -> int:
```

src/gt4py/next/ffront/decorator.py
```diff
@@ -274,6 +279,17 @@ def _compiled_programs(self) -> compiled_program.CompiledProgramsPool:
             static_params=self.static_params,
         )
 
+    def _offset_provider_extended(
```
egparedes:
```suggestion
    def _extend_offset_provider(
```

src/gt4py/next/ffront/decorator.py
```diff
@@ -87,6 +87,11 @@ class Program:
         Sequence[str] | None
     )  # if the user requests static params, they will be used later to initialize CompiledPrograms
 
+    _offset_provider_extended_cache: eve_utils.CustomMapping = dataclasses.field(
+        default_factory=lambda: eve_utils.CustomMapping(common.offset_provider_hash),
+        kw_only=True,
+    )  # cache the extended dictionary of offset providers that includes the implicitly defined ones
```
egparedes: Just a suggestion, but I think this way is more clear that this is just an implementation detail.
```suggestion
    _offset_provider_extended_cache: eve_utils.CustomMapping = dataclasses.field(
        default_factory=lambda: eve_utils.CustomMapping(common.offset_provider_hash),
        kw_only=True,
        init=False
    )  # cache the extended dictionary of offset providers that includes the implicitly defined ones
```

src/gt4py/next/otf/compiled_program.py
```diff
@@ -136,6 +136,10 @@ class CompiledProgramsPool:
         stages.CompiledProgram | concurrent.futures.Future[stages.CompiledProgram],
     ] = dataclasses.field(default_factory=dict)
```
egparedes: Shouldn't `_compiled_programs` be also converted to a `CustomMapping`? That would also make code a bit simpler since `_CompiledProgramsKey` could be just a type alias, right?
```python
CompiledProgramsKey: TypeAlias = tuple[tuple[ScalarOrTupleOfScalars, ...], common.OffsetProviderType]
```

egparedes (commented):
Looks good but I have some suggestions to make it slightly cleaner.

src/gt4py/next/common.py
edopao: I would prefer to keep this method specific for `OffsetProvider` type, if we keep the definition in the current module, close to other offset provider definitions. Otherwise I can follow your suggestion and move it close to the `CustomMapping` definition, in `eve.utils` module.

src/gt4py/next/ffront/decorator.py
edopao: Good idea!

src/gt4py/next/common.py
egparedes: Ok, makes sense. What about using some suffix like `hash_offset_provider_unsafe` or `hash_offset_provider_by_value_id` ... ?

edopao: Good idea.

src/gt4py/next/ffront/decorator.py
edopao: I get a type error:
`error: No overload variant of "__or__" of "dict" matches argument type "Mapping[str, Dimension | NeighborConnectivity]"  [operator]`

src/gt4py/next/otf/compiled_program.py
edopao: Good idea.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -131,10 +129,15 @@ class CompiledProgramsPool:
 
     # TODO(havogt): This dict could be replaced by a `functools.cache`d method
     # and appropriate hashing of the arguments.
```
egparedes: Remove TODO

src/gt4py/next/otf/compiled_program.py
```diff
@@ -131,10 +129,15 @@ class CompiledProgramsPool:
...
-        stages.CompiledProgram | concurrent.futures.Future[stages.CompiledProgram],
-    ] = dataclasses.field(default_factory=dict)
+    _compiled_programs: eve_utils.CustomMapping = dataclasses.field(
+        default_factory=lambda: eve_utils.CustomMapping(_hash_compiled_program_unsafe),
+        kw_only=True,
+        init=False,
+    )
+
+    _offset_provider_type_cache: eve_utils.CustomMapping = dataclasses.field(
+        default_factory=lambda: eve_utils.CustomMapping(common.hash_offset_provider_unsafe)
```
egparedes: For consistency with the previous attribute, I would add:
```
init=False
```

PR #2045, opened 2025-05-22
refactor[cartesian|next] Move imports of implementation down one level

FlorianDeconinck:
## Description

As it stands now we have `gt4py.cartesian` and `gt4py.next `default importing themselves into the top level __init__.py . Now it isn't an issue per se, but it's a double import of a lot of files that ultimately will rarely be used in concert.

This proposes to remove the "all" import and rely on users to actually call `gt4py.next` and `gt4py.cartesian` which I expect is the common case anyway.

## Requirements

### DIFF
--- a/src/gt4py/__init__.py
+++ b/src/gt4py/__init__.py
@@ -6,11 +6,14 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
-"""Python library for generating high-performance implementations of stencil kernels for weather and climate modeling."""
+"""Python library for generating high-performance implementations of stencil kernels for weather and climate modeling.
 
-import sys as _sys
+The library carries to implementations available as submodule imports:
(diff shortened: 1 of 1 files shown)


src/gt4py/__init__.py
```diff
@@ -6,11 +6,14 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
-"""Python library for generating high-performance implementations of stencil kernels for weather and climate modeling."""
+"""Python library for generating high-performance implementations of stencil kernels for weather and climate modeling.
 
-import sys as _sys
+The library carries to implementations available as submodule imports:
```
egparedes, line 11: Maybe?
```suggestion
The library carries two implementations available as submodule imports:
```

egparedes (approved):
Just a typo.

src/gt4py/__init__.py
FlorianDeconinck: Well that's embarrassing...

PR #2046, opened 2025-05-22
Refactor embedded context

egparedes:
## Requirements

### DIFF
--- a/tests/next_tests/integration_tests/cases.py
+++ b/tests/next_tests/integration_tests/cases.py
@@ -493,6 +493,21 @@ def cartesian_case(
 
 
 @pytest.fixture
+def cartesian_case_no_backend():
+    yield Case(
+        backend=None,
+        offset_provider={
+            "Ioff": IDim,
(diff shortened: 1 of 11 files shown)


tests/next_tests/integration_tests/cases.py
```diff
@@ -492,6 +492,21 @@ def cartesian_case(
     )
 
 
+@pytest.fixture
+def cartesian_case_no_backend():
```
tehrengruber, line 496: The `cartesian_case` fixture could use this one to avoid code duplication.

tehrengruber (approved):

tests/next_tests/integration_tests/cases.py
egparedes: I still like it better this way since the original fixture is parametrized with all the backends and here we only want to run it once and using a custom backend.

egparedes: Refactored as discussed offline.

PR #2047, opened 2025-05-23
feat[next][dace]: Use dace cache for compiled programs

edopao:
The workflow step with the persistent `FileCache` store is translating from `CompilableProgram` (ITIR.Program + CompileTimeArgs) to `ProgramSource`, so this step is storing in cache only the result of the SDFG transformations, not the compiled program binary.
With this PR, we rely on dace cache to avoid recompiling the SDFG every time the application is launched.


src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -121,7 +121,13 @@ def __call__(
...
         with dace.config.temporary_config():
             dace.config.Config.set("compiler.build_type", value=self.cmake_build_type.value)
-            dace.config.Config.set("compiler.use_cache", value=False)  # we use the gt4py cache
+
+            # We rely on dace cache to avoid recompiling the SDFG.
+            #   Note that the workflow step with the persistent `FileCache` store
+            #   is translating from `CompilableProgram` (ITIR.Program + CompileTimeArgs)
+            #   to `ProgramSource`, so this step is storing in cache only the result
+            #   of the SDFG transformations, not the compiled program binary.
+            dace.config.Config.set("compiler.use_cache", value=True)
```
philip-paul-mueller, line 130: I would also set the location of the DaCe cache such that it is inside the the gt4py cache and not somewhere else, where it might be wiped.
In that case we should also give it a better name, i.e. not a hidden folder.


By default the cache uses only the name to identify the SDFG, but maybe we should switch to the hash method, although it can lead to [false negative](https://github.com/spcl/dace/issues/1946).
Although, I would like to use the cheap name scheme in release and hash during debug, I think this is not a good idea, because it might lead to hard to find bugs.

philip-paul-mueller (commented):
See my comment about the location of the DaCe cache, that should be changed and my suggestion about going to the hash scheme.
Furthermore, we should wait until [PR#2052](https://github.com/GridTools/gt4py/pull/2052) has been merged.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
edopao: Agreed.

PR #2049, opened 2025-05-23
bug[next]: fix tuple output in direct field operator call with domain argument

nfarabullini:
`out` can be a single or multiple, these different scenarios are now handled by the lambda function


tehrengruber:
This test case could be easily adopted: https://github.com/GridTools/gt4py/blob/main/tests/next_tests/integration_tests/feature_tests/ffront_tests/test_arg_call_interface.py#L302

PR #2052, opened 2025-05-26
refactor/test[dace][next]: Better Testing Infrastructure.

philip-paul-mueller:
This PR updates the set of configuration variables that are always set (by an autouse fixture).
It now also sets `use_cache` to `False`, the main issue is that some tests compile the SDFG multiple times.
It also adds a function that simplifies compilation.
It will give the SDFG a unique name which will avoid the warning that an SDFG was already loaded a and a different name was selected (this will mostly clean the log).

This PR should be merged after [PR#2050](https://github.com/GridTools/gt4py/pull/2050).


edopao (approved):

edopao:
I've tested this change in my PR, it works. Thanks for looking into this dace issue.

PR #2054, opened 2025-05-27
fix[next]: fix nested tuple comparisons in 'cases' utilities

egparedes:
Fix bug in test `cases` utilities introduced in PR #2049 when verifying structs with multiple nested levels of ndarrays.

### DIFF
--- a/tests/next_tests/integration_tests/cases.py
+++ b/tests/next_tests/integration_tests/cases.py
@@ -384,7 +384,8 @@ def tree_mapped_np_all_close(ref: np.ndarray, data: np.ndarray) -> bool:
     """Compare two arrays or nested tuples of arrays using np.allclose."""
     if (is_tuple := isinstance(ref, tuple)) == isinstance(data, tuple):
         if is_tuple:
-            return all(gt_utils.tree_map(np.allclose)(ref, data))
+            all_closed_results = gt_utils.tree_map(np.allclose)(ref, data)
+            return all(gt_utils.flatten_nested_tuple(all_closed_results))
         else:
             return np.allclose(ref, data)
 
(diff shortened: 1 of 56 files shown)


tests/next_tests/integration_tests/cases.py
```diff
@@ -380,13 +380,30 @@ def run(
...
 
 
-def tree_mapped_np_all_close(ref: np.ndarray, data: np.ndarray) -> bool:
+try:
+    RTOL, ATOL, EQUAL_NAN = np.allclose.__wrapped__.__defaults__
+except AttributeError:
+    RTOL, ATOL, EQUAL_NAN = (1e-05, 1e-08, False)
+
+
+def tree_mapped_np_all_close(
```
nfarabullini, line 389:
```suggestion
def tree_mapped_np_allclose(
```
or even
```suggestion
def tree_mapped_allclose(
```

tests/next_tests/integration_tests/feature_tests/test_util_cases.py
```diff
@@ -64,6 +65,43 @@ def test_allocate_const(cartesian_case):
...
+    assert tree_mapped_np_all_close(a, b, rtol=1e-06) is False
+
+
+def test_tree_mapped_np_all_close_nested_tuples():
+    a = (np.array([1.0]), (np.array([2.0]), np.array([3.0])))
+    b = (np.array([1.0]), (np.array([2.0]), np.array([3.0])))
+    assert tree_mapped_np_all_close(a, b) is True
+
+    c = (np.array([1.0]), (np.array([2.1]), np.array([3.0])))
+    assert tree_mapped_np_all_close(a, c) is False
```
nfarabullini, line 96:
```suggestion
    c = (np.array([1.0]), (np.array([2.1]), np.array([3.0])))
    
    assert tree_mapped_np_all_close(a, b) is True
    assert tree_mapped_np_all_close(a, c) is False
```
Just as a personal preference

PR #2055, opened 2025-05-27
test[next]: Enable unittest for performance metrics in dace backend

edopao:
The decorator test for program metrics was only running on gtfn backend, now enabled on dace too.


egparedes (approved):
LGTM

PR #2056, opened 2025-06-02
feat[dace][next]: Added `MultiStateGlobalSelfCopyElimination2`

philip-paul-mueller:
This PR adds the `MultiStateGlobalSelfCopyElimination2` transformation.
It is very similar to `MultiStateGlobalSelfCopyElimination`, however, it uses a different way to identify what to remove.
Instead scanning the global data and then identifying it searches for the transient data.
In fact the two transformation should be merged, however, currently there is no time in doing that, thus they will exist for the time being in parallel.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
@@ -6,7 +6,7 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
-from typing import Any, Optional, Sequence
+from typing import Any, Optional, Sequence, TypeAlias, Union
 
 import dace
 from dace import (
@@ -23,6 +23,11 @@ from dace.transformation.passes import analysis as dace_analysis
 from gt4py.next.program_processors.runners.dace import transformations as gtx_transformations
 
 
+AccessLocation: TypeAlias = tuple[dace.SDFGState, dace_nodes.AccessNode]
+"""An AccessNode and the state it is located in.
+"""
+
+
 def gt_multi_state_global_self_copy_elimination(
     sdfg: dace.SDFG,
     validate: bool = False,
@@ -31,15 +36,24 @@ def gt_multi_state_global_self_copy_elimination(
 
     For the return value see `MultiStateGlobalSelfCopyElimination.apply_pass()`.
     """
-    pipeline = dace_ppl.Pipeline([gtx_transformations.MultiStateGlobalSelfCopyElimination()])
-    res = pipeline.apply_pass(sdfg, {})
+    transforms = [
+        gtx_transformations.MultiStateGlobalSelfCopyElimination(),
+        gtx_transformations.MultiStateGlobalSelfCopyElimination2(),
+    ]
+
+    pipeline = dace_ppl.Pipeline(transforms)
+    pip_res = pipeline.apply_pass(sdfg, {})
 
     if validate:
         sdfg.validate()
 
-    if "MultiStateGlobalSelfCopyElimination" not in res:
-        return None
-    return res["MultiStateGlobalSelfCopyElimination"][sdfg]
+    res: dict[dace.SDFG, set[str]] = {}
+    for trans in transforms:
+        tname = trans.__class__.__name__
+        if tname in pip_res:
+            for nsdfg, processed_data in pip_res[tname].items():
+                res.setdefault(nsdfg, set()).update(processed_data)
+    return res if res else None
 
 
 def gt_remove_copy_chain(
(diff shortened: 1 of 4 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -31,15 +36,24 @@ def gt_multi_state_global_self_copy_elimination(
...
+    pipeline = dace_ppl.Pipeline(transforms)
+    pip_res = pipeline.apply_pass(sdfg, {})
 
     if validate:
         sdfg.validate()
 
-    if "MultiStateGlobalSelfCopyElimination" not in res:
-        return None
-    return res["MultiStateGlobalSelfCopyElimination"][sdfg]
+    res: dict[dace.SDFG, set[str]] = {}
```
edopao: What is the purpose of this loop? Is it to merge the results of the two transformations into a single result? maybe add a comment.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -470,6 +484,293 @@ def _has_read_access_for(
...
 
 
+@dace_properties.make_properties
+class MultiStateGlobalSelfCopyElimination2(dace_transformation.Pass):
+    """Removes self copying across different states.
+
+    This function is very similar to `MultiStateGlobalSelfCopyElimination2`, however,
+    it is a bit more restricted, as `MultiStateGlobalSelfCopyElimination` has a much
+    better way to handle some edge cases.
+    The main difference is that `MultiStateGlobalSelfCopyElimination2` uses an other
```
edopao:
```suggestion
    The main difference is that `MultiStateGlobalSelfCopyElimination2` uses another
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -470,6 +484,293 @@ def _has_read_access_for(
...
+    way to locate the redundant data. Instead of focusing on the on the globals this
```
edopao:
```suggestion
    way to locate the redundant data. Instead of focusing on the globals this
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -470,6 +484,293 @@ def _has_read_access_for(
...
+    transformation focuses on the transients.
+
+    Todo:
+        Merge with the `MultiStateGlobalSelfCopyElimination2`.
```
edopao:
```suggestion
        Merge with the `MultiStateGlobalSelfCopyElimination`.
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -470,6 +484,293 @@ def _has_read_access_for(
...
+            state.remove_node(transient_ac)
+
+            for transient_neighbour in transient_neighbours:
+                if state.degree(transient_neighbour) == 0:
+                    state.remove_node(transient_neighbour)
+
+            if transient_ac.data in sdfg.arrays:
+                sdfg.remove_data(transient_ac.data)
+
+    def _remove_transient_at_definition_point(
```
edopao: Maybe, move this function above `_remove_transient_at_using_location`

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -470,6 +484,293 @@ def _has_read_access_for(
...
+        # Currently we require that there is exactly one location where the temporary
+        #  is defined and written back. This is a simplification, that should be
+        #  removed.
+        if len(write_locations) != 1 and len(read_locations) != 1:
+            return None
+
+        # We actually have to check if the global data is not modified between the
+        #  location where the transient is defined and where it is written back.
+        #  `MultiStateGlobalSelfCopyElimination` does this in a quite sophisticated
+        #  way. We do it cheap we require that the state in which the definition of
```
edopao:
```suggestion
        #  way. Here we do it in a cheap way: we require that the state in which the definition of
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -470,6 +484,293 @@ def _has_read_access_for(
...
+            #  outgoing edge.
+            successor_states = gtx_transformations.utils.find_successor_state(defining_state)
+            if not all(
+                write_back_state in successor_states for write_back_state, _ in read_locations
+            ):
+                return None
+
+        # We now must look at what defines the transient. For that we look at what
+        #  defines it. For that there must be global data that writes into it.
+        # TODO(phimuell): To better handle `concat_where` also allows that more
```
edopao:
```suggestion
        # TODO(phimuell): To better handle `concat_where` also allow multiple producers.
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -470,6 +484,293 @@ def _has_read_access_for(
...
+        #  Currently we only allow that there is a single connection.
+        #  The main issue with allowing multiple connections is that we can no
+        #  longer be sure that everything is written. I encountered some dead
+        #  writes once.
+        assert global_data is not None
+        for state, transient_access_node in read_locations:
+            assert state.in_degree(transient_access_node) == 0
+            if state.out_degree(transient_access_node) != 1:
+                return None
+            if not all(
```
edopao: At this point we know that there is a single outgoing edge, so you could simplify the check.

src/gt4py/next/program_processors/runners/dace/transformations/utils.py
```diff
@@ -571,3 +571,49 @@ def find_upstream_nodes(
...
+
+        curr_out_edges = [oedge.dst for oedge in graph.out_edges(state)]
+
+        # End recursion if we found some successor edges or we have reached the top.
+        if len(curr_out_edges) > 0 or graph.parent_graph is state.sdfg:
+            return curr_out_edges
+        else:
+            # We have not found any successor edges, so we go one level up.
+            #  This time we look for the enclosing graph.
+            # TODO(phimuell): Should loops be handled special?
```
edopao: Yes, I would say so. You could maybe check for the regular case, first:
```
if isinstance(graph.parent_graph, dace.SDFG):
  return _impl(graph, graph.parent_graph)
else:
  assert isinstance(graph.parent_graph, (dace.sdfg.state.ConditionalBlock, dace.sdfg.state.LoopRegion))
  return _impl(graph.parent_graph, graph.parent_graph.parent_graph)
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: good idea.

src/gt4py/next/program_processors/runners/dace/transformations/utils.py
philip-paul-mueller: I will rewrite, but not handle the loop specially.
This was the result from my discussion with Philipp Schaad, where he pointed out that the successor state of a loop could also be the first state in the loop body.
I will not handle that case, but make the comment a bit clearer.

edopao (approved):
LGTM

PR #2057, opened 2025-06-02
fix[next][dace]: Updated `SerialMapPromoter`

philip-paul-mueller:
This PR modifies the `SerialMapPromoter` in such a way that it performs the fusing of the two Maps immediately after the promotion.
This is needed because technically such an SDFG is invalid as memory locations are now written multiple time (it was not critical, because they always write the same value).
By performing the fuse inside the `apply()` function, no invalid SDFG is visible to the outside.

Furthermore, this PR removes the `BasePromoter` that was not used anywhere else.
As an additional restriction it is now only possible to promote the first Map, and no longer the second Map, however, this feature was probably unneeded anyway (in GPU transformations it might be needed).

### DIFF
--- a/tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_serial_map_promoter.py
+++ b/tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_serial_map_promoter.py
@@ -75,14 +75,17 @@ def test_serial_map_promotion():
     assert len(map_entry_2d.map.range) == 2
 
     # Now apply the promotion
-    sdfg.apply_transformations(
+    count = sdfg.apply_transformations(
         gtx_transformations.SerialMapPromoter(
             promote_all=True,
+            # Do not fuse to inspect that the promotion worked.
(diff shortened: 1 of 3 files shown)


tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_serial_map_promoter.py
```diff
@@ -75,14 +75,17 @@ def test_serial_map_promotion():
     assert len(map_entry_2d.map.range) == 2
 
     # Now apply the promotion
-    sdfg.apply_transformations(
+    count = sdfg.apply_transformations(
         gtx_transformations.SerialMapPromoter(
             promote_all=True,
+            # Do not fuse to inspect that the promotion worked.
+            fuse_after_promotion=False,
```
iomaganaris, line 91: Since this transformation is mostly going to be used with `fuse_after_promotion` set to true I would recommend adding a test with this option enabled

iomaganaris (commented):
Looks good overall. I only have a suggestion regarding testing

PR #2058, opened 2025-06-02
feat[dace][next]: Added the "Splitting Tools"

philip-paul-mueller:
This is a set of tools that allows to perform splits of edges and AccessNodes.
They should more be seen as a tool box.

In addition this PR also updates the `SplitAccessNode` transformation, it now also uses the tool box and adds the `SplitConsumerMemlet` that allows to split consumer edges, such that `SplitAccessNode is more applicable.

Note: Must be merged after [PR#2057 "Updated `SerialMapPromoter`"](https://github.com/GridTools/gt4py/pull/2057) has been merged, as it kind of depends on it.


src/gt4py/next/program_processors/runners/dace/transformations/__init__.py
```diff
@@ -69,11 +71,13 @@
     "MapIterationOrder",
     "MoveDataflowIntoIfBody",
     "MultiStateGlobalSelfCopyElimination",
+    "MultiStateGlobalSelfCopyElimination2",
```
iomaganaris: Should be removed

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -252,28 +271,52 @@ def _gt_auto_process_top_level_maps(
...
+                    assume_pointwise=assume_pointwise,
+                ),
+                # TODO(phimuell): Add a criteria to decide if we should promote or not.
+                gtx_transformations.SerialMapPromoter(
+                    only_toplevel_maps=True,
+                    promote_vertical=True,
+                    promote_horizontal=False,
+                    promote_local=False,
+                ),
+            ],
```
iomaganaris: I guess this doesn't need to be run here again since we run it in line 277?

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -285,7 +328,14 @@ def _gt_auto_process_top_level_maps(
...
         # The SDFG was modified by the transformations above. The SDFG was
         #  modified. Call Simplify and try again to further optimize.
-        gtx_transformations.gt_simplify(sdfg, validate=validate, validate_all=validate_all)
+        gtx_transformations.gt_simplify(
+            sdfg,
+            validate=validate,
+            validate_all=validate_all,
+            skip=gtx_transformations.simplify.GT_SIMPLIFY_DEFAULT_SKIP_SET.union(
+                ["ConsolidateEdges"]
+            ),
```
iomaganaris: Could you please add some comment regarding why we disable edge consolidation here?

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
```diff
@@ -122,32 +119,34 @@ def can_be_applied(
...
         number_of_consumers = graph.out_degree(access_node)
-        if number_of_consumers <= 1:
-            return False
-
-        # Furthermore, they must also be different distinct consumers.
-        # NOTE: This is for simplifying the implementation.
-        if number_of_consumers != len({consumer for consumer in graph.out_edges(access_node)}):
-            return False
+        if number_of_consumers == 0:
+            pass
```
iomaganaris: A bit unrelated question to this PR. If we have a transient access node with no consumers, do we have something like dead code elimination to remove the access node and its computation?

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
```diff
@@ -247,16 +269,21 @@ def _find_producer(
...
-                f"Found an invalid SDFG, there are multiple producer for '{self.access_node.data}"
-            )
+            # This might indicate an error (the same memory location is written by
+            #  multiple producer. However, there are some cases where it is not an
+            #  error. For example a Map, with two Tasklets, writes to the node,
+            #  one Tasklet writes `T[__i, 0]` the other `T[__i, 10]`, where `__i`
+            #  is the iteration index. Then Memlet propagation will set the subset
+            #  to something like `T[:, 0:10]`. So it is not an error in that case.
+            return None
+
```
iomaganaris: I'm wondering if we should generate a warning here. We shouldn't fail because this pass is run during the horizontal/vertical map fusion where we allow such case but it shouldn't be valid SDFG in the general

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
```diff
@@ -188,27 +201,36 @@ def _match_consumers_to_producers(
...
+            for producer, assigned_consumers in assignments.items()
+            if len(assigned_consumers) == 0
+        ]
+        if len(unused_producers) == 0:
+            warnings.warn(
+                "'SplitAccessNode': found producers "
+                + ", ".join((str(p) for p in unused_producers))
+                + " that generates data but that is never read.",
+                stacklevel=0,
+            )
```
iomaganaris: I hope that domain inference saves us from this scenario. If not and there is such case then we have to figure out why. I'm wondering if its a good idea to generate an error here even

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
```diff
@@ -157,23 +156,37 @@ def apply(
     ) -> None:
         access_node: dace_nodes.AccessNode = self.access_node
 
-        assignment = self._match_consumers_to_producers(graph)
-        assert assignment is not None
+        assignments = self._match_consumers_to_producers(graph)
```
iomaganaris: Nitpick.. I would prefer a more descriptive name for `assignments` like `matched_node_in_out_edges` or something better if you have any idea

src/gt4py/next/program_processors/runners/dace/transformations/split_memlet.py
```diff
@@ -0,0 +1,178 @@
...
+            desc.subset
+            for desc in gtx_transformations.spliting_tools.describe_incoming_edges(state, node)
+        ]
+        copy_edges_to_split = [
+            oedge
+            for oedge in state.out_edges(node)
+            if isinstance(oedge.dst, dace_nodes.AccessNode)
+            and (not gtx_transformations.utils.is_view(oedge.dst, sdfg))
+            and (oedge.data.dst_subset is not None)
+        ]
```
iomaganaris: Why does this variable has a `copy_` prefix? do we actually need to copy the edges?

src/gt4py/next/program_processors/runners/dace/transformations/split_memlet.py
```diff
@@ -0,0 +1,178 @@
...
+            for split in split_description:
+                if gtx_transformations.spliting_tools.decompose_subset(
+                    producer=split, consumer=copy_edge_to_split.data.src_subset
+                ):
+                    found_edge_to_split = True
+                    break
+            if found_edge_to_split:
+                break
+        if not found_edge_to_split:
+            return False
```
iomaganaris: I'm a bit skeptical about having checks like this in the `can_be_applied` function since those are things we also do at the `apply`. If there's no way to cache such results I'm wondering if it's more beneficial to skip this check from `can_be_applied` and just exit `apply` if there's nothing we can do

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_split_memlet.py
```diff
@@ -0,0 +1,131 @@
...
+    state.add_nedge(tmp, dst, dace.Memlet("tmp[0:100, 13:80] -> [0:100, 13:80]"))
+    state.add_nedge(tmp, dst, dace.Memlet("tmp[0:30, 1:13] -> [0:30, 1:13]"))
+    # In the original case this is a Map.
+    state.add_nedge(tmp, dst, dace.Memlet("tmp[30:100, 1:13] -> [30:100, 1:13]"))
+    sdfg.validate()
+
+    return sdfg, state, tmp
+
+
+def test_split_edge_mock_apply_diffusion_to_w():
```
iomaganaris, line 108: Nice test 👍

iomaganaris (commented):
LGTM overall. Some questions/suggestions/nitpicks

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
philip-paul-mueller: Yes, this was probably a bad merge, at some point I lost track.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
philip-paul-mueller: The comment is at the top of the function.
I guess it does not show up in the diff because it was technically part of another PR.

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
philip-paul-mueller: Yes, this is a pass in DaCe simplify that does this (plus some additional pass from GT4Py).
However, you made the comment on a older version (sorry, for constantly pushing to the branch).
But in the newest version we exit here.

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
philip-paul-mueller: I think at one point there was a warning here.
But I agree that having a warning here, is probably not that bad, but I think that generating an error would be too much.

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
philip-paul-mueller: There is actually a very interesting case that can happen.
Consider your typical `concat_where()` scenario, thus there is a transient, and two Maps write into it, lets say `M1` writes `0:5` and `M2` writes `5:10`.
Then there is a third Map, `M3`, that immediately consumes `0:5`.
It is possible to fuse `M1` and `M3` together, because `M2` also writes into the transient `MapFusion` has to preserve the transient and hence the write to `0:5`, that is unneeded, if no one else also read `0:5`.
Now since `MapFusion` runs before the splitting `SplitAccessNode` will see this situation.

However, I agree that it is quite a rare case, I forgot in which stencil it happened.

iomaganaris: If no one else reads from this transient `5:10` is this going to be removed by some transformation? For example `simplify`?

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
philip-paul-mueller: You are right `assignments` is a bit to general.
After some back and forth I settled  with `edge_reassignment`, I also renamed the function.

src/gt4py/next/program_processors/runners/dace/transformations/split_memlet.py
philip-paul-mueller: The prefix does not say that the edges are copied.
Actually the prefix is `copy_edge` which should indicate that the edge is between two AccessNode and describe a copy operation.

Now that I think about it you are right it is a bit strange, because every Memlet describes kind of a copy.
What I wanted to indicate is, that you have essentially something that results in a `memcpy` operation.
I have removed the prefix.

src/gt4py/next/program_processors/runners/dace/transformations/split_memlet.py
philip-paul-mueller: The transformation system in DaCe was explicitly designed to be stateless, thus you have no guarantee that `apply()` is called after `can_be_applied()` returned `True`.

It is also not possible to gracefully exit `apply()` because `can_be_apply()` would still return `True`, thus you will be trapped in an infinite loop.

What you could do is moving that code into a function, but I in that case I do not think that it makes much sense to do.

iomaganaris: Okay then, I see. This is something we can discuss with DaCe devs in the future to improve a bit the compilation times in case it is a blocker

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
philip-paul-mueller: In our situation there is still `M1` than that writes `0:5` and if that is consumed further down (either in this state or in a successor state) then the transient as a whole is not removed.

iomaganaris (approved):

PR #2059, opened 2025-06-02
fix[next][dace]: Add method to create custom dace backend

edopao:
This PR was developed while debugging why the dace cache was not working in the icon4py granule tests. It was found that the granule test was setting up the dace backend with custom settings in a wrong way, basically disabling the translation cache. Therefore, this PR introduces a function `make_dace_backend()` to create a dace backend with custom setting for SDFG lowering and auto-optimize. Besides, this PR also contains some cleanup in the area of SDFG serialization to JSON and compilation.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
@@ -15,7 +15,7 @@ import dace
 import factory
 
 from gt4py._core import definitions as core_defs
-from gt4py.next import common, config
+from gt4py.next import common, config, metrics
 from gt4py.next.iterator import ir as itir, transforms as itir_transforms
 from gt4py.next.otf import arguments, languages, stages, step_types, workflow
 from gt4py.next.otf.binding import interface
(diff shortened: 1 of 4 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -184,9 +184,12 @@ def generate_sdfg(
                 gtx_transformations.gt_simplify(sdfg)
                 gtx_transformations.gt_gpu_transformation(sdfg, try_removing_trivial_maps=True)
 
-        # Do not use async SDFG call when collecting metrics: we use SDFG instrumentatiom
-        #   and therefore the SDFG execution has to complete before we can retrieve the report.
-        if on_gpu and self.async_sdfg_call and not config.COLLECT_METRICS_LEVEL:
+        if config.COLLECT_METRICS_LEVEL != metrics.DISABLED:
+            # measure execution time of the top-level SDFG
+            sdfg.instrument = dace.dtypes.InstrumentationType.Timer
```
philip-paul-mueller: I do not understand the meaning of this one here.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -216,7 +219,7 @@ def __call__(
         module: stages.ProgramSource[languages.SDFG, languages.LanguageSettings] = (
             stages.ProgramSource(
                 entry_point=interface.Function(program.id, program_parameters),
-                source_code=sdfg.to_json(),
+                source_code=sdfg.to_json(hash=True),
```
philip-paul-mueller, line 228: Why do you specify `hash=True` here?
And I guess you should add a comment why it is important.

philip-paul-mueller (approved):
You should explain why you set `hash=True`.
Beside that it looks good.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
edopao: Comment updated:
```
# We measure the execution time of one program by instrumenting the
#   top-level SDFG with a cpp (std::chrono) timer. This timer measures only
#   the computation time, it does not include the overhead of calling
#   the SDFG from Python.
```

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
edopao: Done.

PR #2060, opened 2025-06-02
feat[dace][next]: Updated `SingleStateGlobalSelfCopyElimination`

philip-paul-mueller:


src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+    `A[2:10] = A[0:8]` would do, as this is invalid to do with a Memlet.
+
+    In the simplest case the function will eliminate the redundant copy. However,
+    the transformation is also able to handle cases where data is written
+    into `T` and both `G` nodes. In that case the function checks if the
+    three nodes act as one and will merge them together.
+
+    However, the transformation will only apply if `T` is not used anywhere else.
+
+    Depending on the situation, the function will all merge all three nodes together.
```
edopao:
```suggestion
    Depending on the situation, the function will merge all three nodes together.
```

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+    Todo:
+        There are two interesting and related cases that should also be handled.
+        - If `T` is accessed downstream, then one could replace `T` with `G`
+            as they are the same (in some cases).
+        - If the pattern was found in two parallel branches, it will not apply.
+            This should be fixed.
+    """
+
+    # Name of all data that is used at only one place. Is computed by the
+    #  `FindSingleUseData` pass and be passed at construction time. Needed until
```
edopao:
```suggestion
    #  `FindSingleUseData` pass and must be passed at construction time. Needed until
```

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+    ) -> bool:
+        node_g1 = self.node_g1
+        node_g2 = self.node_g2
+        node_tmp = self.node_tmp
+        g_desc = node_g1.desc(sdfg)
+        tmp_desc = node_tmp.desc(sdfg)
+
+        if node_g1.data != node_g2.data:
+            return False
+        assert not g_desc.transient
```
edopao, line 94: Isn't this assert a bit too strong? You are asserting on a property of the lowered SDFG. It should be enough to return `False` if `g` is transient.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+        #   is executed before or after another potential read to `g1`. Thus, I think
+        #   that there is no need for further checks.
+        if any(
+            any(e.data.is_empty() for e in graph.out_edges(node) + graph.in_edges(node))
+            for node in [node_g1, node_g2, node_tmp]
+        ):
+            return False
+
+        # Determine which merging strategy should be used, if any.
+        # TODO(phimuell): Figuring out if we the strategy somehow has to influence
```
edopao:
```suggestion
        # TODO(phimuell): Figuring out if the strategy somehow has to influence
```

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+        #   the checks? I do not think so because `can_be_apply()` essentially tries
```
edopao:
```suggestion
        #   the checks? I do not think so because `can_be_applied()` essentially tries
```

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+        #   to figuring out if we can consider the three nodes as a single one.
```
edopao:
```suggestion
        #   to figure out if we can consider the three nodes as a single one.
```

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+        g_subset_start = associated_g_subset.min_element()
+        ss_offset = [
+            g_patch_start - tmp_start
+            for g_patch_start, tmp_start in zip(g_subset_start, tmp_subset_start)
+        ]
+        return ss_offset
+
+
+@dace_properties.make_properties
+class SingleStateGlobalDirectSelfCopyElimination(dace_transformation.SingleStateTransformation):
```
edopao, line 864: I assume that this transformation is the same as in baseline, so I have not reviewed it.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+        for_test: Literal[False],
+    ) -> dict[dace_sbs.Subset, dace_sbs.Subset]: ...
+
+    def _compute_tmp_to_g_mapping(
+        self,
+        state: dace.SDFGState,
+        node_g1: dace_nodes.AccessNode,
+        node_tmp: dace_nodes.AccessNode,
+        node_g2: dace_nodes.AccessNode,
+        for_test: bool,
```
edopao: I propose to rename `for_test` to `check_only`

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+            return self._MERGE_TMP_G2
+
+        return None
+
+    @overload
+    def _compute_offset(
+        self,
+        tdesc: gtx_st.EdgeConnectionSpec,
+        tmp_to_g_mapping: dict[dace_sbs.Subset, dace_sbs.Subset],
+        for_test: Literal[True],
```
edopao: I propose to rename `for_test` to `check_only`

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+        merging_strategy = self._check_merging_strategy(
+            state=graph,
+            node_g1=node_g1,
+            node_tmp=node_tmp,
+            node_g2=node_g2,
+        )
+        if merging_strategy is None:
+            return False
+
+        # To avoid race conditions we require that no other node referring to
```
edopao, line 134: If another node is referring to G, there maybe was already a race condition, or?

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+        #  `G` is inside this state.
+        if any(
+            dnode.data == node_g1.data
+            for dnode in graph.data_nodes()
+            if not (dnode is node_g2 or dnode is node_g1)
+        ):
+            return False
+
+        # Get the mapping, i.e. subsets on `tmp` to subsets on `g`. The function will
+        #  also make sure that every `tmp` subset is translatable to `g` subset.
```
edopao:
```suggestion
        #  check that every `tmp` subset is translatable to `g` subset, otherwise it returns `None`.
```

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+        #  separately) or from `g1` (according to ADR-18 can only be the same
+        #  position). See it as "pure writes"; described in terms of `g`.
+        pure_writes_into_g2 = [
+            gtx_st.describe_edge(edge, incoming_edge=True)
+            for edge in state.in_edges(node_g2)
+            if edge.src not in [node_tmp, node_g1]
+        ]
+
+        for write_into_g2 in pure_writes_into_g2:
+            # We have a conflict if there is a write at `g1` and `g2` that goes
```
edopao: I wonder if this case can really happen? Wouldn't it be an undefined behavior?

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+            # NOTE: We only consider it a conflict if we can show that it intersects.
+            write_write_conflicts = [
+                write_into_g1
+                for write_into_g1 in all_writes_into_g1
+                if gtx_st.are_intersecting(write_into_g2.subset, write_into_g1.subset)
+            ]
+            if len(write_write_conflicts) != 0:
+                return True
+
+            # The second kind of conflict occurs if we read something from `g1`
```
edopao: Same here. Consider that we are in one state. It sounds to me like an invalid SDFG.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+        pure_tmp_writes = [
+            gtx_st.describe_edge(edge, incoming_edge=True)
+            for edge in state.in_edges(node_tmp)
+            if edge.src is not node_g1
+        ]
+
+        # Ensure that everything that is written into `tmp` is also written back.
+        for tmp_write in pure_tmp_writes:
+            if not any(
+                tmp_to_g2_write_back.covers(tmp_write.subset)
```
edopao: Even here, intersecting writes sound strange.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+    ) -> Union[None, dict[dace_sbs.Subset, dace_sbs.Subset]]:
+        """Computes a mapping that describes how `tmp` maps into `g`.
+
+        The function returns a `dict`, that maps subsets of the `tmp_node`
+        to the corresponding subset at the `g`. If this mapping does
+        not exists then the function returns `None`. Each subsets
+        can be seen as a part of `g`.
+
+        The function has two modes, that are selected by specifying
+        `for_test`. If it is `True` then the function assumes that
```
edopao: What about check_only instead of for_test?

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,939 @@
...
+                t_descriptions.append(transfer_g1_desc_t)
+
+        assert not any(tp.other_subset is None for tp in t_descriptions)
+
+        # Now merge all subsets we found together. This describes at `tmp`.
+        # TODO(phimuell): Figuring out what it means if the merge does not
+        #   result in a single subset. I think that the tests are strong
+        #   enough to handle that case.
+        # TODO(phimuell): Add a test for the case.
+        # TODO(edopao): Make sure that I added a test for that case!!
```
edopao: I don't think it should happen in a GTIR SDFG, that you compute a field with undefined values on some slices.. but still good to keep this check.

Anyway, I think the test for this is still missing.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: If we encounter that case we have an invalid SDFG, in the sens of ADR18.
So I think that an error should be maintained, one could even think of turning it into a hard error.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: Yes, I used this PR also to restructure the `redundant_array_removers.py` as the file has become too big.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: You are right it is much better.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: In most of the cases probably yes, but there are some situations where the answer is no, as the data flow itself could act as kind of a serialization.
This check is just there to be absolutely sure that nothing happens we do not want.

Furthermore, in case we will do splitting of global AccessNodes we will have multiple AccessNodes refering to the same (global) data, but writing into different region in the memory.
Thus it also serves us to remind us that this case we must perform more analysis.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: I would say it is not so likely, but my feeling tells me that the more we fuse the more likely it is.
In `1_to_13` we had to use some global data as intermediates and we can do everything with them, even writing multiple times to them.
Thus, I we will might encounter this case.

But I agree, it is probably very paranoid of checking that, I just do not want to be hit by something.
Furthermore, I want to merge this transformation back into DaCe and there it is possible for sure.

I also added a test.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: Same procedure as above.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: I am not sure if I fully understand you.
The function just checks that everything that is written into `tmp` is also written back.
It prevents that everything that has been written to `t` is not just there because it is supposed to be there but not needed, which is similar to the situation Ioannis encountered in `41_to_60`.

I relocated that code to another part, but it is still there and I added a test.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: Sometimes I saw it, but it was super small, my guess was that it was some transformation that it was still there.
And it was indeed a smart idea to put the note there such that you remind me, because I forgot about it.

philip-paul-mueller: I added the `test_multi_t_patch` test for that case.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,1018 @@
...
+    properties as dace_properties,
+    subsets as dace_sbs,
+    symbolic as dace_sym,
+    transformation as dace_transformation,
+)
+from dace.sdfg import graph as dace_graph, nodes as dace_nodes
+from dace.transformation.passes import analysis as dace_analysis
+
+from gt4py.next.program_processors.runners.dace import transformations as gtx_transformations
+from gt4py.next.program_processors.runners.dace.transformations import spliting_tools as gtx_st
```
edopao: I suggest renaming `gtx_st` to `gtx_dace_split`, because `st` makes me think about state. BTX, I see now a spelling error: `spliting`-> `splitting`.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,1018 @@
...
+        node_g1: dace_nodes.AccessNode,
+        node_tmp: dace_nodes.AccessNode,
+        node_g2: dace_nodes.AccessNode,
+        tmp_to_g_mapping: dict[dace_sbs.Subset, dace_sbs.Subset],
+        merging_strategy: int,
+    ) -> bool:
+        """
+        The function checks if the three matched AccessNode (which means the
+        global data and the transient data) can be considered as one single
+        data.
```
edopao: Better to add a description of the returned boolean value.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,1018 @@
...
+            ]
+
+        else:
+            raise NotImplementedError(f"Merging strategy {merging_strategy} is not implemented.")
+
+        # Test for write conflicts that arises because we write something into `t`
+        #  that conflicts with a write to either `g1` or `g2`.
+        # NOTE: We only consider it a conflict if we can show that it intersects.
+        for writes_into_t_wrt_g in pure_writes_into_t:
+            # If we merge `t` and `g2` together and leave `g1` be, then we do not
```
edopao: According to the comment, are we missing an if-statement here?

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
```diff
@@ -0,0 +1,1018 @@
...
+            pure_writes_into_t = [
+                self._compute_transfromed_subset(tdesc, tmp_to_g_mapping)
+                for tdesc in gtx_st.describe_incoming_edges(state, node_tmp)
+                if tdesc.other_node is not node_g1
+            ]
+
+        else:
+            raise NotImplementedError(f"Merging strategy {merging_strategy} is not implemented.")
+
+        # Test for write conflicts that arises because we write something into `t`
```
edopao:
```suggestion
        # Test for write conflicts that arise because we write something into `t`
```

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: You are right.
However, I would propose that we fix that in another PR (I will just open one) because it will produce a huge diff that will dominate everything we did here.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: Good idea.

src/gt4py/next/program_processors/runners/dace/transformations/single_state_global_self_copy_elimination.py
philip-paul-mueller: The comment is wrong.
The `if` you noticed was not there is the one from above, starting at line 245.
It was previous inside `for`, the test was also done slightly different, but I did not like it so I have moved it out.

Thanks for noticing it.

PR #2061, opened 2025-06-03
feat[dace][next]: Updated Edge Consolidation In `MapFusion`

philip-paul-mueller:
Will be a DaCe PR.
Will be superseeded by the DaCe equivalent, if that thing is updated.


iomaganaris (approved):
LGTM

PR #2062, opened 2025-06-04
feat[next]: improve eve caching and hashing utilities

egparedes:
Enhance custom hashing utilities with a `CustomMapping` class and performance improvements in `HashableBy`. Additionally, new unit tests have been added and some of these classes are moved around the `eve.utils` module for more coherent organization.

### DIFF
--- a/src/gt4py/eve/utils.py
+++ b/src/gt4py/eve/utils.py
@@ -303,13 +303,13 @@ class CustomMapping(collections.abc.MutableMapping[_K, _V]):
     retrieving values, while maintaining a mapping-like interface.
 
     Examples:
-        >>> mapping = CustomMapping(lambda x: hash(repr(x))
+        >>> mapping = CustomMapping(lambda x: hash(repr(x)))
         >>> mapping[[1, 2]] = "first"
         >>> mapping[[1, 2]]
-        "first"
+        'first'
         >>> mapping[{1, 2}] = "second"
         >>> mapping[{1, 2}]
-        "second"
+        'second'
         >>> len(mapping)
         2
     """
@@ -354,17 +354,17 @@ class CustomMapping(collections.abc.MutableMapping[_K, _V]):
 
 
 class HashableBy(Generic[_T]):
-    __slots__ = ["func", "value"]
+    __slots__ = ("hashed_value", "value")
 
-    def __init__(self, func: Callable[[_T], int], value: _T) -> None:
-        self.func = func
+    def __init__(self, hash_func: Callable[[_T], int], value: _T) -> None:
         self.value = value
+        self.hashed_value = hash_func(value)
 
     def __hash__(self) -> int:
-        return self.func(self.value)
+        return self.hashed_value
 
     def __eq__(self, other: Any) -> bool:
-        return (self.value is other.value or self.value == other.value) and self.func is self.func
+        return self.value is other.value or self.value == other.value
 
 
 @overload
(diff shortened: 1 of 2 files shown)


src/gt4py/eve/utils.py
```diff
@@ -295,53 +295,104 @@ def __getitem__(self, key: _K) -> _V:
...
+    """
 
-@overload
-def with_fluid_partial(
-    func: Literal[None] = None, *args: Any, **kwargs: Any
-) -> Callable[[Callable[_P, _T]], Callable[_P, _T]]: ...
+    __slots__ = ("key_func", "key_map", "value_map")
 
+    key_func: Callable[[_K], int]
+    key_map: dict[int, _K]
```
edopao, line 320: You need the `key_map` in order to iterate over the original key values, right?

src/gt4py/eve/utils.py
```diff
@@ -295,53 +295,104 @@ def __getitem__(self, key: _K) -> _V:
...
 
-    The `partial` attribute is a function that behaves like `functools.partial`,
-    but also supports partial application of the decorated function. It can be
-    used both as a bare or a parameterized decorator.
+class HashableBy(Generic[_T]):
+    __slots__ = ("hashed_value", "value")
 
-    Arguments:
-        func: The function to decorate.
+    def __init__(self, hash_func: Callable[[_T], int], value: _T) -> None:
```
edopao, line 359: This initialization makes me think whether `HashableBy` should be a dataclass. In previous implementation, the `hashed_value` was computed every time, while now it is computed only at initialization.

tests/eve_tests/unit_tests/test_utils.py
```diff
@@ -151,19 +151,66 @@ def unique_data_items(request):
...
+    # Test override
+    d[5] = 100
+    assert d[5] == 100
+
+    # Test length
+    assert len(d) == 2
+
+    # Test iteration
+    keys = list(d.keys())
+    assert 5 in keys and 3 in keys
```
edopao, line 176: Which of the overridden `abc.Mapping` methods is used for the `keys()` builtin?`` Is it `__iter__`, right?

edopao (approved):
Nice!

src/gt4py/eve/utils.py
egparedes: Yes, to comply with the `Mapping` interface, we need to store the original objects used as keys, since the internal values map only stores the custom keys.

tests/eve_tests/unit_tests/test_utils.py
egparedes: Yes, exactly, `__iter__` is used in the returned `KeysView` instance here: https://github.com/python/cpython/blob/3.13/Lib/_collections_abc.py#L874

src/gt4py/eve/utils.py
egparedes: Yes, the reasoning here is that a `HashableBy` instance is going to be always hashed, because that's the only thing it can do, so doing it at initialization allow us to reduce the initialization time, because we don't need to store the function, and also reduce the hashing time (specially if it's going to be hashed multiple times) because the value is never recomputed.

About your question of making this a dataclass: yes, conceptually this is already a dataclass-like. What would the benefit of making it a real dataclass? In some python versions, the instantiation of a dataclass used to be slightly slower that a barebone class, and I guess the only benefit we would get is a nicer `repr` method, since anyway we want custom implementations of the other operators, so I don't think it pays off to use dataclasses here.

edopao: That's good, thanks for the clarification.

PR #2063, opened 2025-06-04
fix[dace][next]: Allowed Horizontal Promotion in `gt_auto_optimize()`

philip-paul-mueller:
In [PR#1992](https://github.com/GridTools/gt4py/pull/1992) the behaviour of the consolidation of edges during the top level data flow optimization was changed and this caused a series of performance regressions.
One of them was located in `19_to_20` velocity advection stencil in ICON4Py, where a horizontal promotion was previously present, but not any longer, which caused a reduction in performance.
In the end it turned out that we forget to merge a [PR in ICON4Py in which this very optimization was done manually](https://github.com/C2SM/icon4py/pull/735) thus we decided to enable it by default.

The PR also adds several unit tests to fully ensure the correct behaviour.

NOTE:
Merge only after @iomaganaris has confirmed that this was indeed the cause of the degradation.


src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -503,3 +511,73 @@ def _gt_auto_post_processing(
...
+
+    # NOTE: Inclusion of the exit is important that we get the edges that
+    #   actually write the result.
+    first_map_entry = state.entry_node(first_map_exit)
+    first_map_scope = state.scope_subgraph(first_map_entry, include_entry=True, include_exit=True)
+
+    # If there are more than 20 nodes we assume that the promotion is not
+    #  beneficial. This is an arbitrary value.
+    if first_map_scope.number_of_nodes() > 22:
+        return False
```
iomaganaris, line 572: It would be nice if DaCe calculated metrics like `arithmetic intensity`. Maybe it would made more sense to check things like that

src/gt4py/next/program_processors/runners/dace/transformations/serial_map_promoter.py
```diff
@@ -224,9 +265,17 @@ def can_be_applied(
...
+        if self._bypass_fusion_test:
+            pass
+        elif not self._test_if_promoted_maps_can_be_fused(graph, sdfg):
             return False
 
+        if self._promotion_callback is not None:
+            if not self._promotion_callback(
+                graph, sdfg, first_map_exit, second_map_entry, missing_map_parameters
+            ):
+                return False
```
iomaganaris, line 277:
```suggestion
            return self._promotion_callback(
                graph, sdfg, first_map_exit, second_map_entry, missing_map_parameters
            )
```

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_serial_map_promoter.py
```diff
@@ -139,3 +141,273 @@ def test_serial_map_promotion_promote_and_merge():
...
+        inputs={"__in": dace.Memlet("t[__i, 2]")},
+        code="__out = __in + 3.0",
+        outputs={"__out": dace.Memlet("b[__i]")},
+        input_nodes={t},
+        external_edges=True,
+    )
+    sdfg.validate()
+
+    map_promoter = gtx_transformations.SerialMapPromoter(
+        promote_all=True,
```
iomaganaris:
```suggestion
        promote_everything=True,
```

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_serial_map_promoter.py
```diff
@@ -139,3 +141,273 @@ def test_serial_map_promotion_promote_and_merge():
...
+        validate_all=True,
+    )
+    assert count == 0
+
+    # However, if they have the same name, then we can merge them.
+    sdfg_i, _ = _make_sdfg_different_1d_map_name("__i")
+    assert util.count_nodes(sdfg_i, dace_nodes.MapEntry) == 2
+    count = sdfg_i.apply_transformations_repeated(
+        gtx_transformations.SerialMapPromoter(
+            promote_all=True,
```
iomaganaris:
```suggestion
            promote_everything=True,
```

src/gt4py/next/program_processors/runners/dace/transformations/serial_map_promoter.py
```diff
@@ -154,12 +187,20 @@ def __init__(
...
                 "You must select at least one class of dimension that should be promoted."
             )
 
+        self._promotion_callback = None
+        if promotion_callback is not None:
+            self._promotion_callback = promotion_callback
+
+        # This flag is only for testing. It allows to bypass the check if the
+        #  Maps can be fused after promotion.
+        self._bypass_fusion_test = False
```
iomaganaris, line 202: Not a big fan of this but I understand its purpose :(

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_serial_map_promoter.py
philip-paul-mueller: Yea I have forgotten them.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
philip-paul-mueller: This is true, but in the general case you have symbolic quantities that you need to concretize somehow.
However you could compute something based on the number of Tasklets (each Tasklet one operation) and the number of incoming edge (data load).

However, in all cases you need something that you put on the RHS, in this case 22 which is completely arbitrary.

I can implement the proposal above, but what would you use as RHS?

src/gt4py/next/program_processors/runners/dace/transformations/serial_map_promoter.py
philip-paul-mueller: The point of writing in this way is that now you can add an additional check, whatever that might be and you do not have to change this part of the code.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
iomaganaris: For now I'm okay with merging this change with the above logic. What I mentioned probably requires some more thought and work to integrate nice with maps in DaCe in general and take care of all the operations.
Even if we calculate something like this I'm not sure what we should set as limit again. It needs more analysis.

src/gt4py/next/program_processors/runners/dace/transformations/serial_map_promoter.py
iomaganaris: At the moment there's no other additional check however so it becomes a bit more complicated to understand what is the expected value from `_promotion_callback()`. Anyway no hard opinion here. Feel free to keep it as you like

iomaganaris (approved):
Passed verification and performance tests. Good to merge from my side

PR #2067, opened 2025-06-09
feat[cartesian]: DaCe bridge refactor:  OIR ->  TreeIR -> ScheduleTree -> SDFG

FlorianDeconinck:
## Description

Current bridge goes from OIR to SDFG unexpanded, then expand to a full SDFG. The bridge has accumulated tech debt, which we have been cleaning up, but barriers to new feature implementation remains.

We have been thinking about leveraging the Schedule Tree representation of DaCe for some time. The idea behind that representation is that we should have an easier time to do schedule related optimizations (e.g. loop re-ordering based on CPU/GPU target and (same-axis) loop merges).

This PR refactors the GT4Py - DaCe bridge to OIR -> TreeIR -> ScheduleTree -> SDFG -> codegen, details in the [ADR](https://github.com/GridTools/gt4py/blob/06cd753135d1a6caaabe0aca37cf735fb1d96c52/docs/development/ADRs/cartesian/backend-dace-schedule-tree.md).

Note, we might might amend a commit or two to fix/change the data layout to map the current loop structure, but no big changes are planned to go into this branch anymore. So we marked it as ready for review.

## Requirements


romanc:
cscs-ci run

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
```diff
@@ -321,16 +338,16 @@ def visit_Stencil(
...
-                tir.Axis.I: -extent[0][0],
-                tir.Axis.J: -extent[1][0],
+                tir.Axis.I: -field_extent[0][0],
+                tir.Axis.J: -field_extent[1][0],
                 tir.Axis.K: max(k_bound[0], 0),
             }
             containers[field.name] = data.Array(
                 data_type_to_dace_typeclass(field.dtype),  # dtype
-                get_dace_shape(field, extent, k_bound, symbols),  # shape
+                get_dace_shape(field, field_extent, k_bound, symbols),  # shape
```
romanc: Shouldn't temporaries also use `field_without_extent` to derive their shape?

src/gt4py/cartesian/backend/dace_stencil_object.py
```diff
@@ -53,8 +53,6 @@ def add_optional_fields(
         if info.access == AccessKind.NONE and name in kwargs and name not in sdfg.symbols:
             if isinstance(kwargs[name], dace.data.Scalar):
                 sdfg.add_scalar(name, dtype=kwargs[name].dtype)
-            else:
-                sdfg.add_symbol(name, stype=dace.typeclass(type(kwargs[name])))
```
romanc, line 57: Do we need the remaining `if isinstance(kwargs[name], dace.data.Scalar):`?

FlorianDeconinck: I am not sure we need this function at all...

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
FlorianDeconinck: Yes, will do

FlorianDeconinck: CI just died so... no? I guess the temporaries are well scoped because we calculate their full size anyway from the stencils?

pyproject.toml
```diff
@@ -438,7 +439,10 @@ url = 'https://test.pypi.org/simple/'
 
 [tool.uv.sources]
 atlas4py = {index = "test.pypi"}
-dace = {git = "https://github.com/GridTools/dace", tag = "__gt4py-next-integration_2025_07_03", extra = "dace-next"}
+dace = [
+  {git = "https://github.com/romanc/dace", branch = "romanc/stree-to-sdfg", extra = "dace-cartesian"},
+  {git = "https://github.com/GridTools/dace", tag = "__gt4py-next-integration_2025_07_03", extra = "dace-next"}
+]
```
egparedes: This looks correct but it means that `g4py[dace-whatever]` can not longer be installed correctly with anything else than `uv`, since any other build frontend will just pick the latest version of `dace` in PyPI. I understand that this is needed but it is also confusing for users, so maybe we should clearly mark this package extra as internal and only meant for developing tasks. I'm also aware that the same issue happens already for `dace-next` and we haven't done anything yet (probably because we still considered `next` "experimental") but we should also fix it now.

Something like `_dace-cartesian` (and `_dace-next`) would be already ok for me...

egparedes: Also note that using `branch` here is slightly misleading since the lockfile will store the commit which was at the head of that branch when the locking happened, and it will not be automatically updated to the new commits unless using `--refresh` or `--refresh-package` when syncing the venv.

egparedes: Actually the name of the extra needs to start with a letter, so my previous suggestion is not valid. What about `dev_dace-something` ?

romanc: > This looks correct but it means that `g4py[dace-whatever]` can not longer be installed correctly with anything else than `uv`, since any other build frontend will just pick the latest version of `dace` in PyPI.

Yes, as I understand, this is one of the drawbacks of the non-standard `uv.lock` file. I really think we should embrace the concept of lock files. We could export a (bunch of) `pylock.toml` next to `uv.lock` for non-`uv` users. Tbh, the maintenance overhead sounds like overkill to me (given our user base), but I think this would ensure best possible compatibility with other installers. I'd suggest to do this outside of this PR since it's already to big.

> I understand that this is needed but it is also confusing for users, so maybe we should clearly mark this package extra as internal and only meant for developing tasks. I'm also aware that the same issue happens already for `dace-next` and we haven't done anything yet (probably because we still considered `next` "experimental") but we should also fix it now.
> 
> Something like `_dace-cartesian` (and `_dace-next`) would be already ok for me...
> What about dev_dace-something ?

I understand this can be confusing for users and I'd suggest to improve the docs instead of renaming the extra. Fact is that gt4py has no working dace-capabilities without either `dace-cartesian` or `dace-next`. Changing their name doesn't change that fact. As argued above, I think `pylock.toml` files are a better strategy for compatibility. Anyway, since dace backends (in cartesian and next) are treated as second class citizens, I think it's only fair to have more documentation on how to get them up and running.

Getting dace from a fork/branch is a temporary state for gt4py cartesian. It will take some time, but we really want to go back to mainline dace. Meanwhile, users are basically us at NASA/NOAA and Stefano. We both are aware of the change. For NDSL, we anyway have submodules of gt4py and dace such that control these versions as needed.

romanc: Yes, I'm fully aware of that. We have [(internal) docs](https://geos-esm.github.io/SMT-Nebulae/technical/backend/repositories/gt4py/?h=uv#notes-quirks) with the commands needed to update the lock file. While we could use a tag, I think the advantage of the branch name is that is shows where the code is coming from. Given only a tag and a repo, I have to do some work to figure out with what branch this tag is associated. For me, it's a balance between usability and (maintenance) overhead: As argued above, I think everything that needs a reproducible build should rely on a lock file. And since the lock file resolves the tag as well as the branch name to a hash, I don't see a big difference there. On the other hand, maintaining tags for every update (with artificial date name as you do for gt4py-next) is a bit of work on our side (for just getting the same hash again in the lock file).

src/gt4py/cartesian/backend/dace_backend.py
```diff
@@ -233,6 +139,30 @@ def _sdfg_add_arrays_and_edges(
                     None,
                     Memlet(name, subset=subsets.Range(ranges)),
                 )
+        elif isinstance(array, data.Scalar):
+            wrapper_sdfg.add_scalar(
```
philip-paul-mueller, line 153: This will error if by chance `name` is already used.
At least the plain `add_scalar()` has a `find_new_name` (or so) that finds a new name and insert it, I would use that instead to prevent name collision.

src/gt4py/cartesian/backend/dace_backend.py
```diff
@@ -233,6 +139,30 @@ def _sdfg_add_arrays_and_edges(
                     None,
                     Memlet(name, subset=subsets.Range(ranges)),
                 )
+        elif isinstance(array, data.Scalar):
```
philip-paul-mueller, line 152: I would say here is the check that it is a global missing.

src/gt4py/cartesian/backend/dace_backend.py
```diff
@@ -201,7 +101,11 @@ def _sdfg_add_arrays_and_edges(
             ]
 
             wrapper_sdfg.add_array(
-                name, dtype=array.dtype, strides=array.strides, shape=shape, storage=array.storage
+                name,
+                dtype=array.dtype,
+                strides=array.strides,
+                shape=shape,
+                storage=array.storage,
```
philip-paul-mueller, line 111: Assuming `wrapped_sdfg` is the outer SDFG, I would say that the data container should be a transient, because on the outside it is not necessary a global, but could be the result of an intermediate computation.
So I would say you also have to pass `transient=True`.

src/gt4py/cartesian/backend/dace_backend.py
```diff
@@ -201,7 +101,11 @@ def _sdfg_add_arrays_and_edges(
             ]
 
             wrapper_sdfg.add_array(
```
philip-paul-mueller, line 106: Datadescriptors also have the `clone()` method and an SDFG has a `add_datadescriptor()` function, maybe this could simplify the code.

src/gt4py/cartesian/backend/dace_backend.py
```diff
@@ -272,13 +202,37 @@ def _sdfg_specialize_symbols(wrapper_sdfg: SDFG, domain: tuple[int, ...]) -> Non
...
     origin: dict[str, tuple[int, ...]],
     domain: tuple[int, ...],
 ) -> SDFG:
+    """Create a new SDFG by wrapping a _copy_ of the original SDFG and freezing it's
+    origin and domain.
+
+    This wrapping is required because we do not expect any of the inner_sdfg bounds to
+    have been specialized, e.g. we expect "__I/J/K" symbols to still be present. We wrap
+    the call and specialize at top level, which will then be passed as a parameter to the
+    inner sdfg.
```
philip-paul-mueller: I would be more specific if they are passed as scalars or symbols.

src/gt4py/cartesian/backend/dace_backend.py
```diff
@@ -653,14 +645,20 @@ def generate_functor_args(self, sdfg: SDFG) -> list[str]:
         for name, array in sdfg.arrays.items():
```
philip-paul-mueller, line 671: Consider using [`signature_arglist(False)`](https://github.com/spcl/dace/blob/894079e82da7552bf598993c9a473bb9a2fa771e/dace/sdfg/sdfg.py#L1478)

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
```diff
@@ -0,0 +1,359 @@
...
+
+        # Gather all parts of the variable name in this list
+        name_parts = [tasklet_name]
+
+        # Variable K offset subscript
+        if isinstance(node.offset, oir.VariableKOffset):
+            symbol = tir.Axis.K.iteration_dace_symbol()
+            shift = ctx.tree.shift[node.name][tir.Axis.K]
+            offset = self.visit(node.offset.k, ctx=ctx, is_target=False)
+            name_parts.append(f"[{symbol} + {shift} + {offset}]")
```
philip-paul-mueller:
```suggestion
            name_parts.append(f"[({symbol}) + ({shift}) + ({offset})]")
```
Here it is not important, but the string interpolation is essentially C macros.
I was hit by that issue really hard, so I have made a habit of putting parenthesis around them.
I would recommend to you doing the same.

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
```diff
@@ -0,0 +1,359 @@
...
+            return "".join(filter(None, name_parts))
+
+        # Build Memlet and add it to inputs/outputs
+        data_domains: list[int] = (
+            ctx.tree.containers[node.name].shape[-len(node.data_index) :] if node.data_index else []
+        )
+        memlet = Memlet(
+            data=node.name,
+            subset=_memlet_subset(node, data_domains, ctx),
+            volume=reduce(operator.mul, data_domains, 1),
```
philip-paul-mueller: The constructor of the Memlet should do this on [its own](https://github.com/spcl/dace/blob/894079e82da7552bf598993c9a473bb9a2fa771e/dace/memlet.py#L133).

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
```diff
@@ -0,0 +1,359 @@
...
+    ranges: list[tuple[str, str, int]] = []
+    # Handle cartesian indices
+    for index, axis in enumerate(tir.Axis.dims_3d()):
+        if dimensions[index]:
+            i = f"{axis.iteration_dace_symbol()} + {shift[axis]} + {offset_dict[axis.lower()]}"
+            ranges.append((i, i, 1))
+
+    # Append data dimensions
+    for domain_size in data_domains:
+        ranges.append(("0", f"{domain_size}-1", 1))  # ranges are inclusive
```
philip-paul-mueller:
```suggestion
        ranges.append((0, domain_size - 1, 1))  # ranges are inclusive
```

Why are these strings?

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
```diff
@@ -0,0 +1,528 @@
...
+        code_block = oir.CodeBlock(label=f"masklet_{id(node)}", body=[assignment])
+        self.visit(code_block, ctx=ctx)
+
+        return (condition_name, assignment)
+
+    def visit_HorizontalExecution(self, node: oir.HorizontalExecution, ctx: tir.Context) -> None:
+        block_extent = ctx.block_extents[id(node)]
+
+        axis_start_i = f"0 + {block_extent[0][0]}"
+        axis_start_j = f"0 + {block_extent[1][0]}"
```
philip-paul-mueller: Why the `0 + `?

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
```diff
@@ -0,0 +1,528 @@
...
+            field_extent = field_extents[field.name]
+            k_bound = self._k_bounds[field.name]
+            shift[field.name] = {
+                tir.Axis.I: -field_extent[0][0],
+                tir.Axis.J: -field_extent[1][0],
+                tir.Axis.K: max(k_bound[0], 0),
+            }
+            containers[field.name] = data.Array(
+                utils.data_type_to_dace_typeclass(field.dtype),  # dtype
+                get_dace_shape(field, field_extent, k_bound, symbols),  # shape
```
philip-paul-mueller:
```suggestion
                dtype=utils.data_type_to_dace_typeclass(field.dtype),
                shape=get_dace_shape(field, field_extent, k_bound, symbols),
```

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
```diff
@@ -0,0 +1,528 @@
...
+                strides=get_dace_strides(field, symbols),
+                transient=True,
+                lifetime=dtypes.AllocationLifetime.Persistent,
```
philip-paul-mueller, line 358: Using persistent seems a bit extreme, it means that it will be allocated the first time the SDFG is called and be kept around until (essentially) the interpreter shuts down.

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
```diff
@@ -0,0 +1,528 @@
...
+    k_bound: tuple[int, int],
+    symbols: tir.SymbolDict,
+) -> list[symbolic.symbol]:
+    shape = []
+    for index, axis in enumerate(tir.Axis.dims_3d()):
+        if field.dimensions[index]:
+            symbol = axis.domain_dace_symbol()
+            symbols[axis.domain_symbol()] = dtypes.int32
+
+            if axis == tir.Axis.I:
```
philip-paul-mueller, line 497: I would use `elif` here then you can skip the `continue` and it is more clear what you are doing.

philip-paul-mueller (commented):
I have looked at it but I only focused on the DaCe part, since I do not know much about the cartesian backend.
If you have further question, just contact me.

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
romanc: I don't understand. What are you trying to guard against?

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
romanc: Interesting. Let me double check if I messed up something in the definition of the subset.

pyproject.toml
egparedes: Good, I just wanted to let you know in case you were not yet aware.

Another random though it just occurred to me now: would it make sense to use the same `gridtools/dace` repo for the `cartesian` and `next` forks of DaCe (in different branches)? Maybe it would be good for exchanging knowledge and learning from each other... Just a thought, really, I have no idea

philip-paul-mueller: From my side there is no problem in that regard.
It helped us, especially to reduce stalling.

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
FlorianDeconinck: If those boil down to C macros, then parenthesis everywhere indeed.

pyproject.toml
FlorianDeconinck: We definitely can. A note though is that what we are pushing in branch of `v1` will within ~6 months be pushed in `v2` with the refactor that are needed.

FlorianDeconinck: `*` timeline contingent on DaCe v2 fixes & stabilization + our team is alive in 6 months

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
philip-paul-mueller: consider the following:
```python
a = "5 + 3"
b = "7"
c = f"{a} * {b}"
print(eval(a) * eval(b))  # 8 * 3 = 24
print(eval(c))  # 5 + 3 * 7 = 5 * 21 = 27
```
Thus the most safest way to write `c` is `f"({a}) * ({b})"`, in your case it does not mater because you have only `+` with the lowest precedence.
But as soon as you have more complex expressions you will get in trouble without having parentheses.
So my recommendation was: "start using them now all the time to never forget them".

pyproject.toml
egparedes: You're right that changing the name of the extra doesn't fix anything.  I came to the same conclusion after writing this yesterday and now I think the right solution is to remove for now the `dace-whatever` extras and define them as dependency groups. This way, the `dace` extras will not be available to users installing gt4py from PyPI (since they are broken anyway) but, using the dependency groups, they will still be available when installing  gt4py from a copy of the repository. I quickly tried this approach here: https://github.com/GridTools/gt4py/pull/2131

I agree standard lockfiles are a perfect solution for deployment of applications (staff like ICON4Py or PACE) but don't think they are good for libraries like GT4Py. Lockfiles also wouldn't solve the problem for users trying to install `gt4py[dace]` from PyPI. 

Standard lockfiles could also be used for having stable development environments, true, but I don't see much of an advantage of using standard lockfiles instead of the uv lock, since uv is anyway required for development and the uv.lock is way more powerful. Anyway, using uv.lock only in the gt4py repo shouldn't cause any harm since it doesn't prevent at all the use of standard lockfiles in any other project using gt4py like NDSL or PACE.

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
romanc: I see. Today I learned ... parenthesis everywhere is!

pyproject.toml
romanc: Oh, going to dependency groups is a very elegant solution. I like that. Feel free to go ahead with it in #2131. I'd just ask for one thing: Can we leave the version of `dace-cartesian` in #2131 to point to DaCe v1.0.2, such that switching to the branch containing the stree work is associated with this PR?

romanc: Yeah, we can if you prefer to. I have no strong feelings as to where this branch should live. As Florian said, this is a temp solution and we expect to be back in mainline DaCe soonish.

src/gt4py/cartesian/backend/dace_backend.py
romanc: This code runs when we "freeze" origin/domain information into the SDFG. If there was a name conflict, it would already arise in the "inner" (not yet frozen) SDFG. The should be no new names introduced in this process and I thus think it's okay to not prevent name collisions. In fact, if name collisions were to occur here, I'd want to know (even if this means people are screaming at me because "nothing works (anymore)").

pyproject.toml
egparedes: Sure, no problem, I was only testing if the lock would have problems with your version. Just restored the normal dace version in #2131

src/gt4py/cartesian/backend/dace_backend.py
romanc: Ah yes, you are right, we should skip all transients as we do when we define `inputs` and `outputs` in

https://github.com/GridTools/gt4py/blob/11ce4b6c8f1b16a5bd8b75833eb40f971ad29649/src/gt4py/cartesian/backend/dace_backend.py#L239-L248

@FlorianDeconinck we should double check validation again after this change because this is solving "the transient problem" another way. I don't remember, but I think there were things in physics that got tripped by wiring a transient into the inner sdfg and then not declaring it a transient. We shouldn't wire them into the inner sdfg in the first place. My guess is that we got lucky here and some cleanup pass removed the extra inputs/outputs of transients into the inner sdfg.

src/gt4py/cartesian/backend/dace_backend.py
romanc: Again, this code runs when we freeze origin / domain. We don't expect any computation to happen "outside". This is why we explicitly skip transients.

@FlorianDeconinck Writing this brings up the following question in my head: why are we using a wrapper sdfg and a nested SDFG just to freeze origin / domain? The code (later down) even tries to inline the previously created nested SDFG again. Could we just copy the SDFG and specialize the symbols instead? Sounds way simpler to me, but maybe I'm missing something here ...

src/gt4py/cartesian/backend/dace_backend.py
romanc: I left a note here https://github.com/GridTools/gt4py/issues/2082#issuecomment-3060731044 to re-evaluate the nested SDFG tactic a follow-up.

src/gt4py/cartesian/backend/dace_backend.py
romanc: Re-evaluate the nested SDFG as part of https://github.com/GridTools/gt4py/issues/2082#issuecomment-3060731044.

src/gt4py/cartesian/backend/dace_backend.py
romanc: Noted in https://github.com/GridTools/gt4py/issues/2082#issuecomment-3060731044.

src/gt4py/cartesian/backend/dace_backend.py
romanc: Noted in https://github.com/GridTools/gt4py/issues/2082#issuecomment-3060731044 to be applied if we can't get rid of that nested SDFG in the first place.

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
romanc: Ah, I remember why I did this. The thing is this. In case of `VariableKOffset`, we know that we are going to access 1 element inside the K axis. We don't know which one, but we know it's only one. The subset that we have to pass is thus the full K axis (because we only know at runtime where we will read). We reduce the volume back to "1 * data_dimensions" to get the correct the memory analysis numbers.

pyproject.toml
philip-paul-mueller: It is also a temp solution for us.
This PR describes how we do it: https://github.com/GridTools/dace/pull/1

For clarity I would recommend a branch name `gt4py-cartesian-integration` to distinguish it from our stuff.

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
romanc: very good question ... thanks for catching!

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
romanc: yep - we don't need these anymore - they're gone now :+1:

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
romanc: Good point!

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
romanc: I undid the `elif` again because I only add a symbol before `continue` in case there's padding. All other (non padded) symbols get added at the end. That doesn't make it easy to add `elif` and avoid the `continue`. And it's anyway only code style.

docs/development/ADRs/cartesian/backend-cuda-feature-freeze.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Context
+
+The introduction of the [`dace:*`](./backend-dace.md) backends brought up the question of backend redundancy. In particular, it seems that `cuda` and `dace:gpu` backends serve similar purposes.
+
+`dace:gpu` backends not only generate code for different graphics cards, they also share substantial code paths with the `dace:cpu` backend. This simplifies (backend) feature development.
+
+## Decision
+
+We decided to put a feature freeze on the `cuda` backend, focusing on the `dace:*` backends instead. While we don't drop the backend, new DSL features won't be able in the `cuda` backend. New features will error out cleanly and suggest to use the `dace:gpu` backend instead.
```
edopao:
```suggestion
We decided to put a feature freeze on the `cuda` backend, focusing on the `dace:*` backends instead. While we don't drop the backend, new DSL features won't be available in the `cuda` backend. New features will error out cleanly and suggest to use the `dace:gpu` backend instead.
```

docs/development/ADRs/cartesian/backend-dace-version.md
```diff
@@ -0,0 +1,26 @@
+# DaCe backends: DaCe version
+
+In the context of the [DaCe backend](./backend-dace.md) and the [schedule tree](./backend-dace-schedule-tree.md), facing time pressure, we decided to stay at the `v1.x` branch of DaCe to minimize up-front cost and deliver CPU performance as fast as possible. We considered updating to the mainline version of DaCe and accept follow-up cost of partial rewrites once DaCe `v2` releases.
+
+## Context
+
+The currently released version of DaCe is on the `v1.x` branch. However, the mainline branch moved on (with breaking changes) to what is supposed to be DaCe `v2`. All feature development is supposed to be merged against mainline. Only bug fixes are allowed on the `v1.x` branch.
+
+The [schedule tree](./backend-dace-schedule-tree.md) feature will need changes in DaCe, in particular to translate schedule trees into SDFG. We are unfamiliar with the breaking changes in DaCe.
```
edopao, line 9: Do you know whether ScheduleTree IR will be the same across dace-v1 and v2?

docs/development/ADRs/cartesian/backend-dace.md
```diff
@@ -0,0 +1,25 @@
+# DaCe backends
+
+In the context of performance optimization, facing the fragmentedness of NWP code, we decided to implemented a backend based on DaCe to unlock full-program optimization. We accept the downside of having to maintain that (additional) performance backend.
```
edopao:
```suggestion
In the context of performance optimization, facing the fragmentedness of NWP code, we decided to implement a backend based on DaCe to unlock full-program optimization. We accept the downside of having to maintain that (additional) performance backend.
```

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
edopao: I think they have a pass to turn these arrays into scope-lifetime and allocate them in the CUDA memory pool.

src/gt4py/cartesian/gtc/dace/treeir_to_stree.py
```diff
@@ -0,0 +1,229 @@
...
+        with ContextPushPop(ctx, map_scope):
+            self.visit(node.children, ctx=ctx)
+
+    def visit_VerticalLoop(self, node: tir.VerticalLoop, ctx: Context) -> None:
+        # In any case, define the iteration symbol
+        ctx.tree.symbols[tir.Axis.K.iteration_symbol()] = dtypes.int32
+
+        # For serial loops, create a ForScope and add it to the tree
+        if node.loop_order != common.LoopOrder.PARALLEL:
+            for_scope = tn.ForScope(header=_for_scope_header(node), children=[])
```
edopao, line 97: Just a question. By using ScheduleTree in dace-v1, will `tn.ForScope` be lowered to state transitions? while using it in dace-v2, the same node will be lowered to a control flow region?

edopao (commented):
Very interesting, I have learnt a lot from this code review.

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
romanc: You are right - let me change that back to `AllocationLifetime.Scope`, which is the default.

docs/development/ADRs/cartesian/backend-dace-version.md
romanc: From what I can extrapolate, I don't expect big changes go the schedule tree IR. If we built the abstractions correctly, changes will be scoped to `treeir_to_stree.py` when we update from DaCe v1 -> v2.

src/gt4py/cartesian/gtc/dace/treeir_to_stree.py
romanc: Yes, the schedule tree -> SDFG transformation in dace will `tn.ForScope`s into the classical state machine as shown in the [control flow sample](https://github.com/spcl/dace/blob/5cbed3ed20e92b62bc6cb5738d1f3382355ee050/samples/sdfg_api/control_flow.py) of the SDFG API. Once we port this to DaCe v2, `tn.ForLoop`s (as well as other control flow) can be directly mapped to control flow regions. It will definitely streamline the schedule tree -> SDFG transformation. The future is bright :sparkles:

pyproject.toml
egparedes: Ok, so let's use the same fork for both. I've just added the `gt4py-collaborators` group to the DaCe fork repo so from now you should have the same access permissions as for the gt4py repo

FlorianDeconinck:
Thank you @edopao @philip-paul-mueller and @egparedes for engaging here. We are well aware of you busy schedule and your time spend helping us improve this is highly appreciated!

pyproject.toml
romanc: Thanks for setting this up, @egparedes. I can confirm that I have access as I pushed the branch `romanc/stree-to-sdfg` to `GridTools/dace`. I could also instruct the `pyproject.toml` to pull from there.

romanc:
Alright - the issue with temporaries is squashed (and worked around where we can't control it). I think this is ready for show. I've tested with pyFV3 and the AI2 data and this validates all major stencils over there in orchestration as well as stencil mode.

romanc (approved):
I did "one last code review" resulting in the last commit https://github.com/GridTools/gt4py/pull/2067/commits/c34ccc2d4cab713cc9c42c2c77209d5ff4e388c4.

@FlorianDeconinck have a quick look at the changes to memlet range computation when adding fields to the wrapper sdfg (in `dace_backend`).

Also, I figured, we missed to offset the i/j axes for nodes with `VariableKOffset`. The K-axis was okay (we add the offset when we build the `FieldAccess` in the code generation of the Tasklet. This might help with problems that we have seen in physics codes.

romanc:
cscs-ci run

romanc:
cscs-ci run

twicki (approved):
Great first step!

romanc:
cscs-ci run

romanc:
cscs-ci run

PR #2070, opened 2025-06-10
fix[dace][next]: Updated Handling of Not Supported Memlets

philip-paul-mueller:
If the DaCe GPU code generator encounters a Memlet that can not be expressed as a `cudaMemcpy*()` call, then it converts it to a Map.
However, the issue is that these Maps have the wrong iteration order, i.e. wrong memory access pattern and it might not even launch, because of too many blocks in the `y` direction of the compute grid.
For this reason GT4Py has to handle these Memlets explicitly.

However, [DaCe PR#1976](https://github.com/spcl/dace/pull/1976) changed this slightly and thus GT4Py had to follow.
Note, that some of these changes were already introduced by [GT4Py PR#2004](https://github.com/GridTools/gt4py/pull/2004), however, they were made for the original version of the DaCe PR (and the GT4Py PR had to be merged before the DaCe PR was merged).

Furthermore, this PR fixes a different issue, also related to the expansion of Memlets, which can be found in [DaCe PR#2033](https://github.com/spcl/dace/pull/2033) (it is not yet merged and currently at commit `19b6bba`).
It fixes a bug in how the Memlets are expanded.
The DaCe PR also adds the possibility to generate an error instead of slightly converting Memlets into Maps and this PR enables this feature.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/common.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/common.py
@@ -34,6 +34,10 @@ def set_dace_config(
     dace.config.Config.set("default_build_folder", value=str(config.BUILD_CACHE_DIR / "dace_cache"))
     dace.config.Config.set("compiler.use_cache", value=True)
 
+    # Prevents the implicit change of Memlets to Maps. Instead they should be handled by
+    #  `gt4py.next.program_processors.runners.dace.transfromations.gpu_utils.gt_gpu_transform_non_standard_memlet()`.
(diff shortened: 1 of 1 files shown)


edopao (changes requested):
As you suggest, I think we should set `allow_implicit_memlet_to_map=False` in `gtx.program_processors.runners.dace.workflow.common.set_dace_config()`

philip-paul-mueller:
Thanks for the hint, I forgot to ask you where you set these values.

edopao:
> Thanks for the hint, I forgot to ask you where you set these values.

It is actually the place you suggested during review of my PR 😄

PR #2071, opened 2025-06-10
fix[next][dace]: Do not apply map_promoter on symbolic range

edopao:
When a 1d temporary is consumed by different 2d maps, with symbolic range, we can promote the 1d producer map to a 2d map only if the range of the new dimension is known to be not empty. The issue found in icon4py (stencil `39_to_40`, without constant substitution) was that the temporary was no longer produced, after applying MapPromotion, because the new dimension was empty. The other consumer maps, that actually needed the temporary data, found uninitialized data (it happened to be zeros).


src/gt4py/next/program_processors/runners/dace/transformations/map_promoter.py
```diff
@@ -214,19 +214,29 @@ def can_be_applied(
         #  value is written and if we fuse the write to `a` disappears, if `a` is
         #  a single use transient. Thus we have to make sure that we do not write
         #  into global memory.
+        # We also collect all maps that consume the data produced by the first map.
```
philip-paul-mueller: I think you should add a comment explaining why, because even I who kind of suggested that forgot about that.

src/gt4py/next/program_processors/runners/dace/transformations/map_promoter.py
```diff
@@ -214,19 +214,29 @@ def can_be_applied(
...
         for oedge in graph.out_edges(first_map_exit):
             if not isinstance(oedge.dst, dace_nodes.AccessNode):
                 return False
             if not oedge.dst.desc(sdfg).transient:
                 return False
-
+            # If present, store the map on this edge that consume the transient.
+            for oedge_next in graph.out_edges(oedge.dst):
+                if isinstance(oedge_next.dst, dace_nodes.MapEntry):
+                    consumer_maps.add(oedge_next.dst)
```
philip-paul-mueller: The `else` branch here is actually very interesting, currently we do not care, but the question is should we care?
Because, if we find something else than the second Map it implies that we consume whatever is generated, therefore we should care.

After spending an embarrassing amount of time going back and forth, I think we should change what we are doing and try to be less smart.
I see two different paths:
- We first check if the second Map is executed, i.e. `second_map_entry.map.range.num_elements() > 0) != True`. If it is, then we do not impose any further restrictions. If we can not show that, then we change the check on line 229 to `if oedge_next.dst is not second_map_entry: return False`, because consumption by something other than the second Map means that it can not be eliminated and is needed somewhere else. The drawback of this approach is, however, that in case the temporary is retained, the same memory location is written multiple times.
-  The second possibility would be to always do `if oedge_next.dst is not second_map_entry: return False`, this would guarantee that the transient is eliminated. This approach does not have the draw back of the above one, but is applicable in less places.

I am not sure which approach is better, in the spirit of ADR18 I would say the second one, but I do not have a problem with the first one and personally (currently) I do not care as long as we document the ADR18 violation.

However, in each cases your current checks, if the transient can be eliminated are incomplete, as they only cover the case of "used within the state". To also capture the case of "used in a different state" you have to use the single use data, you can find an example [here](https://github.com/GridTools/gt4py/blob/db11163a4c91c3fbcad957e8d73500434e8f57a8/src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py#L469)

src/gt4py/next/program_processors/runners/dace/transformations/map_promoter.py
```diff
@@ -214,19 +214,29 @@ def can_be_applied(
...
         # Test if after the promotion the maps could be fused.
         if not self._test_if_promoted_maps_can_be_fused(graph, sdfg):
             return False
 
+        if len(consumer_maps) > 1 and (second_map_entry.map.range.num_elements() > 0) != True:  # noqa: E712 [true-false-comparison]  # SymPy fuzzy bools.
```
philip-paul-mueller: See my comment above.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_serial_map_promoter.py
```diff
@@ -21,11 +21,12 @@
...
 
+N = 10
 
-def _make_serial_map_promotion_sdfg() -> (
-    tuple[dace.SDFG, dace.SDFGState, dace_nodes.MapEntry, dace_nodes.MapEntry]
-):
-    N = 10
+
+def _make_serial_map_promotion_sdfg(
+    use_symbolic_range: bool = False,
```
philip-paul-mueller, line 30:
```suggestion
    use_empty_symbolic_range: bool = False,
```

philip-paul-mueller (commented):
I think we have to refine the approach a little bit but we are close.

If you need a discussion or if you need help in merging, as we have modified the transformation in [PR#2063](https://github.com/GridTools/gt4py/pull/2063) please drop me a message.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_serial_map_promoter.py
edopao: Not necessarily empty. It is enough that the range is symbolic, and the transformation cannot infer whether it is empty or not at compile time.

src/gt4py/next/program_processors/runners/dace/transformations/map_promoter.py
edopao: OK

src/gt4py/next/program_processors/runners/dace/transformations/map_promoter.py
edopao: Totally correct!

PR #2073, opened 2025-06-11
fix[next]: Fix symbol name collision in CollapseTuple pass

tehrengruber:


src/gt4py/next/iterator/ir_utils/misc.py
```diff
@@ -216,3 +216,21 @@ def extract_projector(
         projector = projector if cur_projector is None else im.compose(cur_projector, projector)
         projector = inline_lambdas.InlineLambdas.apply(projector)
         return extract_projector(expr, projector, _depth + 1)
+
+
+SymOrStr = TypeVar("SymOrStr", itir.Sym, str)
+
+
+def unique_symbol(sym: SymOrStr, reserved_names: Iterable[str]) -> SymOrStr:
```
havogt, line 224: I thought we already have this function somewhere. Did you have this in a some PR long time ago but in the end that part was not merged?

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -106,6 +106,13 @@ def _flattened_as_fieldop_param_el_name(param: str, idx: int) -> str:
     return f"{prefix}{idx}_{suffix}"
 
 
+def _unique_symbol(reserved_names: Sequence[str], sym: itir.Sym):
+    name: str = sym.id
+    while name in reserved_names:
+        name = name + "_"
+    return im.sym(name, sym.type)
+
+
```
havogt:
```suggestion
```
leftover

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -477,16 +484,25 @@ def transform_propagate_to_if_on_tuples_cps(
...
-                    inner_vars[arg_sym] = arg.fun.expr
+                        unique_sym = ir_misc.unique_symbol(sym, [s.id for s in outer_vars.keys()])
+                        if sym != unique_sym:
+                            rename_map[sym.id] = unique_sym.id
+
+                        outer_vars[unique_sym] = val
+
+                    new_expr = arg.fun.expr
+                    if rename_map:
+                        new_expr = inline_lambda(im.let(*rename_map.items())(new_expr))
```
havogt: this line is worth a comment, I guess it does:
- in case we have the name in the other scope, we introduce a let that hides the name collision
- inline_lambda will remove the let, but keeps name of the let target, i.e. the unique name

src/gt4py/next/iterator/ir_utils/misc.py
tehrengruber: Indeed the `InlineLift` pass has another `unique_symbol` function. I'd rather not touch that one, since the pass is obsolete and will be replaced anyway. In the frontend we also have `unique_name` which uses the `ᐞ`, but works differently. I like that format more, but this is on the AST level not ITIR and I would need to build the mechanism.

havogt (approved):

PR #2079, opened 2025-06-16
fix[next]: Frontend test for negative domain temporary

havogt:
Adds a regression test, driven from the frontend, for the gtfn negative-domain temporary allocation bug fixed in #2092.

Triggering the bug requires an intermediate temporary with a negative domain start. To force one inside a chain of single-argument `as_fieldop`s (a 4x-nested laplacian), `fuse_as_fieldop` now stops inlining a single-arg applied fieldop once it accumulates too much overcompute (`len(shifts) >= 6`). Without this the whole chain fully fuses into one stencil and trips the recursion limit during compilation.

The codegen fix originally in this PR was merged separately in #2092 and has been dropped; only the heuristic and the test remain.

### DIFF
--- a/src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
+++ b/src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
@@ -208,7 +208,10 @@ def _arg_inline_predicate(node: itir.Expr, shifts: set[tuple[itir.OffsetLiteral,
     ) or cpm.is_call_to(node, "if_"):
         # always inline arg if it is an applied fieldop with only a single arg
         if is_applied_fieldop and len(node.args) == 1:
-            return True
+            # ... and not too much overcompute
+            # TODO(havogt): tune this value or improve heuristic
+            if len(shifts) < 6:
+                return True
         # argument is never used, will be removed when inlined
         if len(shifts) == 0:
             return True
--- a/tests/next_tests/integration_tests/multi_feature_tests/ffront_tests/test_laplacian.py
+++ b/tests/next_tests/integration_tests/multi_feature_tests/ffront_tests/test_laplacian.py
@@ -47,6 +47,11 @@ def laplap(in_field: gtx.Field[[IDim, JDim], "float"]) -> gtx.Field[[IDim, JDim]
     return lap(lap(in_field))
 
 
+@gtx.field_operator
+def laplaplaplap(in_field: gtx.Field[[IDim, JDim], "float"]) -> gtx.Field[[IDim, JDim], "float"]:
+    return laplap(laplap(in_field))
+
+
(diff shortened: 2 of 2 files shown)


src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -238,17 +238,21 @@ def visit_TemporaryAllocation(self, node: gtfn_ir.TemporaryAllocation, **kwargs:
         assert isinstance(node.domain, (gtfn_ir.CartesianDomain, gtfn_ir.UnstructuredDomain))
         assert node.domain.tagged_offsets.tags == node.domain.tagged_sizes.tags
         tags = node.domain.tagged_offsets.tags
-        new_sizes = []
-        for size, offset in zip(node.domain.tagged_offsets.values, node.domain.tagged_sizes.values):
```
havogt: here was a bug: offsets and sizes are swapped

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -231,24 +231,24 @@ def visit_FunctionDefinition(self, node: gtfn_ir.FunctionDefinition, **kwargs: A
     )
 
     def visit_TemporaryAllocation(self, node: gtfn_ir.TemporaryAllocation, **kwargs: Any) -> str:
-        # TODO(tehrengruber): Revisit. We are currently converting an itir.NamedRange with
-        #  start and stop values into an gtfn_ir.(Cartesian|Unstructured)Domain with
-        #  size and offset values, just to here convert back in order to obtain stop values again.
-        # TODO(tehrengruber): Fix memory alignment.
```
havogt: not sure about this todo

havogt:
we need
- a test
- that checks reliably for oob access

havogt:
Added a test that should always create a temporary (with domain starting in the negative, which was the bug). This will most likely crash before the fix with a segfault due to oob (but no guarantee).

src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
```diff
@@ -198,7 +198,10 @@ def _arg_inline_predicate(node: itir.Expr, shifts: set[tuple[itir.OffsetLiteral,
     ) or cpm.is_call_to(node, "if_"):
         # always inline arg if it is an applied fieldop with only a single arg
         if is_applied_fieldop and len(node.args) == 1:
-            return True
+            # ... and not too much overcompute
+            # TODO(havogt): tune this value or improve heuristic
```
havogt, line 212: leave 4xlap and this problem for later, but use a itir test for the temporary problem

PR #2080, opened 2025-06-17
refactor[cartesian]: fix mypy issues for pyext_builder

romanc:
## Description

`mypy` was configured to ignore all errors in `pyext_builder`. This was a temporary workaround (I guess from when `mypy` was first added). It was easy enough to "fix" what `mypy` was complaining about:

1. The overload that would "tells mypy to accept unpacking kwargs" wasn't really working. It also seems like the code isn't used, so I'm getting rid of it.
2. `mypy` can't know that `setuptools_args["script_args"]` is a list because `setuptools_args` is a dictionary that does also contain values that aren't lists. This is fine, we know, so we can just ignore the error with a line override.

## Requirements

  All good as long as the CI is still green.
  N/A

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -213,10 +213,6 @@ module = 'gt4py.cartesian.*'
 
 [[tool.mypy.overrides]]
 ignore_errors = true
-module = 'gt4py.cartesian.backend.pyext_builder'
-
-[[tool.mypy.overrides]]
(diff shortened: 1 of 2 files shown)


havogt (approved):
lgtm (in case you would like to get rid of the type ignores, you could use a TypedDict, but probably too much for this use-case)

PR #2081, opened 2025-06-17
bug[next]: Fix OOB access in scan operator

tehrengruber:


havogt:
After thinking about it again this workaround makes cases work that are semantically wrong, but we don't have a nice way to express it semantically correct. So tests for this case should actually not necessarily pass.

PR #2084, opened 2025-06-18
fix[next]: offset_provider optional

havogt:
Makes the offset_provider optional in cases where previously we required `offset_provider={}`.

Additional: add the implicit offsetprovider in field_operator as programs call.

### DIFF
--- a/src/gt4py/next/ffront/decorator.py
+++ b/src/gt4py/next/ffront/decorator.py
@@ -54,6 +54,30 @@ from gt4py.next.type_system import type_info, type_specifications as ts, type_tr
 DEFAULT_BACKEND: next_backend.Backend | None = None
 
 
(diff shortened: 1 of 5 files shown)


src/gt4py/next/ffront/decorator.py
```diff
@@ -749,6 +755,8 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any:
                 **kwargs,
             )
         else:
+            if not next_embedded.context.within_valid_context():
```
havogt, line 772: add comment

havogt, line 772:
```suggestion
            if not next_embedded.context.within_valid_context():
                # field_operator as program
```

havogt:
> TODO: write tests in test_arg_call_interface

deleted the offset_provider where it was not needed in test_arg_call_interface

tehrengruber (approved):

PR #2085, opened 2025-06-19
fix[next][dace]: Fix compile time symbol substitution

iomaganaris:
- In `gt_substitute_compiletime_symbols` when we call `gt_simplify` we have to make sure that `ScalarToSymbolPromotion` is executed before the `ConstantPropagation` pass
  - If we don't pass any option in `skip` then by default `ScalarToSymbolPromotion` and `ConstantPropagation` are skipped in `gt_simplify`
  - We can skip `ConstantPropagation` when we call `gt_simplify` since this is called right after

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
@@ -318,6 +318,7 @@ def gt_substitute_compiletime_symbols(
             sdfg=sdfg,
             validate=validate,
             validate_all=validate_all,
+            skip=["ConstantPropagation"],  # Avoid skipping "ScalarToSymbolPromotion"
         )
 
     # We will use the `replace` function of the top SDFG, however, lower levels


src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
```diff
@@ -318,6 +318,7 @@ def gt_substitute_compiletime_symbols(
             sdfg=sdfg,
             validate=validate,
             validate_all=validate_all,
+            skip=["ConstantPropagation"],  # Avoid skipping "ScalarToSymbolPromotion"
```
philip-paul-mueller:
```suggestion
            skip=GT_SIMPLIFY_DEFAULT_SKIP_SET.difference(["ScalarToSymbolPromotion"]),
```

You should also add a comment why you run the transformation, because we blocked it since it is supposed to be evil.

iomaganaris: It seems like it's necessary evil 😅

PR #2086, opened 2025-06-20
chore: use precommit local hooks with uv and update dev scripts

egparedes:
Update pre-commit configuration to use local hooks for python tools tracked by  `uv` to keep versions in sync without duplication of the version information.

Additionally, put in place a new developer-friendly infrastructure to run custom development scripts in a simple and documented way. For now, only a script to update dependencies has been added. The main entry point for the scripts is the `./dev-scripts.py` file. If run without options (or with the `--help` option) displays the usage help for all commands, and the same applies for the subcommands. For example:

```bash
./dev-scripts.py --help    # Show list of commands
./dev-scripts.py update --help    # Show usage for the 'update' command (including list of subcommands)
./dev-scripts.py update pre-commit --help    # Show usage help for the pre-commit 'subcommand' of the 'update' command
```

Finally, update the versions of tools in the pre-commit hooks which affects the formatting of several source code files.


egparedes:
> LGTM, but I would appreciate some examples in the PR description of commands to run the new **update** script.

@edopao I updated the PR description according to your suggestion. Does it work for you now? Note that the commands I'm adding (here and in the related #2087 PR) have full usage documentation to make them trivial to use without extra documentation.

.pre-commit-config.yaml
```diff
@@ -1,85 +1,99 @@
...
 minimum_pre_commit_version: 3.8.0
 repos:
-# - repo: meta
-#   hooks:
-#   - id: check-hooks-apply
-#   - id: check-useless-excludes
+  # - repo: meta
+  #   hooks:
+  #   - id: check-hooks-apply
+  #   - id: check-useless-excludes
```
romanc: These have been commented before. Is anyone using/un-commenting them on a regular basis or could they be deleted?

.pre-commit-config.yaml
```diff
@@ -1,85 +1,99 @@
...
-- repo: https://github.com/gauge-sh/tach-pre-commit
-  rev: v0.23.0
-  hooks:
-  - id: tach
+      - <<: *uv-managed-hook
+        id: ruff-check
+        name: ruff checker
+        entry: uv run --frozen --isolated --group dev ruff check --force-exclude --fix
+        types_or: [python, pyi, jupyter]
+        files: ^src/ # TODO(egparedes): add the `tests` folder here
```
romanc, line 67: Is the `tests/` folder supposed to come in a follow-up PR? Or was the idea to include it in this PR?

romanc (approved):
LGTM - just two questions to clarify commented code / inserted TODO statements.

.pre-commit-config.yaml
egparedes: This TODO already existed, and I would like to tackle it soon but it should go in another PR since it will involve a lot of format and code changes or a lot of `noqa`s comments.

.pre-commit-config.yaml
egparedes: Good point. I've used them in the past to debug and check the pre-commit configuration locally, but left them commented out because they were complaining about something that didn't really like to remove from the config file and otherwise they would make pre-commit fail. I've tested them again now and they behave well, so they are now active.

romanc: :tada:

.pre-commit-config.yaml
romanc: Ah okay - all good then. Let's sync if in case you want help with this.

PR #2090, opened 2025-06-21
refactor[cartesian]: Delete CLIBackenMixin and gt4py/cartesian/cli.py

romanc:
## Description

According to @egparedes, 

> [The CLI backend] was a tool to use gt4py (cartesian) as a c++ code generator, in case the user didn't want to call the generate code from python. I think basically it reads python files with gt4py stencils and uses some backend to generate files with c++ code.
> The original use-case was a way for MeteoSwiss to use gt4py as a replacement of another tool they had doing something similar.
> I don't think it has ever been used and IMHO it should be perfectly fine to remove it.

Since this code generator is not used anymore, we'd like to get rid of it to avoid having to maintain unused code. To be discussed in the upcoming gt4py community meeting. Note that this is a breaking change.

## Requirements

  Tests referencing the cli feature have been deleted.
  N/A


romanc:
From the notes of last week's community meeting, I'm taking it that nobody has an issue with this PR. I'll thus go ahead in prepping this PR to be merged soon.

PR #2094, opened 2025-06-23
fix[next][dace]: Missing symbolic args in lambda nested scope

edopao:
Some scalar arguments of a lambda nested SDFG were not represented as symbols, although used in domain symbolic expressions. The unittest `test_concat_where.py:test_lap_like` In the `concat_where` PR (#1713) was failing in lowering to SDFG because of this bug.


philip-paul-mueller (approved):
LGTM

PR #2096, opened 2025-06-23
feat[next]: synchronous and synchronize compilation

havogt:
Adds:
- synchronous compilation if `GT4PY_BUILD_JOBS=0`.
- `wait_for_compilation()` waits for all pending compilations to finish

### DIFF
--- a/src/gt4py/next/otf/compiled_program.py
+++ b/src/gt4py/next/otf/compiled_program.py
@@ -24,12 +24,37 @@ from gt4py.next.type_system import type_info, type_specifications as ts, type_tr
 
 
 # TODO(havogt): We would like this to be a ProcessPoolExecutor, which requires (to decide what) to pickle.
-_async_compilation_pool = concurrent.futures.ThreadPoolExecutor(max_workers=config.BUILD_JOBS)
+_async_compilation_pool: concurrent.futures.Executor | None = None
+
+
+def _init_async_compilation_pool() -> None:
+    global _async_compilation_pool
+    if _async_compilation_pool is None and config.BUILD_JOBS > 0:
+        _async_compilation_pool = concurrent.futures.ThreadPoolExecutor(
+            max_workers=config.BUILD_JOBS
+        )
+
+
+_init_async_compilation_pool()
 
 ScalarOrTupleOfScalars: TypeAlias = extended_typing.MaybeNestedInTuple[core_defs.Scalar]
 CompiledProgramsKey: TypeAlias = tuple[
     tuple[ScalarOrTupleOfScalars, ...], common.OffsetProviderType
 ]
+
+
+def wait_for_compilation() -> None:
+    """
+    Waits for all ongoing compilations to finish.
+
+    This is useful to ensure that all compiled programs are ready before
+    proceeding with further operations. E.g. when the first call is included in timings.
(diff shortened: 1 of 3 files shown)


src/gt4py/next/otf/compiled_program.py
```diff
@@ -214,9 +239,15 @@ def _compile_variant(
...
-        self._compiled_programs[key] = _async_compilation_pool.submit(
-            self.backend.compile, self.definition_stage, compile_time_args=compile_time_args
-        )
+        if _async_compilation_pool is None:
+            # synchronous compilation
+            self._compiled_programs[key] = self.backend.compile(
+                self.definition_stage, compile_time_args=compile_time_args
+            )
+        else:
+            self._compiled_programs[key] = _async_compilation_pool.submit(
```
havogt: - think about adding a warning in case the async compilation fails (or even better add mechanism that errors in the main thread)
- ~~add no-jit option~~ we have it already...
both is to prevent that a `compile()`ed program is not used by accident, e.g. because a different offset_provider is used in the call.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -360,9 +385,15 @@ def _compile_variant(
...
+                self.backend.compile, self.definition_stage, compile_time_args=compile_time_args
+            )
```
egparedes: I would try to avoid duplicating the compile call logic using something like `functools.partial`:
```suggestion
        compile_call = functools.partial(
            self.backend.compile, self.definition_stage, compile_time_args=compile_time_args
        )
        if _async_compilation_pool is None:
            # synchronous compilation
            self._compiled_programs[key] = compile_call()
        else:
            self._compiled_programs[key] = _async_compilation_pool.submit(compile_call)
```

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_compiled_program.py
```diff
@@ -812,3 +815,45 @@ def test_compile_variants_tuple(cartesian_case, compile_variants_testee_tuple):
...
+
+        start_wait = time.time()
+        gtx.wait_for_compilation()
+        end_wait = time.time()
+        wait_time = end_wait - start_wait
+
+        if synchronous:
+            assert compile_call_time > wait_time
+        else:
+            assert wait_time > compile_call_time
```
egparedes, line 864: Are you sure `wait_time > compile_call_time` will always be true for any backend in the asynchronous case? Anyway, it's worth explaining this logic quickly with a comment

egparedes (approved):
Just a couple of non-blocking comments, so approving the PR now.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_compiled_program.py
havogt: ah, it seems I deleted a comment before commit

havogt: I think so (note embedded is excluded) as there is always some passes being run and `compile` should return quickly. Of course we consider hash computations in compile as non-critical and this could break in the future... I added a comment.

PR #2097, opened 2025-06-23
feat[next]: refreshing_lock

havogt:
Wraps flufl.lock into a contexthandler that does refreshing.

Previously the compilation lock had an arbitrary and huge lifetime of 600s. In case compilation is aborted and restarted, the lock would be blocked for up to 600s. Now the default lifetime is 10s with refresh every 5s.

### DIFF
--- a/src/gt4py/eve/lock.py
+++ b/src/gt4py/eve/lock.py
@@ -15,7 +15,9 @@ from flufl import lock as flufl_lock
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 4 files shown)


src/gt4py/eve/lock.py
```diff
@@ -0,0 +1,32 @@
+# GT4Py - GridTools Framework
```
havogt, line 1: or in `gt4py._core`?

PR #2100, opened 2025-06-25
feat: Python 3.12 support

havogt:


noxfile.py
```diff
@@ -146,6 +153,7 @@ def test_cartesian(
         "*.toml",
         "*.yml",
         "noxfile*.py",
+        ".python-versions",
```
egparedes: Either we add this to all sessions or to none of them, but only add it to `cartesian` doesn't make sense to me. Personally, I wouldn't add anything since if we change the supported python versions we are also changing the `pyproject.toml` (and most likely `noxfile.py` too), so it shouldn't be needed to trigger all tests.

```suggestion
```

ci/cscs-ci.yml
```diff
@@ -1,6 +1,9 @@
 include:
   - remote: 'https://gitlab.com/cscs-ci/recipes/-/raw/master/templates/v2/.ci-ext.yml'
 
+.py312: &py312
```
egparedes, line 4: Why not switching 3.11 to 3.12 instead of just adding 3.12? Adding more jobs to CI will make the CI queue grow even longer...

egparedes (commented):
Just two comments, otherwise looks good to me.

noxfile.py
havogt: I remove it, we always have to touch pyproject.toml if we touch this file. In the beginning some jobs where skipped because I only changed .python-versions

egparedes (approved):
LGTM

PR #2108, opened 2025-06-30
feat: Python 3.13 support

havogt:
Adds support for Python 3.13.

- Bump minimum versions of matplotlib, atlas4py, pyyaml, coverage[toml], cytoolz, deepdiff, mako, numpy, typing-extensions, xxhash, scipy, black for compatibility with Python 3.13.
- gh action get-python-versions return lowest and highest by default; all versions if `all = true` is set; daily gh action uses all, other actions use lowest and highest
- cscs ci: run python 3.13 instead of 3.12
- fixes some typing related changes in dependency
- fixes a bug in type_translation triggered by changes in Python 3.13
- adds a workaround in `test_cli.py` (the file will be deleted in #2090)


ci/cscs-ci.yml
```diff
@@ -1,8 +1,8 @@
 include:
   - remote: 'https://gitlab.com/cscs-ci/recipes/-/raw/master/templates/v2/.ci-ext.yml'
 
-.py312: &py312
-  PYVERSION_PREFIX: py312
+.py313: &py313
+  PYVERSION_PREFIX: py313
   PYVERSION: 3.12.7
```
havogt:
```suggestion
  PYVERSION: 3.13.3
```

noxfile.py
```diff
@@ -153,6 +165,7 @@ def test_cartesian(
         "*.toml",
         "*.yml",
         "noxfile*.py",
+        ".python-versions",
```
egparedes: Not needed as discussed in another PR
```suggestion
```

noxfile.py
```diff
@@ -235,7 +248,12 @@ def test_next(
 
     nox_utils.install_session_venv(
         session,
-        extras=["performance", "testing", *codegen_settings["extras"], *device_settings["extras"]],
+        extras=[
+            "performance",
+            "testing",
+            *codegen_settings["extras"],
+            *device_settings["extras"],
```
egparedes: Did you add the trailing comma here on purpose? If not, you can try removing the trailing comma again to undo the automatic formatting
```suggestion
            *device_settings["extras"]
```

noxfile.py
```diff
@@ -299,7 +317,9 @@ def test_storage(
     device_settings = DeviceTestSettings[device]
 
     nox_utils.install_session_venv(
-        session, extras=["performance", "testing", *device_settings["extras"]], groups=["test"]
+        session,
```
egparedes: Same as above

noxfile.py
havogt: Probably happened when I accidentally bumped the ruff version

PR #2109, opened 2025-06-30
feat[core]: lock api

havogt:
Adds `locking` module to gt4py._core which provides a simple file lock API: `lock(path)`.

The default lock is switched back to `filelock` as
- Lustre should support flock
- flufl.lock didn't seem to work nicely for our use-cases (e.g. performance not perfect and link_count errors)

This PR already allows to customize the locker by overwriting `locking._default_lock` but it's not currently publicly exposed.


src/gt4py/_core/locking.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import ContextManager, Protocol
+
+import filelock
+
+
+class _LockConstructor(Protocol):
+    def __call__(self, path: str, /) -> ContextManager: ...
+
+
+_default_lock: _LockConstructor = filelock.FileLock
```
egparedes: Maybe document that this is private/experimental API for now but might be made public in the future.

src/gt4py/_core/locking.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def lock(path: pathlib.Path | str) -> ContextManager:
```
egparedes: Another option  be to pass here the lock_constructor as as keyword argument with a default set to `_default_lock`, right? Or do do you think it could error prone?

```suggestion
def lock(path: pathlib.Path | str, *, lock_constructor: _LockConstructor = _default_lock) -> ContextManager:
```

egparedes (commented):
I think it makes sense, just a couple of comments.

src/gt4py/_core/locking.py
havogt: What advantage do you see? Probably it would be simpler to directly use the custom lock?
I mean why
```
with gt4py.lock(path, lock_constructor=mylock):
    ...
```
over 
```
with mylock(path):
    ...
```

src/gt4py/_core/locking.py
havogt:
```suggestion
# Changing the default is private API, but could be made public in the future
_default_lock: _LockConstructor = filelock.FileLock
```

edopao (commented):
LGTM

src/gt4py/_core/locking.py
egparedes: I think you're right and my suggestion doesn't have any advantage. Thinking about it, I got distracted by the `_default_lock` name since it's not really a default but the __only__ lock constructor. Woud it be better to rename it to `_lock_constructor`?

edopao (approved):

src/gt4py/_core/locking.py
```diff
@@ -0,0 +1,36 @@
...
+    def __call__(self, path: str, /) -> ContextManager: ...
+
+
+# Changing the lock is private API, but could be made public in the future
+# The second element (str) is an identifier of the lock implementation to avoid conflicts
+# when switching between different lock implementations.
+_lock: tuple[_LockConstructor, str] = (filelock.FileLock, "filelock")
+
+
+def lock(directory: pathlib.Path | str, suffix: str = ".lock") -> ContextManager:
```
havogt, line 30: maybe directory is too restrictive?

egparedes: What do you mean by restrictive? `directory` here is supposed to be the parent folder for the lock file, right? Maybe renaming it could make it more evident

src/gt4py/_core/locking.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class _LockConstructor(Protocol):
+    def __call__(self, path: str, /) -> ContextManager: ...
+
+
+# Changing the lock is private API, but could be made public in the future
+# The second element (str) is an identifier of the lock implementation to avoid conflicts
+# when switching between different lock implementations.
+_lock: tuple[_LockConstructor, str] = (filelock.FileLock, "filelock")
```
egparedes: I don't think the custom identifier is really needed. An identifier can be easily computed from the lockConstructor class name. For example using the "abbreviation" (a10n) from eve utils (or directly the original implementation from boltons: https://boltons.readthedocs.io/en/latest/strutils.html#boltons.strutils.a10n)
```python
identifier = f"{eve.utils.a10n(_lock.__module__)}_{eve.utils.a10n(_lock.__class__.__name__)}"
```

```

egparedes (commented):
Minor questions

src/gt4py/_core/locking.py
havogt: Yes, so basically with this interface the idea is to lock a directory. However you could also have multiple locks for different purposes in the same directory. That's not our use-case, still it restricts the general file locking to lock a directory. It's a minor style issue, because you can still have multiple locks in the same directory by changing the suffix...

src/gt4py/_core/locking.py
havogt: thanks, I wanted to something like this, but didn't know which hash to use.

havogt: I used `_lock.__name__` instead of `_lock.__class__.__name__` to make it also useful if `_lock` is a function

egparedes (approved):
LGTM

havogt:
I thought about the name again in my last commit, but was not obvious why lock_constructor is better. I don't change now because I don't want to wait for another ci run, but I am not opposed to changing it later.

PR #2112, opened 2025-07-01
fix[next][dace]: Update parent reference on cloned NestedSDFG

edopao:


src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -293,6 +293,11 @@ def _replicate_dataflow_into_branch(
         new_nodes: dict[dace_nodes.Node, dace_nodes.Node] = {
             old_node: copy.deepcopy(old_node) for old_node in nodes_to_move
         }
+        for node in new_nodes.values():
+            if isinstance(node, dace_nodes.NestedSDFG):
+                # ensure the correct reference to parent
+                node.sdfg.parent_nsdfg_node = node
```
philip-paul-mueller: I would also set `parent_sdfg`, this is what `SDFGState::add_node()` does.

However, it is strange, `add_node()` already does this, however, it only does it, when `NestedSDFG.sdfg` is set to something other than `None` and it does not override it, which I think is strange.
This should be a question for the chat.

edopao: On a second thought, `parent_sdfg` should remain unchanged.

PR #2116, opened 2025-07-03
feat[next]: enable_jit on Program.__call__

havogt:
Instead of setting `enable_jit` on the program definition, we should set it on `__call__` for more flexibility.

### DIFF
--- a/src/gt4py/next/config.py
+++ b/src/gt4py/next/config.py
@@ -112,4 +112,4 @@ COLLECT_METRICS_LEVEL: int = env_flag_to_int("GT4PY_COLLECT_METRICS_LEVEL", defa
 
 #: The default for whether to allow jit-compilation for a compiled program.
 #: This default can be overriden per program.
-DEFAULT_ENABLE_JIT: bool = env_flag_to_bool("GT4PY_ENABLE_JIT_DEFAULT", default=True)
+ENABLE_JIT_DEFAULT: bool = env_flag_to_bool("GT4PY_ENABLE_JIT_DEFAULT", default=True)
--- a/src/gt4py/next/ffront/decorator.py
+++ b/src/gt4py/next/ffront/decorator.py
@@ -106,7 +106,6 @@ class Program:
     connectivities: Optional[
         common.OffsetProvider
     ]  # TODO(ricoh): replace with common.OffsetProviderType once the temporary pass doesn't require the runtime information
-    enable_jit: bool
     static_params: (
         Sequence[str] | None
     )  # if the user requests static params, they will be used later to initialize CompiledPrograms
@@ -123,7 +122,6 @@ class Program:
         definition: types.FunctionType,
         backend: next_backend.Backend | None,
         grid_type: common.GridType | None = None,
-        enable_jit: bool = config.DEFAULT_ENABLE_JIT,
         static_params: Sequence[str] | None = None,
         connectivities: Optional[
             common.OffsetProvider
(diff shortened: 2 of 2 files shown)


havogt:
Or alternatively implement:
- if `enable_jit` is explicitly set on the program (i.e. not just defaulted), then use that as the default in the `__call__`

We should additionally enable config to be modifiable with contextmanager + contextvar.

src/gt4py/next/config.py
```diff
@@ -112,4 +112,4 @@ def env_flag_to_int(name: str, default: int) -> int:
 
 #: The default for whether to allow jit-compilation for a compiled program.
 #: This default can be overriden per program.
-DEFAULT_ENABLE_JIT: bool = env_flag_to_bool("GT4PY_ENABLE_JIT_DEFAULT", default=True)
+ENABLE_JIT_DEFAULT: bool = env_flag_to_bool("GT4PY_ENABLE_JIT_DEFAULT", default=True)
```
havogt, line 115: this is just a fix (we changed the name of the env var, but not the internal variable).

havogt:
And also add it to `compile`

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_compiled_program.py
```diff
@@ -436,6 +436,62 @@ def test_compile_variants_not_compiled(cartesian_case, compile_variants_testee):
...
+    assert np.allclose(out[0].ndarray, field_a.ndarray - 3)
+    assert np.allclose(out[1].ndarray, field_b.ndarray - 4.0)
+
+
+def test_compile_variants_config_default_disable_jit(cartesian_case, compile_variants_testee):
+    field_a = cases.allocate(cartesian_case, compile_variants_testee, "field_a")()
+    field_b = cases.allocate(cartesian_case, compile_variants_testee, "field_b")()
+    out = cases.allocate(cartesian_case, compile_variants_testee, "out")()
+
+    # the default in config is to allow jit (if that changes, this test has to change)
```
egparedes: Why not just patching the default to make sure is True as it is done in the False case?

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_compiled_program.py
```diff
@@ -436,6 +436,62 @@ def test_compile_variants_not_compiled(cartesian_case, compile_variants_testee):
...
+    cartesian_case, compile_variants_testee
+):
+    # disable jit on the program
+    object.__setattr__(compile_variants_testee, "enable_jit", False)
+
+    field_a = cases.allocate(cartesian_case, compile_variants_testee, "field_a")()
+    field_b = cases.allocate(cartesian_case, compile_variants_testee, "field_b")()
+    out = cases.allocate(cartesian_case, compile_variants_testee, "out")()
+
+    compile_variants_testee(
```
egparedes, line 461: Shouldn't be also tested here that if `enable_hit` is  not passed  at runtime the call fails?

egparedes (commented):
LGTM, just two questions

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_compiled_program.py
havogt: Makes sense. Since the test is about changing the `config` default before call, I thought I have to rely on the current value, but obviously one of the 2 will always be the non-default.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_compiled_program.py
havogt: makes sense. done

egparedes (approved):
LGTM

PR #2087, opened 2025-06-20
ci: use dynamic test matrix definitions in CI via scripts

egparedes:
## Requirements

- [ ] All fixes and/or new features come with corresponding tests.
- [ ] Important design decisions have been documented in the appropriate ADR inside the [docs/development/ADRs/](docs/development/ADRs/README.md) folder.

If this PR contains code authored by new contributors please make sure:

- [ ] The PR contains an updated version of the `AUTHORS.md` file adding the names of all the new contributors.

PR #2119, opened 2025-07-03
Extract base cscs-ci changes from PR #2087 without the dynamic pipeline

egparedes:
## Requirements


ci/cscs-ci.yml
```diff
@@ -1,185 +1,135 @@
...
     - if: $SUBPACKAGE == 'next' && $VARIANT == 'dace' && $DETAIL == 'nomesh'
       variables:
         # TODO: investigate why the dace tests seem to hang with multiple jobs
         GT4PY_BUILD_JOBS: 1
-        SLURM_TIMELIMIT: '00:15:00'
+        SLURM_TIMELIMIT: "00:15:00"
     - when: on_success
+  script:
+    # Since the image does not contain the repo, we need to clone it before running the tests
+    #- mkdir -p "${WORKDIR}/gt4py" && git clone --depth 1 --revision "${CI_COMMIT_SHA}" "${CSCS_CI_ORIG_CLONE_URL}" "${WORKDIR}/gt4py"
```
havogt: this didn't work? do I read correctly that you do clone default branch + fetch current branch now?

ci/cscs-ci.yml
```diff
@@ -1,185 +1,135 @@
...
   rules:
     - if: $SUBPACKAGE == 'next' && $VARIANT == 'dace' && $DETAIL == 'nomesh'
       variables:
         # TODO: investigate why the dace tests seem to hang with multiple jobs
         GT4PY_BUILD_JOBS: 1
-        SLURM_TIMELIMIT: '00:15:00'
+        SLURM_TIMELIMIT: "00:15:00"
     - when: on_success
+  script:
+    # Since the image does not contain the repo, we need to clone it before running the tests
```
havogt, line 118: Again I am wondering if this could be a introduces as a change in cicd-ext. Why are we mirroring to gitlab, if we can just clone the repo from github? Or does this script have to be on gitlab? Or maybe it's a historic thing?

ci/cscs-ci.yml
egparedes: The commented version works, but requires git >= 2.49, which is not available in the docker image (and I miserably failed when I tried to install it in the dockerfile). The current line is the best thing I could find to download the minimal amount of information from the git repo that works with older git versions.

ci/cscs-ci.yml
```diff
@@ -1,185 +1,139 @@
...
-  # image, they can be added too.
-  # Since the base image name is runtime dependent, we need to carry the value of it to
-  # the following jobs via a dotenv file.
+# -- Build stage --
+# The build job creates a the base image for the CSCS CI only if the external
+# dependencies change. The actual repository is cloned at the test stage.
+.build_common:
+  stage: build
+  variables:
+    BASE_IMAGE: docker.io/ubuntu:${UBUNTU_VERSION}
```
finkandreas: This could be `jfrog.svc.cscs.ch/dockerhub/ubuntu:${UBUNTU_VERSION}`, which is proxied through JFrog, and cached. Not really a big saving when it comes to network, when the image is as tiny as an ubuntu base image.

ci/cscs-ci.yml
```diff
@@ -1,185 +1,139 @@
...
-    # are running.
-    PERSIST_IMAGE_NAME: $CSCS_REGISTRY_PATH/public/$ARCH/gt4py/gt4py-ci:$CI_COMMIT_SHA-$PYVERSION
-    DOCKERFILE: ci/checkout.Dockerfile
-    DOCKER_BUILD_ARGS: '["PYVERSION=$PYVERSION", "BASE_IMAGE=${BASE_IMAGE_${PYVERSION_PREFIX}}"]'
-.build_image_x86_64:
-  extends: [.container-builder-cscs-zen2, .build_image]
-.build_image_aarch64:
-  extends: [.container-builder-cscs-gh200, .build_image]
-
+    BASE_IMAGE: docker.io/nvidia/cuda:${CUDA_VERSION}-devel-ubuntu${UBUNTU_VERSION}
```
finkandreas: This could be `jfrog.svc.cscs.ch/dockerhub/nvidia/cuda:${CUDA_VERSION}-devel-ubuntu${UBUNTU_VERSION}`, which is proxied through JFrog, and cached. Network is then all CSCS internal and probably faster.

ci/cscs-ci.yml
```diff
@@ -1,185 +1,139 @@
...
+# TODO: rocm steps are in draft state for now to show how to add support in the future
+# .build_extra_rocm:
 #   variables:
-#     <<: *py313
-build_py313_image_aarch64:
-  extends: .build_image_aarch64
-  needs: [build_py313_baseimage_aarch64]
-  variables:
-    <<: *py313
+#     BASE_IMAGE: docker.io/rocm/dev-ubuntu-${UBUNTU_VERSION}:${ROCM_VERSION}-complete
```
finkandreas: can be proxied through jfrog, like the others.

ci/cscs-ci.yml
finkandreas: We are using gitlab-runners, which only work with gitlab, hence the need for a gitlab mirror.

Also all this syntax is parsed by gitlab, not cicd-ext.

finkandreas: To be clear, you could instruct gitlab-runner to clone the source code for you by setting the variable `GIT_STRATEGY: fetch` (by default `GIT_STRATEGY: none` for container-runner).

The downside of letting the gitlab-runner do the cloning will be that the source code is on $SCRATCH, and you might run out of inodes quickly.

ci/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+if [ ! -z "${EXTRA_UV_ENV_VARS}" ]; then
+    eval "export ${EXTRA_UV_ENV_VARS}"
+fi
+uv pip install --compile-bytecode setuptools wheel pip
+echo "EXTRA_UV_SYNC_ARGS=${EXTRA_UV_SYNC_ARGS}"
+uv sync --compile-bytecode --frozen --no-install-project \
+    --group dev \
+    --extra formatting \
+    --extra jax \
+    --extra performance \
```
edopao, line 68: On a second review, I understand now. These extras come from the repo on the mirror. Very nice, we save some time in the test stage.

ci/cscs-ci.yml
```diff
@@ -1,185 +1,142 @@
...
+    DOCKER_BUILD_ARGS: '["BASE_IMAGE", "CACHE_DIR", "EXTRA_APTGET", "EXTRA_UV_ENV_VARS", "EXTRA_UV_PIP_ARGS", "EXTRA_UV_SYNC_ARGS", "PY_VERSION", "UV_VERSION", "WORKDIR_PATH" ]'
   before_script:
-  # include build arguments in hash since we use a parameterized Docker file
-    - DOCKER_TAG=`echo "$(cat $DOCKERFILE) $DOCKER_BUILD_ARGS" | sha256sum | head -c 16`
-    - export PERSIST_IMAGE_NAME=$CSCS_REGISTRY_PATH/public/$ARCH/base/gt4py-ci:$DOCKER_TAG-$PYVERSION
-    - echo "BASE_IMAGE_${PYVERSION_PREFIX}=$PERSIST_IMAGE_NAME" >> build.env
+    # We create a tag that depends on the SHA value of ci/Dockerfile,
+    # the docker build arguments since the Docker file is heavily parameterized,
+    # and the contents of the `uv.lock` with the python dependencies versions.
+    - DOCKER_TAG=`echo "$(cat ${DOCKERFILE}) ${DOCKER_BUILD_ARGS} $(cat uv.lock)" | sha256sum | head -c 16`
```
edopao: I see `uv.lock` in the list, so this uv.lock is the one on the mirror, right?

edopao (commented):
Very nice!

ci/Dockerfile
egparedes: Yes, the build stage is run from a checked out copy of the repository branch, so all files there are available for docker/podman build and for scripts. At the test stage this is not the case in our setup, so we explicitly clone the repo inside the container, to run the tests for the specific commit we want to test.

ci/cscs-ci.yml
egparedes: Yes, see my previous comment.

ci/cscs-ci.yml
egparedes: @finkandreas I'm now thinking: could it be possible to use another environment variable to define the target location for the git clone? That could solve the issue with the inodes and simplify users' CI code, right?

finkandreas: The problem is that the gitlab runner is running in different stages (with a stage I mean a bash script that is being run). I.e. there is a bash script that is doing the clone, then there is a bash script that is doing the run (i.e. whatever you have defined in `script`), then there is a bash script that is uploading artifacts.
It will go through all the different stages step by step, maybe repeating the stage, if it failed.

Having all this said: Running inside a container (maybe even with multiple ranks) does not support to execute many processes, one after the other. I.e. our container runtime does not allow to spawn a container, and then execute one process after the other inside that container. It is also a bit unclear how a multi node setup then would look like.

ci/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+ENV LC_ALL=C.UTF-8
+
+ARG WORKDIR_PATH=/workspace
+ENV WORKDIR=${WORKDIR_PATH}
+WORKDIR ${WORKDIR}
+
+ARG DEBIAN_FRONTEND=noninteractive
+ARG EXTRA_APTGET=""
+RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
+    # strace \
```
havogt: remove commented packages

havogt (approved):

ci/cscs-ci.yml
```diff
@@ -1,185 +1,143 @@
...
-    - echo "GITHUB_PR=${GITHUB_PR}"
-    - export CI_NOX_RUN_ONLY_IF_CHANGED_FROM="gh/$(curl https://api.github.com/repos/gridtools/gt4py/pulls/${GITHUB_PR} 2>/dev/null | jq -r '.base.ref')"
-    - echo "CI_NOX_RUN_ONLY_IF_CHANGED_FROM=${CI_NOX_RUN_ONLY_IF_CHANGED_FROM}"
-    - NOX_SESSION_ARGS="${VARIANT:+($VARIANT}${SUBVARIANT:+, $SUBVARIANT}${DETAIL:+, $DETAIL}${VARIANT:+)}"
-    - nox -s "test_$SUBPACKAGE-${PYVERSION:0:4}$NOX_SESSION_ARGS"
+  image: ${GT4PY_IMAGE_PREFIX}-${PY_VERSION}
   variables:
     CSCS_CUDA_MPS: 1
     SLURM_JOB_NUM_NODES: 1
     SLURM_TIMELIMIT: 5
```
havogt, line 93: because of santis misconfiguration

```suggestion
    SLURM_TIMELIMIT: 5
    SLURM_GRES: "gpu:4"
```
or 1

ci/cscs-ci.yml
```diff
@@ -1,185 +1,144 @@
...
-#     - SUBPACKAGE: [cartesian, storage]
-#       VARIANT: [-internal, -dace]
-#       SUBVARIANT: [-cuda11x, -cpu]
-#     - SUBPACKAGE: eve
-#     - SUBPACKAGE: next
-#       VARIANT: [-nomesh, -atlas]
-#       SUBVARIANT: [-cuda11x, -cpu]
-.test_helper_aarch64:
-  extends: [.container-runner-santis-gh200, .test_helper]
+    SLURM_GRES: "gpu:4"
```
egparedes:
```suggestion
    SLURM_GPUS_PER_NODE: 4
```

ci/cscs-ci.yml
```diff
@@ -1,185 +1,144 @@
...
-#     - SUBPACKAGE: [cartesian, storage]
-#       VARIANT: [-internal, -dace]
-#       SUBVARIANT: [-cuda11x, -cpu]
-#     - SUBPACKAGE: eve
-#     - SUBPACKAGE: next
-#       VARIANT: [-nomesh, -atlas]
-#       SUBVARIANT: [-cuda11x, -cpu]
-.test_helper_aarch64:
-  extends: [.container-runner-santis-gh200, .test_helper]
+    SLURM_GPUS_PER_NODE: 4
```
edopao: I am testing this other setting: https://github.com/GridTools/gt4py/pull/2133

edopao: but you are probably right, we should set `SLURM_GPUS_PER_NODE`

PR #2121, opened 2025-07-07
ci: fix GitHub ci on push events to main

egparedes:
Fix the "Test code components" workflow when the trigger event is not a pull request by running all test sessions instead of only the sessions affected by the source code changes.

### DIFF
--- a/.github/workflows/test-components.yml
+++ b/.github/workflows/test-components.yml
@@ -35,7 +35,6 @@ jobs:
         with:
           enable-cache: true
           cache-dependency-glob: "uv.lock"
-          # python-version: "3.12" # Use a fixed Python version for defining test sessions
 
       - name: Define test sessions based on changed files
         id: test-sessions


.github/workflows/test-components.yml
```diff
@@ -39,11 +40,21 @@ jobs:
       - name: Define test sessions base on changed files
         id: test-sessions
         run: |
-          ./scripts-cli.py github-ci matrix \
-            --config nox-sessions-config.yml \
-            --base "origin/${{ github.base_ref }}" \
-            --output gh-test-sessions.json \
-            --verbose
+          #if [ "${{ github.event_name }}" == "push" ]; then
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
```
havogt: maybe we can actually leave this in to have a way to run all
```suggestion
          if [ "${{ github.event_name }}" == "push" or "${{ github.event_name }}" == "workflow_dispatch" ]; then
```

.github/workflows/test-components.yml
```diff
@@ -36,14 +37,23 @@ jobs:
           cache-dependency-glob: "uv.lock"
           # python-version: "3.12" # Use a fixed Python version for defining test sessions
```
romanc: unrelated: is this comment still valid? looks like a leftover from the last commit

.github/workflows/test-components.yml
```diff
@@ -36,14 +37,23 @@ jobs:
...
+      - name: Define test sessions based on changed files
         id: test-sessions
         run: |
-          ./scripts-cli.py github-ci matrix \
-            --config nox-sessions-config.yml \
-            --base "origin/${{ github.base_ref }}" \
-            --output gh-test-sessions.json \
-            --verbose
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
+            # Run only sessions related to the changes
```
romanc, line 43: In theory, could you move the `if` to the step level or would mess up something down the line? I mean something like

```yaml
  - name: "PR: Define test sessions based on changed files"
    if: github.event_name == "pull_request"
    run: |
      ./scripts-cli.py ...
  - name: "Merge to main: Run all test sessions"
    if: github.event_name != "pull_request"
    run: |
      ./scripts-cli.py ...
```

egparedes: Yes, that should be equivalent. I don't have a strong preference for any of the two options, just thought that it could be simpler to keep the logic inside one step to avoid repeating the part about exporting the generated file...

.github/workflows/test-components.yml
egparedes: You're right. I think we can now delete the whole line

.github/workflows/test-components.yml
romanc: I also don't have a strong opinion. Let's leave it as is. I was just curios if that was an option.

havogt (approved):

PR #2122, opened 2025-07-07
feat[next][dace]: Map symbolic args to nested SDFG

edopao:
**1st and main change**
This PR tries to solve the lowering of symbol assignment in the context of let statements like `let nlev = vertical_end - 1`.
The let statement is lowered as a nested SDFG. The `nlev` symbol can be represented on the nested SDFG as either a scalar (the previous behavior) or a dace symbol. The condition applied by this PR to use the dace symbolic representation is that all arguments to a scalar expression are already dace symbols. Thus, the symbolic expression `vertical_end - 1` can be mapped to the inner symbol `nlev` on the nested SDFG.

**2nd change**
Suppose a let statement like `let nlev = vertical_end + val - 1` where `vertical_end` is a dace symbol but `val` is a scalar. In this case, the lowering would still choose to lower the scalar expression as a tasklet, because not all arguments are dace symbols, and write the result to a transient data node. To overcome this limitation, this PR introduces another change: represent all program inputs as dace symbols. Observe that this change only applies to the scalar program arguments, that is the inputs to the top-level SDFG. For scalars on nested SDFGs the same rules as before still apply.

**3rd change**
The above change enables some cleanup in the check of symbolic arguments. Since all program scalar arguments are now represented as dace symbols, we do not need to scan _on the top-level SDFG_ which symbols are used inside domain expressions. Therefore, this PR moves the collection of domain symbols to the setup of nested SDFG context.

Note that this PR creates a new module `gtir_lower_types`. In a separate refactoring PR existing data types should be moved here.


src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -785,20 +797,38 @@ def visit_FunCall(
...
+                if (
+                    isinstance(lambda_arg.type, ts.ScalarType)
+                    and all(
+                        symref in sdfg.symbols
+                        for symref in symbol_ref_utils.collect_symbol_refs(
+                            lambda_arg, self.global_symbols.keys()
+                        )
+                    )
+                )
+            }
```
philip-paul-mueller: It might be simpler to digest if it is written as a normal loop.

philip-paul-mueller (commented):
Ioannis and me also observed that in the old lowering the symbol (that was created on an interstate edge) was also used outside the nested SDFG.
It was then turned into a free symbol and added to the calling signature leading to a launch error.
Does this PR also addresses this issue?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -520,7 +520,11 @@ def _add_storage(
             dc_dtype = gtx_dace_utils.as_dace_type(gt_type)
             # note that `symbolic_arguments` contains the name of the tuple a scalar belongs to
             sym_ref = name if tuple_name is None else tuple_name
-            if gtx_dace_utils.is_range_symbol(name) or sym_ref in symbolic_arguments:
+            if (
+                scalars_as_dace_symbols
+                or gtx_dace_utils.is_range_symbol(name)
+                or sym_ref in symbolic_arguments
```
philip-paul-mueller: `Iterable` is a too weak concept here, because an `Iterable` is a single pass thing.
What you need, based on what I have seen, is a [`Collection`](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes), or even better a plain `set`.
- At several point you cast them to a `set`.
- Checking if something is there is fastest in a `set`, or at least faster than iterating through the whole thing.

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
```diff
@@ -316,7 +316,9 @@ def _lower_lambda_to_nested_sdfg(
     # This property is used by pattern matching in SDFG transformation framework
     # to skip those transformations that do not yet support control flow blocks.
     nsdfg.using_explicit_control_flow = True
-    lambda_translator = sdfg_builder.setup_nested_context(lambda_node, nsdfg, sdfg, lambda_symbols)
+    lambda_translator = sdfg_builder.setup_nested_context(
+        nsdfg, sdfg, lambda_symbols, symbolic_arguments={}
```
philip-paul-mueller:
```suggestion
        nsdfg, sdfg, lambda_symbols, symbolic_arguments=set()
```
According to the annotation `symbolic_arguments` is `Iterator[str]` which `dict[str, Any]` is (because by default you iterate over its keys).
However, I guess you wanted to create an empty `set`, but this must be done by `set()`.
See also my other comment, regarding the `_add_storage()` function.

Beside that, I was also wondering why you initialize it as empty.
I would have expected that there is only one place where you set that to the empty container, at the very beginning.
I suspect that there is some outer logic that I do not see.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -375,24 +356,26 @@ def is_column_axis(self, dim: gtx_common.Dimension) -> bool:
...
         scope_symbols: dict[str, ts.DataType],
+        symbolic_arguments: Iterable[str],
     ) -> SDFGBuilder:
         nsdfg_builder = GTIRToSDFG(self.offset_provider_type, self.column_axis, scope_symbols)
         # We pass to the nested SDFG all symbols in scope, which includes the values
         # mapped to the parameters of the lambda expression (`node.params`) and
         # the symbols defined in the parent SDFG.
         params = [gtir.Sym(id=p_name, type=p_type) for p_name, p_type in scope_symbols.items()]
-        symbolic_arguments = _collect_symbols_in_domain_expressions(node, params) | {
+        parent_symbols = {
```
philip-paul-mueller, line 329: If I understand this here correctly you have moved out the functionality of `_collect_symbols_in_domain_expressions()` outside of this function and now only the result in the guise  `symbolic_arguments` is passed instead?
The functionality has been moved into `visit_Program()`?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -478,6 +463,8 @@ def _add_storage(
             symbolic_arguments: Set of GT4Py scalars that must be represented as SDFG symbols.
             name: Symbol Name to be allocated.
             gt_type: GT4Py symbol type.
+            scalars_as_dace_symbols: Set to `True` for the top-level SDFG, where
+                scalar parameters should be represented as dace symbols.
```
philip-paul-mueller: I would also add its relation to `symbolic_arguments`.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -497,15 +484,28 @@ def _add_storage(
                 sym_ref = name if tuple_name is None else tuple_name
                 tuple_fields.extend(
                     self._add_storage(
-                        sdfg, symbolic_arguments, sym.id, sym.type, transient, tuple_name=sym_ref
+                        sdfg,
```
philip-paul-mueller: I would either pass all arguments by position or by name if not forced by the language, i.e. positional only arguments.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -600,24 +604,29 @@ def _add_sdfg_params(
...
         Helper function to add storage for node parameters and connectivity tables.
 
-        GT4Py field arguments will be translated to `dace.data.Array` objects.
-        GT4Py scalar arguments will be translated to `dace.data.Scalar` objects,
-        except when they are listed in 'symbolic_arguments', in which case they
-        will be represented in the SDFG as DaCe symbols.
+        For details about storage allocation of each data type, see `_add_storage()`.
         """
+        if symbolic_arguments is None:
+            symbolic_arguments = {}
```
philip-paul-mueller:
```suggestion
            symbolic_arguments = set()
```

See also my note in the `visit_Program()` function where this function is called.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -669,9 +679,7 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
         else:
             head_state = entry_state
 
-        domain_symbols = _collect_symbols_in_domain_expressions(node, node.params)
-        sdfg_arg_names = self._add_sdfg_params(sdfg, node.params, symbolic_arguments=domain_symbols)
-
+        sdfg_arg_names = self._add_sdfg_params(sdfg, node.params, scalars_as_dace_symbols=True)
```
philip-paul-mueller: This here is the "genesis point" of `symbolic_arguments` and you rely use the default argument to initialize it to the empty set.
However, I think default arguments should be used if you want to modify a certain behaviour in a few cases.
However, as I see it, passing nothing to `_sdd_sdfg_params()` is more like an error.
I would recommend the following:
- Removing the `Optional` and the default value from `symbolic_arguments`.
- Explicitly create the empty set and add a note that this here is the place where we begin with it and pass it.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -785,20 +793,36 @@ def visit_FunCall(
...
+                    lambda_arg, self.global_symbols.keys()
                 )
-                for arg in node.args
-            ]
-
+                if all(symref in sdfg.symbols for symref in symrefs):
+                    symbolic_args[str(p.id)] = gtir_sdfg_utils.get_symbolic(lambda_arg)
+            # All other lambda arguments are lowered to taskgraphs that produce a data node.
+            args = {
+                str(p.id): (
```
philip-paul-mueller, line 773:
```suggestion
                param: (
```
Although I am not sure if this is possible, i.e. if `param` is still in scope.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -785,20 +793,36 @@ def visit_FunCall(
...
+                    continue
+                symrefs = symbol_ref_utils.collect_symbol_refs(
+                    lambda_arg, self.global_symbols.keys()
                 )
-                for arg in node.args
-            ]
-
+                if all(symref in sdfg.symbols for symref in symrefs):
+                    symbolic_args[str(p.id)] = gtir_sdfg_utils.get_symbolic(lambda_arg)
+            # All other lambda arguments are lowered to taskgraphs that produce a data node.
```
philip-paul-mueller: I kind of understand what you mean by "taskgraph" but is there a particular reason why you do not call it "dataflow (graph)"?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -810,24 +834,39 @@ def visit_Lambda(
...
 
-        All arguments to lambda functions are fields (i.e. `as_fieldop`, field or scalar `gtir.SymRef`,
-        nested let-lambdas thereof). The reason for creating a nested SDFG is to define local symbols
-        (the lambda paremeters) that map to parent fields, either program arguments or temporary fields.
+        The `data_args` are passed to a lambda function as data access nodes (i.e.
+        `as_fieldop`, field or scalar `gtir.SymRef`, nested let-lambdas thereof).
+        The reason for creating a nested SDFG is to define local symbols (the lambda
+        paremeters) that map to parent fields, either program arguments or temporaries.
+        Arguments that can be evaluated as symbolic expressions (`symbolic_args`)
+        are instead mapped to symbols on the nested SDFG.
```
philip-paul-mueller: The description here is very strongly tied to the names inside the function, I would use more neutral terms.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg_utils.py
```diff
@@ -127,3 +128,13 @@ def visit_SymRef(self, node: gtir.SymRef, *, symtable: Dict[str, str]) -> gtir.S
     # assert that the new symbol names are not used in the IR
     assert ir_sym_ids.isdisjoint(invalid_symbols_mapping.values())
     return ReplaceSymbols().visit(ir, symtable=invalid_symbols_mapping)
+
+
+def get_symbolic(node: gtir.Expr) -> dace.symbolic.SymbolicType:
```
philip-paul-mueller, line 135: Is every other usage of `pystr_to_symbolic()` removed?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -141,10 +153,10 @@ def is_column_axis(self, dim: gtx_common.Dimension) -> bool:
     @abc.abstractmethod
     def setup_nested_context(
         self,
-        node: gtir.Lambda,
         sdfg: dace.SDFG,
         parent: dace.SDFG,
         scope_symbols: dict[str, ts.DataType],
+        symbolic_arguments: Iterable[str],
```
philip-paul-mueller: See comment in `_add_storage()`.

philip-paul-mueller (commented):
There are some smaller things.

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
edopao: About the empty set, you are absolutely right. Fixed.

I am passing an empty set, here, because I am lazy. It is possible to apply the same (or similar) changes as those of let-statements to scan expressions. However, to minimize the changes I am skipping it for now. Besides, I think that the nested SDFG of a scan expression cannot contain domain expressions, and therefore does not _suffer_ from the same issue as the let statements that modify the range of the vertical domain. I will add an explanation.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: I agree, I'll change to `set` (which I actually used initially).

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: You will see it in next review. I still need the functionality of `_collect_symbols_in_domain_expressions()` but only for the setup of a nested SDFG context. I'll move the code here.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: I was not sure about what doing here. I will follow your suggestion.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: I tried that: it does not work.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: Good point!

src/gt4py/next/program_processors/runners/dace/gtir_sdfg_utils.py
edopao: No, and `pystr_to_symbolic()` is still used for more general parsing of strings. This utility function only accepts `gtir.Expr` as input, and it internally lowers the expression to a string.

On the other hand, I need to replace a few places where we parse `gtir.Expr`.

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
```diff
@@ -316,7 +317,22 @@ def _lower_lambda_to_nested_sdfg(
     # This property is used by pattern matching in SDFG transformation framework
     # to skip those transformations that do not yet support control flow blocks.
     nsdfg.using_explicit_control_flow = True
-    lambda_translator = sdfg_builder.setup_nested_context(lambda_node, nsdfg, sdfg, lambda_symbols)
+    # We pass an empty set as symbolic arguments, which implies that all scalar
+    # inputs of the scan nested SDFG will be represented as scalar data containers.
+    # The reason why we do not check for dace symbolic expressions and do not map
+    # them to inner symbols is that the scan expression should not contain any domain
+    # expression (no field operator inside)..
```
philip-paul-mueller:
```suggestion
    # expression (no field operator inside).
```

src/gt4py/next/program_processors/runners/dace/gtir_scan_translator.py
```diff
@@ -316,7 +317,22 @@ def _lower_lambda_to_nested_sdfg(
...
+    assert (
+        len(
+            eve.walk_values(lambda_node)
+            .filter(lambda node: cpm.is_call_to(node, ("cartesian_domain", "unstructured_domain")))
+            .to_list()
+        )
+        == 0
```
philip-paul-mueller: I think we should ask Enrique if he could add a "count()` predicate or so to `eve`?
Alternatively you could also rewrite the check to something like `assert not any(...)`.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -375,24 +357,42 @@ def is_column_axis(self, dim: gtx_common.Dimension) -> bool:
...
+        symbolic_arguments: set[str],
     ) -> SDFGBuilder:
         nsdfg_builder = GTIRToSDFG(self.offset_provider_type, self.column_axis, scope_symbols)
         # We pass to the nested SDFG all symbols in scope, which includes the values
         # mapped to the parameters of the lambda expression (`node.params`) and
         # the symbols defined in the parent SDFG.
         params = [gtir.Sym(id=p_name, type=p_type) for p_name, p_type in scope_symbols.items()]
-        symbolic_arguments = _collect_symbols_in_domain_expressions(node, params) | {
+
+        domain_symbols = set(
```
philip-paul-mueller: Since you want to turn it into a `set` anyway why do you not start with a `set()` you first create `list`s that you ten turn into `set`s.
I guess it could be a limitation of expression?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -669,8 +687,8 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
         else:
             head_state = entry_state
 
-        domain_symbols = _collect_symbols_in_domain_expressions(node, node.params)
-        sdfg_arg_names = self._add_sdfg_params(sdfg, node.params, symbolic_arguments=domain_symbols)
+        # by setting `symbolic_arguments=None`, we represent all scalars as dace symbols
```
philip-paul-mueller: You should also add why you do this.

philip-paul-mueller (approved):
LGTM.
There are some small things more like suggestions, but they do not warrant an additional round.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: Now that I understand Python better, I can simplify this expression. Thanks for pointing it.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -495,16 +449,13 @@ def _add_storage(
...
                 assert isinstance(sym.type, ts.DataType)
-                # for nested tuples, we always pass the root tuple symbol name
-                sym_ref = name if tuple_name is None else tuple_name
                 tuple_fields.extend(
                     self._add_storage(
                         sdfg=sdfg,
-                        symbolic_arguments=symbolic_arguments,
-                        name=sym.id,
+                        symbolic_params=symbolic_params,
+                        name=str(sym.id),
```
philip-paul-mueller, line 456: You added it so I guess it is needed, but I wonder, why?

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
```diff
@@ -803,18 +755,18 @@ def visit_FunCall(
...
             symbolic_args = {}
             for p, lambda_arg in zip(node.fun.params, node.args, strict=True):
                 if not isinstance(lambda_arg.type, ts.ScalarType):
                     continue
-                symrefs = symbol_ref_utils.collect_symbol_refs(
-                    lambda_arg, self.global_symbols.keys()
-                )
-                if all(symref in sdfg.symbols for symref in symrefs):
-                    symbolic_args[str(p.id)] = gtir_sdfg_utils.get_symbolic(lambda_arg)
+                # convert the scalar argument to a dace symbolic expression
```
philip-paul-mueller:
```suggestion
                # convert the scalar argument to a dace symbolic expression if all of its dependencies are symbols to.
```

I wonder what happens if you have the following case.
- You have symbol `a := symbolicParam1+ 3` and `b := a + 3`. 
- It is clear that you can handle `a` in that way because `param1` is a symbolic argument.
- However, for `b` you can do that too, however, after you have figured out that `a` is also a symbol.

This means that the order of processing is important, i.e. `a` must be processed before `b` is processed.
I am actually not sure if this can happen at all (and if I was able to express myself in an understandable way).

philip-paul-mueller (approved):
LGTM.
I have two small things.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: This cannot happen within the same context: `a` must be defined before `b`. The GTIR would express this using two nested let-lambdas: `let (a := symbolicParam1+ 3)( let (b := a + 3)( a + b ) )` and it would result in two levels of nested SDFGs.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: It is not strictly needed and I should probably remove it. It was annoying, while debugging, to see the dictionary keys be sometimes pure strings and other times `eve.type_definitions.ConstrainedStr`.

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
philip-paul-mueller: You are right!

src/gt4py/next/program_processors/runners/dace/gtir_sdfg.py
edopao: There are several places where I already do this casting.. I would keep it for now and eventually address all casts in a separate PR.

PR #2123, opened 2025-07-07
fix[next][dace]: Filter symbolic expressions as right hand-side assignments

edopao:
This PR fixes a error observed in icon4py runs when `async_sdfg_call` is set to `True`.

```
any(
        str(sym) in sdfg.arrays
        for v in oedge.data.assignments.values()
>       for sym in dace.symbolic.pystr_to_symbolic(v).free_symbols
    ):
E   AttributeError: 'bool' object has no attribute 'free_symbols'
```


edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

PR #2129, opened 2025-07-09
fix[dace][next]: Miscellaneous DaCe Optimizer Optimization

philip-paul-mueller:
This PR contains miscellaneous changes that improve the performance of the optimization pipeline.
The original PR was split into different parts and essentially this contains the remains that could not be moved to any of it.

### DIFF
--- a/src/gt4py/cartesian/utils/base.py
+++ b/src/gt4py/cartesian/utils/base.py
@@ -239,8 +239,7 @@ def make_dir(dir_name, *, mode=0o777, is_package=False, is_cache=False):
     if is_cache:
         with open(os.path.join(dir_name, "CACHEDIR.TAG"), "w") as f:
             f.write(
-                """Signature: 8
-a477f597d28d172789f06886806bc55
+                """Signature: 8a477f597d28d172789f06886806bc55
 # This file is a cache directory tag created by GT4Py.
 # For information about cache directory tags, see:
 #	http://www.brynosaurus.com/cachedir/


src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -137,17 +137,19 @@ def gt_auto_optimize(
                 repl=constant_symbols,
                 simplify=True,  # Simplify again after.
                 simplify_at_entry=False,
-                validate=validate,
+                validate=False,
```
havogt: to me it feels like for debugging you might want to have the option to turn it on?

philip-paul-mueller: It is kind of strange, but the `validate` argument passed to `gt_auto_optimizer()` means "only validate at the end of the function".
Thus `gt_auto_optimizer()` has to pass `validate=False` to every substep it calls, otherwise validation would be performed at intermediate steps.
However, one has to make sure that `validate_all` implies `validate`.

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -44,20 +44,24 @@ def gt_eliminate_dead_dataflow(
...
         single_use_data=single_use_data,
+        validate=False,
+        validate_all=validate_all,
     )
 
     if run_simplify:
         gtx_transformations.gt_simplify(
             sdfg=sdfg,
-            validate=validate,
+            validate=False,
```
edopao:
```suggestion
            validate=validate,
```
since you are using `elif` below?

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_horizontal_map_split_fusion.py
```diff
@@ -213,10 +213,11 @@ def test_horizontal_map_fusion():
     N = 20
     sdfg = _make_sdfg_with_multiple_maps_that_share_inputs(N)
 
-    _ = gtx_transformations.gt_horizontal_map_fusion(
-        sdfg=sdfg,
-        run_simplify=False,
-        consolidate_edges_only_if_not_extending=False,
+    _ = gtx_transformations.gt_horizontal_map_split_fusion(
+        sdfg,
+        False,
```
edopao: Probably better to use keyword syntax here.

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
philip-paul-mueller: It is essentially the same reason that Hannes pointed out above.
However, you are right, there is an error here.
Instead of an `elif` it must be an `if`, because when `validate` and `run_simplify` were both `True` it will currently not run validation.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_horizontal_map_split_fusion.py
philip-paul-mueller: You are right.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
```diff
@@ -72,15 +72,16 @@ def _new_name(old_name: str) -> str:
...
         elif isinstance(node, dace_nodes.NestedSDFG):
             node_ = graph.add_nested_sdfg(
-                copy.deepcopy(node.sdfg),
-                node.in_connectors,
-                node.out_connectors,
-                node.symbol_mapping,
+                sdfg=copy.deepcopy(node.sdfg),
+                inputs=set(node.in_connectors.keys()),
+                outputs=set(node.out_connectors.keys()),
+                symbol_mapping=node.symbol_mapping.copy(),
```
edopao, line 83: Interesting change. Passing the original `symbol_mapping` could have been a source of bugs.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -835,9 +836,8 @@ def _rewire_map_scope(
         # There is an invalid cache state in the SDFG, that makes the memlet
         #  propagation fail, to clear the cache we call the hash function.
         #  See: https://github.com/spcl/dace/issues/1703
-        _ = sdfg.hash_sdfg()
-        # TODO(phimuell): Use a less expensive method.
-        dace.sdfg.propagation.propagate_memlets_state(sdfg, state)
+        _ = sdfg.reset_cfg_list()
+        dace_propagation.propagate_memlets_map_scope(sdfg, state, outer_entry)
```
edopao, line 840: OK, here we reduce the scope of `propagate_memlets()`

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -226,13 +226,8 @@ def apply(
...
         # Readjust the Subsets.
-        # TODO(phimuell): Technically only needed if we patched some data from
-        #   beyond the Map inside the SDFG.
-        dace_propagation.propagate_memlets_nested_sdfg(
-            parent_sdfg=sdfg,
-            parent_state=graph,
-            nsdfg_node=if_block,
-        )
+        #  As a side effect this call will also properly propagate the nested SDFG `if_block`.
+        dace_propagation.propagate_memlets_map_scope(sdfg, graph, enclosing_map)
```
edopao, line 230: OK, here we reduce the scope of `propagate_memlets()`

src/gt4py/next/program_processors/runners/dace/transformations/local_double_buffering.py
```diff
@@ -138,7 +138,9 @@ def _add_local_double_buffering_to_single_data(
         storage=dace_dtypes.StorageType.Register,
         find_new_name=True,
     )
-    new_double_inner_buff_node = state.add_access(new_double_inner_buff_name)
+    new_double_inner_buff_node = state.add_access(
+        new_double_inner_buff_name, copy.copy(input_node.debuginfo)
```
edopao, line 142: Interesting change. As discussed on Slack, we can avoid creating the debug info from dace self-reflection, which is expensive. Besides, the original debug info is probably more useful.

edopao (approved):
I only added some notes for myself. LGTM

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
philip-paul-mueller: Actually we increase the scope of the Memlet propagation, before it was only updating the Memlets inside the nested SDFG and the Memlets that are immediately attached to the nested SDFG object.
Now we also propagate them in the surrounding scope.
It is more a bug fix.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: Yes, here we indeed reduce the scope of the Memlet propagation to the minimum.

PR #2130, opened 2025-07-09
feat[next][dace]: Enable cuda memory pool for temporary arrays

edopao:
Add argument to `auto_optimize()` call to enable memory pool in cuda codegen. Default `False`, but set to `True` in translation stage.


src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -30,6 +30,7 @@ def gt_auto_optimize(
     blocking_size: int = 10,
     blocking_only_if_independent_nodes: bool = True,
     reuse_transients: bool = False,
+    use_memory_pool: bool = False,
```
philip-paul-mueller: I would name it `use_gpu_memory_pool` or simply `gpu_memory_pool`.
The current name indicates that it is also usable by CPU.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -74,6 +74,7 @@ def make_dace_backend(
     gpu: bool,
     async_sdfg_call: bool,
     make_persistent: bool,
+    use_memory_pool: bool,
```
philip-paul-mueller, line 77: This is optional, but I strongly recommend you to rename this argument as well.

philip-paul-mueller (approved):
LGTM

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
edopao: My intention was to keep this argument generic. As you see in `translation.py`, I raise a `NotImplementedError` if this argument is set for CPU.

PR #2133, opened 2025-07-10
CI: explicitly set number of GPUs in slurm request

edopao:
This setting is now required on Santis vCluster, after the vCluster configuration has been updated with sharing capability.


edopao:
cscs-ci run default

edopao:
cscs-ci run default

ci/cscs-ci.yml
```diff
@@ -121,6 +121,7 @@ build_py310_image_aarch64:
   variables:
     CSCS_CUDA_MPS: 1
     SLURM_JOB_NUM_NODES: 1
+    SLURM_GPUS_PER_NODE: 4
```
edopao, line 124: Should we request 1 or 4 GPUs? We only need 1 GPU in current test configuration.

edopao: I propose to set 1 GPU, to match the (current) real requirement.

PR #2134, opened 2025-07-11
fix[next]: symbol clash in inline_lambda

havogt:
Before this PR `(λ(x, y) → f(x, x + y))(y + y, x)` with `opcount_preserving=True` would inline to `(λ(x) → f(x, x + x))(y + y)` because the outer `x` vs the lambda parameter `x` was not properly handled.
Now, in case of such a conflict, the lambda is first transformed to `λ(x_, y) → f(x_, x_ + y)`.


egparedes (approved):
LGTM

PR #2135, opened 2025-07-11
ci: try node sharing

havogt:


havogt:
I wouldn't enable it for now, because there are only 2 nodes in the "shared" partition.

PR #2137, opened 2025-07-11
feat[next][dace]: Lowering of concat_where to SDFG

edopao:
Lowering to SDFG of `concat_where` primitive, follows GTIR PR #1713.
Test coverage provided in previous PR.


src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -1090,13 +1090,16 @@ def _visit_list_get(self, node: gtir.FunCall) -> ValueExpr:
             )
         elif isinstance(index_arg, ValueExpr):
             tasklet_node = self._add_tasklet(
-                "list_get", inputs={"index", "list"}, outputs={"value"}, code="value = list[index]"
+                "list_get",
+                inputs={"__index", "list"},
```
philip-paul-mueller: I would also use `__list` here.
Also since `list` is a Python keyword I would say we should avoid it.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,421 @@
...
+    f: gtir_to_sdfg_types.FieldopData,
+    f_desc: dace.data.Array,
+    concat_dim: gtx_common.Dimension,
+    concat_dim_index: int,
+    concat_dim_origin: dace.symbolic.SymbolicType,
+) -> tuple[gtir_to_sdfg_types.FieldopData, dace.data.Array]:
+    """
+    Helper function called by `_translate_concat_where_impl` to create a slice along
+    the concat dimension, that is a new array with an extra dimension and a single
+    level. This allows to treat 'f' as a slice and concatanate it to the other field.
```
philip-paul-mueller: I would give `f` a better name.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,421 @@
...
+    The scalar value can come from either a scalar node or from a 1D-array (assuming
+    the array represents a field in the concat dimension).
+    """
+    assert isinstance(inp.gt_type, ts.FieldType)
+    assert len(inp.gt_type.dims) == 1
+    out_dims, out_origin, out_shape = gtir_domain.get_field_layout(domain)
+    out_type = ts.FieldType(dims=out_dims, dtype=inp.gt_type.dtype)
+
+    # The strict order of lower and upper bounds of output domain is not guaranteed,
+    #   for concat_where expressions, so we apply a runtime check.
```
philip-paul-mueller: I am actually not sure what you mean with this exactly.
But I guess it means that the lower bound is not necessarily smaller than the upper, which is why we use dynamic Memlets.
I think you should be a bit more detailed.

I have probably asked you that already but do you know why this is the case.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,421 @@
...
+    out_shape[concat_dim_index] = dace.symbolic.pystr_to_symbolic(
+        f"max(0, {out_shape[concat_dim_index]})"
```
philip-paul-mueller: I am not fully sure but shouldn't you use `origin` here.
Is it because you incorporated origin in the index computation below?
If so I would add a note to explain that.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,421 @@
...
+        sdfg: The SDFG where the primitive subgraph should be instantiated
+        state: The SDFG state where the result of the primitive function should be made available
+        sdfg_builder: The object responsible for visiting child nodes of the primitive node.
+        mask_domain: Domain (only for concat dimension) of the true branch, infinite
+            on lower or upper boundary.
+        node_domain: Domain (all dimensions) of output field.
+        tb_node_domain: Domain of the field passed on the true branch.
+        fb_node_domain: Domain of the field passed on the false branch.
+        tb_field: Input field on the true branch.
+        fb_field: Input field on the false branch.
```
philip-paul-mueller, line 176: I think that `tb` and `fb` are a bit too short.
I would propose to name them `true_domain` and `true_field`, respectively.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,421 @@
...
+    if mask_lower_bound == gtir_to_sdfg_utils.get_symbolic(gtir.InfinityLiteral.NEGATIVE):
+        concat_dim_bound = mask_upper_bound
+        lower, lower_desc, lower_domain = (tb_field, tb_data_desc, tb_domain)
+        upper, upper_desc, upper_domain = (fb_field, fb_data_desc, fb_domain)
+    elif mask_upper_bound == gtir_to_sdfg_utils.get_symbolic(gtir.InfinityLiteral.POSITIVE):
+        concat_dim_bound = mask_lower_bound
+        lower, lower_desc, lower_domain = (fb_field, fb_data_desc, fb_domain)
+        upper, upper_desc, upper_domain = (tb_field, tb_data_desc, tb_domain)
+    else:
+        raise ValueError(f"Unexpected concat mask {mask_domain[0]}.")
```
philip-paul-mueller, line 205: I think you should elaborate here a but more, why the assignment true/false to lower/upper is not fix.
After some thinking I think I have guessed it but I am not fully sure.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,421 @@
...
+        tb_field: Input field on the true branch.
+        fb_field: Input field on the false branch.
+
+    Returns:
+        The field resulted from concatanating the input fields on the lower and upper domain.
+    """
+    tb_data_desc, fb_data_desc = (inp.dc_node.desc(sdfg) for inp in [tb_field, fb_field])
+    assert tb_data_desc.dtype == fb_data_desc.dtype
+
+    tb_domain, fb_domain = (
```
philip-paul-mueller, line 184: I would actually write it in two lines, it might be cleaner.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,421 @@
...
+        gtir_domain.extract_domain(domain) for domain in [tb_node_domain, fb_node_domain]
+    )
+    concat_dim, mask_lower_bound, mask_upper_bound = mask_domain[0]
```
philip-paul-mueller, line 188:
```suggestion
    assert len(mask_domain) == 1
    concat_dim, mask_lower_bound, mask_upper_bound = mask_domain[0]
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,421 @@
...
+    # we use the concat domain, stored in the annex, as the domain of output field
+    output_domain = gtir_domain.extract_domain(node_domain)
+    output_dims, output_origin, output_shape = gtir_domain.get_field_layout(output_domain)
+    concat_dim_index = output_dims.index(concat_dim)
+
+    # The strict order of lower and upper bounds of output domain is not guaranteed,
+    #   for concat_where expressions, so we apply a runtime check.
+    output_shape[concat_dim_index] = dace.symbolic.pystr_to_symbolic(
+        f"max(0, {output_shape[concat_dim_index]})"
+    )
```
philip-paul-mueller: I think you should make a helper function out of it.
In that way fixes are applied more uniformly.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,425 @@
...
+        else (
+            f"({map_variables[concat_dim_index]} + {out_origin[concat_dim_index] - inp.origin[0]})"
+        )
+    )
+    state.add_mapped_tasklet(
+        "broadcast",
+        map_ranges={
+            index: r
+            for index, r in zip(map_variables, dace_subsets.Range.from_array(out_desc), strict=True)
+        },
```
philip-paul-mueller:
```suggestion
        map_ranges=dict(zip(map_variables, dace_subsets.Range.from_array(out_desc), strict=True)),
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,425 @@
...
+
+    # The strict order of lower and upper bounds of output domain is not guaranteed,
+    #   for concat_where expressions, so we apply a runtime check.
+    output_shape[concat_dim_index] = dace.symbolic.pystr_to_symbolic(
+        f"max(0, {output_shape[concat_dim_index]})"
+    )
+
+    # in case one of the arguments is a scalar value, we convert it to a single-element
+    # 1D field with the dimension of the concat expression
+    if isinstance(lower.gt_type, ts.ScalarType):
```
philip-paul-mueller, line 223: What if both are scalars?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,425 @@
...
+        assert isinstance(lower.gt_type, ts.FieldType)
+        upper = gtir_to_sdfg_types.FieldopData(
+            upper.dc_node,
+            ts.FieldType(dims=[concat_dim], dtype=upper.gt_type),
+            origin=(concat_dim_bound,),
+        )
+        upper_bound = output_domain[concat_dim_index][2]
+        upper_domain = [(concat_dim, concat_dim_bound, upper_bound)]
+
+    if concat_dim not in lower.gt_type.dims:  # type: ignore[union-attr]
```
philip-paul-mueller, line 244: What does these cases mean.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,425 @@
...
+        assert len(upper_domain) == 1 and upper_domain[0][0] == concat_dim
+        upper_domain = [
+            *output_domain[:concat_dim_index],
+            upper_domain[0],
+            *output_domain[concat_dim_index + 1 :],
+        ]
+        upper, upper_desc = _make_concat_scalar_broadcast(
+            sdfg, state, upper, upper_desc, upper_domain, concat_dim_index
+        )
+    elif lower.gt_type.dims != upper.gt_type.dims:  # type: ignore[union-attr]
```
philip-paul-mueller: What is with the `else` branch, it is not needed?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,425 @@
...
+    lower_range_0 = output_domain[concat_dim_index][1]
+    lower_range_1 = dace.symbolic.pystr_to_symbolic(
+        f"max({lower_range_0}, {lower_domain[concat_dim_index][2]})"
+    )
+    lower_range_size = lower_range_1 - lower_range_0
+
+    upper_range_1 = output_domain[concat_dim_index][2]
+    upper_range_0 = dace.symbolic.pystr_to_symbolic(
+        f"min({upper_range_1}, {upper_domain[concat_dim_index][1]})"
+    )
```
philip-paul-mueller, line 353:
```suggestion
    upper_range_0 = dace.symbolic.pystr_to_symbolic(
        f"min({upper_range_1}, {upper_domain[concat_dim_index][1]})"
    )
    upper_range_1 = output_domain[concat_dim_index][2]
```
Now they have the same order as above.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,425 @@
...
+    output_node = state.add_access(output)
+
+    lower_subset = dace_subsets.Range(
+        [
+            (
+                lower_range_0 - lower.origin[dim_index],
+                lower_range_1 - lower.origin[dim_index] - 1,
+                1,
+            )
+            if dim_index == concat_dim_index
```
philip-paul-mueller: Either I would create a big `if-else` like:
```python
if dim_index == concat_dim_index:
    lower_subset = ...
    upper_subset = ...
    lower_output_subset = ...
    upper_output_subset = ...
else:
    lower_subset = ...
    upper_subset = ...
    lower_output_subset = ...
    upper_output_subset = ...
```
Or define a boolean where you set the result of the comparison.
Because now this decision is actually made at a lot of different places.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,425 @@
...
+    sdfg_builder: gtir_to_sdfg.SDFGBuilder,
+) -> gtir_to_sdfg_types.FieldopResult:
+    """
+    Lowers a `concat_where` expression to a dataflow where two memlets write
+    disjoint subsets, for the lower and upper domain, on one data access node.
+
+    Implements the `PrimitiveTranslator` protocol.
+
+    In case of tuples, this function calls `_translate_concat_where_impl()` on all
+    fields by means of `tree_map`.
```
philip-paul-mueller: I think these two lines are too implementation related for a "public" function.
However, you could add it as a normal comment of to `_translate_concat_where_impl()` which is an internal one.

philip-paul-mueller (commented):
I have some comments, but I do not see a major problem.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
edopao: Correct, I had missed that.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: This is the array shape, so the minimum size is always zero. I use `origin` in the computation of the memlet subset.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: I think that the lowering of nested `concat_where` to GTIR in the frontend is not able to detect empty copy branches. For field operators like:
```
    @gtx.field_operator
    def testee(interior: cases.IField, boundary: cases.IField, N: gtx.int32) -> cases.IField:
        interior = concat_where(IDim == 0, boundary, interior)
        interior = concat_where((1 <= IDim) & (IDim < N - 1), interior * 2, interior)
        interior = concat_where(IDim == N - 1, boundary, interior)
        return interior
```
The lowering to GTIR creates all possible branches because it has no domain knowledge. Afterwords, after the code is lowered to GTIR, we run domain inference and add domain information to all branches.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: There is an assert below on line 190, it should not happen.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: I had the same idea but I was too lazy. Done now.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: Good point.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: I use `upper_range_1` in the computation of `upper_range_0`, that is why they appear in this order.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: The `else` branch means all good, no special handling needed.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: Good idea!

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: Description added.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: I added a note in the function description:
```
    Note that 'tb' and 'fb' stand for true/false branch and refer to the two branches
    of the concat_where expression, passed as second and third argument, respectively.
```

edopao (commented):
Thanks for the review comments, I agree it looks better now.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
philip-paul-mueller: I see it now.
I guess there was some tricky thinking involved to get that right.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,463 @@
...
+    assert concat_dim not in field.gt_type.dims
+    dims = [
+        *field.gt_type.dims[:concat_dim_index],
+        concat_dim,
+        *field.gt_type.dims[concat_dim_index:],
+    ]
+    origin = tuple(
+        [*field.origin[:concat_dim_index], concat_dim_origin, *field.origin[concat_dim_index:]]
+    )
+    shape = tuple([*field_desc.shape[:concat_dim_index], 1, *field_desc.shape[concat_dim_index:]])
```
philip-paul-mueller, line 56: I do not understand that the `concat_dim_index` dimension alsways has length 1?
Or rather what happens if a slice is higher than one level?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -0,0 +1,463 @@
...
+        lower_domain = [(concat_dim, lower_bound, concat_dim_bound)]
+    elif isinstance(upper.gt_type, ts.ScalarType):
+        assert len(upper_domain) == 0
+        assert isinstance(lower.gt_type, ts.FieldType)
+        upper = gtir_to_sdfg_types.FieldopData(
+            upper.dc_node,
+            ts.FieldType(dims=[concat_dim], dtype=upper.gt_type),
+            origin=(concat_dim_bound,),
+        )
+        upper_bound = output_domain[concat_dim_index][2]
```
philip-paul-mueller, line 241: In case you refactor that thing, I would propose that you replace the tuple with a dataclass, because then we could remove this magic `[2]` and `[1]`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
philip-paul-mueller: Sorry for being persistent and a bit low here, but I think you should also add some typical `concat_where()` expressions to make it simpler to understand such as `concat_where(k < 0, ij_field, 1.0)` or something.
Because, I think that it is on a conceptional level simple, but its actual implementation is rather involved.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: All unit tests still pass even if I switch the order. Honestly, I do not remember why I chose this order in the end. Probably, I wanted to follow the principle that the far left-hand side and right-hand side correspond exactly to the output domain, without any symbolic manipulation. Instead, I only manipulate the neighborhood of the concat boundary:

In other words, I could have written:
```
    lower_range_0, upper_range_1 = output_domain[concat_dim_index][1:3]
    lower_range_1 = dace.symbolic.pystr_to_symbolic(
        f"max({lower_range_0}, {lower_domain[concat_dim_index][2]})"
    )
    upper_range_0 = dace.symbolic.pystr_to_symbolic(
        f"min({upper_range_1}, {upper_domain[concat_dim_index][1]})"
    )
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: Good idea.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: I am preparing a refactoring change to introduce the dataclass, but I would prefer to deliver it as a separate PR. Is that OK?

PR #2138, opened 2025-07-11
refactor[next][dace]: Split and rename modules of SDFG lowering

edopao:
This is a pure refactoring PR, it only splits and renames some module used for SDFG lowering. Below some guiding principles:
- Data types used during lowering (e.g. `FieldopData`, `FieldopResult`, `SymbolicData`) are moved to `gtir_to_sdfg_types`, as well as some utility functions to access/modify data types.
- Domain-related types (`FieldopDomain`) and functions are moved to `gtir_domain`
- The two above changes allow to reduce the size of `gtir_to_sdfg_primitives` (previously called `gtir_builtin_translator`) and also to remove several dependencies from `gtir_scan_translator` to `gtir_builtin_translator` for type definitions.
- The entry point of lowering is renamed from `gtir_sdfg` to `gtir_to_sdfg`
- Separate module for large lowering functions of specific primitives like `gtir_to_sdfg_scan` and in future `gtir_to_sdfg_concat_where`


philip-paul-mueller (approved):
LGTM.
However, I think that `gtir_lower` is a bit to short and I would call it `sdfg_lowering`, `gtir_lowering` or something.

PR #2140, opened 2025-07-11
build: fix dace groups

egparedes:
Delete unnecessary  `dev-cartesian`/`dev-next` dependency groups.


edopao (approved):
I tried the following commands and it works.
```
uv sync --extra next --group dace-next
uv sync --extra cartesian --group dace-cartesian
```

PR #2142, opened 2025-07-14
refactor[next][dace]: Introduce dataclass for field operator domain range

edopao:
This PR introduces a dataclass type to represent the domain range in the lowering to SDFG.
It follows a suggestion from @philip-paul-mueller in the review of #2137.


philip-paul-mueller (approved):
LGTM

PR #2146, opened 2025-07-16
fix[dace][next]: Removed Special Case For `CopyChainRemover`.

philip-paul-mueller:
`CopyChainRemover` removes patterns that are generated by `concat_where` expressions.
It checks if the source is not used anywhere else and if it is _fully_ copied into the destination.
Before there was an exception to that rule for cases when it was not possible to verify.
However, in `8_to_13_corrector` we encountered a case where this lead to an error.
In fact the error was probably present in more stencils, but it only surfaced there, the issue was that something was generated that was not used and then some memory location was written twice.
The correctness then dependent on the order in which the code was generated.

This PR removes this special rule. It will not impact SL2, but might impact SL1.


edopao (approved):
LGTM

At some point the lowering of `concat_where` was enforcing the relation `vertical_start < vertical_end`, but then I removed this because in production we are going to replace these variables with constant values. The additional complexity of the lowering is then not needed.

PR #2148, opened 2025-07-16
feat[dace][next]: Switch to DaCe MapFusion

philip-paul-mueller:
Before GT4Py shipped with its own Map fusion transformation, which were a copy of the one in a [DaCe PR](https://github.com/spcl/dace/pull/1987).
However, the PR has now caught up and the implementation provided inside it, actually surpasses the transformations that ships with GT4Py.

Thus it was decided to [update the DaCe fork GT4Py is using](https://github.com/GridTools/gt4py/pull/2149) to include the new version of the transformation and actually using them.

However, GT4Py still creates wrapper around the DaCe versions because, because some default values needs to be changed and also a callback mechanism is needed, to restrict the fusion.
However, in the long term, probably a better solution should be found.

In addition the Map fusion transformations are no longer named serial/parallel but vertical/horizontal, which is also in accordance with the splitting transformations and DaCe itself.


src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -819,8 +831,11 @@ def can_be_applied(
         org_trivial_map_range = copy.deepcopy(trivial_map.range)
         try:
             self._promote_map(graph, replace_trivail_map_parameter=False)
-            if not gtx_transformations.MapFusionSerial.can_be_applied_to(
+            if not gtx_transformations.MapFusionVertical.can_be_applied_to(
                 sdfg=sdfg,
+                options={
+                    "assume_always_shared": True,  # Avoids a scan.
```
edopao: Maybe write here a longer explanation like what you wrote in Slack to me:

> The fusion transformation, when applied, will scan the SDFG to decide whether it is possible to remove an intermediate transient. However, when only checking `can_be_applied()`, we can avoid the scan and assume it is shared.

src/gt4py/next/program_processors/runners/dace/transformations/serial_map_promoter.py
edopao, line 1: Maybe vertical_map_promoter? but I am also OK with keeping the current name

src/gt4py/next/program_processors/runners/dace/transformations/serial_map_promoter.py
```diff
@@ -419,12 +419,13 @@ def _test_if_promoted_maps_can_be_fused(
...
             self._promote_first_map(first_map_exit, second_map_entry)
 
-            if not gtx_transformations.MapFusionSerial.can_be_applied_to(
+            if not gtx_transformations.MapFusionVertical.can_be_applied_to(
                 sdfg=sdfg,
                 expr_index=0,
                 options={
                     "only_inner_maps": self.only_inner_maps,
                     "only_toplevel_maps": self.only_toplevel_maps,
+                    "assume_always_shared": True,  # Avoid a scan.
```
edopao: Same comment.

src/gt4py/next/program_processors/runners/dace/transformations/serial_map_promoter.py
philip-paul-mueller: I realized that even `SerialMapPromoter` is wrong, it just be simply `MapPromoter`.
I will open a new PR, since it is different, although it was started by this PR.

edopao:
LGTM, already approved.

PR #2150, opened 2025-07-16
fix[next][dace]: Ensure non-negative shape of temporaries in concat_where expressions

edopao:
This PR fixes a bug discovered in icon4py stencil tests. The implementation in main was performing the check on non-negative array shape only for the result of concat_where expressions. The problem is that the strange domain with range `start > stop` can be propagated to the arguments of concat_where, and therefore it can appear in regular field operators. This is why the new unittest computes `b * 2` as argument to concat_where.

The fix consists of applying the check in the general utility function, also used by regular field operators, for extraction of field shape from domain range.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_domain.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_domain.py
@@ -107,11 +107,15 @@ def get_field_layout(
     domain: FieldopDomain,
 ) -> tuple[list[gtx_common.Dimension], list[dace.symbolic.SymExpr], list[dace.symbolic.SymExpr]]:
     """
-    Parse the field operator domain and generates the shape of the result field.
+    Parse the field operator domain and generate the shape of the result field.
 
     Note that this function also ensures that the array shape computed from the
     domain range is non-negative. A negative shape can occur in concat_where
     expressions, where it can happen that 'stop' value is smaller than 'start'.
+    Also note that this _strange_ domain with 'start' > 'stop' is usually propagated
+    to the input arguments of a concat_where expression (the child nodes), thus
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -117,28 +117,6 @@ def _make_concat_scalar_broadcast(
...
-    bound is not necessarily smaller than the upper. When the upper bound is smaller,
-    it indicates an empty range to copy. This can only be resolved at runtime, which
-    is why we use dynamic memlets.
-    """
-    out_dims, out_origin, out_shape = gtir_domain.get_field_layout(domain)
-    concat_dim_index = (
-        out_dims.index(concat_dim) if isinstance(concat_dim, gtx_common.Dimension) else concat_dim
-    )
-    out_shape[concat_dim_index] = dace.symbolic.pystr_to_symbolic(
-        f"max(0, {out_shape[concat_dim_index]})"
```
philip-paul-mueller, line 137: The changes looks good to me, but can you explain why this in expression here, that ensures a non negative size, is not enough?

philip-paul-mueller (commented):
LGTM.
But I think we have to run some tests.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: The explanation is in the PR description (or at least I tried). I will write a better comment also in the code to not forget it.

philip-paul-mueller (approved):
LGTM.

Sorry forget to switch to `Approve`.

PR #2153, opened 2025-07-17
ci: add MacOS to daily ci

havogt:
Summary of changes:
- Enable daily CI on `macos` image. However, the default stable version (`macos-14`) has a compiler issue which is fixed in the next version` macos-15`. The `macos-15` version will become the default stable version very soon (in August'25).
- Minimum version of `nanobind` bumped to `1.9.0` in order to include fix for a linker issue on macos.
- Skipping check in a dace unittest in gt4py-next because it fails randomly. The failure becomes consistent after fixing the reduce initialization in the SDFG. This issue will be addressed in a separate PR.
- Skipping `test_dace_parsing` in gt4py-cartesian because it fails in CI.


edopao (dismissed):

.github/workflows/daily-ci.yml
```diff
@@ -34,8 +34,11 @@ jobs:
         # [arg: --resolution, env: UV_RESOLUTION=]
         dependencies-strategy: ["lowest-direct", "highest"]
         gt4py-module: ["cartesian", "eve", "next", "storage"]
-        os: ["ubuntu-latest"] #, "macos-latest"]
+        os: ["ubuntu-latest", "macos-latest"]
```
edopao: The way I run gt4py on my mac is using `gcc` from homebrew installation (`CC=$(brew --prefix gcc)` and similar for `CXX`). In that way, `libomp` installed from brew also works out of the box.

edopao: Alternatively, I also use Apple-`clang`, in which case we need to set `CC`/`CXX` differently.

.github/workflows/daily-ci.yml
```diff
@@ -49,6 +54,19 @@ jobs:
           cache-dependency-glob: "uv.lock"
           python-version: ${{ matrix.python-version }}
 
+      - name: Install latest-stable xcode for macos
+        if: ${{ matrix.os == 'macos-15' }}
+        uses: maxim-lobanov/setup-xcode@v1
```
edopao, line 56: I hope it is OK to use third-party actions.

pyproject.toml
```diff
@@ -99,8 +99,8 @@ dependencies = [
   'jinja2>=3.0.0',
   'lark>=1.1.2',
   'mako>=1.3',
-  'nanobind>=1.4.0 ',
-  'ninja>=1.10',
+  'nanobind>=1.9.0 ',
```
edopao: I see an extra space at the end, to be removed.

edopao (commented):
@havogt @egparedes Now all test jobs are passing in daily CI: https://github.com/GridTools/gt4py/actions/runs/16369575514

I have updated the PR description with a list of changes.

.github/workflows/daily-ci.yml
```diff
@@ -49,6 +54,19 @@ jobs:
...
+        with:
+          xcode-version: latest-stable
+
+      - name: Install OpenMP for macos
+        if: ${{ matrix.os == 'macos-15' }}
+        run: |
+          gcc --version
+          brew install libomp
```
egparedes, line 69: Why is not needed to also `brew install gcc` ?

.github/workflows/daily-ci.yml
```diff
@@ -34,8 +34,13 @@ jobs:
...
         dependencies-strategy: ["lowest-direct", "highest"]
         gt4py-module: ["cartesian", "eve", "next", "storage"]
-        os: ["ubuntu-latest"] #, "macos-latest"]
+        # TODO: switch to macos-latest once macos-15 is default, see https://github.com/actions/runner-images/issues/12520
+        #   On macos-14 we see the error: call to 'isfinite' is ambiguous.
+        os: ["ubuntu-latest", "macos-15"]
         python-version: ${{ fromJson(needs.get-python-versions.outputs.python-versions) }}
+        exclude:
+          - os: macos-15
+            gt4py-module: cartesian
```
egparedes: In theory this PR is trying to also fix this case, right?

pyproject.toml
egparedes: 🙏

.github/workflows/daily-ci.yml
edopao: I can try to re-enable it, I don't know why it was disabled (before I started to work on this PR)

.github/workflows/daily-ci.yml
edopao: Because I think we want to use the default Apple gcc compiler, which is based on clang.

.github/workflows/daily-ci.yml
edopao: Here is the explanation:
`clang: error: unsupported option '-fopenmp'`

We would have to install and use gcc from brew.

havogt: Pointing cartesianto gcc should then fix it I guess?

.github/workflows/daily-ci.yml
egparedes: Ah, now I understand. I thought `libomp` was adding OpenMP support to GCC, but it is the other way around, it is adding OpenMP support to clang, right?

.github/workflows/daily-ci.yml
```diff
@@ -49,12 +51,32 @@ jobs:
...
+          echo "CC=gcc-14" >> $GITHUB_ENV
+          echo "CXX=g++-14" >> $GITHUB_ENV
+
+      - name: Install latest-stable xcode for macos
+        if: ${{ (matrix.os == 'macos-15') && (matrix.gt4py-module != 'cartesian') }}
+        uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: latest-stable
+
+      - name: Install libomp for macos
```
egparedes: Maybe more explicit? 
```suggestion
      - name: Install libomp for macos clang
```

.github/workflows/daily-ci.yml
```diff
@@ -49,12 +51,32 @@ jobs:
           cache-dependency-glob: "uv.lock"
           python-version: ${{ matrix.python-version }}
 
+      - name: Install gcc for macos
```
egparedes: Maybe?
```suggestion
      - name: Install gcc (with OpenMP) for macos
```

.github/workflows/test-components.yml
```diff
@@ -20,65 +20,65 @@ jobs:
...
-  # Job to define test sessions based on changed files
-  define-test-sessions:
-    runs-on: ubuntu-latest
-    outputs:
-      test-sessions: ${{ steps.test-sessions.outputs.test-sessions }}
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0 # Use a deep clone to get target PR branch
+  # # Job to define test sessions based on changed files
```
egparedes: It might be easier to just temporarily rename the file to `test_foo.yaml.disabled` during the debugging tasks in the PR

.github/workflows/daily-ci.yml
edopao: Correct.

havogt:
What's the reason for using a different compiler in cartesian and next?

edopao:
> > What's the reason for using a different compiler in cartesian and next?
> 
> I wanted to use Apple gcc-clang, as a general choice. However, one of the cartesian backends was complaining, during cmake, because gcc-clang does not recognize `-fopenmp`: `clang: error: unsupported option '-fopenmp'`
> 
> Therefore, I made a exception for cartesian and used gcc.

Here is the job failure: https://github.com/GridTools/gt4py/actions/runs/16372332763/job/46263566278

havogt:
Sorry, forgot about this. Do you know what the correct flag would be? We could set compiler and linker flags here https://github.com/GridTools/gt4py/blob/98f728065e20825ece5f3a8f295626d5d9c06093/src/gt4py/cartesian/config.py#L46

havogt:
But also ok, to merge like it is.

edopao:
> Sorry, forgot about this. Do you know what the correct flag would be? We could set compiler and linker flags here
> 
> https://github.com/GridTools/gt4py/blob/98f728065e20825ece5f3a8f295626d5d9c06093/src/gt4py/cartesian/config.py#L46

Now all tests use MacOS clang, for both cartesian and next. All tests pass on the CI job, see https://github.com/GridTools/gt4py/actions/runs/16634316303

I just have to try one more thing, to revert a change to timers in cartesian test.

havogt:
From my side it's approved. Just add your approval once you are happy.

edopao (approved):

PR #2154, opened 2025-07-17
fix[dace][next]: Get correct symbol mapping in MoveDataflowIntoIfBody

edopao:
The `MoveDataflowIntoIfBody` transformation moves part of a dataflow into the nested SDFG of an If-Block. Some symbols need to be mapped to symbols in the parent SDFG. The function `dace.codegen.tools.type_inference.infer_expr_type()` was returning `int` type for unknown symbols, while we expected `None`. The fix is to first check if the symbol exists in parent SDFG, in which case we can directly map it; otherwise, we rely on `infer_expr_type()` to infer the type.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
@@ -460,7 +460,6 @@ class MoveDataflowIntoIfBody(dace_transformation.SingleStateTransformation):
             if new_sym in if_block_symbols:
                 # The symbol is already known, so we check that it is the same type as in the
                 #  parent SDFG.
-                # TODO(phimuell, edopao): Is this check legal.
                 assert if_block_symbols[new_sym] == parent_symbols[new_sym]
 
             elif new_sym in parent_symbols:


src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -456,11 +457,13 @@ def _update_symbol_mapping(
         symbols = if_block.sdfg.symbols
         for sym, symval in if_block.symbol_mapping.items():
```
philip-paul-mueller: I am not sure but I would say it is enough to iterate over `missing_symbols`?
What do you think?

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -456,11 +457,13 @@ def _update_symbol_mapping(
...
             if sym not in symbols:
-                if_block.sdfg.add_symbol(
-                    sym,
-                    dace.codegen.tools.type_inference.infer_expr_type(symval, if_block.sdfg.symbols)
-                    or dace_dtypes.typeclass(int),
-                )
+                if sym == str(symval):
+                    sym_type = parent.symbols.get(sym, dace_dtypes.typeclass(int))
```
philip-paul-mueller: I would not specify a default here, this should either be an error or we should do the same as before.

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -456,11 +457,13 @@ def _update_symbol_mapping(
...
+                else:
+                    sym_type = dace.codegen.tools.type_inference.infer_expr_type(
+                        symval, if_block.sdfg.symbols
```
philip-paul-mueller:
```suggestion
                    sym_type = dace.codegen.tools.type_inference.infer_expr_type(
                        symval, parent.symbols
```

I am actually not sure, but I would say that we should use the parent SDFG here, because the parent sdfg defines them.
It might even be an error in DaCe, I will ask in the channel.

philip-paul-mueller: I just checked with the DaCe source and there they use the parent, so my suggestion was correct and I adapted the DaCe code in a wrong way.

edopao: It depends whether you can have inter-dependent symbol definitions. I am thinking at this case:
```
symbol_mapping = {
    a: 1,
    b: a + 1,
}
```

In this case, you can infer thee type of `b` from `a`.

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
edopao: The default is actually to reproduce the same behavior as before. The map range variables are not defined on the parent SDFG.

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
edopao: Yes, for this specific case it should be enough. We are constructing the nested SDFG and we now that there are not strange symbol mappings like `b = a + 1`

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
philip-paul-mueller: I think this is not possible, because this would imply a certain order in which you can do something.
This the left hand side (key) of the mapping always refers to the symbols inside the nested SDFG and the right hand side (value) always refers to the symbols in the parent scope.
I made the same error when I reviewed the "scope PR".

PR #2160, opened 2025-07-19
ci: update development and GitHub Actions CI infrastructure 

egparedes:
There are two main changes in this PR related to the development infrastructure.

First, the dev scripts use now a more standard python package structure with a `__main__` module, and their dependencies have been moved back to `pyproject.toml` as a `scripts` dependency group.

Second, change the `test-component` GitHub Actions workflow to be more explicit about the sessions which are tested, because all test sessions are listed in the workflow file, and only the (optional) exclusions are loaded from a dynamically generated JSON file.

Finally, this PR also contains minor documentation cleanups.

### DIFF
--- a/scripts/github_ci.py
+++ b/scripts/github_ci.py
@@ -13,10 +13,11 @@ from __future__ import annotations
 
 import json
 from collections.abc import Sequence
-from typing import Annotated, cast
+from typing import Annotated, Literal, cast
 
 import rich
 import typer
+from rich import pretty
 
 from . import _common as common, nox_sessions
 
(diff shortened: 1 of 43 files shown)


scripts/github_ci.py
```diff
@@ -13,18 +13,78 @@
...
+        rich.print(
+            f"Found {len(affected)} test sessions in changes from {base_commit!r}: {affected}"
+        )
+
+    return affected, all_session_names
+
+
+def compute_output(
+    sessions: Sequence[str],
+    kind: Literal["affected", "excluded"],
```
havogt, line 67: Do you want to keep both? Since there are no tests, there is the danger of the unused one to break. Or maybe we want to extract this into a package that could be used in other projects, e.g. icon4py.

havogt (approved):
lgtm. Just one question...

scripts/github_ci.py
egparedes: Yes, I would like to keep them for now. For now it should be easy to keep both working, I'd like to eventually add tests to the scripts and we are still kind of exploring different options, so I want to keep all options open a bit longer.

PR #2161, opened 2025-07-21
build[dace][next]: Updated the DaCe dependency. 

philip-paul-mueller:
This brings DaCe to [`__gt4py-next-integration_2025_07_21_I`](https://github.com/GridTools/dace/tree/__gt4py-next-integration_2025_07_21_I).
The main feature that is introduced is, that it is now possible to specify `assume_always_single_use_data`, which might speed up some calls, as it is now possible to avoid some scans.


philip-paul-mueller:
This PR has been replaced [PR#2167](https://github.com/GridTools/gt4py/pull/2167)

PR #2163, opened 2025-07-22
feat[next][dace]: Config the gpu backend with async sdfg call

edopao:
Make asynchronous SDFG call the default config for gpu backend, in order to align the behavior of the dace and gtfn backends. Besides, this is also the icon4py config.


philip-paul-mueller (approved):
Technically there is no problem with it, however, default arguments have the tendency that they are hidden and not well remembered (because you do not have to type them).
So I think that for something that is so important as persistent and asynchronousity, there should be no default in my opinion.
But I will not object the change.

PR #2166, opened 2025-07-22
perf[next][dace] Remove scalar views

iomaganaris:
Remove scalar views that exist some times in the SDFG after `Reduce` library nodes. The views are generated indeterministically and seemed to cause regressions in performance

TODO

- [x] Run all stencils with the changes below


src/gt4py/next/program_processors/runners/dace/transformations/remove_views.py
```diff
@@ -0,0 +1,102 @@
...
+            return False
+        if not gtx_transformations.utils.is_view(src_desc, sdfg):
+            return False
+        if graph.in_degree(src_node) > 1:
+            return False
+        if not dst_desc.transient:
+            return False
+        first_node_out_edges = graph.out_edges(src_node)
+        if len(first_node_out_edges) != 1 and first_node_out_edges[0].dst != dst_node:
+            return False
```
philip-paul-mueller:
```suggestion
        if graph.out_degree(src_node) != 1:
            return False
```

src/gt4py/next/program_processors/runners/dace/transformations/remove_views.py
```diff
@@ -0,0 +1,102 @@
...
+        dst_node: dace_nodes.AccessNode = self.destination_node
+
+        dst_in_edge = graph.in_edges(dst_node)[0]
+        edge_to_redirect = graph.in_edges(src_node)[0]
+        helpers.redirect_edge(
+            graph,
+            edge_to_redirect,
+            new_dst=dst_in_edge.dst,
+            new_dst_conn=dst_in_edge.dst_conn,
+            new_memlet=dace.Memlet(dst_node.data),
```
philip-paul-mueller: The adjustion of the subsets is missing here, I would say.

philip-paul-mueller (commented):
I see a small issue, but it generally looks good.

src/gt4py/next/program_processors/runners/dace/transformations/remove_views.py
iomaganaris: I am not sure what you mean by adjustion of the subsets. The edge subset should be `[0]` in any way since we have a scalar. Is there something extra we can do with this function?

philip-paul-mueller: You have not made sure that `src` not `dst` are scalars.
So the subsets can be anything.
Consider:
```
[REDUCTION] ---> (scalar) [0] ----> [0] (view) [0] ------> [i] (array)
```

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -368,6 +368,12 @@ def _gt_auto_process_top_level_maps(
             ),
         )
 
+    sdfg.apply_transformations_once_everywhere(
+        gtx_transformations.RemovePointwiseViews,
+        validate=validate,
+        validate_all=validate_all,
+    )
+
```
philip-paul-mueller: Technically this function only makes sense for optimizations inside Maps.
Thus I would move it to the appropriate function.
Probably also add a check in `can_be_applied()` if it is in the right scope.

iomaganaris: Which one is the right scope you think? Conceptually this transformation can run in any scope I believe

philip-paul-mueller: I would say something that is not at global scope.
Also something like:
```python
if state.scope_dict()[access_node] is None:
    return False  # Not inside a Map.
else:
    # right scope, inside a Map or something.
    pass
```

src/gt4py/next/program_processors/runners/dace/transformations/remove_views.py
iomaganaris: Should be handled in [3f6ea37](https://github.com/GridTools/gt4py/pull/2166/commits/3f6ea370d414df7697bb4bbd5f6322356462df14)

src/gt4py/next/program_processors/runners/dace/transformations/remove_views.py
```diff
@@ -76,6 +83,10 @@ def can_be_applied(
             return False
         if len(graph.in_edges(src_node)) != 1:
             return False
+        if isinstance(src_desc, dace_data.ArrayView) and src_desc.total_size != 1:
+            return False
+        if graph.out_degree(dst_node) != 1:
+            return False
```
philip-paul-mueller: I am not sure why you check `dst_node` here.
As far as I understand you remove `src_node` and modify the connection between `src_node` and `dst_node`, thus I would guess that you do not care for the downstream nodes of `dst_node`.

I expect however a check that `src_node` has only one outgoing connection, to `dst_node` as you already have.

src/gt4py/next/program_processors/runners/dace/transformations/remove_views.py
```diff
@@ -88,13 +99,14 @@ def apply(
...
         dst_in_edge = graph.in_edges(dst_node)[0]
+        dst_out_edge = graph.out_edges(dst_node)[0]
         edge_to_redirect = graph.in_edges(src_node)[0]
         helpers.redirect_edge(
             graph,
             edge_to_redirect,
             new_dst=dst_in_edge.dst,
             new_dst_conn=dst_in_edge.dst_conn,
-            new_memlet=dace.Memlet(dst_node.data),
+            new_memlet=dace.Memlet(data=dst_node.data, subset=dst_out_edge.data.subset),
```
philip-paul-mueller: This is actually a bit more complicated that what you have.
However, since you assume that `src_node` only has size one, you can simplify the modification a bit.
```
dst_in_edge_dst_subset = copy.deepcopy(dst_in_edge.data.get_dst_subset(dst_in_edge))
new_edge = helpers.redirect_edge(
    graph,
    edge_to_redirect,
    new_dst=dst_in_edge.dst,
    new_dst_conn=dst_in_edge.dst_conn)

if new_edge.data.data == src_node.data:
    new_edge.data.data = dst_node.data
    new_edge.data.subset = dst_in_edge_dst_subset
else:
    new_edge.data.other_subset = dst_in_edge_dst_subset
```
If you think that the above code looks cryptic, this is because it is.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_remove_point_view.py
```diff
@@ -129,3 +129,114 @@ def test_horizontal_map_fusion():
...
+    )
+
+    # Now create the map using the above tasklets
+    mentry, mexit = state.add_map(
+        "simple_map",
+        ndrange={"i": f"0:{N}", "j": "horizontal_start:horizontal_end"},
+    )
+
+    tmp_access_node = state.add_access("tmp")
+    tmp_access_node.transient = True
```
philip-paul-mueller:
```suggestion
```

`.transient` is a property of a data descriptor not of an AccessNode.
You have to look at it like a UNIX compatible filesystem.
The data descriptors are the inodes, that contains all the important information, without the names (and the link count) and the AccessNodes are like the links.

Furthermore,
- I would suggest to first create the data descriptor before you create an AccessNode refering to it.
- You should give them better names, for example here I would name it `view` or something, such that it is clear that you are dealing with a view.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_remove_point_view.py
```diff
@@ -129,3 +129,114 @@ def test_horizontal_map_fusion():
...
+        shape=(1,),
+        dtype=dace.float64,
+    )
+
+    tmp_access_node2 = state.add_transient("tmp2", shape=shape, dtype=dace.float64)
+
+    state.add_edge(a, None, mentry, "IN_a", dace.Memlet("a[i, j]"))
+    state.add_edge(mentry, "OUT_a", task1, "__in0", dace.Memlet("a[i, j]"))
+    state.add_edge(task1, "__out0", tmp_access_node, None, dace.Memlet("tmp[0]"))
+    state.add_edge(tmp_access_node, None, tmp_access_node2, None, dace.Memlet("tmp[0]"))
```
philip-paul-mueller:
```suggestion
    state.add_edge(tmp_access_node, None, tmp_access_node2, None, dace.Memlet("tmp[0] -> [i, j]"))
```

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_remove_point_view.py
```diff
@@ -129,3 +129,114 @@ def test_horizontal_map_fusion():
...
+    state.add_edge(tmp_access_node2, None, mexit, "IN_out", dace.Memlet("tmp2[i, j]"))
+
+    mentry.add_scope_connectors("a")
+    mentry.add_in_connector("IN_a")
+    mexit.add_in_connector("IN_out")
+    mexit.add_out_connector("OUT_out")
+    state.add_edge(mexit, "OUT_out", out, None, dace.Memlet("out[i, j]"))
```
philip-paul-mueller:
```suggestion
    state.add_edge(mexit, "OUT_out", out, None, dace.Memlet("out[0:20, 0:20]"))
```
You are technically outside the map, thus the Map parameters do not exist anymore.
Why it works I do not know, but it is wrong.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_remove_point_view.py
```diff
@@ -129,3 +129,114 @@ def test_horizontal_map_fusion():
...
+            [
+                access_node
+                for access_node, _ in sdfg.all_nodes_recursive()
+                if isinstance(access_node, dace_nodes.AccessNode)
+                and isinstance(access_node.desc(sdfg), dace.data.ArrayView)
+            ]
+        )
+        == 0
+    )
+    assert util.compare_sdfg_res(ref=ref, res=res)
```
philip-paul-mueller, line 179: Your test passes here because your initial version was wrong (the connection from the MapExit to `out`.
In cases such as that I would recommend to write a Python reference instead of running the unprocessed SDFG.

philip-paul-mueller (commented):
There are some issues, but they are rather minor, smallish details.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_remove_point_view.py
iomaganaris: What does `-> [i,j]` mean here?

iomaganaris: Based on what we discussed I guess it's the `other_subset`

philip-paul-mueller: Yes
- `[0]` is `subset`
- `[i, j]` becomes `other_subset`.

Under the assumption that `assert tmp_access_node.data == "tmp"` holds.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_remove_point_view.py
iomaganaris: I fixed the SDFG so it has now the expected behavior. Unfortunately I don't know how to reproduce the unnecessary `View` creation using python code

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_remove_point_view.py
iomaganaris: It actually didn't do the proper thing in the code but only wrote the first element because of another issue. The `subset` in this particular edge data doesn't play a role in the generated code

src/gt4py/next/program_processors/runners/dace/transformations/remove_views.py
iomaganaris: Thanks a lot for the suggestion and the explanation 🙏

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_remove_point_view.py
philip-paul-mueller: What do you mean by that?

iomaganaris: I mean that now the generated SDFG adds `3.0` to all the elements of the array `a`. Before the latest changes the generated code didn't have the expected behavior.
By defining the function that does `c = a + 3.0` in python and decorating with `@dace.program` I'm not sure I can reproduce the generation of the unnecessary view

philip-paul-mueller: In my experience the Python frontend is not stable, and in my view it does not need to be, as long as it is correct.
If you want to assert on a structural/behaviour change of a transformation, you should always use the graph API.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_remove_point_view.py
iomaganaris: >.transient is a property of a data descriptor not of an AccessNode.

I shouldn't have beed able to set it then? 😅 
I have properly initialized `tmp_view` as well

src/gt4py/next/program_processors/runners/dace/transformations/remove_views.py
iomaganaris: Actually in [6242867](https://github.com/GridTools/gt4py/pull/2166/commits/6242867651118fdd069b0ca9ec3e9a5127e07d43)

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_remove_point_view.py
philip-paul-mueller: Python has dynamic attributes unlike C++.

src/gt4py/next/program_processors/runners/dace/transformations/remove_views.py
```diff
@@ -0,0 +1,122 @@
...
+        # or dst_node.data. In this case we have to update the `other_subset` to match the dst_node.data other_subset
+        # which is the subset which gets written to the dst of the edge.
+        if new_edge.data.data == src_node.data:
+            new_edge.data.data = dst_node.data
+            new_edge.data.subset = dst_in_edge_dst_subset
+        else:
+            new_edge.data.other_subset = dst_in_edge_dst_subset
+
+        graph.remove_edge(dst_in_edge)
+        graph.remove_node(src_node)
```
philip-paul-mueller, line 123: The SDFG also has a `remove_data()` function, which removes the data descriptor from the registry.
I think you should also do that.

Be sure to pass `validate=False` to the function.
Without it it will scan the entire SDFG to make sure that it is not used anywhere else.
However from the context we can deduce that this is the case.

iomaganaris: Good catch 👍

PR #2171, opened 2025-07-24
fix[next][dace]: Specify `other_subset` in lowering of scan

edopao:
Not really an issue, more a code improvement. This PR was initially created to address some test failure with the latest dace version, but that will be fixed in dace repository.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
@@ -269,7 +269,7 @@ class DataflowOutputEdge:
             self.state.remove_node(self.result.dc_node)
         else:
             src_node = write_edge.dst
-            src_node_connector = write_edge.dst_conn
+            src_node_connector = None
             src_subset = write_edge.data.dst_subset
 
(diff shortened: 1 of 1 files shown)


philip-paul-mueller (dismissed):
LGTM.

The issue with the no longer working `+` operator was most likely caused by https://github.com/spcl/dace/pull/2096/files

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -263,28 +263,28 @@ def connect(
...
+                dace.Memlet(data=dest.data, subset=dest_subset),
             )
         else:
             self.state.add_memlet_path(
-                last_node,
+                src_node,
                 map_exit,
                 dest,
-                src_conn=last_node_connector,
-                memlet=dace.Memlet(data=dest.data, subset=subset),
```
philip-paul-mueller, line 287: Would it be possible to specify the `other_subset` here as well?

edopao: Done

PR #2173, opened 2025-07-24
feat[next]: Infer unstructured domains of temporaries

SF-N:


tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -2370,7 +2375,11 @@ def test_gtir_concat_where():
         )
 
         # run domain inference in order to add the domain annex information to the concat_where node.
-        testee = infer_domain.infer_program(testee, offset_provider=CARTESIAN_OFFSETS)
+        testee = infer_domain.infer_program(
+            testee,
+            offset_provider=CARTESIAN_OFFSETS,
+            symbolic_domain_sizes=pass_manager._max_domain_range_sizes(CARTESIAN_OFFSETS),
+        )
```
edopao: Cleanup: you can remove the call to `infer_domain.infer_program` because it is already called by default in `build_dace_sdfg()`.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -2454,7 +2463,11 @@ def test_gtir_concat_where_two_dimensions():
     }
 
     # run domain inference in order to add the domain annex information to the concat_where node.
-    testee = infer_domain.infer_program(testee, offset_provider=CARTESIAN_OFFSETS)
+    testee = infer_domain.infer_program(
+        testee,
+        offset_provider=CARTESIAN_OFFSETS,
+        symbolic_domain_sizes=pass_manager._max_domain_range_sizes(CARTESIAN_OFFSETS),
+    )
```
edopao: Same here

src/gt4py/next/ffront/decorator.py
```diff
@@ -170,20 +173,31 @@ def _compiled_programs(self) -> compiled_program.CompiledProgramsPool:
         if self.backend is None or self.backend == eve.NOTHING:
             raise RuntimeError("Cannot compile a program without backend.")
 
-        if self.static_params is None:
-            object.__setattr__(self, "static_params", ())
+        def path_to_expr(path: Sequence[int]) -> str:
+            return "".join(map(lambda idx: f"[{idx}]", path))
```
havogt: move to non-local scope

src/gt4py/next/ffront/decorator.py
```diff
@@ -170,20 +173,31 @@ def _compiled_programs(self) -> compiled_program.CompiledProgramsPool:
         if self.backend is None or self.backend == eve.NOTHING:
             raise RuntimeError("Cannot compile a program without backend.")
 
-        if self.static_params is None:
```
havogt: at some point this ensured that we fixed static_params to the empty set (and cannot be changed afterwards). Is this still checked somewhere?

src/gt4py/next/ffront/decorator.py
```diff
@@ -170,20 +173,31 @@ def _compiled_programs(self) -> compiled_program.CompiledProgramsPool:
...
-        }
+        if self.static_domains:
+            static_domain_args = []
+            func_type = self.past_stage.past_node.type.definition  # type: ignore[union-attr] # type inference done at this point
+            param_types = func_type.pos_or_kw_args | func_type.kw_only_args
+            for name, type_ in param_types.items():
+                for el_type_, path in type_info.primitive_constituents(type_, with_path_arg=True):
+                    if isinstance(el_type_, ts.FieldType):
+                        static_domain_args.append(f"{name}{path_to_expr(path)}")
+            argument_descriptor_mapping[arguments.FieldDomainDescriptor] = static_domain_args
```
havogt: Can we extract that into a free function for readability?

src/gt4py/next/ffront/past_to_itir.py
```diff
@@ -120,6 +120,16 @@ def past_to_gtir(inp: AOT_PRG) -> stages.CompilableProgram:
             body=body,
         )
 
+    # TODO(tehrengruber): Put this in a dedicated transformation step.
+    if arguments.FieldDomainDescriptor in inp.args.argument_descriptor_contexts:
+        context = inp.args.argument_descriptor_contexts[arguments.FieldDomainDescriptor]
```
havogt:
```suggestion
    if (context := inp.args.argument_descriptor_contexts.get(arguments.FieldDomainDescriptor)) is not None:
```
at least to make @egparedes happy.

src/gt4py/next/ffront/past_to_itir.py
```diff
@@ -120,6 +120,16 @@ def past_to_gtir(inp: AOT_PRG) -> stages.CompilableProgram:
...
+        field_domains = {
+            param: utils.tree_map(lambda x: x.domain if x else x)(v) for param, v in context.items()
```
havogt: What is the lambda doing? Especially what when is `not x`?

src/gt4py/next/iterator/transforms/transform_get_domain_range.py
havogt: this name is not perfect. not easy to guess on the call site what that is doing without going here.

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -10,34 +10,27 @@
...
-            sizes[conn_type.source_dim.value] = max(
-                sizes.get(conn_type.source_dim.value, 0), provider.ndarray.shape[0]
-            )
-            sizes[conn_type.codomain.value] = max(
-                sizes.get(conn_type.codomain.value, 0),
-                provider.ndarray.max() + 1,  # type: ignore[attr-defined] # TODO(havogt): improve typing for NDArrayObject
-            )
-    return sizes
+#: Skip printing warnings after exceeding `NON_CONTIGUOUS_DOMAIN_WARNING_THRESHOLD` this many times.
+NON_CONTIGUOUS_DOMAIN_MAX_WARNINGS: int = 5
```
havogt: I would either make this private or we should consider exposing it in `config`. But I guess then we should briefly discuss how we want to handle such cases in general. Also not sure about the MAX_WARNINGS in general. What about warn once and notify that later errors will be suppressed? Or warn always. Why did you decide to go with this mechanism?

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -312,7 +312,7 @@ def create_global_tmps(
     offset_provider: common.OffsetProvider | common.OffsetProviderType,
     #: A dictionary mapping axes names to their length. See :func:`infer_domain.infer_expr` for
     #: more details.
-    symbolic_domain_sizes: Optional[dict[str, str]] = None,
```
havogt, line 318: note to myself: is this new semantics and why was it introduced?

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_program.py
```diff
@@ -82,7 +82,9 @@ def shift_by_one_program(in_field: cases.IFloatField, out_field: cases.IFloatFie
 
 
 def test_copy_execution(cartesian_case, copy_program_def):
-    copy_program = gtx.program(copy_program_def, backend=cartesian_case.backend)
+    copy_program = gtx.program(
+        copy_program_def, backend=cartesian_case.backend, static_domains=True
```
havogt: should we add both cases (otherwise I would keep the default case)

tests/next_tests/unit_tests/otf_tests/test_compiled_program.py
```diff
@@ -49,18 +50,16 @@ def test_sanitize_static_args_wrong_type():
 
 
 @gtx.field_operator
-def fop(cond: bool, a: gtx.Field[gtx.Dims[TDim], float], b: gtx.Field[gtx.Dims[TDim], float]):
-    return a if cond else b
+def fop(cond: bool):
+    return broadcast(cond, (TDim,))
```
havogt: why this change (also not sure why the original case)...

havogt: ok, probably just simplified the test setup...

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -128,7 +129,11 @@ def build_dace_sdfg(
     """
     if not skip_domain_inference:
         # run domain inference in order to add the domain annex information to the IR nodes
-        ir = infer_domain.infer_program(ir, offset_provider=offset_provider)
+        ir = infer_domain.infer_program(
+            ir,
+            offset_provider=offset_provider,
+            symbolic_domain_sizes=pass_manager._max_domain_range_sizes(offset_provider),
```
havogt, line 135: Why this change?

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_domain.py
```diff
@@ -20,11 +20,64 @@
 from gt4py.next.iterator.ir_utils import domain_utils, ir_makers as im
 
 from next_tests.integration_tests.feature_tests.ffront_tests.ffront_test_utils import (
+    Cell,
     KDim,
     Vertex,
 )
 
 
+@pytest.mark.parametrize(
```
havogt: this is adding a test for existing functionality?

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -144,28 +137,79 @@ def translate(
...
-
                 old_dim = connectivity_type.source_dim
                 new_dim = connectivity_type.codomain
-
                 assert new_dim not in new_ranges or old_dim == new_dim
+                if symbolic_domain_sizes is not None and new_dim.value in symbolic_domain_sizes:
+                    new_range = SymbolicRange(
+                        im.literal(str(0), builtins.INTEGER_INDEX_BUILTIN),
+                        im.ensure_expr(symbolic_domain_sizes[new_dim.value]),
+                    )
```
havogt, line 210: Are we still using this code path? I forgot what we discussed...

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -144,28 +137,79 @@ def translate(
...
+                else:
+                    assert common.is_offset_provider(offset_provider)
```
havogt, line 212: Would it make sense to move this into a function (maybe also the true branch). Maybe at some point we had the idea of extracting the unstructured shift domain inference into function that the user could pass, no?

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -43,6 +45,70 @@ def __call__(
     ) -> itir.Program: ...
 
 
+def _max_domain_range_sizes(offset_provider: Mapping[str, Any]) -> dict[str, itir.Literal]:
```
havogt:
```suggestion
def _max_domain_range_sizes(offset_provider: common.OffsetProvider) -> dict[str, itir.Literal]:
```
?

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_domain.py
```diff
@@ -20,11 +20,64 @@
...
+            Cell: ("horizontal_start", "horizontal_end"),
+            KDim: ("vertical_start", "vertical_end"),
+        },
+    )
+    domain = gtx_dace_domain.extract_domain(domain_expr)
+
+    expr = dace.symbolic.pystr_to_symbolic(param[0])
+    expected_expr = dace.symbolic.pystr_to_symbolic(param[1])
+
+    assert gtx_dace_domain.simplify_domain_expr(expr, domain) == expected_expr
```
tehrengruber: @edopao The `simplify_domain_expr` has been removed in main, could you take a look on how to fix this failing test?

edopao: I think you should just revert all changes in this file. It looks like the result of rebasing from an older baseline.

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -144,28 +137,79 @@ def translate(
...
+                    range_exprs = tuple(
+                        collapse_tuple.CollapseTuple.apply(
+                            expr,
+                            within_stencil=False,
+                            allow_undeclared_symbols=True,
+                        )
+                        for expr in range_exprs
+                    )  # type: ignore[assignment]  # mypy not smart enough
+                    assert all(isinstance(expr, itir.Literal) for expr in range_exprs)
+                    start, stop = (int(literal.value) for literal in range_exprs)  # type: ignore[attr-defined]  # mypy does not understand assert above
```
havogt: minor (but I find it more readable to keep the start/stop separate, instead of wrapping it in a (non-named) tuple with start/stop semantics.

```suggestion
                    (start_expr, stop_expr) = new_ranges[old_dim].start, new_ranges[old_dim].stop
                    (start_expr, stop_expr) = tuple(
                        collapse_tuple.CollapseTuple.apply(
                            expr,
                            within_stencil=False,
                            allow_undeclared_symbols=True,
                        )
                        for expr in (start_expr, stop_expr)
                    )  # type: ignore[assignment]  # mypy not smart enough
                    assert all(isinstance(expr, itir.Literal) for expr in range_exprs)
                    start, stop = (int(literal.value) for literal in (start_expr, stop_expr))  # type: ignore[attr-defined]  # mypy does not understand assert above
```

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -144,28 +137,79 @@ def translate(
...
+                    range_exprs = new_ranges[old_dim].start, new_ranges[old_dim].stop
+                    range_exprs = tuple(
+                        collapse_tuple.CollapseTuple.apply(
+                            expr,
+                            within_stencil=False,
+                            allow_undeclared_symbols=True,
+                        )
+                        for expr in range_exprs
+                    )  # type: ignore[assignment]  # mypy not smart enough
+                    assert all(isinstance(expr, itir.Literal) for expr in range_exprs)
```
havogt: probably (see other comment) will make the type ignore work by spending a few more characters
```suggestion
                    assert isinstance(start_expr, itir.Literal)
                    assert isinstance(stop_expr, itir.Literal)
```

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -144,28 +137,79 @@ def translate(
...
+                            UserWarning(
+                                f"Translating '{self.as_expr()}' using '{shift[0].value}' has "
+                                f"an out-of-bounds access."
+                            ),
+                            stacklevel=2,
+                        )
+
+                    new_start, new_stop = accessed.min(), accessed.max() + 1  # type: ignore[attr-defined]  # TODO(havogt): improve typing for NDArrayObject
+
+                    fraction_accessed = np.unique(accessed).size / (new_stop - new_start)  # type: ignore[call-overload]  # TODO(havogt): improve typing for NDArrayObject
```
havogt:
```suggestion
                    fraction_accessed = xp.unique(accessed).size / (new_stop - new_start)  # type: ignore[call-overload]  # TODO(havogt): improve typing for NDArrayObject
```
?
or where do you make sure it lives on host?

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_compiled_program.py
```diff
@@ -849,3 +850,54 @@ def test_wait_for_compilation(cartesian_case, compile_testee, compile_testee_dom
...
+
+    class CaptureCompileTimeArgsBackend:
+        def __getattr__(self, name):
+            return getattr(cartesian_case.backend, name)
+
+        def compile(self, program, compile_time_args):
+            nonlocal captured_cargs
+            captured_cargs = compile_time_args
+
+            return cartesian_case.backend.compile(program, compile_time_args)
```
havogt, line 961: Should probably use `mock`?

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -57,12 +123,19 @@ def apply_common_transforms(
     force_inline_lambda_args=False,
     #: A dictionary mapping axes names to their length. See :func:`infer_domain.infer_expr` for
     #: more details.
-    symbolic_domain_sizes: Optional[dict[str, str]] = None,
+    symbolic_domain_sizes: Optional[dict[str, str | itir.Expr]] = None,
+    # TODO(tehrengruber): Remove this option again as soon as we have the necessary builtins
+    #  to work with / translate domains.
+    use_max_domain_range_on_unstructured_shift: Optional[bool] = None,
```
havogt: Can you document somewhere (not sure where, ADR?) all the different code paths that we allow for unstructured domain inference and related topics. It's a while that we talked about it and it seems I need your explanation again. But then it's probably better to write it down.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_domain.py
havogt: I wrote the above before Edoardo's comment.

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: Idea from discussion with Enrique: Raise once per tag.

src/gt4py/next/ffront/past_to_itir.py
tehrengruber: I've made the `None` value more explicit with:
```
utils.tree_map(lambda x: x.domain if x is not None else x)(v)
```
I could also revert to the non walrus operator version and add a type annotation for `context`, but this also didn't look pretty.

tests/next_tests/unit_tests/otf_tests/test_compiled_program.py
tehrengruber: Exactly, this was just to simplify the test setup.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
tehrengruber: Broken merge.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_domain.py
tehrengruber: Broken merge.

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: Before this PR we didn't use the mechanism anymore, here I hijacked the "old" mechanism to provide the *compute everywhere* semantics. See `_process_symbolic_domains_option` where the values are populated.

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: See https://numpy.org/devdocs/user/basics.interoperability.html#example-cupy-arrays
> The cupy.ndarray object implements the __array_ufunc__ interface. This enables NumPy ufuncs to be applied to CuPy arrays (this will defer operation to the matching CuPy CUDA/ROCm implementation of the ufunc):

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_compiled_program.py
tehrengruber: Sadly doesn't work, since `backend.transforms` needs to be an instance of `Transforms` which is not the case with mock.

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -43,6 +45,70 @@ def __call__(
...
+        .filter(lambda node: cpm.is_call_to(node, "as_fieldop") and len(node.args) == 2)
+    ):
+        domains.add(as_fop.args[1])
+    return len(symbol_ref_utils.collect_symbol_refs(domains)) > 0
+
+
+def _process_symbolic_domains_option(
+    ir: itir.Program,
+    offset_provider: common.OffsetProvider | common.OffsetProviderType,
+    symbolic_domain_sizes: Optional[dict[str, str | itir.Expr]],
```
```suggestion
    symbolic_domain_sizes: Optional[dict[str, itir.Expr]],
```
Try if this is possible everywhere with low effort.

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -43,6 +45,70 @@ def __call__(
...
+    for as_fop in (
+        ir.walk_values()
+        .if_isinstance(itir.FunCall)
+        .filter(lambda node: cpm.is_call_to(node, "as_fieldop") and len(node.args) == 2)
+    ):
+        domains.add(as_fop.args[1])
+    return len(symbol_ref_utils.collect_symbol_refs(domains)) > 0
+
+
+def _process_symbolic_domains_option(
```
tehrengruber, line 83: Document in this docstring the current behavior.

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: Result from discussion: Make this private until things have settled.

src/gt4py/next/ffront/decorator.py
tehrengruber: Yes, this check here was redundant after #/2258. The check is here: https://github.com/GridTools/gt4py/blob/main/src/gt4py/next/otf/compiled_program.py#L368 It works slightly different: In case there is no `StaticArg` in the argument descriptor mapping, no static params are folded, which is equivalent to no static params.

src/gt4py/next/iterator/transforms/transform_get_domain_range.py
tehrengruber: How about `ReplaceGetDomainRangeWithConstants`? Or alternatively something like `TransformDynamicIntoStaticDomains`.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_program.py
tehrengruber: This was a left-over, `test_compile_variants_decorator_static_domains` has good enough coverage I would say.

src/gt4py/next/iterator/transforms/pass_manager.py
tehrengruber: Done here: https://github.com/SF-N/gt4py/blob/infer_unstructured_domains_of_temporaries/src/gt4py/next/iterator/transforms/pass_manager.py#L90

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: I put the else into a function which improved readability.

src/gt4py/next/iterator/ir_utils/domain_utils.py
havogt: Wasn't aware.

src/gt4py/next/iterator/transforms/pass_manager.py
havogt: Still open?

src/gt4py/next/iterator/transforms/transform_get_domain_range.py
havogt: :+1: Verbose, but easy to understand.

src/gt4py/next/ffront/past_to_itir.py
havogt: that helps!

src/gt4py/next/iterator/ir_utils/domain_utils.py
egparedes: Are we really using `ufunc` dispatching anywhere else in our codebase? I'm not sure but I'd say not, which probably explains why this looked confusing to Hannes.

I would argue that unless we take a clear (and documented) decision of using array dispatching mechanisms like  `__array_ufunc__`, `__array_function__` or `__array_namespace__`, it's better to stick to a old-fashioned explicit `np`/`xp` dispatching.

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -19,25 +22,12 @@
 from gt4py.next.iterator.transforms.constant_folding import ConstantFolding
 
 
-def _max_domain_sizes_by_location_type(offset_provider: Mapping[str, Any]) -> dict[str, int]:
-    """
-    Extract horizontal domain sizes from an `offset_provider`.
+#: Threshold fraction of domain points which may be added to a domain on translation in order
+#: to have a contiguous domain before a warning is raised.
+_NON_CONTIGUOUS_DOMAIN_WARNING_THRESHOLD: float = 1 / 4
```
egparedes, line 27: Question: would it make sense to expose this as a configuration option?

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -68,6 +58,72 @@ def empty(self) -> bool | None:
...
+    assert isinstance(start_expr, itir.Literal) and isinstance(stop_expr, itir.Literal)
+    start, stop = int(start_expr.value), int(stop_expr.value)
+
+    nb_index: slice | int
+    if val in [trace_shifts.Sentinel.ALL_NEIGHBORS, trace_shifts.Sentinel.VALUE]:
+        nb_index = slice(None)
+    else:
+        nb_index = val.value  # type: ignore[assignment]  # assert above
+
+    accessed = connectivity.ndarray[start:stop, nb_index]
```
egparedes, line 90: Is guaranteed at this point that the connectivity is well formed (i.e. 2d with the second axis being the neighbor dimension) or it should be asserted/checked?

src/gt4py/next/iterator/ir_utils/ir_makers.py
```diff
@@ -453,7 +453,7 @@ def _impl(*its: itir.Expr) -> itir.FunCall:
 
 def domain(
     grid_type: Union[common.GridType, str],
-    ranges: dict[common.Dimension, tuple[itir.Expr, itir.Expr]],
+    ranges_or_domain: dict[common.Dimension, tuple[itir.Expr, itir.Expr]] | common.Domain,
```
egparedes, line 456: Would it make sense to add an example of the new case to the doctests?

src/gt4py/next/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -52,7 +52,8 @@ class GTFNTranslationStep(
     enable_itir_transforms: bool = True
     use_imperative_backend: bool = False
     device_type: core_defs.DeviceType = core_defs.DeviceType.CPU
-    symbolic_domain_sizes: Optional[dict[str, str]] = None
+    symbolic_domain_sizes: Optional[dict[str, itir.Expr]] = None
+    use_max_domain_range_on_unstructured_shift: Optional[bool] = None
```
egparedes:
```suggestion
    symbolic_domain_sizes: dict[str, itir.Expr] | None = None
    use_max_domain_range_on_unstructured_shift: bool | None = None
```

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -369,6 +369,7 @@ class DaCeTranslator(
 
     disable_itir_transforms: bool = False
     disable_field_origin_on_program_arguments: bool = False
+    use_max_domain_range_on_unstructured_shift: Optional[bool] = None
```
egparedes:
```suggestion
    use_max_domain_range_on_unstructured_shift: bool | None = None
```

src/gt4py/next/ffront/decorator.py
```diff
@@ -191,6 +198,17 @@ def compile(
         return self
 
 
+def _field_domain_descriptor_mapping_from_func_type(func_type: ts.FunctionType) -> list[str]:
+    static_domain_args = []
+    param_types = func_type.pos_or_kw_args | func_type.kw_only_args
+    for name, type_ in param_types.items():
+        for el_type_, path in type_info.primitive_constituents(type_, with_path_arg=True):
+            if isinstance(el_type_, ts.FieldType):
+                path_as_expr = "".join(map(lambda idx: f"[{idx}]", path))
```
egparedes: I think this can be done in an easier way:
```suggestion
                path_as_expr = "".join(f"[{idx}]" for idx in path))
```
Question: named collections at this point have been already transformed to pure tuples, right? Otherwise this won't work for named collections.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_compiled_program.py
```diff
@@ -936,3 +937,54 @@ def test_wait_for_compilation(cartesian_case, compile_testee, compile_testee_dom
...
+
+    @gtx.field_operator
+    def identity_like(inp: tuple[cases.IField, cases.IField, float]):
+        return inp[0], inp[1]
+
+    # the float argument here is merely to test that static domains work for tuple arguments
+    # of inhomogeneous types
+    @gtx.program(backend=CaptureCompileTimeArgsBackend(), static_domains=True)
```
egparedes, line 969: Should we add a test case for named collections or do you think it's not needed?

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -42,26 +44,120 @@ def __call__(
...
+    sizes: dict[str, int] = {}
+    for provider in offset_provider.values():
+        if common.is_neighbor_connectivity(provider):
+            conn_type = provider.__gt_type__()
+            sizes[conn_type.source_dim.value] = max(
+                sizes.get(conn_type.source_dim.value, 0), provider.ndarray.shape[0]
+            )
+            sizes[conn_type.codomain.value] = max(
+                sizes.get(conn_type.codomain.value, 0),
+                int(provider.ndarray.max()) + 1,  # type: ignore[attr-defined] # TODO(havogt): improve typing for NDArrayObject
```
egparedes, line 62:
```suggestion
        if common.is_neighbor_connectivity(provider):
            src_dim = provider.__gt_type__().source_dim.value,
            codomain_dim = provider.__gt_type__().codomain.value,
            sizes[src_dim] = max(sizes.get(src_dim, 0), provider.ndarray.shape[0])
            sizes[codomain_dim] = max(
                sizes.get(codomain_dim, 0),
                int(provider.ndarray.max()) + 1,  # type: ignore[attr-defined] # TODO(havogt): improve typing for NDArrayObject
```

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: I don't know if this is 
- a hypothetical case and most meshes are ordered properly anyway
- if one should always reorder the mesh since the overhead becomes to large
- it depends and users want to configure this
I therefore made it private for now and we can revisit this when we know more.

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: Good catch, when we wrote this I was actually not aware a Connectivity is not necessarily two dimensional with a regular and a local index. I test for `NeighborConnectivity` now.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_compiled_program.py
tehrengruber: Yes, it actually revealed a small but.

PR #2174, opened 2025-07-24
feat[next[dace]: Simplify symbolic expression based on domain range

edopao:
This PR introduces a `SubgraphContext` for the lowering to SDFG. In addition to the existing `sdfg` and `head_state`, his structure also stores the domain of the target field found in the top-level `SetAt` expression. The range of the target field is used to make assumptions on some symbols, such as `range_end >= range_start`. These assumptions are applied through sympy to simplify the symbolic shape of temporary arrays and memlet subsets.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_domain.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_domain.py
@@ -12,6 +12,7 @@ import dataclasses
 from typing import Optional, Sequence, TypeAlias
 
 import dace
+import sympy
 from dace import subsets as dace_subsets
 
 from gt4py.next import common as gtx_common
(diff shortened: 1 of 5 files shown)


src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -77,6 +78,31 @@ def extract_domain(node: gtir.Expr) -> FieldopDomain:
...
+    Returns:
+        A new symbolic expression.
+    """
+    for dim_range in domain:
+        # We want to enforce the constraint `ub = lb + size`. The actual constraint
+        # is given by the assumption that the `size` variable is integer and non-negative.
+        size = sympy.var(f"__gtir_{dim_range.dim.value}_size", integer=True, negative=False)
+        expr = expr.subs(dim_range.start, dim_range.stop - size).subs(
+            size, dim_range.stop - dim_range.start
+        )
```
philip-paul-mueller, line 100:
```suggestion
        expr = expr.subs({dim_range.start: dim_range.stop - size, size: dim_range.stop - dim_range.start})
```
As far as I know this should works and should also be faster, because it is batched.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -77,6 +78,31 @@ def extract_domain(node: gtir.Expr) -> FieldopDomain:
...
+    return expr.simplify()
```
philip-paul-mueller: DaCe has its [own version](https://github.com/spcl/dace/blob/956312fa4220e0ee934e693103db5f515d86944d/dace/symbolic.py#L1012) of simplify.
I think it does not matter much, but for consistency I would suggest you to use it.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -104,7 +130,8 @@ def get_domain_indices(
 
 
 def get_field_layout(
-    domain: FieldopDomain,
+    field_domain: FieldopDomain,
```
philip-paul-mueller, line 131: I would use `source_domain` instead of `field_domain`, "field" is a bit too neutral compared to the super specific `target_domain`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -686,14 +709,14 @@ def visit_SetAt(
           The SDFG head state, eventually updated if the target write requires a new state.
         """
 
-        source_fields = self._visit_expression(stmt.expr, sdfg, state)
+        # visit the domain expression
+        domain = gtir_domain.extract_domain(stmt.domain)
+
+        source_fields = self._visit_expression(stmt.expr, domain, sdfg, state)
```
```suggestion
        domain = gtir_domain.extract_domain(stmt.domain)
        source_fields = self._visit_expression(stmt.expr, domain, sdfg, state)
```

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
edopao: I tried that but it does not work.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
edopao: I agree, done.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
edopao: Although `source_domain` looks better in the function signature, this is actually the domain of a field operator.

philip-paul-mueller (approved):
LGTM.

PR #2177, opened 2025-07-25
feat[dace][next]: Less Extensive Validation During `gt_auto_optimizer()`

philip-paul-mueller:
The original idea of the `validate` and `validate_all` flags (which are inherited from DaCe) are to either perform validation at the very end or extensively during the operation.
This means that `validate` must be set to `False` for every subset the auto optimizer does.
This was missing before.


src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
```diff
@@ -348,14 +357,19 @@ def gt_substitute_compiletime_symbols(
...
         )
     dace.sdfg.propagation.propagate_memlets_sdfg(sdfg)
 
+    if validate:
+        sdfg.validate()
+
 
 def gt_reduce_distributed_buffering(
     sdfg: dace.SDFG,
+    validate: bool = True,
```
edopao, line 399: Why is `validate` default True here, but False in the method above (`gt_substitute_compiletime_symbols`)?

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -165,46 +172,66 @@ def split_maps(
         first_map_exit: dace_nodes.MapExit,
         second_map_entry: dace_nodes.MapEntry,
         second_map_exit: dace_nodes.MapExit,
-    ) -> None:
-        """Split the map range in order to obtain an overlapping range between the first and second map."""
-        splitted_range = map_fusion_utils.split_overlapping_map_range(
+    ) -> tuple[list[dace_nodes.MapEntry], list[dace_nodes.MapEntry]]:
```
edopao: Should maybe the changes from this line and below remain in the original PR?

src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
philip-paul-mueller: Because I think that a method should check if it worked correctly.

edopao: I agree. Should we set the same default on the other method?

philip-paul-mueller: Good point.
Done.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
philip-paul-mueller: You are right.
I have reverted these changes.

PR #2178, opened 2025-07-25
feat[dace][next]: Optimized `gt_inline_nested_sdfg()`.

philip-paul-mueller:
Before we can inline a nested SDFG, we have to perform a cleaning step (`PruneSymbols` and `PruneConnectors`) on it.
Originally this was done in a two stage way, i.e. first calling `sdfg.apply_transformations_repeated([dace_dataflow.PruneSymbols, dace_dataflow.PruneConnectors], ...)` and then the inline transformation.

In the new design these two stages were unified.
Thus we scan the SDFG recursively and apply apply the cleaning transformation and the inline transformation manually.
Thus only a single pass through the SDFG is needed.

For this it was necessary to replicate some functionality inside DaCe into GT4Py.
However, a major difference to DaCe is that we first try to use `InlineSDFG`, which is only able to handle nested SDFG with a single state, before we try `MultistateInlineSDFG`.
The main reason is that the majority of our SDFGs only have one state and that `InlineSDFG` does not create additional states.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_domain.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_domain.py
@@ -12,6 +12,7 @@ import dataclasses
 from typing import Optional, Sequence, TypeAlias
 
 import dace
+import sympy
 from dace import subsets as dace_subsets
 
 from gt4py.next import common as gtx_common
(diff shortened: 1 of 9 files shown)


philip-paul-mueller:
NOTE:
This PR introduces a bug into the toolchain that might lead to the elimination of writes.
The bug originates from within DaCe and is currently under investigation.

PR #2180, opened 2025-07-25
feat[dace][next]: Introduce different Skip Categorieds During Optimization

philip-paul-mueller:
Before all parts of simplify where run every time.
This PR introduces a more fine grained behaviour, by using the `skip` mechanism, that is already there.
For this three different lists were defined:
- For the initial simplification.
- For phase one, in which the top level dataflow is optimized.
- For phase two, in which the dataflow inside maps is optimized.

The main objective is to improve the performance of the tool chain by deactivating parts of simplify that are not needed.


src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -20,6 +20,55 @@
...
+        "RemoveUnusedSymbols",
+        "ReferenceToView",
+        "ConsolidateEdges",
+    }
+)
+"""Simplify stages disabled during the optimization of dataflow of the top level Maps"."""
+
+
+_GT_AUTO_OPT_INNER_DATAFLOW_STAGE_SIMPLIFY_SKIP_LIST: Final[set[str]] = (
+    _GT_AUTO_OPT_INITIAL_STEP_SIMPLIFY_SKIP_LIST
```
edopao: As far as I can see, this set contains all stages defined in `_GT_AUTO_OPT_TOP_LEVEL_STAGE_SIMPLIFY_SKIP_LIST`. Should we extend this one, instead of `_GT_AUTO_OPT_INITIAL_STEP_SIMPLIFY_SKIP_LIST`?

edopao (commented):
LGTM. Was this PR included in the latest check of performance regression?

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
philip-paul-mueller: This is an interesting point.
I think your solution is cleaner.
I am also unsure if we should remove `_GT_AUTO_OPT_INNER_DATAFLOW_STAGE_SIMPLIFY_SKIP_LIST` as it is the default?

edopao: You were probably referring to `_GT_AUTO_OPT_INITIAL_STEP_SIMPLIFY_SKIP_LIST`. Yes, if it is the same as the default, you can remove it.

src/gt4py/next/program_processors/runners/dace/transformations/constants.py
```diff
@@ -0,0 +1,71 @@
...
+
+The following passes are included:
+- `ScalarToSymbolPromotion`: The lowering has sometimes to turn a scalar into a
+    symbol or vice versa and at a later point to invert this again. However, this
+    pass has some problems with this pattern so for the time being it is disabled.
+- `ConstantPropagation`: Same reasons as `ScalarToSymbolPromotion`.
+"""
+
+
+_GT_AUTO_OPT_INITIAL_STEP_SIMPLIFY_SKIP_LIST: Final[set[str]] = GT_SIMPLIFY_DEFAULT_SKIP_SET
```
edopao, line 25: OK, so by naming convention this constant is supposed to be used only inside the `transformations` module.

edopao (approved):
Better, I like the new solution.

PR #2181, opened 2025-07-25
feat[dace][next]: Optimized Split-Fusion Transformations

philip-paul-mueller:
This PR optimizes the performance of the split transformation (not the result but their runtime).
It renames them to avoid confusion with the regular fusion transformations.


src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -283,16 +337,80 @@ def apply(
...
+            for second_map_entry in unmatched_second_map_entries:
+                if first_map_entry.map.range == second_map_entry.map.range:
+                    matching_maps.append(
+                        (
+                            first_map_entry,
+                            graph.exit_node(first_map_entry),
+                            second_map_entry,
+                            graph.exit_node(second_map_entry),
+                        )
+                    )
```
iomaganaris:
```suggestion
                    matching_maps.append(
                        (
                            first_map_entry,
                            first_map_exit,
                            second_map_entry,
                            second_map_exit,
                        )
                    )
```
Does this make any difference here?

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -261,16 +306,25 @@ def can_be_applied(
             if isinstance(iedge.src, dace_nodes.AccessNode)
         }
 
+        # Test if the Maps are parallel.
+        # TODO(iomaganaris): Before this test was located in the `apply()` function. However,
+        #   Its comment indicated that the `can_be_applied()` function, has performed that
+        #   test already, but I can not see it. Because I, phimuell, also thinks that this
+        #   check should be done here, I relocated it here. Could you conform that this move
+        #   is valid.
```
iomaganaris: I agree that this check makes sense here

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
philip-paul-mueller: The variable `first_map_entry` (on line 362) and by that `first_map_exit` are different from the variables defined on line 335-336.
Here they refer to the result of the splitting of Maps.

So the answer to your question is yes, it would make a difference.
However, the answer to the implied question "is the current naming good" is certainly no.
So I will change the naming.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
philip-paul-mueller: I removed the todo.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
philip-paul-mueller: I have updated the name to reflect what the different objects are now.

iomaganaris: Ah right. I got confused by the naming. Thanks for fixing this 👍

philip-paul-mueller: No problem, the naming _was_ bad.

PR #2182, opened 2025-07-25
feat[next]: Check inout field

SF-N:
This new pass checks within a `SetAt` if any fields that are written are also read with an offset and raises a ValueError in case.
It should run before `create_global_tmps`, since the shifts of the temporary targets are not forwarded.

Motivation: Warn user if he uses an in-out field
```
@gtx.field_operator
def field_op(v_field):
    return v_field(e2v[0])

@gtx.program
def program(v_field):
    field_op(v_field, out=v_field)
```

TODOs:
- [ ] Mechanism to also warn if inputs are aliased, e.g. for the case where the field operator is called directly -> New PR
- [ ] Fix tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 Notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [1.0.7] - 2025-08-12
+
+### Cartesian
+
+- Introduce switches for the default `int` and `float` precision.
+- Introduce `erf` and `erfc` functions.
+- Make CUDA compilation thread-safe.
+
+### Development
+
+- Add MacOS to daily CI.
+
+### Next
+
+See commit history.
+
 ## [1.0.6] - 2025-07-30
 
 ### Cartesian
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -449,7 +449,7 @@ url = 'https://test.pypi.org/simple'
 atlas4py = {index = "test.pypi"}
 dace = [
   {git = "https://github.com/GridTools/dace", branch = "romanc/stree-to-sdfg", group = "dace-cartesian"},
-  {git = "https://github.com/GridTools/dace", tag = "__gt4py-next-integration_2025_08_06", group = "dace-next"}
+  {git = "https://github.com/GridTools/dace", tag = "__gt4py-next-integration_2025_08_12", group = "dace-next"}
 ]
 
 # -- versioningit --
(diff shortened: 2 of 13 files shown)


src/gt4py/next/iterator/transforms/check_inout_field.py
```diff
@@ -0,0 +1,106 @@
...
+                                raise ValueError(
+                                    f"The target {node.target} is also read with an offset."
+                                )
+                if cpm.is_applied_as_fieldop(arg):
+                    check_expr(arg.fun, arg.args, offset_provider)
+
+        if cpm.is_applied_as_fieldop(node.expr):
+            check_expr(node.expr.fun, node.expr.args, offset_provider)
+        else:  # Account for im.make_tuple
+            if hasattr(node.expr, "args"):
```
tehrengruber:
```suggestion
            if cpm.is_call_to(node, "make_tuple"):
```
How about tuple_get? Error if expression is not understood?

src/gt4py/next/iterator/transforms/check_inout_field.py
```diff
@@ -0,0 +1,106 @@
...
+    ):
+        return cls().visit(program, offset_provider=offset_provider)
+
+    def visit_SetAt(self, node: itir.SetAt, **kwargs) -> itir.SetAt:
+        offset_provider = kwargs["offset_provider"]
+
+        def extract_subexprs(expr):
+            """Return a list of all subexpressions in expr.args, including expr itself."""
+            subexprs = [expr]
+            if hasattr(expr, "args"):
```
tehrengruber: What cases are these?

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_check_inout_field.py
```diff
@@ -0,0 +1,338 @@
+# GT4Py - GridTools Framework
```
tehrengruber, line 1: Without reading the tests, these cases come up to my mind:
```
{a, b} <- {as_fieldop(...), as_fieldop(...)}
{a, b} <- as_fieldop(...)
a[0] <- as_fieldop(...)
a[0] <- {as_fieldop(...), as_fieldop(...)}
{a[0], a[1]} <- as_fieldop(...)
{a[0], a[1]} <- {as_fieldop(...), as_fieldop(...)}
a <- {as_fieldop(...), as_fieldop(...)}
```

src/gt4py/next/iterator/transforms/check_inout_field.py
SF-N: `make_tuple`, `tuple_get` and all functions like `plus`, `maximum`, ...
I am checking for `FunCall` now.

PR #2184, opened 2025-07-28
refactor[cartesian, dace]: avoid duplicate transient check in `replace_strides()`

romanc:
## Description

We replacing transient strides, we assemble a list of transient arrays in `replace_transient_strides()`, which we then pass to `replace_strides()`. Replace strides will (contrary to it's name) check again and only replace strides of transient arrays. This is unexpected and duplicate work. We don't use `replace_strides()` from anywhere else than `replace_transient_strides()`, so it's "safe" to remove `if array.transient` even if this alters the functionality of that function because it doesn't matter in the one use-case that we have.

The change includes fully typing `replace_strides()`, which is achieved by changing the second argument from `get_layout_map` to include the full `layout_info` object. That `layout_info` object just has a simpler type. Doing so, `mypy` started complaining about inconsistent types of `layout_map` functions, which I fixed as part of this PR.

## Requirements

  Covered by existing tests.
  N/A

### DIFF
--- a/src/gt4py/cartesian/backend/dace_backend.py
+++ b/src/gt4py/cartesian/backend/dace_backend.py
@@ -62,7 +62,7 @@ def _specialize_transient_strides(
             for array in sdfg.arrays.values()
             if isinstance(array, data.Array) and array.transient
         ],
-        layout_info["layout_map"],
+        layout_info,
     )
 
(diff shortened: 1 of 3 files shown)


twicki (approved):
Changes make sense. Thanks for cleaing this up!

PR #2185, opened 2025-07-28
refactor[cartesian, dace]: simplify after specializing strides in stencil mode

romanc:
## Description

The `dace:*` backends have two modes: classic "stencil mode" and "jit mode" used for what [NDSL](https://github.com/NOAA-GFDL/NDSL) calls "orchestration". This PR concerns the classic stencil mode. In that mode, we use the oir -> schedule tree -> sdfg pipeline to get a symbolic SDFG of the given stencil. In that symbolic SDFG, we can specialize (read "replace symbols with numbers") strides of transient arrays. This PR suggest to run `SDFG.simplify()` after specializing the strides, which potentially allows for further optimizations after the strides have been propagated.

## Requirements

  Covered by existing tests.
  N/A

### DIFF
--- a/src/gt4py/cartesian/backend/dace_backend.py
+++ b/src/gt4py/cartesian/backend/dace_backend.py
@@ -18,6 +18,7 @@ from dace import SDFG, Memlet, SDFGState, config, data, dtypes, nodes, subsets, 
 from dace.codegen import codeobject
 from dace.sdfg.analysis.schedule_tree import treenodes as tn
 from dace.sdfg.utils import inline_sdfgs
+from dace.transformation.passes import SimplifyPass
 
 from gt4py._core import definitions as core_defs
 from gt4py.cartesian import config as gt_config, definitions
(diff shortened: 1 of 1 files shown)


romanc:
Moving back to draft since - apparently - this breaks tests. To be investigated.

romanc:
> This sounds like a step in the right direction. Do we know that this helps us anywhere in particular? Or is this a hunch-based change?

No hard data on this one. Just a note in the [laundry list](https://geos-esm.github.io/SMT-Nebulae/technical/laundry-list/#planned-optimizations). @FlorianDeconinck found this while looking at code. We agreed that `simplify()` generally makes codegen easier and thus might lead to simpler generated code, which makes the task of the general compiler easier. It probably won't break the bank, just good manners.

FlorianDeconinck:
We are specializing transient strides in stencil mode, isn't that breaking caching ? GT4py stencils should be dimensionless, are we sure only strides get touched and not sizes?

PR #2187, opened 2025-07-28
feat[cartesian]: literal precision (float and int)

romanc:
## Description

This PR adds support for specifying the default literal precision (32 or 64 bits) when using `int` or `float` declarations/casts. The default literal precision determined from the system's architecture and can be overwritten by the environment variables `GT4PY_LITERAL_INT_PRECISION` and `GT4PY_LITERAL_FLOAT_PRECISION`. Build options allow to specify the precision per stencil if needed.

 The PR also adds support for explicit casting through `int32` / `int64` and `float32` / `float64` casts.

[Compared](https://github.com/GridTools/gt4py/pull/2187/files/5d9a37e21a44e199421611e6f77a91df7d725f9f..606f8a29d43c4dfa93eca4858c184740cd4b07c0) to work by @FlorianDeconinck, we have

- separated `int`  and  `float` precision specification
- consolidated the many literal precision defaults into `definitions.LITERAL_{INT, FLOAT}_PRECISION`
- don't expose `int` and `float` in `gtscript`'s `__ALL__` field. They aren't defined and doctests thus complained.
- use `ast.unparse` directly (the wrapper function was removed recently because we dropped support for python < 3.10)
- remove print debug statements

## Requirements

### DIFF
--- a/src/gt4py/cartesian/frontend/gtscript_frontend.py
+++ b/src/gt4py/cartesian/frontend/gtscript_frontend.py
@@ -736,7 +736,8 @@ class IRMaker(ast.NodeVisitor):
         self.parameters = parameters
         self.local_symbols = local_symbols
         self.domain = domain or nodes.Domain.LatLonGrid()
-        self.literal_precision = options.literal_precision
+        self.literal_int_precision = options.literal_int_precision
+        self.literal_float_precision = options.literal_float_precision
         self.temp_decls = temp_decls or {}
         self.parsing_context = None
         self.iteration_order = None
@@ -779,12 +780,12 @@ class IRMaker(ast.NodeVisitor):
             "float64": nodes.NativeFunction.FLOAT64,
             "int": (
                 nodes.NativeFunction.INT32
-                if self.literal_precision == 32
+                if self.literal_int_precision == 32
                 else nodes.NativeFunction.INT64
             ),
             "float": (
                 nodes.NativeFunction.FLOAT32
-                if self.literal_precision == 32
+                if self.literal_float_precision == 32
                 else nodes.NativeFunction.FLOAT64
             ),
             "erf": nodes.NativeFunction.ERF,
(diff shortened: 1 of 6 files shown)


src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -773,10 +773,10 @@ def __init__(
...
             "ceil": nodes.NativeFunction.CEIL,
             "trunc": nodes.NativeFunction.TRUNC,
-            "i32": nodes.NativeFunction.I32,
-            "i64": nodes.NativeFunction.I64,
-            "f32": nodes.NativeFunction.F32,
-            "f64": nodes.NativeFunction.F64,
+            "int32": nodes.NativeFunction.I32,
+            "int64": nodes.NativeFunction.I64,
+            "float32": nodes.NativeFunction.F32,
+            "float64": nodes.NativeFunction.F64,
```
romanc: Should those be `NativeFunction.INT32` and friends?

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -1034,9 +1034,11 @@ def visit_Constant(
                     value_type = np.dtype(f"i{int(self.literal_precision / 8)}")
                 elif isinstance(value, float):
                     value_type = np.dtype(f"f{int(self.literal_precision / 8)}")
+                elif hasattr(value, "dtype") and isinstance(value.dtype, np.dtype):
+                    value_type = value.dtype
```
romanc, line 1043: This is for Chris' remapping problem, right?  I think it would be good to include a test case for this.

romanc (commented):
I think we should keep 

https://github.com/GridTools/gt4py/blob/98f728065e20825ece5f3a8f295626d5d9c06093/src/gt4py/cartesian/frontend/nodes.py#L10-L59

up to date and add the new native functions (data types should be covered).

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -759,7 +773,32 @@ def __init__(
...
             "trunc": nodes.NativeFunction.TRUNC,
-        }
+            "int32": nodes.NativeFunction.I32,
+            "int64": nodes.NativeFunction.I64,
+            "float32": nodes.NativeFunction.F32,
+            "float64": nodes.NativeFunction.F64,
+            "int": (
+                nodes.NativeFunction.I32
+                if self.literal_precision == 32
+                else nodes.NativeFunction.I64
```
havogt: I would either introduce 2 variables `literal_float_precision` and `literal_int_precision` or just drop support for switching int precision with the flag. I don't see that switching both together is a useful operation.

romanc: For the validation of the GEOS port, we need a way to control both integer and float precision. So let's introduce two configuration variables.

romanc:
Addressed all reviewer comments, i.e.

- changed `literal_precision` to `literal_int_precision` and `literal_float_precision`
- full propagated numpy-standard names
- added the missing test case for numpy-typed constants

PR #2189, opened 2025-07-28
fix[next][dace]: Correct dace setting to use default cuda stream

edopao:
The setting 'max_concurrent_streams=-1' configures dace to only use the default stream. Previously, we were setting the value `1`, but this just selects a random cuda stream.


src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -55,13 +55,14 @@ def set_dace_config(
         value="-std=c++17 -fPIC -O3 -march=native -Wno-unused-parameter",
     )
 
+    # Setting 'max_concurrent_streams=-1' configures dace to only use the default stream.
     # In some stencils, mostly in `apply_diffusion_to_w` the cuda codegen messes
```
philip-paul-mueller:
```suggestion
    #  In some stencils, mostly in `apply_diffusion_to_w` the cuda codegen messes
```

src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -55,13 +55,14 @@ def set_dace_config(
...
     #  up with the cuda streams, i.e. it allocates N streams but uses N+1.
-    #  This is a workaround until this issue if fixed in DaCe.
+    #  Using only one stream is a workaround until this issue if fixed in DaCe.
     # However, when using asynchronous SDFG call (see `backend.make_dace_backend()`),
```
philip-paul-mueller:
```suggestion
    #  However, when using asynchronous SDFG call (see `backend.make_dace_backend()`),
```

philip-paul-mueller (approved):
LGTM.

Only formatting stuff.
However, you should also add the difference between `1` and `-1`:
- Use default stream as opposed to just one (random) stream that is different for any stencil (but I think we solved that)?
- Different code paths in DaCe as every synchronization is skipped.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_translation.py
```diff
@@ -98,34 +98,17 @@ def test_find_constant_symbols(has_unit_stride):
         assert len(constant_symbols) == 0
 
 
-def _is_present_async_sdfg_init_code(sdfg: dace.SDFG) -> bool:
```
philip-paul-mueller: Why have you deleted that?

edopao: I've deleted this check because I removed the emission of this additional code in the init function, see lines 122-124 in `translation.py`. The unit test is testing only the translation stage (lowering to SDFG and SDFG transformations), while the setting `max_concurrent_streams=-1` takes effect during dace codegen, in the compilation stage.

edopao: I've changed my mind, I prefer to keep this functionality.

philip-paul-mueller:
@edopao I would actually add the description you wrote about into the code.
In the end it is the code you look at first and when other commits (for whatever reason) modifies the line, it is no longer trivial, as in using `git blame`, to figuring out which PR introduced that behaviour.

What I also think is important and should be documented, is that if you use the _default stream_ then codegen will use [different codepaths](https://github.com/spcl/dace/blob/f5dd7c8e77f91269a7dab3a1c08b827559a22bf0/dace/codegen/targets/cuda.py#L1400).

PR #2191, opened 2025-07-28
fix[next][dace]: Do not crash if SDFG report not found

edopao:
This PR handles the case where the SDFG are precompiled without performance metrics `(config.COLLECT_METRICS_LEVEL == metrics.DISABLED)` but `config.COLLECT_METRICS_LEVEL` is set on another run, when the stencils are loaded from the binary library. The main baseline was crashing because the SDFG would return `None` as profiling report.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
@@ -9,6 +9,7 @@
 from __future__ import annotations
 
 import functools
+import warnings
 from typing import Any, Sequence
 
 import dace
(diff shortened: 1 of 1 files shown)


philip-paul-mueller (commented):
I agree that crashing is not good, but I am also not sure that the new behaviour is better.
In my view the situation Give-Me-The-Job-Report-But-The-SDFG-was-Compiled-without-It more like an error to me, thus I would say that issuing a simple warning and continue is not the right action.
Since the user especially requested the report and at the end of the run the report is not there.
From the users perspective this is an error!
Thus, I would say that the thing to do is to exit with a proper error.

philip-paul-mueller (approved):
LGTM

PR #2192, opened 2025-07-29
build: new-style license, bump setuptools

romanc:
## Description

Use new-style license format as suggested by the official python packaging user guide [1]. To do so, we need to bump the minimal version of setuptools.

[1] https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files

## Requirements

  N/A
  N/A

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,11 +2,11 @@
 
 [build-system]
 build-backend = 'setuptools.build_meta'
-requires = ['cython>=3.0.0', 'setuptools>=70.0.0', 'versioningit>=3.1.1', 'wheel>=0.33.6']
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -108,7 +107,7 @@ dependencies = [
   'numpy>=1.26.4',
   'packaging>=20.0',
   'pybind11>=2.10.1',
-  'setuptools>=70.0.0',
+  'setuptools>=77.0.3',
```
havogt, line 110: this bump is probably not necessary, but I don't see problems

havogt (approved):
thanks!

PR #2193, opened 2025-07-29
build[next]: Update dace version

edopao:
The main change introduced by dace main is the removal of `parent` argument from `add_nested_sdfg()`, see https://github.com/spcl/dace/pull/2109


philip-paul-mueller (approved):
LGTM.

PR #2198, opened 2025-07-30
feat[cartesian]: Support for `round()` / ` round_away_from_zero()` in gtscript

romanc:
# Description

This PR adds support `round()` and `round_away_from_zero()` in `gtscript`. This PR got a bit of traction because there are basically two predominant ways to round ties (e.g. when rounding 1.5 to an integer it's as close to 1.0 as it is to 2.0). The default in python, Kotlin, Julia, and C# is to split ties by rounding to the nearest even value, e.g.

- `round(-0.5) == 0.0`
- `round(0.5) == 0.0`
- `round(1.5) == 2.0`
- `round(2.5) == 2.0`

This is also what the IEEE754 floating point standard defines. In contrast, programming languages like Go, Rust, Ruby, C++, and Fortran default to round away from zero, e.g.

- `round(-0.5) == -1.0`
- `round(0.5) == 0.0`
- `round(1.5) == 2.0`
- `round(2.5) == 3.0`

From a DSL point of view, it is paramount to implement the same behavior in all backends, regardless of the programming language of the backend. We choose to implement `round()` with the tie breaking as suggested by the standard. In addition, we implement a function `round_away_from_zero()` which breaks ties by rounding away from zero. This allows, for example, to truthfully port FORTRAN-based code. The full discussion and evaluation of other alternatives can be found in the ADR.

## Requirements

### DIFF
--- a/docs/development/ADRs/cartesian/frontend-round-functions.md
+++ b/docs/development/ADRs/cartesian/frontend-round-functions.md
@@ -1,36 +1,37 @@
-# Behavior of the `round()` function in GT4Py
+# Round functions in gtscript
 
-TODO: To be written after we came to a conclusion in the discussions.
+In the context of adding support for a `round()` function in `gtscript`, facing divergent implementations in backend languages, we decided to implement `round()` with tie-breaking to even and `round_away_from_zero()` with tie-breaking away from zero to achieve consistent results while leaving the choice of behavior up to users. We considered multiple alternatives (see below), which all have their advantages and disadvantages.
+
+Why-statement: In the context of [use case/user story], facing [concern], we decided to [do thing X] to achieve [system qualities/desired consequences]. We considered [thing Y] and accept [downsides / undesired consequences].
 
 ## Context
 
-The default behavior of the `round()` function in python is to split ties by rounding to the nearest even value, e.g.
+There are basically two predominant ways to round ties (e.g. when rounding 1.5 to an integer it's as close to 1.0 as it is to 2.0). The default in python, Kotlin, Julia, and C# is to split ties by rounding to the nearest even value, e.g.
 
 - `round(-0.5) == 0.0`
 - `round(0.5) == 0.0`
 - `round(1.5) == 2.0`
 - `round(2.5) == 2.0`
 
-This is in contrast to C++ and Fortran where the default is to split ties by rounding away from zero, e.g.
+This is also what the IEEE754 floating point standard defines. In contrast, programming languages like Go, Rust, Ruby, C++, and Fortran default to round away from zero, e.g.
 
 - `round(-0.5) == -1.0`
 - `round(0.5) == 0.0`
 - `round(1.5) == 2.0`
 - `round(2.5) == 3.0`
 
-From a DSL point of view, it is paramount to implement a consistent rounding behavior in all backends, especially in the context of chaotic systems such as NWPs.
+From a DSL point of view, it is paramount to implement the same behavior in all backends, regardless of the programming language of the backend.
 
 ## Decision
 
-TODO: To be written after we came to a conclusion in the discussions.
+We choose to implement `round()` with the tie breaking as suggested by the standard. In addition, we implement the function `round_away_from_zero()` which breaks ties by rounding away from zero. This allows, for example, to truthfully port FORTRAN-based code.
 
 ## Consequences
 
 As a consequence
 
 - all backends use the same rounding behavior
-
-TODO: The above is a given (we all agree that all backends should show the same behavior). Other consequences depend on the decision we take.
+- users can choose which rounding behavior suits their needs
 
 ## Alternatives considered
 
@@ -50,7 +51,7 @@ One argument against this option is that is doesn't align with the behavior fo t
 
 From a technical point of view, this option can be implemented with standard library calls in the `gt:*` and `dace:*` backends. The `numpy` and `debug` backends will need a custom implementation of the `round()` function.
 
-### Allow users to configure the behavior of the `round()` function
+### Allow users to configure the behavior of the `round()` function with an argument
 
 One argument in favor of this option is that users can choose: Given that there are good reasons for python and C++ behavior of the `round()` function, we let our users choose what works best for their use-case. By choosing a default, GT4Py can still nudge its users towards one option or another.
 
@@ -102,3 +103,5 @@ ROUND_MODE_DEFAULT: RoundingMode = _get_default_rounding_mode()
 ```
 
 That way, users of GT4Py can adjust the rounding mode to their use-cases, both globally and on a case-by-case basis.
+
+One technical argument against this alternative is the relatively high maintenance overhead compared to having two separate functions.
--- a/docs/development/ADRs/cartesian/frontend-round-functions.md
+++ b/docs/development/ADRs/cartesian/frontend-round-functions.md
@@ -2,7 +2,6 @@
 
 In the context of adding support for a `round()` function in `gtscript`, facing divergent implementations in backend languages, we decided to implement `round()` with tie-breaking to even and `round_away_from_zero()` with tie-breaking away from zero to achieve consistent results while leaving the choice of behavior up to users. We considered multiple alternatives (see below), which all have their advantages and disadvantages.
 
-Why-statement: In the context of [use case/user story], facing [concern], we decided to [do thing X] to achieve [system qualities/desired consequences]. We considered [thing Y] and accept [downsides / undesired consequences].
 
 ## Context
 
--- a/docs/development/ADRs/cartesian/frontend-round-functions.md
+++ b/docs/development/ADRs/cartesian/frontend-round-functions.md
@@ -2,7 +2,6 @@
 
 In the context of adding support for a `round()` function in `gtscript`, facing divergent implementations in backend languages, we decided to implement `round()` with tie-breaking to even and `round_away_from_zero()` with tie-breaking away from zero to achieve consistent results while leaving the choice of behavior up to users. We considered multiple alternatives (see below), which all have their advantages and disadvantages.
 
-
 ## Context
 
 There are basically two predominant ways to round ties (e.g. when rounding 1.5 to an integer it's as close to 1.0 as it is to 2.0). The default in python, Kotlin, Julia, and C# is to split ties by rounding to the nearest even value, e.g.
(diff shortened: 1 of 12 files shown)


docs/development/ADRs/cartesian/frontend-rounding-behavior.md
```diff
@@ -0,0 +1,39 @@
...
+
+- all backends use the same rounding behavior,
+- the `numpy` and `debug` backend get a performance overhead,
+- the performance backends (`gt:*` and `dace:*`) are not impacted.
+
+## Alternatives considered
+
+### Adopt banker's rounding in C++ backends
+
+We usually align with python/numpy standards (link to literal precision ADR) in the frontend following the argument of least surprises. In this case, we argue that the performance of performance backends is too important to adopt banker's rounding in C++ powered backends.
```
romanc: Link to be fixed if https://github.com/GridTools/gt4py/pull/2187 merges first or to be amended with https://github.com/GridTools/gt4py/pull/2187 in case this one merges first.

src/gt4py/cartesian/gtc/gtcpp/gtcpp_codegen.py
```diff
@@ -176,6 +176,9 @@ def visit_NativeFunction(self, func: NativeFunction, **kwargs: Any) -> str:
                 NativeFunction.FLOOR: "std::floor",
                 NativeFunction.CEIL: "std::ceil",
                 NativeFunction.TRUNC: "std::trunc",
+                NativeFunction.ROUND: "std::round",
```
havogt: What about 
```suggestion
                NativeFunction.ROUND: "std::nearbyint",
```
I only did a quick research, but as far as I understand it should give the python/numpy rounding?

havogt: Or are you saying this is slower than std::round?

romanc: Thanks for the pointers. I must have been searching with the wrong keywords. I didn't want to say `std::nearbyint()` is slower than `std::round()`. I meant to say that if we had to come up with a custom `round()` function in C++ to adopt banker's rounding, then this would be slower than a native function call provided by the language.

I agree, `std::nearbyint()` and `std::rint()` will result in the python/numpy behavior of the `round()` function, _if the current rounding mode is set to `FE_TONEAREST`_. So yes, we could adopt python's behavior of the  `round()` function without performance overhead. We'd just need to ensure that the rounding mode is set correctly (even if `FE_TONEAREST` is the default).

Given that option, I think there's still an argument to make for keeping the C++ behavior of the `round()` function: Currently, I expect the majority of (non-AI) NWP codes to be written in either Fortran or C++. Let's say we want as many of these models to be ported to GT4Py with as little friction as we can arrange. The argument of "least surprise" will thus work in favor of the of adopting the C++ behavior in GT4Py.

Porting an NWP model from Fortran/C++ to GT4Py could also involve some kind of end-to-end tests, running the ported model with input/output data serialized from the old model. Adopting python's default for the `round()` function would complicate such tests (which are - in our experience - already hard, but manageable to control).

Would you argue for adopting the python behavior in all backends?

romanc: We could also build a zero cost abstraction layer, letting users decide which behavior they prefer. `round() ` gets an optional second argument and at compile time, we dispatch to either `std::round` or `std::nearbyint`

havogt: I think in general the idea of GT4Py is to avoid inheriting legacy, therefore I would be in favor of the Python behavior. I would like to hear @egparedes opinion who is currently on vacation.

> We'd just need to ensure that the rounding mode is set correctly (even if FE_TONEAREST is the default).

I personally wouldn't go that far.

> We could also build a zero cost abstraction layer, letting users decide which behavior they prefer. round() gets an optional second argument and at compile time, we dispatch to either std::round or std::nearbyint

Since I have never seen a `round` in the wild, I don't know if we need it. Feels like a very niche feature that could be added on demand.

romanc: Okay, then let me make a separate PR for `erf` and `erfc` since it seems we can all agree to merge those two.

> > We could also build a zero cost abstraction layer, letting users decide which behavior they prefer. round() gets an optional second argument and at compile time, we dispatch to either std::round or std::nearbyint
> 
> Since I have never seen a `round` in the wild, I don't know if we need it. Feels like a very niche feature that could be added on demand.

We are adding the `round` function because we encountered one in the wild (while porting GEOS to NDSL, which includes GT4Py). Since we are validating our GEOS port against data serialized from the original model, we'd like to have the option to replicate the FORTRAN numerics. As long as this is configurable, we can make a decision at the NDSL level and I'm happy to adopt the python behavior of `round()` as the default behavior in GT4Py. @FlorianDeconinck is currently also on vacation. He also might have an opinion on the matter, so let's wait for his opinion too.

> I think in general the idea of GT4Py is to avoid inheriting legacy, therefore I would be in favor of the Python behavior.

In general, I agree and e.g. in case of the [mixed precision PR](https://github.com/GridTools/gt4py/pull/2187), I am grateful for your suggestion to use numpy-standard `int32` instead of `i32`. In this case, I'd argue that the behavior of a `round()` function in case of ties is arbitrary and I wouldn't call the choice of the C++ standard "legacy". I would also like to point out that GT4Py is supposed to be a DSL for the domain of weather and climate models. It is my understanding that GT4Py adapts some common idioms from that domain, e.g. the horizontal loops are denoted with `I` and `J`, while the vertical loop is denoted by `K`. This naming choice is arbitrary, and chosen such that it fits the domain. Inside the domain of atmospheric science, my understanding is that users of GT4Py are surprised if both 1.5 and 2.5 round to 2.0. And while I see the benefits of doing so (from a computer science / numerical point of view after spending a day on the topic), I myself still find it to be surprising.

havogt: Disclaimer: Not an expert in the topic either.

Re legacy and just for bikeshedding: it seems rounding in both directions is better for statistics which is probably also what you want in weather and climate. I assume weather and climate is using the Fortran behavior because it's using Fortran, aka legacy. Does Fortran (<= 90) have an equivalent to `nearbyint`? C is not famous for breaking old behavior to get rid of legacy. I didn't check, but most likely some libraries used the `round` behavior before it was standardized.

No problem for me if you want to add a switch for the rounding mode.

romanc:
Now that everybody is back from vacation, I'd like to take up the discussion here again. Since there are multiple options, I did a quick survey of what other languages do. Here's the result

| default `round()` behavior | languages |
| --- | --- |
| Away from zero | [Go](https://yourbasic.org/golang/round-float-to-int/) / [Rust](https://github.com/rust-lang/rust/issues/96710) / [C++](https://en.cppreference.com/w/cpp/numeric/math/round.html) / [Ruby](https://apidock.com/ruby/Float/round) |
| To positive inf | [Java](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#round-double-) / [Js](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round#description) |
| To even | [python](https://docs.python.org/3/library/functions.html#round) / [Kotlin](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.math/round.html) / [Julia](https://www.jlhub.com/julia/manual/en/function/round) / [C#](https://learn.microsoft.com/en-us/dotnet/api/system.math.round?view=net-9.0) |

Many languages support setting the rounding mode and many of the above links will point out that "away from zero" is "what people commonly learn at school" while "to even" is the standardized version with better mathematical properties (and thus used e.g. in financial applications).

My suggestion is as follows: I would allow users to specify how the want to split ties for floats between two integers. I'd suggest `away-from-zero` as the default because I think this is most appropriate for the domain in the sense of least surprises.

FlorianDeconinck:
+1 from `away-from-zero` as the default, with `to-even` introduced as well.

Do we want
- `round(x, to_even=True)`
- OR, `round(x)` and `round_to_even(x)`

?

src/gt4py/cartesian/gtc/ufuncs.py
```diff
@@ -29,6 +28,27 @@
     erfc_.types = ["f->f", "d->d", "F->F", "D->D"]
 
 
+def _round_away_from_zero(num):
+    """Computes the nearest integer value to num, rounding halfway cases away from zero.
```
FlorianDeconinck: Can we move this doc to the `gtscript.py` which is more "user" facing - and cross reference the other round (and vice-versa)

romanc: sure - will do!

docs/development/ADRs/cartesian/frontend-round-functions.md
```diff
@@ -0,0 +1,107 @@
+# Round functions in gtscript
+
+In the context of adding support for a `round()` function in `gtscript`, facing divergent implementations in backend languages, we decided to implement `round()` with tie-breaking to even and `round_away_from_zero()` with tie-breaking away from zero to achieve consistent results while leaving the choice of behavior up to users. We considered multiple alternatives (see below), which all have their advantages and disadvantages.
+
+Why-statement: In the context of [use case/user story], facing [concern], we decided to [do thing X] to achieve [system qualities/desired consequences]. We considered [thing Y] and accept [downsides / undesired consequences].
```
egparedes: Shouldn't this paragraph from the template be removed?
```suggestion
```

docs/development/ADRs/cartesian/frontend-round-functions.md
```diff
@@ -0,0 +1,107 @@
...
+The default rounding mode will be centrally defined in `src/gt4py/cartesian/definitions.py` (value subject to discussions) and can be changed with the environment variable `GT4PY_ROUND_MODE_DEFAULT` (let me know if you find a better name).
+
+```py
+@enum.unique
+class RoundingMode(enum.Enum):
+    ROUND_AWAY_FROM_ZERO = enum.auto()
+    ROUND_TO_EVEN = enum.auto()
+
+def _get_default_rounding_mode() -> RoundingMode:
+    # TODO: fill in the default once the discussion settles
```
egparedes: Open TODO
```suggestion
    # TODO: fill in the default once the discussion settles
```

egparedes (commented):
I did a very quick review and it looks good to me, other than a couple of small forgotten issues in the ADR.

romanc:
cscs-ci run

PR #2201, opened 2025-08-04
fix[dace][next]: "Fixed" Asynchronous SDFG

philip-paul-mueller:
This PR is related to an earlier change, introduced by [PR#2189](https://github.com/GridTools/gt4py/pull/2189), which changed the configuration such that instead of one (random) stream the default stream of the GPU was used.
An unintended side effect was that asynchronous execution, i.e. returning from `CompiledSDFG.fast_call()` once all work has been _scheduled_, i.e. not wait until the work has been completed, which is an [intended](https://github.com/GridTools/gt4py/pull/2035) behaviour, is no longer an opt-in feature but has become an "always on" feature.

This PR re-enables "synchronous" execution, by adding a final state to the SDFG that contains a Tasklet which calls the synchronization primitive of the GPU runtime.
This state is added by default, if GPU execution is enabled, to preserve the previous behaviour.

Note that the changes this PR introduces are potentially dangerous as they might lead to race conditions.
This happens if the result of a GPU Map is, in any way, not used by a Map.
For example, if a GPU Map computes data, which is then copied back to CPU, where it is used, either in a Tasklet or on an interstate edge.
Because the async `mempcy()` function is used and there is no synchronization, the copy might not have concluded before the CPU data is used.
However, the expected impacts are minimal, as in the GT4Py setting this should not happen as scalar computation are usually done on inputs of Maps and we do not use interstate edges.

Due to the changes described above some of the DaCe related aspects of the workflow have been changed.
This in turn lead to the discovery of several bugs, in GT4Py and DaCe, with further consequences.
The most severe one is [DaCe issue#2125](https://github.com/spcl/dace/issues/2125).
The issue is that `dace.Config` is a global object that is shared among all threads (which is an issue for the tests).


src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -60,79 +61,91 @@ def find_constant_symbols(
...
 
-def make_sdfg_async(sdfg: dace.SDFG) -> None:
-    """Make an SDFG call immediatly return, without waiting for execution to complete.
-    This allows to run the SDFG asynchronously, thus letting GPU kernel execution
-    to overlap with host python code.
+def make_sdfg_call_async(sdfg: dace.SDFG, gpu: bool) -> None:
+    """Configure an SDFG to immediately return once all work has been scheduled.
 
-    The asynchronous call is implemented by the following changes to SDFG:
+    This means that `CompiledSDFG.fast_call()` will return only after all computations
```
havogt: What does that mean? It sounds like the opposite of what the function name suggests.

havogt: After seeing the same sentence in `make_sdfg_call_sync` I guess it's a copy-paste error.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -172,54 +195,56 @@ def generate_sdfg(
...
             # We measure the execution time of one program by instrumenting the
             #   top-level SDFG with a cpp timer (std::chrono). This timer measures
             #   only the computation time, it does not include the overhead of
             #   calling the SDFG from Python.
             sdfg.instrument = dace.dtypes.InstrumentationType.Timer
-        elif on_gpu and self.async_sdfg_call:
-            # Do not use async SDFG call when collecting metrics: we use SDFG instrumentatiom
-            #   and therefore the SDFG execution has to complete before we can retrieve the report.
-            make_sdfg_async(sdfg)
+        elif self.async_sdfg_call:
```
havogt, line 254: `self.async_sdfg_call` is set in the backend right?
Maybe the logic here would be slightly cleaner if we overwrite it in case `config.COLLECT_METRICS_LEVEL != metrics.DISABLED`.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
philip-paul-mueller: You are right.
Copy past is indeed the root of all evil.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
philip-paul-mueller: You mean something like this:
```python
async_sdfg_call = True
if config.COLLECT_METRICS_LEVEL != metrics.DISABLED:
    sdfg.instrument = ...
    async_sdfg_call = False
```

I have nothing against it, but I would like to raise the following points:
- I have the impression that `self.async_sdfg_call` would become useless because "asynchronous mode" would be determined by the "metric level"; which might be a good idea since timings are useless without it.
- If you do it, you can not run the thing in "synchronous mode" without enabling metric. I do not see a use case for it.

Therefore, I would propose to leave it as it is and let Edoardo, once he is back, decide what to do with it.

havogt: No, I meant the user-specifies the behavior, but `config.COLLECT_METRICS_LEVEL != metrics.DISABLED` overrides it (possibly with warning). But fine for now.

PR #2204, opened 2025-08-05
fix[cartesian]: fix setuptools for parallel CUDA compilation

SF-N:
This adds a global lock around the entire block that temporarily modifies `setuptools.distutils.core._setup_stop_after`, calls `setuptools.setup()`, updates `cmdclass`, and restores the original value. Since `_setup_stop_after` is a shared global variable controlling setup flow, concurrent access caused race conditions and errors during parallel stencil compilation in PMAP-LES. 
This fix ensures these related operations execute atomically, preventing race conditions and making stencil compilation thread-safe during parallel builds also on GPUs.

### DIFF
--- a/src/gt4py/cartesian/backend/pyext_builder.py
+++ b/src/gt4py/cartesian/backend/pyext_builder.py
@@ -11,6 +11,7 @@ import copy
 import io
 import os
 import shutil
+import threading
 from typing import Any, Dict, List, Literal, Optional, Tuple, Type, TypedDict, Union
 
 import pybind11
@@ -20,6 +21,9 @@ from setuptools.command.build_ext import build_ext
 
 from gt4py._core import definitions as core_defs
 from gt4py.cartesian import config as gt_config
+
+
+_setuptools_lock = threading.Lock()
 
 
 class SetuptoolsArgs(TypedDict):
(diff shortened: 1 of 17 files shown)


src/gt4py/cartesian/backend/pyext_builder.py
```diff
@@ -22,6 +23,9 @@
 from gt4py.cartesian import config as gt_config
 
 
+_setuptools_lock = threading.Lock()
```
tehrengruber:
```suggestion
_SETUPTOOLS_LOCK = threading.Lock()
```

tehrengruber (approved):
@havogt Any objections against merging this? This fixes parallel compilation in PMAP-LES.

SF-N:
Stencil compilation is parallelised using a global `ThreadPoolExecutor`, with each stencil submitted to a background thread at decoration time via `ThreadPoolExecutor.submit()`. Compilation tasks are tracked in a global set and remove themselves upon completion. When the set is empty, the thread pool shuts down automatically. The main thread only waits for compilation by calling `future.result()` when the stencil is used for the first time.

I am not sure if there is an easy way to include a testcase in gt4py, but we are testing this in the PMAP-LES CI, so we will notice regressions when updating the gt4py version.

PMAP-LES is not yet open source.

romanc (approved):
Thanks for the fix and the details answers to my questions. All good from my side.

PR #2209, opened 2025-08-07
prototype: multiple output domains

havogt:
a hacky vertical slice for multiple output domains

(DaCe lowering is missing)


havogt:
replaced by #2225

PR #1847, opened 2025-02-05
refactor[next]: Global ordering relation of dimensions

SF-N:
This PR introduces a global ordering relation of dimensions, replacing the previous mechanism in promote_dims. The ordering relation is: sorting first by `Dimension.kind` (`HORIZONTAL` < `VERTICAL` < `LOCAL`) and then lexicographically by `Dimension.value`.

Reason: 
An as_fieldop call as emitted by the frontend has no domain, and inferring the type of the domain was not possible, since no global ordering relation of dimensions existed, e.g. for an as_fieldop operating on a `Vertex` and a `K` field it was unclear if the dimensions of the output were `(Vertex, K)` or (`K, Vertex)`, which lead to many negative consequences in other places, that will be tackled in [PR 1853](https://github.com/GridTools/gt4py/pull/1853) and following ones.

PR #2217, opened 2025-08-13
Do not merge: Follow New Canonical Field Order

philip-paul-mueller:
In [PR#1847](github.com/GridTools/gt4py/pull/1847) the convention was established that the local dimension is always at the end.
However, in [PR#1922](https://github.com/GridTools/gt4py/pull/1847) this convention was changed such that the local dimension is between the horizontal and vertical.
However, this convention change was not propagated to the DaCe backend, which continued to work, since the case, where a field has a horizontal, vertical and a local was never hit.

This PR introduces this new canonical order into the DaCe backend.


philip-paul-mueller:
Depends on the [new DaCe version](https://github.com/GridTools/gt4py/pull/2218).

philip-paul-mueller:
I think we need unit test were the order is a bit different of the dimensions.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -2150,6 +2150,9 @@ def test_gtir_index():
 
     sdfg = build_dace_sdfg(testee, CARTESIAN_OFFSETS)
 
+    # TODO(phimuell, edopao): I, phimuell, am not sure if this is correct, since you add
+    #   `i` essentially to `i`. So I would expect that the result is `2 * i` or something,
+    #   maybe a `+1` because of the shift.
```
edopao: Good catch! I have opened a separate PR, because unrelated from the main issue: #2234

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -42,7 +42,11 @@ def decorated_program(
             size_args = arguments.iter_size_args(args)
             args = (*args, *size_args)
 
-        if not fun.sdfg_program._lastargs:
+        # TODO(phimuell, edopao): Currently the `fast_call` API is not working. The main reason
```
edopao: I have re-enabled SDFG fast-call and the tests are passing. The real issue was probably the scope of the binding function, which you fixed in `compilation.py`. Good catch!

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -261,12 +266,16 @@ def {_cb_get_stride}(ndarray, dim_index):
             arg3=_cb_last_call_args,
         )
     )
-    code.indent()
+
+    # TODO(phimuell, edopao): I think that it is more safe to iterate through the argument
+    #   list of the SDFG. As it would ensure that everything is updated. I would change the
+    #   iteration such that we iterate through all arguments the SDFG is needing and then
+    #   look where we get it from.
```
edopao: I tried to change the logic to loop through the SDFG arguments list but it is more complicated, because for each GTIR argument (e.g. field "a") we need to set N arguments in the SDFG (the array pointer, the array shape symbols and the strides in all dimensions), which are found at different locations in the SDFG arguments list. For some of these arguments, under certain conditions (e.g. the connectivity arrays, or regular fields with `make_persistent=True`), we just emit some assert code.

Note that the binding function is only used with fast-call, to update the SDFG arguments list, that is from the second time the SDFG is called. On the first time we use the regular SDFG call, which validates that all data containers and free symbols are passed. Besides, the SDFG argument list is generated from `program_source`, so we can _safely_ rely on looping through the GTIR parameters list. I will replace the TODO comment with this explanation.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -43,9 +43,18 @@ def _get_sdfg_ctype_arglist_callback(
...
     assert spec is not None
     module = importlib.util.module_from_spec(spec)
-    exec(python_code, module.__dict__)
-    assert bind_func_name in module.__dict__
-    return module.__dict__[bind_func_name]
+
+    # By copying the global scope we make sure that the function `bind_func_name`
+    #   is not put into the `module.__dict__` scope.
+    #   This can be an issue if two threads in parallel try to generate the name.
+    #   Essentially the function is now an object that only lives inside `scope`.
```
edopao, line 50: Issue addressed in #2235, good catch!

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -128,42 +139,89 @@ def _create_scan_field_operator_impl(
...
-
-    # allocate local temporary storage
+        # TODO(phimuell, edopao): I think this case is not properly handled.
+        #   Something that was not done is for example the update of `field_origin`.
+        #   Furthermore, if we have a list, then `scan_dim_index` would technically
+        #   be more than one index.
+        raise NotImplementedError("The case for List output in scans is not supported.")
+
+    # Create the final data storage, that is outside of the surrounding Map.
+    # TODO(phimuell, edopao): In a previous version the stride was adapted. However,
```
edopao, line 152: You are right, I have applied your solution in a separate PR: https://github.com/GridTools/gt4py/pull/2241

edopao:
Most changes has been implemented as smaller PRs and have been merged to main:
- test[next][dace]: fix test expectation #2234 
- bug[next][dace]: fix scope for SDFG binding function #2235 
- fix[next][dace]: update strides of scan output in SDFG lowering #2241 
- fix[next][dace]: follow new canonical field domain order #2240

This PR contains several suggestions for code refactoring. I will close the PR because outdated with respect to main, but we can still refer to it.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
edopao: Three weeks later, I realize that you were right about the update of connectivity args in SDFG fast-call. I have create a gt4py issue and I am working on it: #2266

PR #2225, opened 2025-08-22
feat[next]: Multiple output domains

SF-N:
This PR builds on [PR 2209](https://github.com/GridTools/gt4py/pull/2209).

### DIFF
--- a/src/gt4py/next/embedded/operators.py
+++ b/src/gt4py/next/embedded/operators.py
@@ -108,8 +108,18 @@ def field_operator_call(op: EmbeddedOperator[_R, _P], args: Any, kwargs: Any) ->
 
         domain = kwargs.pop("domain", None)
 
-        out_domain = common.domain(domain) if domain is not None else _get_out_domain(out)
+        if domain is not None:
+            if isinstance(out, tuple) and not isinstance(domain, tuple):
+                out_domain = tuple([domain] * len(out))
+            else:
+                out_domain = domain
+        else:
+            if isinstance(out, tuple):
+                out_domain = tuple(_get_out_domain(o) for o in out)
+            else:
+                out_domain = _get_out_domain(out)
 
+        # TODO?
         new_context_kwargs["closure_column_range"] = _get_vertical_range(out_domain)
 
         with embedded_context.update(**new_context_kwargs):
@@ -128,26 +138,32 @@ def field_operator_call(op: EmbeddedOperator[_R, _P], args: Any, kwargs: Any) ->
         return op(*args, **kwargs)
 
 
-def _get_vertical_range(domain: common.Domain) -> common.NamedRange | eve.NothingType:
-    vertical_dim_filtered = [nr for nr in domain if nr.dim.kind == common.DimensionKind.VERTICAL]
-    assert len(vertical_dim_filtered) <= 1
-    return vertical_dim_filtered[0] if vertical_dim_filtered else eve.NOTHING
+def _get_vertical_range(
+    domain: common.Domain | tuple[common.Domain, ...],
+) -> common.NamedRange | eve.NothingType | tuple[common.NamedRange | eve.NothingType, ...]:
+    if isinstance(domain, tuple):
+        return tuple(_get_vertical_range(sub) for sub in domain)
+    else:
+        vertical_dim_filtered = [nr for nr in domain if nr.dim.kind == common.DimensionKind.VERTICAL]
+        assert len(vertical_dim_filtered) <= 1
+        return vertical_dim_filtered[0] if vertical_dim_filtered else eve.NOTHING
 
 
 def _tuple_assign_field(
     target: tuple[common.MutableField | tuple, ...] | common.MutableField,
     source: tuple[common.Field | tuple, ...] | common.Field,
-    domain: common.Domain,
+    domain: common.DomainLike | tuple[common.DomainLike | tuple, ...],
 ) -> None:
     @utils.tree_map
-    def impl(target: common.MutableField, source: common.Field) -> None:
+    def impl(target: common.MutableField, source: common.Field, domain: common.DomainLike) -> None:
+        domain = common.domain(domain)
         if isinstance(source, common.Field):
             target[domain] = source[domain]
         else:
             assert core_defs.is_scalar_type(source)
             target[domain] = source
 
-    impl(target, source)
+    impl(target, source, domain)
 
 
 def _intersect_scan_args(
--- a/src/gt4py/next/ffront/past_passes/type_deduction.py
+++ b/src/gt4py/next/ffront/past_passes/type_deduction.py
@@ -8,7 +8,7 @@
 from typing import Any, Optional, cast
 
 from gt4py.eve import NodeTranslator, traits
-from gt4py.next import errors
+from gt4py.next import common, errors
 from gt4py.next.ffront import (
     dialect_ast_enums,
     program_ast as past,
@@ -57,28 +57,35 @@ def _validate_operator_call(new_func: past.Name, new_kwargs: dict) -> None:
         _ensure_no_sliced_field(new_kwargs["out"])
 
         domain_kwarg = new_kwargs["domain"]
-        if not isinstance(domain_kwarg, past.Dict):
+        if not isinstance(domain_kwarg, (past.Dict, past.TupleExpr)):
             raise ValueError(f"Only Dictionaries allowed in 'domain', got '{type(domain_kwarg)}'.")
+        if isinstance(domain_kwarg, past.Dict):
+            domain_kwarg_ = [domain_kwarg]
+        elif isinstance(domain_kwarg, past.TupleExpr):
+            out_kwarg = new_kwargs["out"]
+            assert isinstance(out_kwarg, past.TupleExpr)
+            assert len(out_kwarg.elts) == len(domain_kwarg.elts)
+            domain_kwarg_ = domain_kwarg.elts
+        for dom in domain_kwarg_:
+            if len(dom.values_) == 0 and len(dom.keys_) == 0:
+                raise ValueError("Empty domain not allowed.")
 
-        if len(domain_kwarg.values_) == 0 and len(domain_kwarg.keys_) == 0:
-            raise ValueError("Empty domain not allowed.")
-
-        for dim in domain_kwarg.keys_:
-            if not isinstance(dim.type, ts.DimensionType):
-                raise ValueError(
-                    f"Only 'Dimension' allowed in domain dictionary keys, got '{dim}' which is of type '{dim.type}'."
-                )
-        for domain_values in domain_kwarg.values_:
-            if len(domain_values.elts) != 2:
-                raise ValueError(
-                    f"Only 2 values allowed in domain range, got {len(domain_values.elts)}."
-                )
-            if not _is_integral_scalar(domain_values.elts[0]) or not _is_integral_scalar(
-                domain_values.elts[1]
-            ):
-                raise ValueError(
-                    f"Only integer values allowed in domain range, got '{domain_values.elts[0].type}' and '{domain_values.elts[1].type}'."
-                )
+            for dim in dom.keys_:
+                if not isinstance(dim.type, ts.DimensionType):
+                    raise ValueError(
+                        f"Only 'Dimension' allowed in domain dictionary keys, got '{dim}' which is of type '{dim.type}'."
+                    )
+            for domain_values in dom.values_:
+                if len(domain_values.elts) != 2:
+                    raise ValueError(
+                        f"Only 2 values allowed in domain range, got {len(domain_values.elts)}."
+                    )
+                if not _is_integral_scalar(domain_values.elts[0]) or not _is_integral_scalar(
+                    domain_values.elts[1]
+                ):
+                    raise ValueError(
+                        f"Only integer values allowed in domain range, got '{domain_values.elts[0].type}' and '{domain_values.elts[1].type}'."
+                    )
 
 
 class ProgramTypeDeduction(traits.VisitorWithSymbolTableTrait, NodeTranslator):
@@ -133,9 +140,16 @@ class ProgramTypeDeduction(traits.VisitorWithSymbolTableTrait, NodeTranslator):
 
     def visit_TupleExpr(self, node: past.TupleExpr, **kwargs: Any) -> past.TupleExpr:
         elts = self.visit(node.elts, **kwargs)
-        return past.TupleExpr(
-            elts=elts, type=ts.TupleType(types=[el.type for el in elts]), location=node.location
-        )
+        ttype: ts.TupleType
+        if any(isinstance(elt, past.Dict) for elt in node.elts):
+            assert all(isinstance(elt, past.Dict) for elt in node.elts)
+            # TODO: add check that Dict is DomainLike
+            ttype = ts.TupleType(
+                types=[ts.DomainType(dims=[common.Dimension(elt.keys_[0].id)]) for elt in elts]
+            )
+        else:
+            ttype = ts.TupleType(types=[elt.type for elt in elts])
+        return past.TupleExpr(elts=elts, type=ttype, location=node.location)
 
     def _deduce_binop_type(
         self, node: past.BinOp, *, left: past.Expr, right: past.Expr, **kwargs: Any
--- a/src/gt4py/next/ffront/past_to_itir.py
+++ b/src/gt4py/next/ffront/past_to_itir.py
@@ -160,15 +160,17 @@ def _range_arg_from_field(field_name: str, dim: int) -> str:
     return f"__{field_name}_{dim}_range"
 
 
-def _flatten_tuple_expr(node: past.Expr) -> list[past.Name | past.Subscript]:
-    if isinstance(node, (past.Name, past.Subscript)):
+def _flatten_tuple_expr(node: past.Expr) -> list[past.Name | past.Subscript | past.Dict]:
+    if isinstance(node, (past.Name, past.Subscript, past.Dict)):
         return [node]
     elif isinstance(node, past.TupleExpr):
         result = []
         for e in node.elts:
             result.extend(_flatten_tuple_expr(e))
         return result
-    raise ValueError("Only 'past.Name', 'past.Subscript' or 'past.TupleExpr' thereof are allowed.")
+    raise ValueError(
+        f"Only 'past.Name', 'past.Subscript' or 'past.TupleExpr' thereof are allowed, got '{type(node)}'."
+    )
 
 
 @dataclasses.dataclass
@@ -346,7 +348,7 @@ class ProgramLowering(
 
     def _construct_itir_domain_arg(
         self,
-        out_field: past.Name,
+        out_field: past.Name | past.Subscript | past.Dict,
         node_domain: Optional[past.Expr],
         slices: Optional[list[past.Slice]] = None,
     ) -> itir.FunCall:
@@ -471,30 +473,37 @@ class ProgramLowering(
             flattened = _flatten_tuple_expr(out_arg)
 
             first_field = flattened[0]
-            assert all(
-                self.visit(field.type).dims == self.visit(first_field.type).dims
-                for field in flattened
-            ), "Incompatible fields in tuple: all fields must have the same dimensions."
 
             field_slice = None
             if isinstance(first_field, past.Subscript):
+                raise AssertionError  # TODO support slicing of multiple fields with different domain
                 assert all(isinstance(field, past.Subscript) for field in flattened), (
                     "Incompatible field in tuple: either all fields or no field must be sliced."
                 )
                 assert all(
                     concepts.eq_nonlocated(
                         first_field.slice_,
-                        field.slice_,  # type: ignore[union-attr] # mypy cannot deduce type
+                        field.slice_,
                     )
                     for field in flattened
                 ), "Incompatible field in tuple: all fields must be sliced in the same way."
                 field_slice = self._compute_field_slice(first_field)
                 first_field = first_field.value
 
-            return (
-                self._construct_itir_out_arg(out_arg),
-                self._construct_itir_domain_arg(first_field, domain_arg, field_slice),
-            )
+            if isinstance(domain_arg, past.TupleExpr):
+                domain_args = [
+                    self._construct_itir_domain_arg(field, domain, None)
+                    for field, domain in zip(
+                        flattened, _flatten_tuple_expr(domain_arg), strict=True
+                    )
+                ]
+                domain_expr = im.make_tuple(*domain_args)
+                return self._construct_itir_out_arg(out_arg), domain_expr
+            else:
+                return (
+                    self._construct_itir_out_arg(out_arg),
+                    self._construct_itir_domain_arg(first_field, domain_arg, field_slice),
+                )
         else:
             raise AssertionError(
                 "Unexpected 'out' argument. Must be a 'past.Subscript', 'past.Name' or 'past.TupleExpr' node."
--- a/src/gt4py/next/iterator/embedded.py
+++ b/src/gt4py/next/iterator/embedded.py
@@ -1674,8 +1674,12 @@ def _validate_domain(domain: Domain, offset_provider_type: common.OffsetProvider
 
 
 @runtime.set_at.register(EMBEDDED)
-def set_at(expr: common.Field, domain: common.DomainLike, target: common.MutableField) -> None:
-    operators._tuple_assign_field(target, expr, common.domain(domain))
+def set_at(
+    expr: common.Field,
+    domain: common.DomainLike | tuple[common.DomainLike | tuple, ...],
+    target: common.MutableField,
+) -> None:
+    operators._tuple_assign_field(target, expr, domain)
 
 
 @runtime.if_stmt.register(EMBEDDED)
--- a/src/gt4py/next/iterator/ir_utils/misc.py
+++ b/src/gt4py/next/iterator/ir_utils/misc.py
@@ -227,8 +227,16 @@ def grid_type_from_domain(domain: itir.FunCall) -> common.GridType:
         return common.GridType.UNSTRUCTURED
 
 
+def _flatten_tuple_expr(domain_expr: itir.Expr) -> tuple[itir.Expr]:
+    if cpm.is_call_to(domain_expr, "make_tuple"):
+        return sum((_flatten_tuple_expr(arg) for arg in domain_expr.args), start=())
+    else:
+        return (domain_expr,)
+
+
 def grid_type_from_program(program: itir.Program) -> common.GridType:
-    domains = program.walk_values().if_isinstance(itir.SetAt).getattr("domain").to_set()
+    domain_exprs = program.walk_values().if_isinstance(itir.SetAt).getattr("domain").to_set()
+    domains = sum((_flatten_tuple_expr(domain_expr) for domain_expr in domain_exprs), start=())
     grid_types = {grid_type_from_domain(d) for d in domains}
     if len(grid_types) != 1:
         raise ValueError(
--- a/src/gt4py/next/iterator/transforms/infer_domain.py
+++ b/src/gt4py/next/iterator/transforms/infer_domain.py
@@ -519,6 +519,13 @@ def infer_expr(
     return expr, accessed_domains
 
 
+def _make_symbolic_domain_tuple(domains: itir.Node) -> DomainAccess:
+    if cpm.is_call_to(domains, "make_tuple"):
+        return tuple(_make_symbolic_domain_tuple(arg) for arg in domains.args)
+    else:
+        return SymbolicDomain.from_expr(domains)
+
+
 def _infer_stmt(
     stmt: itir.Stmt,
     **kwargs: Unpack[InferenceOptions],
@@ -528,9 +535,9 @@ def _infer_stmt(
         # between the domain stored in IR and in the annex
         domain = constant_folding.ConstantFolding.apply(stmt.domain)
 
-        transformed_call, _ = infer_expr(
-            stmt.expr, domain_utils.SymbolicDomain.from_expr(domain), **kwargs
-        )
+        symbolic_domain = _make_symbolic_domain_tuple(domain)
+
+        transformed_call, _ = infer_expr(stmt.expr, symbolic_domain, **kwargs)
 
         return itir.SetAt(
             expr=transformed_call,


src/gt4py/next/ffront/decorator.py
```diff
@@ -626,6 +625,18 @@ def program_inner(definition: types.FunctionType) -> Program:
 OperatorNodeT = TypeVar("OperatorNodeT", bound=foast.LocatedNode)
 
 
+def _slice_outs(
+    outs: common.Field | tuple[common.Field | tuple, ...],
+    domains: common.Domain | tuple[common.Domain | tuple, ...],
```
havogt:
```suggestion
    domains: common.DomainLike | tuple[common.DomainLike | tuple, ...],
```

src/gt4py/next/embedded/operators.py
```diff
@@ -128,26 +128,38 @@ def field_operator_call(op: EmbeddedOperator[_R, _P], args: Any, kwargs: Any) ->
         return op(*args, **kwargs)
 
 
-def _get_vertical_range(domain: common.Domain) -> common.NamedRange | eve.NothingType:
-    vertical_dim_filtered = [nr for nr in domain if nr.dim.kind == common.DimensionKind.VERTICAL]
-    assert len(vertical_dim_filtered) <= 1
-    return vertical_dim_filtered[0] if vertical_dim_filtered else eve.NOTHING
+def _get_vertical_range(
+    domain: common.Domain | tuple[common.Domain, ...],
```
havogt:
```suggestion
    domain: xtyping.MaybeNestedInTuple[common.DomainLike],
```

src/gt4py/next/embedded/operators.py
```diff
@@ -108,7 +108,7 @@ def field_operator_call(op: EmbeddedOperator[_R, _P], args: Any, kwargs: Any) ->
 
         domain = kwargs.pop("domain", None)
 
-        out_domain = common.domain(domain) if domain is not None else _get_out_domain(out)
+        out_domain = domain if domain is not None else _get_out_domain(out)
```
havogt: Alternatively (instead of doing the `common.domain` in many places now), we could implement a function that does `MaybeNestedInTuple[DomainLike]` -> `MaybeNestedInTuple[Domain]`. Now many of the utility functions need to deal with `DomainLike`s. So far `DomainLike` was something that the user can pass, but internally, we only use `Domain`, i.e. the place where to do the transformation was only at the user interface.

src/gt4py/next/embedded/operators.py
havogt: or, if we decide to do domainlike->domain promotion outside.
```suggestion
    domain: xtyping.MaybeNestedInTuple[common.Domain],
```

src/gt4py/next/ffront/decorator.py
havogt:
```suggestion
    outs: xtyping.MaybeNestedInTuple[common.Field],
    domains: xtyping.MaybeNestedInTuple[common.Domain],
```

havogt: (or move the domainlike -> domain transformation outside)

tests/next_tests/integration_tests/feature_tests/ffront_tests/ffront_test_utils.py
```diff
@@ -220,6 +224,7 @@ def offset_provider_type(self) -> common.OffsetProviderType: ...
 def simple_mesh(allocator) -> MeshDescriptor:
     num_vertices = 9
     num_cells = 8
+    num_levels = 10
```
havogt: I wouldn't add the k levels to the mesh, why not just leave it in the cases.unstructured?

SF-N: Yes, there is no benefit, so I reverted the changes.

edopao: I think we can remove `num_levels` from `MeshDescriptor` protocol, it is not used anyway.

SF-N: Yes, I removed it.

src/gt4py/next/embedded/operators.py
SF-N: Yes, that makes sense, I tried to refactor it a bit.

src/gt4py/next/common.py
```diff
@@ -592,6 +593,16 @@ def __getstate__(self) -> dict[str, Any]:
 )  # `Domain` is `Sequence[NamedRange]` and therefore a subset
 
 
+def normalize_domains(domain_like: MaybeNestedInTuple[DomainLike]) -> MaybeNestedInTuple[Domain]:
```
havogt: absorb in `domain` (and remove `if isinstance(domain_like, Sequence)` case)?

tehrengruber: After implementing this [here](https://github.com/GridTools/gt4py/pull/2319) I think this is not a good idea anymore. We need to distinguish between places where only a single domain is allowed, e.g. `allocate`, and places where a tuple of domain likes, e.g. all places where this `normalize_domains` is used. It feels simpler to just do `tree_map(domain)(nested_domain_like)` in all places where `normalize_domains` is used and then to remove this function. Then it is clear that in all other places a single domain is allowed and here multiple.

tehrengruber (approved):
I made some cleanups in `past_to_itir` using two new arguments added to `tree_map`. Ready from my side as soon as we have decided and resolved on how to handle `normalize_domains`,

src/gt4py/next/utils.py
```diff
@@ -92,6 +107,8 @@ def tree_map(
     *,
     collection_type: type | tuple[type, ...] = tuple,
     result_collection_constructor: Optional[type | Callable] = None,
+    unpack: bool = False,
```
havogt, line 116: I don't like the unpack. Do we have a use-case where the `unpack=False` makes sense?

src/gt4py/next/common.py
havogt: sounds good

edopao:
The dace lowering now works, also tested in ICON4Py CI. Tomorrow I will run a performance regression test on ICON4Py blueline.

@philip-paul-mueller Could you please review only the changes in the dace backend? Thanks

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -57,6 +60,26 @@ def get_field_domain(domain: domain_utils.SymbolicDomain) -> FieldopDomain:
     ]
 
 
+TargetDomain: TypeAlias = MaybeNestedInTuple[domain_utils.SymbolicDomain]
```
philip-paul-mueller, line 63: I think we should use this type also in other locations.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -683,7 +683,7 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
 
     def visit_SetAt(
         self, stmt: gtir.SetAt, sdfg: dace.SDFG, state: dace.SDFGState
-    ) -> dace.SDFGState:
+    ) -> dace.SDFGState | None:
```
philip-paul-mueller: Please update the description what the return value `None` means.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
```diff
@@ -57,6 +60,26 @@ def get_field_domain(domain: domain_utils.SymbolicDomain) -> FieldopDomain:
...
+"""Symbolic domain which defines the range to write in the target field.
+
+For tuple output, the corresponding domain in fieldview is a tuple of domains.
+"""
+
+
+class TargetDomainParser(eve.visitors.NodeTranslator):
```
philip-paul-mueller: Is there a specific reason why this is a class and not a free function?
To be more clearer I think this class should be a nested class inside a free function such that you can write `parse_target_domain(node)` instead of the more verbose `TargetDomainParser().apply(node)`.
However, since it gets a GTIR expression, I would propose to change its name to something like "extract_target_domain()`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -694,8 +694,7 @@ def visit_SetAt(
         """
 
         # Visit the domain expression.
-        assert isinstance(stmt.domain.type, ts.DomainType)
-        domain = domain_utils.SymbolicDomain.from_expr(stmt.domain)
+        domain = gtir_domain.TargetDomainParser().apply(stmt.domain)
```
philip-paul-mueller: See my comment from above about `TargetDomainParser`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -754,11 +753,24 @@ def _visit_target(
                     ),
                 )
 
-        gtx_utils.tree_map(
-            lambda source, target, _domain=domain, _target_state=target_state: _visit_target(
-                source, target, _domain, _target_state
-            )
-        )(source_tree, target_tree)
+        if isinstance(target_tree, tuple) and not isinstance(domain, tuple):
```
philip-paul-mueller: I do not like this and the comment is indicating that it is only needed to support some tests so I am not sure about the relevance, i.e. I missing a todo "remove once iterator IR is gone" or something.

I am also not sure if it would be better in this case to modify `domain` such that it has the same tree structure as `target_tree` but all values are the original `domain` value?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -846,35 +858,41 @@ def visit_Lambda(
...
-            for param, arg in args.items()
-            if not isinstance(arg, gtir_to_sdfg_types.SymbolicData)
-        }
         symbolic_args = {
-            param: arg
+            str(param.id): arg
             for param, arg in args.items()
             if isinstance(arg, gtir_to_sdfg_types.SymbolicData)
         }
+        data_args: dict[str, gtir_to_sdfg_types.FieldopResult] = {
```
philip-paul-mueller: Why do you now filtering out the `symbolic_args`?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -846,35 +858,41 @@ def visit_Lambda(
...
 
         # inherit symbols from parent scope but eventually override with local symbols
         lambda_symbols = {
             sym: self.global_symbols[sym]
             for sym in symbol_ref_utils.collect_symbol_refs(node.expr, self.global_symbols.keys())
-        } | {
-            param: gtir_to_sdfg_types.get_tuple_type(arg) if isinstance(arg, tuple) else arg.gt_type
-            for param, arg in args.items()
-        }
+        } | {str(param.id): param.type for param, arg in args.items() if arg is not None}
```
philip-paul-mueller, line 868: Just a question, when can `arg` be `None`?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -893,11 +911,21 @@ def visit_Lambda(
...
+                if arg_node is None:
+                    # This argument has empty domain, which means that it should not be
+                    # used inside the nested SDFG, and does not need to be connected outside.
+                    assert all(
+                        node.data != nsdfg_dataname
+                        for node in lambda_ctx.sdfg.all_nodes_recursive()
+                        if isinstance(node, dace.nodes.AccessNode)
+                    )
+                    lambda_ctx.sdfg.arrays[nsdfg_dataname].transient = True
+                    continue
```
```suggestion
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -893,11 +911,21 @@ def visit_Lambda(
...
             if nsdfg_datadesc.transient:
                 continue
-
-            if nsdfg_dataname in lambda_arg_nodes:
-                src_node = lambda_arg_nodes[nsdfg_dataname].dc_node
-                dataname = src_node.data
-                datadesc = src_node.desc(ctx.sdfg)
+            elif nsdfg_dataname in lambda_arg_nodes:
+                arg_node = lambda_arg_nodes[nsdfg_dataname]
+                if arg_node is None:
```
philip-paul-mueller: Why has this case appeared I do not understand it.
I would have expected that stuff you do not need has been filtered out.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -893,11 +911,21 @@ def visit_Lambda(
...
+                    # This argument has empty domain, which means that it should not be
+                    # used inside the nested SDFG, and does not need to be connected outside.
+                    assert all(
+                        node.data != nsdfg_dataname
+                        for node in lambda_ctx.sdfg.all_nodes_recursive()
+                        if isinstance(node, dace.nodes.AccessNode)
+                    )
+                    lambda_ctx.sdfg.arrays[nsdfg_dataname].transient = True
```
philip-paul-mueller: I guess you use here the same trick as with the transients.

The assert has the problem that it will also decent into nested SDFG, if one of them happen to have a data descriptor with `nsdfg_dataname` since we do not have unique names (something I would like to have), this assert might be triggered.

The canonical way of doing it using `sdfg.remove_data(nsdfg_dataname, validate=__debug__)`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -939,7 +967,8 @@ def visit_Lambda(
         nsdfg_symbols_mapping = {}
         for sym in lambda_ctx.sdfg.free_symbols:
             if (sym_id := str(sym)) in lambda_arg_nodes:
-                assert isinstance(lambda_arg_nodes[sym_id].gt_type, ts.ScalarType)
+                arg_node = lambda_arg_nodes[sym_id]
+                assert arg_node and isinstance(arg_node.gt_type, ts.ScalarType)
```
philip-paul-mueller: Again I am not a fan of allowing `None` here.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
```diff
@@ -234,8 +231,8 @@ def _create_field_operator(
         # handle tuples of fields
         output_symbol_tree = gtir_to_sdfg_utils.make_symbol_tree("x", node_type)
         return gtx_utils.tree_map(
-            lambda output_edge, output_sym: _create_field_operator_impl(
-                ctx, sdfg_builder, domain, output_edge, output_sym.type, map_exit
+            lambda edge, sym, ctx_=ctx: _create_field_operator_impl(
+                ctx_, sdfg_builder, domain, edge, sym.type, map_exit
```
philip-paul-mueller: I do not fully get why you have this lambda trick with `ctx` now.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -88,11 +90,11 @@ def _parse_fieldop_arg_impl(
...
     ctx: gtir_to_sdfg.SubgraphContext,
     sdfg_builder: gtir_to_sdfg.SDFGBuilder,
-    field_domain: gtir_domain.FieldopDomain,
-    output_edge: gtir_dataflow.DataflowOutputEdge,
+    output_edge: gtir_dataflow.DataflowOutputEdge | None,
+    output_domain: infer_domain.NonTupleDomainAccess,
     output_type: ts.FieldType,
     map_exit: dace.nodes.MapExit,
-) -> gtir_to_sdfg_types.FieldopData:
+) -> gtir_to_sdfg_types.FieldopData | None:
```
philip-paul-mueller, line 96: This function is allowed to return `None` while `_create_field_operator_impl()` is not, thus there is no explanation what `None` means!

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -201,8 +210,8 @@ def _create_scan_field_operator(
     node_type: ts.FieldType | ts.TupleType,
     sdfg_builder: gtir_to_sdfg.SDFGBuilder,
     input_edges: Iterable[gtir_dataflow.DataflowInputEdge],
-    output_tree: gtir_dataflow.DataflowOutputEdge
-    | tuple[gtir_dataflow.DataflowOutputEdge | tuple[Any, ...], ...],
+    output: MaybeNestedInTuple[gtir_dataflow.DataflowOutputEdge | None],
```
philip-paul-mueller, line 221: What does `None` means here.
As far as I can see, you have not modified `_create_field_operator()`, to which you refer to to, and that did not accept `None`.
Please update the description.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -246,27 +255,28 @@ def _create_scan_field_operator(
...
+        assert isinstance(output_domain, domain_utils.SymbolicDomain)
         return _create_scan_field_operator_impl(
-            ctx, sdfg_builder, field_domain, output_tree, node_type, map_exit
+            ctx, sdfg_builder, output, output_domain, node_type, map_exit
         )
     else:
-        # handle tuples of fields
-        # the symbol name 'x' in the call below is not used, we only need
-        # the tree structure of the `TupleType` definition to pass to `tree_map()`
+        # Handle tuples of fields. note that the symbol name 'x' in the call below
```
philip-paul-mueller:
```suggestion
        # Handle tuples of fields. Note that the symbol name 'x' in the call below
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -246,27 +255,28 @@ def _create_scan_field_operator(
...
+        # is not used, we only need the tree structure of the `TupleType` definition
+        # to pass to `tree_map()` in order to retrieve the type of each field.
         output_symbol_tree = gtir_to_sdfg_utils.make_symbol_tree("x", node_type)
```
philip-paul-mueller: You should not use `x` here as it might result in a clash.
I would propose something like `__gtir_guaranteed_unused_dummy_variable`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -558,6 +569,15 @@ def _connect_nested_sdfg_output_to_temporaries(
     return gtir_dataflow.DataflowOutputEdge(outer_ctx.state, output_expr)
 
 
+def _remove_nested_sdfg_connector(
+    inner_ctx: gtir_to_sdfg.SubgraphContext,
+    nsdfg_node: dace.nodes.NestedSDFG,
+    inner_data: gtir_to_sdfg_types.FieldopData,
+) -> None:
+    inner_data.dc_node.desc(inner_ctx.sdfg).transient = True
+    nsdfg_node.out_connectors.pop(inner_data.dc_node.data)
```
philip-paul-mueller: What if it is an input?
I would also propose to remove the data, see a comment at the top.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -677,24 +690,37 @@ def translate_scan(
...
+    input_edges = []
     for input_connector, outer_arg in lambda_arg_nodes.items():
-        arg_desc = outer_arg.dc_node.desc(ctx.sdfg)
-        input_subset = dace_subsets.Range.from_array(arg_desc)
-        input_edge = gtir_dataflow.MemletInputEdge(
-            ctx.state, outer_arg.dc_node, input_subset, nsdfg_node, input_connector
-        )
-        lambda_input_edges.append(input_edge)
+        assert not lambda_ctx.sdfg.arrays[input_connector].transient
+        if outer_arg is None:
```
philip-paul-mueller: Again I do not like `None`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -677,24 +690,37 @@ def translate_scan(
...
-    lambda_output_tree = gtx_utils.tree_map(
-        lambda lambda_output_data: _connect_nested_sdfg_output_to_temporaries(
-            lambda_ctx, ctx, nsdfg_node, lambda_output_data
+    output_tree = gtx_utils.tree_map(
+        lambda output_data, output_domain: _connect_nested_sdfg_output_to_temporaries(
+            lambda_ctx, ctx, nsdfg_node, output_data
         )
-    )(lambda_output)
+        if output_domain != infer_domain.DomainAccessDescriptor.NEVER
+        else _remove_nested_sdfg_connector(lambda_ctx, nsdfg_node, output_data)
```
philip-paul-mueller: I can barely read that, you should do the selection outside and then pass a single function to `tree_map()`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_types.py
```diff
@@ -178,7 +180,7 @@ def get_symbol_mapping(
         return symbol_mapping
 
 
-FieldopResult: TypeAlias = FieldopData | tuple[FieldopData | tuple, ...]
+FieldopResult: TypeAlias = MaybeNestedInTuple[FieldopData | None]
```
philip-paul-mueller, line 183: What does `None` mean?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_types.py
```diff
@@ -192,28 +194,19 @@ class SymbolicData:
...
+        assert isinstance(sym.type, ts.TupleType)
+        tuple_symbols = gtir_to_sdfg_utils.flatten_tuple_fields(sym.id, sym.type)
         tuple_data_fields = gtx_utils.flatten_nested_tuple(arg)
         return [
             (str(tsym.id), tfield)
             for tsym, tfield in zip(tuple_symbols, tuple_data_fields, strict=True)
         ]
     else:
-        return [(name, arg)]
+        return [(sym.id, arg)]
```
philip-paul-mueller:
```suggestion
        return [(str(sym.id), arg)]
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_utils.py
```diff
@@ -52,6 +52,8 @@ def get_arg_symbol_mapping(
         A mapping from inner symbol names to values or symbolic definitions
         in the parent SDFG.
     """
+    if arg is None:
```
philip-paul-mueller: What does `None` mean?

philip-paul-mueller (commented):
My main point of concern that a lot of checks `is None` where introduced, which I do not like at all.
They should have been filtered out before, this is what I would do/suggest.
If this is not possible, then some explanation, in doc strings, are missing what this mean.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
edopao: Good idea.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: This was a  mistake in rebasing.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: The iterator IR will remain as long as the GTFN backend exists. 
When I started implementing the lowering to SDFG, I always assumed that it had to be compatible with mixed fieldview/iterator IR, because I did not exclude that we would run some transformations on the GTIR, before lowering to SDFG. Therefore, the current lowering supports patterns that we do not hit, for real programs, in the current workflow.
A big `TODO` would be to rewrite the lowering and make it simpler, and only support what is needed in fieldview-GTIR.
Anyway, I think I can start the simplification here: I will remove this branch and instead I will skip some iterator tests.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Before I was filtering `if not isinstance(arg, gtir_to_sdfg_types.SymbolicData)`, now `param.id not in symbolic_args`, which is the same.
Maybe it was more clear before, I can revert the change.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: `arg is None` when the domain inferred by gt4py is empty. This can happen, for example, if a field argument is not accessed. I have added a comment.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: For regular fields, yes, the arguments with empty domain have been filtered out. However, in case of nested tuples, we have not filtered out nested fields with empty domain.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Good catch about the name aliasing in nested SDFGs. However, I cannot call `sdfg.remove_data` here because I am looping on the sdfg data. I will add a cleanup step.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Fixed

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: I have changed the code so that I can avoid the `continue` here.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
edopao: Not needed, indeed, if we drop the support for tuple iterator.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: Good idea.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_utils.py
edopao: I removed this function, instead.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_types.py
edopao: Fixed

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: Done.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_types.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: This function is only called on the output (sink) data nodes inside the nested SDFG.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: Good idea.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: It was not needed here, actually. However, I also see some room for improvement of this code.

src/gt4py/next/program_processors/runners/dace/gtir_domain.py
edopao: Done

edopao (commented):
Thanks @philip-paul-mueller for the review. I will now push an updated version of the SDFG lowering, which addresses your comments. However, I will pull part of the changes into a separate PR, which can be merge separately, before this PR.

edopao:
@philip-paul-mueller The dace part is ready for another round of review.

@havogt Yesterday I ran a performance regression test on icon4py blueline and I see no issues, see plot.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
```diff
@@ -601,7 +601,7 @@ def translate_tuple_get(
     if isinstance(data_nodes, gtir_to_sdfg_types.FieldopData):
         raise ValueError(f"Invalid tuple expression {node}")
     unused_arg_nodes: Iterable[gtir_to_sdfg_types.FieldopData] = gtx_utils.flatten_nested_tuple(
-        tuple(arg for i, arg in enumerate(data_nodes) if i != index)
+        tuple(arg for i, arg in enumerate(data_nodes) if arg is not None and i != index)
```
philip-paul-mueller: Maybe an explanation here would be helpful.
I do not know why but it looks a bit strange to me, probably nothing.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -628,13 +675,18 @@ def translate_scan(
     lambda_args = [sdfg_builder.visit(arg, ctx=ctx) for arg in node.args]
     lambda_args_mapping = [
         (im.sym(_scan_input_name(scan_carry), scan_carry_type), init_data),
-    ] + [(param, arg) for param, arg in zip(stencil_expr.params[1:], lambda_args, strict=True)]
+    ] + [
+        (param, arg)
+        for param, arg in zip(stencil_expr.params[1:], lambda_args, strict=True)
```
philip-paul-mueller: `param` is a name so I would name it something like `pname`.
You see "argument" and "parameter" are kind of interchangeable so I would propose to have a big difference between them.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_types.py
```diff
@@ -180,8 +180,14 @@ def get_symbol_mapping(
         return symbol_mapping
 
 
-FieldopResult: TypeAlias = MaybeNestedInTuple[FieldopData]
-"""Result of a field operator, can be either a field or a tuple fields."""
+FieldopResult: TypeAlias = MaybeNestedInTuple[FieldopData | None]
+"""Result of a field operator, can be either a field or a tuple fields.
+
+For tuple of fields, any of the nested fields can be None, in case it is not udes
```
philip-paul-mueller:
```suggestion
For tuple of fields, any of the nested fields can be None, in case it is not used
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -754,10 +753,10 @@ def _visit_target(
                 )
 
         gtx_utils.tree_map(
-            lambda source, target, domain_=domain, target_state_=target_state: _visit_target(
+            lambda source, target, domain_, target_state_=target_state: _visit_target(
                 source, target, domain_, target_state_
```
```suggestion
            lambda source, target, domain, target_state_=target_state: _visit_target(
                source, target, domain, target_state_
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -754,10 +753,10 @@ def _visit_target(
                 )
 
         gtx_utils.tree_map(
-            lambda source, target, domain_=domain, target_state_=target_state: _visit_target(
+            lambda source, target, domain_, target_state_=target_state: _visit_target(
```
philip-paul-mueller: Another question I have here why do you have to copy `target_state` here into an argument and can not simply capture it?

philip-paul-mueller (commented):
I think the lowering is kind of okay, some refactoring might be needed.
Although I still do not like the `None`s but the explanation makes it clear what is going on.
I have some smallish suggestions.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Right, I can capture it. I think it was not possible, at some point, in the past. I have fixed it here and in other places.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: `param` is a `gtir.Sym` symbol, in this version of the code. I will rename it to `gt_symbol`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
edopao: Please check my attempt to make it more clear.

PR #2229, opened 2025-08-28
feat[next]: Introduce typing module, expose more types, cleanup deprecated stuff

havogt:
Introduce `gt4py.next.typing` for types that are useful for type annotation in user-code, but should not be used at runtime. It's exported as `Annotated` to disallow runtime usage for now, but the cleaner mechanism could be a Protocol in the future.

Additionally:
- expose some runtime types from `_core.definitions` in `gt4py.next`.
- add `is_field_allocation_tool_for` which was a missing convenience function.
- remove deprecated functions and types in `iterator.embedded`

### DIFF
--- a/tests/next_tests/integration_tests/feature_tests/ffront_tests/test_scalar_if.py
+++ b/tests/next_tests/integration_tests/feature_tests/ffront_tests/test_scalar_if.py
@@ -6,35 +6,19 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
-from functools import reduce
-
-import numpy as np
 import pytest
 
-import gt4py.next as gtx
-from gt4py.next import Field, errors, field_operator, float64, index_field
+from gt4py.next import Field, errors, field_operator, float64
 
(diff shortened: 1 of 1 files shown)


src/gt4py/core.py
```diff
@@ -0,0 +1,18 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+from gt4py._core.definitions import CUPY_DEVICE_TYPE, Device, DeviceType, Scalar, is_scalar_type
+
+
+__all__ = [
+    "CUPY_DEVICE_TYPE",
+    "Device",
+    "DeviceType",
+    "Scalar",
```
havogt: `Scalar` would be purely for typing, should we create a `typing` module? It would be a bit confusing with other `typing` modules as they I believe are pure typing, while here I would export the real thing. Other example is `decorator.Program` which should not be used by the user at runtime, but is needed for typing.

egparedes: We could export aliases to the real thing purely for typing, for example using `Annotated`:
```
Program: TypeAlias = Annotated[decorator.Program, "Only for typing"]
```

PR #2235, opened 2025-09-02
bug[next][dace]: fix scope for SDFG binding function

edopao:
Issue found by @philip-paul-mueller in #2217, which could be encountered in parallel test execution. See code comment in `compilation.py`.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
@@ -44,13 +44,11 @@ def _get_sdfg_ctype_arglist_callback(
     assert spec is not None
     module = importlib.util.module_from_spec(spec)
 
-    # By copying the global scope we make sure that the function `bind_func_name`
-    #   is not put into the `module.__dict__` scope.
-    #   This can be an issue if two threads in parallel try to generate the name.
-    #   Essentially the function is now an object that only lives inside `scope`.
-    scope = module.__dict__.copy()
-    assert bind_func_name not in scope
-    exec(python_code, scope)
+    # We use a different dict for the locals namespace, in order to avoid that
+    #   the function `bind_func_name` is added to the current scope.
+    #   This can be an issue if two threads in parallel generate the same name.
+    exec(python_code, scope := {})  # type: ignore[var-annotated]
+    assert bind_func_name not in module.__dict__
 
     assert bind_func_name in scope
     assert bind_func_name not in module.__dict__
--- a/src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
@@ -43,9 +43,18 @@ def _get_sdfg_ctype_arglist_callback(
     spec = importlib.util.spec_from_loader(module_name, loader=None)
     assert spec is not None
     module = importlib.util.module_from_spec(spec)
-    exec(python_code, module.__dict__)
-    assert bind_func_name in module.__dict__
-    return module.__dict__[bind_func_name]
+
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -43,9 +43,18 @@ def _get_sdfg_ctype_arglist_callback(
...
-    assert bind_func_name in module.__dict__
-    return module.__dict__[bind_func_name]
+
+    # By copying the global scope we make sure that the function `bind_func_name`
+    #   is not put into the `module.__dict__` scope.
+    #   This can be an issue if two threads in parallel try to generate the name.
+    #   Essentially the function is now an object that only lives inside `scope`.
+    scope = module.__dict__.copy()
+    assert bind_func_name not in scope
+    exec(python_code, scope)
```
egparedes: Wouldn't it be simpler to use a different dict for the locals namespace?

```suggestion
    exec(python_code, module, scope:={})
    assert bind_func_name not in module.__dict__
```

edopao: Yes, that makes sense. I have changed accordingly.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -43,9 +43,16 @@ def _get_sdfg_ctype_arglist_callback(
...
-    return module.__dict__[bind_func_name]
+
+    # We use a different dict for the locals namespace, in order to avoid that
+    #   the function `bind_func_name` is added to the current scope.
+    #   This can be an issue if two threads in parallel generate the same name.
+    exec(python_code, scope := {})  # type: ignore[var-annotated]
+    assert bind_func_name not in module.__dict__
+
+    assert bind_func_name in scope
+    assert bind_func_name not in module.__dict__
```
egparedes, line 54: Besides removing the repeated assertion, I'm now a bit confused about what is `module` used for here, if the module dict is not passed as the global namespace for `exec()`. If my understanding is not wrong, I can see only two different scenarios:

- `module` is indeed not needed at all since the code does not reference any global definition from the module, so it's maybe only passed to set the `__module__` attribute of the function properly: in this case I would not load the module at all and will just set `scope[find_func_name].__module__ = module_name` before returning the function
- `module` is needed but we don't want to modify it: I would pass `module.__dict__` as the `globals` argument of `exec()` and an empty dict for the locals, as I proposed earlier. If this doesn't work because of the limitations of the `exec()` implementation when passing both globals and locals, I then would revert to passing only a globals namespace with a shallow copy of the `module.__dict__`
    ```python
    exec(python_code, scope := {**module.__dict__}))
```

egparedes: Besides removing the repeated assertion, I'm now a bit confused about what is `module` used for here, if the module dict is not passed as the global namespace for `exec()`. If my understanding is not wrong, I can see only two different scenarios:

- `module` is indeed not needed at all since the code does not reference any global definition from the module, so it's maybe only passed to set the `__module__` attribute of the function properly: in this case I would not load the module at all and will just set `scope[find_func_name].__module__ = module_name` before returning the function
- `module` is needed but we don't want to modify it: I would pass `module.__dict__` as the `globals` argument of `exec()` and an empty dict for the locals, as I proposed earlier. If this doesn't work because of the limitations of the `exec()` implementation when passing both globals and locals, I then would revert to passing only a globals namespace with a shallow copy of the `module.__dict__`
    ```python
    exec(python_code, scope := {**module.__dict__}))
    ```

egparedes (approved):
LGTM

PR #2236, opened 2025-09-02
bug[next]: fix exponential runtime in gtfn codegen

havogt:


tehrengruber (approved):
Certainly a good improvement, but that this fixes the performance problem feels more like a side effect then a proper solution. Wouldn't it make sense to invest a couple of minutes to remove the transformation of children whose values are ignored anyway in `TemplatedGenerator.generic_visit`, e.g. something like this + a comment why we skip:

```
    def transform_children(self, node: Node, **kwargs: Any) -> Dict[str, Any]:
        return {key: self.visit(value, **kwargs) for key, value in node.iter_children_items() if key not in kwargs}  # type: ignore[misc]

    def transform_annexed_items(self, node: Node, **kwargs: Any) -> Dict[str, Any]:
        return {key: self.visit(value, **kwargs) for key, value in node.annex.items() if key not in kwargs}
```

havogt:
> Certainly a good improvement, but that this fixes the performance problem feels more like a side effect then a proper solution. Wouldn't it make sense to invest a couple of minutes to remove the transformation of children whose values are ignored anyway in `TemplatedGenerator.generic_visit`, e.g. something like this + a comment why we skip:
> 
> ```
>     def transform_children(self, node: Node, **kwargs: Any) -> Dict[str, Any]:
>         return {key: self.visit(value, **kwargs) for key, value in node.iter_children_items() if key not in kwargs}  # type: ignore[misc]
> 
>     def transform_annexed_items(self, node: Node, **kwargs: Any) -> Dict[str, Any]:
>         return {key: self.visit(value, **kwargs) for key, value in node.annex.items() if key not in kwargs}
> ```

Enrique wanted to look into improving the TemplatedGenerator. I don't understand how your proposal would fix the issue.

PR #2239, opened 2025-09-03
ci: switch to macos-latest in daily-CI

edopao:
This PR addresses a TODO introduced in #2153. The previous PR fixed the macOS image version to `macOS-15`, in prder to address a compiler issue in Github daily-CI, since `macos-latest` was pointing to `macos-14` at that time. We can now switch to `macos-latest` since the default image version has been upgraded.


edopao:
All tests pass in the daily-CI workflow: https://github.com/GridTools/gt4py/actions/runs/17430636882

PR #2247, opened 2025-09-05
fix[cartesian]: memlets of data dims in jit workflow

romanc:
## Description

Memlets of data_dims in jit workflows were off by one. The memlet was configured to include one too many because ranges in dace are inclusive on both ends (in contrast to many half-open range definitions that are inclusive at the start and exclusive at the end).

Note: This is part of untangling the [`milestone2`](https://github.com/GridTools/gt4py/pull/2202) branch.

## Requirements

  New test added at the NDSL-level where we use the jit workflow
  N/A


twicki (approved):
Does this solve some of the out of bounds issues that the compiler complained about?

PR #2249, opened 2025-09-05
perf: cache dispatch results in eve.visitors  

egparedes:
Performance optimizations for `eve.visitors` classes. The visitor method name lookups per node class are now cached at class level. Every visitor class gets a new cache to avoid caching incorrect names in hierarchies of visitors.

Additionally, a draft attempt to add support for benchmarking the internals of gt4py is also included, together with other minor cleanups in the `eve` subpackage.


havogt (approved):
lgtm

PR #2251, opened 2025-09-08
bug[next][dace]: fix symbol mapping in lowering of let-statements

edopao:
This PR solves a bug which slipped through in #2240. The mapping of symbols for array shape needs to follow the new canonical order of field dimensions, for which the vertical dimensions follow the local one. Therefore, we apply the gt4py `order_dimensions()` utility on the full field domain, which includes also the local dimension.


src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_types.py
```diff
@@ -152,23 +152,30 @@ def get_symbol_mapping(
...
+            globals_size = [
+                size
+                for dim, size in zip(full_dims, outer_desc.shape, strict=True)
+                if dim != local_dim
+            ]
+        else:
+            globals_size = outer_desc.shape
+
+        symbol_mapping: dict[str, dace.symbolic.SymbolicType] = {}
+        for dim, origin, size in zip(self.gt_type.dims, self.origin, globals_size, strict=True):
```
iomaganaris, line 169:
```suggestion
        for dim, origin, size in zip(gtx_common.order_dimensions(self.gt_type.dims), self.origin, globals_size, strict=True):
```
Should this be like this? I guess there are no local dimensions in `self.gt_type.dims` there but still

edopao: I see your point. However, `self.gt_type.dims` should be already ordered. The datatype itself has an implicit validator:
```
class FieldType(DataType, CallableType):
    dims: list[common.Dimension]
    dtype: ScalarType | ListType

    def __str__(self) -> str:
        dims = "..." if self.dims is Ellipsis else f"[{', '.join(dim.value for dim in self.dims)}]"
        return f"Field[{dims}, {self.dtype}]"

    @eve_datamodels.validator("dims")
    def _dims_validator(
        self, attribute: eve_datamodels.Attribute, dims: list[common.Dimension]
    ) -> None:
        common.check_dims(dims)
```

iomaganaris: Discussed offline. `self.gt_type.dims` should already be ordered

PR #2255, opened 2025-09-09
feat[cartesian]: Add backend `dace:cpu_kfirst`

romanc:
## Description

In the process of performance optimizations, we realized that the naive loop structure

```py
for k in range(K):   # vertical loop
    for i in range(I):  # horizontal loops
        for j in range(J):
            # do stencil computation here
```

of the `dace:cpu` backend

- might not be optimal in terms of cache access and
- isn't a 1:1 comparison with existing backends like `gt:cpu_kfirst`

We thus decided to write a small pass to flip vertical and horizontal loops the loops conditionally. This is reflected in the `dace:cpu_kfrist` backend, named after `gt:cpu_kfirst`.

Note: This is part of untangling the `milestone2` branch in PR https://github.com/GridTools/gt4py/pull/2202.

## Requirements

### DIFF
--- a/src/gt4py/cartesian/gtc/dace/passes/utils.py
+++ b/src/gt4py/cartesian/gtc/dace/passes/utils.py
@@ -16,11 +16,5 @@ def list_index(collection: list[tn.ScheduleTreeNode], node: tn.ScheduleTreeNode)
     Comparing with the `is` operator ensures memory comparison. The function `list.index()`
     uses value comparison and might thus yield different results.
     """
-    index = 0
-    for element in collection:
-        # compare with "is" to get memory comparison. ".index()" uses value comparison
-        if element is node:
-            return index
-        index += 1
-
-    raise StopIteration
+    # compare with "is" to get memory comparison. ".index()" uses value comparison
+    return next(index for index, element in enumerate(collection) if element is node)


FlorianDeconinck (approved):
LGTM

twicki (approved):
Looks great, thanks!

src/gt4py/cartesian/gtc/dace/passes/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    uses value comparison and might thus yield different results.
+    """
+    index = 0
+    for element in collection:
+        # compare with "is" to get memory comparison. ".index()" uses value comparison
+        if element is node:
+            return index
+        index += 1
+
+    raise StopIteration
```
egparedes: I'm not sure if this function is really needed since exactly the same behavior can be implemented with a one-liner (including the `StopIteration` in case the element is not found):

```suggestion
    # compare with "is" to get memory comparison. ".index()" uses value comparison
    return next(index for index, element in enumerate(collection) if element is node)
```

romanc: Thanks for the simplification. It's great to learn the pythonic way of doing things. I'd like to keep the function since at least for me it's non-trivial and I can see other passes having the same need.

PR #2256, opened 2025-09-09
bug[next][dace]: fix dimensions order in `find_constant_symbols`

edopao:
This PR solves a bug which slipped through in #2240. The new canonical order of field dimensions was not applied in the `find_constant_symbols` helper function, which searches the strides in the horizontal dimension when `UNSTRUCTURED_HORIZONTAL_HAS_UNIT_STRIDE` is set in gt4py config.

This bug was not caught by icon4py tests because `UNSTRUCTURED_HORIZONTAL_HAS_UNIT_STRIDE` is not set in regular tests, only in benchmark mode. In gt4py, the unit test is now updated to catch this case.

Additional change:
In case of fields with more than one horizontal dimension, the previous implementation threw an exception. This exception is now turned into a warning.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
@@ -9,7 +9,6 @@
 from __future__ import annotations
 
 import dataclasses
-import warnings
 from typing import Any, Optional
 
 import dace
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -40,15 +41,23 @@ def find_constant_symbols(
...
+                h_dims = [dim for dim in p.type.dims if dim.kind == common.DimensionKind.HORIZONTAL]
+                if len(h_dims) == 0:
                     continue
-                elif len(dims) == 1:
-                    dim_index = 0
-                elif len(dims) == 2:
-                    dim_index = 0 if dims[0].kind == common.DimensionKind.HORIZONTAL else 1
+                elif len(h_dims) > 1:
+                    warnings.warn(
+                        f"Unsupported field with multiple horizontal dimensions '{p}'.",
```
iomaganaris: Maybe throwing an error is not a bad idea since this is something unsupported. It can force the user to fix their code instead of keeping something wrong (even if it accidentally works)

iomaganaris (approved):

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
edopao: I was not sure either. I see your point and agree, I will revert the change.

PR #2258, opened 2025-09-11
refactor[next]: Refactor static args mechanism

tehrengruber:
In preperation of the compile time domains #2173 this PR introduces a generic mechanism to add additional information about arguments beyond their type. Static parameters are refactored to use this mechanism.

### DIFF
--- a/src/gt4py/next/ffront/past_to_itir.py
+++ b/src/gt4py/next/ffront/past_to_itir.py
@@ -15,7 +15,7 @@ from typing import Any, Optional, cast
 import devtools
 
 from gt4py.eve import NodeTranslator, concepts, traits, utils as eve_utils
-from gt4py.next import common, config, errors
+from gt4py.next import common, config, errors, utils as gtx_utils
 from gt4py.next.ffront import (
     fbuiltins,
     gtcallable,
@@ -100,14 +100,15 @@ def past_to_gtir(inp: AOT_PRG) -> stages.CompilableProgram:
     if arguments.StaticArg in inp.args.argument_descriptor_contexts:
         static_arg_descriptors = inp.args.argument_descriptor_contexts[arguments.StaticArg]
         if not all(
-            isinstance(arg_descriptor, arguments.StaticArg) or arg_descriptor is None
+            isinstance(arg_descriptor, arguments.StaticArg)
+            or all(el is None for el in gtx_utils.flatten_nested_tuple(arg_descriptor))  # type: ignore[arg-type]
             for arg_descriptor in static_arg_descriptors.values()
         ):
             raise NotImplementedError("Only top-level arguments can be static.")
         static_args = {
             name: im.literal_from_tuple_value(descr.value)  # type: ignore[union-attr]  # type checked above
             for name, descr in static_arg_descriptors.items()
-            if descr
+            if not any(el is None for el in gtx_utils.flatten_nested_tuple(descr))  # type: ignore[arg-type]
         }
         body = remap_symbols.RemapSymbolRefs().visit(itir_program.body, symbol_map=static_args)
         itir_program = itir.Program(


src/gt4py/next/otf/arguments.py
```diff
@@ -9,24 +9,72 @@
...
-class StaticArg(Generic[core_defs.ScalarT]):
-    value: core_defs.ScalarT | tuple[core_defs.ScalarT | tuple, ...]
-    type_: ts.TypeSpec
+class StaticArg(ArgumentDescriptor, Generic[core_defs.ScalarT]):
+    value: extended_typing.MaybeNestedInTuple[core_defs.ScalarT]
+
+    def __post_init__(self) -> None:
+        # transform enum value into the actual value
+        if isinstance(self.value, enum.Enum):
+            object.__setattr__(self, "value", self.value.value)
```
tehrengruber, line 83: @havogt I think this was the actual reason for the cast in `sanitize_args`.

havogt: Can you elaborate?

src/gt4py/next/otf/arguments.py
```diff
@@ -57,21 +109,20 @@ def offset_provider_type(self) -> common.OffsetProviderType:
...
             kwargs={
                 k: type_translation.from_value(v) for k, v in kwargs_copy.items() if v is not None
             },
+            argument_descriptors={},
         )
 
     @classmethod
     def empty(cls) -> Self:
-        return cls(tuple(), {}, {}, None)
+        return cls(tuple(), {}, {}, None, {})
```
havogt, line 153: where is this used?

src/gt4py/next/otf/arguments.py
```diff
@@ -9,24 +9,72 @@
...
 from gt4py.next.otf import toolchain, workflow
-from gt4py.next.type_system import type_specifications as ts, type_translation
+from gt4py.next.type_system import type_info, type_specifications as ts, type_translation
 
 
 DATA_T = typing.TypeVar("DATA_T")
+T = typing.TypeVar("T")
+
+
+class ArgumentDescriptor:
```
havogt:
```suggestion
class ArgumentDescriptor(abc.ABC):
```

src/gt4py/next/otf/arguments.py
```diff
@@ -9,24 +9,72 @@
...
+    def validate(self, name: str, type_: ts.TypeSpec) -> None:
```
havogt:
```suggestion
    @abc.abstractmethod
    def validate(self, name: str, type_: ts.TypeSpec) -> None:
```

src/gt4py/next/otf/arguments.py
havogt: Why not make this abstract?

src/gt4py/next/otf/arguments.py
```diff
@@ -9,24 +9,72 @@
...
+        """
+        Validate argument descriptor.
+
+        This function is called when the type of the argument is available. The name is merely
+        given to give good error messages.
+        """
+        pass
+
+    @classmethod
```
havogt, line 56:
```suggestion
    @classmethod
    @abc.abstractmethod
```

src/gt4py/next/otf/arguments.py
havogt: I think I understand what you mean, but it doesn't ring a bell.

havogt: Hmm. Ok, there is the test.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -164,45 +131,142 @@ def __call__(
...
+            }
+            self._validate_argument_descriptor_mapping()
+        else:
+            for descr_cls, descriptor_expr_mapping in argument_descriptors.items():
+                if (expected := set(self.argument_descriptor_mapping[descr_cls])) != (
+                    got := set(descriptor_expr_mapping.keys())
+                ):
+                    raise ValueError(
+                        f"Argument descriptor {descr_cls.__name__} must be the same for all compiled programs, got {list(got)} expected {list(expected)}."
+                    )
```
havogt: maybe this deserves its own function for better structure?

src/gt4py/next/otf/compiled_program.py
```diff
@@ -164,45 +131,142 @@ def __call__(
...
+        return descriptors
+
+    def _validate_argument_descriptors(
+        self,
+        all_descriptors: ArgumentDescriptors,
+    ) -> None:
+        for descriptors in all_descriptors.values():
+            for expr, descriptor in descriptors.items():
+                param_type = _get_type_of_param_expr(self.program_type, expr)
+                descriptor.validate(expr, param_type)
```
havogt: free function?

src/gt4py/next/otf/arguments.py
tehrengruber: This should have never existed in the first place, but sometimes steps of `backend.Transforms` are executed with empty compile time args which are not used in that step. This is a design flaw of the every step takes a `CompilableProgram` approach in `backend.Transforms`. See `decorator.FieldOperator.foast_stage`

src/gt4py/next/otf/arguments.py
tehrengruber: This function is not abstract. If it is not implemented no validation is needed.

src/gt4py/next/otf/compiled_program.py
tehrengruber: Since I'm using `self` I would keep this as is.

src/gt4py/next/otf/arguments.py
```diff
@@ -8,25 +8,87 @@
...
 from gt4py.next.otf import toolchain, workflow
-from gt4py.next.type_system import type_specifications as ts, type_translation
+from gt4py.next.type_system import type_info, type_specifications as ts, type_translation
 
 
 DATA_T = typing.TypeVar("DATA_T")
+T = typing.TypeVar("T")
+
+
+class ArgumentDescriptor(abc.ABC):
```
egparedes: I think this name is too generic and something more specific will help understanding how is this used. Something reflecting that this is used purely for static (i.e. available at compile-time) information could be better, like for example: `StaticArgDescriptor`, `StaticArgInfo`, `StaticArgSpec`, `CompiledParamSpec`,  ...

Maybe the reason why finding a good name for this is hard is that this class is used for different purposes: instance data is meant to contain concrete static information for a parameter, while the class itself is meant to contain meta-data describing the transformation from concrete data to the restricted compile-time data.

src/gt4py/next/otf/arguments.py
```diff
@@ -8,25 +8,87 @@
...
+    """
+
+    def validate(self, name: str, type_: ts.TypeSpec) -> None:  # noqa: B027  # method is not abstract, but just empty when not implemented
+        """
+        Validate argument descriptor.
+
+        This function is called when the type of the argument is available. The name is merely
+        given to give good error messages.
+        """
+        pass
```
egparedes, line 54: This feels like a secondary use case for this class (at best) so I would add it at the end of the class. Additionally, the name `validate` is super-generic and misleading since this is meant to validate only the type spec, and don't see why should receive the name from outside just for the error message, instead of relying purely in the instance data.

src/gt4py/next/otf/arguments.py
```diff
@@ -8,25 +8,87 @@
...
+
+    @classmethod
+    @abc.abstractmethod
+    def attribute_extractor(cls, arg_expr: str) -> dict[str, str]:
```
egparedes: I think the name is wrong. I would expect that this returns a `Callable` not a dictionary of expressions. IMHO, something like `extractor_exprs` would be more accurate. I also find `in_arg_expr` slightly easier to understand for the argument name, but it's just a personal preference.

src/gt4py/next/otf/arguments.py
```diff
@@ -8,25 +8,87 @@
...
+        a cache key by just hashing `my_arg.value` instead of first constructing the descriptor.
+        """
+        ...
 
 
 @dataclasses.dataclass(frozen=True)
-class StaticArg(Generic[core_defs.ScalarT]):
-    value: core_defs.ScalarT | tuple[core_defs.ScalarT | tuple, ...]
-    type_: ts.TypeSpec
+class StaticArg(ArgumentDescriptor, Generic[core_defs.ScalarT]):
```
egparedes: I think this should be renamed to something like `StaticValueArg` or similar, depending on the final name used for the superclass

src/gt4py/next/otf/arguments.py
```diff
@@ -8,25 +8,87 @@
...
+    value: extended_typing.MaybeNestedInTuple[core_defs.ScalarT]
+
+    def __post_init__(self) -> None:
+        # transform enum value into the actual value
+        if isinstance(self.value, enum.Enum):
+            object.__setattr__(self, "value", self.value.value)
+
+    def validate(self, name: str, type_: ts.TypeSpec) -> None:
```
egparedes, line 85: As mentioned above, I don't fully understand why `name` needs to be passed from outside. Shouldn't it be part of the instance data?

src/gt4py/next/otf/arguments.py
```diff
@@ -8,25 +8,87 @@
...
+                message=f"Invalid static argument '{name}', expected '{type_}', but static value '{self.value}' has type '{actual_type}'.",
+                location=None,
+            )
+
+    @classmethod
+    def attribute_extractor(cls, arg_expr: str) -> dict[str, str]:
+        return {"value": arg_expr}
 
 
 @dataclasses.dataclass(frozen=True)
```
egparedes, line 104: I know this is not part of thus PR but since this is a refactoring, it could be a good opportunity to rework this class. 

This is just a static version of [`inspect.BoundArguments`](https://devdocs.io/python~3.13/library/inspect#inspect.BoundArguments) (I mean _static_ because in `inspect.BoundArguments` both `.args` and `kwargs` are recomputed dynamically everytime) so I'd move it to a common utilities module, rename it to `CallArgs` and move `from_signature` to be the default constructor.

src/gt4py/next/otf/arguments.py
```diff
@@ -45,10 +107,14 @@ def from_signature(cls, *args: Any, **kwargs: Any) -> Self:
...
-    args: tuple[ts.TypeSpec | StaticArg, ...]
-    kwargs: dict[str, ts.TypeSpec | StaticArg]
+    args: tuple[ts.TypeSpec, ...]
+    kwargs: dict[str, ts.TypeSpec]
     offset_provider: common.OffsetProvider  # TODO(havogt): replace with common.OffsetProviderType once the temporary pass doesn't require the runtime information
     column_axis: Optional[common.Dimension]
+    argument_descriptors: Mapping[
+        type[ArgumentDescriptor],
+        dict[str, ArgumentDescriptor],
+    ]
```
egparedes: Can you explain why this needs to be added here? It doesn't look like this belongs here in my opinion.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -38,94 +46,52 @@ def _hash_compiled_program_unsafe(cp_key: CompiledProgramsKey) -> int:
...
     backend: gtx_backend.Backend
     definition_stage: ffront_stages.ProgramDefinition
     program_type: ts_ffront.ProgramType
-    static_params: Sequence[str] | None = None  # not ordered
+    #: mapping from an argument descriptor type to a list of parameters or expression thereof
+    #: e.g. `{arguments.StaticArg: ["static_int_param"]}`
+    #: Note: The list is not ordered.
+    argument_descriptor_mapping: dict[type[arguments.ArgumentDescriptor], Sequence[str]] | None
 
     _compiled_programs: eve_utils.CustomMapping = dataclasses.field(
```
egparedes, line 203: I would also make `_compiled_programs` and `_offset_provider_type_cache` cached properties instead of _fake_ dataclass fields just to use the factory

src/gt4py/next/otf/compiled_program.py
```diff
@@ -164,45 +133,163 @@ def __call__(
...
                     *args, offset_provider=offset_provider, enable_jit=False, **kwargs
                 )  # passing `enable_jit=False` because a cache miss should be a hard-error in this call`
             raise RuntimeError("No program compiled for this set of static arguments.") from e
 
     @functools.cached_property
-    def _static_arg_indices(self) -> tuple[int, ...]:
-        if self.static_params is None:
-            # this could also be done in `__call__` but would be an extra if in the fast path
-            self.static_params = ()
+    def _argument_descriptor_cache_key_from_args(self) -> Callable:
```
egparedes, line 256: I would extract the low-level implementation details of all these cached properties building the extractors and similar stuff to utility functions outside the class, and then call them in the cached property.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -164,45 +133,163 @@ def __call__(
...
+                attr_extractor = descriptor_cls.attribute_extractor(arg_expr)
+                attrs = attr_extractor.keys()
+                for attr in attrs:
+                    elements.append(
+                        getattr(eval(f"{arg_expr}", descriptor_context[descriptor_cls]), attr)
+                    )
+        return tuple(elements)
+
+    @functools.cached_property
+    def _argument_descriptor_attr_retrievers(
```
egparedes: Shouldn't this be part of the `ArgumentDescriptor` class? I think the implementation should be in the Abstract Base class, which would call `argument_extractor` expressions method defined in the concrete classes.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -164,45 +133,163 @@ def __call__(
...
+        for descriptor_cls, arg_exprs in self.argument_descriptor_mapping.items():  # type: ignore[union-attr]  # can never be `None` at this point
+            for arg_expr in arg_exprs:
+                attr_exprs = descriptor_cls.attribute_extractor(arg_expr)
+                retrievers[descriptor_cls][arg_expr] = eval(
+                    f"""lambda {",".join(params)}: {make_dict_expr(attr_exprs)}"""
+                )
+
+        return retrievers
+
+    def _make_argument_descriptors(self, *args: Any, **kwargs: Any) -> ArgumentDescriptors:
```
egparedes: I also think it would be cleaner to move these `_make`/`_validate` utility functions outside of the class, since they do not contribute to the main responsibility of this class, they do not modify its state, and they only use very concrete pieces of information stored in the class, which can be passed as arguments.

src/gt4py/next/otf/compiled_program.py
tehrengruber: Use final

src/gt4py/next/otf/compiled_program.py
tehrengruber: As discussed this will happen in your PR.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -164,45 +133,163 @@ def __call__(
...
+                    )
+
+    def _validate_argument_descriptor_mapping(self) -> None:
+        if self.argument_descriptor_mapping is None:
+            return
+        context = _make_param_context_from_func_type(self.program_type.definition, lambda x: None)
+        for descr_cls, exprs in self.argument_descriptor_mapping.items():
+            for expr in exprs:
+                try:
+                    if eval(expr, context) is not None:
```
tehrengruber:
```suggestion
                    # TODO(tehrengruber): Reevaluate the way we validate here when we add support for containers.
                    #  Building an `ast` looks like the better solution when we have more than just tuples.
                    if eval(expr, context) is not None:
```

src/gt4py/next/otf/arguments.py
tehrengruber: We decided to use `ArgStaticDescriptor`

src/gt4py/next/otf/arguments.py
tehrengruber: As discussed I don't agree with this characterization. This method validates that the instance data is valid and valid in the context of an actual program. It is not only validating the type spec. Passing the name is an easy way to give descriptive error messages to the user.

src/gt4py/next/otf/arguments.py
tehrengruber: Out of scope. I think we are massively overthinking here. This name was decided in a PR reviewed by you very recently. Changing this name would mean changing all docstrings, error messages and derived names. Should I also rename `static_params` now and all user code? This seems inefficient to me and is really makes refactoring PRs like this one no fun.

src/gt4py/next/otf/arguments.py
tehrengruber: As discussed this is inconvenient either for the user configuring an argument descriptor or adds additional boilerplate that is needed nowhere except for in this method.

src/gt4py/next/otf/arguments.py
tehrengruber: Out-of-scope.

src/gt4py/next/otf/arguments.py
tehrengruber: This is used to pass the information down the toolchain.

src/gt4py/next/otf/compiled_program.py
tehrengruber: I started this and it did not improve the code-quality. I kept it as is.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -250,17 +332,18 @@ def compile(
...
-                    dict(zip(static_args.keys(), static_values, strict=True)),
+                    argument_descriptors={
+                        arguments.StaticArg: dict(
+                            zip(
+                                static_args.keys(),
+                                [arguments.StaticArg(value=v) for v in static_values],
+                                strict=True,
+                            )
+                        ),
+                    },
```
havogt, line 412: I would move this logic into the `_compile_variant`

src/gt4py/next/ffront/stages.py
```diff
@@ -129,7 +129,10 @@ def add_func_to_fingerprint(obj: types.FunctionType, hasher: xtyping.HashlibAlgo
 
 @add_content_to_fingerprint.register
 def add_dict_to_fingerprint(obj: dict, hasher: xtyping.HashlibAlgorithm) -> None:
-    for key, value in sorted(obj.items()):
+    for key, value in sorted(
+        obj.items(),
+        key=lambda x: (x[0].__module__, x[0].__qualname__) if isinstance(x[0], type) else x[0],
+    ):
```
havogt: I don't understand, I think the lambda deserves a name.

src/gt4py/next/otf/arguments.py
```diff
@@ -45,10 +117,14 @@ def from_signature(cls, *args: Any, **kwargs: Any) -> Self:
...
-    args: tuple[ts.TypeSpec | StaticArg, ...]
-    kwargs: dict[str, ts.TypeSpec | StaticArg]
+    args: tuple[ts.TypeSpec, ...]
+    kwargs: dict[str, ts.TypeSpec]
     offset_provider: common.OffsetProvider  # TODO(havogt): replace with common.OffsetProviderType once the temporary pass doesn't require the runtime information
     column_axis: Optional[common.Dimension]
+    argument_descriptors: Mapping[
+        type[ArgStaticDescriptor],
+        dict[str, ArgStaticDescriptor],
+    ]
```
havogt, line 131: I think it would be much easier to understand if the information of the args/kwargs would be stored in the same structure. Why is TypeSpec stored linearly, but the rest in the argument_descriptors mapping?

src/gt4py/next/otf/compiled_program.py
```diff
@@ -38,94 +47,84 @@ def _hash_compiled_program_unsafe(cp_key: CompiledProgramsKey) -> int:
...
-    ):
-        raise errors.DSLTypeError(
-            message=f"Invalid static argument types when trying to compile '{program_name}' with type '{program_type}':\n"
-            + ("\n".join([f"  - {error}" for error in mismatch_errors])),
-            location=None,
+def _make_param_context_from_func_type(
+    func_type: ts.FunctionType,
+    type_map: Callable[[ts.TypeSpec], T] = lambda x: x,  # type: ignore[assignment, return-value]  # mypy not smart enough to narrow type for default
+) -> dict[str, extended_typing.MaybeNestedInTuple[T]]:
+    params = func_type.pos_or_kw_args | func_type.kw_only_args
```
havogt, line 77: Please add a small description to the function.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -154,7 +154,9 @@ def __call__(
         it is an error.
         """
         args, kwargs = type_info.canonicalize_arguments(self.program_type, args, kwargs)
-        static_args_values = tuple(args[i] for i in self._static_arg_indices)
+        static_args_values = self._argument_descriptor_cache_key_from_args(*args, **kwargs)
+        # TODO: Dispatching over offset provider type is wrong, especially when we use compile time
```
havogt: should this TODO stay?

src/gt4py/next/otf/compiled_program.py
tehrengruber: TODO

src/gt4py/next/ffront/stages.py
tehrengruber: Conclusion: Iterate over keys only, then use dict access, then key function can be a dedicated function with a descriptive name.

src/gt4py/next/otf/compiled_program.py
tehrengruber: Yes.

src/gt4py/next/ffront/stages.py
tehrengruber: Used a small lambda with a comment.

havogt (approved):

PR #2259, opened 2025-09-12
fix[cartesian]: dace backend fixes

romanc:
## Description

This PR fixes two issues with the DaCe backends of `gt4py.cartesian`:

1. In the Tree IR, we were relying arrays shapes to define the domain symbols `__I`, `__J`, and `__K`. While normally the set of all fields passed to the stencil cover all three Cartesian dimensions, this doesn't have to be. See the stencil `assign_2d(...)` below for a valid stencil with only passing an IJ-Field. In that particular case, the `__K` symbol  was not defined. That's a problem because `__K` is used in the bounds of the vertical loop. Commit 801c5263ae34c94cc68c977f191fd8ed0b9dc5ef fixes this by defining the domain symbols up-front.
2. When freezing origin/domain into the SDFG, we weren't properly considering some corner cases of masked cartesian dimensions and data dimensions. This PR untangles the "let's try to be smart" `zip()` call over three arrays/tuples of potentially different length and clearly separates cartesian and data dimensions. I removed the check for `origins` to be a `tuple` because types say that this never happens (and tests agree). In case the origin of a field is not defined (and can't be derived from the `__all__` entry), we now `raise`  a `ValueError` as precaution instead of failing silently. This case shouldn't happen (tm).

```py
def assign_2d(field: gtscript.Field[gtscript.IJ, np.float32], scalar: np.float32):
        field = scalar
```

With commit e5a089b73ee120a901b9a649ad1bb9bb2f6ecd74, we add some types to test utils and do additional checks on dimensions. Useful for further test development.

This PR brings back fixes from the "milestone2" branch, see https://github.com/GridTools/gt4py/pull/2202. 

## Requirements

  N/A

### DIFF
--- a/src/gt4py/cartesian/backend/dace_backend.py
+++ b/src/gt4py/cartesian/backend/dace_backend.py
@@ -117,16 +117,14 @@ def _sdfg_add_arrays_and_edges(
             ranges = []
             origin = origins.get(name, origins.get("_all_", None))
 
+            if origin is None:
+                raise ValueError("Freeze origin/domain: Unspecified origin for field `{name}`.")
+
             # Read boundaries for axis-bound fields
             index = 0
             for cartesian_index, axis in enumerate(CartesianSpace.names):
                 if axis not in axes:
                     continue
-                # TODO: Not sure yet how we can get into situations where
-                #       origin is None in the first place. Tests fail if we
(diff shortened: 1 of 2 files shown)


romanc:
cscs-ci run

PR #2261, opened 2025-09-16
fix[next]: Fix multi-node scaling issues of `FileCache`

tehrengruber:
Multi-node runs frequently hanged with the SQLite based implementation from the `diskcache` package, rendering runs beyond 16 nodes impossible. Here we replace the previous external implementation with a simple one-file-per-cache-entry implementation that uses `filelock.FileLock` for locking.


havogt:
cscs-ci run

src/gt4py/eve/extended_typing.py
```diff
@@ -542,6 +542,18 @@ class _FrozenDataclassParamsABC(_DataclassParamsABC):
     frozen: Literal[True]
 
 
+_KT = TypeVar("_KT", contravariant=True)
+_VT = TypeVar("_VT")
+
+
+class OpaqueMutableMapping(Protocol[_KT, _VT]):
```
tehrengruber, line 549: I introduced this new protocol since we don't need and want to implement the `__iter__`, `__len__` functionality required for a full `MutableMapping`.

src/gt4py/_core/filecache.py
```diff
@@ -0,0 +1,49 @@
+# GT4Py - GridTools Framework
+#
+# Copyright (c) 2014-2024, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+import pathlib
```
egparedes, line 11:
```suggestion
from __future__ import annotations

import pathlib
```

src/gt4py/_core/filecache.py
```diff
@@ -0,0 +1,49 @@
...
+from gt4py._core import locking
+from gt4py.eve import utils as eve_utils
+
+
+class FileCache:
+    path: pathlib.Path
+
+    def __init__(self, path: str):
+        self.path = pathlib.Path(path)
+        self.path.mkdir(parents=True, exist_ok=True)
```
egparedes: I would at least support that path is a `PathLike` (e.g. `pathlib.Path`)  instance, and also normalize it before store it.

```suggestion
    def __init__(self, path: str | os.PathLike):
        self.path = pathlib.Path(path).resolve()
        self.path.mkdir(parents=True, exist_ok=True)
```

src/gt4py/_core/filecache.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def _get_path(self, key: Hashable) -> pathlib.Path:
+        """Return the path where an item with `key` is stored."""
+        key = eve_utils.content_hash(key)
+        path = self.path.joinpath(key).with_suffix(".pkl")
```
egparedes: Just a suggestion which I find more readable, feel free to ignore:
```suggestion
        path = self.path / f"{key}.pkl"
```

src/gt4py/_core/filecache.py
```diff
@@ -0,0 +1,49 @@
...
+        path.parent.mkdir(parents=True, exist_ok=True)
```
egparedes: Is this needed if the parent folder has already been created in the `__init__`?

src/gt4py/_core/filecache.py
```diff
@@ -0,0 +1,49 @@
...
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        with locking.lock(self._get_path(key)):
+            with open(self._get_path(key), "wb") as f:
+                pickle.dump(value, f, protocol=pickle.HIGHEST_PROTOCOL)
+
+    def __delitem__(self, key: Hashable) -> None:
+        if key not in self:
+            raise KeyError(key)
+        with locking.lock(self._get_path(key)):
+            self._get_path(key).unlink()
```
egparedes: Likely faster and more readable but feel free to ignore
```suggestion
        with locking.lock(path := self._get_path(key)):
            path.unlink()
```

src/gt4py/_core/filecache.py
```diff
@@ -0,0 +1,49 @@
...
+        path = self.path.joinpath(key).with_suffix(".pkl")
+        path.parent.mkdir(parents=True, exist_ok=True)
+        return path
+
+    def __getitem__(self, key: Hashable) -> Any:
+        if key not in self:
+            raise KeyError(key)
+        with locking.lock(self._get_path(key)):
+            with open(self._get_path(key), "rb") as f:
+                return pickle.load(f)
```
egparedes:
```suggestion
        with locking.lock(path := self._get_path(key)):
            with open(path, "rb") as f:
                return pickle.load(f)
```

src/gt4py/_core/filecache.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
+        with locking.lock(self._get_path(key)):
+            with open(self._get_path(key), "wb") as f:
```
egparedes:
```suggestion
        with locking.lock(path := self._get_path(key)):
            with open(path, "wb") as f:
```

Using `pickle.HIGHEST_PROTOCOL` is not future-proof because if the pickle protocol version changes across python versions, reading from the older version a file created with the newer version will fail. I think this should be set to a python version independent constant or the version of the protocol should be added to the pickle file name, to make sure the right protocol version is used.

src/gt4py/_core/locking.py
```diff
@@ -28,13 +28,20 @@ def __name__(self) -> str: ...
 
 
 def lock(directory: pathlib.Path | str, suffix: str = ".lock") -> ContextManager:
```
egparedes, line 31:
```suggestion
def lock(path: str | os.PathLike, suffix: str = ".lock") -> ContextManager:
```
Consequently, change the `_LockConstructor` to also accept a `str | os.PathLike` argument (`filelock` already supports exactly this interface). Ideally, I would also rather pass the path as a PathObject and not as str to the `_LockConstructor` function ...

tests/core_tests/unit_tests/test_filecache.py
```diff
@@ -0,0 +1,48 @@
...
+
+@pytest.fixture
+def temp_cache():
+    """Fixture to create a temporary FileCache directory and clean it up."""
+    tmpdir = tempfile.mkdtemp()
+    cache = filecache.FileCache(tmpdir)
+    yield cache
+    shutil.rmtree(tmpdir)
+
+
```
egparedes, line 24: I would also add a test to make sure that the key computation does not rely on `hash()` and thus is stable across interpret initializations. Something like:

```python
hash_key == hash("key")
filecache_key = str(filecache._get_path("key"))

with mock.patch("builtins.hash", lambda *x: f"__mocked_hash_for_{x!s}"):
    assert hash("key") != hash_key
    assert str(filecache._get_path("key")) == filecache_key
```

src/gt4py/eve/extended_typing.py
egparedes, line 549: Missing docstring

tests/core_tests/unit_tests/test_filecache.py
egparedes: It would be maybe more accurate to just run another instance of the python interpreter with `subprocess.run()` and double check that it generates the same key.

src/gt4py/_core/filecache.py
tehrengruber: > reading from the older version a file created with the newer version will fail

As it should. Anyway, I put version 5 here.

tests/core_tests/unit_tests/test_filecache.py
tehrengruber: This is too much effort and not a priority for me. Feel free to implement this.

src/gt4py/_core/filecache.py
egparedes: My point was that if the load fails because of the wrong protocol version but the file exists, it might be impossible to run the same code with different versions of python in the same folder. If you're sure that's not the case (and you have tested it), add a comment and leave the original code untouched.

src/gt4py/_core/filecache.py
```diff
@@ -0,0 +1,51 @@
...
+import pathlib
+import pickle
+from typing import Any, Hashable
+
+from gt4py._core import locking
+from gt4py.eve import utils as eve_utils
+
+
+class FileCache:
+    path: pathlib.Path
```
egparedes, line 24: Missing docstring here and/or in the module level (at least one of them is needed).

PR #2267, opened 2025-09-19
refactor[next]: unroll_reduce shift via typesystem

havogt:
incomplete

### DIFF
--- a/src/gt4py/next/iterator/transforms/unroll_reduce.py
+++ b/src/gt4py/next/iterator/transforms/unroll_reduce.py
@@ -16,6 +16,7 @@ from gt4py.next import common
 from gt4py.next.iterator import ir as itir
 from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm, ir_makers as im
 from gt4py.next.iterator.ir_utils.common_pattern_matcher import is_applied_lift
+from gt4py.next.type_system import type_specifications as ts
 
 
 def _is_neighbors(arg: itir.Expr) -> TypeGuard[itir.FunCall]:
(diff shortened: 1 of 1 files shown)


src/gt4py/next/iterator/transforms/unroll_reduce.py
```diff
@@ -59,7 +60,11 @@ def _get_partial_offset_tag(arg: itir.FunCall) -> str:
 
 
 def _get_partial_offset_tags(reduce_args: Iterable[itir.Expr]) -> Iterable[str]:
```
havogt, line 43: todo cleanup non used functions

src/gt4py/next/iterator/transforms/unroll_reduce.py
```diff
@@ -42,11 +42,10 @@ def _get_neighbors_args(reduce_args: Iterable[itir.Expr]) -> Iterator[itir.FunCa
...
-    return [
-        arg.type.offset_type.value
-        for arg in reduce_args
-        if isinstance(arg.type, ts.ListType) and arg.type.offset_type is not None
-    ]
+    assert all(isinstance(arg.type, ts.ListType) for arg in reduce_args)
+    assert all(arg.type.offset_type is not None for arg in reduce_args)  # type: ignore[union-attr] # checked in previous line
```
havogt: TODO: try to remove optional from offset_type in ListType

src/gt4py/next/iterator/embedded.py
```diff
@@ -1416,8 +1416,10 @@ def neighbors(offset: runtime.Offset, it: ItIterator) -> _List:
 
 
 @builtins.list_get.register(EMBEDDED)
-def list_get(i, lst: _List[Optional[DT]]) -> Optional[DT]:
-    return lst[i]
+def list_get(i, lst: _List[Optional[DT]]) -> Optional[DT] | Undefined:
+    if i < len(lst.values):
+        return lst[i]
+    return _UNDEFINED  # might happen for lists originating from neighbors with skip values
```
tehrengruber, line 1422: Shouldn't this be protected by a `can_deref` and then not happen?

havogt: >Shouldn't this be protected by a can_deref

Yes
>and then not happen?

No.

Because `if_(can_deref(...), deref(shift(...)), 0)`, i.e. eager execution. In our tests it only happens if we do `unroll_reduce` for itir embedded, but still a bug I guess.

Issue #2266, opened 2025-09-19
next: segmentation fault in SDFG call in icon4py diffusion granule


edopao:
This issue manifests itself as random failures with a segmentation fault in data tests. In most cases, retriggering the test makes it pass. However, in icon4py PR [#747](https://github.com/C2SM/icon4py/pull/747) this issue appears consistently with the dace cpu backend. The PR is adding a benchmark test, parameterized with 2 different grids. As a benchmark, this test will run sequentially, first with one offset provider for `grid0`, then another one for `grid1`. The first test passes, the second always fails because of the segmentation fault.

**First hypothesis (didn't work)**
The connectivity fields used in the first test are released before the second test is run. However, the dace backend uses SDFG fast call: not only the program handle is cached in the compiled-program cache, but also the tuple of arguments that are passed to the SDFG call. This tuple (a dace data structure) contains array pointers for the program fields. The gt4py invocation of fast-call was already updating the array pointers of data fields, because one program can be invoked with different fields. For connectivities, it was not updating the array pointers based on the assumption that the offset provider does not change during the application lifetime. This is true for a real application, and it works in icon4py for both greenline and blueline. However, it does not hold for a pytest session, where we test a stencil program with different offset providers. It can happen that a program handle in cache still references a connectivity array (the raw data pointer) which has been released. This happens because the cache key for compiled programs is computed based on `offset_provider_type`, which can be the same for two different offset providers (same attributes in `NeighborConnectivityType`, different array pointers).

The gt4py PR #2268 tried to always set the connectivity fields, even in SDFG fast call. However, it did not help.

edopao:
The origin of the segmentation fault was found in wrong size of the arrays allocated by the icon4py test. This was not a gt4py/dace issue.

PR #2268, opened 2025-09-19
fix[next][dace]: always update connectivities in fast-call for pytest session

edopao:
This PR addresses issue #2266.


edopao:
This change didn't fix the segmentation fault.

PR #2270, opened 2025-09-24
fix[next][dace]: ensure unique names for conditional blocks and loop regions

edopao:
This PR is not fixing any known issue, however it is a code improvement that prevents potential issues in the future.

This was found while analyzing a dace issue, for which we have a bugfix in dace repository: https://github.com/spcl/dace/pull/2158


philip-paul-mueller (approved):
Are these the only places where it happens?
I would also suggest to apply this to the data descriptors, although I think they are simpler to handle, because they are centralized, but it would make debugging simpler.

PR #2273, opened 2025-09-24
feat[next][dace]: extend CopyChainRemover to remove full-write copies

edopao:
The current design of `CopyChainRemover` allowed to remove temporary arrays which are used as source node of full-read copies. Suppose a temporary `__tmp` which is single-use data, with shape `(10,)`: `__tmp` can be removed in `__tmp[0:10] -> A[x:x+10]`

With this change, `CopyChainRemover` can also be applied in the _write_ direction, that is to remove temporary arrays which are used as destination node of full-write copies. For example, still using the single-use temporary `__tmp` as above: `__tmp` can be removed in `A[x:x+10] -> __tmp[0:10]`.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
@@ -59,7 +59,7 @@ def gt_remove_copy_chain(
     return result if result != 0 else None
 
 
-class _CopyChainRemoverMode(Enum):
+class CopyChainRemoverMode(Enum):
     """Switch for `CopyChainRemover` mode.
 
     The `NONE` mode is invalid, it is there for signaling that the transformation
@@ -96,7 +96,7 @@ class CopyChainRemover(dace_transformation.SingleStateTransformation):
     - `A1` is a transient and must have the same dimensionality as `A2`.
     - `A1` is fully read by `A2`.
 
-    This transformation corresponds to the `PULL` mode of `_CopyChainRemoverMode`.
+    This transformation corresponds to the `PULL` mode of `CopyChainRemoverMode`.
     In a second iteraton, the transformation was extended to support the mirror
     case (`PUSH`), where A2 is removed if the following requirements are satisfied:
     - Through the merging of `A1` and `A2` no cycles are created.
@@ -205,7 +205,7 @@ class CopyChainRemover(dace_transformation.SingleStateTransformation):
         if dst_subset is None:
             return False
 
-        if self._get_copy_chain_mode(sdfg, graph) == _CopyChainRemoverMode.NONE:
+        if self._get_copy_chain_mode(sdfg, graph) == CopyChainRemoverMode.NONE:
             return False
 
         # We have to ensure that no cycle is created through the removal of one node.
@@ -265,7 +265,7 @@ class CopyChainRemover(dace_transformation.SingleStateTransformation):
         old_node: dace_nodes.AccessNode
         new_node_offsets: Sequence[dace_sym.SymbolicType]
         copy_chain_mode = self._get_copy_chain_mode(sdfg, graph)
-        if copy_chain_mode == _CopyChainRemoverMode.PULL:
+        if copy_chain_mode == CopyChainRemoverMode.PULL:
             old_node = a1
             new_node = a2
             # Now we compose the new subset.
@@ -282,7 +282,7 @@ class CopyChainRemover(dace_transformation.SingleStateTransformation):
             #   have to modify this computation slightly.
             new_node_offsets = a1_to_a2_dst_subset.min_element()
         else:
-            assert copy_chain_mode == _CopyChainRemoverMode.PUSH
+            assert copy_chain_mode == CopyChainRemoverMode.PUSH
             old_node = a2
             new_node = a1
             # We compose the new subset in a similar way, but using the index where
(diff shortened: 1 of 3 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -249,82 +281,95 @@ def apply(
...
         #       (s1) ------[c:d]-> (A1) -[0:N]------[a:b]-> (A2)
         #       (s1) ---------[(a + c):(a + c + (d - c))]-> (A2)
         #  Thus the offset is simply given by `a`, the start where `a1` is written into
         #  `a2`.
         #  NOTE: If we ever allow the that `a1` is not fully read, then we would have
         #   to modify this computation slightly.
-        a2_offsets: Sequence[dace_sym.SymExpr] = a1_to_a2_dst_subset.min_element()
+        new_node: dace_nodes.AccessNode
+        old_node: dace_nodes.AccessNode
+        new_node_offsets: Sequence[dace_sym.SymExpr]
```
philip-paul-mueller:
```suggestion
        new_node_offsets: Sequence[dace_sym.SymbolicType]
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -249,82 +281,95 @@ def apply(
...
-        #  `a1` writes into `a2`.
+        #  copied into `a2` for the read direction. Thus the destination of the
+        #  original source, i.e. whatever write into `a1`, is just offset by the
+        #  beginning of the range `a1` writes into `a2`.
         #       (s1) ------[c:d]-> (A1) -[0:N]------[a:b]-> (A2)
         #       (s1) ---------[(a + c):(a + c + (d - c))]-> (A2)
         #  Thus the offset is simply given by `a`, the start where `a1` is written into
         #  `a2`.
         #  NOTE: If we ever allow the that `a1` is not fully read, then we would have
         #   to modify this computation slightly.
```
philip-paul-mueller: This whole explanation belongs to the "read" direction.
Thus it should be inside the respective branch of an `if` and there should be an equivalent description for the "write" direction.
Alternatively, you can also reformulate it in such a way that it applies to both cases, in that case it can remain here.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -105,10 +117,14 @@ class CopyChainRemover(dace_transformation.SingleStateTransformation):
     def __init__(
         self,
         *args: Any,
+        direction: Literal["read", "write"],
         single_use_data: dict[dace.SDFG, set[str]],
         **kwargs: Any,
     ) -> None:
+        if direction not in {"read", "write"}:
+            raise ValueError("Direction must be either 'read' or 'write'.")
```
philip-paul-mueller: I do not like the name "read" and "write" for the direction.
However, currently I also do not have better names for them, I was thinking about "pull" and "push" but I do not remember why I thought they were good.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -97,6 +107,8 @@ class CopyChainRemover(dace_transformation.SingleStateTransformation):
     node_a1 = dace_transformation.PatternNode(dace_nodes.AccessNode)
     node_a2 = dace_transformation.PatternNode(dace_nodes.AccessNode)
 
+    _direction: str
```
philip-paul-mueller:
```suggestion
    direction = dace_properties.Property(dtype=str, ...)
```

However, if you follow with my proposal to remove it as an _argument_, then you no longer need it.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -34,6 +35,8 @@ def gt_remove_copy_chain(
 
     Args:
         sdfg: The SDFG to process.
+        direction: One of "read" or "write". When "read", try to remove a transient
+            node which is fully read; otherwise a transient which is fully written.
```
philip-paul-mueller: I would remove it from the arguments.
Instead I think that transformation should figuring out the mode on its own.

philip-paul-mueller (changes requested):
I am kind of okay with the proposed changes, but there should be some modifications, see comments.

Furthermore, I am thinking that `direction` should not be an argument to the transformation.
Instead the transformation should figuring out which mode/direction to use on its own.
It might allow to write it in a bit better way.

One of my main concern is, however, that this change is a bit invasive.
With that I mean is that the transformation was originally written to handle `concat_where` expressions, where each "level" consecutively gets bigger.
However, I do not think that this is a big problem.

edopao:
CI is failing on Beverin because of a vCluster issue, but it passes on Santis.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -58,6 +59,22 @@ def gt_remove_copy_chain(
     return result if result != 0 else None
 
 
+class _CopyChainRemoverMode(Enum):
```
philip-paul-mueller: You mentioned this in the doc string of `CopyChainRemover` thus it should not be a private class, but is kind of public.
You should either make this public, i.e. remove the `_` or reformulate the doc string of `CompyChainRemover`.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -247,84 +258,160 @@ def apply(
...
+            # We compose the new subset in a similar way, but using the index where
+            #  `a2` starts reading from `a1`.
+            new_node_offsets = a1_to_a2_src_subset.min_element()
 
         # Handle the producer side of things.
-        for producer_edge in list(graph.in_edges(a1)):
+        for producer_edge in list(graph.in_edges(old_node)):
             producer: dace_nodes.Node = producer_edge.src
             producer_conn = producer_edge.src_conn
+            if old_node == a2 and producer is new_node:
```
philip-paul-mueller: Optional: Insert an assert to check the `copy_chain_mode`.

philip-paul-mueller (approved):
LGTM.
I although have some minor comments.

PR #2276, opened 2025-09-26
feature[cartesian]: absolute indexing in K (experimental)

romanc:
## Description

This PR drafts an experimental feature to add support for absolute indexing in the `K` dimension. First version, to be cleaned up ... 

## Requirements


romanc:
cscs-ci run

docs/development/ADRs/cartesian/frontend-indexing-absolute-k.md
```diff
@@ -0,0 +1,74 @@
+# ⚠️ Absolute K indexing
+
+In the context of porting physics parametrizations, some field accesses were at absolute indices in the K-axis. We thus decided to expand the DLS be able to port these stencils. We accept the increase in DSL surface and test it as an [experimental feature](./general-experimental-features.md).
+
+## Context
+
+Porting physics parametrizations, we found two following two access patterns
+
+```fortran
+do L = 1, LM
```
twicki: I know the L is describing layers here. Should we stick with IJK in the GT-world neverthenless for it to be easier to translate?

docs/development/ADRs/cartesian/frontend-indexing-absolute-k.md
```diff
@@ -0,0 +1,74 @@
...
+  do J = 1, JM
+   do I = 1, IM
+      ...
+      Field(i, j, LM)  ! access top layer
```
twicki: small i,j with `I,J` as loop-variables

docs/development/ADRs/cartesian/frontend-indexing-absolute-k.md
```diff
@@ -0,0 +1,74 @@
...
+do L = 1, LM
+  do J = 1, JM
+   do I = 1, IM
+      ...
+      computed_k_index = FindK(...)  ! call a function or search all (previous) levels
+      Field(i, j, computed_k_index)  ! then, access a pre-computed absolute index (e.g. boundary layer)
+      ...
+```
+
+In some cases, porting these codes might be possible by using temporary helper fields and by breaking loop apart. In other cases (e.g. looping over previous levels) that might not be possible and the code can't be expressed as a GT4Py stencil.
```
twicki: the previous level loops were always done with while-loops, which was not ideal but worked. 
After some thinking I could not find an example that was not describebale with the old concepts, even though doing so adds a ton of bloat and makes the code unreadable

docs/development/ADRs/cartesian/frontend-indexing-absolute-k.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Decision
+
+We decided to allow absolute indexing on the K axis in certain circumstances with the following syntax:
+
+```py
+@stencil(backend="dace:cpu", externals={K4: "4"})
+def absolute_k_indexing(in_field: Field[np.float64], k_field: Field[IJ, np.int64], out_field: Field[np.float64], idx: int) -> None:
```
twicki: `k_field` is an interesting name for an `FieldIJ`. 
```suggestion
def absolute_k_indexing(in_field: Field[np.float64], index_field: Field[IJ, np.int64], out_field: Field[np.float64], idx: int) -> None:
```

docs/development/ADRs/cartesian/experimental-features.md
twicki, line 1: I am a bit worried that we might get to a state where we keep experimental features experimental forever. 
To prevent that problem: do we want to set ourselves time-lines for making decisions on those to either go in or out? Or do you think that is not needed?

romanc: Thanks for bringing this up. It's a fear I share and I agree, we should guard against it. I'd be against a strict timeline for features to go in or out. Mostly because I fear we (in the current constellation) won't have the resources to move features in (and thus would have to drop them out after the deadline expires). I'd suggest to re-evaluate the state of each experimental feature on a regular basis instead. For example, we could take it up as a discussion point twice a year in / before the gt4py community meeting.

docs/development/ADRs/cartesian/frontend-indexing-absolute-k.md
romanc: yep - let's stick with `i`, `j`, `k` as loop variables and `I`, `J`, `K` as max values. will change

docs/development/ADRs/cartesian/frontend-indexing-absolute-k.md
romanc: Thanks. I'll reword that paragraph.

docs/development/ADRs/cartesian/frontend-indexing-absolute-k.md
romanc: Nice catch!

docs/development/ADRs/cartesian/frontend-indexing-absolute-k.md
```diff
@@ -0,0 +1,79 @@
...
+Users of GT4Py can now succinctly write the patterns shown above in the context section.
+
+One major drawback of absolute indexing is that we loose the opportunity to warn users at compile time if they read out fo bounds. Since we allow absolute K indices that aren't know at compile time, the extent analysis can't warn us anymore. We plan to mitigate that by supporting an option to check every absolute read (at runtime) in debug mode, see references below.
+
+We ship absolute K indexing as an [experimental features](./general-experimental-features.md). The feature is only available in `debug` and `dace:*` backends. Other backends raise an error at compile time when transpiling from `oir` into the respective backend IR.
+
+## References
+
+- [Issue](https://github.com/GridTools/gt4py/issues/1684) to add an out of bounds check at runtime. The issue is old because already now, with `VariableKOffset`, we can't catch every out of bounds access at compile time.
+- TODO [Issue](<>) about unifying access patterns with global tables.
```
romanc: To be changed once that issue is written. /cc @twicki

FlorianDeconinck:
cscs-ci run

PR #2278, opened 2025-09-30
[perf][next][dace] Avoid unnecessary vertical map splitting

iomaganaris:
Avoid splitting the intermediate `AccessNode` between 2 maps if part of the `AccessNode` in between them is accessed by some other `AccessNode` (if there was a map it wouldn't be a problem because all maps would have been split any way).
Below is an example:
`__tmp1` intermediate `AccessNode`
image
Split `Map`s
image
IUUC the second maps are split but because part of `__tmp1` is then copied to `next_w` the fusion between the two `Map`s can't happen so there's no reason to split them in the first place.
An extension to this would be to avoid splitting the part of the `AccessNode` which is copied to the other `AccessNode` but I'm not sure where this is applicable and how to measure it's performance.
Also it might be that this transformation is unnecessary or not applicable tin `vertically_implicit_solver`s after the copies shown above to `next_w` are removed (PR about that incoming).
The improvement for this stencil from this PR is `2,024,564.5ns` vs `1,111,942.0ns`


src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,27 @@ def can_be_applied(
 
         # TODO(phimuell): Implement a check that the array is single use data?
 
+        # Check that the data read by the nested map are not accessed by any other outgoing edge
```
philip-paul-mueller: What do you need with "nested Map"?

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,27 @@ def can_be_applied(
...
+        # of the access node.
+        access_node_out_edges = graph.out_edges(self.access_node)
```
philip-paul-mueller:
```suggestion
        access_node_out_edges = list(graph.out_edges(self.access_node))
```
I am not sure if it will always be a `list` or if they will change it at some point to an iterator.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,27 @@ def can_be_applied(
...
+        access_node_out_edges_data = [edge.data.data for edge in access_node_out_edges]
```
philip-paul-mueller: This is not guaranteed to work, because in an `AccessNode -> AccessNode` situation it is not clear (nor deterministic) to which of the two AC `edge.data.data` is referring to.
If I understand your correctly then your goal is to find all AC that read _directly_ from `self.access_node` (and by that ignoring all AC that are inside a Map).
In that case you have to do something like this:
```suggestion
        access_node_out_edges_data = [
        oedge.dst.data for oedge in access_node_out_edges if (not oedge.data.is_empty()) and isinstance(oedge, dace_nodes.AccessNode)]  
```
However, the above code handles view in the same way as a normal data container.
This might be a desirable behaviour.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,27 @@ def can_be_applied(
...
+        second_map_in_edges = graph.in_edges(self.second_map_entry)
+        matching_data_subsets = {
+            second_map_edge.data.data: second_map_edge.data.subset
```
philip-paul-mueller: OPTIONAL:
```suggestion
            second_map_edge.data.data: second_map_edge.data.src_subset
```
Using subset here is not wrong and I am sure that it would be correct all the times, but using `src_subset` is in view a bit cleaner as it is immediately clear what side you want to access.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,27 @@ def can_be_applied(
...
+            for second_map_edge in second_map_in_edges
+            if second_map_edge.data.data in access_node_out_edges_data
```
philip-paul-mueller:
```suggestion
            if  (not second_map_edge.data.is_empty()) and second_map_edge.data.data in access_node_out_edges_data
```
Empty edges are used to enforce a certain ordering, but they do not transfer any data, this is just a precaution.

As a side note.
Above, when you collected `access_node_out_edges_data`, I mentioned that `edge.data.data` in general does not work for `AC -> AC` connections.
However, in this specific case it is valid, because for syntactical reasons all upstream adjacent nodes to a MapEntry node have to be AccessNodes.
So you could keep it that way.
However, I am not sure if changing it to `edge.src.data` might be more readable, but for me both are fine.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,27 @@ def can_be_applied(
...
+        }
+        for out_edge in access_node_out_edges:
+            if out_edge.dst == self.second_map_entry:
+                continue
+            for matching_data, matching_subset in matching_data_subsets.items():
+                if out_edge.data.data != matching_data:
+                    continue
+                if gtx_dace_split.are_intersecting(
+                    out_edge.data.subset, matching_subset
+                ) and isinstance(out_edge.dst, dace_nodes.AccessNode):
```
philip-paul-mueller: The `isinstance()` check only depends on the outer loop variable so I would pull it out.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,27 @@ def can_be_applied(
...
+        second_map_in_edges = graph.in_edges(self.second_map_entry)
+        matching_data_subsets = {
+            second_map_edge.data.data: second_map_edge.data.subset
+            for second_map_edge in second_map_in_edges
+            if second_map_edge.data.data in access_node_out_edges_data
+        }
+        for out_edge in access_node_out_edges:
+            if out_edge.dst == self.second_map_entry:
+                continue
+            for matching_data, matching_subset in matching_data_subsets.items():
```
philip-paul-mueller: I am thinking that you could condense that as an `if not any(...)` statement.

philip-paul-mueller (changes requested):
My comments are all minor things.

The changes makes sense to me. 
I think the right solution would be to actually run `MapFusionVertical.can_be_applied_to()`.
But as far as I remember we already discussed this and concluded that it is unfeasible to do.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
iomaganaris: This is a typo. I meant to write `second`

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
iomaganaris: > This is not guaranteed to work, because in an AccessNode -> AccessNode situation it is not clear (nor deterministic) to which of the two AC edge.data.data is referring to.

☠️ ☠️

iomaganaris: I've used instead `oedge.src.data` so that they all refer to the same data of the intermediate `AccessNode`

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
iomaganaris: I've removed it from there because we anyway look if the dst node is an `AccessNode` when populating `matching_data_subsets`

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
iomaganaris: I've changed the logic to check for `edge.src.data` which should be an `AccessNode` so I believe the check you mentioned is not applicable any more

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,32 @@ def can_be_applied(
...
+            oedge.src.data
+            for oedge in access_node_out_edges
+            if isinstance(oedge.dst, dace_nodes.AccessNode)
+        ]
+        second_map_in_edges = list(graph.in_edges(self.second_map_entry))
+        matching_data_subsets = {
+            second_map_edge.src.data: second_map_edge.data.src_subset
+            for second_map_edge in second_map_in_edges
+            if second_map_edge.src.data in access_node_out_edges_data
+            and isinstance(second_map_edge.src, dace_nodes.AccessNode)
```
philip-paul-mueller:
```suggestion
            if isinstance(second_map_edge.src, dace_nodes.AccessNode)
            and second_map_edge.src.data in access_node_out_edges_data
```
Technically you also have to check if the `dst_conn` starts with `IN_` because it could also be [Dynamic Map Ranges](https://spcldace.readthedocs.io/en/latest/sdfg/ir.html#parametric-parallelism).
I think it is safe to ignore them, if want to be on the safe side, then check if you have any of them and if so, do not apply.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,32 @@ def can_be_applied(
 
         # TODO(phimuell): Implement a check that the array is single use data?
 
+        # Check that the data read by the second map from the intermediate AccessNode
+        # are not copied to another AccessNode.
+        access_node_out_edges = list(graph.out_edges(self.access_node))
+        access_node_out_edges_data = [
+            oedge.src.data
```
philip-paul-mueller: `oedge` is an edge that has its source at `self.access_node` thus `oedge.src` is always `self.access_node`.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,32 @@ def can_be_applied(
...
+        }
+        if any(
+            out_edge.dst != self.second_map_entry
+            and any(
+                out_edge.src.data == matching_data
+                and gtx_dace_split.are_intersecting(out_edge.data.src_subset, matching_subset)
+                for matching_data, matching_subset in matching_data_subsets.items()
+            )
+            for out_edge in access_node_out_edges
+        ):
```
philip-paul-mueller, line 678:
```suggestion
        if any(
            out_edge.src.data  in matching_data_subsets
            and gtx_dace_split.are_intersecting(out_edge.data.src_subset, matching_data_subsets[out_edge.src.data ])
            for out_edge in access_node_out_edges
            if out_edge.dst != self.second_map_entry
        ):
```
See my comment about `out_edge.src.data` from above too.

philip-paul-mueller (commented):
Some small things.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
iomaganaris: I would have never thought about this..

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
iomaganaris: I cleaned up a bit the logic I think

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_vertical_map_split_fusion.py
```diff
@@ -67,7 +70,18 @@ def test_vertical_map_fusion():
 
     st.add_nedge(C_node, B_node, dace.Memlet(data=C, subset="0", other_subset="0"))
 
+    if extra_intermediate_edge:
+        D_node = st.add_access(D)
+        st.add_nedge(tmp_node, D_node, dace.Memlet(data=tmp, subset="78:80", other_subset="0:2"))
```
edopao:
```suggestion
        assert N >= 2
        st.add_nedge(tmp_node, D_node, dace.Memlet(data=tmp, subset="N-2:N", other_subset="0:2"))
```

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,26 @@ def can_be_applied(
...
+        second_map_in_edges = list(graph.in_edges(self.second_map_entry))
+        access_node_subset_into_second_map = [
+            second_map_edge.data.src_subset
+            for second_map_edge in second_map_in_edges
+            if second_map_edge.src.data == self.access_node.data
+            and second_map_edge.dst_conn.startswith("IN_")
+        ]
+        if any(
+            out_edge.dst != self.second_map_entry
+            and any(
```
philip-paul-mueller:
```suggestion
            any(
```

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_extended.py
```diff
@@ -459,6 +460,26 @@ def can_be_applied(
...
+                gtx_dace_split.are_intersecting(out_edge.data.src_subset, matching_subset)
+                for matching_subset in access_node_subset_into_second_map
+            )
+            for out_edge in access_node_out_edges
```
philip-paul-mueller, line 672:
```suggestion
            for out_edge in access_node_out_edges if out_edge.dst != self.second_map_entry
```

PR #2280, opened 2025-10-02
feat[next][dace]: collect sdfg execution time with tasklet

edopao:
This PR provides an alternative way to retrieve the SDFG execution time, which corresponds to the stencil compute time metric in GT4Py.
The previous solution relied on SDFG instrumentation, which is a complex framework designed to profile full-application SDFGs, to collect different kinds of counters, and to report metrics across several iterations of a benchmark run. This framework already exists in GT4Py, we only need to timestamp the start and end time of the SDFG execution.
The solution presented in this PR is much more lightweight. More importantly, it surpasses some limitation of the dace framework, namely the small precision of the time measurements (dace truncates the SDFG execution time to microseconds).
Another advantage is that the collection of compute time metric, and the corresponding synchronization point for SDFG execution, can be enabled or disabled by means of the GT4Py config option without recompiling the program. This aligns the behavior of the dace backend with the gtfn backend and makes it easier to collect metrics.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
@@ -14,14 +14,17 @@ from typing import Any, Sequence
 import dace
 
 from gt4py._core import definitions as core_defs
-from gt4py.next import common as gtx_common, config, metrics, utils as gtx_utils
+from gt4py.next import common as gtx_common, config, utils as gtx_utils
 from gt4py.next.otf import stages
-from gt4py.next.program_processors.runners.dace import sdfg_callable, workflow as dace_worflow
-from gt4py.next.program_processors.runners.dace.workflow import common as gtx_wfdcommon
+from gt4py.next.program_processors.runners.dace import sdfg_callable
+from gt4py.next.program_processors.runners.dace.workflow import (
+    common as gtx_wfdcommon,
+    compilation as gtx_wfdcompilation,
+)
 
 
 def convert_args(
-    fun: dace_worflow.compilation.CompiledDaceProgram,
+    fun: gtx_wfdcompilation.CompiledDaceProgram,
     device: core_defs.DeviceType = core_defs.DeviceType.CPU,
 ) -> stages.CompiledProgram:
     # We use the callback function provided by the compiled program to update the SDFG arglist.
@@ -48,6 +51,7 @@ def convert_args(
                 *flat_args,
                 filter_args=False,
             )
+            this_call_args[gtx_wfdcommon.SDFG_ARG_METRIC_LEVEL] = config.COLLECT_METRICS_LEVEL
             with dace.config.set_temporary("compiler", "allow_view_arguments", value=True):
                 fun(**this_call_args)
 
@@ -56,30 +60,5 @@ def convert_args(
             #  so we just update it with the current call arguments.
             update_sdfg_call_args(args, fun.sdfg_program._lastargs[0])
             fun.fast_call()
-
-        metric_collection = metrics.get_active_metric_collection()
-        if (metric_collection is not None) and (
-            config.COLLECT_METRICS_LEVEL >= metrics.PERFORMANCE
-        ):
-            # Observe that dace instrumentation adds runtime overhead:
-            # DaCe writes an instrumentation report file for each SDFG run.
-            with gtx_wfdcommon.dace_context(device_type=device):
-                # We need to set the cache folder and key config in order to retrieve
-                # the SDFG report file.
-                prof_report = fun.sdfg_program.sdfg.get_latest_report()
-            if prof_report is None:
-                raise RuntimeError(
-                    "Config 'COLLECT_METRICS_LEVEL' is set but the SDFG profiling report was not found."
-                    " This might indicate that the backend is using a precompiled SDFG from persistent cache"
-                    " without instrumentation."
-                )
-            assert len(prof_report.events) == 1
-            # The event name gets truncated in dace, so we only check that
-            # it corresponds to the beginning of SDFG label.
-            assert f"SDFG {fun.sdfg_program.sdfg.label}".startswith(prof_report.events[0].name)
-            duration_secs = (
-                prof_report.events[0].duration / 1e6
-            )  # dace timer returns the duration in microseconds
-            metric_collection.add_sample(metrics.COMPUTE_METRIC, duration_secs)
 
     return decorated_program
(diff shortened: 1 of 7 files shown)


edopao:
Not a high priority task, I'll abandon this change.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -9,31 +9,52 @@
...
+        result = fun(*args, **kwargs)
+        assert len(result) == 1
+        assert isinstance(result[0], np.float64)
+        if not collect_time:
+            return
+        metric_collection = metrics.get_active_metric_collection()
+        if metric_collection is not None:
+            metric_collection.add_sample(metrics.COMPUTE_METRIC, result[0].item())
+
+    return inner
```
havogt: I think implementing this as a decorator is a bit confusing, because it's not a decorator that can be applied on any callable, but probably can decorate only exactly one function. Therefore, the more readable implementation would probably be to call `decorated_program` directly from the collect compute time function. Does it make sense?

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -113,69 +113,125 @@ def _has_gpu_schedule(sdfg: dace.SDFG) -> bool:
...
+        assert dace_gpu_backend in ["cuda", "hip"], f"GPU backend '{dace_gpu_backend}' is unknown."
 
-    assert dace.Config.get("compiler.cuda.max_concurrent_streams") == -1, (
-        f"Expected `max_concurrent_streams == -1` but it was `{dace.Config.get('compiler.cuda.max_concurrent_streams')}`."
-    )
+        # NOTE: We should actually wrap the `StreamSynchronize` function inside a
+        #   `DACE_GPU_CHECK()` macro. However, this only works in GPU context, but
+        #   here we are in CPU context. Thus we cannot do it.
+        sync_code = f"{dace_gpu_backend}StreamSynchronize({dace_gpu_backend}StreamDefault);"
+        has_side_effcts = True
```
havogt:
```suggestion
        has_side_effects = True
```

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -113,69 +113,125 @@ def _has_gpu_schedule(sdfg: dace.SDFG) -> bool:
...
+        dace_gpu_backend = dace.Config.get("compiler.cuda.backend")
+        assert dace_gpu_backend in ["cuda", "hip"], f"GPU backend '{dace_gpu_backend}' is unknown."
 
-    assert dace.Config.get("compiler.cuda.max_concurrent_streams") == -1, (
-        f"Expected `max_concurrent_streams == -1` but it was `{dace.Config.get('compiler.cuda.max_concurrent_streams')}`."
-    )
+        # NOTE: We should actually wrap the `StreamSynchronize` function inside a
+        #   `DACE_GPU_CHECK()` macro. However, this only works in GPU context, but
+        #   here we are in CPU context. Thus we cannot do it.
+        sync_code = f"{dace_gpu_backend}StreamSynchronize({dace_gpu_backend}StreamDefault);"
```
havogt: For a single stream is there a difference between `cudaStreamSynchronize` and `cudaDeviceSynchronize`? If not, we can replace this by `cudaDeviceSynchronize` and remove the checks that we have a single stream.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -113,69 +113,125 @@ def _has_gpu_schedule(sdfg: dace.SDFG) -> bool:
...
+        "gt_stop_timer",
+        inputs={"run_cpp_start_time"},
+        outputs={"duration"},
+        code=sync_code
+        + """
+double run_cpp_end_time = static_cast<double>(
+    std::chrono::duration_cast<std::chrono::nanoseconds>(
+        std::chrono::high_resolution_clock::now().time_since_epoch()).count()
+) / 1e9;
+duration = run_cpp_end_time - run_cpp_start_time;
```
havogt, line 209: for my understanding: in the generated code `duration` is a parameter of some c++ function because it's in `outputs`? Similarly, the `time` from line 184 is output and gets passed as input to this piece?

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
edopao: I agree, there should not be much difference. Besides, if we use `cudaDeviceSynchronize` this function would be generic, not limited to default stream only.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
edopao: More precisely:

`duration` is the name of the output connector of this tasklet (as defined on line 214). The output of the tasklet, on this connector, is written by an edge to a data node `__return` (see line 234). `__return` is a global parameter (the direction as output is only given by the fact that `__return` is a reserved symbol for output data, in dace).

`time` is the name of an output connector of the first tasklet. The edge on line 191 writes to a temporary scalar node.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
edopao: Yes, it makes sense. The decorator served also to retrieve `config.COLLECT_METRICS_LEVEL` only one time, when the function is decorated, but I can achieve the same in the way you suggest.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -113,6 +113,116 @@ def _has_gpu_schedule(sdfg: dace.SDFG) -> bool:
...
+    time, it does not include the overhead of calling the SDFG from Python.
+
+    The execution time is measured in seconds and represented as a 'float64' value.
+    It is returned from the SDFG as a one-element array in the '__return' data node.
+    """
+    output, _ = sdfg.add_array("__return", [1], dace.float64)
+    start_time, _ = sdfg.add_scalar("gt_start_time", dace.float64, transient=True)
+    metrics_level = sdfg.add_symbol(gtx_wfdcommon.SDFG_ARG_METRIC_LEVEL, dace.int32)
+
+    #### 1. Set CUDA stram to default and create synchronization code for GPU target.
```
edopao:
```suggestion
    #### 1. Synchronize the CUDA device, in order to wait for kernels completion.
```

havogt (approved):
lgtm

PR #2283, opened 2025-10-06
feat[dace][next]: Added Inline Fuser

philip-paul-mueller:
This PR adds an "Inline Fuser" functionality to the transformations.
This functionality is related to classical MapFusionVertical, however, while in the classical case the two Maps have to be compatible, i.e. the data generated by the first Map must be consumed in the _same_ iteration by the second Map, the inline fuser is much more flexible, as it essentially removes the word _same_ from the definition.
An extended description can be found here: https://hackmd.io/ot702TuJT7OJXp07WveiPA


tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_inline_fuser.py
```diff
@@ -0,0 +1,359 @@
...
+        output_nodes={c},
+        external_edges=True,
+    )
+    sdfg.validate()
+
+    return sdfg, state, me2, a, b, c, next(iter(state.out_edges(me2)))
+
+
+def test_inline_fuse_simple_case():
+    sdfg, state, second_map_entry, a, b, c, edge_to_replace = _create_simple_fusable_sdfg()
```
edopao, line 82: If we happen to apply InlineFusion to such a trivial case, instead of VerticalMapFusion, we should probably issue a warning. Don't you think?

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_inline_fuser.py
```diff
@@ -0,0 +1,359 @@
...
+        "laplap",
+        shape=(6,),
+        dtype=dace.float64,
+        transient=False,
+    )
+    lap = state.add_access("lap")
+
+    state.add_mapped_tasklet(
+        "lap",
+        map_ranges={"__i": "0:8"},
```
edopao:
```suggestion
        map_ranges={"__i": "1:9"},
```
In this way we mimic a more common case in GTIR, where the domain of the temporary does not start from zero.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_inline_fuser.py
```diff
@@ -0,0 +1,359 @@
...
+            "__in2": dace.Memlet("a[__i + 2]"),
+        },
+        code="__out = __in2 - __in1",
+        outputs={"__out": dace.Memlet("lap[__i]")},
+        output_nodes={lap},
+        external_edges=True,
+    )
+    laplap_tlet, laplap_me, _ = state.add_mapped_tasklet(
+        "laplap",
+        map_ranges={"__i": "0:6"},
```
edopao:
```suggestion
        map_ranges={"__i": "1:7"},
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+) -> Optional[tuple[dace_nodes.NestedSDFG, dace_nodes.AccessNode]]:
+    """Tries to perform the inlining of `edge`.
+
+    The function replaces the transmission, i.e. reading, of data through `edge`
+    with an on-the-fly computation of the same data. The term inlining is used
+    because the dataflow, that computes the transmitted data is embedded/inlined
+    into the scope of `edge`.
+
+    This function performs inlining in a single step or returns `None` if that
+    is not possible. It is also possible the perform these two steps, determining
```
edopao:
```suggestion
    is not possible. It is also possible to perform these two steps separately, determining
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    state: dace.SDFGState,
+    edge: dace_graph.MultiConnectorEdge[dace.Memlet],
+) -> Optional[tuple[set[dace_nodes.Node], InlineSpec]]:
+    """First step of dataflow inlining, computing the inline specification.
+
+    The inline specification describes how the inlining of dataflow has to be done.
+    The values computed by this function must be passed to `perform_dataflow_inlining()`
+    for actually inlining dataflow.
+    The function will analyze how a read through `edge` can be replaced by direct
+    computations. If this is not possible then the function return `None`.
```
edopao:
```suggestion
    computations. If this is not possible then the function returns `None`.
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+
+    Currently the following restriction (not exhaustive) are imposed on the dataflow:
```
edopao:
```suggestion
    Currently the following restrictions (not exhaustive) are imposed on the dataflow:
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    - `edge` must be an outgoing edge of a top level MapEntry.
+    - The generating dataflow, i.e. the one that is inlined, must be inside a Map.
+    - The intermediate between the two Maps can only have one producer, i.e. it must
+        be single use data (not checked) and can only have an incoming degree of 1.
```
edopao: Why it is not checked here? In this way, we are checking the conditions partially here and partially on the caller side.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+        return None
+    if edge.data.wcr is not None:
+        return None
+
+    # We require that every access in the edge is either a literal digit, such as
+    #  `1` but also `1:3` or that exactly one symbol that is also map parameter is
+    #  used to do the index. Note that this does allow accesses such as
+    #  `a[map_index:(map_index + 4)]` but not accesses such as `a[map_index:(map_index + symb)]`
+    #  or `a[some_other_symbol]`. Furthermore, we require that all parameters of the
+    #  second Map are used for the access (exception see bellow). These rule are used
```
edopao:
```suggestion
    #  second Map are used for the access (exception see below). These rule are used
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    #  to ensure that one iteration of the first Map are enough to satisfy all data
```
edopao:
```suggestion
    #  to ensure that one iteration of the first Map is enough to satisfy all data
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    #  dependencies of the second Map.
+    avial_second_map_param = {param for param in second_map_entry.map.params}
```
edopao:
```suggestion
    avail_second_map_param = {param for param in second_map_entry.map.params}
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    full_symbolic_access = True
+    full_symbolic_access_dims: list[int] = []
+    for dim, (start, stop, step) in enumerate(edge.data.src_subset):
+        if (step != 1) == True:  # noqa: E712 [true-false-comparison]  # SymPy comparison
+            return None
+
+        if str(start).isdigit():
+            # The start index is a digit, we require that the end is also a digit.
```
edopao, line 196: I initally had some doubts about this restriction (fully symbolic or fully literal, not mixed) but it should probably be OK. The interpolation formulation on K half levels should still work `[K, K+1]`. I guess that this was the main case you are addressing in diffusion, the writes to `w`.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    # Usually we require that all parameters of the second Map are used for the
+    #  access. The main reason is to prevent some slicing, i.e. to ensure that
+    #  the access is not done later. However, this prevents the second Map to read
+    #  from a Map that has one dimension less. Thus as an exception we allow that
+    #  in case all accesses are done using parameters, then not all of them have
+    #  to be used.
+    if (not full_symbolic_access) and (len(avial_second_map_param) != 0):
+        return None
+
+    # We restrict us to them because these node types are kind of simple to modify.
```
edopao: We restrict us to them

->

We restrict the destination nodes to taklets and access nodes

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    #  However, even the Tasklet could be hard to handle. The hardest is possible
+    #  the nested SDFG. Which is possible to handle in some cases but not in all
+    #  cases.
```
edopao: Rewrite:
The hardest is possible the nested SDFG. Which is possible to handle in some cases but not in all cases.

to something like:
The hardest case is the nested SDFG, which is possible to handle under certain conditions but not in all cases.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    # TODO(phimuell): Allow more and allow NestedSDFGs in certain cases.
+    for leave_edge in state.memlet_tree(edge).leaves():
+        if not isinstance(leave_edge.dst, (dace_nodes.AccessNode, dace_nodes.Tasklet)):
+            return None
+
+    # There must be an intermediate that is only generated by a single source.
+    #  NOTE: We omit the check that the intermediate is single use data here.
+    intermediate_node: dace_nodes.Node = next(
+        iter(state.in_edges_by_connector(edge.src, "IN_" + edge.src_conn[4:]))
```
edopao, line 235: Maybe add a note that 4: is to exclude `OUT_` from the connector name on the `MapEntry` node.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    )
+
+    # Check the data dependencies. It is a little bit different compared to classical
+    #  MapFusionVertical, where the two iterations are coupled. Here the restriction
+    #  is that there exists _one_ iteration of the first Map that, generates the data
+    #  that _another_ iteration of the second Map needs. This is a much weaker restriction
+    #  but also harder to check. As a first approximation we look at the shape of the
+    #  data that is generated and consumed.
+    #  To avoid dynamic allocation we require that the exchange set has a constant size.
+    # TODO(phimuell): Improve that.
```
edopao: This TODO is too generic, in my opinion. Either there is something clear that could be implemented as improvement, or we can skip the comment.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    consumer_shape = consumer_subset.size()
+    producer_subset = writing_edges[0].data.get_dst_subset(writing_edges[0], state)
+    producer_shape = producer_subset.size()
+    for cshp, pshp in zip(consumer_shape, producer_shape, strict=True):
+        if not str(pshp).isdigit():
+            return None
+        if (cshp == pshp) == True:  # noqa: E712 [true-false-comparison]  # SymPy comparison
+            continue
+        if (cshp < pshp) == True:  # noqa: E712 [true-false-comparison]  # SymPy comparison
+            continue
```
edopao:
```suggestion
        if (cshp <= pshp) == True:  # noqa: E712 [true-false-comparison]  # SymPy comparison
            continue
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+        return None
+
+    # If a dimension of the intermediate is accessed not by a Map parameter, we require
+    #  that that range is known at generation time, see above. In addition to that we
+    #  require that the range start at zero. The only reason for that is that it makes
```
edopao:
```suggestion
    #  require that the range starts at zero. The only reason for that is that it makes
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+            return None
+
+        if str(start).isdigit():
+            # The start index is a digit, we require that the end is also a digit.
+            if not str(stop).isdigit():
+                return None
+
+            # There is not a full symbolic access.
+            full_symbolic_access = False
+            full_symbolic_access_dims.append(dim)
```
edopao: it seems that `full_symbolic_access_dims` contains the dimensions that are not symbolically accessed, so I would call it `literal_index_access_dims`.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+        if (cshp == pshp) == True:  # noqa: E712 [true-false-comparison]  # SymPy comparison
+            continue
+        if (cshp < pshp) == True:  # noqa: E712 [true-false-comparison]  # SymPy comparison
+            continue
+        return None
+
+    # If a dimension of the intermediate is accessed not by a Map parameter, we require
+    #  that that range is known at generation time, see above. In addition to that we
+    #  require that the range start at zero. The only reason for that is that it makes
+    #  computing the correction offset much simpler.
```
edopao, line 297: Maybe here is a good place for a `TODO: allow non-zero start index for intermediate access`

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+        start=first_map_exit,
+        state=state,
+        start_connector="IN_" + producing_edge.src_conn[4:],
+        limit_node=state.entry_node(first_map_exit),
+    )
+    assert first_map_exit not in generating_nodes
+    assert state.entry_node(first_map_exit) not in generating_nodes
+    assert len(generating_nodes) >= 1
+
+    # Fore syntactical reasons we require that all outgoing edges of the nodes in
```
edopao:
```suggestion
    # For syntactical reasons we require that all outgoing edges of the nodes in
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    sdfg: dace.SDFG,
+    state: dace.SDFGState,
+    edge: dace_graph.MultiConnectorEdge[dace.Memlet],
+    nodes_to_inline: set[dace_nodes.Node],
+    inline_spec: InlineSpec,
+) -> Optional[tuple[dace_nodes.NestedSDFG, dace_nodes.AccessNode]]:
+    """Performs the second step, i.e. the actual inlining, of the dataflow.
+
+    The function will use the information of `nodes_to_inline` and `inline_spec`,
+    which was computed my `find_nodes_to_inline()`, to perform the inlining. The
```
edopao:
```suggestion
    which was computed by `find_nodes_to_inline()`, to perform the inlining. The
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    inlining essentially works by created a nested SDFG inside the Map, where `edge`
```
edopao:
```suggestion
    inlining essentially works by creating a nested SDFG inside the Map, where `edge`
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+            if outside_input_node not in replicated_node_map:
+                if outside_input_node.data not in nsdfg.arrays:
+                    # We have not yet mapped this data into the nested SDFG. For that
+                    #  we use the same descriptor as on the outside, but we have to
+                    #  mark it as non-transient.
+                    nested_input_desc = outside_input_node.desc(sdfg).clone()
+                    nested_input_desc.transient = False
+
+                    if isinstance(nested_input_desc, dace_data.View):
+                        raise NotImplementedError("Mapping in of views is not supported.")
```
edopao:
```suggestion
                        raise NotImplementedError("Mapping of views is not supported.")
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+                    replicated_node_map[outside_input_node] = nstate.add_access(
+                        outside_input_node.data
+                    )
+                    input_node_map[outside_input_node.data] = outside_input_node
+
+                else:
+                    # We have not found this AccessNode yet but the data it refers
+                    #  to has been found. In that case we reuse the AccessNode on
+                    #  the inside we have already created.
+                    # NOTE: That this will slightly change things as before, on the
```
edopao:
```suggestion
                    # NOTE: That this will slightly change the dataflow, on the
```

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+            if src_node is not first_map_entry:
+                # This is a connection from a node that is also replicated into the
+                #  nested SDFG. We will handle it later during the output nodes, thus
+                #  there is nothing to do.
+                assert src_node in nodes_to_replicate
+                continue
+
+            # If we are here then we have a connection from the MapEntry node. Thus
+            #  we have to read from outside data.
+            outside_input_node: dace_nodes.AccessNode = state.memlet_path(iedge)[0].src
```
edopao: `outside_input_node` or `original_input_node`

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+                replicated_node_map[outside_input_node],
+                None,
+                replicated_node_map[iedge.dst],
+                iedge.dst_conn,
+                dace.Memlet.from_memlet(iedge.data),
+            )
+
+        # Recreate the connections between the nodes that we have replicated.
+        #  It is important that above we only handled the connections that were now
+        #  because they were coming from the MapEntry node of the first Map.
```
edopao: Should we rewrite the second sentence as below?
`It is important to note that above we only handled the connections that were coming from the MapEntry node of the first Map.`

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+    }
+
+    nsdfg_node = state.add_nested_sdfg(
+        sdfg=nsdfg,
+        inputs=input_node_map.keys(),
+        outputs={output_name},
+        symbol_mapping=first_map_param_mapping,
+    )
+
+    for input_name, requiered_node in input_node_map.items():
```
edopao:
```suggestion
    for input_name, required_node in input_node_map.items():
```

edopao (commented):
LGTM, I only have some minor comments.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_inline_fuser.py
philip-paul-mueller: The first thing is that I need a definition of "simple".
I implemented it as a free function, such that the user has to decide when to apply, thus it is technically his task of ensuring that the case is not "simple".

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
philip-paul-mueller: Actually the description is wrong, thanks for spotting it.
What I wanted to express is that the data is only generated by a single producer, "single use data" is irrelevant for that.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
philip-paul-mueller: I also think that it works, if not we will modify this check.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
philip-paul-mueller: I removed it, but updated the TODO.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
philip-paul-mueller: I would have to put then everywhere, which is kind of overblown.
However, it never shows up out of the blue, but always with `"IN_" + ...[4:]` I think this should be hint enough , do you agree?

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
philip-paul-mueller: you are right, I skipped it.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
philip-paul-mueller: You are right my name was even wrong.
I think I selected it based on the `full_symbolic_access` variable.
I also decided to remove the `full_symbolic_access` variable because it is redundant.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
philip-paul-mueller: I do not understand this comment.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
philip-paul-mueller: You are right the explanation was really strange.
I tried to improve it but I am not sure if it is better.
Note that I have swapped the order of incoming and outgoing edge, so you will have a huge diff. However, the code has not changed only the comment has, because it is more logical that way.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_inline_fuser.py
edopao: OK, I get it. We will have to call this transformation at the proper stage in the `auto_optimize` pipeline.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
edopao: Maybe we could use `replace` instead, but as you prefer.
```suggestion
        iter(state.in_edges_by_connector(edge.src, edge.src_conn.replace("OUT_", "IN_", 1))
```

edopao (approved):

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
edopao: It was just a suggestion for a different name, but I am OK with `outside_input_node`.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
```diff
@@ -0,0 +1,747 @@
...
+            nstate.add_edge(
+                replicated_node_map[outside_input_node],
+                None,
+                replicated_node_map[iedge.dst],
+                iedge.dst_conn,
+                dace.Memlet.from_memlet(iedge.data),
+            )
+
+        # Recreate the connections between the nodes that we have replicated.
+        #  It is important that above we only handled the connections that were now
```
edopao: "that were now" still unclear

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_inline_fuser.py
philip-paul-mueller: There is not so much differences as the GT4Py way is just adding some shifting to all accesses and we already have shifting in that test.
But I will look into it.

src/gt4py/next/program_processors/runners/dace/transformations/inline_fuser.py
philip-paul-mueller: If we would use `replace()` then we would replace every occurrence of `IN_` inside the string.
Although, it is very unlikely that we encounter that case, it is also not absolutely unlikely.

edopao: I added the `count=1` argument in the call to `replace()`.

philip-paul-mueller: I do not like it because it is not the canonical way of doing it.

PR #2284, opened 2025-10-06
perf[dace][next]: Updated Top Level Dataflow Optimization Stage

philip-paul-mueller:
This PR slightly modifies how the "top level dataflow optimization" phase of the optimizer works, regarding `MapFusionHorizontal`.
While horizontal Map fusion removes Maps/kernel launches, it also creates dependencies (kind of) between unrelated Maps.
An example for this behaviour can be found in `compute_advection_in_horizontal_momentum_equation`, where the following result is created:

2025-10-06-123835_1193x592_scrot

As it can be seen, this 1D Map is the result of the merging of multiple other (very simple) Maps into one.
It is important to note that each of the four different outputs is consumed by a different 2D Map.
It would be desirable if `MapPromoter` would have merged the individual 1D Maps into the consumer, thus removing a kernel launch and and the need to allocate four 1D transients.

This PR solves this by moving horizontal Map fusion after Map promotion is called.

NOTE:
This PR also tried to change the `only_if_common_ancestor` flag of `MapFusionHorizontal` from `False` (i.e. fuse any possible two Maps together) to `True` (only fuse them if they read from the same data).
However, this lead to very obscure performance regressions and was hence abandoned.

Should only be merged once [PR#2294](https://github.com/GridTools/gt4py/pull/2294) and [PR#2304](https://github.com/GridTools/gt4py/pull/2304) has been merged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ dace-cartesian = [
   'dace>=1.0.2,<2'  # renfined in [tool.uv.sources]
 ]
 dace-next = [
-  'dace==2025.10.23'  # refined in [tool.uv.sources]
+  'dace==2025.10.28'  # refined in [tool.uv.sources]
 ]
 dev = [
   {include-group = 'build'},
@@ -453,7 +453,7 @@ url = 'https://test.pypi.org/simple'
 atlas4py = {index = "test.pypi"}
 dace = [
   {git = "https://github.com/GridTools/dace", branch = "romanc/stree-roundtrip", group = "dace-cartesian"},
-  {git = "https://github.com/GridTools/dace", tag = "__gt4py-next-integration_2025_10_23", group = "dace-next"}
+  {git = "https://github.com/GridTools/dace", tag = "__gt4py-next-integration_2025_10_28", group = "dace-next"}
 ]
 
 # -- versioningit --
--- a/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
@@ -528,13 +528,12 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
     four iteration in the second dimension, will get a block size of `(32, 4, 1)`.
     Note that this modification will not influence the launch bound value.
 
-    The block size and launch bounrds are set according to the aadditional following rules:
-        - Determine the number of map parameters that have range > 1.
-        - If there is only one such parameter then handle the map as 1D map,
-          else if there are more than 2 parameters with range > 1 then handle it as ordinary map.
-        - For 1D maps set the block size such that only the dimension with the largest
-          range is set to the specified block size, all other dimensions are set to 1.
-        - If launch_bounds_1d is not set then set the launch_bounds of `scan` maps to 512 to limit their register usage.
+    Note there are some special rules:
+    - If a Map with (at most 3 dimension) has only one non trivial dimension, i.e.
+        > 1, then the Map is handled as a 1D Map.
+    - If there is a 1D Map that contains a nested SDFG whose label starts with `scan_`
+        and `launch_bounds_1d` was not specified, then a value of 512 is assumed to
+        limit register usage.
 
     Args:
         block_size_Xd: The size of a thread block on the GPU for `X` dimensional maps.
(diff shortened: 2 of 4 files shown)


edopao (commented):
This sounds like the right thing to do. Have you measured the speedup?

philip-paul-mueller:
I agree we should wait and I have not measured yet, but I think that MapPromotion is quite important, I remember this stencil where for some reason MapPromotion no longer worked and the performance penalty was quite significant, but it would be interesting to know.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -275,26 +276,11 @@ def _gt_auto_process_top_level_maps(
         #   has been solved.
         vertical_map_fusion._single_use_data = single_use_data
 
-        horizontal_map_fusion = gtx_transformations.MapFusionHorizontal(
-            only_toplevel_maps=True,
-            consolidate_edges_only_if_not_extending=True,
-            # TODO(phimuell): Should we enable this flag?
-            only_if_common_ancestor=False,
```
iomaganaris, line 465: Do you think it would make sense to keep this here with `only_if_common_ancestor=True` option?

philip-paul-mueller: I am actually not sure, I tried it and setting it to `True` and in this case the result is the same as with the PR.
However, I am not sure if "reads the same field" really implies "they are related" as this ignores how the output is used.
I agree from an optimization point of view it might make sense to fuse them, because potentially you might have to load something once.
But I am not sure if you get any advantages when you have the fuser inside the loop compared to having it at the end?

philip-paul-mueller:
I made some measurements where I compared the current `main` and the proposed solution in the PR.
```
MAIN:
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00023434 | 0.00023437 | 0.00000165 | 1363
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00023362 | 0.00023365 | 0.00000174 | 1346
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00024135 | 0.00024128 | 0.00000148 | 1337
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00023483 | 0.00023484 | 0.00000171 | 1339
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00024016 | 0.00024009 | 0.00000163 | 1301
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00023612 | 0.00023627 | 0.00000191 | 1333

PROPOSAL:
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00023984 | 0.00023985 | 0.00000163 | 1282
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00023129 | 0.00023127 | 0.00000134 | 1317
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00022165 | 0.00022149 | 0.00000141 | 1380
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00023719 | 0.00023699 | 0.00000143 | 1374
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00022716 | 0.00022697 | 0.00000168 | 1340
FusedVelocityAdvectionStencilsHMomentum[compile_time_domain-apply_extra_diffusion_on_vn[True]] | 0.00023482 | 0.00023484 | 0.00000150 | 1343
```

The results are not very conclusive, but I think it is fair to say that the proposed solution is not worse than the current one and might be better in some cases.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -308,36 +309,17 @@ def _gt_auto_process_top_level_maps(
...
-        #  on the horizontal are very complicated, it might degrade performance
-        #  but such cases should be rare.
-        # TODO(phimuell): Add a criteria to decide if we should promote or not.
-        # TODO(phimuell): Think about relocating this before horizontal Map fusion.
+        # Promote Maps. This will remove transients between 1D and 2D Maps, at the
+        #  code of more data loads from memory. Empirical observations have shown
+        #  that this is beneficial; Especially for Nabla4 type kernel in conjunction
+        #  with `LoopBlocking`.
+        # NOTE: We have to promote before we do horizontal Map fusion.
         sdfg.apply_transformations_repeated(
```
iomaganaris, line 473: Wouldn't it make sense to move this before the vertical map fusion to fuse the promoted maps with their dependent maps?

philip-paul-mueller: The promoter already fuses the Maps together.
However, according to Edoardo, the results are now even worse.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -458,39 +459,17 @@ def _gt_auto_process_top_level_maps(
...
-        #  importantly the split transformations run.
-        #  We allow promotion of vertical (which is beneficial for Nabla4 type
-        #  stencils and if `LoopBlocking` is enabled. However, we also allow
-        #  horizontal promotion, this is an empirical choice. If the computation
-        #  on the horizontal are very complicated, it might degrade performance
-        #  but such cases should be rare.
-        # TODO(phimuell): Add a criteria to decide if we should promote or not.
-        # TODO(phimuell): Think about relocating this before horizontal Map fusion.
+        # Promote Maps. This will remove transients between 1D and 2D Maps, at the
+        #  code of more data loads from memory. Empirical observations have shown
```
```suggestion
        #  cost of more data loads from memory. Empirical observations have shown
```

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -458,39 +459,17 @@ def _gt_auto_process_top_level_maps(
...
+        #  that this is beneficial; Especially for Nabla4 type kernel in conjunction
```
```suggestion
        #  that this is beneficial; especially for Nabla4-type kernel in conjunction
```

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -507,6 +486,31 @@ def _gt_auto_process_top_level_maps(
             validate_all=validate_all,
         )
 
+        # Now run horizontal Map fusion, also run vertical Map fusion, because it
+        #  might have become possible.
+        # TODO(phimuell): Think of moving horizontal Map fusion after the splitting.
```
edopao, line 491: It should be pretty simple to try the idea in this TODO. Can you try it and if it is not good, remove this TODO?

philip-paul-mueller: I think we should move this experiment into another PR.

edopao: OK for me.

edopao (approved):
LGTM

PR #2232, opened 2025-09-01
feat[next]: adding custom named collections of scalars and fields

egparedes:
Adding support for typed namedtuples and Dataclasses as valid structures for scalar and Fields. Changes initially prototyped in #2211

PR #2285, opened 2025-10-06
feat[next]: upgrade tree_map with custom constructor per original type

havogt:
Preparation for #2232 to allow constructing different containers in tree_map.


tehrengruber (approved):
I have the feeling that the curried interface will not be very nice when used in anything but the simplest cases. This
```
lambda value: tuple if isinstance(value, list) else list
```
is certainly nicer than
```
lambda value, elts: tuple(elts) if isinstance(value, list) else list(elts)
```
but for more complicated cases I would first consider writing something like this
```
def collection_maker(original_collection, elements):
  if isinstance(original_collection, ...):
     # do something with elements
  elif isinstance(original_collection, ...):
     # do something with elements
  ...
```
not
```
def collection_maker(original_collection):
  if isinstance(original_collection, ...):
     def impl(elements):
       # do something with elements
    return impl
  elif isinstance(original_collection, ...):
     def impl(elements):
       # do something with elements
    return impl
  ...
```
The two argument version could also be done similar to `with_path_arg` -> `result_constructor_original_collection_arg` (or a less clumsy name) making the argument optional thus not needing to change any other usage of `result_collection_constructor`.

PS: This is a program of python syntax not currying as a concept.

PR #2286, opened 2025-10-06
feat[next]: Prune empty concat where branches pass

tehrengruber:


src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
```diff
@@ -0,0 +1,59 @@
...
+def _filter_domain(domain: domain_utils.SymbolicDomain, dims: KeysView[common.Dimension]):
+    return domain_utils.SymbolicDomain(
+        grid_type=domain.grid_type,
+        ranges={d: r for d, r in domain.ranges.items() if d in dims},
+    )
+
+
+@dataclasses.dataclass
+class _PruneEmptyConcatWhere(PreserveLocationVisitor, NodeTranslator):
+    PRESERVED_ANNEX_ATTRS = ("domain",)
```
tehrengruber, line 46: Docstring still missing + Add comment why concat_where tuples are not allowed.

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -184,5 +186,8 @@ def apply_fieldview_transforms(
     ir = ConstantFolding.apply(ir)  # type: ignore[assignment]  # always an itir.Program
 
     ir = infer_domain.infer_program(ir, offset_provider=offset_provider)
+    # required for prune_empty_concat_where pass
+    ir = concat_where.expand_tuple_args(ir, offset_provider_type=offset_provider_type)  # type: ignore[assignment]  # always an itir.Program
```
tehrengruber: @edopao This is required for the case `concat_where(cond, some_tuple, some_other_tuple)` and might, but most likely doesn't, have performance implications. You might want to simplify the lowering since after this PR the dace backend will only ever get non-tuple `concat_where`s.

edopao: I've taken note, thank you.

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -53,6 +53,17 @@ def __post_init__(self) -> None:
     def translate(self, distance: int) -> SymbolicRange:
         return SymbolicRange(im.plus(self.start, distance), im.plus(self.stop, distance))
 
+    def empty(self) -> Optional[bool]:
```
egparedes:
```suggestion
    def empty(self) -> bool | None:
```

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -53,6 +53,17 @@ def __post_init__(self) -> None:
...
+        if isinstance(self.start, itir.Literal) and isinstance(self.stop, itir.Literal):
+            start, stop = int(self.start.value), int(self.stop.value)
+            if start >= stop:
+                return True
+            else:
+                return False
```
egparedes:
```suggestion
            return start >= stop
```

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -68,6 +79,13 @@ class SymbolicDomain:
     def __hash__(self) -> int:
         return hash((self.grid_type, frozenset(self.ranges.items())))
 
+    def empty(self) -> Optional[bool]:
```
egparedes:
```suggestion
    def empty(self) -> bool | None:
```

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
```diff
@@ -0,0 +1,61 @@
...
+import dataclasses
+from typing import KeysView
+
+from gt4py.eve import NodeTranslator, PreserveLocationVisitor
+from gt4py.next import common
+from gt4py.next.iterator import ir as itir
+from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm, domain_utils
+
+
+def _filter_domain(domain: domain_utils.SymbolicDomain, dims: KeysView[common.Dimension]):
```
egparedes: Missing return type annotation (Why is mypy not complaining here??)

egparedes: Why not just
```suggestion
def _filter_domain(domain: domain_utils.SymbolicDomain, dims: set[common.Dimension]):
```
or even
```suggestion
def _filter_domain(domain: domain_utils.SymbolicDomain, dims: Container[common.Dimension]):
```
??

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@dataclasses.dataclass
+class _PruneEmptyConcatWhere(PreserveLocationVisitor, NodeTranslator):
+    """ """
+
+    PRESERVED_ANNEX_ATTRS = ("domain",)
+
+    @classmethod
+    def apply(cls, node: itir.Node):
```
egparedes:
```suggestion
    def apply(cls: type[Self], node: itir.Node) -> Self:
```

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
```diff
@@ -0,0 +1,61 @@
...
+            cond_expr, tb, fb = node.args
+
+            if tb == fb:
+                # note: as long as we visited the args we have a copy here, so no need to copy again
+                tb.annex.domain = node.annex.domain
+                return tb
+
+            cond = domain_utils.SymbolicDomain.from_expr(cond_expr)
+            if cond.empty():
+                return node.args[2]
```
egparedes, line 65: Just a question: why is it returned the false branch in this case?
```suggestion
                return node.args[2]
```

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_prune_empty_concat_where.py
```diff
@@ -0,0 +1,69 @@
...
+from gt4py.next.iterator.ir_utils import domain_utils
+
+Vertex = common.Dimension(value="Vertex", kind=common.DimensionKind.HORIZONTAL)
+K = common.Dimension(value="K", kind=common.DimensionKind.VERTICAL)
+
+
+@pytest.mark.parametrize(
+    "accessed_domain, cond_domain, expected",
+    [
+        # TODO(tehrengruber): Implement and enable symbolic test cases.
```
egparedes: Does the TODO mean to enable the commented out cases?

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
tehrengruber: If in a `concat_where(cond, tb, fb)` the `cond` is empty, none of the values come from `tb`, but `fb` is always taken.

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
tehrengruber: I think because it is a private function, but not sure.

tehrengruber: The result of `Dict.keys()` is not a set, but only a `KeyView`. `Container` is better though.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_prune_empty_concat_where.py
tehrengruber: Yes, I improved the comment.

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
egparedes: FYI, key views are actually sets, in the sense they implement the `collections.abc.Set` interface (check here: https://devdocs.io/python~3.13/library/stdtypes#dict-views)

tehrengruber: Not sure why mypy doesn't accept them them.

PR #2287, opened 2025-10-07
feat[next]: flexible metric collections

egparedes:
This PR extends the `metrics` collection mechanism to track as different metric sources  the different compiled variants of a same program.

The main change in design is that metrics sources can contain any kind of extra metadata and can collect metrics samples before they have a properly assigned key for the global store of metrics sources. This means that the key is not always defined in the outermost level of a program callable, but in the innermost.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/__init__.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/__init__.py
@@ -21,6 +21,7 @@ from .gpu_utils import (
     gt_gpu_transformation,
     gt_set_gpu_blocksize,
 )
+from .inline_fuser import inline_dataflow_into_map
 from .local_double_buffering import gt_create_local_double_buffering
 from .loop_blocking import LoopBlocking
 from .map_fusion import MapFusionHorizontal, MapFusionVertical
@@ -103,5 +104,6 @@ __all__ = [
     "gt_split_access_nodes",
     "gt_substitute_compiletime_symbols",
     "gt_vertical_map_split_fusion",
+    "inline_dataflow_into_map",
     "splitting_tools",
 ]
--- a/src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
@@ -8,7 +8,8 @@
 
 """Fast access to the auto optimization on DaCe."""
 
-from typing import Any, Optional, Sequence
+import enum
+from typing import Any, Callable, Optional, Sequence, TypeAlias
 
 import dace
 from dace import data as dace_data
@@ -18,6 +19,29 @@ from dace.transformation.passes import analysis as dace_analysis
 
 from gt4py.next import common as gtx_common
 from gt4py.next.program_processors.runners.dace import transformations as gtx_transformations
+
+
+class GT4PyAutoOptHook(enum.Enum):
+    """Allows to hook into different stages of the auto optimizer.
+
+    The hook system allows to inject certain additional behaviour into `gt_auto_optimize()`.
+    There are multiple hooks supported, see list below, that are called at different
+    stages of the optimization. A hook function receives the SDFG and should modify
+    it inplace.
+
+    The supported values are:
+    - TopLevelDataFlowPre: Called before the top level dataflow is optimized.
+    - TopLevelDataFlow: Called during the top level optimization stage, which might be
+        called multiple times.
+    - TopLevelDataFlowPost: Called after the top level dataflow has been optimized.
+    """
+
+    TopLevelDataFlowPre = enum.auto()
+    TopLevelDataFlow = enum.auto()
+    TopLevelDataFlowPost = enum.auto()
+
+
+GT4PyAutoOptHookFun: TypeAlias = Callable[[dace.SDFG], None]
 
 
 def gt_auto_optimize(
@@ -37,6 +61,7 @@ def gt_auto_optimize(
     assume_pointwise: bool = True,
     validate: bool = True,
     validate_all: bool = False,
+    optimization_hooks: Optional[dict[GT4PyAutoOptHook, GT4PyAutoOptHookFun]] = None,
     **kwargs: Any,
 ) -> dace.SDFG:
     """Performs GT4Py specific optimizations on the SDFG in place.
@@ -92,6 +117,8 @@ def gt_auto_optimize(
             respective value inside the SDFG. This might increase performance.
         assume_pointwise: Assume that the SDFG has no risk for race condition in
             global data access. See the `GT4PyMapBufferElimination` transformation for more.
+        optimization_hooks: A `dict` containing the hooks that should be called,
+            see `GT4PyAutoOptHook` for more information.
         validate: Perform validation during the steps.
         validate_all: Perform extensive validation.
 
@@ -119,6 +146,7 @@ def gt_auto_optimize(
             not if we have too much internal space (register pressure).
     """
     device = dace.DeviceType.GPU if gpu else dace.DeviceType.CPU
+    optimization_hooks = optimization_hooks or {}
 
     with dace.config.temporary_config():
         # Do not store which transformations were applied inside the SDFG.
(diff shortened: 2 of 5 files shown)


src/gt4py/next/otf/compiled_program.py
```diff
@@ -253,7 +260,9 @@ def __call__(
             raise RuntimeError("No program compiled for this set of static arguments.") from e
 
     @functools.cached_property
-    def _argument_descriptor_cache_key_from_args(self) -> Callable:
+    def _argument_descriptor_cache_key_from_args(
+        self,
+    ) -> Callable[..., tuple[ScalarOrTupleOfScalars, ...]]:
```
tehrengruber:
```suggestion
    def _argument_descriptor_cache_key_from_args(
        self,
    ) -> Callable:
```
The output type depends on what the argument descriptor extracts and can be anything. In #2173 this consequently changes to also include domains.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -352,6 +361,16 @@ def _compile_variant(
         if key in self._compiled_programs:
             raise ValueError(f"Program with key {key} already exists.")
 
+        # If we are collecting metrics, create a new metrics entity for this compiled program
+        if config.COLLECT_METRICS_LEVEL:
+            assert metrics.in_collection_mode()
```
tehrengruber: This assert will fail when pre-compiling. Moving this into the `__call__` would be one solution.

src/gt4py/next/otf/compiled_program.py
egparedes: Funny thing is that I didn't intend to make this change, but Copilot suggested it to fix typing issues.

src/gt4py/next/ffront/decorator.py
```diff
@@ -310,16 +310,15 @@ def __call__(
...
-        with metrics.metrics_collection(program_name) as metrics_collection:
-            collect_metrics = metrics_collection is not None and (
-                config.COLLECT_METRICS_LEVEL >= metrics.INFO
-            )
-            if collect_metrics:
+
+        with metrics.collect() as metrics_source:
+            if config.COLLECT_METRICS_LEVEL:
+                assert metrics_source is not None
+                metrics_source.append_to_key(
```
tehrengruber: If collecting just the metric values without a key, the key could just be set later (in CompiledProgram) instead of being assembled in multiple steps. (as discussed with Hannes)

egparedes: Refactored  as discussed with Hannes

src/gt4py/next/otf/compiled_program.py
```diff
@@ -257,8 +256,17 @@ def __call__(
...
         try:
-            self._compiled_programs[key](*args, **kwargs, offset_provider=offset_provider)
+            program = self._compiled_programs[key]
+
+            if config.COLLECT_METRICS_LEVEL:
+                metrics_source = metrics.get_current_source()
+                # This key should be the same as in _compile_variant()
+                metrics_source.key = f"{self.definition_stage.definition.__name__}[{self._compiled_programs.internal_key(key)}]"
+
+            program(*args, **kwargs, offset_provider=offset_provider)
```
havogt, line 274: this could be dangerous in case something raises a TypeError before the `program` call, right?

src/gt4py/next/otf/compiled_program.py
```diff
@@ -377,6 +387,30 @@ def _compile_variant(
         if key in self._compiled_programs:
             raise ValueError(f"Program with key {key} already exists.")
 
+        # If we are collecting metrics, create a new metrics entity for this compiled program
+        if config.COLLECT_METRICS_LEVEL:
+            metric_source: metrics.Source | metrics.SourceHandler
+            compiled_program_pool_key = self._compiled_programs.internal_key(key)
+
+            if metrics.in_collection_mode():
+                # Jitting within a metrics collection context
```
havogt: is this special case needed?

egparedes: Not strictly if we are sure no metadata or metrics have been collected earlier, because that would have triggered the creation of a source without a key inside the handler, which then would collide with the source created in the global sources dict. So I think it's safer and cleaner to do it in the current way.

src/gt4py/next/otf/compiled_program.py
egparedes: Yes, indeed. But how should we fix it? We could add a nested try-except block to convert any exception between extracting program and calling program to a  `RuntimeError`, because I don't see any other recovery strategy, but is it worth it?

havogt: I guess it's ok. Did Till talk to you about this area? He would like to improve this piece of code such that we don't get a key error in case something goes wrong in compilation (because it's the first error that you see). Maybe we can have another look when looking into python overhead.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -377,6 +397,33 @@ def _compile_variant(
         if key in self._compiled_programs:
             raise ValueError(f"Program with key {key} already exists.")
 
+        # If we are collecting metrics, create a new metrics entity for this compiled program
+        if config.COLLECT_METRICS_LEVEL:
```
tehrengruber, line 411: Can we put this into a function?

src/gt4py/next/otf/compiled_program.py
egparedes: I added a small fix for the original TypeError issue originally mentioned here.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -257,12 +264,23 @@ def __call__(
...
         try:
-            self._compiled_programs[key](*args, **kwargs, offset_provider=offset_provider)
-        except TypeError:  # 'Future' object is not callable
-            # ... otherwise we resolve the future and call again
-            program = self._resolve_future(key)
+            program = self._compiled_programs[key]
+            if config.COLLECT_METRICS_LEVEL:
+                metrics_source = metrics.get_current_source()
+                # This key should be the same as in _compile_variant()
+                metrics_source.key = f"{self.definition_stage.definition.__name__}[{self._compiled_programs.internal_key(key)}]"
```
tehrengruber: Can we put this into a function so that the JIT and pre-compile variants use the same convention and refactoring of the key shows it being used in two code-paths?
```
def _metrics_key(self, key):
  return f"{self.definition_stage.definition.__name__}[{self._compiled_programs.internal_key(key)}]"
```
And shouldn't the backend be part of the key as it is for embedded? Otherwise running with multiple backends leads to the same key with different metadata.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -257,12 +264,23 @@ def __call__(
...
+
             program(*args, **kwargs, offset_provider=offset_provider)
+
+        except TypeError:
+            if "program" in locals() and isinstance(program, concurrent.futures.Future):
```
tehrengruber, line 277: Isn't this silently swallowing other TypeErrors?

egparedes: Yes! I forgot to add the `else` branch. Fixed.

src/gt4py/next/otf/compiled_program.py
egparedes: I was a bit reluctant to create a function just to save a function call but it's probably premature optimization. I've fixed it now and will address performance concerns in other PRs after profiling.

src/gt4py/next/otf/compiled_program.py
egparedes: Done.

src/gt4py/next/otf/compiled_program.py
tehrengruber: Keeping it is also fine with me, but I would add a comment then.

src/gt4py/next/otf/compiled_program.py
tehrengruber: That is a nice suggestion actually.

PR #2288, opened 2025-10-07
feat[dace][next]: Demotion of Fields

philip-paul-mueller:
Adding the `demote_fields` argument to `gt_auto_optimizer()`.

The flag instructs the optimizer to handle SDFG arguments, which are globals, as transients and by that allows more possible transformations.
After the has concluded they _might_ be put back, i.e. turned back into a non-transient.
It is not generally possible to put them back as the optimization process might have modified them.

NOTE:
This feature is an experiential unsafe expert feature.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
@@ -8,6 +8,7 @@
 
 """Fast access to the auto optimization on DaCe."""
 
+import warnings
 from typing import Any, Optional, Sequence
 
 import dace
@@ -35,6 +36,7 @@ def gt_auto_optimize(
     gpu_memory_pool: bool = False,
     constant_symbols: Optional[dict[str, Any]] = None,
     assume_pointwise: bool = True,
+    demote_fields: Optional[list[str]] = None,
     validate: bool = True,
     validate_all: bool = False,
     **kwargs: Any,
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -182,6 +190,40 @@ def gt_auto_optimize(
                 validate_all=validate_all,
             )
 
+        # Demote the fields.
+        #  Actually they should probably be at the very start of this function, however,
+        #  they have to be after constant substitution in case the descriptor is modified.
+        original_demoted_descriptors: dict[str, dace_data.Data] = {}
+        if demote_fields:
+            assert isinstance(demote_fields, list)
```
edopao: Would this work?
```suggestion
        if demote_fields is not None:
```

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -119,6 +121,8 @@ def gt_auto_optimize(
             global data access. See the `GT4PyMapBufferElimination` transformation for more.
         optimization_hooks: A `dict` containing the hooks that should be called,
             see `GT4PyAutoOptHook` for more information.
+        demote_fields: Consider these fields as transients for the purpose of optimization.
+            Use at your own risk. See Notes for all implication.
```
```suggestion
            Use at your own risk. See Notes for all implications.
```

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -130,6 +134,10 @@ def gt_auto_optimize(
             has unit strides.
         - In case GPU optimizations are enabled, the function assumes that all
             global fields are already on the GPU.
+        - When using the `demote_fields` feature the user has to ensure that the fields
+            meet all constraints of a transient, see ADR18 for more. Furthermore, there
+            is no guarantee if the demoted field remain parts of the SDFG's call
```
```suggestion
            is no guarantee whether the demoted field remains part of the SDFG's call
```

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
philip-paul-mueller: It is even more correct in that way, thanks for spotting it.

PR #2290, opened 2025-10-07
cartesian[feat]: Expand push vertical map down to include `ForScope`

FlorianDeconinck:
## Description

When writing the vertical map transformation for `dace:cpu_kfirst` we have missed the serial vertical loop, a.k.a. `ForScope` in schedule tree. This PR completes the intended work.

Also: 
- renaming pass file to match PEP.

## Requirements

### DIFF
--- a/src/gt4py/cartesian/definitions.py
+++ b/src/gt4py/cartesian/definitions.py
@@ -154,16 +154,16 @@ class GTError(Exception):
 
 
 class GTSyntaxError(GTError):
-    def __init__(self, message, *, frontend):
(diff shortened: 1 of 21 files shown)


romanc (commented):
Nice! Thanks for renaming the file.

Could you extend test coverage in https://github.com/GridTools/gt4py/blob/main/tests/cartesian_tests/unit_tests/backend_tests/test_dace_backend.py to include the fact that we now also handle `ForScope`s? If I see correctly, this should be as simple as adding a copy kernel with a FORWARD or BACKWARD loop.

PR #2291, opened 2025-10-08
feat[next][dace]: Hooksystem for `gt_auto_optimizer()`

philip-paul-mueller:
The system allows to inject hooks at different stage of the optimization, which allows to perform more specific optimizations.
Note that only an initial set of hook points was added yet.


edopao (approved):
LGTM

PR #2292, opened 2025-10-08
bug[next]: Fix dead-code elimination on some `concat_where` expressions

tehrengruber:
Execute `concat_where.expand_tuple_args` before dead-code-elimination such in an expressions like
```
let tmp = concat_where(cond, {a, b, c}, {d, e, f})
  non_trivial_expr(tmp[0], tmp[2])
end
```
the `tmp[1]` value is removed. Otherwise domain inference fails. That `tmp` is referenced twice in this example is important as otherwise function inlining and `tuple_get` propagation would also remove the tuple. Additionally, `non_trivial_expr`, must be a non-trivial expression as otherwise cps tuple inlining would also resolve this case.

### DIFF
--- a/src/gt4py/next/iterator/transforms/collapse_tuple.py
+++ b/src/gt4py/next/iterator/transforms/collapse_tuple.py
@@ -63,8 +63,8 @@ def _is_trivial_or_tuple_thereof_expr(node: itir.Node) -> bool:
     """
     Return `true` if the expr is a trivial expression (`SymRef` or `Literal`) or tuple thereof.
 
-    Let forms with trivial body and args as well as `if` calls with trivial branches are also
-    considered trivial.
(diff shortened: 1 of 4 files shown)


src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -83,7 +83,7 @@ def _is_trivial_or_tuple_thereof_expr(node: itir.Node) -> bool:
         return _is_trivial_or_tuple_thereof_expr(node.args[1])
     # This will duplicate the condition and increase the size of the tree, but this is probably
     # acceptable.
-    if cpm.is_call_to(node, "if_"):
+    if cpm.is_call_to(node, ("if_", "concat_where")):
```
tehrengruber, line 86: This doesn't fix the bug, but is a small improvement I stumbled upon while writing the fix.

PR #2294, opened 2025-10-09
feat[next][dace]: Updated `gt_auto_optimizer()`'s Hook-System

philip-paul-mueller:
Adding more hooks to further influence how optimization is done.
The PR mainly adds the possibility to pass callbacks to the vertical and horizontal Map fusion transformation during top level dataflow optimization as well as the callback to control map promotion.

Note this PR depends on: https://github.com/GridTools/gt4py/pull/2296

PR #2295, opened 2025-10-09
feat[next][dace]: Enable customization of auto-optimize in dace backend workflow

edopao:
This PR exposes the `gt_auto_optimize()` parameters to application, in order to allow setting up custom dace backends. The parameters include regular function arguments as well as the new hook functions introduced by #2293 and #2294.

Note that this PR removes the setting of `bindings.make_persistent` (and just relies on its default value) because this option is going to be deprecated soon in GT4Py.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/backend.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/backend.py
@@ -119,14 +119,8 @@ def make_dace_backend(
         else None
     }
 
-    if (
-        optimization_args.get("gpu_memory_pool", False)
-        and core_defs.CUPY_DEVICE_TYPE == core_defs.DeviceType.ROCM
-    ):
-        warnings.warn(
-            "Unsupported configuration, 'gpu_memory_pool=True' not compatible with RoCM GPU.",
-            stacklevel=2,
-        )
+    if gpu and core_defs.CUPY_DEVICE_TYPE == core_defs.DeviceType.ROCM:
+        # Backend supports gpu memory pool only on CUDA target.
         optimization_args["gpu_memory_pool"] = False
 
     return DaCeBackendFactory(  # type: ignore[return-value] # factory-boy typing not precise enough


src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -154,6 +154,9 @@ def gt_auto_optimize(
     device = dace.DeviceType.GPU if gpu else dace.DeviceType.CPU
     optimization_hooks = optimization_hooks or {}
 
+    if gpu_memory_pool and device != dace.DeviceType.GPU:
+        raise NotImplementedError("Memory pool only available for GPU device.")
+
```
philip-paul-mueller: I would turn it into a warning.
My reason is that we also do not check if `gpu_block_size` is `None` when on CPU, we simply ignore the parameter.
Furthermore, the parameter clearly indicates that it is meant for GPU.
However, I do not have strong opinions on this.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -8,12 +8,15 @@
...
 from __future__ import annotations
 
+from typing import Any, Final
+
 import factory
 
 import gt4py.next.allocators as next_allocators
 from gt4py._core import definitions as core_defs
-from gt4py.next import backend, common
+from gt4py.next import backend, common as gtx_common, config
```
philip-paul-mueller: Why have you aliased `common` but not `config`?

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -69,112 +67,117 @@ class Params:
...
-        blocking_dim: When not 'None', apply 'LoopBlocking' SDFG transformation
-            on this dimension.
-        blocking_size: Block size to use in 'LoopBlocking' SDFG transformation,
-            when enabled.
-        use_metrics: Add SDFG instrumention to collect the stencil compute time.
+        cached: Cache the lowered SDFG as a JSON file and the compiled programs.
+        auto_optimize: Enable the SDFG auto-optimize pipeline.
+        async_sdfg_call: Make an asynchronous SDFG call on GPU to allow overlapping
+            of GPU kernel execution with the Python driver code.
+        optimization_args: A `dict` containing configuration parameters for
```
philip-paul-mueller, line 86: Technically `optimization_hooks` is part of `optimization_args`, thous you should be more clear in the description that only "the ones that do not have a dedicated parameter" are included here.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -69,112 +67,117 @@ class Params:
...
         use_zero_origin: Can be set to `True` when all fields passed as program
             arguments have zero-based origin. This setting will skip generation
             of range start-symbols `_range_0` since they can be assumed to be zero.
 
     Returns:
-        A custom dace backend object.
+        A dace backend with custom configuration for the target device.
     """
+    fixed_optimization_args: Final[dict[str, Any]] = {
+        "assume_pointwise": True,
```
philip-paul-mueller: What if I want to set `assume_pointwise` to `False`?
You should only check for the values for which you have provided a real argument.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -69,112 +67,117 @@ class Params:
...
+            else None  # let `gt_auto_optimize` select `unit_strides_kind` based on `gpu` argument
+        ),
+        "validate": False,
+        "validate_all": False,
+    }
+
+    if optimization_hooks and not auto_optimize:
+        raise ValueError("Optimizations hook given, but auto-optimize pipeline is disabled.")
+    if optimization_args and not auto_optimize:
+        raise ValueError("Optimizations args given, but auto-optimize pipeline is disabled.")
```
philip-paul-mueller: I would turn this into a warning.
Because it makes disabling auto optimization, for any reason, super hard, instead of just changing one value, I may have to change the stencil in a substantiation way.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
edopao: I agree. I will actually remove this check.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
edopao: I think because `common` exists also inside dace workflow, but it is not strictly needed. I will remove the alias for common.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
edopao: I agree.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
edopao: I will improve the description, as you suggest.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
edopao: My original idea was to not allow changes to this parameter, because we know that by construction all gt4py SDFGs should respect `assume_pointwise=True`.

philip-paul-mueller: It was more like an example, it could be any of them.
But the same is true for `validate`.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -69,112 +68,117 @@ class Params:
...
+        "assume_pointwise": True,  # SDFGs built from GTIR reaspect this assumption
+        "gpu_memory_pool": (use_memory_pool if gpu else False),
+        "optimization_hooks": optimization_hooks,
+        "unit_strides_kind": (
+            common.DimensionKind.HORIZONTAL
+            if config.UNSTRUCTURED_HORIZONTAL_HAS_UNIT_STRIDE
+            else None  # let `gt_auto_optimize` select `unit_strides_kind` based on `gpu` argument
+        ),
+        "validate": False,
+        "validate_all": False,
```
philip-paul-mueller: This violates the description of the parameter, because `validate` does not have a "dedicated parameter" (I know I invented the term and you understood it, but do you think an explanation would be good?) so it is legal to list it in `optimization_args`.
The only thing that should be in that dict should be "unit_strides_kind" everything else not (`gpu_memory_pool` could be a  default).

edopao: I see your point. I have now made `fixed_optimization_args` smaller, it contains `gpu_memory_pool`, `optimization_hooks` and `unit_strides_kind`.

edopao: Update: now I have applied your comment, the dict contains only `unit_strides_kind`.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -69,112 +67,98 @@ class Params:
...
+
+    if config.UNSTRUCTURED_HORIZONTAL_HAS_UNIT_STRIDE:
+        arg_value = optimization_args.get("unit_strides_kind", None)
+        if arg_value is None:
+            optimization_args = optimization_args | {
+                "unit_strides_kind": common.DimensionKind.HORIZONTAL
+            }
+        elif arg_value != common.DimensionKind.HORIZONTAL:
+            raise ValueError(
+                f"Configuration mismatch, received unit_strides_kind='{arg_value}', expected '{common.DimensionKind.HORIZONTAL}'."
```
philip-paul-mueller:
```suggestion
        if "unit_strides_kind" in optimization_args:
            raise ValueError(
                f"Configuration mismatch, received unit_strides_kind='{arg_value}', expected '{common.DimensionKind.HORIZONTAL}'."
        optimization_args = optimization_args | {
                "unit_strides_kind": common.DimensionKind.HORIZONTAL
        }
```
The only value that is allowed to specify here is the value that the `env` dictates, thus it is in my view meaningless to provide in that case, thus we should simply forbid it in that case.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -69,112 +67,98 @@ class Params:
...
-        use_metrics: Add SDFG instrumention to collect the stencil compute time.
+        cached: Cache the lowered SDFG as a JSON file and the compiled programs.
+        auto_optimize: Enable the SDFG auto-optimize pipeline.
+        async_sdfg_call: Make an asynchronous SDFG call on GPU to allow overlapping
+            of GPU kernel execution with the Python driver code.
+        optimization_args: A `dict` containing configuration parameters for
+            the SDFG auto-optimize pipeline. Note that parameters that are derived
+            from GT4Py configuration cannot be overriden and therefore cannot appear
+            here. For example, this function will throw an exception if called with
+            'unit_strides_kind' different from the value in GT4Py configuration.
```
philip-paul-mueller: I would not use `unit_strides_kind` as an example, as it is different, I think that the real example is `gpu`.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -69,112 +67,98 @@ class Params:
...
-            on this dimension.
-        blocking_size: Block size to use in 'LoopBlocking' SDFG transformation,
-            when enabled.
-        use_metrics: Add SDFG instrumention to collect the stencil compute time.
+        cached: Cache the lowered SDFG as a JSON file and the compiled programs.
+        auto_optimize: Enable the SDFG auto-optimize pipeline.
+        async_sdfg_call: Make an asynchronous SDFG call on GPU to allow overlapping
+            of GPU kernel execution with the Python driver code.
+        optimization_args: A `dict` containing configuration parameters for
+            the SDFG auto-optimize pipeline. Note that parameters that are derived
```
philip-paul-mueller:
```suggestion
            the SDFG auto-optimize pipeline, see `gt_auto_optimize()`. Note that parameters that are derived
```

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -84,34 +84,41 @@ def make_dace_backend(
...
+    # The `gt_optimization_args` set contains the parameters of `gt_auto_optimize()`
+    # that are derived from the gt4py configuration, and therefore cannot be customized.
+    gt_optimization_args: Final[set[str]] = {"gpu", "constant_symbols", "unit_strides_kind"}
+
     if optimization_args is None:
         optimization_args = {}
     elif optimization_args and not auto_optimize:
         warnings.warn("Optimizations args given, but auto-optimize is disabled.", stacklevel=2)
+    elif any(arg in gt_optimization_args for arg in optimization_args):
+        intersect_args = gt_optimization_args.intersection(optimization_args.keys())
```
philip-paul-mueller:
```suggestion
    elif intersect_args := gt_optimization_args.intersection(optimization_args.keys()):
```

edopao: Good idea!

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -329,20 +329,13 @@ class DaCeTranslator(
...
     auto_optimize: bool
+    auto_optimize_args: dict[str, Any] | None
     async_sdfg_call: bool
     use_metrics: bool
+
     disable_itir_transforms: bool = False
     disable_field_origin_on_program_arguments: bool = False
 
-    # auto-optimize arguments
-    gpu_block_size: tuple[int, int, int] = (32, 8, 1)
```
philip-paul-mueller, line 338: It seems that the deletion of this line has caused a slight performance regression, because now DaCe default is again used.
See https://github.com/GridTools/gt4py/pull/2313

PR #2299, opened 2025-10-10
refactor[next]: Cleanup implicit offsets

havogt:
Use a lookup function instead of appending implicit offsets to the offset_provider dict.

This is a temporary solution with some quirks until we implement a dedicated class for the offset provider.


src/gt4py/next/common.py
```diff
@@ -1010,6 +1006,8 @@ def is_neighbor_table(obj: Any) -> TypeGuard[NeighborTable]:
 
 OffsetProviderElem: TypeAlias = Dimension | NeighborConnectivity
 OffsetProviderTypeElem: TypeAlias = Dimension | NeighborConnectivityType
+# Note: `OffsetProvider` and `OffsetProviderType` should not be accessed directly,
+# use the `get_offset` and `get_offset_type` functions instead.
 OffsetProvider: TypeAlias = Mapping[Tag, OffsetProviderElem]
 OffsetProviderType: TypeAlias = Mapping[Tag, OffsetProviderTypeElem]
```
havogt, line 1012: @egparedes If you are motivated, for safety we could replace `Mapping` by a Protocol that hides `__getitem__`/`get` for normal usage and type ignores the `__getitem__` access in `get_offset`/`get_offset_type`.

src/gt4py/next/common.py
```diff
@@ -1034,6 +1032,40 @@ def offset_provider_to_type(
     }
 
 
+def _get_dimension_name_from_implicit_offset(offset: str) -> str:
+    assert offset.startswith(_IMPLICIT_OFFSET_PREFIX)
+    return offset[len(_IMPLICIT_OFFSET_PREFIX) :]
+
+
+def get_offset(offset_provider: OffsetProvider, offset_tag: str) -> OffsetProviderElem:
```
tehrengruber, line 1040: What is the purpose of splitting the two functions? The implementation looks identical.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -131,17 +132,28 @@ def _collect_offset_definitions(
...
         node.walk_values()
         .if_isinstance(itir.OffsetLiteral)
         .filter(lambda offset_literal: isinstance(offset_literal.value, str))
         .getattr("value")
     ).to_set()
-    if not used_offset_tags.issubset(set(offset_provider_type.keys())):
+    if not used_offset_tags.issubset(
+        offset_provider_type.keys()
+        | {common.dimension_to_implicit_offset(dim) for dim in dimension_tags}
+    ):
```
tehrengruber: Beside the already poor state of this function, this becomes even hard to understand now. If you revert the changes in this file and instead iterate over all `used_offset_tags` in the loop below that should also work right? 
```
for offset_name in used_offset_tags:
  dim_or_connectivity_type = get_offset(offset_provider_type, offset_name)
  ...
```

src/gt4py/next/common.py
tehrengruber: I think this is worth the effort, I already see the future me forgetting to use this function. If I find time after my debugging session, I'll do it.

src/gt4py/next/common.py
tehrengruber: I combined them.

src/gt4py/next/common.py
tehrengruber: This is sadly not so easy and was therefore skipped. I'll skip an elaborate answer since this is a temporary mechanism anyway.

Issue #208, opened 2020-10-06
Expose the iteration index in GTScript


egparedes:
Expose a new builtin GTScript function exposing the iteration index per axis as a regular `int` value. In the Syntax Workshop, it was decided to use `index(AXES_NAME)` (where AXIS_NAME should be `I`,`J` or `K`)

Example:

```python
with computation(), interval(...):
    field_a = field_b[-2, 0, 0] + field_b[1, 0, 0]
    current_k = index(K)
    if current_k < 3:
         field_a = 1
```

Functionality: the returned value can only be used in the same places as a regular `int` run-time parameter (e.g., not as an offset in a field indexing operation)

romanc:
@FlorianDeconinck / @twicki this issue is (partially) getting addressed in the experimental "THIS_K" functionality, right? I like the idea of extending the concept to all axis.

FlorianDeconinck:
Yes the experimental `THIS_K` is a first pass at this. It's implemented in `debug` and `dace:X` backends for now and live on NASA's experimental branch

FlorianDeconinck:
Poke @oelbert for future discussion on that syntax.

PR #2300, opened 2025-10-10
feature[cartesian]: expose iteration index  in `K` (experimental)

romanc:
## Description

This PR adds the experimental feature to access the iteration index in `K`. Syntax is just `K` for now, e.g. 

```py
    # Flip mid-level variables
    k_inv = k_end - K
    pmid0_in = pmid0_inv.at(K=k_inv)
```

or 

```py
"""
krel [IntField]: Release layer where buoyancy sorting first occurs
kinv [IntField]: Inversion layer with PBL top interface as lower interface
"""
with computation(), interval():
    # ...
    if K >= kinv - 1 and K <= max(krel - 1, kinv - 1):
        uplus_3D = uplus + PGFc * ssu0 * (pifc0[0, 0, 1] - pifc0)
        vplus_3D = vplus + PGFc * ssv0 * (pifc0[0, 0, 1] - pifc0)
    # ...
```

The feature allows for dynamic (i.e. non-static) interval bounds and works in combination with [absolute indexing](https://github.com/GridTools/gt4py/blob/main/docs/development/ADRs/cartesian/experimental/indexing-absolute-k.md).

:warning: This feature is immature and is likely to change (as allowed by the [rules of experimental features](https://github.com/GridTools/gt4py/blob/main/docs/development/ADRs/cartesian/experimental-features.md). For example, we are using issue #208 to find a less loaded name. Be careful when using this feature. It might break your code (as allowed by the rules) in unforeseen ways.

Please read the ADR for a full discussion of the stability and the pros/cons of this experimental feature.

## Requirements

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
@@ -8,20 +8,17 @@
 
 from __future__ import annotations
 
-import re
 from typing import Final
 
 import dace
 
 from gt4py.eve import codegen
-from gt4py.next import common as gtx_common
 from gt4py.next.iterator import builtins as itir_builtins
 from gt4py.next.otf import languages, stages
 from gt4py.next.program_processors.runners.dace import utils as gtx_dace_utils
 from gt4py.next.type_system import type_specifications as ts
 
 
-FIELD_RANGE_PARAM_RE: Final[re.Pattern] = re.compile(r"^__(.+)_(\S+)_range$")
 FIELD_SYMBOL_GT_TYPE: Final[ts.ScalarType] = ts.ScalarType(
     kind=getattr(ts.ScalarKind, itir_builtins.INTEGER_INDEX_BUILTIN.upper())
 )
@@ -50,19 +47,28 @@ def _update_sdfg_scalar_arg(
     code.append(f"{_cb_last_call_args}[{sdfg_arg_index}] = {actype_call}({call_arg})")
 
 
-def _validate_sdfg_scalar_arg(
-    code: codegen.TextBlock,
-    sdfg_arg_desc: dace.data.Data,
-    sdfg_arg_index: int,
-    call_arg: str,
-) -> None:
+def _unpack_args(code: codegen.TextBlock, num_args: int, arg_name: str) -> list[str]:
+    """Unpack a sequence of arguments (either a list or a tuple) into variables.
+
+    Each element of the given sequence gets a name 'arg_name' with an index-based suffix.
+
+    >>> code = codegen.TextBlock()
+    >>> _unpack_args(code, 3, "var")
+    ['var_0', 'var_1', 'var_2']
+    >>> code.lines
+    ['var_0, var_1, var_2, = var']
+    >>> _unpack_args(code, 1, "var_2")
+    ['var_2_0']
+    >>> code.lines
+    ['var_0, var_1, var_2, = var', 'var_2_0, = var_2']
     """
-    Emit Python asserts to validate a scalar argument in the SDFG arglist
-    against the argument value passed to the gt4py program call.
-    """
-    assert isinstance(sdfg_arg_desc, dace.data.Scalar)
-    code.append(f"assert isinstance({_cb_last_call_args}[{sdfg_arg_index}], ctypes._SimpleCData)")
-    code.append(f"assert {_cb_last_call_args}[{sdfg_arg_index}] == {call_arg}")
+    tuple_args = [f"{arg_name}_{i}" for i in range(num_args)]
+    if num_args == 0:
+        raise ValueError("Cannot unpack argument with length zero.")
+    else:
+        # The trailing comma is needed to unpack single-element tuples
+        code.append(f"{', '.join(tuple_args)}, = {arg_name}")
+    return tuple_args
 
 
 def _parse_gt_param(
@@ -71,7 +77,6 @@ def _parse_gt_param(
     arg: str,
     code: codegen.TextBlock,
     sdfg_arglist: dict[str, dace.data.Data],
-    make_persistent: bool,
 ) -> None:
     """Emit Python code to parse a program argument and set the required fields in the SDFG arglist.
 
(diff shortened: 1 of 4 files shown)


src/gt4py/cartesian/frontend/defir_to_gtir.py
```diff
@@ -415,6 +417,17 @@ def visit_ComputationBlock(self, node: ComputationBlock) -> gtir.VerticalLoop:
             loc=location_to_source_location(node.loc),
         )
 
+    def visit_IteratorAccess(self, iterator_access: IteratorAccess) -> gtir.IteratorAccess:
+        warnings.warn(
```
FlorianDeconinck: Can we move this along side the check that the iterator is `K` not `I` or `J`? I'd like to keep the logic of the feature compact and as close to the beginning of the pipeline as possible.

FlorianDeconinck: Either in the `_is_iterator_access` below or in the `nodes.py` has a factory checker thingy

romanc: sure, we can put this wherever. I wasn't sure myself and just put it in the first opportunity that I had reading the PR top-down. That was probably not a smart choice ... let' me move down into `_is_iterator_access()` :+1:

FlorianDeconinck: This typically highlight an issue with the modularity of the frontend, which is a direct knock-on effect from the visitor pattern. Nothing's perfect, but I'd like to see if we can offload to the frontend a lot of the check, leaving the subsequent code to be a fairly clean transformation layer

romanc:
> > (the failing cscs-ci is most likely a fluke in the system. I've had green runs before on this PR. The only changes since the last green run are adding a type hint and moving the warning message - that shouldn't break ci in that way)
> 
> Yes, there are every now and then some flukes in the system availability. Additionally, we realized we introduced recently some flaky unit tests in `next` that failed randomly, so we just merged a PR that should fix them for now

Thank you for managing and maintaining the CI system, @egparedes!  :pray:

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -710,6 +711,24 @@ def _is_datadims_indexing_name(name: str):
...
 
 
+def _is_iterator_access(name: str, loc: nodes.Location) -> bool:
+    if name in ["I", "J", "IJ"]:
+        raise GTScriptSyntaxError(
+            f"Parallel axis {name} can't be queried - only K - at line {loc.line} (column {loc.column})",
+            loc=loc,
+        )
+
+    warnings.warn(
```
FlorianDeconinck: Nit pick: should we have a function that format the experimental  feature, something like:

```python
warn_experimental_feature(name="Iterator access in `K`", github_adr="experimental/iteration-index-k.md")
```

Am I over-engineering warning message? Maybe.

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
```diff
@@ -497,6 +497,14 @@ def visit_TernaryOp(self, node: oir.TernaryOp, **kwargs: Any) -> str:
 
         return f"({if_code} if {condition} else {else_code})"
 
+    def visit_IteratorAccess(
+        self, node: oir.IteratorAccess, ctx: tir.Context, **kwargs: Any
+    ) -> str:
+        if node.name == "K":
```
FlorianDeconinck:
```suggestion
        if node.name == tir.Axis.K:
```

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
```diff
@@ -257,6 +257,12 @@ def visit_NativeFuncCall(self, node: oir.NativeFuncCall, **kwargs: Any) -> str:
 
         return f"{function_name}({arguments})"
 
+    def visit_IteratorAccess(self, node: oir.IteratorAccess, ctx: Context, **kwargs: Any) -> str:
+        if node.name == "K":
```
FlorianDeconinck:
```suggestion
        if node.name == tir.Axis.K:
```

src/gt4py/cartesian/gtc/dace/oir_to_treeir.py
romanc: uh ... good find!

src/gt4py/cartesian/frontend/gtscript_frontend.py
romanc: yeah, I think that can be useful since we are starting to have a couple experimental features. Gives consistency to the warning messages, which is nice. Let me make a separate PR and update the existing experimental features.

romanc: PR to add the function: https://github.com/GridTools/gt4py/pull/2324. I'll update this PR as soon as the other is merged ...

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -710,6 +711,24 @@ def _is_datadims_indexing_name(name: str):
     return name.endswith(f".{_DATADIMS_INDEXER}")
 
 
+def _is_iterator_access(name: str, loc: nodes.Location) -> bool:
+    if name in ["I", "J", "IJ"]:
```
twicki: should the check not just be 
```suggestion
    if name != "K":
```

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
```diff
@@ -257,6 +257,12 @@ def visit_NativeFuncCall(self, node: oir.NativeFuncCall, **kwargs: Any) -> str:
 
         return f"{function_name}({arguments})"
 
+    def visit_IteratorAccess(self, node: oir.IteratorAccess, ctx: Context, **kwargs: Any) -> str:
+        if node.name == tir.Axis.K:
+            return tir.k_symbol(ctx.scope)
+
+        return tir.Axis(node.name).iteration_symbol()
```
twicki, line 264: Do we access to other symbols for loops?

src/gt4py/cartesian/gtc/passes/gtir_dtype_resolver.py
```diff
@@ -51,6 +51,11 @@ def visit_FieldAccess(
             loc=node.loc,
         )
 
+    def visit_IteratorAccess(
+        self, iterator_access: gtir.IteratorAccess, **kwargs: Any
+    ) -> gtir.IteratorAccess:
+        return gtir.IteratorAccess(name=iterator_access.name, dtype=DataType.INT64)
```
twicki: can we do int64 even if we assign to int32?

src/gt4py/cartesian/frontend/gtscript_frontend.py
romanc: That sounds simpler. The way I read the code, I think the original idea here was to return a targeted error for parallel axis (i.e. `I` and `J`) and thus they were spelled out. Since `K` is the only remaining axis, the proposed solution is equivalent (and simpler). In the (unlikely) case that we ever extend axis names, the proposed solution is safer with a potentially confusing error message (for the newly added dimension). I prefer that to the alternative where the error message would be consistent (because we keep the current formulation) and then we risk undefined behavior in case we forget to catch the newly added dimension. So, tldr, let's change it.

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
romanc: Are you asking why we need to treat the `tir.Axis.K` differently? If so, the answer is because the K-iteration symbol in the generated (schedule) tree isn't always the same. It's due to how DaCe does variable replacement (in the resulting SDFG), which assumes that variables aren't shadowed. The way how `for`-loops are implemented in DaCe v1 forces us to generate a new "k-iteration symbol" for each loop to be on the safe side. The commit message of https://github.com/GridTools/gt4py/commit/5ed21e42eeef034f6f5e59b4edaf0e0f4e581928 has an example.

src/gt4py/cartesian/gtc/passes/gtir_dtype_resolver.py
romanc: Good question. I don't know what the policy is at the gt4py level for iteration variables. Should they adhere to the literal integer precision? Maybe ... I have have to defer to @FlorianDeconinck for this one (since I'm just doing the PR here and he did the early works on this).

src/gt4py/cartesian/frontend/gtscript_frontend.py
romanc: updated version is pushed now

src/gt4py/cartesian/gtc/passes/gtir_dtype_resolver.py
FlorianDeconinck: Yeah I think this is bad. This is probably a case of `LITERAL_INT_PRECISION` check to get the `DataType`

romanc: I'll have a look tomorrow. It's not going to be straight forward checking of `LITERAL_INT_PRECISION` because we chose to allow users to change the literal precision per stencil (in the `BuildOptions` - details in the [ADR](https://github.com/GridTools/gt4py/blob/main/docs/development/ADRs/cartesian/frontend/literal-precision.md#decision)). I'll have to see where we are here (pretty far down in the stack as far as I can see now) and how to access or forward this information.

romanc: So, it turns out, treating iterator access special in the `gtir_dtype_resolver` is a bit of a red flag itself. I went around that by defining a `data_type` on the `IteratorAccess` at the definition IR level and propagating that type downwards. That seems to work nicely and allows us to respect the per-stencil `literal_integer_precision` easily. Details in https://github.com/GridTools/gt4py/pull/2300/commits/27667905a6728ec10c962f06be0194443580f461.

romanc:
cscs-ci run

src/gt4py/cartesian/gtc/dace/oir_to_tasklet.py
twicki: Makes sense, thanks

src/gt4py/cartesian/gtc/passes/gtir_dtype_resolver.py
twicki: looks good now, thanks

PR #2301, opened 2025-10-12
feat[bug]: Fix name collision in collapse tuple pass

tehrengruber:
Fixes a bug in the propagate nested let transformation of the collapse tuple pass, where a name collision between a let param to be propagated upwards and a reference inside another let argument or the let body would lead to wrong results. This
```
let a = (let c = 1 in c end)
  {a, c}
end
```
now correctly transformed into
```
{1, c}
```

### DIFF
--- a/src/gt4py/next/common.py
+++ b/src/gt4py/next/common.py
@@ -71,6 +71,8 @@ class DimensionKind(StrEnum):
 
 _DIM_KIND_ORDER = {DimensionKind.HORIZONTAL: 0, DimensionKind.LOCAL: 1, DimensionKind.VERTICAL: 2}
 
+_IMPLICIT_OFFSET_PREFIX: Final[str] = "_Off"
+
 
 def dimension_to_implicit_offset(dim: str) -> str:
     """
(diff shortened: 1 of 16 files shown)


src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -481,15 +486,28 @@ def transform_propagate_nested_let(self, node: itir.FunCall, **kwargs) -> Option
...
+                        reserved_refs = symbol_ref_utils.collect_symbol_refs(
+                            [*node.args, original_inner_expr]
+                        )
                     rename_map: dict[
                         str, ir.SymRef
                     ] = {}  # mapping from symbol with a collision to its new (unique) name
                     for sym, val in zip(arg.fun.params, arg.args):
-                        unique_sym = ir_misc.unique_symbol(sym, [s.id for s in outer_vars.keys()])
+                        unique_sym = ir_misc.unique_symbol(
+                            sym, [s.id for s in outer_vars.keys()] + reserved_refs
```
egparedes: Not important at all, but a single list comprehension should be enough here:
```suggestion
                            sym, [*(s.id for s in outer_vars.keys()), *reserved_refs]
```

PR #2303, opened 2025-10-13
refactor[next][dace]: Move zero-origin constant substitution out of lowering module

edopao:
This PR moves one "SDFG transformation" out of the lowering module `gtir_to_sdfg`. The design principle is that the `gtir_to_sdfg` module should only do the lowering to SDFG, while SDFG transformations should be applied outside. One such transformation is the pruning of field origin symbols for the program arguments, when the field arguments are known to have zero origin. After this PR change, the pass will be applied by `gt_auto_optimize()` together with other constant symbols.

### DIFF
--- a/tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_translation.py
+++ b/tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_translation.py
@@ -56,6 +56,7 @@ def device_type(request) -> str:
 
 def _translate_gtir_to_sdfg(
     ir: itir.Program,
+    offset_provider: gtx_common.OffsetProvider,
     device_type: core_defs.DeviceType,
     auto_optimize: bool,
     async_sdfg_call: bool,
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -363,12 +388,22 @@ def _generate_sdfg_without_configuring_dace(
...
+            # Besides, `gt_simplify()` will bring the SDFG into a canonical form
+            # that the GPU transformations can handle. This is a workaround for
+            # an issue with scalar expressions that are promoted to symbolic expressions
+            # and computed on the host (CPU), but the intermediate result is written
+            # to a GPU global variable (https://github.com/spcl/dace/issues/1773).
+            gtx_transformations.gt_substitute_compiletime_symbols(
+                sdfg, constant_symbols, validate=True
+            )
             gtx_transformations.gt_gpu_transformation(sdfg, try_removing_trivial_maps=True)
+        elif len(constant_symbols) != 0:
```
philip-paul-mueller, line 402: I am not sure if I like that, i.e. that the SDFG is touched.
I agree it is not stated anywhere that `auto_optimize=False` implies that the SDFG is not touched, but I would still say that we do nothing in that case.
But I do not have particular strong opinions on that, but you should document that behaviour somewhere.

edopao: I see your point, and I agree. However, this specific branch (CPU, skip auto_optimize) is used by the dace orchestration. This feature assumes that the field arguments (global arrays) have zero origin, and it also skips auto-optimize on the SDFGCovertible objects, because it targets the full-application SDFG. I will add a note to explain this use case. Anyway, it is still possible to achieve what you say (untouched SDFG) by leaving `disable_field_origin` to its default value (`False`).

PR #2304, opened 2025-10-13
feat[dace][next]: Map Splitter

philip-paul-mueller:
The `SplitMap` transformation is able to split a Map into different smaller ones, in that sense it is similar to the `VerticalMapSplitRange` transformation, but targets different cases.
The transformation targets the following pattern, `MapExit -> (AccessNode)`, in addition `AccessNode` is only written to by the Map and there is only one consumer.
Furthermore, there is more data written to than is read from the `AccessNode`, such situation are created by `broadcast()` expression and `concat_where`s.

Note: This transformation does not integrate the transformation into the Pipeline, this is instead done in [PR#2284](https://github.com/GridTools/gt4py/pull/2284).

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
@@ -11,6 +11,7 @@
 from __future__ import annotations
 
 import copy
+import warnings
 from typing import Any, Callable, Final, Optional, Sequence, Union
 
 import dace
@@ -527,6 +528,14 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
     four iteration in the second dimension, will get a block size of `(32, 4, 1)`.
     Note that this modification will not influence the launch bound value.
 
+    The block size and launch bounrds are set according to the aadditional following rules:
+        - Determine the number of map parameters that have range > 1.
+        - If there is only one such parameter then handle the map as 1D map,
+          else if there are more than 2 parameters with range > 1 then handle it as ordinary map.
+        - For 1D maps set the block size such that only the dimension with the largest
+          range is set to the specified block size, all other dimensions are set to 1.
+        - If launch_bounds_1d is not set then set the launch_bounds of `scan` maps to 512 to limit their register usage.
+
     Args:
         block_size_Xd: The size of a thread block on the GPU for `X` dimensional maps.
         launch_bounds_Xd: The value for the launch bound that should be used for `X`
@@ -604,8 +613,18 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
         super().__init__()
         if block_size_1d is not None:
             self.block_size_1d = block_size_1d
+            if self.block_size_1d[1] != 1 or self.block_size_1d[2] != 1:
+                warnings.warn(
+                    f"1D map block size specified with more than one dimension larger than 1. Configured 1D block size: {self.block_size_1d}.",
+                    stacklevel=0,
+                )
         if block_size_2d is not None:
             self.block_size_2d = block_size_2d
+            if self.block_size_2d[2] != 1:
+                warnings.warn(
+                    f"2D map block size specified with more than twi dimensions larger than 1. Configured 2D block size: {self.block_size_2d}.",
+                    stacklevel=0,
+                )
         if block_size_3d is not None:
             self.block_size_3d = block_size_3d
         self.launch_bounds_1d = _gpu_launch_bound_parser(
@@ -635,7 +654,6 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
         - If the block size of the map is already set.
         - If the map is at global scope.
         - If if the schedule of the map is correct.
-        - If launch_bounds_1d is not set then set the launch_bounds of `scan` maps to 512 to limit their register usage.
         """
         scope = graph.scope_dict()
         if scope[self.map_entry] is not None:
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -136,34 +136,97 @@ def apply(
...
+    sdfg: dace.SDFG,
+    state: dace.SDFGState,
+    map_entry: dace_nodes.MapEntry,
+    remove_unused_data: bool = False,
+    single_use_data: Optional[dict[dace.SDFG, set[str]]] = None,
+) -> int:
+    """Remove the Map denoted by `map_entry` _unconditionally_.
+
+    The function can only operate on top level Maps, nested Maps are not supported.
+    Furthermore, adjacent nodes that ave become isolated by the Map removal will
```
edopao:
```suggestion
    Furthermore, adjacent nodes that have become isolated by the Map removal will
```

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -136,34 +136,97 @@ def apply(
...
+    also be removed.
+    If `remove_unused_data` is specified, then the function will also remove data
+    descriptors that are no longer used and were referenced by the Map.
```
edopao:
```suggestion
    descriptors that were referenced by the Map and are no longer used.
```

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -136,34 +136,97 @@ def apply(
...
+    adjacent_nodes.update(oedge.dst for oedge in state.out_edges(map_exit))
+
+    if any(not isinstance(ac, dace_nodes.AccessNode) for ac in adjacent_nodes):
+        raise ValueError(
+            f"Expected that all adjacent nodes of '{map_entry.map.label}' are AccessNode"
+        )
+
+    # Find all the data that is accessed inside the Map scope.
+    if remove_unused_data:
+        map_scope_datas = {ac.data for ac in map_scope.data_nodes()}
```
edopao:
```suggestion
        map_scope_data_names = {ac.data for ac in map_scope.data_nodes()}
```

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -136,34 +136,97 @@ def apply(
...
+    # Now remove the nodes of the scope.
+    removed_number_nodes = map_scope.number_of_nodes()
+    state.remove_nodes_from(map_scope.nodes())
+
+    # Now check for potentially isolated nodes.
+    #  Process them in deterministic order.
+    for adjacent_node in sorted(adjacent_nodes, key=lambda ac: ac.data):
+        if state.degree(adjacent_node) == 0:
+            if not isinstance(adjacent_node, dace_nodes.AccessNode):
+                map_scope_datas.add(adjacent_node.data)
```
edopao: I don't understand this check. We have already tested that all adjacent nodes are AccessNodes, on line 196

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -136,34 +136,97 @@ def apply(
...
+    map_exit: dace_nodes.MapExit = state.exit_node(map_entry)
+    map_scope = state.scope_subgraph(map_entry, include_exit=True, include_entry=True)
+
+    # Now find all notes that are producer or consumer of the Map, after we removed
+    #  the nodes of the Maps we need to check if they have become isolated.
+    # NOTE: Needs to be a `set` to handle multiple connections with the MapEntry node.
+    adjacent_nodes: set[dace_nodes.AccessNode] = {iedge.src for iedge in state.in_edges(map_entry)}
+    adjacent_nodes.update(oedge.dst for oedge in state.out_edges(map_exit))
+
+    if any(not isinstance(ac, dace_nodes.AccessNode) for ac in adjacent_nodes):
```
edopao: (Optional, just a matter of taste) It is a little easier to read:
```suggestion
    if not all(isinstance(ac, dace_nodes.AccessNode) for ac in adjacent_nodes):
```

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
```diff
@@ -136,34 +136,97 @@ def apply(
...
+    if remove_unused_data:
+        map_scope_datas = {ac.data for ac in map_scope.data_nodes()}
+
+    # Now remove the nodes of the scope.
+    removed_number_nodes = map_scope.number_of_nodes()
+    state.remove_nodes_from(map_scope.nodes())
+
+    # Now check for potentially isolated nodes.
+    #  Process them in deterministic order.
+    for adjacent_node in sorted(adjacent_nodes, key=lambda ac: ac.data):
```
edopao, line 212: Since we are going to remove only isolated nodes, I think we really don't need to sort here.

src/gt4py/next/program_processors/runners/dace/transformations/map_splitter.py
```diff
@@ -0,0 +1,240 @@
...
+    splitting_tools as gtx_stools,
+)
+
+
+@dace_properties.make_properties
+class MapSplitter(dace_transformation.SingleStateTransformation):
+    """Split a Map based on a transient.
+
+    The transformation matches the following pattern `MapExit -> (AccessNode)`,
+    where `AccessNode` needs to a transient single use data whose only producer
```
edopao:
```suggestion
    where `AccessNode` needs to be a transient single use data whose only producer
```

src/gt4py/next/program_processors/runners/dace/transformations/map_splitter.py
```diff
@@ -0,0 +1,240 @@
...
+    #  [issue#1911](https://github.com/spcl/dace/issues/1911) has been solved.
+    _single_use_data: Optional[dict[dace.SDFG, set[str]]]
+
+    @classmethod
+    def expressions(cls) -> Any:
+        return [dace.sdfg.utils.node_path_graph(cls.map_exit, cls.access_node)]
+
+    def __init__(
+        self,
+        remove_dead_dataflow: Optional[bool] = None,
```
edopao, line 74:
```suggestion
        remove_dead_dataflow: bool = True,
```

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
```diff
@@ -90,12 +100,13 @@ def __init__(
         self,
         *args: Any,
         single_use_data: Optional[dict[dace.SDFG, set[str]]] = None,
+        assume_single_use_data: Optional[bool] = None,
```
edopao, line 103:
```suggestion
        assume_single_use_data: bool = False,
```

src/gt4py/next/program_processors/runners/dace/transformations/map_splitter.py
```diff
@@ -0,0 +1,240 @@
...
+                sdfg=sdfg,
+                state=graph,
+                map_entry=sub_me,
+            )
+            new_map_exits.append(sub_mx)
+        gtx_mfutils.delete_map(graph=graph, map_entry=map_entry, map_exit=map_exit)
+
+        # Now perform the split of `self.access_node`. Parts that are not read will
+        #  later be eliminated by dead dataflow elimination.
+        # TODO(phimuell): Make it simpler to call the split function.
```
edopao: I think that this TODO does not belong here.
```suggestion
```

src/gt4py/next/program_processors/runners/dace/transformations/map_splitter.py
```diff
@@ -0,0 +1,240 @@
...
+            assert (
+                isinstance(output_node, dace_nodes.AccessNode) and output_node.desc(sdfg).transient
+            )
+
+            if graph.out_degree(output_node) == 0:
+                gtx_transformations.gt_remove_map(
+                    sdfg=sdfg,
+                    state=graph,
+                    map_entry=graph.entry_node(mx),
+                    # We would need to update single use data to really use it.
```
edopao, line 237: I get this comment, and I agree. However, we can safely remove `output_node.data`, as a post step, because the original access node was single-use data.

edopao (commented):
Just some minor comments

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
philip-paul-mueller: I usually follow the practice that `_data` indicates a name/string.
But I can understand that the `datas` (with `s`) is bugging you.

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
philip-paul-mueller: You are right, this thing stems back from the time when there when the check on 196 was not there.
I actually allowed it before, but realized that it is an error.

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
philip-paul-mueller: What are you betting on that?

src/gt4py/next/program_processors/runners/dace/transformations/map_splitter.py
philip-paul-mueller: I agree it is nicer that way, the main issue I have with it is that it is redundant.
Because the default value is set in two places, either in the `remove_dead_dataflow` property of the class or in the default argument of `__init__()`.
I agree it is very unlikely that we change it, but I still would like to have a single source of truth.

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
philip-paul-mueller: Same as above.

src/gt4py/next/program_processors/runners/dace/transformations/map_splitter.py
philip-paul-mueller: You are technically right, but there is no other place to put it.
I decided to remove it.

src/gt4py/next/program_processors/runners/dace/transformations/map_splitter.py
philip-paul-mueller: I kind of agree, however, I would still propose to keep it here.
You see everything that you remove from the graph does not need to be processed later.
In order to remove these Map fragments you would first to locate them, decide that you can remove them and actually remove them.
Here you get that for free.
Essentially it is cleaning up after yourself.

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
edopao: Have you seen side effects? I understand that there could places where the order matters, but here we only remove isolated nodes.

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
edopao: Exactly, the `s` was bugging me.

src/gt4py/next/program_processors/runners/dace/transformations/map_splitter.py
edopao: OK, I see.

src/gt4py/next/program_processors/runners/dace/transformations/dead_dataflow_elimination.py
philip-paul-mueller: I have not seen it, but I think in a system where removing "trivial" tasklets caused the dynamic of the system to change, one should expect everything.

edopao (approved):
LGTM

PR #2305, opened 2025-10-13
refactor[next]: Reduce programs runtime overhead

egparedes:
This PR changes the implementation of the Python side of program calls to reduce its overhead.

Argument canonicalization now uses a set of cached dynamically compiled functions depending on the shape of the arguments provided at run-time. Compiled variant selection skips the step of converting the offset provider instance to a (pseudo) type and hash it. Additionally, some `NDArrayField` properties have been converted to `cached_properties`, and `gtfn` extract args has been also touched to use more efficient list comprehensions.

### DIFF
--- a/tests/next_tests/benchmarks/benchmark_program_call.py
+++ b/tests/next_tests/benchmarks/benchmark_program_call.py
@@ -111,15 +111,14 @@ def benchmark_horizontal_copy_01_arg_program(
     input_field = gtx.empty([(Cell, size)], dtype=gtx.float64)
     out_field = gtx.empty([(Cell, size)], dtype=gtx.float64)
 
-    # Compilation and warm-up
+    # Initial compilation
     compiled_program = horizontal_copy_01_arg_program.with_backend(backend)
-    for _ in range(WARMUP_ITERS):
-        compiled_program(
-            input_field,
-            out=out_field,
-            horizontal_start=0,
-            horizontal_end=size,
-        )
+    compiled_program(
+        input_field,
+        out=out_field,
+        horizontal_start=0,
+        horizontal_end=size,
+    )
 
     benchmark(
         compiled_program,
@@ -290,7 +289,7 @@ def benchmark_horizontal_copy_25_arg_program(
     input_fields = [gtx.empty([(Cell, size)], dtype=gtx.float64) for _ in range(25)]
     out_field = gtx.empty([(Cell, size)], dtype=gtx.float64)
 
-    # Compilation and warm-up
+    # Initial compilation
     compiled_program = horizontal_copy_25_arg_program.with_backend(backend)
     compiled_program(
         *input_fields,


msimberg:
> Do you know if latest main also fails at the same time steps? If main works, then we should look into more details the changes in this PR

I only tested with 1.0.9 and this branch, so indeed main could also be broken.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -117,7 +117,22 @@ class NdArrayField(
...
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        super().__init_subclass__(**kwargs)
+        if hasattr(cls, "array_ns") and not hasattr(cls, "array_byte_bounds"):
+            try:
+                cls.array_byte_bounds = staticmethod(
+                    getattr(cls.array_ns, "byte_bounds", None)
+                    or cls.array_ns.lib.array_utils.byte_bounds
+                )
+            except AttributeError:
+                pass
```
havogt, line 136: Why is this not a cached_property?

tests/next_tests/unit_tests/test_utils.py
```diff
@@ -57,3 +58,51 @@ def testee(x):
...
+    canonicalizer = utils.make_args_canonicalizer(signature, **options)
+    canonical_form = ((1, 2), {"c": 3, "d": 4})
+
+    in_place_kwargs = options["allow_kwargs_mutation"] is True and options["sort_kwargs"] is False
+
+    output = canonicalizer((1, 2), kwargs := {"c": 3, "d": 4})
+    assert output == canonical_form
+    assert (output[1] is kwargs) == in_place_kwargs
+    assert canonicalizer.cache_info().currsize == 1
+    del output, kwargs
```
havogt: Why not just split into different tests?

tests/next_tests/benchmarks/benchmark_program_call.py
```diff
@@ -0,0 +1,314 @@
...
+    def const_no_args(out: gtx.Field[Dims[IDim], gtx.float64]):
+        const(out=out)
+
+    size = 1
+    out_field = gtx.empty([(IDim, size)], dtype=gtx.float64)
+    compiled_program = const_no_args.with_backend(backend)
+
+    # Compilation and warm-up
+    for _ in range(WARMUP_ITERS):
+        compiled_program(out=out_field)
```
havogt:
```suggestion
    compiled_program = const_no_args.with_backend(backend).compile()
    gtx.wait_for_compilation()
```
pytest benchmark itself does warmup, I think it can be removed.

src/gt4py/next/utils.py
```diff
@@ -217,3 +232,193 @@ def equalize_tuple_structure(
...
+    def cache_clear(self) -> None: ...
+
+
+class CustomCallArgsCanonicalizerFactory(Protocol):
+    def __call__(
+        self, passed_args_count: int, passed_kwargs_keys: tuple[str, ...]
+    ) -> CallArgsCanonicalizer: ...
+
+
+def make_args_canonicalizer_factory(
```
havogt, line 260: - Is this now the 3rd version of the argument canonicalization?
- Can the other 2 be replaced by this in a follow up PR? 
- Maybe add a comment that `inspect.Signature.bind()` is not good enough for our use-case.

src/gt4py/next/utils.py
```diff
@@ -217,3 +232,193 @@ def equalize_tuple_structure(
...
+            canonical_args_expr = "args"
+            if canonical_kwargs_expr == "kwargs":
+                # Args are already canonical, return identity
+                return cast(CallArgsCanonicalizer, lambda args, kwargs: (args, kwargs))
+
+        canonicalizer_func_name = f"canonicalizer_for_{signature_id}_{passed_args_count}_{str.join('_', passed_kwargs_keys)}"
+        canonicalizer_src = f"""
+from __future__ import annotations
+
+def {canonicalizer_func_name}(
```
havogt, line 373: does this need to have a unique name?

src/gt4py/next/utils.py
```diff
@@ -217,3 +232,193 @@ def equalize_tuple_structure(
...
+        canonical_kwargs_expr = "kwargs"
+    else:
+        kwargs_items_exprs = [f"{name!r}: kwargs[{name!r}]" for name in kwonly_names]
+        canonical_kwargs_expr = f"{{ {str.join(', ', kwargs_items_exprs)} }}"
+
+    signature_id = abs(hash(tuple([pos_args_count, *kwonly_names, *all_names])))
+
+    @functools.cache
+    def canonicalizer_factory(
+        passed_args_count: int, passed_kwargs_keys: tuple[str, ...]
```
havogt:
```suggestion
        passed_pos_args_count: int, passed_kwargs_keys: tuple[str, ...]
```
makes the code below slightly easier to read

src/gt4py/next/utils.py
```diff
@@ -217,3 +232,193 @@ def equalize_tuple_structure(
...
+    ) -> CallArgsCanonicalizer:
+        # This function and the generated canonicalizer are optimized for performance,
```
havogt: this function doesn't need to be super fast right? Only the generated one.

src/gt4py/next/utils.py
```diff
@@ -217,3 +232,193 @@ def equalize_tuple_structure(
...
+        # at the cost of readability.
+        if passed_args_count > pos_args_count:
+            raise ValueError(
+                f"Too many positional arguments (expected {pos_args_count}, got {passed_args_count})."
+            )
+        passed_kwargs_set = {*passed_kwargs_keys}
+        if unexpected_kwargs := (passed_kwargs_set - all_names):
+            raise ValueError(f"Got unexpected keyword arguments: {unexpected_kwargs}.")
```
havogt: This would miss the case where we pass a pos_only arg as kwarg, no?

tests/next_tests/unit_tests/test_utils.py
```diff
@@ -57,3 +58,51 @@ def testee(x):
...
+
+    assert canonicalizer((), {"d": 4, "a": 1, "b": 2, "c": 3}) == canonical_form
+    assert canonicalizer.cache_info().currsize == 3
+
+    with pytest.raises(ValueError, match="Missing keyword arguments: {'c'}"):
+        canonicalizer((1, 2), {"d": 4})
+
+    with pytest.raises(ValueError, match="Got unexpected keyword arguments: {'foo'}"):
+        canonicalizer((1, 2), {"d": 4, "foo": 5})
+
```
havogt:
```suggestion
    with pytest.raises(ValueError, match="Got unexpected keyword arguments: {'a'}"):
        canonicalizer({"a": 1, "d": 2, "b": 3, "c": 4})
```
i.e. passing pos_only as kwarg

src/gt4py/next/utils.py
```diff
@@ -217,3 +232,193 @@ def equalize_tuple_structure(
...
+            unpack_args_stmt = (
+                (f"{str.join(', ', (f'a{i}' for i in range(passed_args_count)))}, = args")
+                if passed_args_count
+                else "# No args to unpack"
+            )
+
+            passed_args_indices = iter(range(passed_args_count))
+            if allow_kwargs_mutation:
+                canonical_args_comprehension = (
+                    f"kwargs.pop({key!r})"
```
havogt, line 346: is popping faster than getting?

src/gt4py/next/utils.py
```diff
@@ -217,3 +232,193 @@ def equalize_tuple_structure(
...
+        sort_kwargs=sort_kwargs,
+    )
+
+    def canonicalizer(args: tuple, kwargs: dict[str, Any]) -> tuple[tuple, dict[str, Any]]:
+        return canonicalizer_factory(
+            passed_args_count=len(args), passed_kwargs_keys=tuple(sorted(kwargs.keys()))
+        )(args, kwargs)
+
+    canonicalizer.cache_info = canonicalizer_factory.cache_info  # type: ignore[attr-defined] # adding new attribute
+    canonicalizer.cache_clear = canonicalizer_factory.cache_clear  # type: ignore[attr-defined] # adding new attribute
```
havogt, line 442: why?

src/gt4py/next/embedded/nd_array_field.py
egparedes: This is stored at a class level, not at instance level, since it only depends on the ndarray library used in the concrete NdArrayField subclass. I've added a comment to make it more clear in the code.

src/gt4py/next/utils.py
egparedes: Yes, the idea is that this version (or an enhanced version of it with additional functionality to give better error messages in the typing passes) should replace the others in a follow-up PR, to keep this PR smaller and focused in overhead-related issues.

I've added the requested comment.

src/gt4py/next/utils.py
egparedes: Basically yes, but this function would be callthe first time a different call signature is used by the user in the hot path, so it's still good to be as fast as possible. I've updated the comment to drop the `at the cost of readability` part, since it doesn't not really apply here.

src/gt4py/next/utils.py
egparedes: Yes, fixed now.

src/gt4py/next/utils.py
egparedes: Not really, but by popping we also remove the key from the passed kwargs dict, which means that the kwargs dict will be already canonicalized at the end, since it will only contain keywork-only args.

src/gt4py/next/utils.py
egparedes: Not really, but it's useful to give a half-meaningful name to the generated canonicalizer functions for debugging purposes, since the `inspect.Signature` object doesn't contain the name of the function. I have change it now to accept an optional argument with a explicitly provided name

src/gt4py/next/utils.py
egparedes: The `canonicalizer_factory` is functools.cache instance but it is never directly exposed to the user, so here we just assign the cache-related methods to the function returned to the caller (mostly for debugging purposes).

tests/next_tests/benchmarks/benchmark_program_call.py
egparedes: Done

tests/next_tests/unit_tests/test_utils.py
egparedes: Done

tests/next_tests/unit_tests/test_utils.py
egparedes: Done

tests/next_tests/benchmarks/benchmark_program_call.py
```diff
@@ -0,0 +1,308 @@
...
+
+    # Compilation and warm-up
+    compiled_program = horizontal_copy_01_arg_program.with_backend(backend)
+    for _ in range(WARMUP_ITERS):
+        compiled_program(
+            input_field,
+            out=out_field,
+            horizontal_start=0,
+            horizontal_end=size,
+        )
```
havogt: to fix the failure

havogt (approved):
lgtm

havogt:
cscs-ci run

PR #2306, opened 2025-10-13
fix[next][dace]: Avoid name conflict between tasklet connector and data node

edopao:
SDFG validation checks that there are no name conflicts between tasklet connector and data entities in one SDFG. Because of some naming convention, it could happen that `__arg0` was used as a tasklet connector and as a data node, which cause SDFG validation to fail. This PR tries to avoid name conflicts by creating unique names for the tasklet connectors.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
@@ -663,7 +663,7 @@ class LambdaToDataflow(eve.NodeVisitor):
                     dace.Memlet(data=index_expr.dc_node.data, subset="0"),
                 )
             else:
-                assert index_connector not in connector_mapping
+                assert isinstance(index_expr, SymbolExpr)
 
         return self._construct_tasklet_result(
             field_desc.dtype, deref_node, connector_mapping["val"]
--- a/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
@@ -124,6 +124,8 @@ class DataflowBuilder(Protocol):
 
         if inputs is None:
             inputs = set()
+        else:
+            assert inputs.isdisjoint(outputs)
 
         connector_mapping = {conn: self.unique_tasklet_connector() for conn in (inputs | outputs)}
         new_code = _replace_connectors_in_code_string(code, language, connector_mapping)
@@ -159,6 +161,8 @@ class DataflowBuilder(Protocol):
 
         if inputs is None:
             inputs = {}
+        else:
+            assert inputs.keys().isdisjoint(outputs.keys())
 
         connector_mapping = {
             conn: self.unique_tasklet_connector() for conn in (inputs.keys() | outputs.keys())
--- a/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
@@ -248,9 +248,10 @@ duration = run_cpp_end_time - run_cpp_start_time;
     end_state.add_edge(
         tlet_stop_timer, "duration", end_state.add_access(output), None, dace.Memlet(f"{output}[0]")
     )
-
+    # We normally wrap `add_tasklet()` in order to ensure unique tasklet connector
+    # names, see `gtir_to_sdfg.DataflowBuilder.add_tasklet()`. Since this wrapper
+    # is not available here, we run `validate()` on the instrumented SDFG.
     sdfg.validate()
-    # End of `add_instrumentation` function.
 
 
 def make_sdfg_call_sync(sdfg: dace.SDFG, gpu: bool) -> None:


src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -51,6 +52,21 @@
 from gt4py.next.type_system import type_specifications as ts, type_translation as tt
 
 
+TAKSLET_IN_CONNECTOR: Final[str] = "__gtir_inp"
```
philip-paul-mueller: I am actually not sure if that helps.
Before the connectors where named `__arg`, `__inp`, `__out` and so on, however, now they are either named `__gtir_inp` or `__gtir_out` and by that are not globally unique.

edopao: The problem was not that they were not globally unique. The problem was that GTIR uses `__arg` (and `__arg0` etc.), and by accident this was also used as connector name on the tasklets.

edopao: I forgot one detail. Sometimes the GTIR arguments are lowered to data nodes, which created the name conflict with tasklet connectors.

philip-paul-mueller: After some thinking I realized that the problem was that sometimes connectors of nested SDFG were named `__arg1` which were then also access nodes.
So this PR solved the issue by changing the name scheme of Tasklet connectors, however, it will fail again once GT4Py changed its naming scheme to use `__gtir_inp` for input connectors that end up in nested SDFGs.
It is therefore not a permanent solution.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -51,6 +40,17 @@
...
 
+def _replace_connectors_in_code_string(
+    code: str, language: dace.dtypes.Language, connector_mapping: Mapping[str, str]
+) -> str:
+    """Helper function to replace connector names in the code of a Python tasklet."""
+    code_block = dace.properties.CodeBlock(code, language)
+    transformed_code_stmts = [
+        dace_astutils.ASTFindReplace(connector_mapping).visit(stmt) for stmt in code_block.code
+    ]
+    return dace.properties.CodeBlock(transformed_code_stmts, language).as_string
```
philip-paul-mueller, line 51: Since that thing uses Python's `ast` module I am pretty sure that it can only handle Python and not `C++`.
Thus I would add an assert for those cases.
An alternative, probably faster too, would be to use a regex like `s/\b${OLD_NAME}\b/${NEW_NAME}/g`, but I am not sure how safe it is.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -102,30 +105,73 @@ def add_map(
...
+        from any other tasklet connector or data entity in the SDFG. Therefore,
+        this fuction returns the mapping from the given connector names, as they
+        appear in `inputs` and `outputs` arguments, to the new unique names.
+        """
         unique_name = self.unique_tasklet_name(name)
-        return state.add_tasklet(unique_name, inputs, outputs, code, **kwargs)
+
+        if inputs is None:
+            inputs = set()
+
```
philip-paul-mueller, line 129: As an additional test `assert inputs.isdisjoint(outputs)`.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -194,9 +194,9 @@ def add_instrumentation(sdfg: dace.SDFG, gpu: bool) -> None:
...
         "gt_start_timer",
         inputs={},
-        outputs={"time"},
+        outputs={"__time"},
         code="""\
-time = static_cast<double>(
+__time = static_cast<double>(
     std::chrono::duration_cast<std::chrono::nanoseconds>(
         std::chrono::high_resolution_clock::now().time_since_epoch()).count()
 ) / 1e9;
```
philip-paul-mueller, line 202: I kind of understand why you do not use the wrapper here, but you should add an explanation why.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -246,9 +246,16 @@ def add_instrumentation(sdfg: dace.SDFG, gpu: bool) -> None:
...
-        tlet_stop_timer, "duration", end_state.add_access(output), None, dace.Memlet(f"{output}[0]")
+        tlet_stop_timer,
+        "__duration",
+        end_state.add_access(output),
+        None,
+        dace.Memlet(f"{output}[0]"),
     )
 
+    sdfg.validate()
+    # End of `add_instrumentation` function.
```
philip-paul-mueller:
```suggestion
```

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -629,43 +629,45 @@ def _visit_deref(self, node: gtir.FunCall) -> DataExpr:
...
                     index_expr.dc_node,
                     None,
                     deref_node,
-                    deref_connector,
+                    connector_mapping[index_connector],
                     dace.Memlet(data=index_expr.dc_node.data, subset="0"),
                 )
             else:
-                assert isinstance(index_expr, SymbolExpr)
+                assert index_connector not in connector_mapping
```
philip-paul-mueller: What happened to the `isinstance()` test.

edopao: I think it is not strictly needed because the type annotation of `index_expr` is `MemletExpr | ValueExpr | SymbolExpr`. However, I will put it back for readability.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
edopao: I will add a comment on the call to `validate()`, I call it just to ensure that there are no name conflicts.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Regarding the C++ case, that is checked by the dace implementation. The CodeBlock constructor will raise an exception.

Regarding the regex, I tried that and I got a wrong substitution (OK, I was not using `\b`, good to know). I think this way should be more "robust".

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
philip-paul-mueller: I am never sure, but probably in that case `\<`/`\>` should be used: https://stackoverflow.com/questions/27723018/whats-the-difference-between-b-and-in-regex
But I always confuse them.

PR #2308, opened 2025-10-14
bug[next]: Add support for `nan` and `inf` literals in GTIR

tehrengruber:
This surfaced in icon4py where after some new optimizations the constant folding created `inf` literals, which then failed in code generation in the gtfn & roundtrip backend.


src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -92,26 +92,31 @@ def visit_SymRef(self, node: gtfn_ir_common.SymRef, **kwargs: Any) -> str:
...
+                cpp_value = node.value
+
+                if node.value in ["inf", "-inf"]:
+                    cpp_value = (
+                        f"std::numeric_limits<{cpp_utils.pytype_to_cpptype(node.type)}>::infinity()"
+                    )
+                    return f"-{cpp_value}" if node.value == "-inf" else cpp_value
+                elif node.value == "nan":
+                    cpp_value = f"std::numeric_limits<{cpp_utils.pytype_to_cpptype(node.type)}>::signaling_NaN()"
+                    return cpp_value
```
egparedes, line 109: I think it is more readable this other way but feel free to ignore:
```suggestion
            case "float32" | "float64" as float_type:
                cpp_value = node.value

                if node.value in ["inf", "-inf"]:
                    sign = node.value[:-3]
                    cpp_value = (
                        f"{sign}std::numeric_limits<{cpp_utils.pytype_to_cpptype(float_type)}>::infinity()"
                    )
                    return cpp_value
                elif node.value == "nan":
                    cpp_value = f"std::numeric_limits<{cpp_utils.pytype_to_cpptype(node.type)}>::signaling_NaN()"
                    return cpp_value
```

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
```diff
@@ -92,26 +92,31 @@ def visit_SymRef(self, node: gtfn_ir_common.SymRef, **kwargs: Any) -> str:
...
+
+                if "." not in cpp_value and "e" not in cpp_value and "E" not in cpp_value:
```
egparedes: Feel free to ignore:
```suggestion
                if not {".", "e", "E"} & set(cpp_value):
```

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
tehrengruber: I picked the `sign` suggestion, but the `node.value[:-3]` makes me nervous.

src/gt4py/next/program_processors/codegens/gtfn/codegen.py
tehrengruber: I found this more readable:
```python
if not any(c in cpp_value for c in ".eE"):
    cpp_value = f"{cpp_value}."
```

PR #2312, opened 2025-10-16
refactor[next][dace]: cleanup update_sdfg_args

edopao:
This PR is mainly a code refactoring.
The only change is the way the program call decorator iterates on the arguments: before it accessed the arguments list by index (`args[i]`), which led to multiple accesses at the same index for field arguments, because we need to retrieve the field range and the memory strides for all dimensions. With this change, the arguments list is unpacked first, then each argument is accessed as a variable (`arg0`, `arg1` etc.). Since this code was on the call path, the code change should bring some small speedup.

### DIFF
--- a/tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_bindings.py
+++ b/tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_bindings.py
@@ -8,6 +8,7 @@
 
 """Test the bindings stage of the dace backend workflow."""
 
+import functools
 import numpy as np
 import pytest
 
@@ -151,14 +152,33 @@ def {_bind_func_name}(device, sdfg_argtypes, args, last_call_args):
 assert _binding_source_with_zero_origin != _binding_source
 
 
-@pytest.mark.parametrize(
-    "static_domain_config",
-    [(False, _binding_source), (True, _binding_source_with_zero_origin)],
-)
+_dace_compile_call = dace_workflow.compilation.DaCeCompiler.__call__
+
+
+def mocked_compile_call(
+    self,
+    inp: stages.CompilableSource[languages.SDFG, languages.LanguageSettings, languages.Python],
+    use_metrics: bool,
+    use_zero_origin: bool,
+):
+    assert len(inp.library_deps) == 0
+
+    # ignore assert statements
+    binding_source_pruned = "\n".join(
+        line
+        for line in inp.binding_source.source_code.splitlines()
+        if not line.lstrip().startswith("assert")
+    )
+
+    binding_source_ref = _binding_source_with_zero_origin if use_zero_origin else _binding_source
+    assert binding_source_pruned == binding_source_ref(use_metrics)
+    return _dace_compile_call(self, inp)
+
+
 @pytest.mark.parametrize("use_metrics", [False, True])
-def test_bind_sdfg(static_domain_config, use_metrics, monkeypatch):
+@pytest.mark.parametrize("use_zero_origin", [False, True])
+def test_bind_sdfg(use_metrics, use_zero_origin, monkeypatch):
     M, N, K = (30, 20, 10)
-    use_zero_origin, binding_source_ref = static_domain_config
 
     @gtx.field_operator
     def testee_op(
@@ -179,22 +199,6 @@ def test_bind_sdfg(static_domain_config, use_metrics, monkeypatch):
     ):
         testee_op(a, b, out=out, domain={IDim: (1, M - 1), JDim: (2, N - 2), KDim: (3, K - 3)})
 
-    compilable_source: (
-        stages.CompilableSource[languages.SDFG, languages.LanguageSettings, languages.Python] | None
-    ) = None
-
-    dace_compile_call = dace_workflow.compilation.DaCeCompiler.__call__
-
-    def mocked_compile_call(
-        self,
-        inp: stages.CompilableSource[languages.SDFG, languages.LanguageSettings, languages.Python],
-    ):
-        nonlocal compilable_source
-        compilable_source = inp
-        return dace_compile_call(self, inp)
-
-    monkeypatch.setattr(dace_workflow.compilation.DaCeCompiler, "__call__", mocked_compile_call)
-
     backend = dace_runner.make_dace_backend(
         gpu=False,
         cached=False,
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -84,48 +65,24 @@ def _parse_gt_param(
...
-                    tuple_param_name,
-                    tuple_param_type,
-                    tuple_arg,
-                    code,
-                    sdfg_arglist,
-                    make_persistent,
-                )
+        # Each element of a tuple gets a name with an index-based suffix and it is recursively visited.
+        tuple_args = [f"{arg}_{i}" for i in range(len(param_type.types))]
+        code.append("{} = {}".format(", ".join(tuple_args), arg))
```
egparedes:
```suggestion
        code.append(f"{', '.join(tuple_args)}, = {arg}"
```
or 
```suggestion
        code.append(f"{str.join(', ', tuple_args)}, = {arg}"
```
Whatever your prefer, both are equivalent. The first option is more common but I lately started using the second one, since I find it more readable.

Also note that a trailing comma is needed to unpack the tuple in case there is only one argument.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -281,10 +211,11 @@ def {_cb_get_stride}(ndarray, dim_index):
     #   On the first time, we use the regular SDFG call, which constructs the SDFG
     #   arguments list and validates that all data containers and free symbols are set.
     with code.indented():
-        for i, param in enumerate(program_source.entry_point.parameters):
-            arg = f"{_cb_args}[{i}]"
+        arg_vars = [f"arg{i}" for i in range(len(program_source.entry_point.parameters))]
+        code.append("{} = {}".format(", ".join(arg_vars), _cb_args))
```
egparedes: Same as above:
```suggestion
        code.append(f"{', '.join(arg_vars)}, = {_cb_args}")
```

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_bindings.py
```diff
@@ -36,92 +36,119 @@ def _get_stride(ndarray, dim_index):
...
-    last_call_args[5] = ctypes.c_double(args[2][1][1])
-    last_call_args[6] = ctypes.c_double(args[3])
-    last_call_args[3].value = args[4].data_ptr()
-    last_call_args[12] = ctypes.c_int(args[4].domain.ranges[0].start)
-    last_call_args[13] = ctypes.c_int(args[4].domain.ranges[0].stop)
-    last_call_args[14] = ctypes.c_int(args[4].domain.ranges[1].start)
-    last_call_args[15] = ctypes.c_int(args[4].domain.ranges[1].stop)
-    last_call_args[16] = ctypes.c_int(_get_stride(args[4].ndarray, 0))
-    last_call_args[17] = ctypes.c_int(_get_stride(args[4].ndarray, 1))\
+    arg0, arg1, arg2, arg3, arg4 = args
```
egparedes: Update examples to also have a  trailing comma and add one test for the one argument case.
```suggestion
    arg0, arg1, arg2, arg3, arg4, = args
```

egparedes (changes requested):
Looks good. All my comments are actually the same comment in different places, about using f-strings directly and taking into account the 1-element tuple/args case

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
edopao: I used the explicit `format()` syntax because I didn't like nesting `"` and `'`, but I am going to change it as you suggest.

Good catch about the trailing comma in case there is only one argument. I have modified the logic to handle this case.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -50,19 +47,29 @@ def _update_sdfg_scalar_arg(
...
-    assert isinstance(sdfg_arg_desc, dace.data.Scalar)
-    code.append(f"assert isinstance({_cb_last_call_args}[{sdfg_arg_index}], ctypes._SimpleCData)")
-    code.append(f"assert {_cb_last_call_args}[{sdfg_arg_index}] == {call_arg}")
+    tuple_args = [f"{arg_name}_{i}" for i in range(num_args)]
+    if num_args == 0:
+        raise ValueError("Cannot unpack argument with length zero.")
+    elif num_args == 1:
+        code.append(f"{tuple_args[0]} = {arg_name}[0]")
+    else:
+        code.append(f"{', '.join(tuple_args)} = {arg_name}")
```
egparedes: The special condition for the one element case looks overkill and slightly inconsistent to me. Adding a comma at the end would be enough, consistent and not uncommon in python. I leave the final decision to you but in my opinion something like this would be enough:
```suggestion
    else:
        # The trailing comma is needed to unpack single-element tuples
        code.append(f"{', '.join(tuple_args)}, = {arg_name}")
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -77,13 +49,11 @@ def __init__(
...
-        #   Therefore, we use the name of the build folder as module name.
-        binding_module_name = os.path.basename(program.sdfg.build_folder)
-        self.update_sdfg_ctype_arglist = _get_sdfg_ctype_arglist_callback(
-            binding_module_name, bind_func_name, binding_source.source_code
-        )
+        # Here we dinamically compile this function and add it to the compiled program.
+        exec(binding_source.source_code, global_namespace := {})  # type: ignore[var-annotated]
+        self.update_sdfg_ctype_arglist = global_namespace[bind_func_name]
+        # For debug purpose, we set a unique module name on the compiled function.
+        self.update_sdfg_ctype_arglist.__module__ = os.path.basename(program.sdfg.build_folder)
```
egparedes, line 56: Not super important, but it is weird to use `/` slashes in the `__module__` attribute of a symbol and I'm afraid it might break some tooling. I would replace the slashes by dots and add a virtual prefix path indicating that is related to gt4py-dace. Something like:
```suggestion
        self.update_sdfg_ctype_arglist.__module__ = f"__gt4py_dace_callbacks.{os.path.basename(program.sdfg.build_folder).replace('/', .)}"
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
edopao: The `basename()` function returns the tail of the path, which should not include the slash character.

PR #2298, opened 2025-10-10
feat[dace][next]: Set GPU Thread Block Size Per Dimension from `gt_auto_optimizer()`

philip-paul-mueller:
Before `gt_auto_optimize()` only exposed the `gpu_block_size` argument.
This PR adds the `gpu_block_size_spec` which allows to specify the size of the thread block of a map per dimension.

PR #2313, opened 2025-10-16
perf[next][dace] Set GPU thread block size properly

iomaganaris:
I couldn't figure out how but after https://github.com/GridTools/gt4py/pull/2298 I noticed that the thread block sizes and `launch_bounds` were set to the DaCe default of `(32, 1, 1)`.
This PR fixes this issue by providing some defaults to the backend. Those defaults are:
- `(32, 8 , 1)` for 2D maps
- `(256, 1, 1)` for 1D maps or 2D maps where one of the ranges is has size `1`
- In case there is a scan inside a 1D map then set the `launch_bounds` to `512` to limit register usage and improve occupancy

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
@@ -654,8 +654,14 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
         """Modify the map as requested."""
         gpu_map: dace_nodes.Map = self.map_entry.map
         map_size = gpu_map.range.size()
+        dims_to_inspect = len(map_size)
         num_map_params = 0
-        for axis_size in map_size:
+        for i, axis_size in enumerate(map_size):
+            if i > 0 and map_size[i - 1] == 1:
+                assert axis_size <= 1, (
+                    "GPU thread block size setting currently does not support maps where non-leading "
+                    "dimensions have size greater than one if the previous dimension has size one."
+                )
             if axis_size != 1:
                 num_map_params += 1  # Handle 2D maps where one dimension has range 1 as 1D map
 
@@ -678,17 +684,18 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
             block_size = list(self.block_size_3d)
             launch_bounds = self.launch_bounds_3d
 
-        # If there are more than three dimensions DaCe will condense them into
-        #  the `z` dimension of the block, so we have to ignore the `z` dimension,
-        #  when we modify the block sizes.
-        dims_to_inspect = len(map_size) if len(map_size) <= 3 else 2
+            # If there are more than three dimensions DaCe will condense them into
+            #  the `z` dimension of the block, so we have to ignore the `z` dimension,
+            #  when we modify the block sizes.
+            if num_map_params > 3:
+                dims_to_inspect = 2
 
+        # block size can only have up to three dimensions
+        assert dims_to_inspect <= 3
         # Cut down the block size.
         # TODO(phimuell): Think if it is useful to also modify the launch bounds.
         # TODO(phimuell): Also think of how to connect this with the loop blocking.
         for i in range(dims_to_inspect):
-            if i > 2:  # Block size can have only three dimensions.
-                break
             map_dim_idx_to_inspect = len(gpu_map.params) - 1 - i
             if (map_size[map_dim_idx_to_inspect] < block_size[i]) == True:  # noqa: E712 [true-false-comparison]  # SymPy Fancy comparison.
                 block_size[i] = map_size[map_dim_idx_to_inspect]


src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -123,6 +123,9 @@ def make_dace_backend(
         # Backend supports gpu memory pool only on CUDA target.
         optimization_args["gpu_memory_pool"] = False
 
+    optimization_args["gpu_block_size"] = (32, 8, 1)
+    optimization_args["gpu_block_size_1d"] = (256, 1, 1)
+
```
edopao: These lines are actually overriding the optimization parameters passed as arguments. The place for setting default parameters is the backend factory in icon4py, but it is not merged yet: https://github.com/C2SM/icon4py/pull/897

iomaganaris: I see. Not sure what we should do for now though. Should we keep this here for now until we can specialize this in `icon4py` or should we do remove that from here and keep the non optimal current default values for a while?
The change in the other file should be merged any way I think

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -653,32 +653,36 @@ def apply(
...
+                num_map_params += 1  # Handle 2D maps where one dimension has range 1 as 1D map
 
         # Because of a particularity of the DaCe code generator, the iteration
         #  variable that is associated to the `x` dimension of the block is the
         #  last parameter, i.e. `gpu_map.params[-1]`. The one for `y` the second last.
         if num_map_params == 1:
             block_size = list(self.block_size_1d)
             launch_bounds = self.launch_bounds_1d
-            dims_to_inspect = 1
+            for node in graph.scope_subgraph(
```
philip-paul-mueller: I would only do something if `self.launch_bounds_1d` is not specified, i.e. `None`.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -653,32 +653,36 @@ def apply(
...
+                self.map_entry, include_entry=False, include_exit=False
+            ):
+                if isinstance(node, dace_nodes.NestedSDFG) and "scan" in node.label:
```
philip-paul-mueller: You should document this behaviour in the doc string.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -653,32 +653,36 @@ def apply(
         """Modify the map as requested."""
         gpu_map: dace_nodes.Map = self.map_entry.map
         map_size = gpu_map.range.size()
-        num_map_params = len(gpu_map.params)
+        num_map_params = 0
+        dims_to_inspect = 0
+        for axis_size in map_size:
+            dims_to_inspect += 1
+            if axis_size != 1:
+                num_map_params += 1  # Handle 2D maps where one dimension has range 1 as 1D map
```
philip-paul-mueller, line 666: This is a good idea but I have a question.
Consider these two Maps:
- Map 1: `[i=0:32, j=2]`
- Map 2: `[i=0, j=0:32]`

In both cases this would mean that the block size `(32, 1, 1)` is used.
Which is good for Map 1, but I just want to make sure that it is also okay for Map 2, i.e. if `(1, 32, 1)` is not better?
But I guess  the answer is "no, because `(1, 32, 1)` will effectively occupy a block of size `(32, 32, 1)`".

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
edopao:
```suggestion
    # Set default values for the block size, though these can be customized on the application side.
    if "gpu_block_size" not in optimization_args:
        optimization_args["gpu_block_size"] = (32, 8, 1)
    if "gpu_block_size_1d" not in optimization_args:
        optimization_args["gpu_block_size_1d"] = (64, 1, 1)

```

philip-paul-mueller: You should do that also for 2d and 3d.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
iomaganaris: It's a valid observation however I don't think it should appear or it would matter in reality.
If I'm not mistaken for GPU maps we always set the first dimension to be the horizontal and the second the vertical. Based on this I expect always the horizontal dimension to be larger than the vertical. If the horizontal range is 1 then I would expect anyway bad performance since the memory accesses are not going to be coalesced (since the fields are ordered with horizontal elements being next to each other). In case we have this case `(1, 32, 1)` would be more performant (we would still one warp of 32 threads per threadblock).
If you think we could end up having a case like `Map 2` in our SDFGs we could somehow handle that but it would have to change the logic of how `block_size`s are set

philip-paul-mueller: I actually think that your observation is correct and having Map 2 is very unlikely.
Maybe put a note there, but I think we can lave it that way for the time being.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_gpu_utils.py
```diff
@@ -226,5 +226,5 @@ def test_set_gpu_properties(method: int):
     #  there are more than three dim, it will not regulate the z block.
     #  Even though `map4` is a 4D Map, the launch bound value for 3D map is used.
     assert len(map4.params) == 4
-    assert map4.gpu_block_size == [9, 10, 12]
+    assert map4.gpu_block_size == [9, 10, 10]
```
iomaganaris: @philip-paul-mueller Shouldn't this have been `10` instead of `12` in the first place or have I misunderstood it?

philip-paul-mueller: Because the Map is a a ~~3d~~ 4d Map, thus the last element will be ignored.
The adjustment only takes place for Maps of at least 3 dimensions.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -653,32 +654,38 @@ def apply(
...
         if num_map_params == 1:
             block_size = list(self.block_size_1d)
             launch_bounds = self.launch_bounds_1d
-            dims_to_inspect = 1
+            if launch_bounds is None:
+                for node in graph.scope_subgraph(
+                    self.map_entry, include_entry=False, include_exit=False
+                ):
+                    if isinstance(node, dace_nodes.NestedSDFG) and "scan" in node.label:
+                        launch_bounds = "512"  # Use high launch bound in case of scans to limit register usage and increase occupancy
```
edopao, line 679: This is an optimization, therefore I would not put it here (a utils module). I propose to traverse the SDFG in `_gt_auto_configure_maps_and_strides()` and apply the `launch_bounds=512` there.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_gpu_utils.py
iomaganaris: I'm afraid I haven't understood based on your comment what should be the correct value here 🙈 
`map4` range is `[10, 10, 10, 10]` IIUC and the selected thread block size is `[9, 11, 12]`. I expect then that the actual thread block size will become `[9, 10, 10]`. Is this correct?

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
iomaganaris: I handled this in a different way. I hope it's fine

edopao: Yes, even better.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
iomaganaris: I would argue that it's better to keep setting anything related to setting thread block size and launch bounds to a single place. `GPUSetBlockSize` in this case. Maybe this transformation should be moved from `gpu_utils.py` to its own file to make it more visible.
What do you think?

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -635,6 +635,7 @@ def can_be_applied(
         - If the block size of the map is already set.
         - If the map is at global scope.
         - If if the schedule of the map is correct.
+        - If launch_bounds_1d is not set then set the launch_bounds of maps with `scan`s to 512 to limit their register usage.
```
edopao:
```suggestion
        - If launch_bounds_1d is not set then set the launch_bounds of `scan` maps to 512 to limit their register usage.
```

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
edopao: OK, we can keep it here. However, this code requires #2315, that is avoid inlining scan `NestedSDFG` objects. @philip-paul-mueller we should merge your PR first.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_gpu_utils.py
philip-paul-mueller: It is not modified because the last  value of the CUDA block size is used for the two last dimensions of the Map.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
philip-paul-mueller: The PR Edoardo mentioned sets the `noinline` property of nested SDFGs which are scans to `False` thus they are never inlined.
But for this particular case my PR is not relevant because _this_ PR explicitly targets the case where the scan is _inside_ a Map and it is (currently and probably forever) impossible to inline a nested SDFG that is inside a Map.
The only thing that I would suggest here is that the test might not be enough and I would propose to test for `startswith("scan_")` and hope that you only add prefixes.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -653,32 +654,38 @@ def apply(
...
             block_size = list(self.block_size_2d)
             launch_bounds = self.launch_bounds_2d
-            dims_to_inspect = 2
         else:
             block_size = list(self.block_size_3d)
             launch_bounds = self.launch_bounds_3d
-            # If there are more than three dimensions DaCe will condense them into
-            #  the `z` dimension of the block, so we have to ignore the `z` dimension,
-            #  when we modify the block sizes.
-            dims_to_inspect = 3 if num_map_params == 3 else 2
```
philip-paul-mueller, line 675: This here is the reason why you had to modify the test.
I agree the if condition would be better if it was something like:
```python
dims_to_inspect = 2 if num_map_params > 3 else 3
```

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
edopao: I see, thanks for the clarification. I also agree about the string test proposed by Philip.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
iomaganaris: Ah I missed that somehow 🙈 Thanks for pointing that out

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -653,32 +654,42 @@ def apply(
...
+
+        # If there are more than three dimensions DaCe will condense them into
+        #  the `z` dimension of the block, so we have to ignore the `z` dimension,
+        #  when we modify the block sizes.
+        dims_to_inspect = len(map_size) if len(map_size) <= 3 else 2
```
philip-paul-mueller: This can not be here, because here it also affects 1D Maps.

iomaganaris: I added a few more checks. Hopefully they make sense

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -112,8 +112,8 @@ def gt_auto_optimize(
     gpu: bool,
     unit_strides_kind: Optional[gtx_common.DimensionKind] = None,
     make_persistent: bool = False,
-    gpu_block_size: Optional[Sequence[int | str] | str] = None,
-    gpu_block_size_1d: Optional[Sequence[int | str] | str] = None,
+    gpu_block_size: Optional[Sequence[int | str] | str] = (32, 8, 1),
+    gpu_block_size_1d: Optional[Sequence[int | str] | str] = (64, 1, 1),
```
philip-paul-mueller, line 116: I am not sure if we should set them here as they are quite specific to the problem, but I am not against it.
The best choice is probably to not provide a default, i.e. the user has to specify it explicitly.
But for the time being we can probably go this way.

philip-paul-mueller (approved):
LGTM

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
iomaganaris: I think for now it's not too bad to set these here. When we have the ability to set these from `icon4py` we can move these values there

PR #2314, opened 2025-10-16
feat[cartesian]: 2D temporaries [Experimental]

FlorianDeconinck:
## Description

Canonically, temporaries have been 3D arrays of float. We introduced a few months back the capacity to type hint the type to allow for `integer` and mixed precision temporaries.

Another missing feature is the capacity to ask for a 2D temporary when the semantic of the code asks for it (aggregate, mask, etc.). We introduce with two flavors:

- The default behavior knows how to resolve the `FieldDescriptor` type e.g.
```python
tmp: Field[IJ, np.float64] = ...
```
- A user defined mapping from symbols (as string)  to type allows for applications on top of GT4Py to control the verbosity of the syntax
```python
@stencil(backend=..., dtype={'MyType': Field[IJ, np.float64]})
def stencil(...):
   ...
   tmp: MyType = ...
```

The feature is limited to 2D temporaries for now - and guarded for it.

The feature is experimental because support for `gt:X` backends is missing.

## Requirements


FlorianDeconinck:
cscs-ci run

src/gt4py/cartesian/frontend/nodes.py
```diff
@@ -199,6 +200,20 @@ class Domain(Node):
     parallel_axes = attribute(of=ListOf[Axis])
     sequential_axis = attribute(of=Axis, optional=True)
 
+    @classmethod
+    def from_gtscript(cls, gt_axis: list[gt_script.Axis]):
+        sequential_axis = None
+        parallel_axes = []
+        for axis in gt_axis:
+            if axis == "I" or "J":
```
romanc: I think this should be
```suggestion
            if axis == "I" or axis == "J":
```
or
```suggestion
            if axis in ("I", "J"):
```

src/gt4py/cartesian/frontend/nodes.py
```diff
@@ -199,6 +200,20 @@ class Domain(Node):
     parallel_axes = attribute(of=ListOf[Axis])
     sequential_axis = attribute(of=Axis, optional=True)
 
+    @classmethod
+    def from_gtscript(cls, gt_axis: list[gt_script.Axis]):
```
romanc:
```suggestion
    def from_gtscript(cls, gt_axis: list[gt_script.Axis]) -> Self:
```

src/gt4py/cartesian/gtc/debug/debug_codegen.py
```diff
@@ -202,20 +202,31 @@ def visit_Interval(self, interval: oir.Interval, **kwargs) -> str:
...
+        if temporary_declaration.dimensions[2]:
+            shape.append("k_size")
+            offset.append("0")
+
+        # Data dimensions
         data_dimensions: list[str] = [str(dim) for dim in temporary_declaration.data_dims]
-        shape = shape + data_dimensions
-        shape_decl = ", ".join(shape)
+        if len(temporary_declaration.data_dims) > 0:
+            offset += ["0"] * len(temporary_declaration.data_dims)
```
romanc: can't this be just

```suggestion
        offset += ["0"] * len(temporary_declaration.data_dims)
```

and then python is smart and doesn't do anything if `len(...) == 0`?

src/gt4py/cartesian/gtc/gtcpp/gtcpp.py
```diff
@@ -95,6 +95,7 @@ class Cast(common.Cast[Expr], Expr):
 class Temporary(LocNode):
     name: eve.Coerced[eve.SymbolName]
     dtype: common.DataType
+    dimensions: tuple[bool, bool, bool] = (True, True, True)
```
romanc, line 98: I think we shouldn't touch `gtcpp.py` if we don't add support for 2d temporaries in `gt:*` backends

src/gt4py/cartesian/gtc/numpy/npir.py
```diff
@@ -68,11 +68,14 @@ class TemporaryDecl(Decl):
 
     Parameters
     ----------
+    data_dims: Data dimensions sizes (static)
+    dimensions: Flag which cartesian dimensions is present
```
romanc:
```suggestion
    dimensions: Mask which cartesian dimensions are present
```
For me, a flag is for one value and a mask is a collection of flags.

src/gt4py/cartesian/gtc/numpy/npir_codegen.py
```diff
@@ -88,12 +88,20 @@ def add_declared(self, *args):
...
+        shape = [f"_dI_ + {node.padding[0]}", f"_dJ_ + {node.padding[1]}"]
+        offset = [str(off) for off in node.offset]
+        # Vertical dimension K
+        if node.dimensions[2]:
+            shape += ["_dK_"]
+            offset += ["0"]
+        # Data dimensions
+        if len(node.data_dims) > 0:
+            shape += [str(dim) for dim in node.data_dims]
+            offset += ["0"] * len(node.data_dims)
```
romanc:
```suggestion
        shape += [str(dim) for dim in node.data_dims]
        offset += ["0"] * len(node.data_dims)
```
python should be smart enough to not do anything if `len(node.data_dims)` is 0

tests/cartesian_tests/unit_tests/test_gtc/gtcpp_utils.py
```diff
@@ -170,6 +170,7 @@ class Meta:
         model = gtcpp.Temporary
 
     name = identifier(gtcpp.Temporary)
+    dimensions = (True, True, True)
```
romanc, line 173: I think we shouldn't need to touch the `gt:*` backend if we don't add support in that backend

FlorianDeconinck: We do because `Temporaries` in ` gtc` and `numpy` share a `TemporaryFactory`. The other option is to break that code share or split the code when the factory generates (which is a bit anti-factory).

Meanwhile, this is a legit, though unused, addition that scopes the dims properly.

src/gt4py/cartesian/gtc/gtcpp/gtcpp.py
FlorianDeconinck: See below

src/gt4py/cartesian/gtc/debug/debug_codegen.py
FlorianDeconinck: Yeah it does, good catch

fthaler:
> Can you give a concrete example? I think we concluded at some point that 2D temporaries are not useful, but don't remember why...

AFAIR, we just concluded that for all use cases they can just easily be replaced by ‘normal’ 2D fields, i.e., we can’t do useful optimizations with them (in contrast to the 3D ones which can be cached) and they don’t require extent analysis (as we can’t fuse them with following 3D computations due to the synchronization issues). So we decided the advantage of having them is not worth the implementation effort as we won’t gain any performance.

> I might add that it's a fine answer to us that GridTools, the C++ lib, will not provide those. We can create the necessary infrastructure in GT4Py, which by it's frontend knows that those Field would be accessed badly and therefore can commit to their dimensions.

From the GridTools point of view that’s probably the easiest solution.

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
+# ⚠️ 2 Dimenions Temporaries
+
+In the context of porting physics parametrizations, we have encoutered multiple examples of computation requiring a temporary 2D storage within a stencil.
```
romanc:
```suggestion
In the context of porting physics parametrizations, we have encountered multiple examples of computations requiring a temporary 2D storage within a stencil.
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+
+We thus decided to expand the DLS be able to port these stencils. We accept the increase in DSL surface and test it as an [experimental feature](../experimental-features.md).
+
+## Context
+
+Porting physics parametrizations, we found temporary 2D field used, mostly when computing particular level (surface, phase change, etc.). Previous workaround was to define the temporaries as an argument. The goal is to undo this workaround that complexify calling code.
```
romanc:
```suggestion
Porting physics parametrizations, we found temporary 2D fields used, mostly when computing particular levels (e.g. surface, phase change, etc.). The previous workaround was to define the temporaries outside and pass them as an argument. The goal is to undo this workaround that complexifies calling code.
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+
+Use case in GEOS's shallow convection parametrization, `total_pwo_solid_phase` will be used later in the stencil but not out of it.
```
romanc:
```suggestion
Use case in GEOS's shallow convection parametrization: `total_pwo_solid_phase` will be used later in the stencil but not outside of it.
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+            interpolated_field = field
+
+    # reset masks and temporaries for later use
+    with computation(FORWARD), interval(0, 1):
+        boolean_2d_mask = False
+        pb = 0
+        pt = 0
+```
+
+2D temporaries would clean those code and allow for futher optimization (scalarization) in all performance backend.
```
romanc:
```suggestion
2D temporaries would clean those codes and allow for further optimizations (e.g. scalarization) in all performance backends.
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Decision
+
+All the guardraisls for 2D temporaries are already in place:
```
romanc:
```suggestion
All the guardrails for 2D temporaries are already in place:
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+
+- allocation can only be done under `interval == 1` and `computation` in `FORWARD` or `BACKWARD`,
+- use of 2D fields are fully defined, using the above rules again.
+
+The main change is on the frontend of stencils and proper forwarding of dimensions through to code genreation, except for `gt:X` backend (see [Consequences](#consequences)).
+
+Type hint for temporaries have been introduced before for mixed precision e.g.:
```
romanc:
```suggestion
Type hints for temporaries have been introduced before for mixed precision e.g.:
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+
+```python
+with computation(PARALLEL), inteval(...):
```
romanc:
```suggestion
with computation(PARALLEL), interval(...):
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+    tmp_3D_as_f32: float32 = 0
+```
+
+We propose to extend this type hint to allow specification of the dimensions re-using the `FieldDescriptor`, e.g.
+
+```python
+with computation(FORWARD), inteval(...):
```
romanc:
```suggestion
with computation(FORWARD), interval(...):
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+    tmp_2D_as_f32: Field[IJ, np.float64] = 0
+```
+
+We also guard against any definition of other type of temporaries (e.g. 1D temporary) because they are ill-defined within `gtscript` which pre-suppose that the horizontal dimensions are always computed upon.
```
romanc:
```suggestion
We also guard against any definition of other type of temporaries (e.g. 1D temporaries) because they are ill-defined within `gtscript` which pre-supposes that the horizontal dimensions are always computed upon.
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+
+The type hint can be a little verbose, so we offer to expand upon the `dtypes` dictionnary present in stencil configuration to give a `str: type` pair that can be swapped at parsing time, as long as the type is a derivative of `FieldDescriptor` so the relevant informations can be retrieved.
```
romanc:
```suggestion
The type hint can be a little verbose, so we offer to expand upon the `dtypes` dictionary present in stencil configuration to give a `str: type` pair that can be swapped at parsing time, as long as the type is a derivative of `FieldDescriptor` so the relevant information can be retrieved.
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+
+```python
+@stencil(backend=..., dtype={"My2DType": Field[IJ, np.float64]})
+def the_stencil(...):
+    with computation(FORWARD), inteval(...):
```
romanc:
```suggestion
    with computation(FORWARD), interval(...):
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+        tmp_2D: My2DType = 0
+        ...
+```
+
+## Consequences
+
+Users of GT4Py can now defined 2D temporaries in-stencil.
+
+There's one remaining hiccups: GridTools C++ does not natively offer 2D temporaries. GridTools pre-suppose that all computation of temporaries are done on the 3D grid at least - and won't allocate below that dimensionality.
```
romanc:
```suggestion
There's one remaining hiccup: GridTools C++ does not natively offer 2D temporaries. GridTools pre-supposes that all computations of temporaries are done on the 3D grid at least - and won't allocate below that dimensionality.
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
...
+This is a fair limitation since GridTools doesn't provide the guardrails against race condition that GT4Py does. We can circumvent this by creating a pool of buffers passed as arguments for `gt:X` backends - see [issue](https://github.com/GridTools/gt4py/issues/2322).
+
+## References
+
+- [PR](https://github.com/GridTools/gt4py/pull/2314) PR where the proposal and the discussion occured.
```
romanc:
```suggestion
- [PR](https://github.com/GridTools/gt4py/pull/2314) PR where the proposal and the discussion occurred.
```

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -1607,20 +1615,54 @@ def _resolve_assign(
...
+                                    "2D temporaries (e.g. `tmp: Field[IJ, float] = ...) is an experimental feature "
+                                    "and not yet implemented for the `gt:X` backends."
+                                )
+                            warnings.warn(
+                                "2D temporaries is an experimental feature. Please read "
+                                "<https://github.com/GridTools/gt4py/blob/main/docs/development/ADRs/cartesian/experimental-features.md> "
+                                "to understand the consequences.",
+                                category=UserWarning,
+                                stacklevel=2,
+                            )
```
romanc: Once https://github.com/GridTools/gt4py/pull/2324 merges, you can use that

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,134 @@
+# ⚠️ 2 Dimenions Temporaries
```
romanc:
```suggestion
# ⚠️ Two-dimensional Temporaries
```

src/gt4py/cartesian/frontend/nodes.py
```diff
@@ -143,6 +143,7 @@
 
 import numpy as np
 
+import gt4py.cartesian.gtscript as gt_script
```
romanc: Re-reading this, I asked myself: Is it okay that `cartesian.frontend.nodes` depends on `cartesian.gtscript`? That seems odd, no? Shouldn't `gtscript` depend on `nodes`, but not the other way around? Since it's a `classmethod`, maybe we could just move it to a regular function in `gtscript_frontend.py` that creates a `Domain` from `gtscript` to avoid the link here. @havogt & @egparedes might know more about how the dependencies should be.

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
twicki: I think this is a very convoluted example. In the end we do a simple integration up a column, a very common thing in most codes. The big question that need answers (especially from the gt-side) is why this needs to be a temporary and can't be something else. 
I agree that extent-analysis is not really relevant, as column-based stuff does not go out of bounds ususally. But mentioning that making variables temporary might open up optimization-potential (with pooling for example) might be very relevant here

twicki: an example like this is already making a case for 2d-variables:
image

We specifically need to propagate  bb[i,j,0] everywhere because the 3d accessor below needs acces to it and it can't be a 2d-temp

src/gt4py/cartesian/utils/field.py
```diff
@@ -28,8 +28,8 @@ def __init__(
         self.offsets = offsets
 
     @classmethod
-    def empty(cls, shape, dtype, offset):
-        return cls(np.empty(shape, dtype=dtype), offset, (True, True, True))
+    def empty(cls, shape, dtype, offset, dims: tuple[bool, bool, bool]):
```
twicki, line 31: why do we remove the "default" argument (True, True, True) here?

src/gt4py/cartesian/frontend/nodes.py
FlorianDeconinck: I don't think the `script` should depend on the `nodes`. And vice-versa. Moving it into the `frontend` processor.

src/gt4py/cartesian/utils/field.py
FlorianDeconinck: Because we need the `Field` to respect the dimensions and I have run into a number of "defaults" that are difficult to track when we move to other dimensions. I'd rather have dimensions passed through proper.

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
FlorianDeconinck: Those examples are from actual code - which I think is always superior. I'll add the one you pointed out.

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,146 @@
...
+
+In the context of porting physics parametrizations, we have encountered multiple examples of computations requiring a temporary 2D storage within a stencil.
+
+We thus decided to expand the DLS be able to port these stencils. We accept the increase in DSL surface and test it as an [experimental feature](../experimental-features.md).
+
+## Context
+
+Porting physics parametrizations, we found temporary 2D fields used, mostly when computing particular levels (e.g. surface, phase change, etc.). The previous workaround was to define the temporaries outside and pass them as an argument. The goal is to undo this workaround that complexifies calling code.
+
+Another workaround, driven by how GridTools C++ behave, we forced all temporaries to be 3D. The way to do it would be to create you temporary and then propagate it's value throughout. E.g., from FV3 dynamics code:
```
romanc:
```suggestion
Another workaround, driven by how GridTools C++ behaves, is to force all temporaries to be 3D. The way to do it would be to create you temporary and then propagate it's value throughout. E.g., from FV3 dynamics code:
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,146 @@
...
+2D temporaries would clean those codes and allow for further optimizations (e.g. scalarization) in all performance backends.
+
+## Decision
+
+All the guardrails for 2D temporaries are already in place:
+
+- allocation can only be done under `interval == 1` and `computation` in `FORWARD` or `BACKWARD`,
+- use of 2D fields are fully defined, using the above rules again.
+
+The main change is on the frontend of stencils and proper forwarding of dimensions through to code genreation, except for `gt:X` backend (see [Consequences](#consequences)).
```
romanc:
```suggestion
The main change is on the frontend of stencils and proper forwarding of dimensions through to code generation, except for `gt:X` backend (see [Consequences](#consequences)).
```

docs/development/ADRs/cartesian/experimental/2d-temporaries.md
```diff
@@ -0,0 +1,146 @@
...
+@stencil(backend=..., dtype={"My2DType": Field[IJ, np.float64]})
+def the_stencil(...):
+    with computation(FORWARD), interval(...):
+        tmp_2D: My2DType = 0
+        ...
+```
+
+## Consequences
+
+Users of GT4Py can now defined 2D temporaries in-stencil.
```
romanc:
```suggestion
Users of GT4Py can now define 2D temporaries inside stencils.
```

romanc:
cscs-ci run

PR #2315, opened 2025-10-17
feat[dace][next]: Block Inlining of NestedSDFG Containing Scans

philip-paul-mueller:
Prevent a possible Bug in DaCe from manifesting, see [issue#2182](https://github.com/spcl/dace/issues/2182).
The solution is to block the inlining of NestedSDFG that contains scans.

TODO:
- [ ] Ensure no performance degradation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ dace-cartesian = [
   'dace>=1.0.2,<2'  # renfined in [tool.uv.sources]
 ]
 dace-next = [
-  'dace==2025.10.23'  # refined in [tool.uv.sources]
+  'dace==2025.10.28'  # refined in [tool.uv.sources]
 ]
 dev = [
   {include-group = 'build'},
@@ -453,7 +453,7 @@ url = 'https://test.pypi.org/simple'
 atlas4py = {index = "test.pypi"}
 dace = [
   {git = "https://github.com/GridTools/dace", branch = "romanc/stree-roundtrip", group = "dace-cartesian"},
-  {git = "https://github.com/GridTools/dace", tag = "__gt4py-next-integration_2025_10_23", group = "dace-next"}
+  {git = "https://github.com/GridTools/dace", tag = "__gt4py-next-integration_2025_10_28", group = "dace-next"}
 ]
 
 # -- versioningit --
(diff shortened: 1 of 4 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
```diff
@@ -262,6 +262,21 @@ def gt_inline_nested_sdfg(
...
+        # Block the inlining of NestedSDFG containing a scan.
+        #  we do this to avoid a bug in DaCe simplify transformations, see
+        #  [issue#2182](https://github.com/spcl/dace/issues/2182) for more.
+        #  Before the bug was hidden but once we started running `TrivialTaskletElimination`
+        #  it was no longer the case. The solution is to block the inlining and keep
+        #  scans localized inside their NestedSDFG.
+        # NOTE: Currently there is no transformation that takes advantages in any way
+        #   of a scan and they are mostly inside Maps anyway, except of unit tests,
+        #   where inlining is not possible anyway.
+        if nsdfg_node.label.startswith("scan_"):
```
edopao: Should it maybe be more appropriate to set `no_inline` on the `NestedSDFG` node during lowering?
```
    no_inline = Property(dtype=bool,
                         desc="If True, this nested SDFG will not be inlined during "
                         "simplification",
                         default=False)
```

src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
```diff
@@ -283,6 +284,12 @@ def gt_inline_nested_sdfg(
         if multi_state_inliner.can_be_applied(parent_state, 0, parent_sdfg, permissive=permissive):
             multi_state_inliner.apply(parent_state, parent_sdfg)
             nb_inlines_total += 1
+            if nsdfg_node.label.startswith("scan_"):
+                # See `gtir_to_sdfg_scan.py::translate_scan()` for more information.
+                warnings.warn(
+                    f"Inlined '{nsdfg_node.label}' which might be a scan, this might leads to errors during simplification.",
+                    stacklevel=0,
+                )
```
edopao, line 310: I guess you want to keep this warning. I personally would avoid it because the SDFG name is a weak link, but I am OK to keep it.

src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
philip-paul-mueller: This is a good idea

PR #2318, opened 2025-10-17
fix[next][dace]: Avoid name conflict between tasklet connector and data node

edopao:
This fix replaces the previous change in #2306, which was reverted, to avoid name conflicts between tasklet connector and data node.

The previous PR was quite invasive, and introduced some field validation error in ICON4Py. This PR uses a smaller change, very local to the source of the name conflict: `__arg` is used in GTIR lowering for expression arguments, which are lowered in the SDFG as data nodes. Therefore, these names can collide with the tasklet connectors in the same graph. To avoid this issue, we use a different naming convention for the tasklet arguments.


src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -1723,7 +1723,7 @@ def _visit_generic_builtin(self, node: gtir.FunCall) -> ValueExpr:
             arg_expr = self.visit(arg)
             if isinstance(arg_expr, MemletExpr | ValueExpr):
                 # the argument value is the result of a tasklet node or direct field access
-                connector = f"__arg{i}"
+                connector = f"__var{i}"
```
philip-paul-mueller, line 1726: Have you considered to change all `__arg` to `__var` except for nested SDFGs?
I think this should pretty much remove all issues, regardless if we see it in the unit tests or not.

edopao: Yes, I've checked that. There are no other matches.

PR #2321, opened 2025-10-20
fix[next][dace]: SDFG instrumentation with empty else-branch caused memory leak

edopao:
This PR fixes a bug introduced by #2280. The previous PR used a conditional interstate edge to conditionally execute a sink state, which contained the timer tasklet to produce the stencil compute time. However, there was no else-branch, and no unique sink node in the SDFG. If metrics collection was enabled, the new state would be the sink, otherwise the predecessor state. The lack of a unique sink node caused some error in SDFG transformations that led to memory leak, because the GPU temporary arrays were not properly freed.
This PR builds two conditional blocks, one as unique source and another one as unique sink node for the SDFG. Inside the conditional block, metrics collection is represented as a branch with conditional execution. In this way, the SDFG source and sink nodes are always the same, no matter whether metrics collection is enabled or not.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
@@ -193,7 +193,7 @@ def add_instrumentation(sdfg: dace.SDFG, gpu: bool) -> None:
         has_side_effects = False
 
     #### 2. Timestamp the SDFG entry point.
-    entry_if_region, entry_state = _make_if_region_for_metrics_collection(
+    entry_if_region, begin_state = _make_if_region_for_metrics_collection(
         "program_entry", metrics_level, sdfg
     )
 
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -151,6 +151,24 @@ def add_instrumentation(sdfg: dace.SDFG, gpu: bool) -> None:
...
+
+    def make_if_region(
+        name: str,
+    ) -> tuple[dace.state.ConditionalBlock, dace.state.SDFGState, dace.state.SDFGState]:
+        if_region = dace.sdfg.state.ConditionalBlock(name)
+        sdfg.add_node(if_region, ensure_unique_name=True)
+        then_body = dace.sdfg.state.ControlFlowRegion(f"{if_region.label}_then_body", sdfg=sdfg)
+        then_state = then_body.add_state(f"{if_region.label}_then_branch")
+        if_region.add_branch(
+            dace.sdfg.state.CodeBlock(f"{metrics_level} >= {metrics.PERFORMANCE}"), then_body
```
havogt: I find it slightly confusing that we rely on late-binding of `metrics_level`. Would probably be cleaner if this would be passed as argument, then it could also be defined outside of the function body.

havogt (approved):
Are you positive that this doesn't add overhead in the case where we don't measure performance? I have no feeling for that...

I verified that the memory leak is gone.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
edopao: Good idea, fixed.

PR #2323, opened 2025-10-20
feat[cartesian]: IntEnum support

FlorianDeconinck:
## Description

We  introduced the (strict) support for `IntEnum` in code and arguments of stencils. The strategy is to replace them with proper integer as soon as possible.

Guardrails exists for name clash and `IntEnum` derivation. 

Valid code looks like this

```python
@gtscript.enum
class LocalEnum(IntEnum):
    A = 42
    B = 1000

@gtscript.stencil
def enum(field: gtscript.Field[float], order: LocalEnum):  # type: ignore
    with computation(PARALLEL), interval(0, 1):
        if order > LocalEnum.A:
            field[0, 0, 0] = LocalEnum.B
```

TO CLEAN
- [ ] Register of Enum types should not leave in `gtscript`
- [x] ADR
- [x] Mixed precision

## Requirements


src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -1899,6 +1911,8 @@ def annotate_definition(
                 and param.annotation in gtscript._VALID_DATA_TYPES
             ):
                 dtype_annotation = np.dtype(param.annotation)
+            elif param.annotation in gtscript._ENUM_REGISTER.values():
+                dtype_annotation = int  # We will replace all enums with `int`
```
twicki: just to double-check: This works with int32 / int64 and default low-precision as expected?

src/gt4py/cartesian/gtscript.py
```diff
@@ -159,6 +160,26 @@ def _parse_annotation(arg, annotation):
...
 
 
+_ENUM_REGISTER: dict[str, object] = {}
+"""Register of IntEnum that will be available to parsing in stencils. Register
+with @gtscript.enum()"""
+
+
+def enum(class_: type[IntEnum]):
+    class_name = class_.__name__
+    if class_name in _ENUM_REGISTER:
```
twicki: I think this works and makes for some centralization!

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -309,6 +309,18 @@ def visit_ImportFrom(self, node: ast.ImportFrom):
...
+        #   - value: ast.Name("MyEnum")
+        # We want to replace the entire thing - so we capture the top level
+        # attribute. We don't use the `self.context` because of thise
+        # two-step AST structure which doesn't fit the generic `replace_node`
+
+        if isinstance(node.value, ast.Name) and node.value.id in gtscript._ENUM_REGISTER.keys():
+            int_value = getattr(gtscript._ENUM_REGISTER[node.value.id], node.attr)
+            return ast.Constant(value=int_value)
+
+        # Common replace for all other node in context
```
romanc:
```suggestion
        # Common replace for all other nodes in context.
```

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -309,6 +309,18 @@ def visit_ImportFrom(self, node: ast.ImportFrom):
         return node
 
     def visit_Attribute(self, node: ast.Attribute):
+        # An enum MyEnum.A would come has
+        # > ast.Attribute("A")
+        #   - value: ast.Name("MyEnum")
+        # We want to replace the entire thing - so we capture the top level
+        # attribute. We don't use the `self.context` because of thise
+        # two-step AST structure which doesn't fit the generic `replace_node`
```
romanc:
```suggestion
        # attribute. We don't use the `self.context` because of this
        # two-step AST structure which doesn't fit the generic `replace_node`.
```

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -2024,6 +2038,10 @@ def collect_external_symbols(definition):
         local_symbols = CollectLocalSymbolsAstVisitor.apply(gtscript_ast)
         nonlocal_symbols = {}
 
+        # De-pop from `context in Enum registered with `gtscript`
```
romanc: Maybe mention _why_ the enums are removed, e.g. (as far as I understand this is why)
```suggestion
        # Remove enums from `context` because they are transformed to integers.
```

src/gt4py/cartesian/gtscript.py
```diff
@@ -159,6 +160,26 @@ def _parse_annotation(arg, annotation):
...
+        raise ValueError(
+            f"Cannot register @gtscript.enum {class_name} as a class"
+            "with the same name is already registered."
```
romanc: I'd suggest a slightly less technical error message since this is user facing. The fact that we keep enums in a registry is an implementation detail and shouldn't need to bother users.

```suggestion
            f"Enum names must be unique. @gtscript.enum {class_name} is already taken."
```

romanc (commented):
I think this looks good as proof of concept. Some suggested wording changes inline.

src/gt4py/cartesian/frontend/gtscript_frontend.py
FlorianDeconinck: Probably not, added to clean list

src/gt4py/cartesian/gtscript.py
FlorianDeconinck: I don't like that `ENUM_REGISTER` is on `gtscript`, I will probably move it into it's own little space

FlorianDeconinck:
Test fails on 3.10, fine on 3.11. Looks like the `AST` parsing of `Enum` has moved

src/gt4py/cartesian/definitions.py
```diff
@@ -42,6 +42,21 @@
 """Default literal precision used for unspecific `float` types and casts."""
 
 
+def get_integer_default_type():
+    """Return the integer numpy type corresponding to the LITERAL_INT_PRECISION set."""
```
romanc: That won't work with the per-stencil overrides of `literal_integer_precision` that we allow. In the `GTScriptParser`, you have access to the `BuildOptions` in `self.options` which contains `literal_integer_precision`.  Likewise, in `annotate_definition()` of the `StencilObject` you get the `BuildOptions` as function argument, so you can get the precision from there.

FlorianDeconinck: Indeedy.

FlorianDeconinck:
cscs-ci run default

FlorianDeconinck:
cscs-ci run default

romanc (commented):
Good things will come to those who are patient ...

I think this feature warrants a frontend ADR. And I wonder if `FloatEnum` and `BoolEnum` should also be a thing or not. Why only integers?

FlorianDeconinck:
> Good things will come to those who are patient ...
> 
> I think this feature warrants a frontend ADR. And I wonder if `FloatEnum` and `BoolEnum` should also be a thing or not. Why only integers?

I guess they could. `FloatEnum` and `BoolEnum` hurts my little C feelings but if they are standards then...

FlorianDeconinck:
Hmm it's a bit more complex other type of Enum. It requires inner checks on the parameters to make sure they are "valid". I would move that to another ticket right now

FlorianDeconinck:
ADR added

docs/development/ADRs/cartesian/frontend/IntEnum.md
```diff
@@ -0,0 +1,34 @@
...
+def enum(field: gtscript.Field[float], order: MyEnum):  # type: ignore
+    with computation(PARALLEL), interval(0, 1):
+        if order > MyEnum.A:
+            field[0, 0, 0] = MyEnum.B
+```
+
+## Context
+
+`IntEnum` enforces a type consistency in the elements of `Enum` making it a safe way to express flags and integer constants. It is standard AST as of 3.11+. Supporting in `gt4py.cartesian` would allow a modern python expression to be supported and clean up code around a very common behavior of users to write in hardcoded flag values.
+
```
romanc, line 22: Maybe add here for context that there are other enum types which we might (float / bool) or might not (string) support in the future if a clear need arises from user code.

docs/development/ADRs/cartesian/frontend/IntEnum.md
```diff
@@ -0,0 +1,34 @@
...
+We will replace all instances of those enum in stencils by their constant values, allowing the implementation to be focused on the frontend and not downstream into the IRs.
+
+## Consequences
+
+`IntEnum` are considered frozen - if the value is swapped during runtime it won't be reflected.
+
+## Alternatives considered
+
+We could expand the system to all `Enum` by introducing a type check during registration of the enum in gt4py making sure the type of the parameters are legitimate in the context of `gt4py.cartesian`.
```
romanc, line 39: Maybe it would be worth adding the limitations of using constants / externals here: I'm thinking in the given example above, `MyEnum.A` and `MyEnum.B`  could be constants from a `constants.py` module, but order would have to be an `int` (i.e. not restricted to either `MyEnum.A`  or `MyEnum.B`).

src/gt4py/cartesian/definitions.py
```diff
@@ -42,6 +42,21 @@
 """Default literal precision used for unspecific `float` types and casts."""
 
 
+def get_integer_type(literal_integer_precision: int):
+    """Return the integer numpy type corresponding to the LITERAL_INT_PRECISION set."""
+    # I'd love to return `numpy.signedinteger[LITERAL_INT_PRECISION]` but that won't work
```
romanc, line 47: if you wanna use new python syntax, this is a case where `match` / `case` (see [docs](https://docs.python.org/3.12/tutorial/controlflow.html#match-statements)) would apply.

romanc (approved):
Looks good, if you want to increase test coverage, I'd try to cover

- failure cases for name clash and
- enums that are derived from the "wrong" base type (e.g. `StrEnum`)

FlorianDeconinck:
cscs-ci run default

FlorianDeconinck:
cscs-ci run default

PR #2327, opened 2025-10-22
feat[next][dace]: enable overriding of cxx/cuda compiler arguments

edopao:
Add the possibility to pass user-defined compiler arguments to dace build, by setting the environment variables `CXXFLAGS`/`CUDAFLAGS` for CPU/CUDA compiler.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/common.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/common.py
@@ -71,7 +71,7 @@ def set_dace_config(
         dace.Config.set("compiler.cpu.args", value=gt_cxxargs)
     else:
         dace.Config.set(
-            "compiler.cuda.args",
+            "compiler.cpu.args",
             value="-std=c++14 -fPIC -O3 -march=native -Wall -Wextra -Wno-unused-parameter -Wno-unused-label",
         )
     if gt_cudaargs := os.environ.get("CUDAFLAGS", None):


iomaganaris (approved):
There's a ~40% increase in the library size but the libraries are pretty small any way so I don't think there's a problem keeping this as a default.
LGTM!

iomaganaris:
The flags set in the point of this PR are used in DaCe to set `-DCMAKE_CUDA_FLAGS`. Since this CMake variable is set `CUDAFLAGS` are not taken into account. Maybe instead of adding this particular flag to the config we could just append the `CUDAFLAGS` env variable

src/gt4py/next/config.py
```diff
@@ -113,3 +113,8 @@ def env_flag_to_int(name: str, default: int) -> int:
 #: The default for whether to allow jit-compilation for a compiled program.
 #: This default can be overriden per program.
 ENABLE_JIT_DEFAULT: bool = env_flag_to_bool("GT4PY_ENABLE_JIT_DEFAULT", default=True)
+
+
+#: Extra user-defined compiler arguments to use in compiled backends.
+COMPILE_CXXFLAGS: str | None = os.environ.get("GT4PY_COMPILE_CXXFLAGS", None)
+COMPILE_CUDAFLAGS: str | None = os.environ.get("GT4PY_COMPILE_CUDAFLAGS", None)
```
havogt:
```suggestion
CXXFLAGS: str | None = os.environ.get("CXXFLAGS", None)
CUDAFLAGS: str | None = os.environ.get("CUDAFLAGS", None)
```
my proposal was this, because it will directly work for properly setup cmake projects.

But now I am not sure, because in DaCe we now violate how CMake uses these flags, because DaCe does `-DCMAKE_CXX_FLAGS=...` it should ignore the environment variable.

havogt: Maybe we just do this hack in the dace part and remove it from the config again.

edopao: I have now addressed this comment, I have reverted all changes in gt4py config.

src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -64,15 +65,21 @@ def set_dace_config(
...
-    dace.Config.set(
-        "compiler.cuda.args",
-        value="-Xcompiler -O3 -Xcompiler -march=native -Xcompiler -Wno-unused-parameter",
+    # The dace dafault settings use fast-math in both cpu and gpu compilation,
+    # we don't use it here. We also allow the user to append extra compiler arguments
+    # by means of environment variables.
+    cxxflags = (
+        "-std=c++14 -fPIC -O3 -march=native -Wall -Wextra -Wno-unused-parameter -Wno-unused-label"
     )
+    if gt_cxxargs := os.environ.get("CXXFLAGS", None):
```
edopao, line 70: Should we call these variables`EXTRA_CXXFLAGS` and `EXTRA_CUDAFLAGS`?

havogt: Then we should also enable this for the gt4py cmake build setup which currently works with `CXXFLAGS`/`CUDAFLAGS` because of standard cmake behavior (which I like).

edopao: @iomaganaris I propose keeping `CXXFLAGS`/`CUDAFLAGS`, in order to be consistent with gt4py, but not just adding, replacing the compiler args:
```
if gt_cxxargs := os.environ.get("CXXFLAGS", None):
    dace.Config.set("compiler.cpu.args", value=gt_cxxargs)
else:
    dace.Config.set(
        "compiler.cuda.args",
        value="-Xcompiler -O3 -Xcompiler -march=native -Xcompiler -Wno-unused-parameter",
    )
```

iomaganaris: I think that the last suggestion makes the most sense 👍

PR #2328, opened 2025-10-22
build[dace][next]: Updated DaCe Dependency

philip-paul-mueller:
Update of DaCe to the new tag [`__gt4py-next-integration_2025_10_23`](https://github.com/GridTools/dace/tree/__gt4py-next-integration_2025_10_23) of the fork.
It includes some fixes regarding the removal of redundant arrays.

Important: This is the first DaCe Tag that sets a version, which is always derived from the Tag.


edopao:
> > Important: This is the first DaCe Tag that sets a version, which is always derived from the Tag.
> 
> I haven't followed so far the development of our DaCe version so this might be a naive question but, since the DaCe branch we will point to now has a different version (`2025.10.22` if I understood correctly), shouldn't we also update the `dace-next` dependency to reflect it?
> 
> ```toml
> dace-next = [
>   'dace==2025.10.22'  # or 'dace>=2025.10.22' 
> ]
> ```

Good catch!

edopao (changes requested):
I need to include https://github.com/spcl/dace/pull/2191 in the dace integration branch.

edopao (approved):

PR #2329, opened 2025-10-22
feat[next][dace]: Targeted transformation to remove copies in ICON4Pys vertically implicit solvers

iomaganaris:
## Description
Handles the following case in the `vertically_implicit_solver`s stencils:
image
The transformation replaces the data of the second and third access nodes with the data of the first and fourth so that they can be simplified later and remove the redundant copies

## Requirements

- Needs to be added to `icon4py` as a hook for the vertically implicit solvers

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
@@ -6,6 +6,14 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
+# TODO(iomaganaris): This transformation is very specific to a pattern in the vertically_implicit_solver of ICON4Py.
+# The transformation should be generalized or moved to ICON4Py.
+# One idea to generalize this transformation is: in case we have a chain of 4 AccessNodes which are
+# connected sequentially with memlets (first_node -> second_node -> third_node -> fourth_node),
+# we can split the middle access nodes (second and third) based on their input and output memlets' subsets and those can later
+# be removed because the same data are copied from the first all the way to fourth node.
+# The split should be done using SplitAccessNode transformation.
+
 import warnings
 from typing import Any, Optional
 


src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,240 @@
...
+        first_node: dace_nodes.AccessNode = self.first_node
+        first_desc: dace_data.Data = first_node.desc(sdfg)
+        second_node: dace_nodes.AccessNode = self.second_node
+        second_desc: dace_data.Data = second_node.desc(sdfg)
+        third_node: dace_nodes.AccessNode = self.third_node
+        third_desc: dace_data.Data = third_node.desc(sdfg)
+        fourth_node: dace_nodes.AccessNode = self.fourth_node
+        fourth_desc: dace_data.Data = fourth_node.desc(sdfg)
+
+        # Match the pattern found in "vertically_implicit_solver"
```
philip-paul-mueller, line 90: Since it is super specific to ICON4Py I would also add the possibility to inject a name, i.e. that it checks that the global has a specific name.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,240 @@
...
+
+        # Make sure that second and fourth node have the same shape as the first node
+        if second_desc.shape != first_desc.shape or third_desc.shape != first_desc.shape:
+            warnings.warn(
+                f"[RemoveAccessNodeCopies] Shapes of second or third node do not match the shape of the first node: {second_desc.shape}, {third_desc.shape} vs {first_desc.shape}",
+                stacklevel=0,
+            )
+            # needs https://github.com/GridTools/gt4py/pull/2173 # return False
+
+        # Make sure that there is no other AccessNode with the same data in the SDFG state
```
philip-paul-mueller: I think this is not so much of a problem, but it only checks the current state and not the whole SDFG.
What you could do is something like:
```python
find_single_use_data = dace_analysis.FindSingleUseData()
single_use_data = find_single_use_data.apply_pass(sdfg, None)
if first_node.data not in single_use_data:
    return False
```

BTW the graph also has the `data_nodes()` function which iterates over all AccessNodes in a state.

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
```diff
@@ -215,6 +215,27 @@ def maybe_intersecting(
...
+    total_range = None
+    for r in ranges:
+        if not isinstance(r, dace_sbs.Subset):
+            return None
+        if total_range is None:
+            total_range = r
+        else:
+            total_range = dace_sbs.bounding_box_union(total_range, r)
+            if total_range is None or not isinstance(total_range, dace_sbs.Subset):
+                return None
```
philip-paul-mueller: Is this check necessary?
As far as I can see `bounding_box_union()` never returns `None` and will always return a `Range`.

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
```diff
@@ -215,6 +215,27 @@ def maybe_intersecting(
     return dace_sbs.intersects(sbs1, sbs2) is not False
 
 
+def union_of_subsets(ranges: Iterable[dace_sbs.Subset]) -> Optional[dace_sbs.Subset]:
+    """Computes the union of a list of ranges.
+    The function will compute the union of a list of ranges. If the
+    ranges are not compatible or if the union is not a range, the
+    function will return `None`.
+    Args:
+        ranges: The ranges whose union should be computed.
```
philip-paul-mueller:
```suggestion
    function will return `None`.
    
    Args:
        ranges: The ranges whose union should be computed.
```

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
```diff
@@ -215,6 +215,27 @@ def maybe_intersecting(
     return dace_sbs.intersects(sbs1, sbs2) is not False
 
 
+def union_of_subsets(ranges: Iterable[dace_sbs.Subset]) -> Optional[dace_sbs.Subset]:
```
philip-paul-mueller: You should extend the doc and explain how this function differs from `subset_merger()`.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,240 @@
...
+                "[RemoveAccessNodeCopies] The whole range of the first node is not written.",
+                stacklevel=0,
+            )
+            # needs https://github.com/GridTools/gt4py/pull/2173 # return False
+
+        # Make sure that data written to first node are only copied from first to second and from second to third node (they don't have to be rewritten to the last node)
+        first_node_out_edges = list(graph.out_edges(first_node))
+
+        # Make sure that all the edges from the first node go to the second node
+        if any(edge.dst != second_node for edge in first_node_out_edges):
```
philip-paul-mueller:
```suggestion
        # Make sure that all the edges from the first node go to the second node
        if any(edge.dst != second_node for edge in graph.out_edges(first_node)):
```
I am wondering is this variable is needed?
Also applies to `second_node_out_edges`.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,240 @@
...
+        graph: dace.SDFGState,
+        sdfg: dace.SDFG,
+    ) -> None:
+        first_node: dace_nodes.AccessNode = self.first_node
+        second_node: dace_nodes.AccessNode = self.second_node
+        third_node: dace_nodes.AccessNode = self.third_node
+
+        # Find the edge between first and second access nodes (there should be only one)
+        edge_between_first_and_second = next(
+            edge for edge in graph.edges() if edge.src == first_node and edge.dst == second_node
```
philip-paul-mueller:
```suggestion
            edge for edge in graph.edges_between(first_node, second_node)
```

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,240 @@
...
+        )
+
+        # Compute the offset of the data between first and second node
+        node_offset = [
+            range_first[0] - range_second[0]
```
philip-paul-mueller: I am not sure if it would improve anything here, but `Range` also has the `min_element()` function.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,240 @@
...
+                # Update the edge data to the first node data
+                edge.data.data = first_node.data
+                if edge.data.dst_subset != tuple([0]):
+                    if edge.data.dst_subset is not None:
+                        for i, dst_subset_range in enumerate(edge.data.dst_subset):
+                            edge.data.dst_subset[i] = (
+                                dst_subset_range[0] + node_offset[i],
+                                dst_subset_range[1] + node_offset[i],
+                                dst_subset_range[2],
+                            )
```
philip-paul-mueller: There is `Range.offset(node_o
```suggestion
                            edge.data.dst_subset.offset(node_offset, negative=False)
```

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,240 @@
...
+        # Compute the offset of the data between first and second node
+        node_offset = [
+            range_first[0] - range_second[0]
+            for range_first, range_second in zip(
+                edge_between_first_and_second.data.src_subset,
+                edge_between_first_and_second.data.dst_subset,
+            )
+        ]
+        # Update all edges with the correct subsets that match the first node range
+        for edge in graph.edges():
```
philip-paul-mueller, line 242: Sorry for this wall of text.
But here are some things are not right and kind of violate some internal DaCe things.

Iterating through the edges, as you do is not enough, in the general case (I am sorry if I told you it would be).
The reason for this is that you only update edges that are referring to the data of `second_node`, which does not need to be.
Doing it in general is tricky, but because of the special case we are targeting, we should get away a bit cheaper.
However, for that we need [PR#2189](https://github.com/spcl/dace/pull/2189) to be merged, which is currently blocked, but should soon be under way (however, make sure that `first_node` is on global scope, i.e. add `if graph.scope_dict()[first_node] is not None: return False` to the `can_be_applied()`.
Before you modify the edges you have to run the following code:
```python
scope_dict = graph.scope_dict()
for node in graph.nodes():
    if isinstance(node, (dace_nodes.EntryNode)) and scope_dict[node] is None:
        dace.sdfg.utils.canonicalize_memlet_trees_for_scope(graph, node)
        dace.sdfg.utils.canonicalize_memlet_trees_for_scope(graph, graph.exit_node(node))
```
This function _should ^{(R)}_ ensure that it is okay do perform the manipulation later directly.

Secondly, this here is probably the only case were it is wrong to use `{src, dst}_subset`, instead you must use `subset` and `other_subset`.
The issue is that `{src, dst}_subset` handles the subsets in a way that are meaningful when you care about the direction of an edge, but in this case you are not caring about it.
At least not explicitly, for that you would have to check `edge.{src, dst}`, which is a problem if a scoping node, i.e. a `Map`, is in between.

The canonicalize call above has ensured that edges within a Map scope, that belongs to a Memlet tree (the thing that gets highlighted when you hoover over an edge in the visualization) from the outside, referring to the outside node.
I.e. for every edge on the Memlet tree the following holds `graph.memlet_tree(edge)[0].src.data == edge.data.data`.
If my explanation sounds cryptic to you, this is because it is and hard to explain, so best to ask me in person.
```python
for edge in graph.edges():
    edge_data = edge.data.data
    if edge_data == second_node.data:
        edge.data.subset.offset(node_offset, negative=True)
```
The bottom line is, that thanks to the canonicalization you know that you have to adapt every edge that refers to `second_node.data` you must update and modify `subset`.
There is however a slight problem as this does not necessarily handle connections between `second_node` and another AccessNode.
Worse it might even modify and change the connections between `first_node` and `second_node`, which you must handle/exclude to.

I think you will have to do something like this:
```python
for edge in graph.edges:
    if edge.src is second_node or edge.dst is second_node:
        Handle these cases
    elif edge.data.data == second_node.data:
         Modification of `subset`.
```

philip-paul-mueller (commented):
It generally looks good, but there are some aspects that needs more work.
While the transformation is hyper specific it is also hyper specific of how the graph elements, especially the `data` attribute are created.
Thus any change in DaCe or GT4Py might make the thing fail.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
iomaganaris: If we check for the name here I think we can also enable this transformation by default in auto optimize. (That was my original hacky thought to conditionally enable transformations but I thought the hooks are better 😅 )

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
iomaganaris: `next_w` already exists in this state twice, so it shouldn't be part of `single_use_data`, right?
I can use `data_nodes()` here though indeed 👍 
Also I would say it's not necessary to check in all states because we shouldn't care how `next_w` is handled in other states because there shouldn't be any dependency that matters for replacing the transformation

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
iomaganaris: Valid point

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
iomaganaris: I missed the existence of `subset_merger` 😅 I can also use that

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
iomaganaris: I thought they were going to be used later and I kept them but it's true they are not necessary 👍

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
edopao: FYI https://github.com/spcl/dace/pull/2189 is now merged and available in `gt4py:main`

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
iomaganaris: I have temporarily enabled the transformation in `gt_auto_optimize` to ease banchmarking and check inside `can_be_applied` the name of the sdfg

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -584,6 +584,12 @@ def _gt_auto_process_top_level_maps(
             validate_all=validate_all,
         )
 
+        sdfg.apply_transformations_repeated(
+            gtx_transformations.RemoveAccessNodeCopies(single_use_data=single_use_data),
+            validate=False,
+            validate_all=validate_all,
+        )
+
```
edopao: I would merge this PR without running the new transformation. It will be possible to enable it as a hook from icon4py.
```suggestion

```

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
edopao: I agree with Philip. We need to be able to setup this transformation with name patterns specific to a certain gt4py program. We want then to run this transformation only on the selected program, by configuring the proper hook in icon4py.

iomaganaris: I totally agree with you and I will take care of that. I just wanted to make these changes usable in the short term

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
iomaganaris: Thank you very much once more for the detailed explanation @philip-paul-mueller . I hope I took care properly the edge subsets now

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,249 @@
...
+    )
+
+    # Name of all data that is used at only one place. Is computed by the
+    #  `FindSingleUseData` pass and be passed at construction time. Needed until
+    #  [issue#1911](https://github.com/spcl/dace/issues/1911) has been solved.
+    _single_use_data: Optional[dict[dace.SDFG, set[str]]]
+
+    def __init__(
+        self,
+        stencil_selection: Optional[Sequence[str]] = None,
```
havogt: This shouldn't be needed with the icon4py mechanism to add this transformation to only a particular program, right?

iomaganaris: It's not necessary indeed. It just adds a bit more flexibility in the selection of the stencils. If you think it's too much it can be removed

edopao: I would remove it, because it overlaps with the mechanism in place for program hooks and this may create confusion. Besides, it is the first time I see the concept of stencil (although just in the variable name, here) in our SDFG transformations.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,249 @@
...
+    This transformation helps to reduce the number of data copies in the SDFG."""
+
+    first_node = dace_transformation.PatternNode(dace_nodes.AccessNode)
+    second_node = dace_transformation.PatternNode(dace_nodes.AccessNode)
+    third_node = dace_transformation.PatternNode(dace_nodes.AccessNode)
+    fourth_node = dace_transformation.PatternNode(dace_nodes.AccessNode)
+
+    # If given, only apply the transformation to the selected stencils (by SDFG name)
+    stencil_selection = dace_properties.Property(
+        dtype=Sequence,
```
philip-paul-mueller: I am actually not sure if that work, because as far as I know [they do an `isinstance()` check](https://github.com/spcl/dace/blob/29bcb5d3bda48e3397498360cf2e71f1ce9f7db3/dace/properties.py#L144) and you must pass a concrete class.
But as it seems this will get removed it is irrelevant, so FYI.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,249 @@
...
+            return False
+        if scope_dict[second_node] is not None:
+            return False
+        if scope_dict[third_node] is not None:
+            return False
+        if scope_dict[fourth_node] is not None:
+            return False
+
+        # Make sure that first and fourth node refer to the same data
+        if first_node.data != fourth_node.data:
```
philip-paul-mueller, line 111: I would also check `first_node.data == "next_w"`.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,249 @@
...
+        if second_desc.shape != third_desc.shape:
+            return False
+
+        # Make sure that second and fourth node have the same shape as the first node
+        if second_desc.shape != first_desc.shape or third_desc.shape != first_desc.shape:
+            warnings.warn(
+                f"[RemoveAccessNodeCopies] Shapes of second or third node do not match the shape of the first node: {second_desc.shape}, {third_desc.shape} vs {first_desc.shape}",
+                stacklevel=0,
+            )
+            # needs https://github.com/GridTools/gt4py/pull/2173 # return False
```
philip-paul-mueller: I do not fully understand this.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
philip-paul-mueller: I would make sure that `second_node` and `thrid_node` refer to single use data.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,249 @@
...
+                for fourth_node_range in ranges_written_to_fourth_node
+            ):
+                return False
+
+        # Make sure that the whole data of the first node is written
+        first_node_range = first_desc.shape
+
+        union_written_to_first_node_data = next(
+            iter(gtx_dace_split.subset_merger(ranges_written_to_first_node))
+        )
```
philip-paul-mueller: There is no guarantee that `subset_merger()` will return a single set.
So you would have to check that first.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # Find the edge between first and second access nodes (there should be only one)
+        edge_between_first_and_second = next(
+            edge for edge in graph.edges_between(first_node, second_node)
+        )
+
+        # Compute the offset of the data between first and second node
+        assert len(edge_between_first_and_second.data.src_subset) == len(
+            edge_between_first_and_second.data.dst_subset
+        )
```
philip-paul-mueller:
```suggestion
        assert edge_between_first_and_second.data.src_subset.dims() == edge_between_first_and_second.data.dst_subset.dims()        
```
I agree it is a bit longer, but it is also a bit more direct.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,249 @@
...
+        node_offset = [
+            range_first_min - range_second_min
+            for range_first_min, range_second_min in zip(
+                edge_between_first_and_second.data.src_subset.min_element(),
+                edge_between_first_and_second.data.dst_subset.min_element(),
+            )
+        ]
```
philip-paul-mueller, line 231: Is there a check that ensures that `second_node` and `third_node` have the same shape.
If not then shouldn't there be different offsets?

philip-paul-mueller (commented):
Some small comments.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
iomaganaris: I would prefer not to hardcode values and name of variables in the transformation. It can be that the name of the variable changes in the `icon4py` frontend but the behavior stays the same so the transformation would still be valid

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
iomaganaris: I wanted to check here that the shape of the `data` of the 4 AccessNodes that the transformation is based on are the same. However to do that we need to have the compile time domains for the global AccessNode which is enabled in the mentioned PR.
Now I have realized though that the shapes of the third and fourth node are different than the first and fourth because in the horizontal domain they only have to save a slice of `next_w`. I will remove this check

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
iomaganaris: Yes, this is checked in `can_be_applied`:
```
        # Make sure that second and third node have the same shape
        if second_desc.shape != third_desc.shape:
            return False
```

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,237 @@
...
+    def __init__(
+        self,
+        *args: Any,
+        single_use_data: Optional[dict[dace.SDFG, set[str]]] = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self._single_use_data = None
+        if single_use_data is not None:
+            self._single_use_data = single_use_data
```
philip-paul-mueller: Edoardo made me aware of this possibility, because it is not a DaCe property.
```suggestion
        self._single_use_data = single_use_data
```

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,237 @@
...
+                gtx_dace_split.are_intersecting(edge.data.dst_subset, fourth_node_range)
+                for fourth_node_range in ranges_written_to_fourth_node
+            ):
+                return False
+
+        # Make sure that the whole data of the first node is written
+        first_node_range = first_desc.shape
+
+        first_node_writes_subset = gtx_dace_split.subset_merger(ranges_written_to_first_node)
+        assert len(first_node_writes_subset) == 1
```
philip-paul-mueller: Instead of `assert` I would probably return False and maybe also generate a warning.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,237 @@
...
+        for edge in graph.edges():
+            # Update edges that refer to second or third AccessNodes
+            if edge.data.data == second_node.data or edge.data.data == third_node.data:
+                # Replace data with data of first node
+                edge.data.data = first_node.data
+                # Add offset to subset
+                edge.data.subset.offset(node_offset, negative=False)
+                # Add offset to other_subset if it's necessary (src is either first, second or third node and dst is either second or third node)
+                if edge.dst in [second_node, third_node] and edge.data.other_subset is not None:
+                    edge.data.other_subset.offset(node_offset, negative=False)
```
philip-paul-mueller, line 251: I am a bit confused here.
As far as I can see you are handling here edges that writes into the second or third node, that are not related to the chain.
But I am wondering how you are handling/excluding the edge between the second and or third edge.
Or does that thing need to be handled in the same way? Maybe?

iomaganaris: Maybe I have confused things as well..
The logic here is the following.
Let's say that the ranges of the nodes are:
```
first_node = fourth_node = [4000:41000, 0:80]
second_node = third_node = [0:37000, 0:80]
```
This means that if we wanted to write the whole domain of the second node with a map. Then we would have an edge connecting the map with the second_node with:
```
edge.data.data = second_node.data
edge.data.subset = [0:37000, 0:80]
```
Since we want to replace the second_node.data with first_node.data that have different range we have to add the necessary offset to the edge.data.subset. (Logic of `lines 235:240`)
Now in case we write from first node to second node we would have an edge with:
```
edge.data.data = second_node.data
edge.data.subset = [4000:41000, 80]
edge.data.other_subset = [0:37000, 80]
```
In that case we have to change the other_subset which corresponds to the second_node range and add to it the same offset. (logic of `lines 228:229`).
In the case of `vertically_implicit_solvers` there are no other edges going in or out of these access nodes. However I think that based on the `edge.data` and the destinations of the `edge.data` (in case of incoming edges to second and third_node) the handling of the edges should be correct.
See also:
`Case 1`
image
`Case 2
`image

philip-paul-mueller: No it make sense.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -0,0 +1,251 @@
+# GT4Py - GridTools Framework
+#
+# Copyright (c) 2014-2024, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
```
havogt, line 8:
```suggestion

# TODO(iomaganaris): This transformation is very specific to a pattern in the vertically_implicit_solver of ICON4Py.
# The transformation should be generalized or moved to ICON4Py.

```

PR #2331, opened 2025-10-22
Fix thread block size setting assertion

iomaganaris:
Not sure how the CI (and I) missed that


edopao:
> > Not sure how the CI (and I) missed that
> 
> I think I see the assert error in icon4py CI: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/4455690602105886/4525297225819146/-/jobs/11806902756

Without this fix, of course.

PR #2332, opened 2025-10-22
feat[next]: Implement faster data pointer retrieval in ndarray-based fields

egparedes:
This PR changes the mechanism to access the data pointer in `NdArrayField`s subclasses. Instead of using the `data_ptr()` function call (required for compatibility with DaCe orchestration) we now use the `_data_buffer_ptr_` internal attribute, which allows us to use `property` and `cached_property` when possible to speed up the access.


havogt (approved):
lgtm

PR #2334, opened 2025-10-23
feat[dace][next]: Strides of Views

philip-paul-mueller:
In case there is a view of a global array, we do not have a problem, because we never modify the strides of globals.
The problems start when there is a view of a transient.
The SDFG is constructed using C order, which good for CPU.
However, for GPU we go to FORTRAN order which is essentially swapping the memory order.
But we can not only do that for transients, but must also do it for views, which might be difficult, because views can add or remove certain dimensions.

This PR does _not_ solve it, but instead generates an error if such a case is detected.


philip-paul-mueller:
I think we can conclude that the performance is okay.

philip-paul-mueller:
Verfied correctness in ICON4Py CI: https://github.com/C2SM/icon4py/pull/923

iomaganaris (approved):
The changes look reasonable however I don't know the context of the changes

PR #2342, opened 2025-10-24
perf[next]: cache low level buffer information in fields

egparedes:
Add a new `BufferInfo`/`__gt_buffer_info__` interface to collect the required low-level information to pass Field buffers to backends and cache it on the `Field` side, when the `Field` implementation allows it. The DaCe binding generator already uses this new interface.

Additionally, this PR minor refactorings in the `Field` implementation, and minor code style changes to make the implementation follow the same order of method definitions as the interface.

### DIFF
--- a/src/gt4py/next/common.py
+++ b/src/gt4py/next/common.py
@@ -14,6 +14,7 @@ import dataclasses
 import enum
 import functools
 import math
+import sys
 import types
 from collections.abc import Iterable, Mapping, Sequence
 
@@ -798,7 +799,11 @@ class BufferInfo:
     def from_ndarray(cls, ndarray: core_defs.NDArrayObject) -> BufferInfo:
         # TODO(egparedes): Implement this function using __dlpack__ and ctypes.
         #   The current implementation is messy and only works for numpy and cupy.
-        array_ns = ndarray.__array_namespace__()  # type: ignore[attr-defined]
+        try:
+            array_ns = ndarray.__array_namespace__()  # type: ignore[attr-defined]
+        except AttributeError:
+            array_ns = sys.modules[ndarray.__class__.__module__]
+
         array_byte_bounds_func = (
             getattr(array_ns, "byte_bounds", None) or array_ns.lib.array_utils.byte_bounds
         )
--- a/tests/next_tests/unit_tests/test_common.py
+++ b/tests/next_tests/unit_tests/test_common.py
@@ -649,7 +649,7 @@ class TestBufferInfo:
         buffer_info = common.BufferInfo.from_ndarray(a)
         assert hash(buffer_info) == buffer_info.hash_key
 
-        # Create a view wit the same data pointer and shape
+        # Create a view with the same data pointer and shape
         b = a[0:None]
         assert np.allclose(a, b)
         assert a is not b


src/gt4py/next/common.py
```diff
@@ -784,6 +783,42 @@ class MutableField(Field[DimsT, core_defs.ScalarT], Protocol[DimsT, core_defs.Sc
...
 
+class BufferInfo(NamedTuple):
+    """Holds information about a buffer in memory."""
+
+    data_ptr: int
+    ndim: int
+    shape: tuple[int, ...]
+    elem_strides: tuple[int, ...]
+    byte_strides: tuple[int, ...]
+    buffer_info_id: int
```
egparedes: Should it also contain device information?

src/gt4py/next/common.py
```diff
@@ -784,6 +783,42 @@ class MutableField(Field[DimsT, core_defs.ScalarT], Protocol[DimsT, core_defs.Sc
...
+
```
havogt, line 797: Should we add a `hash` function with cached hash or do you think we don't gain much over hashing the tuple?

egparedes: I've refactored the hash-related parts. It should be fine now.

src/gt4py/next/common.py
egparedes: Done

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -202,6 +146,49 @@ def from_array(
...
         return cls(domain, array)
 
+    @functools.cached_property
+    def __gt_origin__(self) -> tuple[int, ...]:
+        assert common.Domain.is_finite(self.domain)
+        return tuple(-r.start for r in self.domain.ranges)
+
+    @property
+    def __gt_buffer_info__(self) -> common.BufferInfo:
+        return common.BufferInfo.from_ndarray(self.ndarray)
```
havogt, line 166:
```suggestion
        """
        Fallback implementation to retrieve the 'BufferInfo' of a Field.
        
        See cached implementations on the NDArrayField subclasses.
        """
        return common.BufferInfo.from_ndarray(self.ndarray)
```
Something like this?

Or maybe put an a `NotImplemented` here to force the subclasses to implement it?

src/gt4py/next/common.py
```diff
@@ -784,6 +783,69 @@ class MutableField(Field[DimsT, core_defs.ScalarT], Protocol[DimsT, core_defs.Sc
...
+            data_ptr=data_ptr,
+            ndim=ndim,
+            shape=shape,
+            elem_strides=elem_strides,
+            byte_strides=byte_strides,
+            device=device,
+        )
+
+    @functools.cached_property
+    def hash_key(self) -> int:
```
havogt, line 838:
```suggestion
    def _hash_key(self) -> int:
```
maybe?

src/gt4py/next/embedded/nd_array_field.py
egparedes: I merged the suggestion. I don't think we need to make it an abstract method for now, but maybe in the future if we really add more subclasses.

src/gt4py/next/common.py
egparedes: I'd rather leave it as a public attribute since this is useful mostly in performance critical code and accessing it directly saves the unnecessary function call to `hash()`

edopao: This `hash_key` is meant to be used in the future, right? I think @egparedes mentioned some ideas.

edopao (commented):
LGTM

PR #2343, opened 2025-10-24
cartesian[fix]: Remove hard check on ADR existing for experimental feature

FlorianDeconinck:
## Description

Part of our "experimental" methodology is to write ADRs for each. To reinforce this, we check in the experimental feature warning that the ADR exists.

Alas' this doesn't really work on deployment:
- `docs` not deployed on install
- the way the path behaves are different depending on deployed or editable or local code

We remove the check  - we still reference the ADR on the `github`


romanc (approved):
You will also need to remove the test that checks the ValueError. Otherwise, this looks good.

And yes since docs arent deployed we cant (and arguably in user code we shouldnt) check for ADR existence.

PR #2344, opened 2025-10-24
Improve setting of thread block size for 1D maps

iomaganaris:
The block size is set according to the following rules:
- Determine the number of map parameters that have range > 1.
- If there is only one such parameter then handle the map as 1D map, else if there are more than 2 parameters with range > 1 then handle it as ordinary map.
- For 1D maps set the block size such that only the dimension with the largest range is set to the specified block size, all other dimensions are set to 1.

### DIFF
--- a/src/gt4py/cartesian/utils/base.py
+++ b/src/gt4py/cartesian/utils/base.py
@@ -17,7 +17,6 @@ import importlib.util
 import itertools
 import json
 import os
-import pathlib
 import string
 import sys
 import time
@@ -458,11 +457,6 @@ def warn_experimental_feature(*, feature: str, ADR: str) -> None:
 
     # be nice and remove a potential `/` prefixed
     ADR = ADR.removeprefix("/")
-
-    repo_root = pathlib.Path(__file__).parent.parent.parent.parent.parent
-    adr_location = repo_root / "docs" / "development" / "ADRs" / "cartesian" / ADR
-    if not adr_location.exists():
-        raise ValueError(f"Could not find ADR for '{feature}' in. Looked in `{adr_location}`.")
 
     warnings.warn(
         f"{feature} is an experimental feature. Please read "
--- a/tests/cartesian_tests/unit_tests/test_util.py
+++ b/tests/cartesian_tests/unit_tests/test_util.py
@@ -37,8 +37,3 @@ def test_warn_experimental_feature() -> None:
     # even works with pre-fixed slash in (relative) ADR path
     with pytest.warns(UserWarning, match="Test feature is an experimental feature."):
         gt_util.warn_experimental_feature(feature="Test feature", ADR="/README.md")
-
-
-def test_warn_experimental_feature_raise_if_adr_not_found() -> None:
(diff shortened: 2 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -604,6 +605,11 @@ def __init__(
         super().__init__()
         if block_size_1d is not None:
             self.block_size_1d = block_size_1d
+            if self.block_size_1d[1] != 1 or self.block_size_1d[2] != 1:
+                warnings.warn(
+                    f"1D map block size specified with mmore than one dimension larger than 1. Configured 1D block size: {self.block_size_1d}.",
```
philip-paul-mueller:
```suggestion
                    f"1D map block size specified with more than one dimension larger than 1. Configured 1D block size: {self.block_size_1d}.",
```

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -604,6 +605,11 @@ def __init__(
...
+                    stacklevel=0,
+                )
         if block_size_2d is not None:
```
philip-paul-mueller, line 621: The warning from above is missing here.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -651,27 +656,61 @@ def apply(
         graph: Union[dace.SDFGState, dace.SDFG],
         sdfg: dace.SDFG,
     ) -> None:
-        """Modify the map as requested."""
```
philip-paul-mueller, line 654: I think you should put this description into doc string of the class.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -651,27 +656,61 @@ def apply(
...
-                assert axis_size <= 1, (
-                    "GPU thread block size setting currently does not support maps where non-leading "
-                    "dimensions have size greater than one if the next dimension has size one."
-                )
+        for axis_size in map_size:
             if axis_size != 1:
-                num_map_params += 1  # Handle 2D maps where one dimension has range 1 as 1D map
+                num_map_params += 1  # Count how many parameters have range > 1
+        # If the number of map parameters that have range > 1 is 1 then handle the map as 1D map else handle it as full dimensional map
+        if num_map_params > 1 and dims_to_inspect > num_map_params:
```
philip-paul-mueller, line 682: For your information, the pythonic way of doing it would be:
```suggestion
        if dims_to_inspect > num_map_params > 1:
```
However, I am not a fan of this chaining either, so you can keep it the way it is.

iomaganaris: I agree that the 2 separate checks make this more understandable

PR #2345, opened 2025-10-24
Add NVTX marker instrumentation

iomaganaris:
Not sure if this is useful but I'm opening a PR to avoid losing these changes.

It generates the following NVTX ranges:
image

### Requirements

- https://github.com/spcl/dace/pull/2001


src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -267,6 +267,14 @@ def add_instrumentation(sdfg: dace.SDFG, gpu: bool) -> None:
         dace.Memlet(f"{output}[0]"),
     )
 
+    if (config.COLLECT_METRICS_LEVEL == metrics.GPU_TX_MARKERS) and gpu:
+        sdfg.instrument = dace.dtypes.InstrumentationType.GPU_TX_MARKERS
+        for node, _ in sdfg.all_nodes_recursive():
+            if isinstance(node, dace.nodes.MapEntry):
+                node.instrument = dace.dtypes.InstrumentationType.GPU_TX_MARKERS
```
edopao: Adding just a note here. Initially I thought that the instrumentation of GPU maps is not strictly needed, because they already appear in the NVTX profiling as GPU kernels. However, on a second thought I realized that this additional trace allows to see all maps, also the CPU maps, and even maps with sequential schedule (if any), so it is good to keep the map instrumentation. I suggest to add a comment.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -267,6 +267,14 @@ def add_instrumentation(sdfg: dace.SDFG, gpu: bool) -> None:
         dace.Memlet(f"{output}[0]"),
     )
 
+    if (config.COLLECT_METRICS_LEVEL == metrics.GPU_TX_MARKERS) and gpu:
```
edopao, line 270: Your current check is at compile time. The check in the instrumentation above is at runtime. I propose to always add NVTX instrumentation, because the `config.COLLECT_METRICS_LEVEL` value is meant to be a runtime setting. It is still possible to skip SDFG instrumentation by setting `use_metrics=False` on the dace backend, at compile-time.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
iomaganaris: Good point. I'll add a comment. Btw in the instrumentation implementation in DaCe the maps that have a GPUScheduled are skipped from the `GPU_TX_MARKERS` instrumentation

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
iomaganaris: I am not sure I would be up to enabling NVTX ranges by default since they will add some (really minimal) overhead to the CPU side. If `use_metrics=False` can be set by default then it would be fine but it would increase some more complexity for production and profiling runs

egparedes (approved):

PR #2346, opened 2025-10-27
refactor[next][dace]: Refactoring tests of GTIR to SDFG lowering

edopao:
This PR is refactoring the GTIR-to-SDFG test code:
- Most tests were still setting the domain in `as_fieldop`, although the domain is inferred by gt4py in the call to `build_sdfg_from_gtir()`. This PR removes unnecessary domains from the IR, and instead passes `skip_domain_inference=True` for those IRs where the domains need to be manually set.
- Replace custom-made `get_domain_range()` with gt4py helper function `get_field_domain()`.
- Inline `stmt` domain in IR, without using a separate variable, whenever possible.
- Remove one test case `test_gtir_neighbors_as_output` that was never supported in gt4py, and there is no plan to support.

Additionally, this PR enables the tuple iterator tests, for dace backend because they work without any code change.


tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_gtir_to_sdfg.py
```diff
@@ -128,27 +128,19 @@ def build_dace_sdfg(
     return dace_backend.build_sdfg_from_gtir(ir, offset_provider_type, column_axis=KDim)
 
 
-def get_domain_range(
-    field: str | gtir.Expr, dim: gtx_common.Dimension, margin: Optional[tuple[int, int]] = None
-) -> tuple[gtir.Expr, gtir.Expr]:
-    r = (
-        im.tuple_get(0, im.call("get_domain_range")(field, im.axis_literal(dim))),
-        im.tuple_get(1, im.call("get_domain_range")(field, im.axis_literal(dim))),
+def apply_margin_on_field_domain(
```
philip-paul-mueller, line 136: I would add a doc string here.

philip-paul-mueller (approved):
LGTM

PR #2349, opened 2025-10-28
feat[dace][next]: Modified How Degenerated 1D Maps Are Handled

philip-paul-mueller:
The PR modifies how degenerated 1d Maps, i.e. Maps with more then 1 dimension, but only one of them is != 1, are handled.
The previous version also handled changes where the Map had more than 4 dimensions, which leaded to a complicated logic that required to sort the sizes.
However, in case of not fully specialized Maps, the Maps still might have symbolic sizes, which made it impossible to sort them, thus this feature was removed.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
@@ -701,13 +701,6 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
                     stacklevel=0,
                 )
 
-            elif non_trivial_1d_map_degenerated_map_dimension is None:
-                warnings.warn(
-                    f"Map '{gpu_map}' has {dims_to_inspect} dimensions but not all are non trivial (range > 1). "
-                    "Handling it as a map with full dimensions.",
-                    stacklevel=0,
-                )
-
         # Because of a particularity of the DaCe code generator, the iteration
         #  variable that is associated to the `x` dimension of the block is the
         #  last parameter, i.e. `gpu_map.params[-1]`. The one for `y` the second last.
@@ -722,6 +715,12 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
                 block_size[len(map_size) - non_trivial_1d_map_degenerated_map_dimension - 1] = (
                     self.block_size_1d[0]
                 )
+                if non_trivial_1d_map_degenerated_map_dimension != 0:
+                    warnings.warn(
+                        f"Blocksize of 1d Map '{gpu_map}' was set to {block_size}, but the iteration index is not the x dimension.",
+                        stacklevel=0,
+                    )
+
             else:
                 block_size = list(self.block_size_1d)
 


src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -528,13 +528,11 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
...
-        - If there is only one such parameter then handle the map as 1D map,
-          else if there are more than 2 parameters with range > 1 then handle it as ordinary map.
-        - For 1D maps set the block size such that only the dimension with the largest
-          range is set to the specified block size, all other dimensions are set to 1.
-        - If launch_bounds_1d is not set then set the launch_bounds of `scan` maps to 512 to limit their register usage.
+    Note there are some special rules:
+    - If a Map with (at most 3 dimension) has only one non trivial dimension, i.e.
+        > 1, then the Map is handled as a 1D Map.
+    - If there is a 1D Map that contains a nested SDFG whose label starts with `scan_`
+        and `launch_bounds_1d` was not specified, then a value of 512 is assumed.
```
iomaganaris:
```suggestion
        and `launch_bounds_1d` was not specified, then a value of 512 is assumed to
        limit its register usage.
```

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -673,63 +671,73 @@ def apply(
...
+                    if non_trivial_1d_map_degenerated_map_dimension is not None:
+                        # Found more than two non trivial dimension.
+                        non_trivial_1d_map_degenerated_map_dimension = None
+                        break
+                    non_trivial_1d_map_degenerated_map_dimension = dim
+            else:
+                # We found at most one non trivial Map dimensions. Find out if it is
+                #  exactly one such Map.
+                if non_trivial_1d_map_degenerated_map_dimension is not None:
+                    is_degenerated_1d_map = True
```
iomaganaris:
```suggestion
                else:
                    # We found at most one non trivial Map dimensions. Find out if it is
                    #  exactly one such Map.
                    if non_trivial_1d_map_degenerated_map_dimension is not None:
                        is_degenerated_1d_map = True
```
IIUC

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_gpu_utils.py
```diff
@@ -230,14 +230,18 @@ def test_set_gpu_properties(method: int):
     assert map4.gpu_launch_bounds == "200"
 
 
-def test_set_gpu_properties_1D():
+@pytest.mark.parametrize("reverse_4d_map_order", [True, False])
```
iomaganaris: Instead of running all the cases again by parametrizing the test I would prefer just adding an extra map with the required logic

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -673,63 +671,73 @@ def apply(
...
+        non_trivial_1d_map_degenerated_map_dimension: Optional[int] = None
+        if 1 < dims_to_inspect <= 3:
+            for dim, dim_size in enumerate(map_size):
+                if dim_size != 1:
+                    if non_trivial_1d_map_degenerated_map_dimension is not None:
+                        # Found more than two non trivial dimension.
+                        non_trivial_1d_map_degenerated_map_dimension = None
+                        break
+                    non_trivial_1d_map_degenerated_map_dimension = dim
+            else:
```
iomaganaris:
```suggestion
            else:  # we haven't found more than 1 dimension > 1
```

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
philip-paul-mueller: I agree it is strange.
I have updated the code and it actually looks nicer now.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_gpu_utils.py
philip-paul-mueller: I agree, but then we should have a different dimension in that case.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -673,63 +672,72 @@ def apply(
...
+                    f"Map '{gpu_map}', size '{map_size}', is a degenerated 1d Map. Handle it as a 1d Map.",
+                    stacklevel=0,
+                )
+
+            elif non_trivial_1d_map_degenerated_map_dimension is None:
+                warnings.warn(
+                    f"Map '{gpu_map}' has {dims_to_inspect} dimensions but not all are non trivial (range > 1). "
+                    "Handling it as a map with full dimensions.",
+                    stacklevel=0,
+                )
```
iomaganaris: Based on the logic above I think that this is going to be printed for every map than has more than 1 range larger than 1 which is going to be most of the time all of the maps. I think it doesn't add much so it can be removed

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -673,63 +672,72 @@ def apply(
...
+            if is_degenerated_1d_map:
+                assert non_trivial_1d_map_degenerated_map_dimension is not None
+                assert 0 <= non_trivial_1d_map_degenerated_map_dimension <= 2
+                assert len(map_size) <= 3
+
+                block_size = [1, 1, 1]
+                # Order of Map parameters is from outer to inner, i.e. z,y,x
+                block_size[len(map_size) - non_trivial_1d_map_degenerated_map_dimension - 1] = (
+                    self.block_size_1d[0]
                 )
```
iomaganaris, line 718: This warning might be helpful in case we see some unexpected behavior in the future so I would keep it

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
philip-paul-mueller: You are right, I removed it.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
philip-paul-mueller: Restored it.

iomaganaris (approved):

PR #2351, opened 2025-10-29
feat[dace][next]: Using the `canonicalize_memlet_trees()` function

philip-paul-mueller:
Especially older DaCe transformations have trouble handling non standard Memlets. i.e. they expect a particular assignment of `Memlet.data` and `Memlet.subset`, among them is Memlet propagation.
The recently introduced helper function [`canonicalize_memlet_trees()`](https://github.com/spcl/dace/pull/2189) allows to transform the SDFG in such a way that it conforms to this format.
This PR introduces calls to this function at particular locations, mostly before Memlet propagation happens.

TODO:
- [x] Verify correctness in ICON4Py CI or verification mode
- [x] Verify performance stability.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/strides.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/strides.py
@@ -703,6 +703,7 @@ def _gt_modify_strides_of_views_non_recursive(sdfg: dace.SDFG) -> None:
         evaluate if this function is still needed.
     """
     for state in sdfg.states():
+        scope_dict = state.scope_dict()
         propagation_record: set[PropagatedStrideRecord] = set()
         for view_node in state.data_nodes():
             view_desc = view_node.desc(sdfg)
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -518,8 +518,9 @@ def _gt_auto_process_top_level_maps(
         )
 
         if not disable_splitting:
-            # TODO(phimuell): Find out how to skip the propagation and integrating it
-            #   into the split transformation.
+            # TODO(phimuell): Implement a data cleaner.
+            dace_sdutils.canonicalize_memlet_trees(sdfg)
+            dace_propagation.propagate_memlets_sdfg(sdfg)
```
edopao, line 525: Would it be useful to have 2 gt4py wrapper functions?
- `gt_propagate_memlet_sdfg()` that also calls `canonicalize_memlet_trees(sdfg)`
- `gt_propagate_memlet_map()` that also calls `canonicalize_memlet_trees_for_scope(map_entry / map_exit)`

philip-paul-mueller: I am not fully sure, as we only call `canonicalize_memlet_trees()` (which seems to have a bug) to fix some other things and in the long run the only reason where it is called is in front of dace simplify or other (known) DaCe transformation.
So I am not sure if it makes sense in the long run.

However I agree that there should be a function that runs the thing for entry and exit but this should life in dace.
I will do a PR tomorrow.

edopao (approved):
LGTM

PR #2353, opened 2025-10-30
feat[dace][next] More Roboust Calling

philip-paul-mueller:
This PR changes how calls to the underlying `CompiledSDFG` objects are carried out.
Before, the implementation heavily relied on the internal data format of the DaCe class.
However a recent [change in DaCe](https://github.com/spcl/dace/pull/2185) changed this internal data leading to errors, that had to [be patched](https://github.com/GridTools/dace/pull/9).
This PR introduces a more stable fix and builds upon a [refactoring in DaCe](https://github.com/spcl/dace/pull/2206) that beside other things, exposes the tools that were needed by GT4Py to work independently of the internals.
For that reason this PR also updates the DaCe dependency to `2025.11.04`.

The main change is, that the argument vector, i.e. the C representation of the arguments used for the call, are no longer managed by `CompiledSDFG` but instead by GT4Py's `CompiledDaceProgram`.

TODO:
- [x] ICON4Py CI must pass: https://github.com/C2SM/icon4py/pull/925
- [x] [DaCe PR#2206](https://github.com/spcl/dace/pull/2206) must be included.


src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -44,16 +43,8 @@ def decorated_program(
...
-            # Initialization of `_lastargs` was done by the `CompiledSDFG` object,
-            #  so we just update it with the current call arguments.
-            update_sdfg_call_args(args, fun.sdfg_program._lastargs[0])
-            fun.fast_call()
-        except IndexError:
-            # First call, the SDFG is not initialized, so forward the call to `CompiledSDFG`
-            # to properly initialize it. Later calls to this SDFG will be handled through
-            # the `fast_call()` API.
-            assert len(fun.sdfg_program._lastargs) == 0  # `fun.sdfg_program._lastargs` is empty
+        if fun.csdfg_args is None:
```
edopao: It was Enrique's idea to use try/except, because in Python there is no overhead if no exception is raised. Our regular case will be in the try, the except is only on first call. The if has a super-little overhead, but in every program call.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -56,12 +61,37 @@ def __init__(
         # For debug purpose, we set a unique module name on the compiled function.
         self.update_sdfg_ctype_arglist.__module__ = os.path.basename(program.sdfg.build_folder)
 
-    def __call__(self, **kwargs: Any) -> None:
-        result = self.sdfg_program(**kwargs)
-        assert result is None
+        # If not called they yet, then it is `None`.
```
edopao:
```suggestion
        # If not called yet, then it is `None`.
```

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -65,8 +56,16 @@ def decorated_program(
                 gtx_wfdcommon.SDFG_ARG_METRIC_LEVEL: config.COLLECT_METRICS_LEVEL,
                 gtx_wfdcommon.SDFG_ARG_METRIC_COMPUTE_TIME: collect_time_arg,
             }
-            with dace.config.set_temporary("compiler", "allow_view_arguments", value=True):
-                fun(**this_call_args)
+            fun.process_arguments(**this_call_args)
```
edopao: I like this `process_arguments()` function.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
philip-paul-mueller: We can copy that idea, although I do not like it that much.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
philip-paul-mueller: Thanks, I think it makes the `convert_args()` function nicer.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -45,15 +44,14 @@ def decorated_program(
...
+            # NOTE: If this is the first call then we will generate an exception because
+            #   `fun.csdfg_args` is still the empty tuple.
+            update_sdfg_call_args(args, fun.csdfg_args[0])  # type: ignore[misc]  # Will error out in first call.
+
         except IndexError:
-            # First call, the SDFG is not initialized, so forward the call to `CompiledSDFG`
-            # to properly initialize it. Later calls to this SDFG will be handled through
-            # the `fast_call()` API.
-            assert len(fun.sdfg_program._lastargs) == 0  # `fun.sdfg_program._lastargs` is empty
+            # First call. Construct the initial argument vector of the `ComiledDaceProgram`.
```
edopao:
```suggestion
            # First call. Construct the initial argument vector of the `CompiledDaceProgram`.
```

pyproject.toml
```diff
@@ -11,7 +11,7 @@ dace-cartesian = [
   'dace>=1.0.2,<2'  # renfined in [tool.uv.sources]
 ]
 dace-next = [
-  'dace==2025.10.30'  # refined in [tool.uv.sources]
+  'dace==2025.05.35'  # refined in [tool.uv.sources] and I like that book.
```
edopao: What book? :)

edopao (commented):
Very nice!

pyproject.toml
philip-paul-mueller: It is a [German children book](https://en.wikipedia.org/wiki/The_35th_of_May;_or,_Conrad%27s_Ride_to_the_South_Seas) and serves as a reminder for me that I have to undo these changes. 
You see I needed a version/date that will never be selected for this hack.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -36,6 +37,12 @@ class CompiledDaceProgram(stages.CompiledProgram):
         None,
     ]
 
+    # Processed argument vectors that are passed to `CompiledSDFG.fast_call()`. If the
+    #  `tuple` is empty then it is not initialized. The first element is used in the
+    #  normal call and we update it, the second element is used only for initialization
+    #  and we do not update it.
+    csdfg_args: Union[tuple[list[Any], Sequence[Any]], tuple[()]]
```
egparedes: Why not just using `None` as usual for an uninitialized attribute?
```suggestion
    csdfg_args: tuple[list[Any], Sequence[Any]] |  None
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -56,12 +63,33 @@ def __init__(
         # For debug purpose, we set a unique module name on the compiled function.
         self.update_sdfg_ctype_arglist.__module__ = os.path.basename(program.sdfg.build_folder)
 
-    def __call__(self, **kwargs: Any) -> None:
-        result = self.sdfg_program(**kwargs)
-        assert result is None
+        # If not called they yet, thus empty tuple.
```
egparedes:
```suggestion
        # Since the sdfg program hasn't been called yet. 
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -56,12 +63,33 @@ def __init__(
...
+        """
+        This function will process the arguments and store the processed values in `self.csdfg_args`,
+        to call them use `self.fast_call()`.
+        """
+        with dace.config.set_temporary("compiler", "allow_view_arguments", value=True):
+            processed_csdfg_args = self.sdfg_program.construct_arguments(**kwargs)
+        assert isinstance(processed_csdfg_args, tuple) and len(processed_csdfg_args) == 2
+        # Note we only care about the first argument vector, that is used in normal call.
+        #  Since we update it, we ensure that it is a `list`.
+        self.csdfg_args = (list(processed_csdfg_args[0]), processed_csdfg_args[1])
```
egparedes: Using explicit unpacking achieves the same effect as the assert and it makes the coder nicer to read.
I wouldn't store the second argument if it is not needed.
```suggestion
            # We only care about the first value, which stores the arguments used in a normal call.
            csdfg_call_args, _ = self.sdfg_program.construct_arguments(**kwargs)
        self.csdfg_args =[*csdfg_call_args]
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -56,12 +63,33 @@ def __init__(
         # For debug purpose, we set a unique module name on the compiled function.
         self.update_sdfg_ctype_arglist.__module__ = os.path.basename(program.sdfg.build_folder)
 
-    def __call__(self, **kwargs: Any) -> None:
-        result = self.sdfg_program(**kwargs)
-        assert result is None
+        # If not called they yet, thus empty tuple.
+        self.csdfg_args = ()
+
+    def process_arguments(self, **kwargs: Any) -> None:
```
egparedes: I find the name of this function not informative at all because it's too generic. What do you think about something like `prepare_args_list` ?

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -56,12 +63,33 @@ def __init__(
...
+            processed_csdfg_args = self.sdfg_program.construct_arguments(**kwargs)
+        assert isinstance(processed_csdfg_args, tuple) and len(processed_csdfg_args) == 2
+        # Note we only care about the first argument vector, that is used in normal call.
+        #  Since we update it, we ensure that it is a `list`.
+        self.csdfg_args = (list(processed_csdfg_args[0]), processed_csdfg_args[1])
 
     def fast_call(self) -> None:
-        result = self.sdfg_program.fast_call(*self.sdfg_program._lastargs)
+        """Perform a call to the compiled SDFG using the processed arguments, see `self.process_arguments()`."""
+        assert isinstance(self.csdfg_args, tuple) and len(self.csdfg_args) == 2
```
egparedes:
```suggestion
```

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -45,15 +44,14 @@ def decorated_program(
             args = (*args, out)
 
         try:
-            # Initialization of `_lastargs` was done by the `CompiledSDFG` object,
-            #  so we just update it with the current call arguments.
-            update_sdfg_call_args(args, fun.sdfg_program._lastargs[0])
-            fun.fast_call()
+            # Not the first call. We will only update the first argument vector (normal call).
```
egparedes:
```suggestion
            # Not the first call. We will only update the first argument list (normal call).
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -36,6 +37,12 @@ class CompiledDaceProgram(stages.CompiledProgram):
         None,
     ]
 
+    # Processed argument vectors that are passed to `CompiledSDFG.fast_call()`. If the
```
egparedes: I don't think `argument vectors` is the best name for this. In Python arguments are tuples or any other sequence which can be unpacked, so arguments list or arguments sequence feels more appropriate and less awkward.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
egparedes: As mentioned below.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -45,15 +44,14 @@ def decorated_program(
...
+            # NOTE: If this is the first call then we will generate an exception because
+            #   `fun.csdfg_args` is still the empty tuple.
```
egparedes:
```suggestion
            #   `fun.csdfg_args` is still the empty
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
philip-paul-mueller: You are right the name is not that good, however, I am also not super happy about the `prepare_args_list`.
I would suggest `prepare_arguments`.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
philip-paul-mueller: I understand your reasoning, however, I think that in this case _vector_ is the better name, because we go to C and there we have `agrv` (okay we do not call `main()`).

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
philip-paul-mueller: Okay there are two things here.
First, I told you that we do not need the initialization arguments, there I was wrong.
We only need it the first time we do the call and then never again, we do not even update it.
So we could technically remove it, but I think it would require some additional logic that makes the code harder to understand so I would propose to keep it.
If you really want it to be removed, I can do that.

The second thing is, that I fully agree with the unpacking.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
philip-paul-mueller: I would keep that because it tests if `prepare_arguments()` has been called or not.
I know you get an error in the unpacking, but I would prefer some more information about that.
However, I added a description to the `assert`.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
philip-paul-mueller: I think changing it into a `None` is actually better.
Because in the first call it will raise a `TypeError` instead an `IndexError`.
And this might hide some other bugs in the update code.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -56,12 +64,33 @@ def __init__(
         # For debug purpose, we set a unique module name on the compiled function.
         self.update_sdfg_ctype_arglist.__module__ = os.path.basename(program.sdfg.build_folder)
 
-    def __call__(self, **kwargs: Any) -> None:
-        result = self.sdfg_program(**kwargs)
-        assert result is None
+        # Since the sdfg program hasn't been called yet.
+        self.csdfg_args = None
+
+    def prepare_arguments(self, **kwargs: Any) -> None:
```
edopao: I am OK with the current name, however I would have called it `construct_arguments()`

pyproject.toml
```diff
@@ -240,6 +240,11 @@ disallow_incomplete_defs = false
 disallow_untyped_defs = false
 module = 'gt4py.next.iterator.*'
 
+[[tool.mypy.overrides]]
+disallow_incomplete_defs = false
+disallow_untyped_defs = false
+module = 'gt4py.next.program_processors.runners.dace_iterator.*'
+
```
edopao, line 247: Remember to remove these lines.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -56,12 +64,33 @@ def __init__(
...
+        """
+        This function will process the arguments and store the processed values in `self.csdfg_args`,
+        to call them use `self.fast_call()`.
+        """
+        with dace.config.set_temporary("compiler", "allow_view_arguments", value=True):
+            csdfg_agrv, csdfg_init_argv = self.sdfg_program.construct_arguments(**kwargs)
```
egparedes:
```suggestion
            csdfg_argv, csdfg_init_argv = self.sdfg_program.construct_arguments(**kwargs)
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -56,12 +64,33 @@ def __init__(
...
+        # Note we only care about the first argument vector, that is used in normal call.
+        #  Since we update it, we ensure that it is a `list`.
+        self.csdfg_args = ([*csdfg_agrv], csdfg_init_argv)
```
egparedes:
```suggestion
        self.csdfg_argv = [*csdfg_argv]
        self.csdfg_init_argv = csdfg_init_argv
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
egparedes: I agree

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -45,15 +44,15 @@ def decorated_program(
...
-        except IndexError:
-            # First call, the SDFG is not initialized, so forward the call to `CompiledSDFG`
-            # to properly initialize it. Later calls to this SDFG will be handled through
-            # the `fast_call()` API.
-            assert len(fun.sdfg_program._lastargs) == 0  # `fun.sdfg_program._lastargs` is empty
+            # Not the first call.
+            #  We will only update the first argument vector (arguments for the normal call).
+            # NOTE: If this is the first time then we will generate an exception because
+            #   `fun.csdfg_args` is `None`
+            update_sdfg_call_args(args, fun.csdfg_args[0])  # type: ignore[index]  # Will error out in first call.
```
egparedes:
```suggestion
            update_sdfg_call_args(args, fun.csdfg_argv])  # type: ignore[index]  # Will error out in first call.
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -56,12 +64,33 @@ def __init__(
...
 
     def fast_call(self) -> None:
-        result = self.sdfg_program.fast_call(*self.sdfg_program._lastargs)
+        """Perform a call to the compiled SDFG using the processed arguments, see `self.prepare_arguments()`."""
+        assert isinstance(self.csdfg_args, tuple) and len(self.csdfg_args) == 2, (
+            "Argument vector was not set properly."
+        )
+        self.sdfg_program.fast_call(*self.csdfg_args)
```
egparedes:
```suggestion
        self.sdfg_program.fast_call(self.csdfg_argv, self.csdfg_init_argv)
```

egparedes (approved):
LGTM

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
philip-paul-mueller: Me too.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -64,26 +65,28 @@ def __init__(
...
+        self.csdfg_init_argv = csdfg_init_argv
 
     def fast_call(self) -> None:
         """Perform a call to the compiled SDFG using the processed arguments, see `self.prepare_arguments()`."""
-        assert isinstance(self.csdfg_args, tuple) and len(self.csdfg_args) == 2, (
+        assert self.csdfg_argv is not None and self.csdfg_init_argv is not None, (
             "Argument vector was not set properly."
         )
-        self.sdfg_program.fast_call(*self.csdfg_args)
+        self.sdfg_program.fast_call(self.csdfg_argv, self.csdfg_init_argv, do_gpu_check=False)
```
edopao, line 89: Are we really sure we want to pass `do_gpu_check=False`? Do you know what is its overhead? Maybe we could pass `do_gpu_check=__debug__`?

pyproject.toml
egparedes: @philip-paul-mueller did you actually remove these lines and the "and I like that book" comment of the `dace-next` version before merging?

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
egparedes: @philip-paul-mueller , any thoughts on this comment?

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -56,12 +65,35 @@ def __init__(
...
+        with dace.config.set_temporary("compiler", "allow_view_arguments", value=True):
+            csdfg_argv, csdfg_init_argv = self.sdfg_program.construct_arguments(**kwargs)
+        # Note we only care about `csdfg_argv` (normal call), since we have to update it,
+        #  we ensure that it is a `list`.
+        self.csdfg_argv = [*csdfg_argv]
+        self.csdfg_init_argv = csdfg_init_argv
 
     def fast_call(self) -> None:
-        result = self.sdfg_program.fast_call(*self.sdfg_program._lastargs)
+        """Perform a call to the compiled SDFG using the processed arguments, see `self.prepare_arguments()`."""
```
egparedes, line 85: @philip-paul-mueller update `self.prepare_arguments()` to `self.construct_arguments()`

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
philip-paul-mueller: https://github.com/GridTools/gt4py/pull/2365

We should have added something here.

PR #2354, opened 2025-10-30
refactor[next][dace]: Cleanup SDFG lowering of tuples

edopao:
Improves the code for SDFG-lowering of tuples:
- Adoption of gt4py `eve.extended_typing` for tuple type annotations (`MaybeNestedInTuple`, `NestedTuple`).
- Simplification of field operator lowering, by dropping the support for tuple iterator (not used in fieldview GTIR).
- Refactoring code to allow removing some redundant functions, e.g. `get_tuple_type()` and `get_arg_symbol_mapping()`.
- Cleanup of test markers.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
@@ -246,13 +246,11 @@ def _create_scan_field_operator(
 
     # Note that `output_symbol` below is not used, we only need the tree-like
     # structure to get the type of each nested field in the `tree_map` visitor.
-    output_symbol: MaybeNestedInTuple[gtir.Sym]
-    if isinstance(node_type, ts.FieldType):
-        output_symbol = im.sym("__gtir_guaranteed_unused_dummy_variable", node_type)
-    else:
-        output_symbol = gtir_to_sdfg_utils.make_symbol_tree(
-            "__gtir_guaranteed_unused_dummy_variable", node_type
-        )
+    dummy_output_symbol = (
+        gtir_to_sdfg_utils.make_symbol_tree("__gtir_unused_dummy_var", node_type)
+        if isinstance(node_type, ts.TupleType)
+        else im.sym("__gtir_unused_dummy_var", node_type)
+    )
 
     return gtx_utils.tree_map(
         lambda edge, sym: (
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -245,28 +244,28 @@ def _create_scan_field_operator(
...
-                    ctx,
-                    sdfg_builder,
-                    field_domain,
-                    output_edge,
-                    output_sym.type,
-                    map_exit,
-                )
+        output_symbol = gtir_to_sdfg_utils.make_symbol_tree(
+            "__gtir_guaranteed_unused_dummy_variable", node_type
+        )
```
philip-paul-mueller: I am not sure but in this case
```python
dummy_output_symbol = x if y else z
```
_might_ be better, but maybe not.
You can keep it that way.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -665,8 +652,8 @@ def translate_scan(
 
     # build the mapping of symbols from nested SDFG to field operator context
     nsdfg_symbols_mapping = {str(sym): sym for sym in lambda_ctx.sdfg.free_symbols}
-    for psym, arg in lambda_args_mapping:
-        nsdfg_symbols_mapping |= gtir_to_sdfg_utils.get_arg_symbol_mapping(psym.id, arg, ctx.sdfg)
+    for pname, arg in lambda_arg_nodes.items():
```
philip-paul-mueller, line 653: This is my personal taste, feel free to ignore.
However, here you changed the name to `pname` to indicate that it is the "parameter name", but `arg` kind of does not "resonate" that much with it.
So I would change `arg` to something like `pvalue` or simply `parameter`.
My main point is, argument and parameter are most likely interchangeable but they are different words, thus I would stick to one of them.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: I like your suggestion, fixed.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: I see. I have applied some renaming, which I hope makes some clarity.

PR #2355, opened 2025-10-31
perf[next][dace]: Remove temporary inside scan map scope

edopao:
This PR removes a transient array inside the scan map scope. This array was used for temporary storage of the scan column computation. It was there in the beginning to overcome some issue with mapping of strides and dynamic array allocation.
- [x] This change verifies in ICON4Py CI and blueline run
- [x] Performance regression test

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
@@ -270,10 +270,12 @@ class DataflowOutputEdge:
         dest: dace.nodes.AccessNode,
         dest_subset: dace_subsets.Range,
     ) -> bool:
-        """Create the connection.
+        """Create a connection to the `dest` node, writing the given `dest_subset`.
 
         Might remove the last data container. Returns `True` if it was removed,
-        `False` if kept.
+        `False` if kept. For edges writing data from a nested SDFG, in case the
+        outside data container is removed, the caller is responsible to propagate
+        the strides of the destination array to the array inside the nested SDFG.
         """
         dest_desc = self.result.dc_node.desc(self.state)
         write_edge = self.state.in_edges(self.result.dc_node)[0]
@@ -285,12 +287,16 @@ class DataflowOutputEdge:
             remove_last_node = True
         elif isinstance(write_edge.src, dace.nodes.NestedSDFG):
             if isinstance(dest_desc, dace.data.Scalar):
-                # This scalar is needed for nested SDFGs implementing reduction
-                # with WCR memlet on the output connection.
+                # We keep scalar temporary storage, as a general rule, since it
+                # does not affect performance of the generated code. This scalar
+                # is only required in some cases, e.g. for nested SDFGs implementing
+                # reduction, which use a WCR memlet for the reduction operation.
                 remove_last_node = False
             else:
                 # We remove the transient array on the output connection of a nested
                 # SDFG and write directly to the destination node.
+                # The caller is responsible to propagate the strides of the destination
+                # array to the array inside the nested SDFG.
                 remove_last_node = True
         else:
             remove_last_node = False
--- a/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
@@ -161,29 +161,33 @@ def _create_scan_field_operator_impl(
     )
     field_node = ctx.state.add_access(field_name)
 
-    # Now connect the output connector on the nested SDFG with the result field.
+    # Now connect the output connector on the nested SDFG with the result field
+    #  outside the scan map scope. For 1D domain, containing only the column dimension,
+    #  there is no map scope, since the map range is only on the horizontal domain.
     #  Up to now the nested SDFG is writing into a transient data container that
     #  has the size to hold one column. The function below, that does the connection,
     #  will remove that transient and write directly to the result field.
     inner_map_output_temporary_removed = output_edge.connect(map_exit, field_node, field_subset)
-    assert inner_map_output_temporary_removed
+    if not inner_map_output_temporary_removed:
+        raise ValueError("The scan nested SDFG is expected to write directly to the result field.")
 
     assert ctx.state.in_degree(field_node) == 1
     field_node_path = ctx.state.memlet_path(next(iter(ctx.state.in_edges(field_node))))
     assert field_node_path[-1].dst is field_node
 
     # The temporary node which the nested SDFG was writing to has been deleted,
-    # and the nested SDFG will write directly to the result field outside. Thus,
-    # we have to modify the stride of the scan column array inside the nested SDFG
-    # to match the strides outside.
+    #  and the nested SDFG will write directly to the result field. Thus, we have
+    #  to modify the stride of the scan column array inside the nested SDFG to match
+    #  the strides outside.
     nsdfg_scan = field_node_path[0].src
     assert isinstance(nsdfg_scan, dace.nodes.NestedSDFG)
     inner_output_name = field_node_path[0].src_conn
     inner_output_desc = nsdfg_scan.sdfg.arrays[inner_output_name]
     assert len(inner_output_desc.shape) == 1
 
-    # The result field on the outside is a transient array, so its stride is constant.
-    # We just need to set it on the inside array, without symbol mapping.
+    # The result field on the outside is a transient array, allocated inside this
+    # function, so we know that its stride is constant. We just need to set it on
+    # the inside array, and we do not need to map any stride symbol.
     outside_output_stride = field_desc.strides[scan_dim_index]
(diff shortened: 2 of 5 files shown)


src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -275,35 +275,22 @@ def connect(
...
-                # Temporary data with compile-time size is allocated on the stack
-                # and therefore is safe to keep. We decide to keep it as a workaround
-                # for a dace issue with memlet propagation in combination with
-                # nested SDFGs containing conditional blocks. The output memlet
-                # of such blocks will be marked as dynamic because dace is not able
-                # to detect the exact size of a conditional branch dataflow, even
-                # in case of if-else expressions with exact same output data.
+            if isinstance(dest_desc, dace.data.Scalar):
+                # This scalar is needed for nested SDFGs implementing reduction
+                # with WCR memlet on the output connection.
```
philip-paul-mueller: Is it always a given that in that case we have a reduction or is it just a 99% of the time thing?
If so I would reformulate the comment a bit.
But I am in favour of keeping the scalar.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -275,35 +275,22 @@ def connect(
...
                 remove_last_node = False
             else:
-                # In case the output data has runtime size it is necessary to remove
-                # it in order to avoid dynamic memory allocation inside a parallel
-                # map scope. Otherwise, the memory allocation will for sure lead
-                # to performance degradation, and eventually illegal memory issues
-                # when the gpu runs out of local memory.
+                # We remove the transient array on the output connection of a nested
```
philip-paul-mueller, line 296: Maybe add a note about the stride and GPU where it is adapted if needed.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -161,51 +161,32 @@ def _create_scan_field_operator_impl(
...
 
-    # Now connect the final output node with the output of the nested SDFG.
+    # Now connect the output connector on the nested SDFG with the result field.
     #  Up to now the nested SDFG is writing into a transient data container that
     #  has the size to hold one column. The function below, that does the connection,
-    #  will remove that data container.
+    #  will remove that transient and write directly to the result field.
     inner_map_output_temporary_removed = output_edge.connect(map_exit, field_node, field_subset)
-    assert ctx.state.in_degree(field_node) == 1
+    assert inner_map_output_temporary_removed
```
philip-paul-mueller: I think this should be a `raise` because `output_edge.connect()` is technically _allowed_ to return `False`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -161,51 +161,32 @@ def _create_scan_field_operator_impl(
...
+    nsdfg_scan = field_node_path[0].src
+    assert isinstance(nsdfg_scan, dace.nodes.NestedSDFG)
+    inner_output_name = field_node_path[0].src_conn
+    inner_output_desc = nsdfg_scan.sdfg.arrays[inner_output_name]
+    assert len(inner_output_desc.shape) == 1
+
+    # The result field on the outside is a transient array, so its stride is constant.
+    # We just need to set it on the inside array, without symbol mapping.
+    outside_output_stride = field_desc.strides[scan_dim_index]
+    assert str(outside_output_stride).isdigit()
```
philip-paul-mueller, line 192: At first I struggled to understand the meaning of this `assert`, until I realized that it was here to ensure that you do not need to update the symbol mapping.

However, I am not sure if this `assert` is legal in all cases.
We know use the outer data container directly, however, that thing might have a symbolic size, if we directly write into an argument, which is global.

Thus, I would say in case `outside_output_stride` is a value then we just do what you already do.
However, in case it is symbolic (which I think is legal, please correct me if this is never ever possible) then I think you have to use the symbol mapping.
I think there might actually be a helper function in `strides.py` that does this already, but I am not sure how brittle they are.
If you you decide to update the symbol mapping manually, then please add a TODO to revisit this once the functions in `strides.py` have been updated, something that I want to do since a long time.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -161,51 +161,32 @@ def _create_scan_field_operator_impl(
...
+    inner_output_desc.set_shape(new_shape=inner_output_desc.shape, strides=(outside_output_stride,))
```
philip-paul-mueller, line 193: I think in that case you could also write:
```suggestion
    inner_output_desc.strides = (outside_output_stride,)
```
It is in my view less confusing.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -724,14 +721,7 @@ def apply(
...
                 block_size = list(self.block_size_1d)
 
-            # Launchbounds, if we have a scan and it is not specified use a special launch bound.
             launch_bounds = self.launch_bounds_1d
-            if launch_bounds is None:
-                for node in graph.scope_subgraph(
-                    self.map_entry, include_entry=False, include_exit=False
-                ):
-                    if isinstance(node, dace_nodes.NestedSDFG) and node.label.startswith("scan_"):
-                        launch_bounds = "512"  # Use high launch bound in case of scans to limit register usage and increase occupancy
```
philip-paul-mueller, line 734: Is this a bad merge?

src/gt4py/next/program_processors/runners/dace/transformations/strides.py
```diff
@@ -63,15 +63,15 @@ def gt_change_transient_strides(
         return sdfg
 
     for nsdfg in sdfg.all_sdfgs_recursive():
-        _gt_change_transient_strides_non_recursive_impl(nsdfg)
```
philip-paul-mueller, line 66: I can not mark it, but you need also to rename the surrounding function, i.e. `gt_change_transient_strides()`.

src/gt4py/next/program_processors/runners/dace/transformations/strides.py
```diff
@@ -104,19 +104,20 @@ def _gt_change_transient_strides_non_recursive_impl(
             continue
 
         # We assume that everything is in C order initially, to get FORTRAN order
-        #  we simply have to reverse the order.
+        #  we simply have to reverse the order. This is necessary only for transient
+        #  access nodes because the non-transients come from outside and have their
+        #  own strides.
         # TODO(phimuell): Improve this.
```
philip-paul-mueller:
```suggestion
        # TODO(phimuell): Set the stride based on the actual access pattern.
```

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: I was also in doubt, and I agree now.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: The result field, which `field_desc` refers to, is allocated inside this function, as a temporary array. So we know by construction that its stride is always constant. I will update the comment.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: True, but I prefer to keep the call to the set function.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
edopao: @iomaganaris please check this comment

iomaganaris: It's an intentional change. The reason why the high launch bound was set was to limit register usage. With the removal of tha large static array with this PR there is no high register usage any more in the scans so we can remove the launch bounds restriction. There might be some other small modificiations in the configuration of the scan maps necessary to improve their performance but they can come in another PR

philip-paul-mueller: Then it is fine.

PR #2359, opened 2025-11-03
fix[next][dace]: More strict check on node sequencing in `gt_create_local_double_buffering`

edopao:
This change addresses some issues encountered in dace transformations, e.g. inside `move_dataflow_into_if_body` transformation, caused by a `null -> null` edge connecting a double-buffer data node to some other node.
This edge is added to enforce sequencing of write after read. The change is to apply a more strict check on when the sequencing edge is needed.


philip-paul-mueller (approved):
LGTM

edopao:
Performance looks good (although some variability):

PR #2363, opened 2025-11-04
fix[next][dace]: Disable dace cache during pytest session

edopao:


src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -54,7 +55,9 @@ def set_dace_config(
     #   is translating from `CompilableProgram` (ITIR.Program + CompileTimeArgs)
     #   to `ProgramSource`, so this step is storing in cache only the result
     #   of the SDFG transformations, not the compiled program binary.
-    dace.Config.set("compiler.use_cache", value=True)
+    #   Additional note: we have observed segfault errors in pytest sessions,
+    #   during loading of compiled SDFG from .so file, so we disable the cache.
+    dace.Config.set("compiler.use_cache", value=("pytest" not in sys.modules))
```
philip-paul-mueller: In the CI this is certainly not a problem, but testing locally is, because even if persient cache is enabled it will not be generated.
So I would say if would fist check if persistent cache is explicitly requested if so no nothing, if not, then I would check with pytest.

As an alternative, we could also define a special environment variable such as `GT4PY_RUN_IN_CI` or something.
But I am not sure if this is an overblown solution.

edopao: Good point, I will add the check on the persistent cache.

philip-paul-mueller:
I think we can close this PR as the [update of DaCe](https://github.com/GridTools/gt4py/pull/2367) should solve the issue.

PR #2364, opened 2025-11-04
bug[next]: fix timer precision loss

havogt:
Previously we were computing the floating point differences of time since epoch, i.e. cancellation such that resulting precision was only in the order of microseconds.

In Python use `time.perf_counter()` instead. In gtfn, we do the computation with std::chrono::duration, in DaCe we go via nanoseconds (as int64) since epoch, as we cannot use a std::chrono::time_point in the interface easily.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ dace-cartesian = [
   'dace>=1.0.2,<2'  # renfined in [tool.uv.sources]
 ]
 dace-next = [
-  'dace==2025.10.30'  # refined in [tool.uv.sources]
+  'dace==2025.11.04'  # refined in [tool.uv.sources] and I like that book.
 ]
 dev = [
   {include-group = 'build'},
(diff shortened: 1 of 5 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -243,11 +243,11 @@ def add_instrumentation(sdfg: dace.SDFG, gpu: bool) -> None:
...
-double run_cpp_end_time = static_cast<double>(
-    std::chrono::duration_cast<std::chrono::nanoseconds>(
-        std::chrono::high_resolution_clock::now().time_since_epoch()).count()
-) / 1e9;
-duration = run_cpp_end_time - run_cpp_start_time;
+auto now = std::chrono::high_resolution_clock::now();
+auto run_cpp_end_time = std::chrono::duration_cast<std::chrono::nanoseconds>(
+        now.time_since_epoch()
+    ).count();
+duration = static_cast<double>(run_cpp_end_time - run_cpp_start_time) / 1.e9;
```
egparedes: Feel free to ignore but I think this makes more sense:
```suggestion
duration = static_cast<double>(run_cpp_end_time - run_cpp_start_time) * 1.e-9;
```

PR #2368, opened 2025-11-05
feat[next]: Compiled variant for field operators

tehrengruber:
In #1968 we introduced pre- and jit-compilation of static arguments. In the first version this was only support for programs, but not field operators which were directly called. This PR extends to functionality to also support field operators and scans. 

A good amount of technical debt which accumulated around the `Backend.__call__` method was also removed as it became obsolete with the unified approach.

Since scans are the only type of callable with generic arguments and there is a good amount of technical debt surrounding them that was beyond the scope of this PR, the first version barely implements something working for them and prints a warning as performance is likely not great. If performance is needed a simple workaround is to wrap the scan into a field operator or program.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/program.py
+++ b/src/gt4py/next/program_processors/runners/dace/program.py
@@ -171,7 +171,9 @@ class Program(decorator.Program, dace.frontend.python.common.SDFGConvertible):
 
         # Build the closure dictionary
         closure_dict: dict[str, dace.data.Array] = {}
-        offset_provider_type = gtx_common.offset_provider_to_type(self.compilation_options.connectivities)
+        offset_provider_type = gtx_common.offset_provider_to_type(
+            self.compilation_options.connectivities
+        )
         for conn_id, conn in used_connectivities.items():
             if conn_id not in self.connectivity_tables_data_descriptors:
                 self.connectivity_tables_data_descriptors[conn_id] = dace.data.Array(


src/gt4py/next/ffront/foast_to_past.py
tehrengruber, line 96: This function used to be wrong. The out argument was previously passed in `kwarg_types`, but this function just ignored all values in `kwarg_types` and just added `out` manually again. With the CompiledProgramPool the out argument is just the third argument and not a keyword argument anymore (consistent with the program definition below). Since this function should be reworked or replaced I didn't bother and just dropped the manual addition and added some sanity checks.

src/gt4py/next/ffront/decorator.py
```diff
@@ -299,105 +353,6 @@ def __call__(
...
-    def freeze(self) -> FrozenProgram:
-        if self.backend is None:
-            raise ValueError("Can not freeze a program without backend (embedded execution).")
-        return FrozenProgram(
-            self.definition_stage if self.definition_stage else self.past_stage,
-            backend=self.backend,
-        )
-
-
-@dataclasses.dataclass(frozen=True)
```
tehrengruber, line 355: This is not used anywhere including in icon4py and PMAP so I removed it instead of investing time into making it compatible with `_CommonProgramLike`.

src/gt4py/next/ffront/decorator.py
```diff
@@ -407,42 +362,6 @@ def __call__(self, *args: Any, offset_provider: common.OffsetProvider, **kwargs:
     pass
 
 
-# TODO(tehrengruber): This class does not follow the Liskov-Substitution principle as it doesn't
-#  have a program definition. Revisit.
-@dataclasses.dataclass(frozen=True)
```
tehrengruber, line 412: This is not used anywhere including in icon4py and PMAP so I removed it instead of investing time into making it compatible with `_CommonProgramLike`. Since `Backend.__call__` was removed it couldn't stay.

docs/user/next/advanced/ToolchainWalkthrough.md
```diff
@@ -1,400 +0,0 @@
-```python
```
tehrengruber, line 1: This file has become mostly obsolete with the introduction of `CompiledProgram`, hence I removed it here.

src/gt4py/next/ffront/decorator.py
```diff
@@ -38,20 +38,123 @@
...
 from gt4py.next.instrumentation import metrics
 from gt4py.next.iterator import ir as itir
-from gt4py.next.otf import arguments, compiled_program, stages, toolchain
+from gt4py.next.otf import arguments, compiled_program, options, toolchain
 from gt4py.next.type_system import type_info, type_specifications as ts, type_translation
 
 
 DEFAULT_BACKEND: next_backend.Backend | None = None
 
+DefinitionT = TypeVar(
```
egparedes: Not criticial but this typevar name is too generic. What about `DSLDefT` or something in that direction?

src/gt4py/next/ffront/decorator.py
```diff
@@ -38,20 +38,123 @@
...
+)
+
+
+@dataclasses.dataclass(frozen=True)
+class _ProgramLikeMixin(Generic[DefinitionT]):
+    """
+    Mixing used by program and program-like objects.
+
+    Contains functionality and configuration options common to all kinds of program-likes.
+    """
```
egparedes: I would suggest a different naming, in the same spirit as the previous suggestion:
```suggestion
class _DSLCallableMixin(Generic[DSLDefT]):
    """
    Mix-in used by directly callable GT4Py DSL artifacts like programs and operators.

    Contains functionality and configuration options common to all kinds of callable artifacts.
    """
```

src/gt4py/next/ffront/decorator.py
```diff
@@ -38,20 +38,123 @@
...
+
+    @abc.abstractmethod
+    def __gt_type__(self) -> ts.CallableType: ...
+
+    def with_backend(self, backend: next_backend.Backend) -> Self:
+        return dataclasses.replace(self, backend=backend)
+
+    def with_connectivities(
+        self,
+        connectivities: common.OffsetProvider,  # TODO(ricoh): replace with common.OffsetProviderType once the temporary pass doesn't require the runtime informatio
```
egparedes:
```suggestion
        connectivities: common.OffsetProvider,  # TODO(ricoh): replace with common.OffsetProviderType once the temporary pass doesn't require the runtime information
```

src/gt4py/next/otf/compiled_program.py
```diff
@@ -259,7 +267,26 @@ def __call__(
         else:
             args, kwargs = canonical_args, canonical_kwargs
         static_args_values = self._argument_descriptor_cache_key_from_args(*args, **kwargs)
-        key = (static_args_values, common.hash_offset_provider_items_by_id(offset_provider))
+
+        if self._is_generic:
+            warnings.warn(
+                "Calling generic programs / direct calls to scan operators are not been optimized. Consider calling a specialized version instead.",
```
egparedes:
```suggestion
                "Calling generic programs / direct calls to scan operators are not optimized. Consider calling a specialized version instead.",
```

src/gt4py/next/ffront/type_info.py
```diff
@@ -349,3 +350,88 @@ def return_type_scanop(
         ts.TypeSpec,
         tree_map_type(lambda arg: ts.FieldType(dims=promoted_dims, dtype=arg))(carry_dtype),
     )
+
+
+def type_in_program_context(callable_type: ts.CallableType) -> ProgramType | ts.FunctionType:
```
egparedes, line 355: Please add a meaningful docstring here, because the name of the function itself is not super clear.

src/gt4py/next/ffront/type_info.py
```diff
@@ -349,3 +350,88 @@ def return_type_scanop(
...
+    callable_type: ts.CallableType,
+) -> inspect.Signature:
+    if isinstance(callable_type, ts_ffront.ProgramType):
+        return _signature_from_callable_in_program_context(callable_type.definition)
+    elif isinstance(callable_type, ts_ffront.FieldOperatorType | ts_ffront.ScanOperatorType):
+        operator_signature = _signature_from_callable_in_program_context(callable_type.definition)
+        params = (
+            [*operator_signature.parameters.values()][1:]
+            if isinstance(callable_type, ts_ffront.ScanOperatorType)
+            else operator_signature.parameters.values()
```
egparedes: Just a suggestion but I find it more readable like this
```suggestion
        params = operator_signature.parameters.values()
        if isinstance(callable_type, ts_ffront.ScanOperatorType):
            params = params[1:]  # Remove the carry state arg
```

src/gt4py/next/ffront/type_info.py
```diff
@@ -349,3 +350,88 @@ def return_type_scanop(
...
+                for i, type_ in enumerate(callable_type.pos_only_args)
+            ]
+            + [
+                inspect.Parameter(name=name, kind=inspect.Parameter.POSITIONAL_OR_KEYWORD)
+                for name, type_ in callable_type.pos_or_kw_args.items()
+            ]
+            + [
+                inspect.Parameter(name=name, kind=inspect.Parameter.KEYWORD_ONLY)
+                for name, type_ in callable_type.kw_only_args.items()
+            ]
```
egparedes, line 444:
```suggestion
            [
                *(inspect.Parameter(name=str(i), kind=inspect.Parameter.POSITIONAL_ONLY)
                for i, type_ in enumerate(callable_type.pos_only_args)),
                *(inspect.Parameter(name=name, kind=inspect.Parameter.POSITIONAL_OR_KEYWORD)
                for name, type_ in callable_type.pos_or_kw_args.items()),
                *(inspect.Parameter(name=name, kind=inspect.Parameter.KEYWORD_ONLY)
                for name, type_ in callable_type.kw_only_args.items())
            ]
```

src/gt4py/next/otf/compiled_program.py
egparedes:
```suggestion
                "Calling generic programs / direct calls to scan operators are not optimized. Consider calling a specialized version instead.",
```

src/gt4py/next/otf/compiled_program.py
```diff
@@ -259,7 +267,26 @@ def __call__(
...
+                stacklevel=2,
+            )
+            arg_specialization_key = eve_utils.content_hash(
+                (
+                    tuple(type_translation.from_value(arg) for arg in canonical_args),
+                    {k: type_translation.from_value(v) for k, v in canonical_kwargs.items()},
+                )
+            )
+        else:
+            arg_specialization_key = None
```
egparedes, line 288: I think the need for this extra key component should be explained in a comment

src/gt4py/next/otf/compiled_program.py
```diff
@@ -295,9 +328,22 @@ def __call__(
                 )  # passing `enable_jit=False` because a cache miss should be a hard-error in this call`
             raise RuntimeError("No program compiled for this set of static arguments.") from e
 
+    @functools.cached_property
+    def _is_generic(self) -> bool:
```
egparedes, line 337: I find this name slightly misleading since it's not really clear what `generic` means in the context of gt4py. I'd rather change it to something less fancy with less potential for confusion like `has_deferred_type_args` or something similar...

src/gt4py/next/otf/compiled_program.py
```diff
@@ -393,6 +439,10 @@ def _compile_variant(
         self,
         argument_descriptors: ArgumentDescriptors,
         offset_provider: common.OffsetProviderType | common.OffsetProvider,
+        arg_types: tuple[ts.TypeSpec, ...] | None = None,
+        kwarg_types: dict[str, ts.TypeSpec] | None = None,
+        # argument used only to validate key computed in a call / dispatch agrees with the
+        # key computed here
```
egparedes: For consistency with the key computation above, I think it could be better to add only a single optional parameter 
```suggestion
        arg_specialization_info: tuple[tuple[ts.TypeSpec, ...], dict[str, ts.TypeSpec]] | None = None,
        # argument used only to validate key computed in a call / dispatch agrees with the
        # key computed here
```

src/gt4py/next/otf/compiled_program.py
```diff
@@ -431,12 +482,22 @@ def _compile_variant(
...
             )
 
+        assert (arg_types is None) == (kwarg_types is None)
+        if arg_types is None or kwarg_types is None:
+            if self._is_generic:
+                raise ValueError(
+                    "Can not precompile generic program or scan operator without argument types."
+                )
+            arg_types = tuple(self.program_type.definition.pos_only_args) + tuple(
+                self.program_type.definition.pos_or_kw_args.values()
```
egparedes:
```suggestion
            arg_types = (*self.program_type.definition.pos_only_args, *self.program_type.definition.pos_or_kw_args.values())
```

src/gt4py/next/otf/options.py
```diff
@@ -0,0 +1,36 @@
...
+    # mostly important for testing. Users should not rely on it.
+    enable_jit: bool = dataclasses.field(default_factory=lambda: config.ENABLE_JIT_DEFAULT)
+    #: if the user requests static params, they will be used later to initialize CompiledPrograms
+    static_params: Sequence[str] | None = (
+        None  # TODO: describe that this value will eventually be a sequence of strings
+    )
+    # TODO(ricoh): replace with common.OffsetProviderType once the temporary pass doesn't require the runtime information
+    #: A dictionary holding static/compile-time information about the offset providers.
+    #: For now, it is used for ahead of time compilation in DaCe orchestrated programs,
+    #: i.e. DaCe programs that call GT4Py Programs -SDFGConvertible interface-.
```
egparedes, line 37: Would it be possible to make this more readable by adding vertical spaces between fields and/or move the code comments after the definition and keep the sphinx comments before or something like that?

src/gt4py/next/otf/options.py
```diff
@@ -0,0 +1,36 @@
...
+    connectivities: common.OffsetProvider | None = None
```
egparedes, line 38: I would add some check to make sure those two do not get out of sync, even as simple as just:

```python
assert CompilationOptionsArgs.__annotations__.keys() == CompilationOptions.__annotations__.keys()
```

src/gt4py/next/otf/compiled_program.py
tehrengruber: Doesn't help with the key since the types are only computed in the `generic` path so the key has to be assembled multiple times as before. Taken the suggestion nonetheless to avoid the potential argument inconsistency with only one being None.

src/gt4py/next/otf/options.py
tehrengruber: Move them after the definition doesn't feel right to me. Maybe in front?

src/gt4py/next/otf/compiled_program.py
tehrengruber: I've added a comment explaining what generic means and a todo that clarifies that the deferred type is merely a hack / workaround.

src/gt4py/next/ffront/decorator.py
tehrengruber: I went for `ProgramLikeDefinitionT` for now. If we really want to do `DSLCallable` we should also rename `CompiledProgram` to `CompiledDSLCallable` etc. This doesn't feel like a good fit for this PR.

src/gt4py/next/ffront/decorator.py
tehrengruber: See comment above.

src/gt4py/next/ffront/decorator.py
egparedes: I think `ProgramLike` is meaningless in this context and that the right solution would be exactly what you propose as non-fitting this PR: use `DSLCallable` here and rename `CompiledProgram` to `CompiledDSLCallable`. Anyway, unless @havogt has a strong opinion towards one of the options, this should not be a blocker for this PR.

src/gt4py/next/ffront/decorator.py
egparedes: See answer above.

egparedes (approved):
LGTM

PR #2370, opened 2025-11-05
feat[dace][next]: Added `DoubleWriteRemover` Transformation

philip-paul-mueller:
The transformation eliminates a pattern observed in some stencils.
Essentially transforming
```python
for i in range(N):
    temp[i] = ...
A[:, level1] = temp
A[:, level2] = temp
```
into
```python
for i in range(N):
    temp_scal = ...
    A[i, level1] = temp_scal
    A[i, level2] = temp_scal
```

TODO:
- [x] Passing [ICON4Py CI](https://github.com/C2SM/icon4py/pull/935)
- [x] Performance validation

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/utils.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/utils.py
@@ -719,7 +719,7 @@ def associate_dimmensions(
         ...     sbs2=dace_sbs.Range.from_string("0:10, 4"),
         ...     allow_trivial_dimensions=False,
         ... )
-        ({1: 0}, [0, 2], [])
+        ({1: 0}, [0, 2], [1])
 
     Note:
         This function can not perform arbitrary reshapes, such as flatten a multi


src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -418,3 +419,326 @@ def _get_copy_chain_mode(self, sdfg: dace.SDFG, state: dace.SDFG) -> CopyChainRe
...
+        temp_node: dace_nodes.AccessNode = self.temp_node
+        temp_desc = temp_node.desc(sdfg)
+
+        # Only the Map can produce data.
+        if graph.in_degree(temp_node) != 1:
+            return False
+
+        producer_edge = next(iter(graph.in_edges(temp_node)))
+        produced_sbs: dace_sbs.Range = producer_edge.data.get_dst_subset(producer_edge, graph)
+        if produced_sbs is None:
```
edopao: Is this even possible? this is a memlet edge passing through a MapExit node.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -418,3 +419,326 @@ def _get_copy_chain_mode(self, sdfg: dace.SDFG, state: dace.SDFG) -> CopyChainRe
...
+        assert outer_producer_edge.src_conn.startswith("OUT_")
+        inner_producer_edge = next(
+            iter(
+                inner_producer_edge
+                for inner_producer_edge in graph.in_edges_by_connector(
+                    map_exit, "IN_" + outer_producer_edge.src_conn[4:]
+                )
+            )
+        )
+        inner_producer_subset = inner_producer_edge.data.get_dst_subset(inner_producer_edge, graph)
```
iomaganaris, line 607: We call this logic in both `can_be_applied` and `apply`. Should it be moved to a separate function?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -418,3 +419,326 @@ def _get_copy_chain_mode(self, sdfg: dace.SDFG, state: dace.SDFG) -> CopyChainRe
...
+        inner_production_sbs = inner_producer_edge.data.get_dst_subset(inner_producer_edge, graph)
+        if inner_production_sbs.num_elements() != 1:
+            return False
+        if inner_producer_edge.data.allow_oob:
+            return False
+        if inner_producer_edge.data.is_empty():
+            return False
+        if inner_producer_edge.data.wcr is not None:
+            return False
+        # TODO(phimuell): Due to the special meaning of `dynamic` Memlets in GT4Py,
```
edopao: I have grepped `.dynamic`in the lowering to SDFG and I found no match. So there is no special meaning in GT4Py, I guess. However, I am in favor of keeping the check. Just reformulate the comment.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -418,3 +419,326 @@ def _get_copy_chain_mode(self, sdfg: dace.SDFG, state: dace.SDFG) -> CopyChainRe
...
+                for inner_producer_edge in graph.in_edges_by_connector(
+                    map_exit, "IN_" + outer_producer_edge.src_conn[4:]
+                )
+            )
+        )
+        inner_producer_subset = inner_producer_edge.data.get_dst_subset(inner_producer_edge, graph)
+
+        # Because we distribute to multiple consumers, we need some temporary storage
+        #  inside the Map scope, aka. AccessNode. We will first check if there is
+        #  already am AccessNode that can be used or if we have to create a new one.
```
iomaganaris:
```suggestion
        #  already an AccessNode that can be used or if we have to create a new one.
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -418,3 +419,326 @@ def _get_copy_chain_mode(self, sdfg: dace.SDFG, state: dace.SDFG) -> CopyChainRe
...
+            consumer_sbs: dace_sbs.Range = consumer_edge.data.get_src_subset(consumer_edge, graph)
+            if consumer_sbs is None:
+                return False
+
+            # What is produced must be consumed.
+            if consumer_sbs != produced_sbs:
+                return False
+
+            # Since `temp_node` is eliminated, the consumer must be another AccessNode.
+            # TODO(phimuell): Think of allowing that _some_ consumer are Maps.
```
edopao: I would remove the TODO. You cannot end up with MapExit-MapEntry edge.

src/gt4py/next/program_processors/runners/dace/transformations/utils.py
```diff
@@ -678,3 +678,108 @@ def _impl(
             all_successors.extend(successor.all_states())
 
     return all_successors
+
+
+def associate_dimmensions(
+    sbs1: dace_sbs.Range,
+    sbs2: dace_sbs.Range,
+    allow_trivail_dimensions: bool = True,
```
iomaganaris:
```suggestion
    allow_trivial_dimensions: bool = True,
```

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_double_write_remover.py
```diff
@@ -0,0 +1,339 @@
...
+        state.add_nedge(b, c, dace.Memlet("b[0:10] -> [0, 1:11]"))
+        state.add_nedge(b, c, dace.Memlet("b[0:10] -> [2, 0:10]"))
+
+    sdfg.validate()
+    return sdfg, state, b, c, mx
+
+
+# @pytest.mark.parametrize("slice_to_second", [True, False])
+def test_remove_double_write_multi_producer():  # slice_to_second: bool):
+    slice_to_second = True
```
iomaganaris: I guess you want to keep the parametrization of it?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -418,3 +419,326 @@ def _get_copy_chain_mode(self, sdfg: dace.SDFG, state: dace.SDFG) -> CopyChainRe
...
+            current_inner_producer_desc = current_inner_producer_node.desc(sdfg)
+            if isinstance(current_inner_producer_desc, dace_data.View):
+                reuse_inner_producer_node = False
+            elif isinstance(current_inner_producer_desc, dace_data.Scalar):
+                reuse_inner_producer_node = True
+            elif (
+                isinstance(current_inner_producer_desc, dace_data.Array)
+                and len(current_inner_producer_desc.shape) == 1
+                and (current_inner_producer_desc.shape[0] == 1) == True  # noqa: E712 [true-false-comparison]  # SymPy comparison
+            ):
```
edopao, line 625: Just for my understanding: how would the case `len(shape) != 1` look like? an inner array which is larger than the size it is written, therefore partially uninitialized?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -418,3 +419,326 @@ def _get_copy_chain_mode(self, sdfg: dace.SDFG, state: dace.SDFG) -> CopyChainRe
...
+
+                    # Compute the correction.
+                    consumer_correction = consumer_destination[consumer_dim][0]
+                    temp_correction = outer_producer_subset[temp_node_dim][0]
+                    correction = consumer_correction - temp_correction
+
+                    # Compose the final subset.
+                    inner_producer_idx = inner_producer_subset[temp_node_dim]
+                    inner_map_output_subset.append(
+                        (inner_producer_idx[0] + correction, inner_producer_idx[1] + correction, 1)
```
edopao, line 704: I see `step=1` here, was it checked anywhere?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
```diff
@@ -418,3 +419,326 @@ def _get_copy_chain_mode(self, sdfg: dace.SDFG, state: dace.SDFG) -> CopyChainRe
...
+                "OUT_" + connector_name,
+                consumer_edge.dst,
+                consumer_edge.dst_conn,
+                dace.Memlet(
+                    data=consumer_edge.dst.data,
+                    subset=consumer_edge.data.dst_subset,
+                    other_subset=None,
+                ),
+            )
+            map_exit.add_scope_connectors(connector_name)
```
edopao, line 745: Just out of curiosity: I have seen already in other transformations that the edges traversing a map are built this way. Why not using `add_memlet_path()` ?

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_associate_dimensions.py
edopao, line 1: Typo in file name: dimmensions -> dimensions

src/gt4py/next/program_processors/runners/dace/transformations/utils.py
```diff
@@ -678,3 +678,108 @@ def _impl(
...
+) -> tuple[dict[int, int], list[int], list[int]]:
+    """Computes the association between dimensions of two subsets.
```
edopao, line 688: One docstring runnable example would be useful to understand what this function does.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: I am actually not sure if it makes sense, because in `can_be_apply()` you do not know yet that there is only a single edge, while in `apply()` you know that.
I tried it, but I think that the current version is cleaner.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_double_write_remover.py
philip-paul-mueller: Yes you are right.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: I am pretty sure but that it is technically illegal.
However, I think that the DaCe code generator would accept it because that subset is technically not needed, since the "writing into the AccessNode (`temp_node`)" is not done through a `memcpy` that needs a traget, but through the subset of the Memlet _inside_ the Map.

However, I agree that such an SDFG should be considered invalid and I will turn it into an `assert`.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: This is interesting as far as I remember `concat_where` used `dynamic` Memlets or do I miss something?

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: The idea would be more to not delete `temp_node` but keep it for these nodes alive.
Thus I would keep it.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: `shape == (1, 1, 1, 1, 1,1 ,1)`
However, I would say we should handle that above, but it is a bit more nasty to handle.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: No you are right.
I added them.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: I tried it, but for some reasons Memlet propagation failed to do things properly.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
edopao: Yes, but at some point I removed it because it was no longer needed.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
edopao: I see, then the comment would be:
```suggestion
            # TODO(phimuell): Allow the consumer to be a Map, in which case `temp_node` should not be removed.`
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: This is new to me, but in that case we can remove them.

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
edopao: Oh right, I confused `len(desc.shape)` with `desc.total_size()`

src/gt4py/next/program_processors/runners/dace/transformations/utils.py
```diff
@@ -678,3 +678,123 @@ def _impl(
...
+    Returns:
+        The function returns a tuple with three elements. The first element is
+        a `dict` that maps each dimension of `sbs1` to a dimensions `sbs2`.
+        The second element is a `list` containing all dimensions that were dropped
+        from `sbs1`, because they where trivial. The third element is the same
+        as the second element, but for `sbs2`.
+
+    Example:
+        >>> import dace
+        >>> from dace import subset as dace_sbs
```
edopao: The problem is a missing `s` in `subsets`:
```suggestion
        >>> from dace import subsets as dace_sbs
```

src/gt4py/next/program_processors/runners/dace/transformations/redundant_array_removers.py
philip-paul-mueller: Thanks for fixing it.

PR #2372, opened 2025-11-05
refactor[next][dace]: Introduce `SDFGBuilder.add_nested_sdfg`

edopao:


src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -34,9 +34,13 @@
 
 from gt4py import eve
 from gt4py.eve import concepts
-from gt4py.next import common as gtx_common, utils as gtx_utils
+from gt4py.next import common as gtx_common, config, utils as gtx_utils
```
philip-paul-mueller: You prefix everything with `gtx` except `config` which is strange.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -248,7 +300,7 @@ def make_field(
         data_type: ts.FieldType,
     ) -> gtir_to_sdfg_types.FieldopData:
         """
-        Helper method to build the field data type associated with a data access node.
+        Implements `SDFGBuilder.make_field()`.
```
philip-paul-mueller: It never occurred to me, but this doc string should not exists, because it implements an interface function it should follow the interface to the letter.
Thus this doc string should either not exist or be an exact copy of the one describing the interface.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -179,22 +184,68 @@ def setup_nested_context(
             expr: The lambda expression to be lowered as a nested SDFG.
             sdfg_name: The name of the nested SDFG where to lower the given expression.
             parent_ctx: The parent SDFG context.
-            scope_symbols: Mapping from symbol name to data type for the GTIR symbols
-                forwarded to the nested context.
+            lambda_params: List of GTIR symbols passed as parameters to the lambda expression.
             symbolic_inputs: Arguments that have to be passed to the nested SDFG
                 as dace symbols.
+            capture_scope_symbols: When True, the lambda expression is allowed to
```
philip-paul-mueller: Is "allowed" the correct word here?
It has the connotation of "maybe it will maybe not", I would say you should use "will".

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -306,41 +360,52 @@ def setup_nested_context(
...
+        lambda_pnames = {str(p.id) for p in lambda_params}
+        assert symbolic_inputs.issubset(lambda_pnames) and all(
+            isinstance(p.type, ts.ScalarType) for p in lambda_params if str(p.id) in symbolic_inputs
+        )
+
+        # If `capture_scope_symbols` is True, besides the values mapped to the parameters
+        # of the lambda expression (`lambda_params`), we also pass to the nested SDFG
+        # all GTIR-symbols in scope, which means the symbols defined in current context.
+        if capture_scope_symbols:
+            scope_symbols = {
```
philip-paul-mueller: Why is this not an update of `lambda_pnames` or `lambda_symbols`?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -351,30 +416,119 @@ def flatten_tuple_syms(params: Iterable[gtir.Sym]) -> list[gtir.Sym]:
...
+        assert data_args.keys().isdisjoint(symbolic_args.keys())
+
+        # Collect the names of all output data, by flattening any tuple structure.
+        lambda_output_data = (
+            gtx_utils.flatten_nested_tuple(inner_result)
+            if isinstance(inner_result, tuple)
+            else [inner_result]
+        )
+        # The output connectors only need to be setup for the actual result of the
+        # internal dataflow that writes to some sink data nodes.
```
```suggestion
        # internal dataflow that writes to some sink data nodes of the nested SDFG.
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -351,30 +416,119 @@ def flatten_tuple_syms(params: Iterable[gtir.Sym]) -> list[gtir.Sym]:
...
+        # We filter out the data nodes that are passed as input arguments (non-transient
```
philip-paul-mueller: I do not see where you filter out the inputs.
I was expecting something like `data in input_data`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -850,113 +1003,31 @@ def visit_Lambda(
...
                 symbolic_args[gt_symbol_name] = arg
             else:
-                data_args[gt_symbol_name] = arg
                 lambda_arg_nodes |= {
-                    str(nested_param.id): nested_arg
-                    for nested_param, nested_arg in gtir_to_sdfg_types.flatten_tuple(gt_symbol, arg)
-                    if nested_arg is not None  # we filter out arguments with empty domain
+                    str(gt_symbol_.id): arg_
+                    for gt_symbol_, arg_ in gtir_to_sdfg_types.flatten_tuple(gt_symbol, arg)
+                    if arg_ is not None  # we filter out arguments with empty domain
```
philip-paul-mueller: I do not like the trailing underscore here, what was wrong with the old names?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -850,113 +1003,31 @@ def visit_Lambda(
...
-        assert all(isinstance(_type, ts.DataType) for _type in lambda_symbols.values())
-
         # lower let-statement lambda node as a nested SDFG
-        lamnda_translator, lambda_ctx = self.setup_nested_context(
+        lambda_translator, lambda_ctx = self.setup_nested_context(
             expr=node,
             sdfg_name="lambda",
             parent_ctx=ctx,
-            scope_symbols=lambda_symbols,  # type: ignore[arg-type]  # lambda_symbols checked by assert above
+            lambda_params=[param for param, arg in args.items() if arg is not None],
```
philip-paul-mueller: Would it make sense to move this filtering into `setup_nested_context()`?

philip-paul-mueller (commented):
I have some small comments, but it is okay.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Good point.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Good point. I was using a complicated way to perform this check.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: I didn't like `param`, I would have had to rename it to `nested_gt_symbol` but it was too long. I am going to change to `gtsym` and `nested_gtsym`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: `scope_symbols` is also used below to retrieve `parent_symbols`. I have modified this code.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: I think I could pass all arguments, indeed. This is also consistent with another upcoming change.

philip-paul-mueller (approved):
LGTM

PR #2373, opened 2025-11-05
fix[next][dace]: Use correct dtype for connectivity tables

edopao:
Fixes the lowering of unstructured shift, by using the `dtype` attribute in the connectivity descriptor instead of the default index builtin type (`int32`). Search in the new code the pattern `tt.from_dtype(conn_type.dtype)`.

Additional change: some code refactoring to make more consistent naming of `NeighborConnectivityType` variables. This PR uses the `conn_` prefix in variable names.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
@@ -31,12 +31,13 @@ from dace import subsets as dace_subsets
 from gt4py import eve
 from gt4py.eve.extended_typing import MaybeNestedInTuple, NestedTuple
 from gt4py.next import common as gtx_common, utils as gtx_utils
-from gt4py.next.iterator import builtins as gtir_builtins, ir as gtir
+from gt4py.next.iterator import ir as gtir
 from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm, ir_makers as im
 from gt4py.next.iterator.transforms import symbol_ref_utils
 from gt4py.next.program_processors.runners.dace import (
     gtir_python_codegen,
     gtir_to_sdfg,
+    gtir_to_sdfg_types,
     gtir_to_sdfg_utils,
     utils as gtx_dace_utils,
 )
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -1684,7 +1687,7 @@ def _visit_shift(self, node: gtir.FunCall) -> IteratorExpr:
         offset_provider_type = self.subgraph_builder.get_offset_provider_type(offset)
         # second argument should be the offset value, which could be a symbolic expression or a dynamic offset
         offset_expr = (
-            SymbolExpr(offset_value_arg.value, gtx_dace_utils.as_dace_type(_INDEX_DTYPE))
+            SymbolExpr(offset_value_arg.value, gtir_to_sdfg_types.INDEX_DTYPE)
```
philip-paul-mueller, line 1690: Where is `INDEX_DTYPE` defined?

philip-paul-mueller (approved):
LGTM.

There is one point I do not understand.

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
edopao: Here: https://github.com/GridTools/gt4py/blob/726a0293ba5c2f7dd2bbb0fae844889d8a45f638/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_types.py#L199

philip-paul-mueller: What confused me was that the it was not edited in this PR (at least I did not found it).
Why is it here the hardedcoded value and not the dynamic one?

edopao: This is the type used for an offset literal (a compile-time integer) to be used in some index computation. The index computation in the generated code is still using the `INTEGER_INDEX_BUILTIN` dtype, which is the same type used for the map variables. This PR changes only the type of storage for the connectivity arrays, but the array indexing remains the same.

PR #2375, opened 2025-11-06
fix[next]: Reuse parameters in direct fo calls

SF-N:
Previously, the parameter names in programs implicitly generated on a direct field operator call where of the form `__sym<uid>` and thus different to the names specified in the field operator definition. This breaks when used in combination with static parameters on field operators (see enabled in #2368). The replacement works by replacing all refs to arguments by their static value. For direct field operator calls this refs are now named the same as in their definition.


src/gt4py/next/ffront/foast_to_past.py
```diff
@@ -99,18 +98,24 @@ def __call__(self, inp: AOT_FOP) -> AOT_PRG:
...
                 location=loc,
             )
-            for arg_type in arg_types
+            for name, arg_type in zip(
+                [
+                    *definition.pos_only_args,
+                    *definition.pos_or_kw_args.keys(),
+                    *definition.kw_only_args.keys(),
+                ],
+                arg_types,
```
tehrengruber, line 120:
```suggestion
                arg_types,
                strict=True
```

tehrengruber (changes requested):
Issues in dace still need to be resolved.

PR #2376, opened 2025-11-06
refactor[next][dace]: Change representation of field symbolic layout

edopao:


edopao:
No performance gain from changing the SDFG representation of symbolic field layout.

PR #2377, opened 2025-11-07
fix[cartesian]: respect percision of constants

romanc:
## Description

In the `IRMaker`, we discard type information that is potentially available from inlined constants because e.g. `np.float64` is of instance `float`, which means, that the old order of the `if` statements would trigger for `isinstance(..., (int, float))` in cases that we don't want. We flip the order to ensure that we keep potential type annotations on e.g. global constants.

This is fallout from literal precision work (see PR #2187) and was reported (first hidden by other issues) in upstream issue https://github.com/NOAA-GFDL/NDSL/issues/195.

## Requirements

  N/A


FlorianDeconinck (approved):
LGTM

PR #2378, opened 2025-11-07
refactor[next][dace]: Rename array symbols for shape and strides

edopao:
This PR does some code refactoring according to the following principles: 
- The name of symbols for field stride now contains the dimension string instead of an index: `__w_stride_0` -> `__w_IDim_stride`. This follows the same principle already used for range symbols.
- For connectivities, the array name already contains information about the connectivity itself (e.g. `__gt_conn_C2E`), so the size/stride symbols contain `source`/`neighbor` string:
  - `__gt_conn_C2E_size_0` -> `__gt_conn_C2E_source_size`
  - `__gt_conn_C2E_stride_1` -> `__gt_conn_C2E_neighbor_stride`

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
@@ -142,10 +142,10 @@ def _parse_gt_param(
                         arg_range = f"{arg}.domain.ranges[{i}]"
                         rstart = gtx_dace_utils.range_start_symbol(param_name, dim)
                         rstop = gtx_dace_utils.range_stop_symbol(param_name, dim)
-                        for suffix, sdfg_symbol in [("start", rstart), ("stop", rstop)]:
+                        for suffix, sdfg_range_symbol in [("start", rstart), ("stop", rstop)]:
                             _parse_gt_param(
-                                param_name=sdfg_symbol.name,
-                                param_type=gtx_dace_utils.as_itir_type(sdfg_symbol.dtype),
+                                param_name=sdfg_range_symbol.name,
+                                param_type=gtx_dace_utils.as_itir_type(sdfg_range_symbol.dtype),
                                 arg=f"{arg_range}.{suffix}",
                                 code=code,
                                 sdfg_arglist=sdfg_arglist,
@@ -160,13 +160,13 @@ def _parse_gt_param(
                             f"assert {_cb_sdfg_argtypes}[{sdfg_arg_index}].strides[{i}] == {arg_stride}"
                         )
                     else:
-                        stride_symbol = gtx_dace_utils.field_stride_symbol(param_name, dim)
-                        assert array_stride == stride_symbol
+                        sdfg_stride_symbol = gtx_dace_utils.field_stride_symbol(param_name, dim)
+                        assert array_stride == sdfg_stride_symbol
                         # The strides of a global array are defined by a sequence
                         # of SDFG symbols.
                         _parse_gt_param(
-                            param_name=stride_symbol.name,
-                            param_type=gtx_dace_utils.as_itir_type(stride_symbol.dtype),
+                            param_name=sdfg_stride_symbol.name,
+                            param_type=gtx_dace_utils.as_itir_type(sdfg_stride_symbol.dtype),
                             arg=arg_stride,
                             code=code,
                             sdfg_arglist=sdfg_arglist,
--- a/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
@@ -52,18 +52,18 @@ def find_constant_symbols(
                     raise NotImplementedError(
                         f"Unsupported field with multiple horizontal dimensions '{p}'."
                     )
-                stride_symbol = gtx_dace_utils.field_stride_symbol(str(p.id), dim)
-                constant_symbols[stride_symbol.name] = 1
+                sdfg_stride_symbol = gtx_dace_utils.field_stride_symbol(str(p.id), dim)
+                constant_symbols[sdfg_stride_symbol.name] = 1
         # Same for connectivity tables, for which the first dimension is always horizontal
         connectivity_types = gtx_dace_utils.filter_connectivity_types(offset_provider_type)
         for offset, conn_type in connectivity_types.items():
             if (conn_id := gtx_dace_utils.connectivity_identifier(offset)) in sdfg.arrays:
                 assert not sdfg.arrays[conn_id].transient
                 assert conn_type.source_dim.kind == common.DimensionKind.HORIZONTAL
-                stride_symbol = gtx_dace_utils.field_stride_symbol(
+                sdfg_stride_symbol = gtx_dace_utils.field_stride_symbol(
                     conn_id, conn_type.source_dim, connectivity_types
                 )
-                constant_symbols[stride_symbol.name] = 1
+                constant_symbols[sdfg_stride_symbol.name] = 1
 
     if disable_field_origin_on_program_arguments:
         # collect symbols used as range start for all program arguments
@@ -84,11 +84,11 @@ def find_constant_symbols(
                     # zero-dimensional field
                     continue
                 # set all range start symbols to constant value 0
-                origin_symbols = [
+                sdfg_origin_symbols = [
                     gtx_dace_utils.range_start_symbol(str(psymbol.id), dim)
                     for dim in psymbol.type.dims
                 ]
-                constant_symbols |= {sdfg_symbol.name: 0 for sdfg_symbol in origin_symbols}
+                constant_symbols |= {sdfg_symbol.name: 0 for sdfg_symbol in sdfg_origin_symbols}
 
     return constant_symbols
 


src/gt4py/next/program_processors/runners/dace/gtir_python_codegen.py
```diff
@@ -76,28 +76,28 @@
...
     return f"{true_val} if {cond} else {false_val}"
 
 
-def builtin_tuple_get(index: str, tuple_name: str) -> str:
+def _builtin_tuple_get(index: str, tuple_name: str) -> str:
     return f"{tuple_name}_{index}"
 
 
-def make_const_list(arg: str) -> str:
+def _make_const_list(arg: str) -> str:
```
philip-paul-mueller: For consistency I would also add `builtin`.

src/gt4py/next/program_processors/runners/dace/gtir_python_codegen.py
```diff
@@ -152,8 +152,8 @@ def visit_FunCall(self, node: gtir.FunCall, args_map: dict[str, gtir.Node]) -> s
                 raise NotImplementedError(f"Unexpected deref with arg type '{type(node.args[0])}'.")
             return self.visit(node.args[0], args_map=args_map)
         elif isinstance(node.fun, gtir.SymRef):
-            args = self.visit(node.args, args_map=args_map)
             builtin_name = str(node.fun.id)
+            args = self.visit(node.args, args_map=args_map)
```
philip-paul-mueller, line 154: Just out of curiosity, why did you change the ordering here?

src/gt4py/next/program_processors/runners/dace/utils.py
```diff
@@ -71,52 +66,66 @@ def is_connectivity_identifier(
...
+
+def field_stride_symbol(
+    field_name: str,
+    dim: gtx_common.Dimension,
+    offset_provider_type: dict[str, gtx_common.NeighborConnectivityType] | None = None,
+) -> dace.symbol:
+    return _field_symbol(field_name, dim, "stride", offset_provider_type)
+
+
+def range_symbol_name(field_name: str, axis: str) -> str:
```
philip-paul-mueller: I have the feeling that this function should not be used directly, only the `range_{start, stop}_symbol()` function should be used.
Thus I am suggesting that you should change it to an internal function. i.e. adding `_`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -486,21 +482,21 @@ def _add_storage(
                 )
             if isinstance(gt_type.dtype, ts.ScalarType):
                 dc_dtype = gtx_dace_utils.as_dace_type(gt_type.dtype)
-                dims = gt_type.dims
+                full_dims = gt_type.dims
```
philip-paul-mueller: I am not sure of "full" is appropriate here, although I understand its meaning.
I would suggest something like `all_dims` (which is not much better) or `full_field_dims`.
However, I do not have strong opinions on it and you can keep it that way.

src/gt4py/next/program_processors/runners/dace/program.py
```diff
@@ -152,63 +153,45 @@ def __sdfg_closure__(self, reevaluate: Optional[dict[str, str]] = None) -> dict[
         closure_dict: dict[str, Any] = {}
```
philip-paul-mueller: As far as I can see this function only returns a non empty `dict` is `self.connectivities` is set.
So I would recommend you to refactor the function into something:
```python
def __sdfg_closure__():
    if not self.connectivities:
        return {}
    ...
```
It is not very readable with this big `if`.

src/gt4py/next/program_processors/runners/dace/program.py
```diff
@@ -152,63 +153,45 @@ def __sdfg_closure__(self, reevaluate: Optional[dict[str, str]] = None) -> dict[
...
         if self.connectivities:
-            symbols = {}
-            with_table = [
-                name for name, conn in self.connectivities.items() if common.is_neighbor_table(conn)
-            ]
-            in_arrays_with_id = [
-                (name, conn_id)
-                for name in with_table
-                if (conn_id := gtx_dace_utils.connectivity_identifier(name))
+            connectivity_types = gtx_dace_utils.filter_connectivity_types(
```
philip-paul-mueller: I would suggest to move the definition of this variable before the `for` loop where it is actually used.
It is too disconnected from its usage.

src/gt4py/next/program_processors/runners/dace/sdfg_callable.py
```diff
@@ -22,10 +22,10 @@
 def get_field_domain_symbols(name: str, domain: gtx_common.Domain) -> dict[str, int]:
     assert gtx_common.Domain.is_finite(domain)
     return {
-        gtx_dace_utils.range_start_symbol(name, dim): r.start
+        gtx_dace_utils.range_start_symbol(name, dim).name: r.start
```
philip-paul-mueller, line 25:
```suggestion
        str(gtx_dace_utils.range_start_symbol(name, dim)): r.start
```
Do you think this could be more readable?

src/gt4py/next/program_processors/runners/dace/utils.py
```diff
@@ -71,52 +66,66 @@ def is_connectivity_identifier(
 
 
 def is_connectivity_symbol(name: str, offset_provider_type: gtx_common.OffsetProviderType) -> bool:
-    m = SIZE_SYMBOL_RE.match(name) or STRIDE_SYMBOL_RE.match(name)
-    if m is None:
+    SYMBOL_RE: Final[re.Pattern] = re.compile(r"^__(\S+)_(\S+)_(size|stride)$")
```
philip-paul-mueller: You perform the recompilation every time you enter this function.
I would thus suggest to put its definition outside.

src/gt4py/next/program_processors/runners/dace/utils.py
```diff
@@ -71,52 +66,66 @@ def is_connectivity_identifier(
...
-def range_symbol(field_name: str, axis: str) -> str:
+def _field_symbol(
+    field_name: str,
+    dim: gtx_common.Dimension,
+    sym: Literal["size", "stride"],
+    offset_provider_type: dict[str, gtx_common.NeighborConnectivityType] | None,
+) -> dace.symbol:
+    if (m := CONNECTIVITY_INDENTIFIER_RE.match(field_name)) is None:
+        name = f"__{field_name}_{dim.value}_{sym}"
+    else:  # a connectivity field
```
philip-paul-mueller, line 89: Might this be dangerous?

src/gt4py/next/program_processors/runners/dace/gtir_python_codegen.py
edopao: Probably a rebase leftover from my previous PR, where I changed the symbolic layout. I will revert it.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: I will apply your suggestion `all_dims`, here and in other places.

src/gt4py/next/program_processors/runners/dace/program.py
edopao: Good idea.

src/gt4py/next/program_processors/runners/dace/sdfg_callable.py
edopao: I am not sure, accessing `.name` ensures - in a loose way though - that the object is a dace symbol.

src/gt4py/next/program_processors/runners/dace/utils.py
edopao: Right, I missed that.

src/gt4py/next/program_processors/runners/dace/utils.py
edopao: It is ensured by the fact that I look up the match result in `offset_provider_type`. I can add an assert to make it clear.

src/gt4py/next/program_processors/runners/dace/utils.py
edopao: It is directly only in one place, currently, in `gtir_python_codegen`. However, I have found a way to revert the dependency, which I like.

src/gt4py/next/program_processors/runners/dace/sdfg_callable.py
philip-paul-mueller: I can understand that; since typing in Python is an oxymoron.

src/gt4py/next/program_processors/runners/dace/utils.py
```diff
@@ -71,52 +71,68 @@ def is_connectivity_identifier(
...
+    dim: gtx_common.Dimension,
+    sym: Literal["size", "stride"],
+    offset_provider_type: Mapping[str, gtx_common.NeighborConnectivityType] | None,
+) -> dace.symbol:
+    if (m := CONNECTIVITY_INDENTIFIER_RE.match(field_name)) is None:
+        name = f"__{field_name}_{dim.value}_{sym}"
+    else:  # a connectivity field
+        assert offset_provider_type is not None
+        offset = m[1]
+        assert offset in offset_provider_type
```
philip-paul-mueller:
```suggestion
        assert m[1] in offset_provider_type
        offset = m[1]
```
Fully optional.
My only point is that now the asserts are consecutive.

PR #2379, opened 2025-11-07
perf[next]: Refactor of Program call arguments canonicalization

egparedes:
Refactor argument canonicalization utilities to leverage builtin cpython argument assignment logic.

Additionally, some properties have been move around in `Program`  to enhance readability.

### DIFF
--- a/src/gt4py/next/utils.py
+++ b/src/gt4py/next/utils.py
@@ -312,10 +312,6 @@ def make_args_canonicalizer(
     var_key_arg: str | None = None
 
     for key, param in signature.parameters.items():
-        # Remove annotations to avoid issues in case `str(annotation)`
-        # produces incorrect python expressions
-        params.append(param.replace(annotation=param.empty))
-
         match param.kind:
             case inspect.Parameter.POSITIONAL_ONLY | inspect.Parameter.POSITIONAL_OR_KEYWORD:
                 pos_args.append(f"{key}")
@@ -327,6 +323,10 @@ def make_args_canonicalizer(
             case inspect.Parameter.VAR_KEYWORD:
                 var_key_arg = key
                 key_args.append(f"**{key}")
+
+        # Remove annotations to avoid issues in case `str(annotation)`
+        # produces incorrect python expressions
+        params.append(param.replace(annotation=param.empty))
 
     canonicalizer_signature = inspect.Signature(parameters=params)
 
@@ -341,7 +341,7 @@ def make_args_canonicalizer(
         pos_args_tuple_expr = f"({str.join(', ', pos_args)})"
 
     if len(key_args) == 1 and var_key_arg:
-        # If there is only an '**kwargs' parameter, we can just return it directly
+        # If there is only a '**kwargs' parameter, we can just return it directly
         key_args_dict_expr = var_key_arg
     else:
         # In the regular case, assemble the output dictionary with all keyword arguments
@@ -384,7 +384,7 @@ def canonicalize_call_args(
         A tuple of positional arguments and a dictionary with keyword arguments.
 
     Note:
-        This is a convenience wrapper around `make_args_canonicalizer`.
+        This is a convenience wrapper around `make_args_canonicalizer_for_function`.
     """
 
     return make_args_canonicalizer_for_function(func)(*args, **kwargs)


src/gt4py/next/ffront/stages.py
```diff
@@ -59,6 +60,45 @@ class ProgramDefinition:
     grid_type: Optional[common.GridType] = None
     debug: bool = False
 
+    @functools.cached_property
```
egparedes: Is this useful? It's currently not used anywhere...

egparedes: No, deleting it.

src/gt4py/next/utils.py
```diff
@@ -285,213 +282,100 @@ def equalize_tuple_structure(
...
-    sort_kwargs: bool = False,
-) -> CustomCallArgsCanonicalizerFactory:
+    name: str | None = None,
+) -> Callable[..., tuple[tuple, dict[str, Any]]]:
     """
-    Create a factory for functions that canonicalize call arguments for a given signature.
+    Create a call arguments canonicalizer function from a given signature.
 
-    For a full description of what canonicalization means, see `make_args_canonicalizer`.
+    The canonicalization means that the returned arguments are as all positional
```
havogt:
```suggestion
    The canonicalization means that the returned arguments are as if all positional
```
is that what you meant?

src/gt4py/next/utils.py
```diff
@@ -285,213 +282,100 @@ def equalize_tuple_structure(
...
+                var_pos_arg = key
+                pos_args.append(f"*{key}")
+            case inspect.Parameter.KEYWORD_ONLY:
+                key_args.append(f"{key!r}: {key}")
+            case inspect.Parameter.VAR_KEYWORD:
+                var_key_arg = key
+                key_args.append(f"**{key}")
+
+    canonicalizer_signature = inspect.Signature(
+        parameters=params, return_annotation="tuple[tuple, dict[str, Any]]"
```
havogt: What's the reason to annotate the return type? For looking at the `str`? It's not useful for tools, right?

src/gt4py/next/utils.py
```diff
@@ -285,213 +282,100 @@ def equalize_tuple_structure(
...
-                else "# No args to unpack"
-            )
+    params = []
+    pos_args = []
+    key_args = []
+    var_pos_arg: str | None = None
+    var_key_arg: str | None = None
+
+    for key, param in signature.parameters.items():
+        params.append(param.replace(annotation=param.empty))  # remove annotations to avoid issues
```
havogt: What are the issues with annotations?

src/gt4py/next/utils.py
```diff
@@ -285,213 +282,100 @@ def equalize_tuple_structure(
...
-                    else f"a{next(passed_args_indices)}"
-                    for key in pos_name_to_index
-                )
-            canonical_args_expr = f"{str.join(', ', canonical_args_comprehension)},"
+    # Optimization for 0 or 1 positional arguments
+    if (pos_args_len := len(pos_args)) == 0:
+        pos_args_tuple = "()"
+    elif pos_args_len == 1 and var_pos_arg:
+        pos_args_tuple = var_pos_arg
+    else:
```
havogt: what's the optimization here?
avoiding the tuple in the 1-case?

src/gt4py/next/utils.py
```diff
@@ -285,213 +282,100 @@ def equalize_tuple_structure(
...
-        )
-        canonicalizer_src = f"""
-from __future__ import annotations
+    # Optimization for 0 or 1 keyword-only arguments
+    if (key_args_len := len(key_args)) == 0:
+        key_args_dict = "{}"
+    elif key_args_len == 1 and var_key_arg:
+        key_args_dict = var_key_arg
+    else:
+        key_args_dict = f"{{ {str.join(', ', key_args)} }}"
```
havogt: The 2 sections about optimizations are the most difficult part to understand. Are they worth it?

src/gt4py/next/utils.py
```diff
@@ -285,213 +282,100 @@ def equalize_tuple_structure(
...
-                )
-            else:
-                canonical_args_comprehension = (
-                    f"kwargs[{key!r}]"
-                    if key in passed_kwargs_keys
-                    else f"a{next(passed_args_indices)}"
-                    for key in pos_name_to_index
-                )
-            canonical_args_expr = f"{str.join(', ', canonical_args_comprehension)},"
+    # Optimization for 0 or 1 positional arguments
```
havogt: I think we should add tests for all these special cases.

src/gt4py/next/utils.py
```diff
@@ -285,213 +282,100 @@ def equalize_tuple_structure(
...
+    for key, param in signature.parameters.items():
+        params.append(param.replace(annotation=param.empty))  # remove annotations to avoid issues
+        match param.kind:
+            case inspect.Parameter.POSITIONAL_ONLY | inspect.Parameter.POSITIONAL_OR_KEYWORD:
+                pos_args.append(f"{key}")
+            case inspect.Parameter.VAR_POSITIONAL:
+                var_pos_arg = key
+                pos_args.append(f"*{key}")
+            case inspect.Parameter.KEYWORD_ONLY:
+                key_args.append(f"{key!r}: {key}")
```
havogt: Why `repr` in just this place?

havogt (changes requested):
Mainly I don't understand the optimizations. I guess we should also add tests for them.

src/gt4py/next/utils.py
egparedes: Added a comment to explain it.

src/gt4py/next/utils.py
egparedes: Done

src/gt4py/next/utils.py
egparedes: I've simplified the code, so it should be now easier to understand.

src/gt4py/next/utils.py
egparedes: Yes, avoiding building a new tuple or dict, since we can just return what we got.

src/gt4py/next/utils.py
egparedes: Not really useful, I agree. Removed

src/gt4py/next/utils.py
egparedes: Added a comment to explain it.

havogt (approved):
lgtm. I am still interested in the answers to the 2 questions from the previous round.

src/gt4py/next/utils.py
havogt: :roll_eyes: that's smart...

src/gt4py/next/utils.py
egparedes: The whole paragraph has been rephrased.

src/gt4py/next/utils.py
egparedes: Maybe too smart, I've changed it back to just using quotes.

PR #2380, opened 2025-11-07
fix[next][dace]: Generate temporary array ids in determistic way

edopao:


edopao:
> LGTM.
> 
> I have a question sometimes you use `add_temporary()` arrays, should we also modify them? Since they are only unique within an SDFG. Thus we might have many `__tmp1` that then get strangely mangled upon inlining.

I think I always use `add_temporary()` for temporary arrays created by the lowering. The _temporary_ fields here are of different kind, they are program symbols. However, they are not necessarily all temporaries, to be exact. That is why in #2382 I change the name prefix `gtir_tmp_` to `gtir_var_` (I could have done in this PR, actually): these symbols are currently mostly the temporaries generated by SSA pass, but they could also appear in function arguments, in which case they would not be lowered as transient arrays - they would become global arrays in nested SDFGs or tasklet connectors.

As per using a single unique ID generator for transient arrays during the lowering, I would currently not change it. It requires using a single visitor for all nested SDFGs (or a global static UID generator). The lowering needs to undergo some code refactory to make it better maintainable, so we can do it then.

PR #2382, opened 2025-11-09
fix[next][dace]: Avoid name conflict between tasklet connector and SDFG data

edopao:
SDFG validation checks for name conflicts between tasklet connector and data entities in the SDFG. This check results in failures like:
`dace.sdfg.validation.InvalidSDFGNodeError: Connector name '__inp' is already used as a symbol, constant, or array name (at state stmt_0, node tlet_3_copy)`

This PR tries to avoid such name conflicts by mangling the names of the tasklet connectors.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
@@ -465,7 +465,7 @@ class LambdaToDataflow(eve.NodeVisitor):
         tasklet_node, connector_mapping = self.subgraph_builder.add_tasklet(
             name, self.sdfg, self.state, inputs, outputs, code, **kwargs
         )
-        if not inputs:
+        if len(inputs) == 0:
             # All nodes inside a map scope must have an in/out path that traverses
             # the entry and exit nodes. Therefore, a tasklet node with no arguments
             # still needs an (empty) input edge from map entry node.


src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -101,30 +101,76 @@ def add_map(
     def add_tasklet(
         self,
         name: str,
+        sdfg: dace.SDFG,
         state: dace.SDFGState,
-        inputs: Union[Set[str], Dict[str, dace.dtypes.typeclass]],
-        outputs: Union[Set[str], Dict[str, dace.dtypes.typeclass]],
+        inputs: set[str] | None,
```
philip-paul-mueller: I kind of understand this change, but was there a particular reason to switch to `None` to indicate "no input"?
What was wrong with using an empty set (beside the fact that there is no set literal in Python)?
As a user I would start to think that passing `None` or the empty set might have different effects.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -101,30 +101,76 @@ def add_map(
...
+        `gtir_to_sdfg_utils.get_tasklet_connector()`), in order to avoid name conflicts
+        with SDFG data. Otherwise, SDFG validation would detect such conflicts and fail.
+        """
         unique_name = self.unique_tasklet_name(name)
-        return state.add_tasklet(unique_name, inputs, outputs, code, **kwargs)
+
+        if inputs is None:
+            inputs = set()
+        else:
+            assert inputs.isdisjoint(outputs)
```
```suggestion
        inputs = inputs or set()
        assert inputs.isdisjoint(outputs)
```
I think that this is more pythonic, although I admit that you have to create a new input set even if you pass an empty one.
You could fix that with: `inputs = set() if inputs is None else inputs`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -101,30 +101,76 @@ def add_map(
...
+        inputs: Mapping[str, dace.Memlet] | None,
         code: str,
-        outputs: Dict[str, dace.Memlet],
+        outputs: Mapping[str, dace.Memlet],
+        language: dace.dtypes.Language = dace.dtypes.Language.Python,
         **kwargs: Any,
-    ) -> tuple[dace.nodes.Tasklet, dace.nodes.MapEntry, dace.nodes.MapExit]:
-        """Wrapper of `dace.SDFGState.add_mapped_tasklet` that assigns unique name."""
+    ) -> tuple[dace.nodes.Tasklet, dace.nodes.MapEntry, dace.nodes.MapExit, dict[str, str]]:
+        """Wrapper of `dace.SDFGState.add_mapped_tasklet` that modifies the connector names.
```
philip-paul-mueller: The part about the "unique name" got removed.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -101,30 +101,76 @@ def add_map(
...
+        tasklet_node = state.add_tasklet(
+            unique_name, inputs, outputs, new_code, language=language, **kwargs
+        )
+        return tasklet_node, connector_mapping
 
     def add_mapped_tasklet(
         self,
         name: str,
+        sdfg: dace.SDFG,
         state: dace.SDFGState,
```
philip-paul-mueller, line 143: Why not passing the context here.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -266,8 +266,9 @@ def add_instrumentation(sdfg: dace.SDFG, gpu: bool) -> None:
         None,
         dace.Memlet(f"{output}[0]"),
     )
-
-    # Check SDFG validity after applying the above changes.
+    # We normally wrap `add_tasklet()` in order to rename tasklet connectors and
```
philip-paul-mueller: I would reformulate that I think something like:
Normally we do not call `SDFGState.add_tasklet()` directly, instead we call the wrapper provided by `DataflowBuilder`, that ensures unique naming blablabla.
However, this method is not available here we have to call the underlying DaCe function directly.
We now run `validate()` to make sure that there was no naming conflict introduced.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: You are right, I will revert this change.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: I like your suggestion, no big overhead. However, not needed if I do not allow `None` in the first place.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: `DataflowBuilder` (which corresponds to the lowering of iterator view) does not use `SubgraphContext` - so far, but it can be revisited with some code refactoring. `SubgraphContext` was initially meant to be used only in the _fieldview_ part of the lowering.

philip-paul-mueller: That is fine, I just saw it "oh now we _could_ use or it would make sense to use the context."

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -451,8 +450,8 @@ def _add_map(
     def _add_tasklet(
         self,
         name: str,
-        inputs: set[str] | None,
-        outputs: set[str],
+        inputs: set[str] | dict[str, dace.dtypes.typeclass],
+        outputs: set[str] | dict[str, dace.dtypes.typeclass],
```
philip-paul-mueller:
```suggestion
        inputs: set[str] | Mapping[str, dace.dtypes.typeclass],
        outputs: set[str] | Mapping[str, dace.dtypes.typeclass],
```
Although I am unsure if we ever need the `dace.dtypes.typeclass` part.
Furthermore, I think the `ValueType` can be `None` (this is wrong in the typing of DaCe).

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -477,10 +476,10 @@ def _add_tasklet(
     def _add_mapped_tasklet(
         self,
         name: str,
-        map_ranges: Mapping[str, str | dace.subsets.Subset],
-        inputs: Mapping[str, dace.Memlet] | None,
+        map_ranges: dict[str, str | dace.subsets.Subset],
+        inputs: dict[str, dace.Memlet],
```
philip-paul-mueller:
```suggestion
        map_ranges: Mapping[str, str | dace.subsets.Subset],
        inputs: Mapping[str, dace.Memlet],
```

src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
```diff
@@ -477,10 +476,10 @@ def _add_tasklet(
...
         code: str,
-        outputs: Mapping[str, dace.Memlet],
+        outputs: dict[str, dace.Memlet],
```
philip-paul-mueller:
```suggestion
        outputs: Mapping[str, dace.Memlet],
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -103,33 +103,34 @@ def add_tasklet(
         name: str,
         sdfg: dace.SDFG,
         state: dace.SDFGState,
-        inputs: set[str] | None,
-        outputs: set[str],
+        inputs: set[str] | dict[str, dace.dtypes.typeclass],
+        outputs: set[str] | dict[str, dace.dtypes.typeclass],
```
```suggestion
        inputs: set[str] | Mapping[str, dace.dtypes.typeclass],
        outputs: set[str] | Mapping[str, dace.dtypes.typeclass],
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -103,33 +103,34 @@ def add_tasklet(
...
+        if isinstance(inputs, set):
+            inputs = {k: None for k in inputs}
+        if isinstance(outputs, set):
+            outputs = {k: None for k in outputs}
+        assert inputs.keys().isdisjoint(outputs.keys())
 
         connector_mapping = {
-            conn: gtir_to_sdfg_utils.get_tasklet_connector(conn) for conn in (inputs | outputs)
+            conn: gtir_to_sdfg_utils.get_tasklet_connector(conn)
+            for conn in (inputs.keys() | outputs.keys())
```
philip-paul-mueller, line 126:
```suggestion
            for conn in sorted(inputs.keys() | outputs.keys())
```
This might be paranoid.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_utils.py
```diff
@@ -45,6 +49,14 @@ def get_map_variable(dim: gtx_common.Dimension) -> str:
     return f"i_{dim.value}_gtx_{dim.kind}{suffix}"
 
 
+def get_tasklet_connector(name: str) -> str:
```
philip-paul-mueller: Feel free to ignore, but I am not super happy with `get_tasklet_connector()` especially the `get` as it has some "look it up somewhere" vibes.
What do you think about `make_tasklet_connector_for()` it transmits the idea that the connector is created anew.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: I see, python `dict` follows insertion order. I don't think it is needed here, but it is probably needed a few lines above.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_utils.py
edopao: I see your point, but I do not find an alternative function name which I really like.
I forgot to check that the current `name` does not already start with the prefix (I mean, we should not call this function more than once).

edopao: On second thought, I can go for `make_tasklet_connector_for()`

PR #2383, opened 2025-11-10
feat[dace][next]: Deterministic `gt_split_access_nodes()`

philip-paul-mueller:
Makes `gt_split_access_nodes()` deterministic by processing it in the order of `AccessNodes`.
However, this means that the name of the data must be predictable and stable.

--------------
### ToDo
- [x] Include in `gt_auto_optimization()`
- [x] Order in which the splits are performed make that deterministic as well.
- [x] Check with @edopao if the names are deterministic
- [x] Verify in ICON4Py CI: https://github.com/C2SM/icon4py/pull/937
- [x] Verify Performance

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
@@ -493,10 +493,6 @@ class LambdaToDataflow(eve.NodeVisitor):
             name, self.sdfg, self.state, map_ranges, inputs, code, outputs, **kwargs
         )
 
-    def unique_nsdfg_name(self, prefix: str) -> str:
-        """Utility function to generate a unique name for a nested SDFG, starting with the given prefix."""
-        return self.subgraph_builder.unique_nsdfg_name(self.sdfg, prefix)
-
     def _construct_local_view(self, field: MemletExpr | ValueExpr) -> ValueExpr:
         if isinstance(field, MemletExpr):
             desc = field.dc_node.desc(self.sdfg)
@@ -778,7 +774,7 @@ class LambdaToDataflow(eve.NodeVisitor):
                     deref_on_input_memlet=deref_on_input_memlet: self._visit_if_branch_arg(
                         if_sdfg,
                         if_branch_state,
-                        tsym.id,
+                        str(tsym.id),
                         targ,
                         deref_on_input_memlet,
                         if_sdfg_input_memlets,
@@ -869,7 +865,7 @@ class LambdaToDataflow(eve.NodeVisitor):
             else (condition_value.dc_dtype == dace.dtypes.bool_)
         )
 
-        nsdfg = dace.SDFG(self.unique_nsdfg_name(prefix="if_stmt"))
+        nsdfg = dace.SDFG(self.subgraph_builder.unique_nsdfg_name("if_stmt"))
         nsdfg.debuginfo = gtir_to_sdfg_utils.debug_info(node, default=self.sdfg.debuginfo)
 
         # create states inside the nested SDFG for the if-branches
@@ -1371,7 +1367,7 @@ class LambdaToDataflow(eve.NodeVisitor):
         assert desc.shape[local_dim_index] == offset_provider_type.max_neighbors
 
         # we lower the reduction map with WCR out memlet in a nested SDFG
-        nsdfg = dace.SDFG(name=self.unique_nsdfg_name("reduce_with_skip_values"))
+        nsdfg = dace.SDFG(self.subgraph_builder.unique_nsdfg_name("reduce_with_skip_values"))
         nsdfg.add_array(
             "values",
             (desc.shape[local_dim_index],),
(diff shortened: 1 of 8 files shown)


edopao:
I was thinking that the assert on unique labels could be part of this PR, but inside `auto_optimize`. What do you think?

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
```diff
@@ -48,12 +58,91 @@ def gt_split_access_nodes(
...
+        return apply_count
+
+    # The splitter transformation. Note that we set `assume_single_use_data` to `True`
+    #  because we do this test outside.
+    access_node_splitter = gtx_transformations.SplitAccessNode(
+        assume_single_use_data=True,
     )
 
+    # Since the transformation only applies to single use data, the order in which the
+    #  states are processed is irrelevant. Furthermore, the fragments generated through
```
edopao: "Furthermore, the fragments generated..." I don't see how this is relevant. Are you considering the case in which `gt_split_access_nodes()` is called multiple times?

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
```diff
@@ -257,6 +257,12 @@ def split_node(
     if already_reconfigured_nodes is None:
         already_reconfigured_nodes = set()
 
+    # Sort the split descriptions such that they are processed in a deterministic way.
+    # NOTE: Turning them into a string is the best solution is probably the only way
+    #   to achieve some stability. The only downside is that the order now depends
+    #   on the specialization level that is used, i.e. to we have numbers or symbols.
+    split_description = sorted(split_description, key=lambda split: str(split))
```
edopao, line 264: `split_description` is a Sequence, so it is already sorted. As alternative, we could ensure the deterministic order in the place where the sequence is created.

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
```diff
@@ -334,6 +340,13 @@ def split_edge(
     #  detail that does not limit the applicability of this function.
     # TODO(phimuell): Implements some check that nothing is lost.
 
+    # Bring the split description in a deterministic order.
+    # NOTE: See note in `split_node()` why the sorting is done in this way.
+    # NOTE: The main benefit of bringing `split_description` into a deterministic
+    #   order is that the output of this function is deterministic as well. I am
+    #   not sure if there is any benefit beside that.
+    split_description = sorted(split_description, key=lambda split: str(split))
```
edopao, line 348: Same

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
```diff
@@ -366,6 +379,8 @@ def split_edge(
                 new_fully_splitted_subsets.append(consumer)
         fully_splitted_subsets = new_fully_splitted_subsets
 
+    # Allocate the return `dict` the order is important, first the reordered split
+    #  description followed by the `None` key.
```
edopao: Can you also explain why it is important?

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
philip-paul-mueller: This is the justification why scanning the state once is enough.
Because you could argue that you should examine the result of the split (which generates more AccessNodes) should be examined as well.
And we would thus need an interative procedure.
However, I am not fully sure now if this might be needed.
I will update the comment to make this more clear.

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
philip-paul-mueller: I do not fully understand your comment, are you saying that sorting is not needed, because it is a sequence?
According to the [Python glossary](https://docs.python.org/3/glossary.html#term-sequence) a `Sequence` extends an `Iterable` by an efficient `__getitem__()` function, which essentially give sense to the concept of "first", "second", etc. element.
This means that they have _some_ (fixed but ultimately random) order but to be stable you need a _specific_ order.

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
philip-paul-mueller: see my other reply.

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
philip-paul-mueller: Good point, actually I do not remember why it is that important and it is probably not key to anything.
The reason for preallocating it was probably to avoid a check if it is already there.
A nice side effect is that it reflects processing order since `dict` remembers insertion order.
Putting `None` at the end is more like a convention, the beginning would also have made sense.

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
edopao: On the second thought, I agree with sorting this sequence here. My motivation is that this function should _in theory_ implement some logic for the order, but we rely on alphabetical order for now.

philip-paul-mueller: I agree this is what the Note should express, that this is the only way I found that gives you something stable.
Because, you can have comparable things such as `symPyVariable1` and `symPyVariable2`.

PR #2384, opened 2025-11-10
fix[next][dace] Set properly strides of `data` in CPU

iomaganaris:
After https://github.com/GridTools/gt4py/pull/2355 we have to update the strides of AccessNode `data` for memlets inside `NestedSDFG`s properly.
`gt_change_strides` used to run only for the GPU backend. Now it's necessary for CPU as well.
The final strides must be set after all the transformations are done because some AccessNodes might be removed and the strides of the corresponding AccessNodes inside `NestedSDFG`s need to be updated

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
@@ -261,4 +261,6 @@ class RemoveAccessNodeCopies(dace_transformation.SingleStateTransformation):
 
         # Update strides after changing the data of access nodes
         for modified_node in [second_node, third_node]:
-                gtx_transformations_strides.gt_propagate_strides_from_access_node(sdfg, graph, modified_node)
+            gtx_transformations_strides.gt_propagate_strides_from_access_node(
+                sdfg, graph, modified_node
+            )


src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
```diff
@@ -257,3 +258,6 @@ def apply(
         # Replace data of second and third node with data of first and fourth node
         second_node.data = first_node.data
         third_node.data = first_node.data
+
+        # Update strides after changing the data of access nodes
+        gtx_transformations_strides.gt_propagate_strides_of(sdfg, first_node.data)
```
philip-paul-mueller:
```suggestion
        for modified_node in [second_node, third_node]:
                gtx_transformations_strides.gt_propagate_strides_from_access_node(sdfg, graph, modified_node)
```
This should be more efficient since it propagates the strides where modified them, i.e. everything that is attached to `first_node` and `second_node` has not changed.

src/gt4py/next/program_processors/runners/dace/transformations/strides.py
```diff
@@ -30,7 +30,7 @@
 
 def gt_change_strides(
     sdfg: dace.SDFG,
-    gpu: bool,
+    gpu: bool = False,
```
philip-paul-mueller:
```suggestion
    gpu: bool,
```
In this case I think the user has to think about if he is on GPU or not.

src/gt4py/next/program_processors/runners/dace/transformations/remove_access_node_copies.py
iomaganaris: Good point 👍

PR #2386, opened 2025-11-11
feat[next-dace]: Use SDFG library node for lowering of broadcast

edopao:
TODO:
- [x] Run ICON4Py CI, see [ICON4Py PR#1240](https://github.com/C2SM/icon4py/pull/1240)
- [x] Run Blueline (there are some degradation but in general it works)
- [ ] Run MuPhys (just to be sure)


src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
```diff
@@ -73,7 +73,7 @@ def _parse_fieldop_arg(
     ctx: gtir_to_sdfg.SubgraphContext,
     sdfg_builder: gtir_to_sdfg.SDFGBuilder,
     domain: gtir_domain.FieldopDomain,
-) -> MaybeNestedInTuple[gtir_dataflow.IteratorExpr | gtir_dataflow.MemletExpr]:
+) -> gtir_dataflow.IteratorExpr | gtir_dataflow.MemletExpr:
```
philip-paul-mueller, line 75: Just out of curiosity why this change.

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
```diff
@@ -0,0 +1,73 @@
...
+from dace import library as dace_library, nodes as dace_nodes
+from dace.transformation import transformation as dace_transform
+
+
+_INPUT_NAME: Final[str] = "_input"
+_OUTPUT_NAME: Final[str] = "_output"
+
+
+@dace_library.expansion
+class ExpandPure(dace_transform.ExpandTransformation):
```
philip-paul-mueller: I would incorporate the name of what you are expanding, something like `FillExpandPure`.

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
```diff
@@ -0,0 +1,73 @@
...
+
+        state.add_mapped_tasklet(
+            f"{node.label}_tasklet", map_rng, inputs, code, outputs, external_edges=True
+        )
+
+        return sdfg
+
+
+@dace_library.node
+class Fill(dace_nodes.LibraryNode):
```
philip-paul-mueller: I would add some more semantic, i.e. an input connector, that collects the value that should be broadcasted and an output connector for the output.

I am also wondering if it would make sense to have two different library nodes.
One where the value that is broadcast is a literal, like `0.0` and one, which is probably the current one, where the value is read from another data descriptor (might be hard to integrate into the lowering).

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
```diff
@@ -0,0 +1,73 @@
...
+    @staticmethod
+    def expansion(node: Fill, parent_state: dace.SDFGState, parent_sdfg: dace.SDFG) -> dace.SDFG:
+        sdfg = dace.SDFG(f"{node.label}_sdfg")
+
+        assert len(parent_state.out_edges(node)) == 1
+        outedge = parent_state.out_edges(node)[0]
+        out_desc = parent_sdfg.arrays[outedge.data.data]
+        inner_out_desc = out_desc.clone()
+        inner_out_desc.transient = False
+        out = sdfg.add_datadesc(_OUTPUT_NAME, inner_out_desc)
```
philip-paul-mueller: Because `add_datadesc()` defaults `find_new_name` to `False` this means that only one fill node can be used.

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
```diff
@@ -0,0 +1,73 @@
...
+
+from typing import Any, Final
+
+import dace
+from dace import library as dace_library, nodes as dace_nodes
+from dace.transformation import transformation as dace_transform
+
+
+_INPUT_NAME: Final[str] = "_input"
+_OUTPUT_NAME: Final[str] = "_output"
```
philip-paul-mueller:
```suggestion
_INPUT_NAME: Final[str] = "gtx_fill_input_temporary_{old_name}"
_OUTPUT_NAME: Final[str] = "gtx_fill_output_temporary_{old_name}"
```

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
```diff
@@ -0,0 +1,73 @@
...
+        assert len(parent_state.in_edges(node)) == 1
+        inedge = parent_state.in_edges(node)[0]
+        inp_desc = parent_sdfg.arrays[inedge.data.data]
+        inner_inp_desc = inp_desc.clone()
+        inner_inp_desc.transient = False
+        inp = sdfg.add_datadesc(_INPUT_NAME, inner_inp_desc)
+        inedge._dst_conn = _INPUT_NAME
+        node.add_in_connector(_INPUT_NAME)
+        inputs = {"_in": dace.Memlet(data=inp, subset="0")}
+        code = "_out = _in"
```
philip-paul-mueller: What are you doing here?

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
```diff
@@ -0,0 +1,73 @@
...
+        sdfg = dace.SDFG(f"{node.label}_sdfg")
+
+        assert len(parent_state.out_edges(node)) == 1
+        outedge = parent_state.out_edges(node)[0]
+        out_desc = parent_sdfg.arrays[outedge.data.data]
+        inner_out_desc = out_desc.clone()
+        inner_out_desc.transient = False
+        out = sdfg.add_datadesc(_OUTPUT_NAME, inner_out_desc)
+        outedge._src_conn = _OUTPUT_NAME
+        node.add_out_connector(_OUTPUT_NAME)
```
philip-paul-mueller: Why are you modifying the node?

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
```diff
@@ -0,0 +1,73 @@
...
+
+        state = sdfg.add_state(f"{node.label}_state")
```
philip-paul-mueller: Why are you adding a state?

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
```diff
@@ -0,0 +1,73 @@
...
+        map_params = [f"__i{i}" for i in range(len(out_desc.shape))]
+        map_rng = {i: f"0:{s}" for i, s in zip(map_params, out_desc.shape)}
+        out_mem = dace.Memlet(expr=f"{out}[{','.join(map_params)}]")
+        outputs = {"_out": out_mem}
```
philip-paul-mueller: You define the Map range based on the whole target array.
In my opinion you should inspect the destination subset of the outgoing edge of the fill node and use that.

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@dace_library.expansion
+class ExpandPure(dace_transform.ExpandTransformation):
+    """Implements pure expansion of the Fill library node."""
+
+    environments: Final[list[Any]] = []
+
+    @staticmethod
+    def expansion(node: Fill, parent_state: dace.SDFGState, parent_sdfg: dace.SDFG) -> dace.SDFG:
```
philip-paul-mueller: It seems that the [interface has changed](https://github.com/spcl/dace/pull/2077).

philip-paul-mueller (commented):
There are some refinements needed.

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
havogt: Any particular reason to introduce `Fill` instead of broadcast? Couldn't there be similar cases where you don't want to materialize a field that is broadcasted to more dimensions?

philip-paul-mueller: I do not fully understand this comment.
Regarding the not materialization, this PR is not complete, because some transformation, to handle it are missing; Which will be in some other PR.
Currently they expand to a Map and then the optimizer has to guess what to do and currently the Maps are handled in a crude way.
The main reason to have this Node is that the optimizer (or some transformation, I think we need two) can handle it in a more appropriate way.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
edopao: This is the correct return type. This function can never return a tuple. You are right that this change is unrelated.

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
edopao: I agree.

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
edopao: Thanks for pointing me to this, I was following the `reduce` library node which is not updated.

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
edopao: Because I was adding a nested SDFG, but we concluded that it is not necessary. We have agreed that I should inline the mapped tasklet in the current state, instead.

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
edopao: Good point.

src/gt4py/next/program_processors/runners/dace/sdfg_library_nodes.py
edopao: I think I get Hannes' comment. After changing the library node to a mapped tasklet inlined in the same state (as suggested by Philip) we can achieve the goal suggested by Hannes. I have spotted some places in the lowering of `concat_where` where I could use the new broadcast node. For now I have just added a TODO.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
```diff
@@ -285,6 +291,86 @@ def translate_as_fieldop(
...
+
+    # The memory layout of the output field follows the field operator compute domain.
+    field_dims, field_origin, field_shape = gtir_domain.get_field_layout(field_domain)
+    assert field_dims == node.type.dims
+    field_name, field_desc = sdfg_builder.add_temp_array(ctx.sdfg, field_shape, field_dtype)
+    field_node = ctx.state.add_access(field_name)
+
+    # Retrieve the scalar argument, which could be either a literal value or the
+    # result of a scalar expression.
+    if isinstance(scalar_arg, gtir.Literal) and np.isfinite(
```
edopao: The filter `np.isfinite()` is currently just a workaround to a dace issue related to parsing of `inf` value in `test_execution::test_infinity`

edopao:
No plan for now to integrate this feature.

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+import copy
+from typing import Any, Final, Iterable, Optional, Sequence
+
+import dace
+from dace import (
+    data as dace_data,
+    library as dace_library,
+    nodes as dace_nodes,
+    properties as dace_properties,
+    subsets as dace_sbs,
```
edopao: In the transformation module, we use the `dace_sbs` alias, in the lowering module we use `dace_subsets`. It's OK to use `dace_sbs` in this module, but let's try to keep it consistent.

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+    """Implements write of a scalar value over an array subset.
+
+    Same as XLA.
+    broadcast_in_dim[i] describes where dimension `i` of the `value_to_broadcast`
+    goes. In case of a scalar it is empty.
+    Furthermore the following has to hold:
+    ```python
+    for i in range(len(broadcast_in_dim):
+        assert output.shape[broadcast_in_dim[i]] == value_to_broadcast.shape[i]
+    ```
```
edopao:
```suggestion
    ```
    In other words, the result array shape has the same size as the broadcast domain.
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+
+    Args:
+        broadcast_in_dim: How to broadcast.
```
edopao:
```suggestion
        broadcast_in_dim: How to broadcast, see the class documentation.
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+        params: The parameters that should be used for the expansion. If given one
```
edopao:
```suggestion
        params: The parameters that should be used for the expansion. If given, one
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+            entry for each dimension of the destination is needed.
+
+    Todo:
+        - While for the output it is probably okay to always require an adjacent
+            AccessNode for the input it might be possible to be on the other side
```
edopao:
```suggestion
            AccessNode, the input nodes might be outside a map scope.
```
However, I don't understand how this could happen.

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+    default_implementation: Final[str | None] = "pure"
+
+    brodcast_in_dims = dace_properties.ListProperty(element_type=int)
+    params = dace_properties.ListProperty(element_type=str, allow_none=True)
+
+    def __init__(
+        self,
+        name: str,
+        broadcast_in_dims: Sequence[int],
+        params: Optional[Iterable[gtx_common.Dimension | str]],
```
edopao:
```suggestion
        params: Iterable[gtx_common.Dimension | str] | None,
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+        debuginfo: dace.dtypes.DebugInfo | None = None,
+    ):
+        # TODO(philip, edopao): I would propose to drop `value` then. This makes it
+        #   simpler to handle in the transformations.
```
edopao: I agree, we can remove the todo comment.

```suggestion
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+            self.params = [
+                gtx_dace_lowering.get_map_variable(param)
+                if isinstance(param, gtx_common.Dimension)
+                else param
+                for param in params
+            ]
+
+    def validate(self, sdfg: dace.SDFG, state: dace.SDFGState) -> None:
+        if len(self.brodcast_in_dims) != len(set(self.brodcast_in_dims)):
+            raise ValueError("`Can not broadcast to multiple dimensions at the same time.")
```
edopao:
```suggestion
            raise ValueError("`Broadcast dimensions must be unique")
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+
+        # TODO(phimuell): Handle empty Memlets in the input.
+        if state.in_degree(self) != 1 and next(iter(state.in_edges(self))).dst_conn == _INPUT_NAME:
```
edopao:
```suggestion
        if state.in_degree(self) != 1 or next(iter(state.in_edges(self))).dst_conn != _INPUT_NAME:
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+            raise ValueError("GT4Py Broadcast node needs exactly one input.")
+        if (
+            state.out_degree(self) != 1
+            and next(iter(state.out_edges(self))).src_conn == _OUTPUT_NAME
```
edopao:
```suggestion
            state.out_degree(self) != 1
            or next(iter(state.out_edges(self))).src_conn != _OUTPUT_NAME
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+            # In fact it would also be possible to broadcast into a Scalar, but this
+            #  does not make much sense.
+            raise ValueError(
+                f"Can only broadcast into an array, but target was `{type(bcast_result_desc).__name__}`."
+            )
+
+        if (self.params is not None) and (len(self.params) != len(bcast_result_desc.shape)):
+            raise ValueError(
+                f"Expected that {len(bcast_result_desc.shape)} parameters are"
+                f" needed but {len(self.params)} were specified."
```
edopao: `The number of dimensions of the result shape {len(bcast_result_desc.shape)} does not match the nmuber of broadcast parameters {len(self.params)}.`

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+            )
+
+        if isinstance(bcast_value_desc, dace_data.Scalar):
+            if len(self.brodcast_in_dims) != 0:
+                raise ValueError("For a scalar `broadcast_in_dims` must be empty.")
+        else:
+            if len(self.brodcast_in_dims) != len(bcast_value_desc.shape):
```
edopao, line 143:
```suggestion
        elif len(self.brodcast_in_dims) != len(bcast_value_desc.shape):
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+                )
+            if len(bcast_result_desc.shape) < len(bcast_value_desc.shape):
+                raise ValueError(
+                    f"The value to broadcast has more dimensions ({len(bcast_value_desc.shape)})"
+                    f" than the result ({len(bcast_result_desc.shape)})."
+                )
+
+            for src_dim, bcast_dst_dim in enumerate(self.brodcast_in_dims):
+                if bcast_dst_dim < 0:
+                    raise ValueError("Negative broadcast")
```
edopao:
```suggestion
                    raise ValueError("Negative dimension index is invalid.")
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+    that this function does not generate nested SDFG and instead performs the
+    expansion in place.
+    """
+
+    input_edge: dace_graph.MultiConnectorEdge[dace.Memlet] = next(iter(state.in_edges(bcast_node)))
+    output_edge: dace_graph.MultiConnectorEdge[dace.Memlet] = next(
+        iter(state.out_edges(bcast_node))
+    )
+
+    # TODO(phimuell): Add warning.
```
edopao: I don't think we should add a warning. Either allow it or not.

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+    map_params: list[str] = (
+        [f"__bcast{dst_dim}" for dst_dim in range(len(output_edge.data.subset))]
+        if bcast_node.params is None
+        else list(bcast_node.params)
+    )
+
+    output_subset: list[str] = map_params.copy()
+    map_ranges: dict[str, dace_sbs.Range] = {
+        map_param: dace_sbs.Range([sbs])
+        for map_param, sbs in zip(map_params, output_edge.data.subset)
```
edopao:
```suggestion
        for map_param, sbs in zip(map_params, output_edge.data.subset, strict=True)
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    input_subset: list[str]
+    if len(bcast_node.brodcast_in_dims) == 0:
+        input_subset = ["0"]
+    else:
+        bcast_value_offset = input_edge.data.subset.min_element()
+        input_subset = [
+            f"{map_params[dst_dim]} + ({offset})"
+            for dst_dim, offset in zip(bcast_node.brodcast_in_dims, bcast_value_offset)
```
edopao:
```suggestion
            for dst_dim, offset in zip(bcast_node.brodcast_in_dims, bcast_value_offset, strict=True)
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+        dace.Memlet.from_array(input_edge.data.data, sdfg.arrays[input_edge.data.data]),
+    )
+    state.add_edge(
+        me,
+        f"OUT_{input_edge.data.data}",
+        bcast_tlet,
+        "__in",
+        dace.Memlet(data=input_edge.data.data, subset=", ".join(input_subset)),
+    )
+    me.add_scope_connectors(input_edge.data.data)
```
edopao, line 231:
```suggestion
    state.add_memlet_path(
        input_edge.src,
        me,
        bcast_tlet,
        src_conn=input_edge.src_conn,
        dst_conn="__in",
        dace.Memlet(data=input_edge.data.data, subset=", ".join(input_subset)),
    )
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+        dace.Memlet(data=output_edge.data.data, subset=", ".join(output_subset)),
+    )
+    state.add_edge(
+        mx,
+        f"OUT_{output_edge.data.data}",
+        output_edge.dst,
+        output_edge.dst_conn,
+        dace.Memlet(data=output_edge.data.data, subset=copy.deepcopy(output_edge.data.subset)),
+    )
+    mx.add_scope_connectors(output_edge.data.data)
```
edopao:
```suggestion
    state.add_memlet_path(
        bcast_tlet,
        mx,
        output_edge.dst,
        src_conn="__out",
        dst_conn=output_edge.dst_conn,
        dace.Memlet(data=output_edge.data.data, subset=copy.deepcopy(output_edge.data.subset)),
    )
```

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
```diff
@@ -0,0 +1,302 @@
...
+class BroadcastExpandInlined(dace_transform.ExpandTransformation):
+    """Implements pure expansion of the Broadcast library node."""
+
+    environments: Final[list[Any]] = []
+
+    @staticmethod
+    def expansion(node: Broadcast, state: dace.SDFGState, sdfg: dace.SDFG) -> dace.SDFG:
+        # TODO:
+        #   - Modify the edges on the outside.
+        #   - Handle the missing symbols.
```
edopao: The missing symbols are automatically mapped to symbols with the same name, aren't they?

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
```diff
@@ -0,0 +1,422 @@
...
+from dace.transformation.passes import analysis as dace_analysis
+
+from gt4py.next.program_processors.runners.dace import (
+    library_nodes as gtx_lib_nodes,
+    transformations as gtx_xtrans,
+)
+
+
+@dace_properties.make_properties
+class ScalarBrodcastInliner(dace_transformation.SingleStateTransformation):
```
edopao:
```suggestion
class ScalarBroadcastInliner(dace_transformation.SingleStateTransformation):
```

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -27,14 +27,18 @@
...
 
 from gt4py import eve
 from gt4py.eve.extended_typing import MaybeNestedInTuple, NestedTuple
 from gt4py.next import common as gtx_common, utils as gtx_utils
 from gt4py.next.iterator import ir as gtir
 from gt4py.next.iterator.ir_utils import common_pattern_matcher as cpm, ir_makers as im
 from gt4py.next.iterator.transforms import symbol_ref_utils
-from gt4py.next.program_processors.runners.dace import sdfg_args as gtx_dace_args
+from gt4py.next.program_processors.runners.dace import (
+    library_nodes as gtir_library_nodes,
```
edopao:
```suggestion
    library_nodes as gtx_dace_library_nodes,
```

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -1471,27 +1342,73 @@ def _visit_reduce(self, node: gtir.FunCall) -> ValueExpr:
...
+            reduce_node = gtir_library_nodes.ReduceWithSkipValues(
+                name,
+                reduce_wcr,
+                identity=reduce_identity.value,
+                init=reduce_init.value,
+                debuginfo=gtir_to_sdfg_utils.debug_info(node),
+            )
+            reduce_node.input_conn = inp_conn
+            reduce_node.output_conn = outp_conn
+            reduce_node.mask_conn = mask_conn
```
edopao:
```suggestion
            reduce_node = gtir_library_nodes.ReduceWithSkipValues(
                name,
                reduce_wcr,
                identity=reduce_identity.value,
                init=reduce_init.value,
                debuginfo=gtir_to_sdfg_utils.debug_info(node),
                input_conn=inp_conn,
                output_conn=outp_conn,
                mask_conn=mask_conn,
            )
```

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
```diff
@@ -292,6 +290,162 @@ def translate_as_fieldop(
...
+
+    fun_node = node.fun
+    assert len(fun_node.args) == 2
+    fieldop_expr, fieldop_domain_expr = fun_node.args
+    assert cpm.is_ref_to(fieldop_expr, "deref")
+
+    # TODO:
+    #   - Include the dimensions of the output, such that we can generate the Map correctly
+    #       this is needed for expansion of the broadcast node such that we can order them correctly.
+    #       This information is stored in `field_dims`.
```
edopao, line 319: Is this TODO still valid?

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
```diff
@@ -292,6 +290,162 @@ def translate_as_fieldop(
...
+    bcast_result_subset = dace_subsets.Range.from_array(field_desc)
+    bcast_value_subset: dace_subsets.Subset | None = None
+
+    # Which dimensions in `bcast_value` corresponds to the ones in `cast_result`.
+    broadcast_in_dims: list[int]
+
+    if isinstance(bcast_arg, gtir.Literal):
+        assert isinstance(bcast_arg.type, ts.ScalarType)
+        bcast_value_tlet, connector_mapping = sdfg_builder.add_tasklet(
+            sdfg_builder.unique_tasklet_name(bcast_node_name),
```
edopao: The name is made unique inside the `add_tasklet()` wrapper:
```suggestion
            bcast_node_name,
```

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
```diff
@@ -292,6 +290,162 @@ def translate_as_fieldop(
...
+        bcast_value_name, bcast_value_desc = sdfg_builder.add_temp_scalar(
+            ctx.sdfg, gtx_dace_args.as_dace_type(bcast_arg.type)
+        )
+        bcast_value = ctx.state.add_access(bcast_value_name)
+        ctx.state.add_edge(
+            bcast_value_tlet,
+            connector_mapping["__out"],
+            bcast_value,
+            None,
+            dace.Memlet.from_array(bcast_value_name, bcast_value_desc),
```
edopao:
```suggestion
            dace.Memlet(data=bcast_value_name, subset="0"),
```

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
```diff
@@ -292,6 +290,162 @@ def translate_as_fieldop(
...
+            assert isinstance(arg.gt_dtype, ts.ListType)
+            raise NotImplementedError("Broadcast of lists is not supported.")
+
+    else:
+        # TODO: What is the exact difference between this case and the one above?
+
+        bcast_value = arg.field
+        bcast_value_desc = bcast_value.desc(ctx.sdfg)
+        if isinstance(bcast_value_desc, dace_data.Scalar):
+            broadcast_in_dims = []
```
edopao, line 387: Have you ever hit this case: `IteratorExpr` with `bcast_value_desc` scalar? I think you can remove this if branch.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
```diff
@@ -292,6 +290,162 @@ def translate_as_fieldop(
...
+        broadcast_in_dims=broadcast_in_dims,
+        params=field_dims,
+        debuginfo=gtir_to_sdfg_utils.debug_info(node),
+    )
+    ctx.state.add_node(bcast_node)
+
+    # If not specified differently use the whole `bcast_value` content.
+    if bcast_value_subset is None:
+        bcast_value_subset = dace_subsets.Range.from_array(bcast_value.desc(ctx.sdfg))
+
```
edopao, line 430:
```suggestion
    # If not specified differently use the whole `bcast_value` content.
    if bcast_value_subset is None:
        assert isinstance(bcast_value_desc, dace_data.Scalar)
        bcast_value_subset = dace_subsets.Range.from_string("0")

```

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -247,6 +252,15 @@ def gt_auto_optimize(
...
 
+        # We need new transformations in order to deal with GTIR library nodes.
+        # For now, we simply expand these nodes before starting optimizing.
+        # TODO: Remove once transformations are ready.
+        for node, state in sdfg.all_nodes_recursive():
+            if isinstance(node, gtir_library_nodes.Broadcast):
+                gtir_library_nodes.inplace_broadcast_expander(node, state, state.sdfg)
+            elif isinstance(node, gtir_library_nodes.GTIR_LIBRARY_NODES):
+                node.expand(state)
+
```
```suggestion
        nb_applied = 0
        for node, state in sdfg.all_nodes_recursive():
            if isinstance(node, gtir_library_nodes.Broadcast):
                gtir_library_nodes.inplace_broadcast_expander(node, state, state.sdfg)
                nb_applied += 1
            elif isinstance(node, gtir_library_nodes.GTIR_LIBRARY_NODES):
                node.expand(state)
                nb_applied += 1
        if validate_all and nb_applied > 0:
            sdfg.validate()
```

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
```diff
@@ -0,0 +1,422 @@
...
+    """
+
+    The transformation performs two kind of operations that are related to
+    broadcast library nodes. However, this transformations only operate on
+    scalar values that are broadcasted.
+
+    The first one is "static access replacement" in that operation the pattern:
```
edopao:
```suggestion
    The transformation performs two kinds of operations that are related to
    broadcast library nodes. However, this transformation only operates on
    broadcast of scalar values.

    The first one is "static access replacement" in that the operation pattern:
```

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
```diff
@@ -0,0 +1,422 @@
...
+
+    The second kind of operation is "AccessNode bypassing" in which the
+    following pattern is matched:
+    ```
+        (value_to_broadcast) -> (bcast_result) -> (access_node_consumer)
+    ```
+    to:
+    ```
+        (value_to_broadcast) -> (access_node_consumer)
+    ```
```
edopao, line 58:
```suggestion
    ```
        (value_to_broadcast)  -> [BroadCast] -> (bcast_result) -> (access_node_consumer)
    ```
    to:
    ```
        (value_to_broadcast) -> [BroadCast] -> (access_node_consumer)
    ```
```

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
```diff
@@ -0,0 +1,422 @@
...
+    Thus, the broadcast is performed directly into the target destination.
+
+    It is important that these two modes can be performed at the same time.
+    Meaning `bcast_result` can have multiple consumer.
```
edopao:
```suggestion
    Meaning `bcast_result` can have multiple consumers.
```

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
```diff
@@ -0,0 +1,422 @@
...
+    properties as dace_properties,
+    subsets as dace_sbs,
+    transformation as dace_transformation,
+)
+from dace.sdfg import graph as dace_graph, nodes as dace_nodes
+from dace.transformation.passes import analysis as dace_analysis
+
+from gt4py.next.program_processors.runners.dace import (
+    library_nodes as gtx_lib_nodes,
+    transformations as gtx_xtrans,
```
edopao: It is the first time I see this alias `gtx_xtrans`. As for the other, we can agree on a consistent alias across files (I have seen another alias above in this PR).

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
```diff
@@ -0,0 +1,422 @@
...
+
+        # The main reason for restricting ourselves to scalar broadcast is that
+        #  we can skip some analysis steps and simplify everything.
+        if not isinstance(bcast_value_desc, dace_data.Scalar):
+            return False
+
+        # A fundamental requirement is that `bcast_result` is only generated by us.
+        #  ADR-18 guarantees us this if it is transient and has a single producer,
+        #  `bcast_node`. However, since we will remove `bcast_result`, we have to
+        #  make sure that it is not used every where else.
```
edopao:
```suggestion
        #  make sure that it is not used anywhere else.
```

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
```diff
@@ -0,0 +1,422 @@
...
+        #   that the Maps consumers can be handled by `InlineBroadcastAccess`.
+
+        # Now we have to inspect all consumers of the result node.
+        for consumer_edge in graph.out_edges(bcast_result):
+            if consumer_edge.data.is_empty():
+                return False
+
+            match consumer := consumer_edge.dst:
+                case dace_nodes.AccessNode():
+                    # TODO(phimuell): Are there more checks needed.
```
edopao: I suggest removing this todo comment before merge, unless there are known cases.
```suggestion
```

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
```diff
@@ -0,0 +1,422 @@
...
+                    return False
+
+                case _:
+                    # This kind of node can not be handled.
+                    return False
+
+        # Check single use data if it was not known at the beginning.
+        if self._single_use_data is None:
+            find_single_use_data = dace_analysis.FindSingleUseData()
+            single_use_data = find_single_use_data.apply_pass(sdfg, None)
```
edopao, line 204: Would it be wrong to now store `single_use_data`? I am asking because it is used again inside `apply()`.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -247,6 +252,15 @@ def gt_auto_optimize(
         #  See also [issue#2095](https://github.com/spcl/dace/issues/2095)
         dace.Config.set("optimizer", "symbolic_positive", value=False)
 
+        # We need new transformations in order to deal with GTIR library nodes.
+        # For now, we simply expand these nodes before starting optimizing.
+        # TODO: Remove once transformations are ready.
```
edopao: Why not calling `ScalarBrodcastInliner` before expanding?

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
```diff
@@ -0,0 +1,422 @@
...
+            return False
+        if gtx_xtrans.utils.is_view(bcast_value_desc):
+            return False
+
+        # NOTE: The big question is if it favourable to perform the transformation
+        #   every time. If there are only AccessNodes consumers then the answer is
+        #   probably yes, as we can remove the read and write of the initial data
+        #   only the write to final destination is left. If the consumers are Maps
+        #   the thing is a bit different. As we have to create the intermediate
+        #   allocation. If the read of the memory is okay the `InlineBroadcastAccess`
```
edopao, line 159: `InlineBroadcastAccess` does  not exist yet.

edopao (commented):
Very good, just some minor comments.

edopao (commented):
Very good, just some minor comments.

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
philip-paul-mueller: I think this comment dates back when there was still the `.value` attribute.

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
philip-paul-mueller: I think the way it is now is a bit clearer because it makes it clear that "this is the Array case and there is the Scalar checks".

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
philip-paul-mueller: See my updated comment.

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
philip-paul-mueller: I am actually not a fan of `add_memlet_path()`.

src/gt4py/next/program_processors/runners/dace/library_nodes/broadcast.py
philip-paul-mueller: Yes, this thing was also very old. You can see the note at the end of the function.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
philip-paul-mueller: I think I hit it.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
philip-paul-mueller: I would rather keep it in this way, because it will work also in the future.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
philip-paul-mueller: Good idea.

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
philip-paul-mueller: No because, there is no guarantee that between now and later there is a change that makes data no longer single use.
By passing it on constructor the user kinds of guarantee that it is safe.

src/gt4py/next/program_processors/runners/dace/transformations/broadcast.py
philip-paul-mueller: I think that was the old name.
Then I removed it to `Scalar...`, because I thought that I onlyneed to handle scalars which was wrong.
So there is a todo to rename it, I think I will now switch to the old name.

philip-paul-mueller:
Here are the newest data from `8edefc0` which show that the speed is now comparable (at least is much better than [before](https://github.com/C2SM/icon4py/pull/1240#issuecomment-4398132484)).
If I look at my experiences with the `compute_advection_in_vertical` then the 3% degradation we observes are most likely caused by some broadcast loops that were (for whatever reasons) not integrated into other kernels.

philip-paul-mueller:
My current guess of why they are not integrated is because of their range.
In this PR the ranges are based on the size of the target location, i.e. the range is `[0, size_of_patch_to_broadcast)`, but before the range was based on the grid coordinates that were written, i.e. the range was `[136, 2000)`.
Furthermore, the Map splitter currently only consider the range and not the sizes.

I think the current design (with size) is better, because it makes handling the nodes much simpler and they contain less state (ideally they would be stateless, but this is not possible).
Thus, we should update the splitting tools to consider the size of the iteration spaces instead of their size.

edopao:
I have moved the library node for reduction with skip values to a separate PR #2603

PR #2387, opened 2025-11-11
style[cartesian]:  Cleanup in error messaging

katrinafandrich:
## Description
- Gt4py error messages have been updated to provide the user with more information to make for easier debugging
## Requirements

- [x] The PR contains an updated version of the `AUTHORS.md` file adding the names of all the new contributors.

### DIFF
--- a/src/gt4py/cartesian/frontend/gtscript_frontend.py
+++ b/src/gt4py/cartesian/frontend/gtscript_frontend.py
@@ -107,9 +107,10 @@ class AxisIntervalParser(gt_meta.ASTPass):
         cls,
         node: Union[ast.Ellipsis, ast.Slice, ast.Subscript, ast.Constant],
         axis_name: str,
+        fields: dict[str, nodes.FieldDecl],
         loc: Optional[nodes.Location] = None,
     ) -> nodes.AxisInterval:
-        parser = cls(axis_name, loc)
+        parser = cls(axis_name, fields, loc)
 
         if isinstance(node, ast.Ellipsis):
             interval = nodes.AxisInterval.full_interval()
@@ -144,8 +145,14 @@ class AxisIntervalParser(gt_meta.ASTPass):
 
         return nodes.AxisInterval(start=start, end=end, loc=loc)
 
-    def __init__(self, axis_name: str, loc: Optional[nodes.Location] = None):
+    def __init__(
+        self,
+        axis_name: str,
+        fields: dict[str, nodes.FieldDecl],
+        loc: Optional[nodes.Location] = None,
+    ):
         self.axis_name = axis_name
+        self.fields = fields
         self.loc = loc
 
         error_msg = "Invalid interval range specification"
@@ -169,16 +176,15 @@ class AxisIntervalParser(gt_meta.ASTPass):
         self,
         value: Union[int, None, gtscript.AxisIndex, nodes.AxisBound, nodes.VarRef],
         endpt: nodes.LevelMarker,
-    ) -> nodes.AxisBound:
+    ) -> nodes.AxisBound | nodes.RuntimeAxisBound:
         if isinstance(value, nodes.AxisBound):
             return value
 
         if isinstance(value, int):
             level = nodes.LevelMarker.END if value < 0 else nodes.LevelMarker.START
             offset = value
-        elif isinstance(value, nodes.VarRef):
-            level = value
-            offset = 0
+        elif isinstance(value, (nodes.VarRef, nodes.FieldRef)):
+            return nodes.RuntimeAxisBound(level=nodes.LevelMarker.START, offset=value, loc=self.loc)
         elif isinstance(value, gtscript.AxisIndex):
             level = nodes.LevelMarker.START if value.index >= 0 else nodes.LevelMarker.END
             offset = value.index + value.offset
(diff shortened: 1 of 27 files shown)


src/gt4py/cartesian/gtc/passes/oir_optimizations/temporaries.py
```diff
@@ -25,7 +25,9 @@ def visit_FieldAccess(
         offsets = node.offset.to_dict()
         if node.name in tmps_name_map:
             assert offsets["i"] == offsets["j"] == offsets["k"] == 0, (
```
FlorianDeconinck: This should be a try/except rather than an assert

FlorianDeconinck (changes requested):

src/gt4py/cartesian/gtc/passes/oir_optimizations/temporaries.py
romanc: do you mean `if` and then `raise`? like

```suggestion
            if offsets["i"] != 0 or offsets["j"] != 0 or offsets["k"] != 0:
                raise ValueError(
                    "No K-offset capabilities of 3D temporaries. Must "
                    f"define '{node.name}' as a FloatField."
                )
```

tests/eve_tests/unit_tests/test_type_definitions.py
```diff
@@ -17,7 +17,7 @@ def test_immutability(self):
 
         fl = FrozenList([0, 1, 2, 3, 4, 5])
 
-        with pytest.raises(TypeError, match="object does not support item assignment"):
+        with pytest.raises(TypeError, match="object does not support item assignment."):
```
romanc: This change fails the [test run](https://github.com/GridTools/gt4py/actions/runs/19270517249/job/55103843339) because the error message that python throws doesn't have a period at the end. With your change `pytest` will expect one and fail if there isn't one. It looks like we don't control this error message (it comes deep from the python stack), so I think you should just revert that change.

src/gt4py/cartesian/gtc/passes/oir_optimizations/temporaries.py
```diff
@@ -25,7 +25,9 @@ def visit_FieldAccess(
...
-                "Non-zero offset in temporary that is replaced?!"
+                "No K-offset capabilities of 3D temporaries. Must define '"
```
twicki: The check up top checks for `i` and `j` to also be offset. Why does the error message change to k only?

And you have the option to offset temporaries in k: 
```python 
        tmp = 100
    with computation(PARALLEL), interval(1, None):
        out_field = tmp[0, 0, -1]
```
is legit gtp4y code

src/gt4py/cartesian/gtc/common.py
twicki, line 1: The gt4py formatting is slightly different from the one we have in NDSL / PyFV3. Could there be an environment mismatch that led to this formatting? Or was something just not caught?

src/gt4py/cartesian/stencil_object.py
```diff
@@ -381,7 +381,7 @@ def _validate_args(  # Function is too complex
             )
         ):
             raise ValueError(
-                f"Compute domain too large (provided: {domain}, maximum: {max_domain})"
+                f"Compute domain too large (provided: {domain}, maximum: {max_domain}. Adjust K interval as needed.)"
```
twicki: I think the wrong `k` inteval is only one of the issues here, it could be a poorly provided `domain` for the computation

twicki (changes requested):

AUTHORS.md
```diff
@@ -35,3 +35,4 @@
 - Vogt, Hannes. ETH Zurich - CSCS
 - Weber, Benjamin. MeteoSwiss
 - Wicky, Tobias. Allen Institute for AI
+- Fandrich, Katrina. SSAI/NASA-GSFC
```
romanc: let's keep the list alphabetically sorted

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -2363,7 +2397,7 @@ def run(self, backend_name: str):
             parameters=[
                 parameter_decls[item.name] for item in api_signature if item.name in parameter_decls
             ],
-            computations=init_computations + computations,
+            computations=(init_computations + computations if init_computations else computations),
```
romanc: This looks like a bad merge. I changed this as part of https://github.com/GridTools/gt4py/pull/2388 because `init_computations` is always defined, so we don't need special handling in this case.

src/gt4py/cartesian/gtc/passes/oir_optimizations/temporaries.py
```diff
@@ -24,9 +24,11 @@ def visit_FieldAccess(
     ) -> Union[oir.FieldAccess, oir.ScalarAccess]:
         offsets = node.offset.to_dict()
         if node.name in tmps_name_map:
-            assert offsets["i"] == offsets["j"] == offsets["k"] == 0, (
-                "Non-zero offset in temporary that is replaced?!"
-            )
+            if offsets["i"] != 0 or offsets["j"] != 0 or offsets["k"] != 0:
+                raise ValueError(
+                    "No K-offset capabilities of 3D temporaries. Must "
```
romanc: From the condition, I'd assume that temporaries don't have any offset capabilities. I'd remove the explicit mention of K dimension.

src/gt4py/cartesian/frontend/gtscript_frontend.py
katrinafandrich: So I can remove the if else?

romanc: yes, it can just be
```py
computations=init_computations + computations,
```

twicki:
cscs-ci run

FlorianDeconinck:
cscs-ci run

Issue #2220, opened 2025-08-18
[cartesian] Variable K offset self-assignment is over guarded in FORWARD/BACKWARD mode


FlorianDeconinck:
A slight regression will occur in **_the next proposed version of the frontend._**

The GTIR will be checking that `absolute K` index are not used for self-assignment. Variable K offset got included as well (check `ParAssignment` in `gtir.py`). When computation is in `FORWARD/BACKWARD` this check is too restrictive, as we have allowed the user to assign to K. `gtir` doesn't have a concept of computation at this stage so we can:

1/ Trust our frontend checks and remove the guard
2/ Move computation-schedule down enough for GTIR to make a more comprehensive guard

Below is reproducer.

```python
from ndsl.dsl.gt4py import computation, interval, FORWARD
from ndsl.boilerplate import (
    get_factories_single_tile,
    get_factories_single_tile_orchestrated,
)
from ndsl.constants import X_DIM, Y_DIM, Z_DIM
from ndsl.dsl.typing import FloatField, Float, Int, IntField
from ndsl import StencilFactory, orchestrate
import numpy as np

domain = (2, 2, 5)

ORCHESTRATE = False

if ORCHESTRATE:
    stcil_fctry, qty_factry = get_factories_single_tile_orchestrated(
        domain[0], domain[1], domain[2], 0, on_cpu=True
    )
else:
    stcil_fctry, qty_factry = get_factories_single_tile(
        domain[0], domain[1], domain[2], 0, backend="dace:cpu"
    )


def the_stencil(in_field: FloatField, out_field: FloatField, lev: IntField):
    with computation(FORWARD), interval(...):
        out_field[0, 0, lev] += 2


class Code:
    def __init__(self, stencil_factory: StencilFactory):
        orchestrate(obj=self, config=stencil_factory.config.dace_config)
        self._the_stencil = stcil_fctry.from_dims_halo(
            func=the_stencil,
            compute_dims=[X_DIM, Y_DIM, Z_DIM],
        )

    def __call__(self, in_field: FloatField, out_field: FloatField, lev: IntField):
        self._the_stencil(in_field, out_field, lev)


if __name__ == "__main__":
    in_arr = qty_factry.zeros(
        dims=[X_DIM, Y_DIM, Z_DIM],
        units="inputs",
    )
    sz = domain[0] * domain[1] * domain[2]
    in_arr.view[:] = np.arange(sz, dtype=Float).reshape(domain)

[...]

PR #2388, opened 2025-11-11
fix[cartesian]: Allow self-assignment with offset in K dimension in sequential vertical loops

romanc:
## Description

This PR is about handling (i.e. allowing/rejecting) self-assignment as well as other read _and_ write situations in the same vertical loop. Some of those are not allowed in parallel mode and we do now guard against things like

```py
@gtscript.stencil(backend=backend)
def self_assignment_parallel(field: Field[np.float32]) -> None:
    with computation(PARALLEL), interval(1, None):
        field = field[0, 0, -1] * 2
```

which has an inherent race condition in parallel execution.

I've added the checks at the `gtir` level. I've done so because the `gtir`-level is the first layer to derive from the `eve` framework, provides a nice API to iterate over all the statements in a vertical loop / interval.

Tested by new unit tests and upstream (in NDSL) with https://github.com/NOAA-GFDL/NDSL/pull/316.

Related issue: https://github.com/GridTools/gt4py/issues/2220

## Requirements

  N/A

### DIFF
--- a/tests/cartesian_tests/integration_tests/multi_feature_tests/test_code_generation.py
+++ b/tests/cartesian_tests/integration_tests/multi_feature_tests/test_code_generation.py
@@ -1475,7 +1475,7 @@ def test_upcasting_both_sides_of_assignment(backend: str) -> None:
 
 
 def test_no_write_and_read_with_horizontal_offset() -> None:
-    with pytest.raises(ValueError, match="Self-assignment with offset is illegal."):
+    with pytest.raises(ValueError, match="Self-assignment with offset in I or J is illegal."):
 
         @gtscript.stencil(backend="debug")
         def self_assign_offset(field: Field[np.float64]) -> None:


tests/cartesian_tests/integration_tests/multi_feature_tests/test_code_generation.py
```diff
@@ -1472,3 +1472,71 @@ def test_upcasting_stencil(
...
+    @gtscript.stencil(backend=backend)
+    def self_assignment_parallel(field: Field[np.int32]) -> None:
+        with computation(FORWARD), interval(1, None):
+            field = field[0, 0, -1] * 2
+
+    @gtscript.stencil(backend=backend)
+    def self_assignment_2_parallel(field: Field[np.int32]) -> None:
+        with computation(FORWARD), interval(1, None):
+            tmp = field[0, 0, -1]
+            field = tmp * 2
```
FlorianDeconinck, line 1568: Things with I and/or J, which are valid errors

```python
def the_stencil(in_field: FloatField, out_field: FloatField):
        out_field = in_field[I + 1]
        in_field = out_field[I + 1]
```

and

```python
def the_stencil(in_field: FloatField, out_field: FloatField):
        out_field = in_field[I + 1]
        in_field = out_field[I + 1]
```

romanc: There's already unit test coverage for this in `test_gtir` and `test_gtir_to_oir`. I've added two integration test with similar stencils as you suggested.

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -2364,7 +2363,7 @@ def run(self, backend_name: str):
             parameters=[
                 parameter_decls[item.name] for item in api_signature if item.name in parameter_decls
             ],
-            computations=init_computations + computations if init_computations else computations,
+            computations=init_computations + computations,
```
romanc, line 2366: `init_computations` comes from `_make_init_compuations()` above which will always return a list. The returned list might be empty if there are no "init computations". In that case, the concatenation still works. Imo no need to have an `if` / `else` here.

src/gt4py/cartesian/gtc/gtir.py
```diff
@@ -99,25 +99,11 @@ def no_write_and_read_with_offset_of_same_field(
...
-                    .to_set()
-                )
+                eve.walk_values(instance.right)
+                .filter(_cartesian_fieldaccess)
+                .filter(lambda acc: acc.offset.i != 0 or acc.offset.j != 0)
+                .getattr("name")
+                .to_set()
             )
             if instance.left.name in offset_reads:
                 raise ValueError("Self-assignment with offset is illegal.")
```
FlorianDeconinck: Let's tailor the message


```suggestion
                raise ValueError("Self-assignment with offset in I or J is illegal.")
```

FlorianDeconinck (approved):
LGTM, small error message improvement

PR #2393, opened 2025-11-13
bug[next]: disallow implicit bool conversion of fields

havogt:
An old NumPy version caught the following mistake `np.cumsum(field)`, which should have been `np.cumsum(field.asnumpy())`. The result was a size-1 NumPy object-array which contained a Field. The mistake was not caught because `bool(field)==True`, `np.allclose` broadcasted the size-1 array to the size of the other array and did math operations that where all supported by the `Field` API which in the end resulted in an np.all check of an array of Fields, which is always `True`.

Long story short, we shouldn't allow implicit conversion of Fields to bool, which revealed more bugs in our tests.


havogt (commented):
I approve your changes (but can't press the button as its my PR).

PR #2394, opened 2025-11-13
refactor[next][dace]: Simplify visitor of `concat_where` expressions

edopao:
This PR reduces the number of lines of code in the lowering of `concat_where` to SDFG. This is achieved mainly by means of two changes:
- Reuse the logic for broadcasting of a scalar and of a field slice. Both kinds of input can be handled as a broadcast expression, which is currently lowered as a mapped tasklet. In a separate PR, we will introduce a broadcast library node for this.
- Handle the lower and upper domain of the concat domain with the same utility function (`_translate_concat_where_branch()`) since the difference between these two cases is minimal. The function argument `is_lower` allows to select one or the other.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
@@ -225,7 +225,7 @@ def translate_concat_where(
     #  - If the domain expression is unbound on range start (negative infinite),
     #    the expression on the true branch is to be considered the input for the
     #    lower domain.
-    #  - Viceversa, if the domain expression is unbound on range stop (positive
+    #  - Vice versa, if the domain expression is unbound on range stop (positive
     #    infinite), the true expression represents the input for the upper domain.
     infinity_literals = (gtir.InfinityLiteral.POSITIVE, gtir.InfinityLiteral.NEGATIVE)
     if mask_domain.ranges[concat_dim].start in infinity_literals:
@@ -246,7 +246,7 @@ def translate_concat_where(
         dtype = gtx_dace_utils.as_dace_type(node.type.dtype)
     else:
         # TODO(edopao): Refactor allocation of fields with local dimension and enable this.
-        raise NotImplementedError("concat_where with list output is not supported")
+        raise NotImplementedError("'concat_where' with list output is not supported")
 
     output, output_desc = sdfg_builder.add_temp_array(ctx.sdfg, output_shape, dtype)
     output_node = ctx.state.add_access(output)


src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -29,401 +31,137 @@
...
-    if concat_dim not in lower.gt_type.dims:  # type: ignore[union-attr]
+    concat_dim: gtx_common.Dimension,
+    concat_dim_bound_expr: gtir.Expr,
+    output_domain: domain_utils.SymbolicDomain,
+    output_type: ts.FieldType,
+    source_expr: gtir.Expr,
+    output_desc: dace.data.Array,
+    output_node: dace.nodes.AccessNode,
+    output_origin: Sequence[dace.symbolic.SymbolicType],
+    is_lower: bool,
```
philip-paul-mueller: This is just my impression, but I would think that the `is_lower` argument should come as first or almost the first argument, because it is determines _how_ the arguments have to be interpreted, just an idea.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -29,401 +31,137 @@
...
+) -> None:
```
philip-paul-mueller, line 51: You should add a top level doc string here.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -29,401 +31,137 @@
...
+    assert isinstance(source_expr.type, (ts.FieldType, ts.ScalarType))
+
+    source_domain = source_expr.annex.domain
+    if isinstance(source_expr.type, ts.ScalarType) or len(source_expr.type.dims) < len(
+        output_type.dims
+    ):
         """
```
philip-paul-mueller: `"""` is not Python's equivalent to `/* */`.
Thus you should use proper Python comments inside functions.
See also here: https://peps.python.org/pep-0008/#block-comments

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -452,18 +190,51 @@ def translate_concat_where(
...
-        sdfg_builder,
-        node.annex.domain,
-        mask_domain,
-        tb,
-        tb_domain,
-        fb,
-        fb_domain,
-    )
+    # Expect unbound range in the concat domain expression on lower or upper range:
+    #  - if the domain expression is unbound on lower side (negative infinite),
```
philip-paul-mueller: Is "lower" appropriate here, I would connect it with the `.start` attribute, but maybe I miss something.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -452,18 +190,51 @@ def translate_concat_where(
...
+    #  - viceversa, if the domain expression is unbound on upper side (positive
+    #    infinite), the true expression represents the input for the upper domain.
+    infinity_literals = (gtir.InfinityLiteral.POSITIVE, gtir.InfinityLiteral.NEGATIVE)
+    if mask_domain.ranges[concat_dim].start in infinity_literals:
+        bound_expr = mask_domain.ranges[concat_dim].stop
+        lower_expr, upper_expr = node.args[1:]
+    elif mask_domain.ranges[concat_dim].stop in infinity_literals:
+        bound_expr = mask_domain.ranges[concat_dim].start
+        upper_expr, lower_expr = node.args[1:]
+    else:
```
philip-paul-mueller, line 237: Maybe a stupid question, but were are "finite" domains handled?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -29,401 +31,137 @@
...
-        upper, upper_desc = _make_concat_scalar_broadcast(
-            ctx=ctx,
-            inp=upper,
-            inp_desc=upper_desc,
-            out_domain=node_domain,
-        )
+        if concat_dim not in source_domain.ranges:
+            source_domain.ranges[concat_dim] = (
+                domain_utils.SymbolicRange(
+                    start=output_domain.ranges[concat_dim].start, stop=concat_dim_bound_expr
```
philip-paul-mueller:
```suggestion
                    start=output_domain.ranges[concat_dim].start,
                    stop=concat_dim_bound_expr,
```
It looks more uniform that way, you could also put the expression below on one line as an alternative.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -29,401 +31,137 @@
...
-        # Same as previous case, but the scalar value is taken from `upper` input.
-        assert upper_domain.ranges.keys() == {concat_dim}
-        upper_domain = domain_utils.promote_domain(upper_domain, node_domain.ranges.keys())
-        upper, upper_desc = _make_concat_scalar_broadcast(
-            ctx=ctx,
-            inp=upper,
-            inp_desc=upper_desc,
-            out_domain=node_domain,
-        )
+        if concat_dim not in source_domain.ranges:
```
philip-paul-mueller, line 111: I Just for my understanding, if `concat_dim` is in `source_domain` then this means we have essentially a `where` expression, where the condition which side is taken, depends on an expression of the indices.
Is this correct?
And if yes, what is the difference from the `else` case below?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -29,401 +31,137 @@
...
+        source_domain = domain_utils.domain_intersection(source_domain, output_domain)
+
+        # Use a 'deref' field operator to broadcast the input expression on the target domain.
+        bcast_expr = im.as_fieldop("deref", source_domain.as_expr())(source_expr)
+        bcast_expr.type = output_type
+
+        source = sdfg_builder.visit(bcast_expr, ctx=ctx)
     else:
         """
         Handle here the _regular_ case, that is concat_where applied to two fields
```
philip-paul-mueller:
```suggestion
        Handle here the _regular_ case, that is `concat_where` applied to two fields
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -29,401 +31,137 @@
...
+    source_subset = []
+    output_subset = []
+    for dim, size, src_origin, dst_origin in zip(
+        all_dims,
+        output_desc.shape,
+        source.origin,
+        output_origin,
+        strict=True,
+    ):
+        if dim == concat_dim:
```
philip-paul-mueller, line 158: I am not sure, but I could imagine that adding some comments here might be beneficial, but feel free to ignore.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -452,18 +190,51 @@ def translate_concat_where(
...
+            concat_dim=concat_dim,
+            concat_dim_bound_expr=bound_expr,
+            output_domain=node.annex.domain,
+            output_type=node.type,
+            source_expr=source_expr,
+            output_desc=output_desc,
+            output_node=output_node,
+            output_origin=output_origin,
+            is_lower=(i == 0),
+        )
```
philip-paul-mueller: I am not sure, but it might be more clear to avoid the loop here.
You could also create a second wrapper that does it.

philip-paul-mueller (commented):
It generally looks good, most if not all of my comments are rather minor.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: Good point.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: The domain expression here is not necessarily a 1:1 mapping of what you see in user code. For user code expressions like `concat_where(IDim == 0,` the gt4py frontend will create two nested expressions, one for `]-oo, 0]` and one for `(0, +oo[` (I am not sure which one of the two includes the zero value, but for sure one of the two).

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: I think it is correct. It would be wrong if we supported field domain in reverse order, where range start > end. However, I also get your point: it more clear to refer to the range start/stop. and use lower/upper only for the domain.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: Short answer: the fact that `concat_dim` is part of the domain does not imply that the input field is defined on the full dimension range of the output domain.
Long answer: in one specific case, when both input expressions are defined on the full domain, we could lower the `concat_where` in the first case as a `where` expression. Note that the `where` expression becomes a Map with a `ConditionalBlock` inside, so the boundary condition is evaluated for each element of the output array. This requires the input fields to also be defined on the full domain of the output field. However, this is not possible in the most common case, because the input expressions are only defined on the lower or upper domain, respectively. The GT4Py domain inference restricts the domains to only what needs to be computed.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
edopao: The reason was to be able to support large code blocks, because I got `ERA001 Found commented-out code`. I moved the code examples to the top-level doc string.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -192,11 +221,11 @@ def translate_concat_where(
...
     concat_dim = next(iter(mask_domain.ranges.keys()))
 
-    # Expect unbound range in the concat domain expression on lower or upper range:
-    #  - if the domain expression is unbound on lower side (negative infinite),
+    # Expect unbound range in the concat domain expression on range start or end:
+    #  - If the domain expression is unbound on range start (negative infinite),
     #    the expression on the true branch is to be considered the input for the
     #    lower domain.
-    #  - viceversa, if the domain expression is unbound on upper side (positive
+    #  - Viceversa, if the domain expression is unbound on range stop (positive
```
philip-paul-mueller:
```suggestion
    #  - Vice versa, if the domain expression is unbound on range stop (positive
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_concat_where.py
```diff
@@ -452,18 +219,66 @@ def translate_concat_where(
...
+    # We use the concat domain, stored in the annex, as the domain of output field.
+    output_domain = gtir_domain.get_field_domain(node.annex.domain)
+    output_dims, output_origin, output_shape = gtir_domain.get_field_layout(output_domain)
+    assert output_dims == node.type.dims
+
+    if isinstance(node.type.dtype, ts.ScalarType):
+        dtype = gtx_dace_utils.as_dace_type(node.type.dtype)
+    else:
+        # TODO(edopao): Refactor allocation of fields with local dimension and enable this.
+        raise NotImplementedError("concat_where with list output is not supported")
```
philip-paul-mueller:
```suggestion
        raise NotImplementedError("`concat_where` with list output is not supported")
```

PR #2357, opened 2025-10-31
fix[cartesian]: Fix absolute field-access in while-conditionals

twicki:
## Description

We found that there was an issue with absolute k-indexing inside the conditional of a while loop. This was due to a missing propagation of the symbol-table into the conditional. This PR fixes that issue.

## Requirements

- [x] All fixes and/or new features come with corresponding tests.
- [ ] Important design decisions have been documented in the appropriate ADR inside the [docs/development/ADRs/](docs/development/ADRs/README.md) folder. N/A

PR #2395, opened 2025-11-13
feat[cartesian]: Runtime Interval Bounds

twicki:
## Description
In the context of porting physics parametrization, a repeating pattern in a lot of solver-like codes are column based loops that only affect parts of the column until a certain condition is met.
This PR implements the feature of allowing specification of the interval at runtime as an experimental feature - meaning we only have an implementation in the debug backend. 
For more details, consut the corresponding ADR.

## Requirements

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_dataflow.py
@@ -953,13 +953,6 @@ class LambdaToDataflow(eve.NodeVisitor):
                 )
             )(out_edges, out_syms)
 
-            # Isolated access node will make validation fail.
-            # Isolated access nodes can be found in `make_tuple` expressions that
-            # construct tuples from input arguments.
-            for data_node in nstate.data_nodes():
-                if nstate.degree(data_node) == 0:
-                    assert not data_node.desc(nsdfg).transient
-                    nsdfg.remove_node(data_node)
         else:
             result = outer_value
 
--- a/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
@@ -19,6 +19,7 @@ import dataclasses
 from typing import Any, Dict, Iterable, List, Mapping, Optional, Protocol, Sequence, Tuple, Union
 
 import dace
+from dace import subsets as dace_subsets
 from dace.frontend.python import astutils as dace_astutils
 
 from gt4py import eve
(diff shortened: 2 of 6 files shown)


src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -195,7 +207,10 @@ def _make_axis_bound(
         return nodes.AxisBound(level=level, offset=offset, loc=self.loc)
 
     def visit_Name(self, node: ast.Name) -> nodes.VarRef:
-        return nodes.VarRef(name=node.id, loc=nodes.Location.from_ast_node(node))
+        assert "K" not in self.fields[node.id].axes
```
FlorianDeconinck: No `assert`, we got burned enough with those in the compiler pipeline, do an if/raise

twicki: done

docs/development/ADRs/cartesian/experimental/runtime-intervals.md
```diff
@@ -0,0 +1,128 @@
...
+        temporary_mask = min(temporary_mask,K)
+with computation(FORWARD), interval(0, -1):
+    if K > temporary_mask: # ignoring the upper bound here for simplicity
+        rhoifc0j = pifc0 / ( r * 0.5 * ( thv0bot[K+1] + thv0top ) * exnifc0)
+```
+
+The problem with this solution is that:
+
+1. Bloats the user code, especially when these masks have to be carried around for long times.
+2. Generates performance issues on the CPUIn the physics we see patterns like this:
```
FlorianDeconinck: Copy-paste gone wrong, it's the same wording as above

docs/development/ADRs/cartesian/experimental/runtime-intervals.md
```diff
@@ -0,0 +1,128 @@
...
+        with interval(0, temporary):
+            out_field[0, 0, 0] = input_data[0, 0, 0]
+```
+
+To bound the interval.
+
+## Consequences
+
+The biggest issue that comes with allowing temporary fields as well as 2d fields is the fact that this enforces an `(i,j)(k)`-loop order. We think it is ok for such a requirement to exist and want to allow the backend to make the decision that is right on a case-by-case basis.
+
```
FlorianDeconinck, line 93: Another consequence is we break the guardrail of `gt4py` that guarantees you won't read or write out of bounds. Another feature (absolute indexing in K) is breaking this guardrail, and [we propose a runtime checker to be implemented](https://github.com/GridTools/gt4py/issues/1684)

docs/development/ADRs/cartesian/experimental/runtime-intervals.md
```diff
@@ -0,0 +1,100 @@
...
+            out_field[0, 0, 0] = input_data[0, 0, 0]
+
+        with interval(0, index_data):
+            out_field[0, 0, 0] = input_data[0, 0, 0]
+
+        with interval(0, temporary):
+            out_field[0, 0, 0] = input_data[0, 0, 0]
+```
+
+To bound the interval.
```
romanc: redundant with line 61?

src/gt4py/cartesian/frontend/defir_to_gtir.py
```diff
@@ -572,6 +574,15 @@ def visit_AxisBound(self, node: AxisBound) -> gtir.AxisBound:
             level=self.GT4PY_LEVELMARKER_TO_GTIR_LEVELMARKER[node.level], offset=node.offset
         )
 
+    def visit_RuntimeAxisBound(self, node: RuntimeAxisBound) -> gtir.AxisBound:
```
romanc:
```suggestion
    def visit_RuntimeAxisBound(self, node: RuntimeAxisBound) -> gtir.RuntimeAxisBound:
```

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -168,16 +175,21 @@ def _make_axis_bound(
...
         elif isinstance(value, nodes.VarRef):
-            level = value
-            offset = 0
+            level = nodes.LevelMarker.START
+            offset = value
+            return nodes.RuntimeAxisBound(level=level, offset=offset, loc=self.loc)
+        elif isinstance(value, nodes.FieldRef):
+            level = nodes.LevelMarker.START
+            offset = value
+            return nodes.RuntimeAxisBound(level=level, offset=offset, loc=self.loc)
```
romanc: how about

```suggestion
            return nodes.RuntimeAxisBound(level=nodes.LevelMarker.START, offset=value, loc=self.loc)
        elif isinstance(value, nodes.FieldRef):
            return nodes.RuntimeAxisBound(level=nodes.LevelMarker.START, offset=value, loc=self.loc)
```
? 

As far as I understand, the other `elif` cases just populate `level` and `offset` because these are used in L207 to construct the `AxisBound` that is returned.

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -194,7 +206,17 @@ def _make_axis_bound(
 
         return nodes.AxisBound(level=level, offset=offset, loc=self.loc)
 
-    def visit_Name(self, node: ast.Name) -> nodes.VarRef:
+    def visit_Name(self, node: ast.Name) -> nodes.Ref:
+        # Handle the field accesses
+        if node.id in self.fields:
+            if "K" in self.fields[node.id].axes:
+                raise ValueError(
+                    "Using a field with a K-Axis as a bound for an interval is invalid."
```
romanc: We could be nice and name the offending field, e.g.

```python
f"Using field `{node.id}` with a K-Axis as a bound for an interval is invalid."
```

src/gt4py/cartesian/frontend/nodes.py
```diff
@@ -728,17 +728,28 @@ def __rshift__(self, steps: int):
         return self.cycle(steps=steps)
 
 
+class BaseAxisBound(Node):
```
romanc: in `common.py`, you opted to inherit from `ABC`, should this one too?

src/gt4py/cartesian/gtc/gtir.py
```diff
@@ -250,6 +253,9 @@ def _size_one(interval: Interval) -> bool:
             if interval.start.level != interval.end.level:
                 # if the levels (start/end) aren't the same, we don't know at this stage
                 return False
+            if not (isinstance(interval.start, AxisBound) and isinstance(interval.end, AxisBound)):
```
romanc, line 256: How about we use `has_runtime_access()` from the interval?

```suggestion
            if interval.has_runtime_access():
```

I feel like this is a pattern that shows in a couple of places in this PR. We could also be more conservative and use `interval.has_static_bounds()` (to be added) instead.

src/gt4py/cartesian/gtc/numpy/oir_to_npir.py
```diff
@@ -208,8 +216,8 @@ def visit_VerticalLoopSection(
     ) -> npir.VerticalPass:
         return npir.VerticalPass(
             body=self.visit(node.horizontal_executions, **kwargs),
-            lower=node.interval.start,
-            upper=node.interval.end,
+            lower=self.visit(node.interval.start),
+            upper=self.visit(node.interval.end),
```
romanc: I think it would be nice to pass the `kwargs` on in case we need them when we actually implement `RuntimeAxisBounds` in the numpy backend. PR https://github.com/GridTools/gt4py/pull/2357 made me sensitive to silently dropping `kwargs`. Maybe I'm overly cautions now ... 

```suggestion
            lower=self.visit(node.interval.start, **kwargs),
            upper=self.visit(node.interval.end, **kwargs),
```

src/gt4py/cartesian/gtc/oir.py
```diff
@@ -19,7 +19,12 @@
 
 from gt4py import eve
 from gt4py.cartesian.gtc import common
-from gt4py.cartesian.gtc.common import AxisBound as AxisBound, LocNode as LocNode
+from gt4py.cartesian.gtc.common import (
+    AxisBound,
+    BaseAxisBound,
+    LocNode as LocNode,
```
romanc:
```suggestion
    LocNode,
```

src/gt4py/cartesian/gtc/oir.py
```diff
@@ -206,6 +237,14 @@ def check(cls: Type[UnboundedInterval], instance: UnboundedInterval) -> None:
         _check_interval(instance)
 
     def covers(self, other: Union[Interval, UnboundedInterval]) -> bool:
+        if not (
+            isinstance(self.start, AxisBound)
+            and isinstance(self.end, AxisBound)
+            and isinstance(other.start, AxisBound)
+            and isinstance(other.end, AxisBound)
+        ):
+            raise NotImplementedError("Can't assess if runtime intervals cover each other")
```
romanc, line 241: Do `UnboundedInterval` generally support `RuntimeAxisBounds`? The class definition currently only supports `Optional[AxisBound]`

https://github.com/GridTools/gt4py/blob/63b819e0753fc3fbe881a0f213103f4218d69119/src/gt4py/cartesian/gtc/oir.py#L230-L232

src/gt4py/cartesian/gtc/oir.py
```diff
@@ -160,37 +165,63 @@ def _check_interval(instance: Union[Interval, UnboundedInterval]) -> None:
...
         _check_interval(instance)
 
     def covers(self, other: Interval) -> bool:
+        if not (
+            isinstance(self.start, AxisBound)
+            and isinstance(self.end, AxisBound)
+            and isinstance(other.start, AxisBound)
+            and isinstance(other.end, AxisBound)
+        ):
+            raise NotImplementedError("Can't determine coverage of runtime intervals")
```
romanc, line 188: How about using `has_runtime_access()`?

```suggestion
        if self.has_runtime_access() or other.has_runtime_access():
            raise NotImplementedError("Can't determine coverage of runtime intervals")
```

src/gt4py/cartesian/gtc/passes/gtir_k_boundary.py
```diff
@@ -11,7 +11,7 @@
 from typing import Any, Dict, Tuple, Union
 
 from gt4py import eve
-from gt4py.cartesian.gtc import gtir
+from gt4py.cartesian.gtc import common, gtir
 from gt4py.cartesian.gtc.common import LevelMarker
```
romanc, line 15: not your code: if you feel like cleaning up, you could do a search & replace of `LevelMarker` -> `common.LevelMarker` in this file and then delete this line

romanc (approved):
Thanks for the write-up - nice ADR. I left a couple minor things inline.

What do you think of the following class hierarchy (e.g. in common.py)?

```python
class AxisBound(eve.Node, ABC):
    level: LevelMarker
    offset: ScalarAccess | FieldAccess | int

class RuntimeAxisBound(AxisBound):
    offset: ScalarAccess | FieldAccess

class StaticAxisBound(AxisBound):
    offset: int = 0
```

I think that's easier to read since both specialized subclasses are also specialized in their name, removing the need for a `Base*` prefix on the base class. As discussed before, no need to change now. I think that's minor and can also be a follow-up assuming you agree.

src/gt4py/cartesian/frontend/gtscript_frontend.py
twicki: done, I also consolidated it into a single `isinstance` check

src/gt4py/cartesian/frontend/nodes.py
twicki: done. 

I am not 100% sure if we want that - not only here, but also in common, because base classes like `Statement`, `Decl` or similar do not do it either. But that discussion can be tabled

src/gt4py/cartesian/gtc/gtir.py
twicki: As discussed, this makes mypy unhappy but I can try to see if a custom TypeGuard could work

src/gt4py/cartesian/gtc/oir.py
twicki: see above

FlorianDeconinck:
Random thought: do we guard against field used in bounds changing? How this looks like with our codegen? Should we disallow any write to it?

romanc:
cscs-ci run

FlorianDeconinck (approved):
Two follow up PRs:
- `dace:X` backend
- Guard against bound re-write in the interval

PR #2396, opened 2025-11-14
fix[next][dace]: Use explicit dataflow in let-lambdas

edopao:
This PR was created in the context of investigating the result validation failure in the `graupel` program. Note that this PR does not solve the bug, but it provides a better SDFG representation of let-lambdas.
The domain inference pass can detect that certain arguments are not used, and do not need to be computed. The corresponding access nodes are not created, in the SDFG, and we use `None` in the construction of the intermediate result for the lowering. However, on the interface to nested SDFGs, which are used to lower let-lambda nodes, the `None` arguments were not properly handled and resulted in transient nodes with uninitialized data. DaCe would just emit a warning, during validation, but would still accept this SDFG. This PR ensures that `None` nodes do not end up in the SDFG, and all arguments passed to a nested SDFG contain defined values, either globals or temporary nodes with the result of some intermediate computation.

- [x] ICON4Py CI tests pass
- [x] Validated in ICON-blueline
- [x] No performance regression in ICON-blueline

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
@@ -374,8 +374,8 @@ class SDFGBuilder(DataflowBuilder, Protocol):
             inner_ctx: The nested SDFG context, containing the state where `inner_result` is written.
             outer_ctx: The parent SDFG context, containing the state where the nested SDFG should be added.
             symbolic_args: Scalar argumemts to be passed to the nested SDFG through symbol mapping.
-            data_args: Data arguments to be passed through edge memlets. If None,
-                domain inference detected that this argument is not used.
+            data_args: Data arguments to be passed through edge memlets. It contains `None` values
+                for those arguments that, based on domain inference, should not be used.
             inner_result: The data produced by the nested SDFG, inside the state specified by `inner_ctx`.
             capture_outer_data: Allow capturing scalars and arrays defined in the parent SDFG.
 
@@ -927,7 +927,8 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
         Helper function to add storage for node parameters and connectivity tables.
 
         If `use_transient_storage=True`, all scalars and arrays are allocated as
-        transient data. When this data is accessed by the `SymRef` visitor, it is
+        transient data. When this data is accessed by the `SymRef` visitor or mapped
+        to input data inside a nested SDFG, during lowering of let-lambdas, it is
         turned into global. This allows, for nested SDFGs, to prune input connectors
         for data that is not used. The remaining unused transients in the SDFG are
         removed by the dace simplify pass. For connectivity arrays, we always use
--- a/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
@@ -635,7 +635,7 @@ def translate_tuple_get(
     access_nodes_to_remove = [
         arg.dc_node
         for arg in unused_data_nodes
-        if arg is not None and not arg.dc_node.desc(ctx.sdfg).transient
+        if not (arg is None or arg.dc_node.desc(ctx.sdfg).transient)
     ]
     assert all(ctx.state.degree(access_node) == 0 for access_node in access_nodes_to_remove)
     ctx.state.remove_nodes_from(access_nodes_to_remove)
--- a/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
+++ b/src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
@@ -436,8 +436,13 @@ def _lower_lambda_to_nested_sdfg(
     for edge in lambda_input_edges:
         edge.connect(map_entry=None)
 
-    # Create the body of the initialization state.
-    # This dataflow will initialize the scan carry variable.
+    # Create the body of the initialization state, which will initialize the scan
+    # carry variable. Note that we do it here, after having visited the lambda node
+    # and having built 'compute_state', for a specific reason: the scan carry variable
+    # was initially a lambda parameter, and threfore lowered as a non-transient node,
+    # but it needs to be changed into transient (see `scan_carry_desc.transient = True`)
+    # because it is only used as internal state. The input data will a different node,
+    # which is copied to the scan carry variable in the initialization state.
     scan_carry_input = (
         gtir_to_sdfg_utils.make_symbol_tree(scan_carry_symbol.id, scan_carry_symbol.type)
         if isinstance(scan_carry_symbol.type, ts.TupleType)
@@ -521,8 +526,9 @@ def _lower_lambda_to_nested_sdfg(
     # Corner case where the scan computation, on one level, does not depend on
     # the result from previous level. In this case, the state information from
     # previous level is not used, therefore we could find isolated access nodes.
-    # Note that this is probably a misuse of scan in application code: it could
-    # have been represented as a pure field ooperator.
+    # In case of tuples, it might be that only some of the fields are used.
+    # In case of scalars, this is probably a misuse of scan in application code:
+    # it could have been represented as a pure field operator.
     for arg in gtx_utils.flatten_nested_tuple((stencil_args[0],)):
         state_node = arg.dc_node
         if compute_state.degree(state_node) == 0:


src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -176,6 +177,58 @@ class SubgraphContext:
...
 
+    def copy_field(
+        self,
+        src: gtir_to_sdfg_types.FieldopData,
+        domain: gtir_domain.FieldopDomain | None = None,
+    ) -> gtir_to_sdfg_types.FieldopData:
+        """Create a copy of some data descriptor in the current SDFG context.
+
+        It allocates transient scalar or array storage with the exact same descriptor
+        as some other data in the same SDFG context, and creates an edge memlet
```
philip-paul-mueller: I think the description is misleading especially the word "some" as you essentially copy `src`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -176,6 +177,58 @@ class SubgraphContext:
     sdfg: dace.SDFG
     state: dace.SDFGState
 
+    def copy_field(
+        self,
+        src: gtir_to_sdfg_types.FieldopData,
+        domain: gtir_domain.FieldopDomain | None = None,
```
```suggestion
        domain: gtir_domain.FieldopDomain | None,
```
I woudl recommend to not provide a default value.
It might be less convenient, but it ensures that the user is always aware of what is going on.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -176,6 +177,58 @@ class SubgraphContext:
...
+                is translated into the subset to be copied.
+
+        Returns:
+            The new access node to the data copy.
+        """
+        data_desc = src.dc_node.desc(self.sdfg)
+        if isinstance(src.gt_type, ts.FieldType):
+            if domain is None:
+                out, out_desc = self.sdfg.add_temp_transient_like(data_desc)
+                out_origin = list(src.origin)
```
philip-paul-mueller, line 205:
```suggestion
                out_origin = tuple(src.origin)
```

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -414,12 +469,35 @@ def setup_nested_context(
...
+        # as dace symbols in parent SDFG. These symbols are mapped to dace symbols
+        # in the nested SDFG.
+        parent_sdfg_symbols = set()
+        for name, arg_type in lambda_symbols.items():
+            if isinstance(arg_type, ts.TupleType):
+                parent_sdfg_symbols |= {
+                    sym_id
+                    for gtsym in gtir_to_sdfg_utils.flatten_tuple_fields(name, arg_type)
+                    if isinstance(gtsym.type, ts.ScalarType)
+                    and (sym_id := str(gtsym.id)) in parent_ctx.sdfg.symbols
```
philip-paul-mueller: An optimization hint, calling `parent_ctx.sdfg.symbols` will start up a Python machinery.
So it is better to create a local variable for the symbols, i.e. `parent_symbs = parent_ctx.sdfg.symbols`, outside the for loop and then use that variable for testing.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -414,12 +469,35 @@ def setup_nested_context(
         else:
             lambda_symbols = {}
 
+        # We flatten all GTIR-symbols in current scope that are captured by the
+        # lambda expression, and filter the scalar symbols that are represented
+        # as dace symbols in parent SDFG. These symbols are mapped to dace symbols
+        # in the nested SDFG.
+        parent_sdfg_symbols = set()
```
philip-paul-mueller: I remember that there is another location where you do something similar.
Nevertheless I would recommend to create a helper function for that.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -482,16 +554,15 @@ def add_nested_sdfg(
         # are expected to be transient data. The caller is responsible for turning
         # them into global data and map it to arrays outside the nested SDFG.
         lambda_outputs = {
-            output.dc_node.data
+            data
```
```suggestion
            data_name
```
I would suggest to append `_name` to clearly transmit the fact that it is only the name and not the data descriptor.
I use this convention inside the optimization.
I think that for the refactoring we should sit together and discuss some uniform naming convention or at least some guide lines how to name things.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -482,16 +554,15 @@ def add_nested_sdfg(
...
             for output in lambda_output_data
-            if output is not None and output.dc_node.desc(inner_ctx.sdfg).transient
+            if output is not None and (data := output.dc_node.data) not in data_args
         }
 
         # Map free symbols to parent SDFG
         nsdfg_symbols_mapping = {}
         for dc_symbol in inner_ctx.sdfg.free_symbols:
             if dc_symbol in data_args:
-                assert isinstance(data_args[dc_symbol].gt_type, ts.ScalarType)
```
philip-paul-mueller, line 494: Out of curiosity why have you removed that assert?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -482,16 +554,15 @@ def add_nested_sdfg(
         # are expected to be transient data. The caller is responsible for turning
         # them into global data and map it to arrays outside the nested SDFG.
         lambda_outputs = {
-            output.dc_node.data
+            data
             for output in lambda_output_data
-            if output is not None and output.dc_node.desc(inner_ctx.sdfg).transient
+            if output is not None and (data := output.dc_node.data) not in data_args
```
philip-paul-mueller: Why was the transient test removed?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -509,26 +581,42 @@ def add_nested_sdfg(
 
         input_memlets = {}
         unused_data = set()
-        for nsdfg_dataname, nsdfg_datadesc in inner_ctx.sdfg.arrays.items():
-            if nsdfg_datadesc.transient:
+        for data, datadesc in inner_ctx.sdfg.arrays.items():
```
philip-paul-mueller: My personal opinion feel free to ignore: I would keep the `nsdfg_` prefix it makes it clear that you operate on things that are nested.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -509,26 +581,42 @@ def add_nested_sdfg(
...
-        for data in sorted(unused_data):  # NOTE: remove the data in deterministic order
-            inner_ctx.sdfg.remove_data(data, validate=gtx_config.DEBUG)
+                if (arg_node := data_args[data]) is None:
+                    unused_data.add(data)
+                else:
+                    input_memlets[data] = outer_ctx.sdfg.make_array_memlet(arg_node.dc_node.data)
+            else:
+                # Capture GTIR symbols (scalars, arrays) from the parent scope.
+                assert data in outer_ctx.sdfg.arrays
+                # Ensure that connectivity tables are non-transient arrays in parent SDFG.
```
philip-paul-mueller: Maybe you should add _why_ you do that.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -509,26 +581,42 @@ def add_nested_sdfg(
...
+                if data in connectivity_arrays:
+                    outer_ctx.sdfg.arrays[data].transient = False
+                input_memlets[data] = outer_ctx.sdfg.make_array_memlet(data)
+
+        if unused_data:
+            for data in sorted(unused_data):  # NOTE: remove the data in deterministic order
+                assert not inner_ctx.sdfg.arrays[data].transient
+                input_nodes_to_remove = [
+                    access_node
+                    for access_node in inner_ctx.state.data_nodes()
```
philip-paul-mueller: This is super in efficient as it will go though every node for every data.
You should go through it once before the loop and then store cache it.
Furthermore, I would add to make sure that there is only state, i.e. an `assert`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -509,26 +581,42 @@ def add_nested_sdfg(
...
+                    if access_node.data == data
+                ]
+                non_isolated_nodes = [
+                    input_node
+                    for input_node in input_nodes_to_remove
+                    if inner_ctx.state.in_degree(input_node) != 0
```
```suggestion
                    if inner_ctx.state.degree(input_node) != 0
```
This will also consider the case were something reads from a node that is supposed to be unused.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -509,26 +581,42 @@ def add_nested_sdfg(
...
+
+        if unused_data:
+            for data in sorted(unused_data):  # NOTE: remove the data in deterministic order
+                assert not inner_ctx.sdfg.arrays[data].transient
+                input_nodes_to_remove = [
+                    access_node
+                    for access_node in inner_ctx.state.data_nodes()
+                    if access_node.data == data
+                ]
+                non_isolated_nodes = [
```
philip-paul-mueller: I would change the name to something like `non_isolated_unused_nodes` it is longer, but much more clearer what it contains.
I would also add a short comment that outlines your intentions.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -721,27 +809,12 @@ def _visit_expression(
...
-                    dace.Memlet(
-                        data=dst,
-                        subset=dace.subsets.Range.from_array(dst_desc),
-                        other_subset=dace.subsets.Range.from_array(src_desc),
-                    ),
-                )
-                return gtir_to_sdfg_types.FieldopData(dst_node, src.gt_type, src.origin)
-
-        return gtx_utils.tree_map(_visit_result)(result)
+        if use_temp:
```
philip-paul-mueller: You should add a description of what `use_temp` does.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -1016,18 +1089,15 @@ def visit_Lambda(
...
-                pass  # domain inference has detetcted that this argument is not used
-            elif isinstance(arg, gtir_to_sdfg_types.SymbolicData):
+            if isinstance(arg, gtir_to_sdfg_types.SymbolicData):
+                gtsym_id = str(gtsym.id)
                 symbolic_args[gtsym_id] = arg
             else:
                 lambda_arg_nodes |= {
                     str(nested_gtsym.id): nested_arg
                     for nested_gtsym, nested_arg in gtir_to_sdfg_types.flatten_tuple(gtsym, arg)
-                    if nested_arg is not None  # we filter out arguments with empty domain
```
philip-paul-mueller, line 1030: Why have you removed the filtering of the unused data?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
```diff
@@ -1075,42 +1146,33 @@ def construct_output_for_nested_sdfg(
...
+            elif lambda_ctx.state.degree(inner_data.dc_node) == 0:
+                # Isolated access node will make validation fail.
+                # Isolated access nodes can be found in the join-state of an if-expression.
                 lambda_ctx.state.remove_node(inner_data.dc_node)
+            # Transient data nodes only exist within the nested SDFG. In order to return some result data,
+            # the corresponding data container inside the nested SDFG has to be changed to non-transient,
+            # that is externally allocated, as required by the SDFG IR. An output edge will write the result
+            # from the nested-SDFG to a new intermediate data container allocated in the parent SDFG.
+            outer_data = inner_data.map_to_parent_sdfg(
+                self, lambda_ctx.sdfg, ctx.sdfg, ctx.state, nsdfg_node.symbol_mapping
```
philip-paul-mueller: Maybe a stupid question, but I would say that this function should operate on the context itself and not on its member.
Maybe not possible due to design restrictions.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
```diff
@@ -599,22 +608,20 @@ def translate_tuple_get(
...
-        if arg is not None
-    )
-    unused_arg_nodes = tuple(arg for arg in unused_arg_nodes if arg is not None)
-    ctx.state.remove_nodes_from(
-        [arg.dc_node for arg in unused_arg_nodes if ctx.state.degree(arg.dc_node) == 0]
-    )
+    access_nodes_to_remove = [
+        arg.dc_node
+        for arg in unused_data_nodes
+        if arg is not None and not arg.dc_node.desc(ctx.sdfg).transient
```
philip-paul-mueller: I have a question why do you not remove _all_ unused data nodes as before?
Furthermore I do not understand the condition you apply, if you apply de-Morgan you arrive at:
```python
not (arg is None or arg.dc_node.desc(ctx.sdfg).transient)
```
I would actually expect that the condition should be the opposite, i.e. without the leading `not`.
Could you explain that?

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -515,18 +515,11 @@ def connect_scan_output(
...
-        if compute_state.degree(data_node) == 0:
-            # By construction there should never be isolated transient nodes.
-            # Therefore, the assert below implements a sanity check, that allows
-            # the exceptional case (encountered in one GT4Py test) where the carry
-            # variable is not used, so not a scan indeed because no data dependency.
-            assert (not data_node.desc(lambda_ctx.sdfg).transient) or data_node.data.startswith(
-                scan_carry_symbol.id
-            )
-            compute_state.remove_node(data_node)
+    # corner case where the scan state is not used in column update (not really a scan)
```
philip-paul-mueller: You should explain that better, because I have no clue what you mean by that.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
```diff
@@ -649,20 +642,18 @@ def translate_scan(
...
-    lambda_arg_nodes: dict[str, gtir_to_sdfg_types.FieldopData] = {}
+    lambda_arg_nodes: dict[str, gtir_to_sdfg_types.FieldopData | None] = {}
     for gtsym, arg in lambda_args_mapping:
-        gtsym_id = str(gtsym.id)
         if isinstance(arg, gtir_to_sdfg_types.SymbolicData):
+            gtsym_id = str(gtsym.id)
             symbolic_args[gtsym_id] = arg
-        elif arg is not None:
+        else:
             lambda_arg_nodes |= {
```
philip-paul-mueller: Stupid questions, but can you have symbols inside tuples.

philip-paul-mueller (commented):
Most comments are small, but there are some locations where I was really confused, especially in some filtering (one of the last comments).

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Good idea.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: `out_origin` has to be a list, because `get_field_layout()` below returns a list.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Done.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Yes, good point. I will add the naming convention to the refactoring.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: It was replaced with the test if the output data name is present among the inputs. It makes more sense, in my opinion, because it does not rely on some assumption how the nested SDFG was built.

I will update the comment and add an `assert`.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: I would have had to first assert that `data_args[dc_symbol] is not None`. I will put it back.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: You are right, I have now improved this search.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: All arguments are now passed to `add_nested_sdfg()`, including `None`, to ensure that the globals for `None` data are removed, but everything else is kept. This allows to detect uninitialized data in the form of dangling input connectors.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: Good point, fixed.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_primitives.py
edopao: If `arg is None`, there is no attribute `dc_node` and therefore no node to remove.

With respect to before, I have shuffled conditions being checked and conditions being asserted. By construction, `make_tuple` should not produce isolated transient nodes. Isolated nodes can only appear for global data, in case `make_tuple` combines multiple inputs into one tuple.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: Done

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg.py
edopao: I think there was before another location, but this should be the only location after #2372 

However, there is another piece of code which is duplicated, on line 1093, for which I can create a helper function.

src/gt4py/next/program_processors/runners/dace/gtir_to_sdfg_scan.py
edopao: This question is not stupid at all. I will create a helper function for this code, since it is duplicated, but I will check the need for symbolic scalars inside tuples in a separate PR. Who knows, maybe this is the source of the problem in the `graupel` program.

PR #2404, opened 2025-11-26
fix[next][dace]: Use symbol mapping for stride propagation across nested SDFGs

edopao:
This PR addresses a validation error observed in the `graupel` program in ICON4Py. The validation of this program failed on the `t` output field, although the computation of `t` was correct. The write to another output field, `pr`, which was not verified in the `graupel` test, was writing out of buffer because of wrong strides. The strides were incorrectly set during SDFG transformations because the symbol mapping was wrong for a specific lowering case, where the same field name `pr` was used inside the scope of a scan field operator and in the parent scope of a let-lambda expression which was accessing the `pr` program argument. This PR fixes the symbol mapping in strides propagation.

- [x] Verified in ICON4Py
- [x] Verified in ICON-blueline
- [x] Check performance in ICON-blueline

### DIFF
--- a/tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_strides.py
+++ b/tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_strides.py
@@ -99,7 +99,7 @@ def _make_strides_propagation_level2_sdfg() -> tuple[dace.SDFG, dace_nodes.Neste
         sdfg=sdfg_level3,
         inputs={"a3"},
         outputs={"c3"},
-        symbol_mapping={s3: s3.replace("3", "2") for s3 in sdfg_level3.free_symbols},
+        symbol_mapping={s3: s3 for s3 in sdfg_level3.free_symbols},
     )
 
     state.add_edge(state.add_access("a2"), None, nsdfg, "a3", dace.Memlet("a2[0:10]"))
@@ -147,7 +147,7 @@ def _make_strides_propagation_level1_sdfg() -> tuple[
         sdfg=sdfg_level2,
         inputs={"a2", "c2"},
         outputs={"a2_alias", "b2", "c2"},
-        symbol_mapping={s: s.replace("2", "1") for s in sdfg_level2.free_symbols},
+        symbol_mapping={s: s for s in sdfg_level2.free_symbols},
     )
 
     for inner_name in nsdfg_level2.in_connectors:
@@ -179,10 +179,9 @@ def test_strides_propagation():
     sdfg_level1, nsdfg_level2, nsdfg_level3 = _make_strides_propagation_level1_sdfg()
 
     # Tests if all strides are distinct in the beginning and match what we expect.
-    for level, sdfg in enumerate([sdfg_level1, nsdfg_level2.sdfg, nsdfg_level3.sdfg], start=1):
+    for sdfg in [sdfg_level1, nsdfg_level2.sdfg, nsdfg_level3.sdfg]:
         for aname, adesc in sdfg.arrays.items():
             exp_stride = f"{aname}_stride"
-            target_symbol = exp_stride.replace(str(level), str(level - 1))
             assert len(adesc.strides) == 1
             assert str(adesc.strides[0]) == exp_stride, (
                 f"Expected that '{aname}' has strides '{exp_stride}', but found '{adesc.strides}'."
@@ -190,7 +189,7 @@ def test_strides_propagation():
 
             if (nsdfg := sdfg.parent_nsdfg_node) is not None:
                 assert exp_stride in nsdfg.symbol_mapping
-                assert str(nsdfg.symbol_mapping[exp_stride]) == target_symbol
+                assert str(nsdfg.symbol_mapping[exp_stride]) == exp_stride
 
     # Now we propagate `a` and `b`, but not `c`.
     gtx_transformations.gt_propagate_strides_of(sdfg_level1, "a1")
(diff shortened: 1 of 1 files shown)


tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_copy_chain_remover.py
```diff
@@ -684,14 +718,21 @@ def test_a1_additional_output():
...
 
-    # Ensure that the SDFG was constructed in the correct way.
-    assert inner_sdfg.arrays["i0"].strides == sdfg.arrays["a"].strides
-    assert inner_sdfg.arrays["o0"].strides == sdfg.arrays["b"].strides
-    assert inner_sdfg.arrays["i0"].shape == inner_sdfg.arrays["o0"].shape
-    assert inner_sdfg.arrays["i0"].shape == sdfg.arrays["a"].shape
+    inner_input_desc = inner_sdfg.arrays["i0"].clone()
+    inner_output_desc = inner_sdfg.arrays["o0"].clone()
+
+    if not use_symbolic:
```
philip-paul-mueller: Why is this `if` here?
I would say this has to hold in all cases, if it is symbolic then the symbol mapping needs to be inspected to ensure that it is right.

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_copy_chain_remover.py
```diff
@@ -684,14 +718,21 @@ def test_a1_additional_output():
...
+def test_linear_chain_with_nested_sdfg(use_symbolic):
+    sdfg, inner_sdfg, nsdfg_node = _make_linear_chain_with_nested_sdfg_sdfg(use_symbolic)
 
-    # Ensure that the SDFG was constructed in the correct way.
-    assert inner_sdfg.arrays["i0"].strides == sdfg.arrays["a"].strides
-    assert inner_sdfg.arrays["o0"].strides == sdfg.arrays["b"].strides
-    assert inner_sdfg.arrays["i0"].shape == inner_sdfg.arrays["o0"].shape
-    assert inner_sdfg.arrays["i0"].shape == sdfg.arrays["a"].shape
+    inner_input_desc = inner_sdfg.arrays["i0"].clone()
+    inner_output_desc = inner_sdfg.arrays["o0"].clone()
```
philip-paul-mueller: you should call them `original_inner_{input, output}_desc`, to avoid confusion later as you keep them as a reference here.

src/gt4py/next/program_processors/runners/dace/transformations/strides.py
```diff
@@ -542,65 +506,39 @@ def _gt_map_strides_into_nested_sdfg(
...
-            assert sym.name in sdfg.symbols, f"Expected that '{sym}' is defined in the parent SDFG."
-            nsdfg_node.sdfg.add_symbol(sym.name, sdfg.symbols[sym.name])
-            nsdfg_node.symbol_mapping[sym.name] = sym
+    # Make sure that the stride symbols from the outer array are not already defined
+    #  inside the nested SDFG, by calling with `add_symbol()` with `find_new_name=True`.
+    for i, dim_ostride in enumerate(new_strides):
+        if str(dim_ostride).isdigit():
+            # A literal stride (e.g. `1`) can be set directly
+            new_strides[i] = dim_ostride
+        elif dim_ostride.is_symbol:
```
philip-paul-mueller: In case `singular_symbol == singular_symbol.free_symbols` holds, which I think it does, you might consider merging this and the `else` branch together.
Having this distinction between "singular" and "composed symbol" is technically nice, but it also complicates things.
So I am favour of removing it, but if you want to keep I am not against it.

src/gt4py/next/program_processors/runners/dace/transformations/strides.py
```diff
@@ -542,65 +506,39 @@ def _gt_map_strides_into_nested_sdfg(
...
-            if not (sym.name in nsdfg_node.sdfg.symbols or sym.name in nsdfg_node.symbol_mapping)
-        }
-
-        # Now propagate the symbols from the parent SDFG to the NestedSDFG.
-        for sym in missing_symbol_mappings:
-            assert sym.name in sdfg.symbols, f"Expected that '{sym}' is defined in the parent SDFG."
-            nsdfg_node.sdfg.add_symbol(sym.name, sdfg.symbols[sym.name])
-            nsdfg_node.symbol_mapping[sym.name] = sym
+    # Make sure that the stride symbols from the outer array are not already defined
+    #  inside the nested SDFG, by calling with `add_symbol()` with `find_new_name=True`.
```
philip-paul-mueller: I would write something like that:
For the strides on the inside we will create a new unique symbol inside the nested SDFG which is initialized, through the symbol mapping to the value it should have on the outside.
The benefit is, that it only the mapped container is affected and nothing else.
The main drawback of this is, that the connection is used, thus if the old stride symbol is used somewhere inside the nested SDFG, with the expectation that it is the stride of the inner container, then this connection is lost.
However, this does probably less of an issue for the strides and is probably more problematic for the shape, whose symbols are more likely to appear as loop bounds.

src/gt4py/next/program_processors/runners/dace/transformations/strides.py
```diff
@@ -542,65 +506,39 @@ def _gt_map_strides_into_nested_sdfg(
...
+            stride_symbol_mapping = {}
+            for osym in dim_ostride.free_symbols:
+                if (
+                    osym.name not in nsdfg_node.symbol_mapping
+                    or str(nsdfg_node.symbol_mapping[osym.name]) != osym.name
+                ):
+                    isym = nsdfg_node.sdfg.add_symbol(osym.name, osym.dtype, find_new_name=True)
+                    stride_symbol_mapping[isym] = osym
+            new_strides[i] = dim_ostride.subs(stride_symbol_mapping)
+            nsdfg_node.symbol_mapping |= stride_symbol_mapping
```
philip-paul-mueller: Maybe I have missed something, but assuming that the stride on the outside is given as `var1_outside + var2_outside`.
Then you create two symbols in the inner SDFG, let's say with names `var1_unique_name_inside` and `var2_unique_name_inside` and then use `var1_unique_name_inside + var2_unique_name_inside` as stride on the inside.
Then you add the following mapping to `symbol_mapping` `{'var1_unique_name_inside': 'var1_outside', 'var2_unique_name_inside': 'var2_outside'}`.

So my question now is, why not create a new unique symbol, let's say `new_unique_symbol_inside` use it as stride and add the following to the symbol mapping: `{'new_unique_symbol_inside': 'var1_outside + var2_outside'}`.
This would be simpler in my opinion.

src/gt4py/next/program_processors/runners/dace/transformations/strides.py
edopao: Very good!

src/gt4py/next/program_processors/runners/dace/transformations/strides.py
edopao: The solution you suggest was actually my initial implementation. The current solution was supposed to keep some logical connection, in case `var1_unique_name_inside` could be used in multiple places. However, given the explanation from your comment above, I think we could ignore the logical connection and go for the simpler solution.

philip-paul-mueller: As I said in our call, I am a bit unsure about this and have the feeling that regardless what we do, it will hit us.

philip-paul-mueller (approved):
LGTM

PR #2407, opened 2025-11-28
refactor[eve,next]: UIDGenerator -> SequentialIDGenerator

havogt:
Removes unused functionality from UIDGenerator, making it a SequentialIDGenerator only.

Refactors next to use a sequential ID per prefix by introducing a `IDGeneratorPool`, which provides a unique increasing index per prefix. Usage:
```
pool = IDGeneratorPool()
next(pool["prefix"]) # -> prefix_0
next(pool["prefix"]) # -> prefix_1
next(pool["other"]) # -> other_0
```

Additional changes:
- Make `uids` parameter in passes non-optional as they should all use (in the same pipeline) the same pool
- Change some test references because the old generator was 1-based, while the new is 0-based.

### DIFF
--- a/tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_dead_code_elimination.py
+++ b/tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_dead_code_elimination.py
@@ -8,7 +8,7 @@
 
 import pytest
 
-from gt4py.next import common
+from gt4py.next import common, utils
 from gt4py.next.type_system import type_specifications as ts
 from gt4py.next.iterator.ir_utils import ir_makers as im
 from gt4py.next.iterator import ir as itir
@@ -46,7 +46,9 @@ def program_factory(expr: itir.Expr) -> itir.Program:
         (im.let("val", "inp1")(im.if_(im.and_(True, True), "val", "inp2")), im.ref("inp1")),
     ],
 )
-def test_let_constant_foldable_if(input, expected, uids):
+def test_let_constant_foldable_if(
+    input: itir.Expr, expected: itir.Expr, uids: utils.IDGeneratorPool
+):
     input_program = program_factory(input)
     inlined = dead_code_elimination.dead_code_elimination(
         input_program, offset_provider_type={}, uids=uids


src/gt4py/eve/utils.py
```diff
@@ -872,63 +870,16 @@ def __hash__(self) -> int:  # type: ignore[override]
...
-        if warn_unsafe is None:
-            warn_unsafe = self.warn_unsafe
-        if warn_unsafe and start < next(self._counter):
-            warnings.warn("Unsafe reset of UIDGenerator ({self})", stacklevel=2)
-        self._counter = itertools.count(start)
-
-        return self
+    prefix: str = ""
+    format: str = "{prefix}_{id}"
+    _id: Iterator[int] = dataclasses.field(default_factory=itertools.count)
```
egparedes: Just a comment, I don't see why this needs to be private. The user may want to plug here any counter (for example a counter starting at a specific number).
```suggestion
    counter: Iterator[int] = dataclasses.field(default_factory=itertools.count)
```
Additionally, I think this could be a frozen dataclass

src/gt4py/eve/utils.py
```diff
@@ -872,63 +870,16 @@ def __hash__(self) -> int:  # type: ignore[override]
...
 
+    def __next__(self) -> str:
+        return self.format.format(prefix=self.prefix, id=next(self._id))
```
egparedes, line 887: I would add a `new_id()` (or `id()` or `next()`) convenience  function to get the new id.

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -194,7 +193,7 @@ def apply(
         enabled_transformations: Optional[Transformation] = None,
         # allow sym references without a symbol declaration, mostly for testing
         allow_undeclared_symbols: bool = False,
-        uids: Optional[eve_utils.UIDGenerator] = None,
+        uids: utils.SequentialPrefixedIDGenerator,
```
havogt: I decided to make all of them non-optional as correctness relies on the user to pass in a correct id generator.

src/gt4py/next/utils.py
```diff
@@ -388,3 +390,8 @@ def canonicalize_call_args(
     """
 
     return make_args_canonicalizer_for_function(func)(*args, **kwargs)
+
+
+class SequentialPrefixedIDGenerator(eve_utils.CustomDefaultDictBase):
```
tehrengruber:
```suggestion
class SequentialPrefixedIDGenerator(eve_utils.CustomDefaultDictBase[str, eve_utils.SequentialIDGenerator]):
```

Maybe something like:
`SequentialIDGeneratorPool`
`CollectiveSequentialIDGenerator`
`SharedSequentialIDGenerators`
The sequential term is also not essential in my opinion as we only care about uniqueness, and since this is in gt4py utils not eve, so close to the caller not the implementation of the generator I would personally go for `IDGeneratorPool`. I'm also fine with the existing name, but it took me a while to understand that there is `SequentialIDGenerator` and `SequentialPrefixedIDGenerator` and which one is which. If you like the original name most, how about appending an `s? That would be easy to memorize.

havogt: Yeah, wasn't happy with the name. I like `IDGeneratorPool`.

src/gt4py/eve/utils.py
```diff
@@ -871,64 +869,20 @@ def __hash__(self) -> int:  # type: ignore[override]
...
-        if start < 0:
-            raise ValueError(f"Starting value must be a positive number ({start} provided).")
-        if warn_unsafe is None:
-            warn_unsafe = self.warn_unsafe
-        if warn_unsafe and start < next(self._counter):
-            warnings.warn("Unsafe reset of UIDGenerator ({self})", stacklevel=2)
-        self._counter = itertools.count(start)
+    prefix: str = ""
+    counter: Iterator[int] = dataclasses.field(default_factory=itertools.count)
+    format: str = "{prefix}_{id}"
```
egparedes, line 880:
```suggestion
    #: A string to be used as template for the new ids.
    #: It should contain the `{prefix}`and `{id}` format keys.
    id_template: str = "{prefix}_{id}"
```

src/gt4py/next/iterator/transforms/collapse_tuple.py
```diff
@@ -194,7 +193,7 @@ def apply(
         enabled_transformations: Optional[Transformation] = None,
         # allow sym references without a symbol declaration, mostly for testing
         allow_undeclared_symbols: bool = False,
-        uids: Optional[eve_utils.UIDGenerator] = None,
+        uids: utils.IDGeneratorPool,
```
tehrengruber, line 196: Another option to the fixtures would be:
```suggestion
        uids: utils.IDGeneratorPool | None,
```
and then in the apply
```python
uids = uids or utils.IDGeneratorPool()
```
This would ensure explicitness in the tests, but avoid the countless duplication of `utils.IDGeneratorPool` calls.

src/gt4py/next/iterator/ir_utils/misc.py
```diff
@@ -30,15 +30,17 @@ class CannonicalizeBoundSymbolNames(eve.NodeTranslator):
...
     >>> cannonicalized_testee2 = CannonicalizeBoundSymbolNames.apply(testee2)
     >>> assert cannonicalized_testee1 == cannonicalized_testee2
     """
 
-    _uids: eve_utils.UIDGenerator = dataclasses.field(
-        init=False, repr=False, default_factory=lambda: eve_utils.UIDGenerator(prefix="_csym")
+    _uids: eve_utils.SequentialIDGenerator = dataclasses.field(
+        init=False,
+        repr=False,
+        default_factory=lambda: eve_utils.SequentialIDGenerator(prefix="_csym"),
```
egparedes, line 43: Just a suggestion in case you didn't think about using `functools.partial` instead of lambdas for the default factories. Feel free to ignore:
```suggestion
        default_factory=functools.partial(eve_utils.SequentialIDGenerator, prefix="_csym")),
```

src/gt4py/next/iterator/transforms/cse.py
```diff
@@ -292,17 +292,17 @@ def extract_subexpression(
...
+        >>> from gt4py.eve.utils import SequentialIDGenerator
         >>> expr = im.plus(im.plus("x", "y"), im.plus(im.plus("x", "y"), "z"))
         >>> predicate = lambda subexpr, num_occurences: num_occurences > 1
         >>> new_expr, extracted_subexprs, _ = extract_subexpression(
-        ...     expr, predicate, UIDGenerator(prefix="_subexpr")
+        ...     expr, predicate, SequentialIDGenerator("_subexpr")
         ... )
         >>> print(new_expr)
-        _subexpr_1 + (_subexpr_1 + z)
+        _subexpr_0 + (_subexpr_0 + z)
```
egparedes, line 302: Another alternative could have been to use a counter starting at `1` in the `SequentialIDGenerator`, right?

egparedes (commented):
LGTM, just a couple of optional suggestions and documentation additions.

src/gt4py/next/iterator/transforms/cse.py
havogt: Yes, but that's quite confusing, isn't it?

src/gt4py/next/iterator/ir_utils/misc.py
havogt: I'll skip this. It's one import more and not shorter, but actually didn't think of it.

src/gt4py/eve/utils.py
havogt: I find `id_template` to clumsy. Isn't `format` used for this kind of things? I added the documentation

PR #2408, opened 2025-12-01
ci[cartesian,next]: Upgrade scipy min version for py3.14

edopao:
Fix for daily CI tests with Python 3.14 and lowest resolution strategy, see: https://github.com/GridTools/gt4py/actions/runs/19793817840


pyproject.toml
```diff
@@ -141,7 +141,7 @@ next = ['gt4py[jax,standard,testing]']
 rocm4_3 = ['cupy-rocm-4-3>=13.3.0']
 rocm5_0 = ['cupy-rocm-5-0>=13.3.0']
 rocm6_0 = ['cupy>=13.4.1']
-standard = ['clang-format>=18.1', 'scipy>=1.14.1']
+standard = ['clang-format>=18.1', 'scipy>=1.14.1 ; python_version < "3.14"', 'scipy>=1.16.1 ; python_version >= "3.14"']
```
havogt: I would go with 1.16.1 for all python versions. What do you think @egparedes ?

havogt: (which is the first version that comes with 3.14 wheels)

edopao: `1.16.1` does no longer support py3.10

edopao: I could change it to:
`scipy>=1.14.1 ; python_version < "3.11"', 'scipy>=1.16.1 ; python_version >= "3.11"]`

havogt: Ok, maybe let's do that and it will disappear once we remove 3.10. maybe add a comment that this is only because of wheels availablility.

edopao:
Daily CI tests pass: https://github.com/GridTools/gt4py/actions/runs/19819292340

havogt (approved):
thanks!

PR #2409, opened 2025-12-01
refactor[next][dace]: Move SDFG-lowering modules into subpackage

edopao:
Move SDFG-lowering modules into a `lowering` subpackage. The objective is a better separation between the lowering to SDFG, the SDFG transformations and the SDFG callable module. Future rewrite or modifications of the SDFG-lowering, which happens inside the lowering module, should not impact on the SDFG callable interface, nor on the SDFG transformations.


philip-paul-mueller (approved):
LGTM

PR #2339, opened 2025-10-23
feat[next]: Support for staggered fields

SF-N:
TODO: add deprecation warning for `Koff[1]`
Maybe: Allow dimension in offset_provider and print depr warning

Stacked on-top of: #2411

PR #2411, opened 2025-12-04
refactor[next]: Use cartesian shift syntax in all tests

tehrengruber:
Prerequisite for staggered fields (#2339), split out so that PR reduces to the staggering feature itself.

### Changes
- **New `CartesianOffset` IR node.** `field(Dim ± n)` now lowers to `shift(cartesian_offset(Dim, Dim), n)` — a node carrying the offset's source/target dimensions *with their `DimensionKind`* — replacing the implicit string offset (`_OffDim`), which dropped the kind and broke vertical `KDim ± n` on the gtfn/dace/roundtrip pipelines. Handled in type inference/synthesis, domain inference, `trace_shifts`, pretty-print/parse, and the gtfn/dace/roundtrip backends (roundtrip cartesian-offset support was previously missing). The shift no longer needs an offset-provider entry — the node is self-describing.
- **Test migration.** All tests use the `field(Dim ± n)` syntax; the `Ioff`/`Joff`/`Koff` `FieldOffset`s and their offset-provider entries are removed, kept only where still required (dynamic `as_offset`, iterator-level string offsets).

### Why the node carries `domain` *and* `codomain`
`CartesianOffset` encodes a source→target dimension mapping. In this PR every cartesian shift is a **translation** (`domain == codomain`, e.g. `IDim → IDim`), enforced by explicit `assert domain == codomain` in the domain-inference and backend handlers. The second dimension exists for #2339's **relocation** shifts (`domain != codomain`, e.g. `IDim → IHalfDim` for half-cell ±0.5 staggered shifts): keeping the node shape here lets #2339 add relocation by dropping those asserts and implementing the `domain != codomain` case — without reshaping the node or re-touching every handler.

### Out of scope (stays in #2339)
No `Connectivity` `Protocol`→concrete refactor, and no staggering (`_Staggered` dims / ±0.5 half shifts).


tests/next_tests/unit_tests/iterator_tests/test_pretty_printer.py
```diff
@@ -197,6 +197,15 @@ def test_shift():
     assert actual == expected
 
 
+def test_cartesian_offset():
+    testee = ir.CartesianOffset(
+        domain=ir.AxisLiteral(value="I"), codomain=ir.AxisLiteral(value="J")
+    )
+    expected = "Iₕ₂Jₕ"
```
havogt:
```suggestion
    expected = "Iₕ->Jₕ"
```
what about using some kind of arrow instead?

tehrengruber: Sounds good. →?

src/gt4py/next/ffront/foast_to_gtir.py
```diff
@@ -304,19 +304,16 @@ def _visit_shift(self, node: foast.Call, **kwargs: Any) -> itir.Expr:
...
                     op=dialect_ast_enums.BinaryOperator.ADD | dialect_ast_enums.BinaryOperator.SUB,
-                    left=foast.Name(id=dimension),  # TODO(tehrengruber): use type of lhs
+                    left=foast.Name() as dim_name,
                     right=foast.Constant(value=offset_index),
                 ):
                     if arg.op == dialect_ast_enums.BinaryOperator.SUB:
                         offset_index *= -1
-                    # TODO(havogt): we rely on the naming-convention for implicit offsets, see `dimension_to_implicit_offset`
+                    assert isinstance(dim_name.type, ts.DimensionType)
+                    dim = dim_name.type.dim
```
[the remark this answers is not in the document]
havogt: It's already checked further up

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -180,6 +181,15 @@ def translate(
...
             off, val = shift
+            if isinstance(off, itir.CartesianOffset):
+                if val is trace_shifts.Sentinel.VALUE:
+                    raise NotImplementedError("Dynamic offsets not supported.")
+                assert isinstance(val, itir.OffsetLiteral) and isinstance(val.value, int)
+                dom = common.Dimension(value=off.domain.value, kind=off.domain.kind)
+                cod = common.Dimension(value=off.codomain.value, kind=off.codomain.kind)
+                assert dom == cod  # relocation (staggering) is not supported here
+                new_ranges[dom] = SymbolicRange.translate(self.ranges[dom], val.value)
+                return SymbolicDomain(self.grid_type, new_ranges)
```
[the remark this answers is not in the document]
tehrengruber: The proper fix is to make `common.Dimension.__eq__` also compare the kind.

src/gt4py/next/iterator/transforms/trace_shifts.py
```diff
@@ -137,7 +137,8 @@ def _can_deref(x):
 
 def _shift(*offsets):
     assert all(
-        isinstance(offset, ir.OffsetLiteral) or offset in [Sentinel.ALL_NEIGHBORS, Sentinel.VALUE]
+        isinstance(offset, (ir.OffsetLiteral, ir.CartesianOffset))
+        or offset in [Sentinel.ALL_NEIGHBORS, Sentinel.VALUE]
         for offset in offsets
     )
```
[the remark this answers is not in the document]
tehrengruber: Indeed this should be fixed.

src/gt4py/next/iterator/ir.py
```diff
@@ -109,6 +109,11 @@ class AxisLiteral(Expr):
     kind: common.DimensionKind = common.DimensionKind.HORIZONTAL
 
 
+class CartesianOffset(Expr):
+    domain: AxisLiteral
```
tehrengruber, line 113: Is this actually a good name?

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -511,6 +511,10 @@ def _resolve_dimensions(
         for off_literal in reversed(
             shift_tuple[::2]
         ):  # Only OffsetLiterals are processed, located at even indices in shift_tuple. Shifts are applied in reverse order: the last shift in the tuple is applied first.
+            if isinstance(off_literal, itir.CartesianOffset):
```
tehrengruber, line 510: The fact that this change is duplicated indicates we should at some point unify `_resolve_dimensions` with the `shift` type synth, as they do the same thing just in forward and reverse order. The loop order here has always been a brain twister to me even though I have been convinced for this to be correct, so I wouldn't do this here.

src/gt4py/next/iterator/embedded.py
```diff
@@ -191,8 +191,11 @@ def skip_value(
 
 
 # Offsets
-OffsetPart: TypeAlias = Tag | common.IntIndex
-CompleteOffset: TypeAlias = tuple[Tag, common.IntIndex]
+# A cartesian shift can be encoded by a `common.Dimension` tag (carrying its kind), shifting the
+# position along that dimension directly; named offsets use a string `Tag` resolved via the
+# offset provider.
+OffsetPart: TypeAlias = Tag | common.Dimension | common.IntIndex
```
tehrengruber: In the IR we have `CartesianOffset`, but here we have `Dimension`. That feels inconsistent. Eventually this should be `CartesianConnectivity`. Maybe this is a hint that this change also belongs in the PR here. Otherwise it is not easy to asses if disallowing `shift(ref_to_cart_conn, distance)` is a good idea.

I did not review the rest of this file.

src/gt4py/next/iterator/embedded.py
```diff
@@ -191,7 +191,10 @@ def skip_value(
 
 
 # Offsets
-OffsetPart: TypeAlias = Tag | common.IntIndex
+# A cartesian shift can be encoded by a `common.Dimension` tag (carrying its kind), shifting the
+# position along that dimension directly; named offsets use a string `Tag` resolved via the
+# offset provider.
+OffsetPart: TypeAlias = Tag | common.Dimension | common.IntIndex
 CompleteOffset: TypeAlias = tuple[Tag, common.IntIndex]
 OffsetProviderElem: TypeAlias = common.OffsetProviderElem
```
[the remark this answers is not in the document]
tehrengruber: This is a valid concern.

tehrengruber (commented):
Note for the discussion: The ITIR test changes are missing, feels like they should be in this PR.

src/gt4py/next/iterator/ir.py
havogt: You mean it should be something like domain_dim, codomain_dim?

havogt (commented):
A few non-blocking comments from a review of the `CartesianOffset` refactor. The correctness items we discussed (the deliberate removal of `Dimension` from `offset_provider`, the embedded-vs-traced `offset` handling, the duplicated `domain == codomain` guards) are intentionally left out — these are cleanup/convention notes.

src/gt4py/next/iterator/ir.py
```diff
@@ -109,6 +109,11 @@ class AxisLiteral(Expr):
     kind: common.DimensionKind = common.DimensionKind.HORIZONTAL
 
 
+class CartesianOffset(Expr):
```
havogt, line 112: `CartesianOffset` carries both `domain` and `codomain`, but every consumer in this PR asserts `domain == codomain` — the second dimension only exists for the relocation/staggering work in #2339. No ADR needed for this prep refactor, but the follow-up staggering PR (#2339) should document the design decisions (why the node carries `domain` + `codomain`, the translation-vs-relocation distinction, why the guards live in each backend) in an ADR under `docs/development/ADRs/next/` (cf. ADR 0005 'Extending_Iterator_IR').

src/gt4py/next/__init__.py
```diff
@@ -22,6 +22,7 @@
 from .._core.definitions import CUPY_DEVICE_TYPE, Device, DeviceType, is_scalar_type
 from . import common, ffront, iterator, program_processors, typing
 from .common import (
+    CartesianConnectivity,
```
havogt, line 25: `CartesianConnectivity` is re-exported here but not added to `__all__` below, unlike its sibling `Connectivity`. Per the next-package convention, public symbols should be listed in `__all__` so `from gt4py.next import *` binds them — and the roundtrip backend emits generated code calling `gtx.CartesianConnectivity(...)`, so it is effectively public. Please add it to `__all__`.

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -180,19 +181,19 @@ def translate(
...
                     raise NotImplementedError("Dynamic offsets not supported.")
                 assert isinstance(val, itir.OffsetLiteral) and isinstance(val.value, int)
-                current_dim = connectivity_type
-                # cartesian offset
-                new_ranges[current_dim] = SymbolicRange.translate(
-                    self.ranges[current_dim], val.value
-                )
-            elif isinstance(connectivity_type, common.NeighborConnectivityType):
+                dom = common.Dimension(value=off.domain.value, kind=off.domain.kind)
+                cod = common.Dimension(value=off.codomain.value, kind=off.codomain.kind)
```
havogt: These two inline `common.Dimension(value=..., kind=...)` reconstructions duplicate `misc.dim_from_axis_literal()`, the helper added in this same PR (and already used in the dace lowering). Suggest:

```python
dom = itir_misc.dim_from_axis_literal(off.domain)
cod = itir_misc.dim_from_axis_literal(off.codomain)
```

so the AxisLiteral→Dimension conversion lives in one place.

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -507,10 +507,12 @@ def _resolve_dimensions(
...
+                if off_literal.domain != off_literal.codomain:
+                    raise NotImplementedError("relocation (staggering) is not supported")
```
havogt: Error-message style (CODING_GUIDELINES §Error messages: sentence case, trailing period): `"Relocation (staggering) is not supported."`. Same string at line 665.

src/gt4py/next/iterator/ir_utils/ir_makers.py
```diff
@@ -588,6 +588,10 @@ def axis_literal(dim: common.Dimension) -> itir.AxisLiteral:
     return itir.AxisLiteral(value=dim.value, kind=dim.kind)
 
 
+def cartesian_offset(domain: common.Dimension, codomain: common.Dimension) -> itir.CartesianOffset:
```
havogt: `cartesian_offset(domain, codomain)` is only ever called with the same dim twice (`im.cartesian_offset(dim, dim)` at the three `foast_to_gtir` sites). Consider a single-argument maker `cartesian_offset(dim)` that sets `codomain=domain` internally — it removes the repeated `dim, dim` at every call site while leaving the `CartesianOffset` node two-dimensional for the #2339 staggering work.

src/gt4py/next/program_processors/runners/roundtrip.py
```diff
@@ -64,6 +64,14 @@ def visit_Literal(self, node: itir.Literal, **kwargs: Any) -> str:
     NoneLiteral = as_fmt("None")
     OffsetLiteral = as_fmt("{value}")
     AxisLiteral = as_fmt("{value}")
+
+    def visit_CartesianOffset(self, node: itir.CartesianOffset, **kwargs: Any) -> str:
+        # render as a `gtx.CartesianConnectivity(domain, codomain=codomain)` constructor call.
+        # The dimensions are injected as `gtx.Dimension`s in the generated module (see
+        # `axis_literals` below) and the embedded `shift` consumes the resulting connectivity
+        # as a self-describing cartesian offset.
```
havogt: This comment mostly restates what the `return` line already shows. The only non-obvious bit is the scope dependency on the injected dimension symbols. Suggest trimming to:

```python
# `node.{domain,codomain}.value` reference `gtx.Dimension`s injected into the generated
# module by the axis-literal collection; they must be in scope at the call site.
```

or dropping it entirely.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -144,31 +164,29 @@ def _collect_offset_definitions(
...
+                offset_definitions[axis.value] = TagDefinition(
+                    name=Sym(id=axis.value), alias=_vertical_dimension
                 )
-        elif isinstance(
-            connectivity_type := dim_or_connectivity_type, common.NeighborConnectivityType
-        ):
+
+    for offset_name, connectivity_type in offset_provider_type.items():
+        # cartesian shifts are encoded as `CartesianOffset` nodes (handled above) and don't
+        # occur in the `offset_provider_type`, which now only holds neighbor connectivities
```
havogt: This block duplicates the comment at line 167, and the 'handled above' cross-reference rots if the blocks move. Suggest removing it — the `isinstance(connectivity_type, common.NeighborConnectivityType)` check is self-explanatory. Keep the line-167 comment, which explains the non-obvious reason for the separate `CartesianOffset` walk.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -1661,31 +1665,39 @@ def _visit_shift(self, node: gtir.FunCall) -> IteratorExpr:
...
+            return self._make_cartesian_shift(it, offset_dim, offset_expr)
+
+        # first argument of the shift node is the offset provider
+        assert isinstance(offset_provider_arg, gtir.OffsetLiteral)
+        offset = offset_provider_arg.value
+        assert isinstance(offset, str)
+        offset_provider_type = self.subgraph_builder.get_offset_provider_type(offset)
+
+        # named offsets are always unstructured shifts; cartesian shifts are encoded as
+        # self-describing `CartesianOffset` nodes and handled above
```
havogt: The 'handled above' cross-reference is fragile. Suggest tightening to one non-obvious statement and dropping the rest (the transient-storage comment that follows is fine):

```python
# reaching here means a named offset → unstructured shift (cartesian shifts took the
# CartesianOffset branch above)
```

src/gt4py/next/iterator/ir_utils/ir_makers.py
tehrengruber: Let's do `cartesian_offset(domain: Dimension, codomain: Dimension | None)` then.

src/gt4py/next/program_processors/runners/roundtrip.py
tehrengruber: Let's drop the comment.

tehrengruber:
Good to go from my side.

havogt (approved):
lgtm

PR #2413, opened 2025-12-05
fix[next][dace]: Better usage of `SubgraphContext` during SDFG lowering

edopao:
Change the visitor class so that we can generate unique labels for temporaries/maps/tasklets across all nested SDFG.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
@@ -190,7 +190,13 @@ class SubgraphContext:
     def get_symbol_type(self, symbol_name: str) -> ts.DataType:
         return self.scope_symbols[symbol_name]
 
-    def input_data(self) -> set[str]:
+    def get_input_data(self) -> set[str]:
+        """Retrieve the names of arrays and scalars which are input data to this context.
+
+        Note that only the data that is currently used in this context, in other words
+        accessed in this SDFG state, is included in the returned set.
+        """
+
         flat_symbols = []
         for sym_name, sym_type in self.scope_symbols.items():
             if isinstance(sym_type, ts.TupleType):
@@ -494,8 +500,9 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
     """Provides translation capability from a GTIR program to a DaCe SDFG.
 
     A single instance of this visitor is used for the entire lowering, across all
-    levels of nested SDFGs. For each nested level, a new `SubgraphContext` is setup
-    with the data symbols in scope.
+    levels of nested SDFGs. The level-specific information, including the data symbols
+    available in the lowering scope, is stored inside a `SubgraphContext` object
+    that can be accessed by the visitor methods.
 
     This class is responsible for translation of `ir.Program`, that is the top level representation
     of a GT4Py program as a sequence of `ir.Stmt` (aka statement) expressions.
@@ -1051,7 +1058,7 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
             target: gtir_to_sdfg_types.FieldopData,
             target_domain: domain_utils.SymbolicDomain,
             target_state: dace.SDFGState,
-            is_target_inout: bool,
+            target_is_also_used_as_input: bool,
         ) -> None:
             assert source.dc_node.desc(ctx.sdfg).transient
             assert source.gt_type == target.gt_type
@@ -1059,13 +1066,14 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
             source_subset = _make_access_index_for_field(field_domain, source)
             target_subset = _make_access_index_for_field(field_domain, target)
 
-            if is_target_inout:
-                # write the field operator result in the separate target state
+            if target_is_also_used_as_input:
+                # write the field operator result in the separate target state,
+                # in order to ensure the correct dataflow for write after read
                 target_state.add_nedge(
                     target_state.add_access(source.dc_node.data),
                     target.dc_node,
                     dace.Memlet(
-                        data=target.dc_node.data, subset=target_subset, other_subset=source_subset
+                        data=source.dc_node.data, subset=source_subset, other_subset=target_subset
                     ),
                 )
             else:
@@ -1074,7 +1082,7 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
                     source.dc_node,
                     ctx.state.add_access(target.dc_node.data),
                     dace.Memlet(
-                        data=target.dc_node.data, subset=target_subset, other_subset=source_subset
+                        data=source.dc_node.data, subset=source_subset, other_subset=target_subset
                     ),
                 )
                 target_state.remove_node(target.dc_node)
@@ -1084,7 +1092,7 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
 
         # Visit the field operator expression.
         source_tree = self._visit_expression(stmt.expr, ctx)
-        ctx_input_data = ctx.input_data()
+        ctx_input_data = ctx.get_input_data()
 
         # In order to support inout argument, write the result in separate next state
         # this is needed to avoid indeterministic behavior for expressions like: X, Y = X + 1, X
@@ -1102,7 +1110,7 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
                 target,
                 target_domain,
                 target_state,
-                is_target_inout=(target.dc_node.data in ctx_input_data),
+                target_is_also_used_as_input=(target.dc_node.data in ctx_input_data),
             )
         )(source_tree, target_tree, domain)
 
--- a/src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
+++ b/src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
@@ -660,6 +660,7 @@ def translate_scalar_expr(
             # e.g. the type name `float64` used in casting expressions like `cast_(variable, float64)`
             visit_expr = str(arg_expr.id) in ctx.scope_symbols
         else:
+            # any other kind of node should always be visited
             visit_expr = True
 
         if visit_expr:


src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -66,23 +66,27 @@ def unique_map_name(self, name: str) -> str: ...
...
-        return sdfg.add_temp_transient(shape, dtype)
+        temp_name = self.unique_temp_name()
+        return sdfg.add_transient(temp_name, shape, dtype)
 
     def add_temp_array_like(
         self, sdfg: dace.SDFG, datadesc: dace.data.Array
     ) -> tuple[str, dace.data.Scalar]:
         """Add a temporary array to the SDFG."""
-        return sdfg.add_temp_transient_like(datadesc)
+        return sdfg.add_temp_transient_like(datadesc, name=self.unique_temp_name())
```
philip-paul-mueller: This is super personal and if feel free to ignore, but in the other function you stored the name in a temporary variable, here the unique name is not unique.
You see there is probably no difference in any way, but it is a bit inconsistent (even that word is too big for that).

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -588,13 +587,6 @@ def setup_nested_context(
 
         sdfg = dace.SDFG(name=self.unique_nsdfg_name(parent_ctx.sdfg, sdfg_name))
```
philip-paul-mueller: I know not part of the review, but I would call this variable `nsdfg` or something.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -615,13 +607,15 @@ def setup_nested_context(
         # When they are accessed, the corresponding data descriptor (scalar or array)
         # will be turned into global by setting `transient=False`. In this way, we remove
         # all unused (and possibly undefined) input arguments.
-        nsdfg_builder._add_sdfg_params(
+        self._add_sdfg_params(
             sdfg,
             node_params=input_params,
             symbolic_params=(domain_symbols | mapped_symbols | symbolic_inputs),
             use_transient_storage=True,
         )
```
philip-paul-mueller, line 647: I do not fully understand this change, more in the sense of "how is that possible"?
It is probably a change of logic that I missed.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -984,18 +971,7 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
         sdfg.debuginfo = gtir_to_sdfg_utils.debug_info(node)
 
         # start block of the stateful graph
-        entry_state = sdfg.add_state("program_entry", is_start_block=True)
-
-        # declarations of temporaries result in transient array definitions in the SDFG
-        if node.declarations:
```
philip-paul-mueller, line 990: Is there a specific reason why you removed.
I guess it is because it is not really used, but why was it added?

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1030,92 +1007,65 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
...
-        Each statement expression results in some sort of dataflow gragh writing to temporary storage.
-        The translation of `SetAt` ensures that the result is written back to the target external storage.
+        Each statement expression results in some sort of dataflow gragh writing to temporary
+        storage, inside the current state. The translation of `SetAt` ensures that the result
+        is written back to the target external storage, in a new state after the current one,
+        which will become the new head state.
 
         Returns:
-          The SDFG head state, eventually updated if the target write requires a new state.
+          The new SDFG head state, where the target fields are written.
```
philip-paul-mueller: I think it is more correct to say something like "Where the writes to the AccessNodes of the final results are located" or something.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1030,92 +1007,65 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
...
                 source, target, target_domain, target_state
             )
         )(source_tree, target_tree, domain)
 
-        if target_state.is_empty():
-            sdfg.remove_node(target_state)
-            return state
-        else:
-            return target_state
+        return target_state
```
philip-paul-mueller: I think you should return a context here
```suggestion
        return SubgraphContext(sdfg, target_state)
```
I think this could be more consistent in the overall design.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1030,92 +1007,65 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
...
+                ctx.state.remove_node(source.dc_node)
+
+        # Visit the domain expression.
+        domain = gtir_domain.extract_target_domain(stmt.domain)
+
+        # Visit the field operator expression.
+        source_tree = self._visit_expression(stmt.expr, ctx)
 
+        # In order to support inout argument, write the result in separate next state
+        # this is needed to avoid undefined behavior for expressions like: X, Y = X + 1, X
```
philip-paul-mueller:
```suggestion
        # this is needed to avoid indeterministic behavior for expressions like: X, Y = X + 1, X
```

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1030,92 +1007,65 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
...
+                target.dc_node,
+                dace.Memlet(
+                    data=target.dc_node.data, subset=target_subset, other_subset=source_subset
+                ),
+            )
+            if ctx.state.degree(source.dc_node) == 0:
+                ctx.state.remove_node(source.dc_node)
+
+        # Visit the domain expression.
+        domain = gtir_domain.extract_target_domain(stmt.domain)
```
philip-paul-mueller, line 1091: Was it purely "esthetic reasons" why you moved this part from the beginning of the function to here?
I do not mind, I actually think it is better, because it separates the sections.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1030,92 +1007,65 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
...
-        expr_input_args = {
-            sym_id
-            for sym in eve.walk_values(stmt.expr).if_isinstance(gtir.SymRef)
-            if (sym_id := str(sym.id)) in sdfg.arrays
-        }
-        state_input_data = {
-            node.data
-            for node in state.data_nodes()
-            if node.data in expr_input_args and state.degree(node) != 0
-        }
```
philip-paul-mueller, line 1067: Why have you removed that?

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1030,92 +1007,65 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
...
-                )
+            target_state.add_nedge(
+                # create in the target state new access nodes to the field operator result
+                target_state.add_access(source.dc_node.data),
+                target.dc_node,
+                dace.Memlet(
+                    data=target.dc_node.data, subset=target_subset, other_subset=source_subset
+                ),
+            )
+            if ctx.state.degree(source.dc_node) == 0:
```
philip-paul-mueller: What is this node?
Was it created in the (eager) assumption that it is an inout node and it is not?
Because it is also a bit strange considering the creation of the edge before.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
```diff
@@ -655,15 +655,12 @@ def translate_scalar_expr(
...
-                sdfg_builder.get_symbol_type(arg_expr.id)
-            except KeyError:
-                # all `SymRef` should refer to symbols defined in the program, except in case of non-variable argument,
-                # e.g. the type name `float64` used in casting expressions like `cast_(variable, float64)`
-                visit_expr = False
+            # all `SymRef` should refer to symbols defined in the program, except in case of non-variable argument,
+            # e.g. the type name `float64` used in casting expressions like `cast_(variable, float64)`
+            visit_expr = str(arg_expr.id) in ctx.scope_symbols
+        else:
+            visit_expr = True
```
philip-paul-mueller, line 664: This change is quite large, but I guess they should be equivalent, but work differently.
Before you checked if the name refers to a type and now you check if the name is known, is this approximately correct?

I would like to ask why it was different from before, because as far as I can tell (please correct me) `scope_symbols` was there before, was it just a over side.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -522,9 +522,6 @@ def make_field(
         )
         return gtir_to_sdfg_types.FieldopData(data_node, field_type, field_origin)
 
-    def get_symbol_type(self, symbol_name: str) -> ts.DataType:
```
philip-paul-mueller, line 525: (Optional) I would actually keep this function, because it is not fully obvious that `scope_symbols` maps symbol names to types.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_scan.py
```diff
@@ -542,7 +545,7 @@ def _handle_dataflow_result_of_nested_sdfg(
         # The field is used outside the nested SDFG, therefore it needs to be copied
         # to a temporary array in the parent SDFG (outer context).
         inner_desc.transient = False
-        outer_dataname, outer_desc = outer_ctx.sdfg.add_temp_transient_like(inner_desc)
+        outer_dataname, outer_desc = sdfg_builder.add_temp_array_like(outer_ctx.sdfg, inner_desc)
```
philip-paul-mueller, line 548: I kind of understand why the `add_temp_*()` things are in the builder, but these expressions looks a bit awkward and it might be worth considering to add them to the context.
But it might not be a good idea.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
edopao: `self` is an instance of `SDFGBuilder`, same as before. It acts on the `sdfg` passed as argument. From the `self` instance, it accesses the offset provider, which is the same through all nested levels.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
edopao: I moved it to `build_sdfg_from_gtir()`. I perform this check before calling the IR visitor.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
edopao: In order to simplify the lowering logic, I can always write the output field in a separate state, after the compute state, and "rely" on state fusion to remove the extra state when not needed.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
edopao: This happens in case of a conditional expression, which ends with a join state containing only an access node to the result, while the dataflow is in the then-else branches.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
edopao: Yes, purely esthetic reasons.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
edopao: Before, `scope_symbols` was a member of the `SDFGBuilder` implementation, but it was not available on the `SDFGBuilder` protocol interface. I used try/except (a bad choice, indeed) to detect failure in retrieving an entry in `scope_symbols`.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_scan.py
edopao: I agree. The problem is that `SDFG`Builder is one for the entire lowering, across all nested levels, while the context is defined per level. We need unique array names across all levels.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
edopao: I tried to use it everywhere, but here it would make things more complicated than needed, I think.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
philip-paul-mueller: Now I get it thanks.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_scan.py
philip-paul-mueller: This makes sense.
I would probably add that somewhere to the documentation.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
philip-paul-mueller: For my curiosity, who/where/why was the node put into that state?

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1030,92 +1011,65 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
...
         ) -> None:
-            target_desc = sdfg.arrays[target.dc_node.data]
-            assert not target_desc.transient
-
             assert source.gt_type == target.gt_type
             field_domain = gtir_domain.get_field_domain(target_domain)
             source_subset = _make_access_index_for_field(field_domain, source)
             target_subset = _make_access_index_for_field(field_domain, target)
 
-            if target.dc_node.data in state_input_data:
```
philip-paul-mueller, line 1088: I am a bit unsure about this logic that has been removed, was its functionality moved somewhere else?

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
edopao: The SDFG will look like this, at this stage, for a simple conditional expression:
image

The access node to `gtir_tmp_4` (and the state) is added by `gtir_to_sdfg_primitives.translate_if()`

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_scan.py
edopao: Added to the doc string of `GTIRToSDFG`.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
edopao: The answer is the same as for the comment above. In order to keep the lowering logic simple, I always write the temp data to the output fields in `target_state`. I rely on the state fusion pass to eliminate `target_state` if not needed. If this is confusing, I can put back the old logic.

edopao: I have reverted this change.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -186,6 +190,19 @@ class SubgraphContext:
     def get_symbol_type(self, symbol_name: str) -> ts.DataType:
         return self.scope_symbols[symbol_name]
 
+    def input_data(self) -> set[str]:
```
philip-paul-mueller: I would add a small description here.
The name is too general in my opinion.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -476,6 +493,10 @@ def flatten_tuple_args(
 class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
     """Provides translation capability from a GTIR program to a DaCe SDFG.
 
+    A single instance of this visitor is used for the entire lowering, across all
+    levels of nested SDFGs. For each nested level, a new `SubgraphContext` is setup
+    with the data symbols in scope.
```
philip-paul-mueller: Maybe reformulate to "The level specific information is stored inside a `SubgraphContext` instance that can be accessed ??"

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1030,28 +1051,40 @@ def _visit_target(
             target: gtir_to_sdfg_types.FieldopData,
             target_domain: domain_utils.SymbolicDomain,
             target_state: dace.SDFGState,
+            is_target_inout: bool,
```
philip-paul-mueller: I know you do not like long names, but maybe you should use `target_is_also_used_as_input` as name or `use_separete_write_state`.
Even if you do not like it, I would suggest to rename it to `target_is_inout`, because `is_target_inout` sounds more like a callback that can be called to find out if this is the case and not like an argument, which says "This is a fact, deal with it".

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1030,28 +1051,40 @@ def _visit_target(
...
-                target_state.add_access(source.dc_node.data),
-                target.dc_node,
-                dace.Memlet(
-                    data=target.dc_node.data, subset=target_subset, other_subset=source_subset
-                ),
-            )
-            if ctx.state.degree(source.dc_node) == 0:
-                ctx.state.remove_node(source.dc_node)
+            if is_target_inout:
+                # write the field operator result in the separate target state
```
philip-paul-mueller: I think you could keep the name if you extend the comment here.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1030,28 +1051,40 @@ def _visit_target(
...
+                target_state.add_nedge(
+                    target_state.add_access(source.dc_node.data),
+                    target.dc_node,
+                    dace.Memlet(
+                        data=target.dc_node.data, subset=target_subset, other_subset=source_subset
```
philip-paul-mueller: I would suggestion to "reverse the Memlet direction", i.e. `data=source.data, subset=source_subset, other_subset=target_subset`.
I think it might trigger less bad code paths in DaCe.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_primitives.py
```diff
@@ -655,15 +655,12 @@ def translate_scalar_expr(
     scalar_expr_args = []
 
     for i, arg_expr in enumerate(node.args):
-        visit_expr = True
         if isinstance(arg_expr, gtir.SymRef):
```
philip-paul-mueller, line 658: I am a bit conflicted here, I think in this case the `?:` operator could improve readability, but I also like the comment and the clear separation.
So feel free to switch to `?:` or keep it as it is, but I would add an expression why `visit_expr = True` is set in the `else` branch.

PR #2416, opened 2025-12-10
fix[next]: type checking with named collections in scans

havogt:
Improves support of named collections in scans.

### DIFF
--- a/src/gt4py/next/ffront/type_info.py
+++ b/src/gt4py/next/ffront/type_info.py
@@ -12,6 +12,7 @@ from typing import Iterator, Sequence, cast
 
 import gt4py.next.ffront.type_specifications as ts_ffront
 import gt4py.next.type_system.type_specifications as ts
+from gt4py.eve import datamodels
 from gt4py.eve.extended_typing import NestedTuple
 from gt4py.next import common, utils
 from gt4py.next.type_system import type_info


src/gt4py/next/iterator/type_system/inference.py
```diff
@@ -33,7 +33,7 @@ def _is_representable_as_int(s: int | str) -> bool:
 
 
 def _set_node_type(node: itir.Node, type_: ts.TypeSpec) -> None:
-    if node.type:
+    if node.type and not isinstance(type_, ts.DeferredType):
```
havogt, line 36: got the following in muphys

```
    def _set_node_type(node: itir.Node, type_: ts.TypeSpec) -> None:
        if node.type:
>           assert type_info.is_compatible_type(node.type, type_), (
                f"Node {node!s} already has a type {node.type} which differs from {type_}."
            )
E           AssertionError: Node precip_stateᐞ0 already has a type tuple[tuple[Field[[Cell, K], float64], Field[[Cell, K], float64], Field[[Cell, K], float64], Field[[Cell, K], bool]], tuple[Field[[Cell, K], float64], Field[[Cell, K], float64], Field[[Cell, K], float64], Field[[Cell, K], bool]], tuple[Field[[Cell, K], float64], Field[[Cell, K], float64], Field[[Cell, K], float64], Field[[Cell, K], bool]], tuple[Field[[Cell, K], float64], Field[[Cell, K], float64], Field[[Cell, K], float64], Field[[Cell, K], bool]], Field[[Cell, K], float64]] which differs from DeferredType(constraint=None).
```

havogt: @tehrengruber can you check this change?

tehrengruber: This is fine.

src/gt4py/next/type_system/type_info.py
```diff
@@ -487,6 +487,14 @@ def is_compatible_type(type_a: ts.TypeSpec, type_b: ts.TypeSpec) -> bool:
             return False
         for el_type_a, el_type_b in zip(type_a.types, type_b.types, strict=True):
             is_compatible &= is_compatible_type(el_type_a, el_type_b)
+    elif isinstance(type_a, ts.NamedCollectionType) and isinstance(type_b, ts.NamedCollectionType):
+        # NOTE: we are not checking if the `original_python_type` is compatible, this is a feature
```
tehrengruber: I'm not sure this is the semantic we want for all callers. In `_set_node_type` we explicitly want to check that nothing has changed, but ignore everything that is "unknown" or concretizable. In other words this function is merely a poor placeholder until we have a generic type system. Just ignoring the `original_python_type` doesn't really fit in that definition I feel (and hence might complicate the introduction of a generic type system). Where is this functionality triggered concretly (let me know if a chat is more convenient).

src/gt4py/next/ffront/type_info.py
```diff
@@ -80,12 +65,12 @@ def promote_arg(param: ts.TypeSpec, arg: ts.TypeSpec) -> ts.TypeSpec:
...
             param_el = param
             for idx in path:
-                if not isinstance(param_el, ts.TupleType):
+                if not isinstance(param_el, ts.COLLECTION_TYPE_SPECS):
                     # The parameter has a different structure than the actual argument. Just return
                     # the argument unpromoted and let the further error handling take care of printing
                     # a meaningful error.
                     return arg_el
-                param_el = param_el.types[idx]
+                param_el = param_el.types[idx]  # type: ignore[attr-defined] # checked in conditiona above
```
egparedes:
```suggestion
                param_el = param_el.types[idx]  # type: ignore[attr-defined] # checked in condition above
```

src/gt4py/next/named_collections.py
```diff
@@ -375,3 +375,33 @@ def make_constructor_expr(
...
+    def impl(
+        *args: Any | NamedCollection,
+    ) -> utils._R | NamedCollection:  # type annotation for human...
+        # TODO assert all inputs have the same (nested) collection structure
+        if is_named_collection_type(type(args[0])):
+            extract = make_named_collection_extractor(type(args[0]))  # type: ignore[arg-type]
+            construct = make_named_collection_constructor(type(args[0]), nested=True)  # type: ignore[arg-type]
+        else:
+            extract = identity
+            construct = identity
```
egparedes: Not important but I'd do:
```suggestion
            extract = construct = identity
```
or
```suggestion
            extract = construct = lambda x: x
```

src/gt4py/next/type_system/type_info.py
tehrengruber: Conclusion: The promotion or demotion where this occurs should create a `NamedCollectionType` with `original_python_type` being something, e.g. a Sentinel, which means that it is compatible with any other value in here, but in case the frontend tries to reconstruct the type an error is raised.

src/gt4py/next/ffront/type_info.py
```diff
@@ -186,13 +173,28 @@ def function_signature_incompatibilities_fieldop(
...
+    value: ts.CollectionTypeSpecT,
+    elems: NestedTuple[ts.DataType],
+) -> ts.CollectionTypeSpecT:
+    return (
+        ts.NamedCollectionType(
+            keys=value.keys, original_python_type=ts.ANY_PYTHON_TYPE_NAME, types=list(elems)
+        )
+        if isinstance(value, ts.NamedCollectionType)
+        else ts.TupleType(types=list(elems))  # type: ignore[return-value]
+    )
```
egparedes: Not super important, but I think it can be implemented in a different way to avoid duplicating the logic of the overriden function:
```suggestion
def _tree_map_type_constructor_drop_python_type(
    value: ts.CollectionTypeSpecT,
    elems: NestedTuple[ts.DataType],
) -> ts.CollectionTypeSpecT:
    result = _tree_map_type_constructor(value, elems)
    if isinstance(value, ts.NamedCollectionType):
        result = datamodels.evolve(result, original_python_type=ts.ANY_PYTHON_TYPE_NAME)
    return result
```

PR #2418, opened 2025-12-10
fix[next][dace]: Remove isolated access nodes to unused lambda args

edopao:
This PR fixes an issue in SDFG lowering of scan field operators, which contain unused arguments in the lambda expression.

Issue reproducer provided by @havogt.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
@@ -1836,9 +1836,12 @@ class LambdaToDataflow(eve.NodeVisitor):
         elif cpm.is_applied_shift(node):
             return self._visit_shift(node)
 
-        elif isinstance(node.fun, gtir.Lambda):
-            # Lambda node should be visited with 'visit_let()' method.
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -1937,7 +1930,16 @@ def visit_let(
             return self.visit_let(let_node.fun, args=let_args)
         else:
             # this lambda node is not a let-statement, but a stencil expression
-            return self.visit(node)
+            output = self.visit(node)
+            # remove access nodes to lambda arguments which are not used
+            flat_arg_nodes = [
+                x.field if isinstance(x, IteratorExpr) else x.dc_node  # type: ignore[attr-defined]
+                for x in gtx_utils.flatten_nested_tuple(tuple(self.symbol_map.values()))
+            ]
```
havogt:
```suggestion
            flat_arg_nodes = (
                x.field if isinstance(x, IteratorExpr) else x.dc_node  # type: ignore[attr-defined]
                for x in gtx_utils.flatten_nested_tuple(tuple(self.symbol_map.values()))
            )
```
I would default to a generator if list semantics is not explicitly needed

PR #2421, opened 2025-12-12
fix[next]: don't cse literal expressions (e.g. scan's init)

havogt:


src/gt4py/next/iterator/transforms/cse.py
```diff
@@ -161,6 +161,8 @@ def generic_visit(self, node, **kwargs):
     def visit(self, node: itir.Node, is_let_form: bool = False, **kwargs) -> None:  # type: ignore[override] # supertype accepts any node, but we want to be more specific here.
         # do not descent into un-applied lambda
         can_collect_children = not isinstance(node, itir.Lambda) or is_let_form
+        if cpm.is_call_to(node, "scan"):
+            can_collect_children = False
```
havogt: just for discussion...

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -496,6 +496,35 @@ def testee(
     )
 
 
+@pytest.mark.uses_scan
+def test_scan_init_duplicated(cartesian_case):
+    @gtx.scan_operator(axis=KDim, forward=True, init=((1.0,), (1.0,)))
```
havogt: in main the `{1.0}` tuple gets extracted

src/gt4py/next/iterator/transforms/cse.py
havogt: Till suggested to forbid cs extract literal expressions

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -496,6 +496,35 @@ def testee(
     )
 
 
+@pytest.mark.uses_scan
```
tehrengruber: This feels like a regression test. Do you mind moving it? It would also be nice to have a comment that explains what it is testing, in particular since (if I remember the reason correctly) the reason might disappear if we allow scalars on the top-level in gtfn.

tehrengruber (approved):

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
havogt: done

PR #2425, opened 2025-12-18
don't merge: workaround caching issue

havogt:
The NDArrayConnectivityField stores the inverse image for embedded in a `_cache` property which is part of the hash for lowering. This hash changes on each (interpreter) run, therefore the cache is not properly re-used.

We should find a more general solution than this workaround... next year...


kotsaloscv:
FYI -> https://github.com/eth-cscs/cscs-reframe-tests/pull/493
By patching on-the-fly GT4Py, the cache works as expected.

egparedes:
PR #2536 should fix the original issue in the proper way.

PR #2427, opened 2025-12-29
fix[cartesian] Remove the `vloop_sections` for a more unqiue `id(node)`

FlorianDeconinck:
## Description

Recent work with `dace:X` and Fortran memory aligned backend showcased some issue where the symbol renamed of K index in vertical loop. Prior we were using a local index to change the name to be able to do simple search and replace.

This breaks in non-obvious ways, leading to symbol straying. Unable to debug the `stree-to-sdfg` bridge, I moved to make the postfix of the index properly unique (via `id(node)`) which stabilized it for now.

Proper work is expected to be done, when we potentially revisit those for DaCe v2 and/or better vertical map merge.


romanc (approved):
LGTM! As discussed yesterday, this is totally fine.

PR #2429, opened 2026-01-09
fix[next]: Fix canonicalize unstructured domain in GTFN lowering

edopao:
The GTFN IR pass `_CannonicalizeUnstructuredDomain` only supports domain expressions with two dimensions, so it should skip domains with only one (vertical) dimension.


tehrengruber:
This particular issue is also addressed in #2403, but we found a more fundamental issue the is still blocking.

edopao:
> This particular issue is also addressed in #2403, but we found a more fundamental issue the is still blocking.

Thanks, this explains why the icon4py stencil (`KDim`-only) produces wrong computation result with this change.

PR #2431, opened 2026-01-10
fix[next]: Fix segfault for nanobind >=2.10

tehrengruber:
Temporary fix for a segfault with at least nanobind 2.10.2. The circumstances of the environment are not really clear yet, but this looks like a nanobind issue on first sight. I could reproduce the error with python 3.10 & 3.12.

Error message:
```
Fatal Python error: Segmentation fault

Current thread 0x00007d4596caeb80 (most recent call first):
  File "/home/tille/Development/gt4py_functional/src/gt4py/next/program_processors/runners/gtfn.py", line 67 in decorated_program
  File "/home/tille/Development/gt4py_functional/src/gt4py/next/backend.py", line 161 in __call__
  File "/home/tille/Development/gt4py_functional/src/gt4py/next/ffront/decorator.py", line 724 in __call__
```

Source of the error:
This innocent looking line in [src/gt4py/next/otf/compilation/compiler.py](https://github.com/Gridtools/gt4py/blob/ccf27ec/src/gt4py/next/otf/compilation/compiler.py#L86) triggers the error
```python
getattr(importer.import_from_path(src_dir / new_data.module), new_data.entry_point_name)
```
After the `getattr` call the module `importer.import_from_path(src_dir / new_data.module)` is garbage collected resulting in a call to `nanobind::detail::nb_module_clear`. This function (for unknown reasons) garbage collects the value stored in `static_pyobjects[pyobj_name::dl_version_tpl]` which is used in `nanobind/include/nanobind/ndarray.h` when calling the compiled program in [src/gt4py/next/program_processors/runners/gtfn.py](https://github.com/Gridtools/gt4py/blob/ccf27ec/src/gt4py/next/program_processors/runners/gtfn.py#L67).

Steps to debug:
1. Make sure you have python with debug symbols and the python extensions for gdb installed
2. ```
    set args -m pytest tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py -k test_copy
    break init_pyobjects
    run
    # stop when `static_pyobjects[pyobj_name::dl_version_tpl]` changes
    watch static_pyobjects[pyobj_name::dl_version_tpl]
    # ignore first change in init_pyobjects to the value
    continue
    # now you should land in `nanobind::detail::nb_module_clear`, printing
    # a python backtrace reveals the line in which the problem occurs
    py-bt
    ```

The proper solution is likely for each function in nanobind to keep a reference to the module.

Update: we opened an issue in the nanobind repo reporting this bug: https://github.com/wjakob/nanobind/issues/1283

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
@@ -12,6 +12,7 @@ from typing import Any, Optional
 
 import dace
 from dace import (
+    data as dace_data,
     dtypes as dace_dtypes,
     properties as dace_properties,
     subsets as dace_sbs,
@@ -152,6 +153,16 @@ class MoveDataflowIntoIfBody(dace_transformation.SingleStateTransformation):
 
         # If no branch has something to inline then we are done.
         if all(len(rel_df) == 0 for rel_df in relocatable_dataflow.values()):
+            return False
+
+        # Check if relatability is possible.
+        if not self._check_for_data_and_symbol_conflicts(
+            sdfg=sdfg,
+            state=graph,
+            relocatable_dataflow=relocatable_dataflow,
+            enclosing_map=enclosing_map,
+            if_block=if_block,
+        ):
             return False
 
         # Because the transformation can only handle `if` expressions that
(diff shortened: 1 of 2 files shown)


tehrengruber:
Other also seem to be running into this issue unless they install with pinned versions, e.g. from `uv.lock`. @egparedes @havogt I would like to not invest more time into investigating the issue in nanobind and potentially fixing it there at this moment. Should we either a) merge this (with an improved comment) or b) restrict nanobind to 2.9 for the time being.

PR #2433, opened 2026-01-12
feat[next][dace]: Enable async memory alloc on DaCe-HIP backend

edopao:
The new dace version`2026_01_13` adds support for async memory allocation in stream order for HIP codegen.


edopao:
cscs-ci run default

PR #2434, opened 2026-01-13
feat[next]: Enable unrolling scan loops

iomaganaris:
The PR adds a pass that unrolls scan loops (`LoopRegion`). It works in combination with https://github.com/spcl/dace/pull/2269.

It gets enabled in `icon4py` using: https://github.com/C2SM/icon4py/pull/992

TODO:
- [x] Decide whether it should be run by default or it should be enabled in icon4py as a hook (see https://github.com/C2SM/icon4py/commit/3247a8b6ae9ba6d31894b54f77ac8d0c0b38d14b)
- [x] Run `blueline` to check performance again

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
@@ -722,13 +722,11 @@ def _gt_auto_process_dataflow_inside_maps(
     )
 
     if scan_loop_unrolling:
-      sdfg.apply_transformations_once_everywhere(
-          gtx_transformations.ScanLoopUnrolling(
-              unroll_factor=scan_loop_unrolling_factor
-          ),
-          validate=False,
-          validate_all=validate_all,
-      )
+        sdfg.apply_transformations_once_everywhere(
+            gtx_transformations.ScanLoopUnrolling(unroll_factor=scan_loop_unrolling_factor),
+            validate=False,
+            validate_all=validate_all,
+        )
 
     return sdfg
 
--- a/src/gt4py/next/program_processors/runners/dace/transformations/scan_loop_unrolling.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/scan_loop_unrolling.py
@@ -16,15 +16,10 @@ from dace.sdfg import state as dace_state
 @dace_properties.make_properties
 class ScanLoopUnrolling(dace_transformation.MultiStateTransformation):
     """
-    Unrolls loops that are part of a GPU scan operation.
+    Unrolls loop regions that are part of a GPU scan operation.
     """
 
-    _default_unroll_factor: Final[int] = 4
-    unroll = dace_properties.Property(
-        dtype=bool,
-        default=True,
-        desc="Whether to unroll GPU scan loops.",
-    )
+    _default_unroll_factor: Final[int] = 0
     unroll_factor = dace_properties.Property(
         dtype=int,
         default=_default_unroll_factor,
@@ -39,20 +34,13 @@ class ScanLoopUnrolling(dace_transformation.MultiStateTransformation):
 
     def __init__(
         self,
-        unroll: Optional[bool] = None,
         unroll_factor: Optional[int] = None,
     ) -> None:
         super().__init__()
-        if unroll is not None:
-            self.unroll = unroll
-        if unroll_factor is not None and unroll_factor > 0:
-            self.unroll = True
+        if unroll_factor is not None and unroll_factor >= 0:
             self.unroll_factor = unroll_factor
-        elif unroll_factor is not None and unroll_factor == 0:
-            self.unroll = True
-            self.unroll_factor = 0
         elif unroll_factor is not None and unroll_factor < 0:
-            self.unroll = False
+            raise ValueError("Unroll factor must be non-negative.")
 
     def can_be_applied(
         self,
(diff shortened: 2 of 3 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -713,6 +721,14 @@ def _gt_auto_process_dataflow_inside_maps(
         validate_all=validate_all,
     )
 
+    sdfg.apply_transformations_once_everywhere(
+        gtx_transformations.ScanLoopUnrolling(
+            unroll=scan_loop_unrolling, unroll_factor=scan_loop_unrolling_factor
+        ),
+        validate=False,
+        validate_all=validate_all,
+    )
```
edopao: It would be more intuitive, in my opinion:
```suggestion
    if scan_loop_unrolling:
      sdfg.apply_transformations_once_everywhere(
          gtx_transformations.ScanLoopUnrolling(
              unroll_factor=scan_loop_unrolling_factor
          ),
          validate=False,
          validate_all=validate_all,
      )
```

It would work straight away because the `unroll` default value is `True`.

iomaganaris: Good point 👍 We save the SDFG traversal as well

src/gt4py/next/program_processors/runners/dace/transformations/scan_loop_unrolling.py
```diff
@@ -0,0 +1,78 @@
...
+
+import dace
+from dace import properties as dace_properties, transformation as dace_transformation
+from dace.sdfg import state as dace_state
+
+
+@dace_properties.make_properties
+class ScanLoopUnrolling(dace_transformation.MultiStateTransformation):
+    """
+    Unrolls loops that are part of a GPU scan operation.
```
edopao:
```suggestion
    Unrolls loop regions that are part of a GPU scan operation.
```

src/gt4py/next/program_processors/runners/dace/transformations/scan_loop_unrolling.py
```diff
@@ -0,0 +1,78 @@
...
+    ) -> None:
+        super().__init__()
+        if unroll is not None:
+            self.unroll = unroll
+        if unroll_factor is not None and unroll_factor > 0:
+            self.unroll = True
+            self.unroll_factor = unroll_factor
+        elif unroll_factor is not None and unroll_factor == 0:
+            self.unroll = True
+            self.unroll_factor = 0
```
edopao:
```suggestion
        if unroll_factor is not None and unroll_factor >= 0:
            self.unroll = True
            self.unroll_factor = unroll_factor
```

src/gt4py/next/program_processors/runners/dace/transformations/scan_loop_unrolling.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    Unrolls loops that are part of a GPU scan operation.
+    """
+
+    _default_unroll_factor: Final[int] = 4
+    unroll = dace_properties.Property(
+        dtype=bool,
+        default=True,
+        desc="Whether to unroll GPU scan loops.",
+    )
```
edopao: I think this just needs to be a class member:
```suggestion
    unroll = True
```

src/gt4py/next/program_processors/runners/dace/transformations/scan_loop_unrolling.py
```diff
@@ -0,0 +1,78 @@
...
+        self,
+        graph: Union[dace.SDFGState, dace.SDFG],
+        expr_index: int,
+        sdfg: dace.SDFG,
+        permissive: bool = False,
+    ) -> bool:
+        if self.scan_loop_region in self.applied_scans:
+            return False
+        if not self.scan_loop_region.label.startswith("scan_"):
+            return False
```
edopao, line 53: I would check the label first.

src/gt4py/next/program_processors/runners/dace/transformations/scan_loop_unrolling.py
iomaganaris: May I ask what is the difference between having the variable as a class member instead of a property?

edopao: Can it be that properties can be serialized to text (e.g. JSON)? Though I don't see why they are used in SDFG transformations, except for the history of transformations. I think I have seen that the SDFG representation can store the transformation history, in which case the transformations also need to be converted to text.

philip-paul-mueller: Sorry to barge in here.
But why should this here not be a property?
I am not even sure if it works, without prefixing it with `_`?
Having a property allows you to write something like:
```python
sdfg.apply_transformation(YourAwesomeTransformation, options={"your_awesome_option_that_is_a_dace_property": ...})
```

The actual use of it is in construction it in certain ways, like the above and to translate it to JSON.
This makes sense for properties on SDFGs but for transformation the only real use case is the undo button on the VS Code plugin.

edopao: In this case, I would say that `unroll` is not a configuration option. Why would I ever want to configure `ScanLoopUnrolling` with `unroll=False`? For me it is more of an internal (private) flag. Or (better IMO) we could just remove it and throw an exception if `ScanLoopUnrolling` is configured with `unroll_factor < 0`.

philip-paul-mueller: You are right, this sounds better.
My main point was more against a class variable, because they are evil.

iomaganaris: I agree that removing the `unroll` variable is the cleanest option 👍 Thanks for the suggestions

PR #2435, opened 2026-01-13
feat[cartesian]: Allow OpenMP to be  deactivated

FlorianDeconinck:
## Description

Move maps to `Sequential` when `GT4PY_CARTESIAN_DEACTIVATE_THREADING=True` for `dace:X` stencil backends

## Requirements

### DIFF
--- a/src/gt4py/cartesian/config.py
+++ b/src/gt4py/cartesian/config.py
@@ -71,6 +71,7 @@ os.environ.setdefault("DACE_CONFIG", os.path.join(os.path.abspath("."), ".dace.c
 
 DACE_DEFAULT_BLOCK_SIZE: str = os.environ.get("DACE_DEFAULT_BLOCK_SIZE", "64,8,1")
 
+"""Remove all OpenMP threading in dace:X backends for gt4py.cartesian"""
 GT4PY_CARTESIAN_DEACTIVATE_THREADING = os.environ.get(
     "GT4PY_CARTESIAN_DEACTIVATE_THREADING", "False"
 ).lower() in ["true", "1"]


edopao:
What is strange to me is that this problem only happens with Python "highest" strategy for dependency resolution, so there might be some newer python package that _somehow_ leads to different generated code. I will try locally on my Mac.

edopao:
> I don't see this problem locally on my Mac, using OpenMP from brew installation as in the CI job.
> 
> I will now try to use a newer MacOS image in the CI, which is the next candidate release (macos-26).

Unfortunately, it still fails on macos-26. Next test, I could switch to gcc, installed through brew.

edopao:
> > I don't see this problem locally on my Mac, using OpenMP from brew installation as in the CI job.
> > I will now try to use a newer MacOS image in the CI, which is the next candidate release (macos-26).
> 
> Unfortunately, it still fails on macos-26. Next test, I could switch to gcc, installed through brew.

I still see the same issue using gcc instead of default clang compiler:
` what():  thread_specific_storage constructor: could not initialize the TSS key!`

romanc:
> @romanc / @twicki : this PR didn't fix our issue, but do we want to keep the OpenMP deactivation alive anyway? Could be worthwhile to force hyperscaling strategy for GEOS

I agree that it could be useful and that this PR is low overhead. I would advice against merging for two reasons

1. We don't have no imminent use-case / need now. In my experience, adding options just because it _could_ be useful in the future often leads to code bloat that is hard to get rid of in the future.
2. If we indeed need it, I agree with @twicki that we could build the same post-opt (in NDSL). Unless we plan different optimization strategies (in NDSL) based on OpenMP availability, I think it's simpler to just do it just before we compile (& cache) the SDFG.

I'd argue let's wait for the need to arise and then driving the implementation from that use-case. We also have a task on the roadmap to centralize configuration and reduce the amount of environment variables. Adding one now just for the option that we might use it in the future seems to work against this goal.

PR #2436, opened 2026-01-14
fix[next][dace]: Make all node labels unique

edopao:
This PR ensures that some temporary scalars get unique name. The same fix was applied to the labels of nested SDFGs.

This should be the last source of duplicated labels. It was tested using this piece of code after the SDFG lowering:
```
all_data = []
all_maps = []
all_tlets = []
all_nsdfgs = []
for nsdfg in sdfg.all_sdfgs_recursive():
    all_data.extend(str(d) for d, desc in nsdfg.arrays.items() if isinstance(desc, (dace.data.Scalar, dace.data.Array)) and (d.startswith("__tmp" or d.startswith("gtir_tmp"))))
    all_maps.extend(n.label for st in nsdfg.states() for n in st.nodes() if isinstance(n, dace.nodes.MapEntry))
    all_tlets.extend(n.label for st in nsdfg.states() for n in st.nodes() if isinstance(n, dace.nodes.Tasklet))
    all_nsdfgs.append(nsdfg.label)
assert len(all_data) == len(set(all_data))
assert len(all_maps) == len(set(all_maps))
assert len(all_tlets) == len(set(all_tlets))
assert len(all_nsdfgs) == len(set(all_nsdfgs))
```

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
@@ -774,7 +774,7 @@ class LambdaToDataflow(eve.NodeVisitor):
                     deref_on_input_memlet=deref_on_input_memlet: self._visit_if_branch_arg(
                         if_sdfg,
                         if_branch_state,
-                        tsym.id,
+                        str(tsym.id),
                         targ,
                         deref_on_input_memlet,
                         if_sdfg_input_memlets,


philip-paul-mueller (commented):
It looks good, I just have two comments here:
- I would add your check, that everything is unique as a check that is checked in an `assert`.
- You forgot to the nested SDFGs in your checks.

philip-paul-mueller (approved):
LGTM.

PR #2437, opened 2026-01-14
feat[next]: Add instrumentation package and user-defineable hooks

egparedes:
This PR adds the instrumentation subpackage to `gt4py.next` containing the already existing performance metrics infrastructure and a new hook infrastructure to allow users to attach custom callbacks to different parts of the gt4py programs execution.

### DIFF
--- a/src/gt4py/next/otf/compilation/compiler.py
+++ b/src/gt4py/next/otf/compilation/compiler.py
@@ -83,11 +83,12 @@ class Compiler(
                     f"On-the-fly compilation unsuccessful for '{inp.program_source.entry_point.name}'."
                 )
 
-        compiled_prog = getattr(
-            importer.import_from_path(src_dir / new_data.module), new_data.entry_point_name
+        m = importer.import_from_path(
+            src_dir / new_data.module, sys_modules_prefix="gt4py.__compiled_programs__."
         )
+        func = getattr(m, new_data.entry_point_name)
 
-        return compiled_prog
+        return func
 
 
 class CompilerFactory(factory.Factory):
--- a/src/gt4py/next/otf/compilation/importer.py
+++ b/src/gt4py/next/otf/compilation/importer.py
@@ -8,11 +8,26 @@
 
 import importlib.util
 import pathlib
+import sys
 from types import ModuleType
 
 
-def import_from_path(module_file: pathlib.Path) -> ModuleType:
-    """Import all function objects from a Python module and return a mapping {function_name: object}."""
+def import_from_path(
+    module_file: pathlib.Path, *, add_to_sys_modules: bool = True, sys_modules_prefix: str = ""
+) -> ModuleType:
+    """
+    Load dynamically a python module from a given file path.
+
+    Args:
+        module_file: The path to the python file to load as a module.
+        add_to_sys_modules: Whether to add the loaded module to 'sys.modules'.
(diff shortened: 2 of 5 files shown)


src/gt4py/next/instrumentation/hook_machinery.py
```diff
@@ -0,0 +1,202 @@
...
+    """Generate a unique name for a callable object."""
+    return (
+        f"{func.__module__}.{getattr(func, '__qualname__', func.__class__.__qualname__)}#{id(func)}"
+    )
+
+
+def _is_empty_function(func: Callable) -> bool:
+    """Check if a callable object is empty (i.e., contains no statements)."""
+    try:
+        assert callable(func)
```
havogt:
```suggestion
    assert callable(func)
    try:
```
maybe. But why at all?

src/gt4py/next/otf/compiled_program.py
```diff
@@ -46,6 +47,61 @@
...
+
+        metrics.sources[source_key].metadata |= dict(
+            name=program_definition.definition.__name__,
+            backend=backend.name,
+            compiled_program_pool_key=hash(key),
+            **{
+                f"{eve_utils.CaseStyleConverter.convert(key.__name__, 'pascal', 'snake')}s": value
+                for key, value in argument_descriptors.items()
+            },
+        )
```
havogt: From a repository structure point of few, wouldn't it make more sense to register this later? Now we have the function `_metrics_prefix_from_pool_root` in this file which seems wrong.

src/gt4py/next/program_processors/runners/roundtrip.py
```diff
@@ -137,7 +137,7 @@ def fencil_generator(
     if cache_key in _FENCIL_CACHE:
         if debug:
             print(f"Using cached fencil for key {cache_key}")
-        return typing.cast(stages.CompiledProgram, _FENCIL_CACHE[cache_key])
+        return _FENCIL_CACHE[cache_key]  # A CompiledProgram is just a Callable
```
havogt, line 140: off topic: maybe `CompiledProgram` -> `ExecutableProgram` to include embedded in the naming?

src/gt4py/next/otf/compiled_program.py
havogt: The comment got detached by your recent changes, but still a question. Also the empty function stuff could just disappear if we don't have a default registered thing.

src/gt4py/next/instrumentation/hook_machinery.py
egparedes: Removed

src/gt4py/next/program_processors/runners/roundtrip.py
egparedes: Yes, you are right. It looks like a better name for the concept.

havogt (approved):
lgtm

src/gt4py/next/otf/compiled_program.py
egparedes: As discussed offline, it's not clear where to put metrics-related pool functions. The built-in implementation of the hooks is meant to be not optional, so user can not remove it. It might worth to revise this in the future, but it seems like a good idea to move away from the current implementation in small steps.

PR #2438, opened 2026-01-14
ci[cartesian]: Disable maocOS test in daily CI

edopao:
The following error happens in daily CI on macOS, only for the uv `highest` resolution strategy.

`libc++abi: terminating due to uncaught exception of type std::runtime_error: thread_specific_storage constructor: could not initialize the TSS key!`

TSS / TLS is the thread-local storage, and apparently the operating system refused to allocate a new thread-specific key. An explanation is that macOS has a low limit on TSS / TLS keys and, unlike Linux, macOS does not scale.

We could not find a way to workaround this issue, so we disable the failing test configuration for now.


edopao:
Awaiting result of daily CI pipeline here: https://github.com/GridTools/gt4py/actions/runs/21000406882/job/60368549910

PR #2440, opened 2026-01-16
refactor[next]: Add config variable `GT4PY_ADD_GPU_TRACE_MARKERS`

edopao:
This PR does not add any new feature, it only changes the way to enable an existing feature introduced in #2345.

The `COLLECT_METRICS_LEVEL` is supposed to be a runtime configuration, to enable or disable collection of the metrics. Besides, the current GPU trace is not producing any metric.

The NVTX/ROC-TX traces require to introduce some calls in the generated code. Therefore we need a separate configuration variable, to be checked at lowering/compile time, to allow the user to introduce the GPU TX markers in the generated code.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
@@ -267,7 +267,7 @@ duration = static_cast<double>(run_cpp_end_time - run_cpp_start_time) * 1.e-9;
         dace.Memlet(f"{output}[0]"),
     )
 
-    if gpu and config.ENABLE_GPU_TRACE:
+    if gpu and config.ADD_GPU_TRACE_MARKERS:
         sdfg.instrument = dace.dtypes.InstrumentationType.GPU_TX_MARKERS
         for node, _ in sdfg.all_nodes_recursive():
             if isinstance(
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -267,7 +267,7 @@ def add_instrumentation(sdfg: dace.SDFG, gpu: bool) -> None:
         dace.Memlet(f"{output}[0]"),
     )
 
-    if (config.COLLECT_METRICS_LEVEL == metrics.GPU_TX_MARKERS) and gpu:
+    if gpu and config.ENABLE_GPU_TRACE:
```
egparedes: How does this interact with the workflow cache(s) key(s)? I mean the code generated with traces should have a key in the cached steps different than the code generated without traces, but it doesn't seem to be any state in the code generator tracking this...

edopao: As discussed offline, it is a limitation the cache in gt4py-next. The same limitation applies to other configuration options such as `CMAKE_BUILD_TYPE` and `UNSTRUCTURED_HORIZONTAL_HAS_UNIT_STRIDE`.

.github/workflows/daily-ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         os: ["ubuntu-latest", "macos-latest"]
         python-version: ${{ fromJSON(needs.get-python-versions.outputs.python-versions) }}
         exclude:
-          # TODO(edopao): re-enable this test configuration once the following issue
+          # FIXME[#2439](edopao): re-enable this test configuration once the following issue
```
edopao, line 43: This is just an improvement of a previous PR, I had not thought about using the `FIXME` comment.

PR #2441, opened 2026-01-16
fix[next]: Test cleanups

havogt:


tests/next_tests/definitions.py
```diff
@@ -138,7 +138,6 @@ class ProgramFormatterId(_PythonObjectIdMixin, str, enum.Enum):
 )
 # Common list of feature markers to skip
 COMMON_SKIP_TEST_LIST = [
-    (REQUIRES_ATLAS, XFAIL, BINDINGS_UNSUPPORTED_MESSAGE),
```
havogt, line 141: in the past we directly used the buffers from atlas via a custom offset_provider. since this is not the case anymore it just works.

tests/next_tests/integration_tests/multi_feature_tests/ffront_tests/test_ffront_fvm_nabla.py
```diff
@@ -88,10 +82,6 @@ def test_ffront_compute_zavgS(exec_alloc_descriptor):
 
 @pytest.mark.requires_atlas
 def test_ffront_nabla(exec_alloc_descriptor):
-    # TODO(havogt): fix nabla setup to work with GPU
-    if exec_alloc_descriptor.allocator.device_type != core_defs.DeviceType.CPU:
```
havogt, line 92: seems to just work...

tests/next_tests/integration_tests/multi_feature_tests/iterator_tests/test_fvm_nabla.py
edopao, line 100: It seems that the iterator tests tests that fail with the dace backend should be marked as `USES_TUPLE_ITERATOR`. This marker was created for tests containing a field operator that returns a tuple. If you have a better name for the marker, please change it.

egparedes: @edopao , just a question for my own understanding: why is a marker needed for iterator tests returning tuples? Shouldn't it be already supported by the DaCe lowering?

havogt: thanks

edopao: The dace backend initially had support for lowering of field operators with tuple return. However, this was an exception, because the fieldview IR contains only tuples of field operators. The reason why I removed the support for the field operators on tuples, besides not being needed for fieldview IR, is that the inferred domain sometimes appeared as a tuple of domains, sometimes as a single domain, the latter not being aligned with the standard domain representation. So the lowering contained logic to identify the exceptional case. In the end I just cleaned up this extra logic.

PR #2442, opened 2026-01-16
feat[next]: add support for array_namespace allocation

havogt:
Refactors the constructor functions to allow construction from array namespaces. This will allow to use the allocators for jax fields in a next PR.

We introduce a new concept `Allocator` which is the only public type for the different kind of allocators (array namespace or custom layout allocators).

In addition to the `emtpy`, etc. functions which are used as `gtx.empty(..., allocator=..., device=..., alignment=...)` by the user, we offer a `FieldConstructor`-class which binds the low-level details of allocation (allocator, device, alignment).

### DIFF
--- a/src/gt4py/cartesian/backend/dace_stencil_object.py
+++ b/src/gt4py/cartesian/backend/dace_stencil_object.py
@@ -125,6 +125,11 @@ class DaCeStencilObject(StencilObject, SDFGConvertible):
             origin=origin,
             domain=domain,
         )
+        # strip history from SDFG
+        for tmp_sdfg in frozen_sdfg.all_sdfgs_recursive():
+            tmp_sdfg.transformation_hist = []
+            tmp_sdfg.orig_sdfg = None
+
         self._frozen_cache[key] = DaCeFrozenStencil(self, origin, domain, frozen_sdfg)
 
         basename = os.path.splitext(self.SDFG_PATH)[0]
--- a/src/gt4py/cartesian/gtc/common.py
+++ b/src/gt4py/cartesian/gtc/common.py
@@ -621,6 +621,11 @@ def native_func_call_dtype_propagation(*, strict: bool = True) -> datamodels.Roo
             NativeFunction.FLOAT64,
         ):
             instance.dtype = _precision_to_datatype(instance.func)  # type: ignore[attr-defined]
+        elif instance.func == NativeFunction.POW:
+            # Use non-strict to derive return type as max(type(base), type(exponent))
+            common_dtype = verify_and_get_common_dtype(cls, instance.args, strict=False)
+            if common_dtype:
+                instance.dtype = common_dtype  # type: ignore[attr-defined]
         else:
             # assumes all NativeFunction args have a common dtype
             common_dtype = verify_and_get_common_dtype(cls, instance.args, strict=strict)
--- a/src/gt4py/cartesian/gtc/passes/gtir_upcaster.py
+++ b/src/gt4py/cartesian/gtc/passes/gtir_upcaster.py
@@ -112,15 +112,17 @@ class _GTIRUpcasting(eve.NodeTranslator):
         )
 
     def visit_NativeFuncCall(self, node: gtir.NativeFuncCall) -> gtir.NativeFuncCall:
-        # Skip upcasting of cast operations
         if node.func in [
+            # Skip upcasting of cast operations
             common.NativeFunction.INT32,
             common.NativeFunction.INT64,
             common.NativeFunction.FLOAT32,
             common.NativeFunction.FLOAT64,
+            # Don't upcast the exponent of a power function because of the base, e.g. `3.0 ** 2`
+            common.NativeFunction.POW,
         ]:
-            # Make sure to upcast arguments of cast operations, e.g.
-            # float(my_int32 + my_int64) -> float(cast(int64, my_int32), my_int64)
+            # Make sure to upcast arguments of those operations anyway, e.g.
+            # float(my_int32 + my_int64) -> float(cast(int64, my_int32) + a my_int64)
             return _update_node(node, {"args": self.visit(node.args)})
 
         upcasting_rule = functools.partial(
--- a/tests/cartesian_tests/integration_tests/multi_feature_tests/test_code_generation.py
+++ b/tests/cartesian_tests/integration_tests/multi_feature_tests/test_code_generation.py
@@ -1600,6 +1600,26 @@ def test_upcasting_both_sides_of_assignment(backend: str) -> None:
     assert (input == output).all()
 
 
+@pytest.mark.parametrize("backend", ("debug",))  # ALL_BACKENDS)
+def test_upcasting_leave_integer_power_arguments_alone(backend: str) -> None:
+    domain = (5, 5, 5)
+
+    input = gt_storage.ones(backend=backend, shape=domain, dtype=np.float32)
+    output = gt_storage.zeros(backend=backend, shape=domain, dtype=np.float32)
+    squared = gt_storage.full(
+        fill_value=2, backend=backend, shape=(domain[0], domain[1]), dtype=np.int32
+    )
+
+    @gtscript.stencil(backend=backend)
+    def test_upcasting_stencil(
+        in_field: Field[np.float32], squared: Field[IJ, np.int32], out_field: Field[np.float32]
+    ) -> None:
+        with computation(FORWARD), interval(...):
+            out_field = in_field**squared
+
+    test_upcasting_stencil(input, squared, output)
+
+
 def test_no_write_and_read_with_horizontal_offset() -> None:
     with pytest.raises(ValueError, match="Self-assignment with offset in I or J is illegal."):
 
--- a/tests/cartesian_tests/unit_tests/test_gtc/test_gtir_upcaster.py
+++ b/tests/cartesian_tests/unit_tests/test_gtc/test_gtir_upcaster.py
@@ -35,8 +35,8 @@ AN_UNIMPORTANT_LITERAL = LiteralFactory(value="", dtype=DataType.DEFAULT)
 
 
 def contains_cast_node(cast_node, expr):
-    # Checks if `expr` contains `cast_node`. If `cast_node` contains `expr=Placeholder()`
-    # we skip equality check of `expr` and `cast_node.expr`
+    """Checks if `expr` contains `cast_node`. If `cast_node` contains `expr=Placeholder()`
+    we skip equality check of `expr` and `cast_node.expr`."""
     return (
         len(
             expr.walk_values()
@@ -51,6 +51,11 @@ def contains_cast_node(cast_node, expr):
     )
 
 
+def contains_no_cast_nodes(expr):
+    """Checks that `expr` contains no `cast_node`s."""
+    return len(expr.walk_values().if_isinstance(Cast).to_list()) == 0
+
+
 def upcast_and_validate(expr, expected_cast_nodes):
     assert isinstance(expected_cast_nodes, List)
     assert all([isinstance(cast, Cast) for cast in expected_cast_nodes])
@@ -60,6 +65,9 @@ def upcast_and_validate(expr, expected_cast_nodes):
 
     for cast in expected_cast_nodes:
         assert contains_cast_node(cast, result)
+
+    if not expected_cast_nodes:
+        assert contains_no_cast_nodes(result)
 
 
 def test_upcast_BinaryOp_BOOL_to_FLOAT():
@@ -119,4 +127,23 @@ def test_upcast_integers_division():
             Cast(dtype=DataType.FLOAT32, expr=LiteralFactory(value="1", dtype=DataType.INT32)),
             Cast(dtype=DataType.FLOAT32, expr=LiteralFactory(value="2", dtype=DataType.INT32)),
         ],
+    )
+
+
+def test_upcast_power_function():
+    testee = NativeFuncCall(func=NativeFunction.POW, args=[A_FLOAT64_LITERAL, A_INT64_LITERAL])
+    upcast_and_validate(
+        testee, []
+    )  # don't cast integer exponents of power function because of the base
+
+
+def test_upcast_power_function_arguments():
+    exponent = BinaryOp(
+        op=ArithmeticOperator.DIV,
+        left=LiteralFactory(value="1", dtype=DataType.INT32),
+        right=LiteralFactory(value="2", dtype=DataType.FLOAT32),
+    )
+    testee = NativeFuncCall(func=NativeFunction.POW, args=[A_FLOAT64_LITERAL, exponent])
+    upcast_and_validate(
+        testee, [Cast(dtype=DataType.FLOAT32, expr=LiteralFactory(value="1", dtype=DataType.INT32))]
     )
--- a/uv.lock
+++ b/uv.lock
@@ -1161,7 +1161,7 @@ wheels = [
 [[package]]
 name = "dace"
 version = "1.0.0"
-source = { git = "https://github.com/GridTools/dace?branch=romanc%2Fstree-v2#852c4bf30d973668ec8ce117a28db183962254a9" }
+source = { git = "https://github.com/GridTools/dace?branch=romanc%2Fstree-v2#8f7601b4e4953a1b785ac676701fc1d4e6540b37" }
 resolution-markers = [
     "python_full_version >= '3.14'",
     "python_full_version == '3.13.*'",
@@ -1707,7 +1707,7 @@ build = [
     { name = "wheel" },
 ]
 dace-cartesian = [
-    { name = "dace", version = "1.0.0", source = { git = "https://github.com/GridTools/dace?branch=romanc%2Fstree-v2#852c4bf30d973668ec8ce117a28db183962254a9" } },
+    { name = "dace", version = "1.0.0", source = { git = "https://github.com/GridTools/dace?branch=romanc%2Fstree-v2#8f7601b4e4953a1b785ac676701fc1d4e6540b37" } },
 ]
 dace-next = [
     { name = "dace", version = "43!2026.2.12", source = { registry = "https://gridtools.github.io/pypi/" } },


src/gt4py/_core/definitions.py
```diff
@@ -72,6 +72,21 @@
...
+    int16: "int16",
+    int32: "int32",
+    int64: "int64",
+    uint8: "uint8",
+    uint16: "uint16",
+    uint32: "uint32",
+    uint64: "uint64",
+    float32: "float32",
+    float64: "float64",
+}
```
egparedes: Not important but it's maybe less error prone to do:
```suggestion
dtype_to_name = {
    getattr(np, name): name
    for name in (
        "bool",
        "int8",
        "int16",
        "int32",
        "int64",
        "uint8",
        "uint16",
        "uint32",
        "uint64",
        "float32",
        "float64",
    )
}
```
Also, I think we should update `bool_ = np.bool_` to just `bool_ = np.bool`

src/gt4py/_core/ndarray_utils.py
```diff
@@ -19,7 +21,112 @@
...
+    def uint64(self) -> Any: ...
+
+    @property
+    def float32(self) -> Any: ...
+
+    @property
+    def float64(self) -> Any: ...
+
+
+def is_array_api_creation_namespace(obj: Any) -> TypeGuard[ArrayCreationNamespace]:
```
egparedes: I still think it would be probably better to get rid of this file (or most of it) and just use  https://data-apis.org/array-api-compat/ (maybe even combined with https://github.com/34j/types-array-api to get rid of the clumsy `NDArrayObject`...)

src/gt4py/_core/ndarray_utils.py
```diff
@@ -34,3 +141,40 @@ def array_namespace(array: core_defs.NDArrayObject) -> types.ModuleType:
...
+        return None
+    if device.device_type == core_defs.DeviceType.CPU:
+        return None  # or Literal['cpu']
+    raise ValueError(f"NumPy does not support device type {device.device_type}.")
+
+
+# TODO(egparedes): Replace by single dispatch on types (instead of instances)
+_device_translation_registry: dict[ArrayCreationNamespace, Callable[[core_defs.Device], Any]] = {
+    cast(ArrayCreationNamespace, np): _numpy_device_translator
+}
```
egparedes, line 122: I just learnt (or re-learnt) about [`__array_namespace_info__`](https://data-apis.org/array-api/2023.12/API_specification/inspection.html#inspection) which I think could help here

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,370 @@
...
+# - translation of dtype/device to target namespace
+# - absolute (domain) to relative (shape), includes aligned index
+# - backing array api OR FieldBufferAllocator
+
+
+# TODO Introduce a protocol and keep the concrete implementation for NDArray-based allocation (both array api and fieldbufferallocator) in a separate class?
+# TODO maybe FieldAllocationNamespace can evolve to FieldNamespace (convering also all Field operations?)
+
+# Type to be used by the end-user
+FieldAllocator: TypeAlias = (
```
egparedes: Maybe ?
```suggestion
FieldBufferAllocator: TypeAlias = (
```

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,370 @@
...
+    core_ndarray_utils.ArrayCreationNamespace | next_allocators.FieldBufferAllocationUtil
+)
+
+# TODO make all other types private? check usage in ICON4Py
+
+
+class FieldAllocationNamespace(abc.ABC):
```
egparedes: Maybe
```suggestion
class FieldCreationNamespace(abc.ABC):
```
or
```suggestion
class FieldConstructorsNamespace(abc.ABC):
```

src/gt4py/_core/ndarray_utils.py
```diff
@@ -34,3 +141,40 @@ def array_namespace(array: core_defs.NDArrayObject) -> types.ModuleType:
...
+            raise ValueError(
+                f"CuPy only supports GPU devices, got device type {device.device_type}."
+            )
+        return cupy.cuda.Device(device.device_id)
+
+    _device_translation_registry[cupy] = _cupy_device_translator
+
+
+@functools.cache
+def get_device_translator(
```
egparedes: Same comment as above about `__array_namespace_info__`

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,370 @@
...
+        assert dtype is not None  # TODO check where to put the default
+        res = common._field(self._empty(domain, dtype=dtype), domain=domain)
+        assert isinstance(res, nd_array_field.NdArrayField)  # for typing
+        return res
+
+    @abc.abstractmethod
+    def _empty(
+        self, domain: common.Domain, *, dtype: core_defs.DType
+    ) -> core_defs.NDArrayObject: ...
+
```
egparedes: So it looks like `FieldCreationNamespace` is just fully implemented and just relies on another concept: `FieldNDArrayCreationNamespace` (implementing all the underscored `_creation_function`s to be provided at initialization, right?

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,370 @@
...
+        dtype: core_defs.DType,
+        copy: bool | None,
+    ) -> core_defs.NDArrayObject: ...
+
+
+_ANS = TypeVar("_ANS", bound=core_ndarray_utils.ArrayCreationNamespace)
+
+
+@dataclasses.dataclass(frozen=True)
+class ArrayNamespaceWrapper(FieldAllocationNamespace, Generic[_ANS]):
```
egparedes: This could just be one implementation of the new `FieldNDArrayCreationNamespace` concept mentioned above...

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,370 @@
...
+            data, dtype=self._to_array_ns_dtype(dtype), device=self.device, copy=copy
+        )
+        assert domain.shape == arr.shape, (
+            f"pre-condition of `_asarray` not met: {data.shape=} and {domain.shape} need to agree."
+        )
+        return arr
+
+
+@dataclasses.dataclass(frozen=True)
+class FieldBufferAllocatorWrapper(FieldAllocationNamespace):
```
egparedes: And this other other too...

src/gt4py/next/constructors.py
havogt: I would prefer to keep this name short because it's the only type the user should need. To me the name `FieldAllocator` makes sense because it's used to allocate a field. In practice it's only concerned with allocating the buffer, but the user doesn't need to know that to work with it.

src/gt4py/next/constructors.py
havogt:
```suggestion
class FieldConstructionNamespace(abc.ABC):
```
to throw in a 3rd option. All 3 I prefer over the current one. `FieldCreationNamespace` less than the other 2 because it doesn't match the module name.

src/gt4py/next/constructors.py
havogt: yes

src/gt4py/next/constructors.py
havogt: yes it is already, right? the concepts that you describe exist, it's just how we implement them.

src/gt4py/_core/definitions.py
havogt: or use `bool` ourselves? But we probably have to move it to a separate file then to not shadow `bool` in this whole module which is also used here.

src/gt4py/_core/ndarray_utils.py
havogt: I checked both libraries again:
- I think we should use array-api-compat for strict checking of what we support in embedded, but that's a different PR. We could use it already to replace the `array_namespace` function in this file. No strong opinion, we can change it here, but maybe I would prefer to introduce it for the other use-case.
- https://github.com/34j/types-array-api cannot be used now, because it's python 3.12+ (although the pyproject.toml doesn't reflect that). 

Maybe the Protocol that I added here is overkill, I can simplify or completly remove it.

src/gt4py/_core/ndarray_utils.py
havogt: Thanks, didn't know it exists. I don't think it helps here. If I remember correctly, you also didn't check how it could help?!

src/gt4py/next/constructors.py
havogt: I split them

tests/next_tests/integration_tests/feature_tests/ffront_tests/ffront_test_utils.py
```diff
@@ -415,7 +419,7 @@ def skip_value_mesh(allocator) -> MeshDescriptor:
         simple_mesh,
         pytest.param(skip_value_mesh, marks=pytest.mark.uses_mesh_with_skip_values),
     ],
-    ids=lambda p: p(None).name,
+    ids=lambda p: p.__name__,  # TODO previously it is allocating the mesh and then throws it away...
```
havogt: TODO

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,375 @@
...
+
+        shape = storage_utils.asarray(
+            data
+        ).shape  # TODO this is a quite expensive throw-away operation...
+        if shape != actual_domain.shape:
+            raise ValueError(f"Cannot construct 'Field' from array of shape '{shape}'.")
+
+        if dtype is None:  # TODO does this make sense?
+            dtype = core_defs.dtype(storage_utils.asarray(data).dtype)
+
```
havogt: Not sure what to do about this. The `storage_utils.asarray` is only there in case we are passing something that is not array but convertible with `asarray`. The current typing (and already before the PR) required `NDArrayObject`. If we want to keep that we can drop the `asarray` calls here, because shape and dtype is already defined.

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,375 @@
...
+        copy: bool | None = None,
+    ) -> core_defs.NDArrayObject:
+        if copy is not None:
+            # TODO(egparedes): allow zero-copy construction (no reallocation) if buffer has
+            #   already the correct layout and device.
+            raise NotImplementedError("The 'copy' argument is not yet supported.")
+        assert dtype is not None
+        arr = self.empty(domain, dtype=dtype)
+
+        arr[...] = core_ndarray_utils.array_namespace(arr).asarray(data)  # type: ignore[index] # # `NDArrayObject` typing is not complete
```
havogt:
```suggestion
        arr[...] = core_ndarray_utils.array_namespace(arr).asarray(data)  # type: ignore[index] # `NDArrayObject` typing is not complete
```

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,375 @@
...
+        domain: common.Domain,
+        data: core_defs.NDArrayObject,
+        *,
+        dtype: core_defs.DType,
+        copy: bool | None = None,
+    ) -> core_defs.NDArrayObject:
+        if copy is not None:
+            # TODO(egparedes): allow zero-copy construction (no reallocation) if buffer has
+            #   already the correct layout and device.
+            raise NotImplementedError("The 'copy' argument is not yet supported.")
```
havogt: we briefly touched on this. Should we remove view support? I think the conclusion was we shouldn't allow it, but for rare cases the user can work around it, but not with public api.

src/gt4py/_core/ndarray_utils.py
egparedes: Ok, let's introduce array-api-compat in a later PR and leave this protocol here for now mostly as documentation for ourselves

src/gt4py/_core/ndarray_utils.py
```diff
@@ -34,3 +141,40 @@ def array_namespace(array: core_defs.NDArrayObject) -> types.ModuleType:
...
+
+def _numpy_device_translator(device: core_defs.Device | None) -> Any:
+    if device is None:
+        return None
+    if device.device_type == core_defs.DeviceType.CPU:
+        return None  # or Literal['cpu']
+    raise ValueError(f"NumPy does not support device type {device.device_type}.")
+
+
+# TODO(egparedes): Replace by single dispatch on types (instead of instances)
```
havogt:
```suggestion
# TODO(egparedes): Replace by single dispatch on `is`-equality.
```
Want to do this in this PR?

src/gt4py/_core/ndarray_utils.py
egparedes: I thought it could for the translation of the device types, did you see a way how to use it or still seems useless to you?

tests/next_tests/integration_tests/feature_tests/ffront_tests/ffront_test_utils.py
havogt: Should I keep this as is? For the current meshes it's fine...

src/gt4py/next/constructors.py
egparedes: Not 100% sure `FieldAllocator` is the way to go but I agree that `FieldBufferAllocator` seems cumbersome so let's keep `FieldAllocator` for now if we don't find something better

src/gt4py/next/constructors.py
```diff
@@ -234,64 +579,24 @@ def as_field(
...
 def as_connectivity(
     domain: common.DomainLike | Sequence[common.Dimension],
     codomain: common.Dimension,
     data: core_defs.NDArrayObject,
-    dtype: Optional[core_defs.DType] = None,
+    dtype: core_defs.DType | None = None,
     *,
-    allocator: Optional[next_allocators.FieldBufferAllocationUtil] = None,
-    device: Optional[core_defs.Device] = None,
+    # TODO what about origin?
```
havogt: I think we should add origin here, right?

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,375 @@
...
+            #   already the correct layout and device.
+            raise NotImplementedError("The 'copy' argument is not yet supported.")
+        assert dtype is not None
+        arr = self.empty(domain, dtype=dtype)
+
+        arr[...] = core_ndarray_utils.array_namespace(arr).asarray(data)  # type: ignore[index] # # `NDArrayObject` typing is not complete
+        return arr
+
+
+# TODO test all code paths
```
havogt: Can we go through this function together to discuss the defaults.

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,375 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class FieldConstructionNamespace:
+    """
+    Public-facing field construction API.
+
+    Delegates array creation to the composed `_array_constructor` and wraps the result in a Field.
+    """
+
+    _array_constructor: _ArrayCreationNamespace
```
egparedes: Shouldn' t we try to align these 2 names?

havogt: Yes, I'll push a change. Didn't realize that I have `_ArrayCreationNamespace` and the protocol `ArrayCreationNamespace`.

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,375 @@
...
+
+# Type to be used by the end-user
+FieldAllocator: TypeAlias = (
+    core_ndarray_utils.ArrayCreationNamespace | next_allocators.FieldBufferAllocationUtil
+)
+
+# TODO make all other types private? check usage in ICON4Py
+
+
+class _ArrayCreationNamespace(abc.ABC):
```
egparedes: I would move this class definition down, to put it together with its concrete implementations

havogt: yes

src/gt4py/next/constructors.py
havogt: I think I drop the `ArrayCreationNamespace` and just keep `ArrayNamespace` with the info that it's incomplete (unless we come to the conclusion we should/can get proper typing from somewhere).

src/gt4py/next/constructors.py
egparedes: Yes, I think we should drop it. The user should take himself to build a ndarray before calling a field constructor

src/gt4py/_core/ndarray_utils.py
havogt: I didn't see a way how to use, did you?

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,375 @@
...
+        domain: common.DomainLike,
+        dtype: core_defs.DTypeLike | None = None,
+    ) -> nd_array_field.NdArrayField:
+        domain = common.domain(domain)
+        dtype = core_defs.dtype(dtype)
+        assert dtype is not None
+        res = common._field(self._array_constructor.ones(domain, dtype=dtype), domain=domain)
+        assert isinstance(res, nd_array_field.NdArrayField)  # for typing
+        return res
+
```
egparedes, line 153: Would it be too convoluted to concentrate the boiler plate around the creation in one single function and then call it from the specific constructor? I'm thinking of something along these lines:

```python
def construct_from_array(self, domain: DomainLike, dtype: DTypeLike, ndarray_constructor: callable):
    domain = common.domain(domain)
    dtype = core_defs.dtype(dtype)
    res = common._field(ndarray_constructor(domain=domain, dtype=type), domain=domain)
    assert isinstance(res, NDArrayField)
    return res

def ones(
    self,
    domain: common.DomainLike,
    dtype: core_defs.DTypeLike | None = None,
) -> nd_array_field.NdArrayField:
    return construct_from_array(
        domain=domain, dtype=dtype, ndarray_constructor=self._array_constructor.ones,
    )

def full(
    self,
    domain: common.DomainLike,
    fill_value: core_defs.Scalar,
    dtype: core_defs.DTypeLike | None = None
):
    return construct_from_array(
        domain=domain,
        dtype=dtype or core_defs.dtype(type(fill_value)),
        ndarray_constructor=functools.partial(self._array_constructor.full, fill_value=fill_value)),
    )
```

src/gt4py/next/constructors.py
egparedes: Fine for me, but just to be clear, if we remove it we would need to change the signature of this function to remove the argument, right?

havogt: Yes, it was not there before this PR. I removed it from the public facing interface. Will clean this up as well I think.

src/gt4py/_core/definitions.py
havogt: separate file

src/gt4py/next/constructors.py
havogt: `Allocator`

src/gt4py/next/constructors.py
havogt: copy removed

havogt: add description why copy not supported: we don't want aliasing

src/gt4py/_core/types.py
```diff
@@ -0,0 +1,44 @@
...
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Extracted from definitions.py to avoid `bool` to shadow the python built-in `bool` type in that file.
+
+import numpy as np
+
+
+bool = np.bool  # noqa: A001
```
egparedes: Question: should we keep the old `bool_` name as an alias for backward compatibility?

src/gt4py/_core/ndarray_utils.py
```diff
@@ -19,13 +21,94 @@
...
 
 
-def array_namespace(array: core_defs.NDArrayObject) -> types.ModuleType:
+class ArrayNamespace(Hashable, Protocol):
+    """
+    Currently only a subset of the Array API standard namespace with functions relevant for array creation.
+
+    TODO(havogt): replace by ... or make it more complete and put next to NDArrayObject definition.
+
+    See also 'ArrayNamespace'.
```
egparedes:
```suggestion
```

src/gt4py/_core/types.py
havogt:
```suggestion
bool = np.bool_  # noqa: A001
```
for numpy 1.x compat

src/gt4py/_core/ndarray_utils.py
```diff
@@ -19,13 +21,94 @@
...
+    int8: type
+    int16: type
+    int32: type
+    int64: type
+    uint8: type
+    uint16: type
+    uint32: type
+    uint64: type
+    float32: type
+    float64: type
```
egparedes: Question: are we sure we need these to be types? I mean, if the Array API doesn't required them to be types, it might be enough to require that they are callables, right?

src/gt4py/_core/types.py
```diff
@@ -0,0 +1,45 @@
...
+
+uint8 = np.uint8
+uint16 = np.uint16
+uint32 = np.uint32
+uint64 = np.uint64
+
+float32 = np.float32
+float64 = np.float64
+
+type_to_name = {
```
havogt, line 32:
```suggestion
# This is used to map between dtypes of different array namespaces, as only the name is defined in the array API.
# We don't use `<type>.__name__` for the mapping to be on the safe side.
type_to_name = {
```

src/gt4py/next/custom_layout_allocators.py
```diff
@@ -6,6 +6,8 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
+# TODO custom_layout_allocators ?
```
havogt: forgot about that one...

src/gt4py/next/custom_layout_allocators.py
```diff
@@ -83,50 +85,34 @@ def is_field_allocator_factory_for(
...
-        return obj
-    elif is_field_allocator_factory(obj):
-        return obj.__gt_allocator__
-    elif not strict or is_field_allocator(default):
-        return default
-    else:
-        raise TypeError(
-            f"Object '{obj}' is neither a field allocator nor a field allocator factory."
-        )
+# TODO do we have this function somewhere? maybe `_get_slices_from_domain_slice`
```
havogt: and that one :(

src/gt4py/_core/ndarray_utils.py
```diff
@@ -19,13 +21,94 @@
...
+        and hasattr(obj, "int32")
+        and hasattr(obj, "int64")
+        and hasattr(obj, "uint8")
+        and hasattr(obj, "uint16")
+        and hasattr(obj, "uint32")
+        and hasattr(obj, "uint64")
+        and hasattr(obj, "float32")
+        and hasattr(obj, "float64")
+    )
+
```
egparedes, line 105: I'm not sure if it works, but I'd rather have a more automated implementation to avoid that the protocol and the typeguard get out of sync.

```python
_array_namespace_members: Final = tuple(vars(ArrayNamespace).keys())

# and then:
def is_array_namespace(obj: Any) -> TypeGuard[ArrayNamespace]:
    return all(hasattr(name) for name in _array_namespace_members) 
```

src/gt4py/_core/ndarray_utils.py
```diff
@@ -34,3 +117,47 @@ def array_namespace(array: core_defs.NDArrayObject) -> types.ModuleType:
...
+def _numpy_device_translator(device: core_defs.Device | None) -> Any:
+    if device is None:
+        return None
+    if device.device_type == core_defs.DeviceType.CPU:
+        return None  # or Literal['cpu']
+    raise ValueError(f"NumPy does not support device type {device.device_type}.")
+
+
+# Currently private as we only support a concrete set of array namespaces.
+_device_translation_registry: dict[ArrayNamespace, Callable[[core_defs.Device], Any]] = {
```
egparedes: Maybe add a TODO mentioning a future PR using array_api_compat...

src/gt4py/_core/ndarray_utils.py
```diff
@@ -34,3 +117,47 @@ def array_namespace(array: core_defs.NDArrayObject) -> types.ModuleType:
...
+
+
+@functools.cache
+def get_device_translator(array_ns: ArrayNamespace) -> Callable[[core_defs.Device], Any]:
+    """
+    Returns a mapping from a GT4Py 'Device' to the corresponding device object for the given array namespace.
+    """
+    if array_ns not in _device_translation_registry:
+        raise ValueError(f"No device translator registered for array namespace {array_ns}.")
+    return _device_translation_registry[array_ns]
```
egparedes: I don't think the `functools.cache` brings any speedup here, specially if rewritten with a try-except:
```suggestion
def get_device_translator(array_ns: ArrayNamespace) -> Callable[[core_defs.Device], Any]:
    """
    Returns a mapping from a GT4Py 'Device' to the corresponding device object for the given array namespace.
    """
    try:
        return _device_translation_registry[array_ns]
    except KeyError:
        raise ValueError(f"No device translator registered for array namespace {array_ns}.")
```

src/gt4py/next/custom_layout_allocators.py
egparedes: Is this TODO still relevant? Should it be solved in this PR?

src/gt4py/next/custom_layout_allocators.py
egparedes, line 1: Should the module be renamed to something less generic (e.g. `buffer_allocators`)?

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,390 @@
...
+from gt4py.eve import extended_typing as xtyping
+
+
+"""
+Field construction API.
+
+The user-facing API consists of the functions 'empty', 'zeros', 'ones', 'full' and 'as_field',
+or the 'FieldConstructor' class for more advanced use cases.
+
+These functions create GT4Py 'Field's backed by arrays created using a specified allocator, which can be either an array namespace
```
egparedes:
```suggestion
These functions create GT4Py Field's backed by arrays (NDArrayField) created using a specified allocator, which can be either an array namespace
```

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,390 @@
...
+(e.g. 'numpy', 'cupy') or a GT4Py field buffer allocator (e.g. a backend).
+
+This module deals with 3 concepts:
+- allocating with a backing array API or 'FieldBufferAllocator'
+- translating from absolute (domain) to relative (shape) indexing, e.g. for handling of 'aligned_index'
+- translating from GT4Py dtypes and devices to the format expected by the backing array API.
+"""
+
+# Type to be used by the end-user
+Allocator: TypeAlias = core_ndarray_utils.ArrayNamespace | next_allocators.FieldBufferAllocationUtil
```
egparedes, line 48: Question: do we still like the name `FieldBufferAllocationUtil`? Maybe `FieldBufferAllocator` would be more consistent with the rest of the API and I think this PR is the right moment to rename it

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,390 @@
...
+
+Accepts either:
+- An array namespace following the Array API standard (e.g. ``numpy``, ``cupy``, ``jax.numpy``),
+  which will be used directly for array creation.
+- A GT4Py field buffer allocator or allocator factory (e.g. a backend), which controls
+  memory layout and alignment.
+"""
+
+DEFAULT_DEVICE: core_defs.Device = core_defs.Device(core_defs.DeviceType.CPU, 0)
+DEFAULT_DTYPE: core_defs.DType = core_defs.Float64DType(())
```
egparedes:
```suggestion
DEFAULT_DEVICE: Final[core_defs.Device] = core_defs.Device(core_defs.DeviceType.CPU, 0)
DEFAULT_DTYPE: Final[core_defs.DType] = core_defs.Float64DType(())
```

src/gt4py/next/custom_layout_allocators.py
havogt: I think it's not exactly the same function (e.g. passing namedindices that are not represented in the domain doesn't make sense in the other context).

havogt: on the other hand I could extend the other function with a "strict" mode.

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,390 @@
...
+                device=device,
+                aligned_index=aligned_index,
+            )
+        else:
+            raise ValueError(f"Invalid field allocator: {allocator}.")
+
+    class _NdArrayConstructor(Protocol):
+        def __call__(
+            self, domain: common.Domain, *, dtype: core_defs.DType
+        ) -> core_defs.NDArrayObject: ...
```
egparedes: Two comments:
- why do we need to define this protocol inside the class? I think it's a bit weird and I'd move it outside the class definition.
- I've just realized we used both `NdArraySomething` (e.g. NDArrayObject) and `NdArraySomething` (e.g. NdArrayField) in our codebase. I always try to follow PEP 8 guidelines on names with acronyms (https://peps.python.org/pep-0008/#descriptive-naming-styles) but I'm aware many people dislikes it, including you I think. In any case, I think we should just pick one and use it consistently

src/gt4py/next/custom_layout_allocators.py
havogt: not now...

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,390 @@
...
+            dtype = core_defs.dtype(data.dtype)
+
+        return self._construct_from_array(
+            actual_domain,
+            dtype,
+            lambda domain, dtype: self._array_constructor.asarray(domain, data, dtype=dtype),
+        )
+
+
+class _FieldArrayConstructionNamespace(abc.ABC):
```
egparedes: I dislike this name but I'm not sure I can come up with a better one..

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,390 @@
...
+        *,
+        dtype: core_defs.DType,
+    ) -> core_defs.NDArrayObject: ...
+
+
+_ANS = TypeVar("_ANS", bound=core_ndarray_utils.ArrayNamespace)
+
+
+@dataclasses.dataclass(frozen=True)
+class _ArrayApiCreationNamespace(_FieldArrayConstructionNamespace, Generic[_ANS]):
```
egparedes: Similar comment as above about acronyms in names. I think it should be:
```suggestion
class _ArrayAPICreationNamespace(_FieldArrayConstructionNamespace, Generic[_ANS]):
```

src/gt4py/_core/types.py
```diff
@@ -0,0 +1,44 @@
...
+        "int32",
+        "int64",
+        "uint8",
+        "uint16",
+        "uint32",
+        "uint64",
+        "float32",
+        "float64",
+    )
+}
```
egparedes, line 47: Maybe:
```suggestion
type_to_name = {
    globals()[name]: name
    for name in (
        "bool",
        "int8",
        "int16",
        "int32",
        "int64",
        "uint8",
        "uint16",
        "uint32",
        "uint64",
        "float32",
        "float64",
    )
}
```

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,390 @@
...
+        *,
+        dtype: core_defs.DType,
+    ) -> core_defs.NDArrayObject:
+        arr = self.empty(domain, dtype=dtype)
+
+        arr[...] = core_ndarray_utils.array_namespace(arr).asarray(data)  # type: ignore[index] # `NDArrayObject` typing is not complete
+        return arr
+
+
+@eve.utils.optional_lru_cache
```
egparedes, line 384: Document why `optional_lru_cache` is used here

tests/next_tests/unit_tests/test_constructors.py
```diff
@@ -20,35 +30,81 @@
...
-# TODO: parametrize with gpu backend and compare with cupy array
-@pytest.mark.parametrize(
-    "allocator, device",
-    [
-        [next_allocators.StandardCPUFieldBufferAllocator(), None],
-        [None, core_defs.Device(core_defs.DeviceType.CPU, 0)],
-    ],
+
+@dataclasses.dataclass
+class ConstructureTestSetup:
```
egparedes:
```suggestion
class ConstructorTestSetup:
```

tests/next_tests/unit_tests/test_constructors.py
```diff
@@ -160,3 +211,97 @@ def test_aligned_index():
 def test_as_connectivity(nd_array_implementation, data, skip_value):
     testee = gtx.as_connectivity([I], J, nd_array_implementation.array(data))
     assert testee.skip_value is skip_value
+
+
+# --- FieldConstructor.__init__ codepath tests ---
```
egparedes:
```suggestion
```

egparedes (changes requested):
Overall looks good. I just have some  comments and questions about style and naming

src/gt4py/_core/types.py
havogt: it's a private type anyway. cartesian didn't export it, and I'd say we just fix our codes. This pr breaks next API anyway.

src/gt4py/_core/ndarray_utils.py
havogt: how? There is the to_device functions but no (public) functions that do the conversion. Or am I missing somehting?

src/gt4py/next/custom_layout_allocators.py
havogt: removed in the meantime

src/gt4py/next/custom_layout_allocators.py
havogt: in the meantime was renamed to custom_layout_allocators as previosuly discussed.

src/gt4py/next/constructors.py
havogt:
```suggestion
These functions create GT4Py 'Field's backed by arrays (NDArrayField) created using a specified allocator, which can be either an array namespace
```
removing the `'` is wrong.

src/gt4py/next/constructors.py
havogt: Should I keep it at all? My friend generated it, but maybe I should just use `Callable` instead of a concrete hing, because it's very internal. Because it's very internal I also didn't complain to my friend to where it put it.

src/gt4py/_core/ndarray_utils.py
egparedes: Ok, maybe I got confused again about what `array_api_compat` will give us.

src/gt4py/next/constructors.py
egparedes: Ah, yes, you're right about the `'`.  I misread it

src/gt4py/next/constructors.py
egparedes: Fine for me to remove it.

src/gt4py/_core/ndarray_utils.py
havogt: Why should they be callable actually? I think the only requirement is that they implement `__eq__()`, see https://data-apis.org/array-api/latest/API_specification/data_types.html

src/gt4py/_core/ndarray_utils.py
havogt: Yeah, probably a leftover of different ideas

src/gt4py/next/constructors.py
havogt: I played with the names, I think it's better now.

src/gt4py/next/constructors.py
havogt: I guess if we want to change it here, we need to talk again, because it will trigger other naming changes.

src/gt4py/_core/ndarray_utils.py
havogt: I introduced _EqualityComparable

src/gt4py/_core/ndarray_utils.py
egparedes: I think this is the correct version now:

```python
_array_namespace_members: frozenset[str]

if sys.version < (3, 13):
    _array_namespace_members = frozenset(typing._get_protocol_attrs(ArrayNamespace))
else:
    _array_namespace_members = typing.get_protocol_members(ArrayNamespace))

def is_array_namespace(obj: Any) -> TypeGuard[ArrayNamespace]:
    return all(hasattr(name) for name in _array_namespace_members) 
```

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,387 @@
...
+from gt4py.eve import extended_typing as xtyping
+
+
+"""
+Field construction API.
+
+The user-facing API consists of the functions 'empty', 'zeros', 'ones', 'full' and 'as_field',
+or the 'FieldConstructor' class for more advanced use cases.
+
+These functions create GT4Py 'Field's backed by arrays (NDArrayField) created using a specified allocator, which can be either an array namespace
```
egparedes:
```suggestion
These functions create GT4Py 'Field's backed by arrays (NDArrayField)
created using a specified allocator, which can be either an array namespace
```

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,387 @@
...
+
+This module deals with 3 concepts:
+- allocating with a backing array API or 'FieldBufferAllocator'
+- translating from absolute (domain) to relative (shape) indexing, e.g. for handling of 'aligned_index'
+- translating from GT4Py dtypes and devices to the format expected by the backing array API.
+"""
+
+# Type to be used by the end-user
+Allocator: TypeAlias = core_ndarray_utils.ArrayNamespace | next_allocators.FieldBufferAllocationUtil
+"""Type for field memory allocators.
```
egparedes:
```suggestion
"""
Type for field memory allocators.
```

src/gt4py/next/constructors.py
```diff
@@ -8,26 +8,387 @@
...
+    def asarray(
+        self,
+        domain: common.Domain,
+        data: core_defs.NDArrayObject,
+        *,
+        dtype: core_defs.DType,
+    ) -> core_defs.NDArrayObject: ...
+
+
+_ANS = TypeVar("_ANS", bound=core_ndarray_utils.ArrayNamespace)
```
egparedes: Just a suggestion, feel free to ignore
```suggestion
_ArrayNST = TypeVar("_ANS", bound=core_ndarray_utils.ArrayNamespace)
```

src/gt4py/_core/ndarray_utils.py
havogt: As discussed: the above needs the pre-array-api `np/cp.bool_` workaround, therefore we decided to keep as is and then use array_api_compat in a next step.

PR #2444, opened 2026-01-18
fix[next-dace]: Remove isolated node generated from splitting

edopao:
This PR fixes an issue observed in icon4py stencil `compute_rho_theta_pgrad_and_update_vn`, which resulted in an isolated access node to global in/out data. Test case with reproducer was also added.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
@@ -1000,7 +1000,9 @@ def _perform_node_split_with_bypass_impl(
     #   dataflow might be in some invalid state.
     state.remove_edge(producer_edge)
     if state.degree(producer_edge.src) == 0:
+        assert len(new_consumer_edges) == 0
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
```diff
@@ -999,6 +999,8 @@ def _perform_node_split_with_bypass_impl(
     # TODO(phimuell): Find a way to avoid doing the propagation here, where the
     #   dataflow might be in some invalid state.
     state.remove_edge(producer_edge)
+    if state.degree(producer_edge.src) == 0:
+        state.remove_node(producer_edge.src)
```
philip-paul-mueller, line 1004: It did not occured to me yesterday, but if the source has become isolated it means that no fragments were created.
This simply adds a check and an early exit, I am kind of okay to drop the early exit, but the `assert` is important.
```suggestion
        assert len(new_consumer_edges) == 0
        state.remove_node(producer_edge.src)
        return new_consumer_edges
```

PR #2483, opened 2026-02-17
feat[next]: Precompilation with static domains

SF-N:
Adds support for precompilation of programs with static domains. 
Previously, `static_domains=True` was only usable in the JIT path, where field domains are available from the runtime arguments.
This change lets callers provide the static field domains explicitly during `compile()`, allowing the precompile path to construct the same `FieldDomainDescriptor` inputs that JIT compilation would normally derive from runtime fields.   
 `static_domains` is specified per field argument: 

```python                                                                                                                                                               
     program.compile(                                                                                                                                                        
         ...,                                                                                                                                                                
         static_domains={                                                                                                                                                    
             "inp": {IDim: (0, nx), JDim: (0, ny)},                                                                                                                          
             "out": {IDim: (1, nx - 1), JDim: (1, ny - 1)},                                                                                                                  
         },                                                                                                                                                                  
     )  
```
Tuple field arguments use indexed field expressions:                                                                                                                    
```python
     static_domains={                                                                                                                                                        
         "out[0]": {IDim: (0, nx)},                                                                                                                                          
         "out[1]": {IDim: (0, nx)},                                                                                                                                          
     }  
```
Note that a consequence of this syntax is that it is not possible to have a parameter called `static_domains` and pass a static value to it, as `static_domains` is now a reserved keyword argument to `compile()`.


havogt:
Just dropping a few ideas to discuss:
- I think tuples should just use normal tuple syntax, i.e. `"out": ({IDim: (0, nx)}, {IDim: (0, nx)})`
- Not sure what to do in case of NamedCollections, maybe "out": {"x": {IDim: (0, nx)}, "y": {IDim: (0, nx)}

Just random thought:
class Point(NamedTuple):
    x: Field
    y: Field
    
"out": Domain[Point](x=domain1, y=domain2)

havogt:
Feedback from a friend for discussion

>static_domains=True also requires every scalar a domain depends on to be static. _compute_rho_theta_pgrad_and_update_vn derives concat_where domains from start_edge_*/end_edge_*/nflatlev; >when those stay runtime, domain inference asserts in domain_utils._unstructured_translate_range_statically (assert isinstance(start_expr, itir.Literal)) instead of falling back to the max-domain path the @program form used. So compile-time domains are all-or-nothing with the scalars that shape them — I now gate static_domains on the compile_time_domain variant. A corollary: the compile_time_vertical variant (vertical bounds static, horizontal dynamic) is not expressible for a direct call, since static_domains is a single per-program flag.

havogt:
And do we actually need to provide input sizes as well? Why?

PR #2531, opened 2026-03-13
feat[next-dace]: Updated `MoveDataflowIntoIfBody`

philip-paul-mueller:
This PR updates `MoveDataflowIntoIfBody` transformation.
It addresses some aspects that were not fully addressed in a [previous PR](https://github.com/GridTools/gt4py/pull/2514). The main change is that now the processing is not done in a per connector fashion but is performed in a unified way, i.e. all at once.
Furthermore, the stability was increased, by ordering the nodes to relocate according to their node id.

PR #2534, opened 2026-03-13
Sort DaCe nodes for deterministic code generation

kotsaloscv:


src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
```diff
@@ -1007,8 +1007,15 @@ def apply(
 
         # This is the tasklet that we will put inside the map, note we have to do it
         #  this way to avoid some name clash stuff.
+        # Use a deterministic hash instead of uuid.uuid1() to ensure stable code
+        # generation across runs. The hash combines properties that are unique to
+        # this specific clone context.
+        _clone_key = (
+            f"{tasklet.label}_{tasklet.code.as_string}_{map_entry.label}_{access_node.data}"
```
philip-paul-mueller:
```suggestion
            f"{tasklet.label}_{tasklet.code.as_string}_{map_entry.label}_{connector_name}_{access_node.data}"
```
`clone_key` is not necessarily unique, as the transformation allows that there are multiple connections between the AccessNode and the Map.
I remember that the transformation was explicitly designed to allow this kind of situation, but I am not sure if it is used.
By also using the connector name the key should become unique.

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -501,7 +512,7 @@ def _remove_outside_dataflow(
         # Before we can clean the original nodes, we must clean the dataflow. If a
         #  node, that was relocated, has incoming connections we must remove them
         #  and the parent dataflow.
-        for node_to_remove in all_relocatable_dataflow:
+        for node_to_remove in sorted(all_relocatable_dataflow, key=_node_sort_key):
```
philip-paul-mueller: Just for my curiosity:
I am a bit puzzled that one needs to process the nodes here in sorted order.
Since you only remove objects and does not change anything else, I would have expected that it is irrelevant in which order you do it.
Have you tried to run it without the sorting here?

philip-paul-mueller (commented):
Thanks for the changes, I have some small comment and a question.

There is also [PR#2531](https://github.com/GridTools/gt4py/pull/2531), which modifies `MoveDataflowIntoIfBody` in a similar way, but introduces some bug fixes and other improvements.
Nevertheless, I would propose to merge the changes in this PR, as they stabilize the code, and the other PR will not ready as fast as this one.

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
kotsaloscv: You are right! I will revert it with all my other changes ;)

PR #2537, opened 2026-03-17
fix[next]: Include compile-time arguments in fingerprint

SF-N:
This pull request refactors the way unique hashes are generated for compilable programs and adds a new unit test to ensure hash uniqueness for different precision arguments. 

* The `fingerprint_compilable_program` function in `src/gt4py/next/otf/stages.py` now generates the hash using the entire `args` object instead of individually hashing specific components like `offset_provider` and `column_axis`.

* A new unit test, `test_different_precisions`, was added to verify that changing the precision of a scalar argument results in a different fingerprint hash.


tehrengruber:
Fixed by #2648

PR #2565, opened 2026-04-02
build[next-dace]: Smaller DaCe Folder

philip-paul-mueller:
Introduces the reduced DaCe folders from [DaCe PR#2331](https://github.com/spcl/dace/pull/2331) into GT4Py.
In in DEBUG mode the `development` mode will be used otherwise the `production` mode is used.
To override this one can export the environment variable `DACE_compiler_build_folder_mode` and set it either to `production` or `development`.

Note:
In `production` mode the SDFG is not stored to disc, however, to create a `CompiledSDFG` one needs the SDFG (at least parts of it). This means that the generated folders are not self contained. This is not a problem for GT4Py as the SDFGs are stored inside the GT4Py cache, but for general DaCe.


havogt:
I tried it. Nice improvement less than 1/10th in number of files and ~ 1/20th in filesize.
E.g. r2b5 single rank cache from  440M -> 24M, 4410 files -> 306 files.

src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -132,6 +132,11 @@ def set_dace_config(
         "progress", value=gtx_config.env_flag_to_bool("DACE_progress", default=gtx_config.DEBUG)
     )
 
+    # In debug mode we generate the full folder version otherwise we use the production version.
+    dace.Config.set(
+        "compiler", "build_folder_version", value=("full" if gtx_config.DEBUG else "production")
```
edopao: On latest dace main:
```suggestion
        "compiler", "build_folder_mode", value=("development" if gtx_config.DEBUG else "production")
```

philip-paul-mueller:
@edopao I am actually not sure if we should switch to DaCe main or stay on our fork, but I am not particular opinionated about that.

edopao:
> @edopao I am actually not sure if we should switch to DaCe main or stay on our fork, but I am not particular opinionated about that.

This was just for testing. We need to wait for a weekly release from DaCe main, then we can update this PR and merge it.

philip-paul-mueller:
I have now changed it back such that it follows the GT4Py DaCe repo with the newest release Ioannis recently (2026-04-20) created.

src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -135,6 +135,11 @@ def set_dace_config(
         "progress", value=gtx_config.env_flag_to_bool("DACE_progress", default=gtx_config.DEBUG)
     )
 
+    # In debug mode we generate all files otherwise we use the production folder mode.
+    dace.Config.set(
+        "compiler", "build_folder_mode", value=("development" if gtx_config.DEBUG else "production")
```
iomaganaris: As far as I understood the source files and SDFG is not saved in the `production` mode, right?
I would like to have an extra option for `GT4Py` like `RelWithDebInfo` in `CMake` for that because if `gtx_config.DEBUG` is set, then the `CMAKE_BUILD_TYPE` is also set to `Debug`

philip-paul-mueller: I think I see the use case, I have to think about how to expose that.

philip-paul-mueller: I solved it by adding an additional configuration open.
The behaviour is now similar to the lifetime of the cache.
I am not super happy, because it is a rather DaCe specific option and there should be some GT4Py option that hides it.

philip-paul-mueller: However, an alternative that would work even without this new addition is to export `DACE_compiler_build_folder_mode` variable with the mode you want.
I am not sure which one is better, I would like to avoid adding a lot of additional configuration option.
On the other hand we already have `SIKP_DACE_WARNINGS`, but this could not be achieved by exporting an env.

What are you thoughts about that @havogt @egparedes ?

iomaganaris (approved):
Looks good from my side 👍

src/gt4py/next/program_processors/runners/dace/workflow/common.py
havogt: I tend to the dace variable, to avoid pollution of the gt4py config for expert-level debug utils.

philip-paul-mueller: I removed the configuration variable, but I added a note to about how to switch to development.

havogt (approved):
lgtm

philip-paul-mueller:
cscs-ci run

PR #2567, opened 2026-04-03
refactor[cartesian]: Better error message for out of bound read on stencils

CharlesKrop:
- Better error message for out of bound read: refactor [cartesian]

## Description

Changed the error message raised when a stencil attempts to read outside of the valid bounds of an array. The new message now provides information about which fields fail, their side, and some tips to guide users resolving the error.

Existing test suite covers this change, new tests should not be required to validate this change.


FlorianDeconinck (approved):
LGTM - co-authored.

FlorianDeconinck:
cscs-ci run

PR #2568, opened 2026-04-09
fix[next]: Deterministic dace auto optimize

tehrengruber:
This PR is a first vertical slice to make `auto_optimize` deterministic. The general idea is to use ordered datastructures everywhere where order matters (e.g. when iterating over elements of a set) and provide an easy to follow workflow to debug and resolve indeterminism (see [here](https://github.com/GridTools/gt4py/blob/f303b2668b637ccc4f105bde556febc292f86209/docs/development/dace_codegen_reproducability.md)). In many places this just means replacing `set` with `OrderedSet` from https://github.com/rspeer/ordered-set and using deterministic naming schemes. In addition to the changes here https://github.com/GridTools/dace/pull/17 is needed, which also introduces an `id` property in all dace nodes. The `id` property is computed from a deterministic counter incremented whenever a new node is constructed. Together with the `reset_node_id_counter` context manager each node gets a unique id, which is stable across runs. This was initially meant as a way to order output from networkx algorithms used in dace that return sets, but surprisingly this was not needed for the stencil used to test this PR. Since this can be different for other stencils and since the id value is also very useful to quickly recognize that order has changed this is kept for now.

Open discussion points:
- Is `OrderedSet` the right package?
- How should we test this? Using icon4py or a fuzzer?
- Guidelines for symbol name generation

The program used to test this is PR:
icon4py @ 0918c3d with `model/atmosphere/dycore/tests/dycore/stencil_tests/test_compute_advection_in_vertical_momentum_equation.py::TestFusedVelocityAdvectionStencilVMomentum::test_TestFusedVelocityAdvectionStencilVMomentum[compile_time_domain]`


docs/development/dace_codegen_reproducability.md
```diff
@@ -0,0 +1,95 @@
+# Debugging indeterministic behavior of dace transformations
+
+- Enable printing each transformation step, e.g. using
+  ```
+  dace.Config.set("progress", value=True)
+  ```
+  TODO: introduce new config var that prints the hash instead of hard-coding it.
+- Execute the program in question twice and compare the output.
+- Set a conditonal breakpoint in beginning of the `apply` method of the first pass where the SDFG
```
philip-paul-mueller, line 9: `apply()` is something that is specific to the [`PatternTransformation`](https://github.com/spcl/dace/blob/0ec62e20d44b6cb5f83f89b2fbc9b61f218e48e5/dace/transformation/transformation.py#L65).
Not all transformations use them (but the majority does).
The most general interface is given by [`Pass`](https://github.com/spcl/dace/blob/0ec62e20d44b6cb5f83f89b2fbc9b61f218e48e5/dace/transformation/pass_pipeline.py#L39), whcih defines the `apply_pass()` method.
But as mentioned above, also functions can be transformations.

docs/development/dace_codegen_reproducability.md
```diff
@@ -0,0 +1,95 @@
...
+    transformations as gtx_transformations,
+)
+from dace.utils import print_sdfg_hash
+
+file = sys.argv[1]
+
+with open(file) as f:
+    data = json.load(f)
+    sdfg = dace.SDFG.from_json(data)
+    print_sdfg_hash(sdfg)
```
philip-paul-mueller, line 50: There is also [`dace.SDFG.from_file()`](https://github.com/spcl/dace/blob/0ec62e20d44b6cb5f83f89b2fbc9b61f218e48e5/dace/sdfg/sdfg.py#L1803).

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
philip-paul-mueller, line 1: I looked at this file and it looks okay, although the `O(N)` deletion is not nice.
One could rework the algorithm to get rid of them, however I would not do this.
There is [PR#2531](https://github.com/GridTools/gt4py/pull/2531) which updates this file and also tries to avoid in determinism.
While it still uses `set`s it sorts the nodes ones in a deterministic way (depending on the node order upon input).

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
```diff
@@ -202,8 +206,10 @@ def split_overlapping_map_range(
...
         for the first and the second map, respectively.
     """
-    first_map_params = set(first_map.params)
-    second_map_params = set(second_map.params)
+    # TODO(tehrengruber): The structure here looks a little funky. We just use an ordered set for
+    #  now, but likely no sets are needed at all.
+    first_map_params = OrderedSet(first_map.params)
+    second_map_params = OrderedSet(second_map.params)
     if first_map_params != second_map_params:
         return None
```
philip-paul-mueller, line 214: sets are needed here, at least for the check that follows because a Map with parameters ["i", "j"] is the same as one with parameters ["j", "i"].
However, you can do something like:
```python
first_map_params = sorted(first_map.params)
second_map_params = sorted(second_map.params)
```
then you can also remove the sorted() calls bellow as well.

src/gt4py/next/program_processors/runners/dace/transformations/simplify.py
```diff
@@ -1033,7 +1033,7 @@ def apply(
 
         # Now we will reroute the edges went through the inner map, through the
         #  inner access node instead.
-        for old_inner_edge in list(
+        for old_inner_edge in list(  # TODO(tehrengruber): Why all these list comprehensions everywhere?
```
philip-paul-mueller, line 1036: It is needed because you _replace_ the edges (removing the old and adding a new one).
Furthermore, the `by_connector()` gives you [back an iterator](https://github.com/spcl/dace/blob/0ec62e20d44b6cb5f83f89b2fbc9b61f218e48e5/dace/sdfg/state.py#L522) thus modyfing the edges will change your iteration source, which leads to a race condition.

src/gt4py/next/program_processors/runners/dace/transformations/split_access_nodes.py
philip-paul-mueller, line 1: I think it should be possible to rework most of it to avoid using `set`s.
I am actually not remembering why I was using them in the first place.

src/gt4py/next/program_processors/runners/dace/transformations/splitting_tools.py
```diff
@@ -263,6 +264,7 @@ def split_node(
     # NOTE: Turning them into a string is the best solution is probably the only way
     #   to achieve some stability. The only downside is that the order now depends
     #   on the specialization level that is used, i.e. if we have numbers or symbols.
+    # TODO(tehrengruber): Is this still needed?
```
philip-paul-mueller, line 267: If `split_description` is stable we can remove the sort.

docs/development/dace_codegen_reproducability.md
```diff
@@ -0,0 +1,95 @@
+# Debugging indeterministic behavior of dace transformations
+
+- Enable printing each transformation step, e.g. using
+  ```
+  dace.Config.set("progress", value=True)
```
philip-paul-mueller, line 5: This will only give you like ~95% of the cases, as transformation can also run through other means that the patter matcher or can be simple functions that do things.

docs/development/dace_codegen_reproducability.md
```diff
@@ -0,0 +1,95 @@
...
+file = sys.argv[1]
+
+with open(file) as f:
+    data = json.load(f)
+    sdfg = dace.SDFG.from_json(data)
+    print_sdfg_hash(sdfg)
+
+    sdfg.apply_transformations_repeated(
+        transformation(
+            ignore_upstream_blocks=False,
```
philip-paul-mueller, line 90: This should be `True` for the correct working.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
```diff
@@ -78,8 +79,11 @@ def _new_name(old_name: str) -> str:
         elif isinstance(node, dace_nodes.NestedSDFG):
             node_ = graph.add_nested_sdfg(
                 sdfg=copy.deepcopy(node.sdfg),
-                inputs=set(node.in_connectors.keys()),
-                outputs=set(node.out_connectors.keys()),
+                # TODO(tehrengruber): What is the performance optimization from Philip about?
```
philip-paul-mueller, line 82: What do you mean with my performance optimization?
What cold be a problem is the copying `node.sdfg` might also copy the surrounding SDFG, because nested SDFGs have a reference to their parent SDFG.

src/gt4py/next/program_processors/runners/dace/transformations/map_fusion_utils.py
```diff
@@ -78,8 +79,11 @@ def _new_name(old_name: str) -> str:
...
+                #  In any case this here leads to an sdfg in which the order in graph.nodes
+                #  is indeterministic, but to_json, then from_json restores it again.
+                inputs={k: None for k in node.in_connectors.keys()},
+                outputs={k: None for k in node.out_connectors.keys()},
```
philip-paul-mueller, line 86: This should be equivelent since the data types should not change.
```suggestion
                inputs=node.in_connectors.copy().
                outputs=node.out_connectors.copy(),
```
It you want to play it safe ignore this suggestion.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -372,7 +409,7 @@ def generate_sdfg(
         *args: Any,
         **kwargs: Any,
     ) -> dace.SDFG:
-        with gtx_wfdcommon.dace_context(device_type=self.device_type):
+        with gtx_wfdcommon.dace_context(device_type=self.device_type), dace.sdfg.nodes.reset_node_id_counter():
```
philip-paul-mueller, line 412: I kind of understand why it is used here, although I do not like it.

philip-paul-mueller (commented):
Generally looks okay, there are some improvements possible.

src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -48,6 +48,7 @@ def set_dace_config(
     # NOTE: Each thread maintains its own set of configuration, i.e. `dace.Config` is
     #   a thread local variable. This means it is safe to set values that are different
     #   for each thread.
+    dace.Config.set("progress", value=True)
```
edopao, line 51: This is not meant to be merged, right?

docs/development/dace_codegen_reproducability.md
edopao: In order to avoid editing the code and adding this line, you can export an environment variable:
`export DACE_progress=1`

(note the the upper/lower case unfortunately matters)

edopao:
cscs-ci run determinism

edopao:
cscs-ci run dace-determinism

edopao:
I propose to close this PR, all changes should be merged into `main`.

edopao:
The `dace-determinism` pipeline now passes on `gt4py:main` + `dace:main`.

PR #2569, opened 2026-04-09
fix[next-dace]: Fix SDFG lowering of neighbors-reduce on field with non-zero origin

edopao:
This PR re-enables a warning which was hit in ICON4Py by the dace backend, because we forced the backend to always compute fields everywhere, as a workaround for an issue in the dace backend, also when gt4py can infer the minimal domain at compile time.

The issue in the dace backed was caused by a bug in de-referencing a field computed on the minimal domain. Tested in icon4py PR https://github.com/C2SM/icon4py/pull/1190


src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -113,15 +114,15 @@ def _process_symbolic_domains_option(
...
     if use_max_domain_range_on_unstructured_shift is None:
         use_max_domain_range_on_unstructured_shift = _has_dynamic_domains(ir)
+    elif use_max_domain_range_on_unstructured_shift:
+        if not _has_dynamic_domains(ir):
+            warnings.warn(
+                "You are using static domains together with "
+                "'use_max_domain_range_on_unstructured_shift'. This is "
+                "likely not what you wanted.",
+                stacklevel=2,
+            )  # noqa: ERA001, RUF100
```
```suggestion
            )
```

tehrengruber (approved):
Can we add an integration test? `test_unstructured_shift` called with a field of non-zero domain start / origin should trigger this too right? Maybe we should also rename the PR since the main part is the fix not really the reintroduction of the warning.

edopao:
> Can we add an integration test? `test_unstructured_shift` called with a field of non-zero domain start / origin should trigger this too right? Maybe we should also rename the PR since the main part is the fix not really the reintroduction of the warning.

Good point. I have added two test cases. One is a variant of `test_unstructured_shift`, as you suggested, however this case was already working. The other one is testing `neighbor_sum` with a field of non-zero origin, and this was failing. Anyway, both are relevant: the first covers the lowering of unstructured `shift` operator and the second `reduce+neighbors` builtins.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_execution.py
```diff
@@ -724,6 +747,29 @@ def combine(a: cases.IField, b: cases.IField) -> cases.IField:
     cases.verify_with_default_data(cartesian_case, combine, ref=lambda a, b: a + a + b)
 
 
+@pytest.mark.uses_unstructured_shift
+@pytest.mark.uses_program_with_sliced_out_arguments
+def test_neighbor_sum_with_non_zero_origin(unstructured_case):
+    if unstructured_case.backend is None:
+        pytest.xfail("Embedded backend requires contiguous inverse image.")
```
edopao: @havogt Would that be OK to merge the tests with this `xfail`?

edopao: I have added a todo comment and I have updated the xfail message.

egparedes (approved):
LGTM

PR #2571, opened 2026-04-13
ci: run pre-commit manually

romanc:
## Description

We are currently using the `pre-commit/action`. The action is in [maintenance mode](https://github.com/pre-commit/action) and effectively unmaintained. Last month, all pending issues and PRs were closed as "duplicates" or [without any reason](https://github.com/pre-commit/action/pull/242), presumably to move users over to [precommit.ci](https://pre-commit.ci/).

The action is causing a warning because it uses `actions/cache@v4` under the hood, which isn't (marked as) compatible with node24. Node24 will be the default in a couple months. We should thus update to avoid potential issues once node24 gets to be the new default.

image

This PR suggests to just drop the dependency on `pre-commit/action` and instead call `pre-commit` by hand. The PR preserves caching the pre-commit cache and thus the functionality is as before.

## Requirements

  N/A
  N/A

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -38,8 +38,16 @@ jobs:
           enable-cache: true
           cache-dependency-glob: "uv.lock"
 
-      - name: "Run pre-commit"
-        uses: pre-commit/action@v3.0.1
+      - name: Restore cache folder
+        uses: actions/cache@v5
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit_${{ env.pythonLocation }}_${{ hashFiles('.pre-commit-config.yaml') }}
+
+      - name: Run pre-commit
+        run: |
+          pip install pre-commit
+          pre-commit run --show-diff-on-failure --color=always --all-files
 
   typing-exports:
     needs: [get-python-versions]


.github/workflows/code-quality.yml
```diff
@@ -38,8 +38,16 @@ jobs:
...
+      - name: Restore cache folder
+        uses: actions/cache@v5
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit_${{ env.pythonLocation }}_${{ hashFiles('.pre-commit-config.yaml') }}
+
+      - name: Run pre-commit
+        run: |
+          pip install pre-commit
+          pre-commit run --show-diff-on-failure --color=always --all-files
```
egparedes, line 51: Just a question: is this more or less the content of the deprecated `pre-commit` action? 

Edit: it seems so. Forget the silly question :facepalm:

egparedes (approved):
LGTM, just a question for my own understanding.

.github/workflows/code-quality.yml
romanc: Yes, the old action really didn't do much. There are a couple of forks, which have been updated, out there. In the end, I figured, it's so easy to do "by hand" that I don't think we really need a dedicated action for this.

PR #2572, opened 2026-04-13
ci: update slackapi/slack-github-action

romanc:
## Description

This PR updates the slack notification of the "Daily CI" workflow on GitHub Action.

Notification was using an outdated version, which had no support for node24. Node24 will be the default node version on GitHub Action runners in a couple months. We should update before to avoid potential issues with the slack notifications.

image

The PR updates the notification part according to the upgrade guides to [v2](https://github.com/slackapi/slack-github-action/releases/tag/v2.0.0) and [v3](https://github.com/slackapi/slack-github-action/releases/tag/v3.0.0).

@egparedes How do I test with `SLACK_BOT_CHANNEL_TEST`? If I change the workflow to use that channel, where do test messages show up?

## Requirements

  N/A
  N/A

### DIFF
--- a/.github/workflows/daily-ci.yml
+++ b/.github/workflows/daily-ci.yml
@@ -7,9 +7,9 @@ on:
 
   ## COMMENTED OUT: only for testing CI action changes.
   ## It only works for PRs to `main` branch from branches in the upstream gt4py repo.
-  # pull_request:
-  #   branches:
-  #     - main
+  pull_request:
+    branches:
+      - main
   ## END
 
 jobs:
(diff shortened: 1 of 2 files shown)


egparedes:
@romanc IIRC the notifications test channel is `_test-app-integrations_` in the GridTools slack

romanc:
testing with the test slack channel failed - more work needed ...

egparedes:
@romanc did you also try with the normal notifications channel? I think it should be fine to receive a bit of spam there during debugging...

romanc:
testing with the test slack channel failed - more work needed ...

> @romanc did you also try with the normal notifications channel? I think it should be fine to receive a bit of spam there during debugging...

I don't think the channel is the problem. [This run](https://github.com/GridTools/gt4py/actions/runs/24395167215/job/71251031715?pr=2572) suggests that the token is not configured correctly. I've changed back to using an environment variable (as before, but with a changed name) to see if that works better.

romanc:
@egparedes I am at the end of my wisdom here. That was supposed to be an easy thing on the side, but I can't seem to make the action read the secret. Maybe a new type of secret is needed? Just a guess, I can't double-check that part because I don't have admin rights on the repo (which is totally fine). Things will continue to work for some time, so this is not super urgent. Feel free to take over. I won't have time to work on this in the coming weeks.

PR #2573, opened 2026-04-14
refactor[cartesian]: appropriate types for `interval` / `horizontal`

romanc:
## Description

GTScript offers context managers like

- `with computation(...)`
- `with interval(...)`
- `with horizontal(...)`

while the first one correctly returns (a trivial) context manager, the other two functions just `pass` and don't return anything. Because these context managers are just used for parsing, this isn't important. However, it trips static code analysis like `pyright` and adds squiggly red lines that could be avoided.

image

This PR suggests for `interval()` and `horizontal()` to also return a trivial context manager, which will avoid the squiggly lines.

## Requirements

  N/A
  N/A

### DIFF
--- a/src/gt4py/cartesian/gtscript.py
+++ b/src/gt4py/cartesian/gtscript.py
@@ -786,12 +786,12 @@ def computation(order):
 
 def interval(*args):
     """Define the interval of computation in the 'K' sequential axis."""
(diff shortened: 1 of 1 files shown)


twicki (approved):
concise, clean, let's go

PR #2574, opened 2026-04-14
feat[cartesian]: Enforce cartesian layout to be given precedence in allocation striding

FlorianDeconinck:
## Description

We change the algorithm to compute strides from dimensions to make sure the cartesian are always given lower strides and data dimensions are pushed after cartesian in stride space. This fix the "kfirst" 4D arrays to actually have K as the smallest stride

e.g.

```
dace:cpu_kfirst:[8, 4, 2, 1]
dace:cpu_KJI:[1, 2, 4, 8]
dace:cpu:[2, 1, 4, 8]
```
becomes
```
dace:cpu_kfirst:[4, 2, 1, 8]
dace:cpu_KJI:[1, 2, 4, 8]
dace:cpu:[2, 1, 4, 8]
```

for a (2, 2, 2, 2) domain.

TODO:
- [x] Unit test that checks the above "cartesian first"

### DIFF
--- a/src/gt4py/storage/cartesian/layout.py
+++ b/src/gt4py/storage/cartesian/layout.py
@@ -6,19 +6,7 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    Dict,
-    Final,
-    Literal,
-    Optional,
-    Sequence,
-    Tuple,
-    TypedDict,
-    Union,
-)
+from typing import TYPE_CHECKING, Any, Callable, Literal, Tuple, TypedDict, Union
 
 import numpy as np
 
@@ -35,34 +23,6 @@ class LayoutInfo(TypedDict):
     device: Literal["cpu", "gpu"]
     layout_map: Callable[[Tuple[str, ...]], Tuple[int, ...]]
     is_optimal_layout: Callable[[Any, Tuple[str, ...]], bool]
-
-
-REGISTRY: Dict[str, LayoutInfo] = {}
-
-
-def from_name(name: str) -> Optional[LayoutInfo]:
-    return REGISTRY.get(name, None)
-
-
-def register(name: str, info: Optional[LayoutInfo]) -> None:
-    if info is None:
-        REGISTRY.pop(name, None)
-    else:
-        assert isinstance(name, str)
-        assert isinstance(info, dict)
-
-        REGISTRY[name] = info
-
-
-def check_layout(layout_map, strides):
-    if not (len(strides) == len(layout_map)):
-        return False
-    stride = 0
-    for dim in reversed(np.argsort(layout_map)):
-        if strides[dim] < stride:
-            return False
-        stride = strides[dim]
-    return True
 
 
 def layout_maker_factory(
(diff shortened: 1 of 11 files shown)


FlorianDeconinck:
cscs-ci run

twicki:
Maybe I misunderstand the code here. But my understanding is that all the allocation of the data does not go through the functions that are modified here but are either hard-coded: https://github.com/GridTools/gt4py/blob/81c864cd80bdcd21cd8f507e2f8fa3c3f2ab17d0/src/gt4py/storage/cartesian/layout.py#L139
or go though the interal functions below:
https://github.com/GridTools/gt4py/blob/81c864cd80bdcd21cd8f507e2f8fa3c3f2ab17d0/src/gt4py/storage/cartesian/layout.py#L148

And this here only notifies the backend about the layout.

Why we have two options of doing that, I have no idea but if we start modifying them, maybe we should consolidate the two?
And I assume this is why it worked "by chance" on some backends

src/gt4py/storage/cartesian/layout.py
```diff
@@ -37,50 +25,30 @@ class LayoutInfo(TypedDict):
     is_optimal_layout: Callable[[Any, Tuple[str, ...]], bool]
 
 
-REGISTRY: Dict[str, LayoutInfo] = {}
```
twicki, line 40: The regristy has been moved to its own central place

src/gt4py/storage/cartesian/layout_registry.py
```diff
@@ -0,0 +1,113 @@
...
+    if info is None:
+        REGISTRY.pop(name, None)
+    else:
+        assert isinstance(name, str)
+        assert isinstance(info, dict)
+
+        REGISTRY[name] = info
+
+
+register(
```
twicki, line 33: we could have a discussion if all backends should support their own layout or if we should minimize the amount of layouts to only the acutally different ones. But I think this would be a pretty significant change on how to use the data-generators that does not fit here

tests/cartesian_tests/integration_tests/feature_tests/test_field_layouts.py
```diff
@@ -74,3 +74,26 @@ def test_bad_layout_warns(backend):
         "provided allocators in `gt4py.storage`.",
     ):
         stencil(field_a=inp, field_b=outp)
+
+
+@pytest.mark.parametrize("backend", ALL_BACKENDS)
```
twicki, line 79: This is the actual test that checks the behavior of the data indices

tests/storage_tests/unit_tests/test_layout.py
```diff
@@ -48,90 +48,84 @@
...
-        (("0", "1"), (1, 0)),
-        (("I", "0", "1"), (2, 1, 0)),
-        (("J", "0", "1"), (2, 1, 0)),
-        (("K", "0", "1"), (2, 1, 0)),
-        (("I", "J", "0", "1"), (3, 2, 1, 0)),
-        (("I", "K", "0", "1"), (3, 2, 1, 0)),
-        (("J", "K", "0", "1"), (3, 2, 1, 0)),
-        (("I", "J", "K", "0", "1"), (4, 3, 2, 1, 0)),
         (("0",), (0,)),
+        (("1",), (0,)),
```
twicki, line 101: The check that "0" and "1" behaves identical is done here and does not need to be repeated on every permutation containing "0"

twicki:
cscs-ci run

FlorianDeconinck:
Looking good! You can ok myself-yourself in, since I can't because I am the author but you are. Yeah.

twicki (approved):

FlorianDeconinck:
cscs-ci run

FlorianDeconinck:
cscs-ci run

PR #2576, opened 2026-04-16
Fix wrong python version selection in some daily CI tasks

egparedes:
Fix a bug in the way `nox` was called to run some specific sessions which made every CI matrix entry run more tests than needed:  `-t` (tag filter) and `-s` (session filter) options do not interact together as an intersection, but `-t` completely replaces `-s`.

Additionally, fix some other minor issues and add comments on how to debug CI issues for this workflow without running all tests.


romanc (approved):
To be completely honest, this is all a bit voodo to me, but in general terms I think this looks good.

Happy to hear that we don't have 80 workflosws running anymore nightly :tada:

edopao (approved):
Thanks for this fix! LGTM

PR #2577, opened 2026-04-17
feat[next-dace]: Clarify use of HIPFLAGS

iomaganaris:
Replace `hip_args` with `HIPFLAGS` to avoid any confusion when `HIPFLAGS` is passed. This variable should be already used by CMake to set the HIP flags, no matter what we set in DaCe


edopao (approved):
LGTM

PR #2578, opened 2026-04-17
perf[next-dace]: Enhance LoopBlocking pass

iomaganaris:
Added the following features to the `LoopBlocking` pass:
- `Memlet`s that have independent data are promoted in the outer `Map` so that they only have to be read once
- Added option `blocking_independent_node_threshold`. `LoopBlocking` is only going to be applied to `Map`s that have more than the threshold number independent variables


src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -672,6 +672,8 @@ def can_be_applied(
             return False
         if self.map_entry.map.gpu_block_size is not None:
             return False
+        if self.map_entry.map.gpu_maxnreg > 0:
+            return False
```
philip-paul-mueller: I am actually not sure if this is needed, as `maxnreg` is [only set if it is not set](https://github.com/GridTools/gt4py/blob/24ed83e034dfdc982487829c6f2f682ca542b862/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py#L765).
However, I assume this changes avoids some other problem, that I don't fully get.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -74,10 +74,21 @@ class LoopBlocking(dace_transformation.SingleStateTransformation):
         default=True,
         desc="If 'True' then blocking is only applied if there are independent nodes.",
     )
+    promote_independent_memlets = dace_properties.Property(
```
philip-paul-mueller, line 79: You also need to add them to the doc string of the class, there I would also add what promotion means, probably creating a scalar.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -137,6 +155,10 @@ def can_be_applied(
 
         block_var_idx = map_params.index(block_var)
         map_range_size = map_range.size()
+
+        if all(map_range_size_i == 1 for map_range_size_i in map_range_size):
```
philip-paul-mueller: There could be a Sympy issue here, so you have to use:
```suggestion
        if all((map_range_size_i == 1) == True for map_range_size_i in map_range_size):
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -74,10 +74,21 @@ class LoopBlocking(dace_transformation.SingleStateTransformation):
...
+    independent_node_threshold = dace_properties.Property(
+        dtype=int,
+        default=0,
+        desc="Minimum number of independent nodes required to apply blocking (non-inclusive).",
+    )
 
     # Set of nodes that are independent of the blocking parameter.
     _independent_nodes: Optional[set[dace_nodes.AccessNode]]
     _dependent_nodes: Optional[set[dace_nodes.AccessNode]]
+    _memlet_to_promote: Optional[set[dace.Memlet]]
```
philip-paul-mueller: This is confusing (and also very strange), but the class `Memlet` only the payload of the edge in the dataflow graph, although everyone is _pretending_ that it is the actual edge.
Thus the correct type signature is:
```suggestion
    _memlet_to_promote: Optional[set[dace_graph.MultiConnectorEdge[dace.Memlet]]]
```
However, as far as I understand it should be possible to change the `set` into a `list`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller, line 199: Currently you couple `can_be_applied()` with `apply()` because you do not set `self._memlet_to_promote` to `None`.
Thus you require that `can_be_applied()` has run immediate before `apply()` has been called.

To solve this you have to reset `_memlet_to_promote` here and in `apply()` populate it again.
You can also do it in `self._prepare_independent_memlets()`, see my other note.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -323,7 +361,7 @@ def partition_map_output(
                 )
 
                 # Check if the partition exists.
-                if class_res is None:
+                if self.require_independent_nodes and class_res is None:
```
philip-paul-mueller: This change does not make sense to me, because `None` means that the partition (for whatever reason) does not exist and the transformation can not handle this node or particular situation.
If a node is independent then `_classify_node()` return `True` and in case it is dependent it returns `False`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -345,8 +383,9 @@ def partition_map_output(
         )
 
         # If requested check if the blocking is a good idea.
-        if self.require_independent_nodes and (not self._check_if_blocking_is_favourable(state)):
+        if not self._check_if_blocking_can_promote_anything():
             self._independent_nodes = None
+            self._memlet_to_promote = None
             return False
```
philip-paul-mueller: I am thinking that the functionality to check if the blocking is okay or not does not belong here.
Instead it should be moved outside of this function.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -533,7 +572,10 @@ def _classify_node(
             dst_subset: dace_subsets.Subset | None = memlet.dst_subset
 
             if memlet.is_empty():  # Empty Memlets do not impose any restrictions.
-                continue
+                if len(in_edges) == 1 and self.require_independent_nodes:
+                    return False
+                else:
```
philip-paul-mueller: I do not understand the purpose of this change.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -552,6 +594,12 @@ def _classify_node(
             if not (in_edge.src is outer_entry or in_edge.src in self._independent_nodes):
                 return False
 
+        if self._memlet_to_promote is not None:
+            # Remove all memlets that are connected to an independent node
+            for edge in state.in_edges(node_to_classify):
+                if edge in self._memlet_to_promote:
+                    self._memlet_to_promote.remove(edge)
```
philip-paul-mueller: I do not think that you have to do it here.
As I understand it, a Memlet that can/should be promoted is one that connects the Map entry node to dependent node.
So you can just run something like:
```python=
for oedge in state.out_edges(map_entry):
    if oedge.src not in self._independent_nodes and self._check_if_edge_can_be_promoted(oedge, ...):
        self._memlet_to_promote.add(oedge)
```
after the classification has been done.
This also makes it simpler to decouple the `apply()` and `can_be_applied()`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -165,6 +195,9 @@ def apply(
         # Now compute the partitions of the nodes.
         self.partition_map_output(graph, sdfg)
 
+        if self.promote_independent_memlets:
+            self._prepare_independent_memlets(graph, sdfg)
```
philip-paul-mueller, line 217: As I wrote above you are currently relying that `apply()` is called right after `can_be_applied()`.
To be conforming you need to recreate the `self._memlet_to_promote` attribute.
However, there are some problems to this, that might need some changes.
I recommend the following:
- You run `self.partition_map_output()`, which is needed to get the correct set of Memlets to promote.
- Then you recompute the set of edges that you want to promote.
- Then you perform the promotion, i.e. generating the temporary storage _and_ update the `self._independent_nodes`.
- Then you continue with `self._prepare_inner_outer_maps()`.

Note that there is a very interesting problem here, that should not affect you.
You see `self.map_entry` is a special object, it is not a reference to the Map entry node.
Instead it is an object that stores a reference to the state and the ID of the node (Map entry).
Thus, writing `self.map_entry` will perform a lookup, but since the lookup according to IDs is implemented very [crappy](https://github.com/spcl/dace/blob/51790b5684511e1a379a90ca3f5a374e99817124/dace/sdfg/graph.py#L657), modifying the nodes of a state is dangerous as it will destroy the node IDs.
However, this only happens for deletion, so it is fine in your case.

Why I am writing that?
Because `self._prepare_inner_outer_maps()` implicitly assumes that `self.map_entry` will give the correct node.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+        subsets_to_inspect: list[dace_subsets.Subset] = []
+        if dst_subset is not None:
+            subsets_to_inspect.append(dst_subset)
+        if src_subset is not None:
+            subsets_to_inspect.append(src_subset)
+
+        if any(self.blocking_parameter in subset.free_symbols for subset in subsets_to_inspect):
+            return False
+
+        if isinstance(edge.dst, dace_nodes.Tasklet) and not edge.data.data.startswith("gt_conn_"):
```
philip-paul-mueller, line 694: This looks a bit brittle to me.
So I would do the following.

First check if we have a scalar read, this can be done by:
```python=
# Check scalar reads.
if (edge.data.subset.num_elements() == 1) == True:
    return True
```
The use of `subset` is intentional here and handles problems of Tasklet vs. AccessNode consumers.
I would make the second check more specific, i.e.:
```python=
# Handle connection tables
if isinstance(edge.dst, dace_nodes.Tasklet) and edge.data.data.startswith("gt_conn_"):
    return True
```
However, I would say, that most of these are enclosed by a Map because they appear in the context of a neighbourhood reduction.
I am not sure maybe it might be enough to also check if `edge.dst` is also a `dace_nodes.MapEntry`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+            # (connection name: `__tlet_field`) and one or two offsets
+            # (connection name: `__tlet_index_Cell` and `__tlet_index_K`).
+            # If there is no `K` dependent offset if we have blocking on `K`, we can promote the memlet.
+            # If there is no `Cell` dependent offset if we have blocking on `Cell`, we can promote the memlet.
+            # However above require checking the tasklet internals without being sure of these conventions.
+            # For that purpose we choose to promote only memlets that are passed to tasklets only
+            # if they are connectivities.
+            return False
+
+        return True
```
philip-paul-mueller, line 705:
```suggestion
        return False
```
This is a bit more safe in my opinion as you check what should be allowed instead of what is forbidden.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+
+        _ = sdfg.reset_cfg_list()
+        dace_sdutils.canonicalize_memlet_trees_for_map(state=state, map_node=outer_map_entry)
+        dace_propagation.propagate_memlets_map_scope(sdfg, state, outer_map_entry)
+
+        for in_edge in self._memlet_to_promote:
+            # Create a temporary AccessNode that will be used to promote the memlet
+            independent_outer_map_indexes = [
+                index for index in outer_map_entry.map.params if index != self.blocking_parameter
+            ]
```
philip-paul-mueller: Can be moved outside the loop.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+            independent_outer_map_as_range = dace_subsets.Range(
+                ranges=[
+                    in_edge.data.subset.ranges[outer_map_entry.map.params.index(idx)]
```
philip-paul-mueller: Since this mapping is fix, you could create a `dict` outside the loop.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+                state=state,
+                edge=in_edge,
+                new_dst=promoted_anode,
+                new_dst_conn=None,
+                new_memlet=dace.Memlet(
+                    data=in_edge.data.data,
+                    subset=in_edge.data.subset,
+                    other_subset=dace_subsets.Range.from_array(sdfg.arrays[promoted_name]),
+                ),
+            )
```
philip-paul-mueller, line 769: This might fail in very strange ways if the consumer is an AccessNode.
The following code should handle that case too:

```python=
assert in_edge.data.wcr is None
simple_direction = not (isinstance(in_edge.dst, dace_nodes.AccessNode) and in_edge.data.data != in_edge.dst.data)
first_new_memlet = deepcopy(in_edge.data)
if simple_direction:
    first_new_memlet = dace.Memlet(
        data=in_edge.data.data,
        subset=deepcopy(in_edge.data.subset),
        other_subset=deepcopy(new_subset_as_range),
    )
else:
    first_new_memlet = dace.Memlet(
        data=promoted_name,
        subset=deepcopy(new_subset_as_range),
        other_subset=deepcopy(in_edge.data.other_subset),
    )

second_new_memlet = deepcopy(in_edge.data)
if simple_direction:
    second_new_memlet.data = promoted_name
    second_name_memlet.subset = deepcopy(new_subset_as_range)
else:
    second_name_memlet.other_subset = deepcopy(new_subset_as_range)

state.add_edge(
    in_edge.src,
    in_edge.src_conn,
    promoted_anode,
    None,
    first_new_memlet,
)
state.add_edge(
    promoted_anode,
    None,
    in_edge.dst,
    in_edge.dst_conn,
    second_new_memlet,
)
state.remove_edge(in_edge)
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+
+            if isinstance(original_dst_of_in_edge, dace_nodes.MapEntry):
+                for edge in state.out_edges(original_dst_of_in_edge):
+                    if edge.data.data == in_edge.data.data:
+                        edge.data.data = promoted_name
+                        assert len(original_dst_of_in_edge.params) == 1, (
+                            "Independent memlets should only be inputs to maps that have a single parameter. "
+                            "Those should always be neighbor reductions."
+                        )
+                        edge.data.subset = next(iter(original_dst_of_in_edge.params))
```
philip-paul-mueller, line 792: It is correct that you have to update `subset` here, but does does not make much sense to me.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+                        edge.data.other_subset = None
```
philip-paul-mueller: You can not set this to `None` you must keep it.
Instead there is an `else` branch missing, there you need to update `other_subset` (you need to set it to the same value you used for `.subset` in the `then` branch.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+            elif isinstance(original_dst_of_in_edge, dace_nodes.NestedSDFG):
+                raise NotImplementedError("Promotion of memlets to NestedSDFG not implemented yet.")
+            elif isinstance(original_dst_of_in_edge, dace_nodes.LibraryNode):
+                raise NotImplementedError(
+                    "Promotion of memlets to LibraryNode not implemented yet."
+                )
```
philip-paul-mueller, line 798: I think you can remove them, since `original_dst_of_in_edge` is always classified as a dependent node.

philip-paul-mueller (commented):
I like the changes but there some things that needs to be addressed.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
iomaganaris: With this check I want to make sure that if the `gpu_maxnreg` is set for the map from some other transformation we don't change it here but the check in the `can_be_applied` is wrong so I'm going to change it

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: Since iterate over `_memlet_to_promote` it should be a `list` indeed. And you're also right for the type of `_memlet_to_promote`. Not sure how the syntax checks didn't catch this issue

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: Why do things have to be that complicated? 🙈 Thank you very much for the detailed comment 👍

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: It was necessary for the way I implemented the `require_independent_nodes` check. It's no longer needed with my new changes

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: This was necessary because I was populating the independent memlets before populating the independent nodes. If we populate the independent memlets after populating the nodes we can add a check there to avoid adding to the independent memlets a memlet that has as dst an independent node

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: Tbh handling what's going on inside the Tasklets is so unclear I don't want to touch this code. I would restrict this as much as possible. Unfortunately the proposed changes didn't work out of the box

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: Not sure I understand the point of this comment

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: > This might fail in very strange ways if the consumer is an AccessNode.

If it's an `AccessNode` then it would be an independent one, so it shouldn't be handled here. Maybe we should add an `assert` instead

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: Not sure I understood your comment. Should I change something?

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: I thought that if the edge memlet data is not the one that I have touched then I won't have to update anything because then the edge data will be the output data so the `subset` would be correct. I don't know if there could be a case that the `other_subset` refers to an outside node data for an edge that stems from a `MapEntry`

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: Not sure why I could remove these checks. Couldn't we have a dependent `NestedSDFG` or `LibraryNode`?

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: DaCe is not typechecked, so every type from it is apparently handled as `Any`.
But you have to ask Enrique if you want more information.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: What I was trying to say is that before `_check_if_edge_can_be_promoted()` returned `True` at the very end, thus you allow the promotion except for very specific cases.
However, what I was suggesting it, was that you forbid the promotion except you allow it.
I think this is more explicit and allows for a more targeted approach.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: You are right I did not consider that.
However, instead of an assert I would raise a `NotImplementedErrror`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+            independent_outer_map_as_range = dace_subsets.Range(
+                ranges=[
+                    in_edge.data.subset.ranges[outer_map_entry.map.params.index(idx)]
+                    for idx in independent_outer_map_indexes
+                ]
+            )
+            new_subset = []
+            for subset_range in in_edge.data.subset.ranges:
+                if subset_range not in independent_outer_map_as_range.ranges:
+                    new_subset.append(subset_range)
```
philip-paul-mueller, line 727: This looks a bit brittle, because you look at sizes.
However, currently I do not have a better idea, beside looking at the subset and checking if it contains the blocking parameter, but I am not sure if this is better.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+            assert len(new_subset) > 0, (
+                "After removing the independent dimensions there should be at least one dimension left to promote."
+            )
```
philip-paul-mueller, line 730: I think the number of dimensions to promote should be one less than before?

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: If there is an AccessNode inside a Map and the Memlet _inside_ the Map scope refers to the AccessNode inside, then `outher_subset` stores what is read from the data outside.
Note that the `data` attribute of the Memlet _outside_ the Map will refer to the data that is outside.
However, this is a case that is unlikely but not impossible to appear.

Here you have to keep `other_subset` in case the Memlet ends in an AccessNode.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+                original_dst_conn_of_in_edge,
+                memlet=dace.Memlet(
+                    data=promoted_name,
+                    subset=dace_subsets.Range.from_array(promoted_desc),
+                    other_subset=original_dst_other_subset_of_in_edge,
+                ),
+            )
+
+            if isinstance(original_dst_of_in_edge, dace_nodes.MapEntry):
+                for edge in state.out_edges(original_dst_of_in_edge):
```
philip-paul-mueller: I did not saw this before, but iterating over the out edges is not enough, as there could be nested Maps.
I think there is something in utils, the reroute or so that can help you do it or at least give you some hints on how to do it.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +648,140 @@ def _post_process_independent_nodes(
...
+                    if edge.data.data == in_edge.data.data:
```
philip-paul-mueller: The `else` branch is missing.
There you have to update `other_subset`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: I mean `original_dst_of_in_edge` is a `MapEntry` and it does not have an attribute called `params`, the `Map` has this, i.e. `original_dst_of_in_edge.map.params` would exists.
`Map::params` stores the iteration variables (ordered accordingly to [this function](https://github.com/GridTools/gt4py/blob/c1b9c80f89fc2e6718d4a55171eaae9ef7282624/src/gt4py/next/common.py#L1390)).
However, now you take the _first_ iteration variable, which at this point should be horizontal dimension.
So I do not understand the logic that is applied here.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: > Note that there is a very interesting problem here, that should not affect you.
You see self.map_entry is a special object, it is not a reference to the Map entry node.
Instead it is an object that stores a reference to the state and the ID of the node (Map entry).
Thus, writing self.map_entry will perform a lookup, but since the lookup according to IDs is implemented very [crappy](https://github.com/spcl/dace/blob/51790b5684511e1a379a90ca3f5a374e99817124/dace/sdfg/graph.py#L657), modifying the nodes of a state is dangerous as it will destroy the node IDs.
However, this only happens for deletion, so it is fine in your case.
Why I am writing that?
Because self._prepare_inner_outer_maps() implicitly assumes that self.map_entry will give the correct node.

Have I understood correctly that the fix to that is to get the `self.outer_entry` in the beginning of the `apply` function and then pass it as an argument to the necessary functions, instead of calling `self.outer_entry` inside them?

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: So you mean changing the function `_check_if_edge_can_be_promoted` to `_check_if_edge_cannot_be_promoted`? It seems to me like the exact same thing unless you mean something different 🙈

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: `original_dst_of_in_edge` is a `MapEntry` indeed but it has a `params` attribute.
You're right that `next(iter(original_dst_of_in_edge.params))` is wrong. Instead it should be I believe:
```
edge.data.subset = dace_subsets.Range.from_indices(original_dst_of_in_edge.params)
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: Yes, and this applies to any node that has been pattern matched and is accessed by `self.pattern_matched_node`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: I think in this situation we can probably check the name of the Tasklet (probably only the beginning; We should come up with a policy on how we modify names) and some attributes.
But we should probably check with Edoaro to make sure that we match the right thing.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: In the beginning of `_prepare_independent_memlets` we call `canonicalize_memlet_trees_for_map` which should avoid the following scenario:
image

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: I mean something different.
The function has essentially this structure:
```python=
def _check_if_edge_can_be_promoted():
    if check_bad_case_1():
        return False
    if check_bad_case_2():
        return False
    return True
```
This means that the spirit of the function is "allow promotion except I know for sure that it is bad".
In my view this is a bit risky and I would change it to:
```python=
def _check_if_edge_can_be_promoted():
    if check_good_case_1():
        return True
    if check_good_case_2():
        return True
    return False
```
Thus the spirit has changed to "allow promotion only when I know it is useful".

I think that this implementation philosophy is a bit better as it can be more targeted.
However, I am not particular attached to it, I just think that it is a more fine grained way.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: In case we have parameters in the ranges I think we actually check the parameters and not the sizes. For example:
```
(Pdb) p in_edge.data.subset.ranges
[(__i0, __i0, 1), (0, 2, 1)]
(Pdb) p independent_outer_map_as_range.ranges
[(__i0, __i0, 1)]
(Pdb) p new_subset
[(0, 2, 1)]
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: If `isinstance(edge.dst, dace_nodes.MapEntry)` would it help if we called
```
dace_sdutils.canonicalize_memlet_trees_for_map(state=state, map_node=edge.dst)
        dace_propagation.propagate_memlets_map_scope(sdfg, state, edge.dst)
```
?

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: Okay there are two things here.
- Setting `other_subset` to `None` is okay if the target is a scalar (most cases), but it is an error if the target is an array, i.e. where it is not obvious where the data should be written to.
    Consider the example attached which essentially does `double t[2]; t[0] = a[i]; t[1] = b[i];`.
    example
    If you would now set `other_subset` to `None` then you would get something like `double t[2]; t[None] = a[i]; t[None] = b[i];`.
    I will be at the office tomorrow, if you have further questions just ask me.
- Using `canonicalize_memlet_trees_for_map()` is a very nice idea, however, it also means that you do not need the `if` because every `.data` attribute you find that does not refer to the data is an error.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: Maybe I miss something, but here you modifying the (downstream) Memlets such that instead reading from the global data they read from `promoted_anode`.
You do this because you know that the read (expressed by the Memlet you promoted) does not depend on `K`.
However, has far as I understand `original_dst_of_in_edge.params` contains `K`?

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: No.
What I meant is that one level of Memlets might not be enough, as they could "continue".
Thus you must use something like [here](https://github.com/GridTools/gt4py/blob/99b79b58b67f97060bab6c82f4c9380b6759ea31/src/gt4py/next/program_processors/runners/dace/transformations/utils.py#L525), i.e. you iterate over the Memlet tree.
Ideally you could use `reconfigure_dataflow_after_rerouting()`, however, that function can only handle simple shifts.
You do not have simple shifts because in line 726 you conditionally add.
What you could do is adding dummy dimensions, i.e. size 0 dimensions.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: I am not fully sure what I meant with it.
But thinking about it now, I think the `can_be_applied()` should make sure that you do not hit that case.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: I asked a friend to create a pre-commit check for this case. See https://github.com/GridTools/gt4py/tree/precommit_check. Let me know if we should add something like this

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: I think that in that case we actually have to implement the handling of those cases instead of silently avoid applying the transformation but I haven't seen any of these cases yet so that's why I left it as it is 🙈

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: The two things above are not actually the same.
Case 1 (current version) implies:
`not check_bad_case_1() and not check_bad_case_2()`
```
== not(check_bad_case_1() or check_bad_case_2())
== not(not check_good_case_1() or not check_good_case_2())
== (check_good_case_1() and check_good_case_2())
```
Case 2 implies:
`check_good_case_1() or check_good_case_2()`
If we assume that `check_good_case_1() = not check_bad_case_1()` and `check_good_case_2() = not check_bad_case_2()` to get the equivalent of the Case 1 we would need to do:
```
def _check_if_edge_can_be_promoted():
    can_be_promoted = True
    can_be_promoted &= check_good_case_1()
    can_be_promoted &= check_good_case_2()
    return can_be_promoted
```
I find this a bit too verbose tbh so I would prefer the first version. If there's no test that checks any other failing case for this function I would say that the current world truth is enough for not checking more cases 😅

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: `original_dst_of_in_edge` is the `dst` of the `edge` before any changes. In case this is a `Map`, I have the impression that based on our lowering it can only be a neighbor reduction. In case it's a neighbor reduction it shouldn't have as parameter a vertical or horizontal index. I think in general in our lowering there should be a `Map` inside another `Map` that refers to vertical or horizontal indexes. In case there is such map it should be one that was created by the `LoopBlocking` pass. However we only call this pass once and even if we called it more than once then the inner `Map` would be `dependent` and also after the latest changes I added, the `edges` pointing to such map shouldn't be considered for promotion.
Let me know if the above is right and makes sense

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: I tried traversing the memlet tree as you mentioned. Hopefully that should be enough? 🙈 Even if there some indirection using another `AccessNode`, based on what I've seen until now, accessing this `AccessNode` will be done using only local indexes which we don't touch

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: Regarding your first point, I had no idea such thing is possible from our lowering but I removed the `other_subset = None` now so it's okay.
Regarding your second point, I think that the `if` statement is necessary because with the current changes we traverse the whole `memlet tree`

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -10,7 +10,7 @@
 
 import enum
 import warnings
-from typing import Any, Callable, Optional, Sequence, TypeAlias, Union
+from typing import Any, Callable, List, Optional, Sequence, TypeAlias, Union
```
philip-paul-mueller:
```suggestion
from typing import Any, Callable, Optional, Sequence, TypeAlias, Union
```

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -118,9 +118,11 @@ def gt_auto_optimize(
     gpu_block_size_2d: Optional[Sequence[int | str] | str] = None,
     gpu_block_size_3d: Optional[Sequence[int | str] | str] = None,
     gpu_maxnreg: Optional[int] = None,
-    blocking_dim: Optional[gtx_common.Dimension] = None,
+    blocking_dims: Optional[List[gtx_common.Dimension]] = None,
```
philip-paul-mueller:
```suggestion
    blocking_dims: Optional[Sequence[gtx_common.Dimension]] = None,
```

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -762,7 +762,7 @@ def apply(
                 block_size[i] = map_size[map_dim_idx_to_inspect]
 
         gpu_map.gpu_block_size = tuple(block_size)
-        if self.maxnreg is not None:
+        if self.maxnreg is not None and gpu_map.gpu_maxnreg == 0:
```
philip-paul-mueller, line 768: Just for my curiosity, what is the intention behind this change?

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -63,46 +67,87 @@ class LoopBlocking(dace_transformation.SingleStateTransformation):
         allow_none=True,
         desc="Size of the inner blocks; 'B' in the above description.",
     )
-    blocking_parameter = dace_properties.Property(
-        dtype=str,
+    blocking_parameters = dace_properties.Property(
+        dtype=list,
         allow_none=True,
```
philip-paul-mueller, line 76:
```suggestion
    blocking_parameters = dace_properties.ListProperty(
        dtype=str,
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -52,6 +54,8 @@ class LoopBlocking(dace_transformation.SingleStateTransformation):
         blocking_parameter: On which parameter should we block.
```
philip-paul-mueller: Needs to be updated as well.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -63,46 +67,87 @@ class LoopBlocking(dace_transformation.SingleStateTransformation):
...
+        self._memlet_to_promote = []
+        for out_edge_outer_entry in graph.out_edges(outer_entry):
+            if self._check_if_edge_can_be_promoted(
+                matched_blocking_var, out_edge_outer_entry, outer_entry
+            ):
+                self._memlet_to_promote.append(out_edge_outer_entry)
+
+    def _get_blocking_parameter(self, map_params: list[str]) -> Optional[str]:
+        if self.blocking_parameters is None:
+            return None
```
philip-paul-mueller:
```suggestion
            raise ValueError("No blocking parameters were specified.")
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -146,10 +195,32 @@ def can_be_applied(
         if (map_range_size[block_var_idx] <= self.blocking_size) == True:  # noqa: E712 [true-false-comparison]  # SymPy Fancy comparison.
             return False
 
-        if not self.partition_map_output(graph, sdfg):
+        if self.promote_independent_memlets:
+            self._populate_memlet_to_promote(matched_blocking_var, graph, outer_entry)
```
philip-paul-mueller, line 215: As far as I understand you need to know which nodes are dependent to make a decision to promote.
However, you first call the `_populate_memlet_to_promote()` before you call compute the partition.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -162,11 +233,21 @@ def apply(
...
         Performs the operation described in the doc string.
         """
+        outer_entry: dace_nodes.MapEntry = self.outer_entry
+        map_params: list[str] = outer_entry.map.params
+        matched_blocking_var: str | None = self._get_blocking_parameter(map_params)
+        assert matched_blocking_var is not None
+
         # Now compute the partitions of the nodes.
-        self.partition_map_output(graph, sdfg)
+        self.partition_map_output(matched_blocking_var, graph, sdfg, outer_entry)
```
philip-paul-mueller, line 259: Here the order of partition and then promote is correct or at least different from the one `can_be_applied()`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +687,225 @@ def _post_process_independent_nodes(
...
+        edge: dace_graph.MultiConnectorEdge[dace.Memlet],
+        outer_entry: dace_nodes.MapEntry,
+    ) -> bool:
+        """Check if a memlet can be promoted to the outer map.
+
+        The function checks if the memlet can be promoted to the outer map.
+        This is the case if the memlet does not depend on the blocking parameter.
+
+        Args:
+            edge: The edge containing the memlet to inspect.
```
philip-paul-mueller, line 722: some arguments are missing.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +687,225 @@ def _post_process_independent_nodes(
...
+        Returns:
+            The function returns `True` if the memlet can be promoted.
+        """
+        assert edge.src == outer_entry
+
+        memlet: dace.Memlet = edge.data
+        src_subset: dace_subsets.Subset | None = memlet.src_subset
+        dst_subset: dace_subsets.Subset | None = memlet.dst_subset
+
+        if self._independent_nodes is not None and edge.dst in self._independent_nodes:
```
philip-paul-mueller: This `is None` check should not be here and is needed because in `can_be_applied()` the Memlets are first checked for promotion before the partition is actually computed.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +687,225 @@ def _post_process_independent_nodes(
...
+        if any(matched_blocking_var in subset.free_symbols for subset in subsets_to_inspect):
+            return False
+
+        # If the memlet is connected to a MapEntry and the MapEnty parameters contain the blocking parameter,
+        # we could promote the memlet to the outer map but at the time there's no case where this can happen
+        # so we leave it as future work.
+        # TODO(iomaganaris): Implement this case if it turns out to be relevant.
+        if isinstance(edge.dst, dace_nodes.MapEntry):
+            if matched_blocking_var in edge.dst.params:
+                return False
```
philip-paul-mueller: This can not happen because the blocking parameter is already a Map parameter of the outer Map.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +687,225 @@ def _post_process_independent_nodes(
...
+
+        if isinstance(edge.dst, dace_nodes.LibraryNode):
+            # We currently do not handle promotion of memlets to library nodes, since it is not clear if this can actually happen in the cases we care about and it would require some work to handle the different cases.
+            warnings.warn(
+                "LoopBlocking: Memlet promotion to LibraryNode is not supported and will be skipped.",
+                stacklevel=2,
+            )
+            return False
+
+        if isinstance(edge.dst, dace_nodes.NestedSDFG):
```
philip-paul-mueller, line 774: The `AccessNode` check is missing.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +687,225 @@ def _post_process_independent_nodes(
...
+        state: dace.SDFGState,
+        sdfg: dace.SDFG,
+        outer_map_entry: dace_nodes.MapEntry,
+    ) -> None:
+        assert self._independent_nodes is not None  # silence MyPy
+        assert self._memlet_to_promote is None  # silence MyPy
+
+        _ = sdfg.reset_cfg_list()
+        dace_sdutils.canonicalize_memlet_trees_for_map(state=state, map_node=outer_map_entry)
+        dace_propagation.propagate_memlets_map_scope(sdfg, state, outer_map_entry)
```
philip-paul-mueller, line 808: I am not sure if Memlet propagation is needed here.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +687,225 @@ def _post_process_independent_nodes(
...
+
+        self._populate_memlet_to_promote(matched_blocking_var, state, outer_map_entry)
+        # Below checks are necessary for MyPy
+        if self._memlet_to_promote and len(self._memlet_to_promote) == 0:
```
philip-paul-mueller:
```suggestion
        if len(self._memlet_to_promote) == 0:
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +687,225 @@ def _post_process_independent_nodes(
...
+            return
+        assert self._memlet_to_promote is not None
+
+        for in_edge in self._memlet_to_promote:
+            if isinstance(in_edge.dst, dace_nodes.AccessNode):
+                raise NotImplementedError(
```
philip-paul-mueller: This case should be rejected already in the `populate_memlet_to_promote()` function.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +687,225 @@ def _post_process_independent_nodes(
...
+                    "Promotion of memlets to AccessNodes is not implemented because "
+                    "this case should already be handled since the destination AccesesNode "
+                    "should already be in the set of independent nodes."
+                )
+            # Create a temporary AccessNode that will be used to promote the memlet
+            promoted_accessnode_shape = []
+            # We have to adjust the subsets of the inner map out edges that are connected to the memlet we want to promote in case the destination of the in_edge is a MapEntry.
+            corresponding_inner_map_out_edges = list(
+                state.out_edges_by_connector(in_edge.dst, "OUT_" + in_edge.dst_conn[3:])
+            )
```
philip-paul-mueller: You only have connectors starting with `IN_` at Map nodes.
It can even be None.
So doing this operation, `dst_conn[3:]`, unprotected by an `if isinstance(node, MapEntry)` is most certainly wrong.

This function should probably look like:
```python=
inode = add_new_intermediate_storage()
add_edge(outer_map, inode, modified_memlet1())
e = add_edge(inode, original_edge.dst, dcpy(original_edge.data))
for mtree in state.memlet_tree(e).traverse(include_self=True):
    apply_correction(mtree.edge)
```

I think that the best idea would be to copy the logic of [`MapFusionVertical::compute_reduced_intermediate()`](https://github.com/spcl/dace/blob/3d76a660c5d770a15318066c48ab28e2428b6dd4/dace/transformation/dataflow/map_fusion_vertical.py#L752).

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -600,6 +687,225 @@ def _post_process_independent_nodes(
...
+        dace_sdutils.canonicalize_memlet_trees_for_map(state=state, map_node=outer_map_entry)
+        dace_propagation.propagate_memlets_map_scope(sdfg, state, outer_map_entry)
+
+        self._populate_memlet_to_promote(matched_blocking_var, state, outer_map_entry)
+        # Below checks are necessary for MyPy
+        if self._memlet_to_promote and len(self._memlet_to_promote) == 0:
+            return
+        assert self._memlet_to_promote is not None
+
+        for in_edge in self._memlet_to_promote:
```
philip-paul-mueller: I took the liberty and tried to come up with something different: https://github.com/iomaganaris/gt4py/pull/2
Feel free to modify.

philip-paul-mueller (commented):
Some additional comments.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
iomaganaris: To make sure that we don't overwrite the value set from some other transformation. My friend left a comment about that

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: I think this is there so that mypy doesn't complain about the second check

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: I thought that the two calls are probably needed to make sure we don't have any strange state of memlets from previous transformations

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: I changed that to:
```
if not self._memlet_to_promote:
            return
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: I left my friend take care of the following based on your changes. I hope it's  okay now 😅 Same for the comments above

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -67,46 +72,87 @@ class LoopBlocking(dace_transformation.SingleStateTransformation):
...
         desc="Size of the inner blocks; 'B' in the above description.",
     )
-    blocking_parameter = dace_properties.Property(
-        dtype=str,
+    blocking_parameters = dace_properties.ListProperty(
+        element_type=str,
         allow_none=True,
-        desc="Name of the iteration variable on which to block (must be an exact match);"
-        " 'I' in the above description.",
+        desc="Names of the iteration variables on which to block. The first one that is found in the map parameters is used",
```
philip-paul-mueller:
```suggestion
        desc="Names of the iteration variables on which to block. The first one that is found in the map parameters is used.",
```
Something is missing in this sentences like "checked in order they are listed" but I do not care much.
However, you should mention it here and in the doc string, that a match only handles a single parameter.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
philip-paul-mueller: Actually I just realized that you were right and there should be an error here.
However, because of the guard in `can_be_applied()` we can condense it to `assert self.blocking_parameters`

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -122,38 +168,83 @@ def can_be_applied(
         - The map range must have step size of 1.
         - The partition must exists (see `partition_map_output()`).
         """
-        if self.blocking_parameter is None:
-            raise ValueError("The blocking dimension was not specified.")
+        if self.blocking_parameters is None:
```
philip-paul-mueller:
```suggestion
        if not self.blocking_parameters:
```
This detects the empty list and the None case.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -122,38 +168,83 @@ def can_be_applied(
...
+            # Compute the partition first so that the independent nodes are known
+            # when checking if memlets can be promoted.
+            partition_succeeded = self.partition_map_output(
+                matched_blocking_var, graph, sdfg, outer_entry
+            )
+            self._populate_memlet_to_promote(matched_blocking_var, graph, outer_entry)
+        else:
+            partition_succeeded = self.partition_map_output(
+                matched_blocking_var, graph, sdfg, outer_entry
+            )
```
philip-paul-mueller:
```python
if not self.partition_map_output(matched_blocking_var, graph, sdfg, outer_entry):
    return False

if self.promote_independent_memlets:
    # Partition must be computed first.
    self._populate_memlet_to_promote(matched_blocking_var, graph, outer_entry)

File: src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py [hunk 42]
```diff
@@ -122,38 +168,83 @@ def can_be_applied(
...
+        ):
+            return False
+
+        # Check if the blocking is a good idea.
+        if not self._check_if_blocking_can_promote_anything(state=graph):
             return False
+
+        # Disable by default scans because there is `ScanLoopUnrolling` for them (if the blocking is done in the same dimension as the scan).
+        # Otherwise blocking the loop in the other dimension shouldn't be beneficial.
```
philip-paul-mueller:
```suggestion
        # Disable by default scans because there is `ScanLoopUnrolling` for them
        # (if the blocking is done in the same dimension as the scan).
        # Otherwise blocking the loop in the other dimension shouldn't be beneficial.
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -122,38 +168,83 @@ def can_be_applied(
...
+                matched_blocking_var, graph, sdfg, outer_entry
+            )
+            self._populate_memlet_to_promote(matched_blocking_var, graph, outer_entry)
+        else:
+            partition_succeeded = self.partition_map_output(
+                matched_blocking_var, graph, sdfg, outer_entry
+            )
+
+        if (
+            not partition_succeeded
```
philip-paul-mueller:
```suggestion
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -420,7 +527,7 @@ def _classify_node(
 
         # Despite its type the node's free symbols can not contain the blocking
```
philip-paul-mueller:
```suggestion
        # Regardless of its type the node's free symbols can not contain the blocking
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -604,6 +711,211 @@ def _post_process_independent_nodes(
...
+        src_subset: dace_subsets.Subset | None = memlet.src_subset
+        dst_subset: dace_subsets.Subset | None = memlet.dst_subset
+
+        if edge.dst in self._independent_nodes:
+            # If the memlet is connected to an independent node, then we can not promote it, since it would be redundant.
+            return False
+
+        if (
+            memlet.is_empty()
+        ):  # Empty Memlets should already be in independent nodes and don't have read dependencies
```
philip-paul-mueller:
```suggestion
        # Empty Memlets should already be in independent nodes and don't have read dependencies
        if memlet.is_empty():
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -604,6 +711,211 @@ def _post_process_independent_nodes(
...
+        if any(matched_blocking_var in subset.free_symbols for subset in subsets_to_inspect):
+            return False
+
+        if isinstance(edge.dst, dace_nodes.AccessNode):
+            # Promotion of memlets to AccessNodes is not needed because the
+            #  destination AccessNode should already be in the set of
+            #  independent nodes.
+            return False
+
+        if isinstance(edge.dst, dace_nodes.Tasklet) and not edge.data.data.startswith("gt_conn_"):
```
philip-paul-mueller, line 755: Maybe we should also check the connector name, i.e. adding something like `not edge.dst_conn.startswith("__tlet_index")`, but I am not fully sure if it is a good idea and we can skip it.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -604,6 +711,211 @@ def _post_process_independent_nodes(
...
+
+            original_data = edge_to_promote.data.data
+            original_data_desc = sdfg.arrays[original_data]
+
+            promote_subset: dace_subsets.Subset = (
+                edge_to_promote.data.subset
+            )  # Works because of canonicalization.
+            # TODO(phimuell): Handle the removal of pseudo dimensions.
+            buffer_shape = tuple(dace_symb.overapproximate(promote_subset.size()))
+            assert all(str(x).isdigit() for x in buffer_shape)
```
philip-paul-mueller: Maybe we should move this check into `_check_if_edge_can_be_promoted()` there we could also ensure that the size is not too big, because currently we would also promote something with the size `a[0:10000000000000, 0:100000000000000]`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -604,6 +711,211 @@ def _post_process_independent_nodes(
...
+                )
+                state.remove_edge(edge_to_promote)
+
+                for mtree in state.memlet_tree(new_consumer_edge).traverse_children(True):
+                    edge_to_correct = mtree.edge
+                    assert edge_to_correct.data.data == original_data  # Because of canonicalize.
+                    edge_to_correct.data.data = buffer_data
+                    edge_to_correct.data.subset.offset(buffer_offset, negative=True)
+
+                continue
```
philip-paul-mueller, line 857: I am actually not sure if we should turn this thing into an `else:` for readability.
But I actually do not care.
Probably add a note or something.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -604,6 +711,211 @@ def _post_process_independent_nodes(
...
+                    buffer_offset_list.append(dim_range[0])
+                else:
+                    # Was not accessed by a Map parameter.
+                    buffer_offset_list.append((lower_bound, lower_bound, 1))
+            buffer_offset = dace_subsets.Range(buffer_offset_list)
+
+            promoted_buffers[dedup_key] = (
+                buffer_node,
+                buffer_data,
+                buffer_desc,
```
philip-paul-mueller: Technically you do not need to store `desc` because with the name you can always retrieve it from `sdfg.array`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -604,6 +711,211 @@ def _post_process_independent_nodes(
...
+                buffer_node,
+                None,
+                dace.Memlet(
+                    data=original_data,
+                    subset=copy.deepcopy(promote_subset),
+                    other_subset=dace_subsets.Range.from_array(buffer_desc),
+                ),
+            )
+
+            # Create the old (uncorrected) consumer.
```
philip-paul-mueller:
```suggestion
```

tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_loop_blocking.py
```diff
@@ -1608,3 +1608,281 @@ def test_loop_blocking_direct_access_node_scalar():
...
+    sdfg.add_array("gt_conn_dummy", shape=(40, 2), dtype=dace.int32, transient=False)
+    sdfg.add_array("ikoffset", shape=(40,), dtype=dace.int32, transient=False)
+    sdfg.add_array("S", shape=(40,), dtype=dace.float64, transient=False)
+    sdfg.add_scalar("t", dtype=dace.float64, transient=True)
+    sdfg.add_scalar("tt", dtype=dace.float64, transient=True)
+    sdfg.add_scalar("ttt", dtype=dace.float64, transient=True)
+    sdfg.add_scalar("tttt", dtype=dace.float64, transient=True)
+    sdfg.add_scalar("ttttt", dtype=dace.float64, transient=True)
+    sdfg.add_scalar("tttttt", dtype=dace.float64, transient=True)
+    sdfg.add_scalar("ttttttt", dtype=dace.float64, transient=True)
```
philip-paul-mueller, line 1633: I like the naming 😺

philip-paul-mueller (commented):
I have some small comments, but in general it looks okay.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: Yes.. I think it makes more sense to through an error so that we don't call the transformation that traverses every map entry if there's no need for it to run.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: Maybe this restricts the check more? We might decide to change the `__tlet_index` name but not the `gt_conn` so this would break even if the behavior is the same

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: Yes, good point 👍

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -736,12 +742,14 @@ def _gt_auto_process_dataflow_inside_maps(
     # Separate Tasklets into dependent and independent parts to promote data
     #  reusability. It is important that this step has to be performed before
     #  `TaskletFusion` is used.
-    if blocking_dim is not None:
+    if blocking_dims is not None and blocking_size is not None and blocking_size > 0:
```
philip-paul-mueller: According to the type annotation `blocking_size` is not `None`.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -53,9 +56,11 @@ class LoopBlocking(dace_transformation.SingleStateTransformation):
 
     Args:
         blocking_size: The size of the block, denoted as `B` above.
-        blocking_parameter: On which parameter should we block.
+        blocking_parameters: On which parameter should we block.
```
philip-paul-mueller:
```suggestion
        blocking_parameters: On which parameter should we block Priority based on the order of the passed dimensions.
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -67,46 +72,86 @@ class LoopBlocking(dace_transformation.SingleStateTransformation):
...
+        element_type=str,
         allow_none=True,
-        desc="Name of the iteration variable on which to block (must be an exact match);"
-        " 'I' in the above description.",
+        desc="Names of the iteration variables on which to block checked in order they are listed. Only the first one that is found is used for blocking.",
     )
     require_independent_nodes = dace_properties.Property(
         dtype=bool,
         default=True,
         desc="If 'True' then blocking is only applied if there are independent nodes.",
```
philip-paul-mueller: I think we can drop all these `desc` things, as we have a description in the doc string.
This is only used by the visual studio code extension.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -604,6 +706,218 @@ def _post_process_independent_nodes(
...
+            The function returns `True` if the memlet can be promoted.
+        """
+        assert edge.src == outer_entry
+        assert self._independent_nodes is not None  # silence MyPy
+
+        memlet: dace.Memlet = edge.data
+        src_subset: dace_subsets.Subset | None = memlet.src_subset
+        dst_subset: dace_subsets.Subset | None = memlet.dst_subset
+
+        if edge.dst in self._independent_nodes:
```
philip-paul-mueller:
```suggestion
        if edge.dst in self._independent_nodes or isinstance(edge.dst, AccessNodes):
```

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -604,6 +706,218 @@ def _post_process_independent_nodes(
...
+            # (connection name: `__tlet_index_Cell` and `__tlet_index_K`).
+            # If there is no `K` dependent offset if we have blocking on `K`, we can promote the memlet.
+            # If there is no `Cell` dependent offset if we have blocking on `Cell`, we can promote the memlet.
+            # However above require checking the tasklet internals without being sure of these conventions.
+            # For that purpose we choose to promote only memlets that are passed to tasklets only
+            # if they are connectivities.
+            return False
+
+        if isinstance(edge.dst, dace_nodes.LibraryNode):
+            # We currently do not handle promotion of memlets to library nodes, since it is not clear if this can actually happen in the cases we care about and it would require some work to handle the different cases.
```
philip-paul-mueller, line 767: I think here a line break would help.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
```diff
@@ -604,6 +706,218 @@ def _post_process_independent_nodes(
...
+            return False
+
+        promote_subset: dace_subsets.Subset = edge.data.subset  # Works because of canonicalization.
+        buffer_shape = tuple(dace_symb.overapproximate(promote_subset.size()))
+        if not all(str(x).isdigit() for x in buffer_shape):
+            return False
+        for buffer_dim in buffer_shape:
+            # We don't want to promote memlets that are too big because then the register or stack usage would increase a lot and we will get worse performance.
+            # TODO(iomaganaris): Maybe find a better suited number
+            if buffer_dim > 100:
```
philip-paul-mueller: Maybe we should introduce a "constexpr static variable to give it a name".

philip-paul-mueller (approved):
LGTM.

src/gt4py/next/program_processors/runners/dace/transformations/loop_blocking.py
iomaganaris: I tried conveying this with a `Final` class variable. Probably it's enough

PR #2585, opened 2026-04-24
fix[cartesian] Use `utf-8` to encore python module code 

FlorianDeconinck:
## Description

An (odd) bug manifested for the first time on the HPC on a GPU build: Using non-ascii characters in docstrings (here a `²`), crashed writing the generated code.

The issue was tracked down the following: In GT4Py, if stencils have docstrings, that docstring is copied over into the generated source code. On the HPC, the default encoding is ascii (for whatever reason). Now when the generated code (including the non-ascii characters in the description) was written down to disk, writing failed because the non-ascii characters could not be represented.

The PR proposes to write and read all code in `utf-8` to alleviate the issue.

## Requirements


romanc (approved):
LGTM. Found and fixes some more occurrences of `{read,write}_text()` without encoding.

PR #2487, opened 2026-02-20
feat[next]: Add support for tuple comprehensions (not stacked)

SF-N:
Adds support for tuple comprehensions, e.g. for usage on tracers.
```python
@gtx.field_operator
def testee(
    tracers: tuple[cases.IField, cases.IField], factor: int32
) -> tuple[cases.IField, cases.IField]:
    return tuple(tracer * factor for tracer in tracers)
```

PR #2586, opened 2026-04-27
feat[next]: Tracer support part 1: tree_map

SF-N:
As a first step towards tracer support (enabling vector operations), this PR adds the infrastructure for mapping a function over the elements of a tuple.

### What this PR does

**New ITIR builtins** (`iterator/builtins.py`), each taking a single tuple argument:
- `tree_map_tuple(f)(t)` — recursively maps `f` over the (possibly nested) tuple structure of `t`, applying it at the leaves.
- `map_tuple(f)(t)` — maps `f` over the top-level elements of `t` only (no recursion).

**New pass `ExpandTupleMaps`** (`iterator/transforms/expand_tuple_maps.py`) — expands both builtins into `make_tuple` / `tuple_get` calls. It runs early in both `apply_common_transforms` and `apply_fieldview_transforms` (`iterator/transforms/pass_manager.py`).

**Type synthesizers** for the new builtins (`iterator/type_system/type_synthesizer.py`).

**Rename `map_` → `map_list`** across the codebase, to disambiguate list/neighbor mapping from the new tuple-mapping operators.

### Not yet included

The frontend lowering (`ffront/foast_to_gtir.py`) does not emit these builtins yet — `where` / casts still expand tuples statically via `process_elements` (see the TODOs there). This PR only lands the builtins, type inference, and expanding infrastructure that subsequent tracer-support steps will build on.


src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -176,6 +177,26 @@ def apply_common_transforms(
...
+                CollapseTuple.Transformation.PROPAGATE_TUPLE_GET
+                | CollapseTuple.Transformation.COLLAPSE_TUPLE_GET_MAKE_TUPLE
+            ),
+            uids=uids,
+            offset_provider_type=offset_provider_type,
+        )  # type: ignore[assignment]  # always an itir.Program
+        if ir == collapsed:
+            break
+    else:
+        raise RuntimeError("'CollapseTuple' did not converge after `UnrollTreeMap`.")
```
SF-N: Without this `test_reduction_expression_with_where_and_tuples` fails with `ValueError: 'target_domain' cannot be 'NEVER' unless "allow_uninferred=True".`

src/gt4py/next/ffront/foast_to_gtir.py
SF-N, line 392: I think  `_visit_astype` cannot use `tree_map `because it needs type-dependent lowering per leaf: fields use `_map(cast)` while scalars use `cast` directly. Let's discuss if you have something else in mind.

SF-N, line 439: As far as I see, `_visit_concat_where` already has its own `expand_tuple_args` pass for handling nested tuples, and each branch can have a different domain. Attempting to wrap it in `tree_map` caused type inference failures. Let's discuss if you have something else in mind.

src/gt4py/next/iterator/transforms/unroll_tree_map.py
```diff
@@ -0,0 +1,85 @@
...
+
+        tup_refs = [next(self.uids["_utm"]) for _ in tup_args]
+        body = _unroll(f, tup_types, [im.ref(r) for r in tup_refs])
+
+        result = body
+        for ref_name, tup in reversed(list(zip(tup_refs, tup_args))):
+            result = im.let(ref_name, tup)(result)
+
+        itir_inference.reinfer(result)
+        return result
```
tehrengruber:
```suggestion
        tup_refs = [next(self.uids["_utm"]) for _ in tup_args]
        @tree_map(
          collection_type=ts.TupleType
          result_collection_constructor=im.make_tuple,
          with_path_arg=True
        )
        def mapper(*el_types, path: tuple[int, ...]):
          return im.call("unique_name_for_f")(*(operator.reduce(lambda expr, i: im.tuple_get(expr, i), path, tuple_ref) for tuple_ref in tuple_refs))
        return im.let(*zip(tuple_refs, tup_args))(
          mapper(*tup_types)
        )
```
`tree_map` is a perfect fit for this.

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -633,6 +633,47 @@ def applied_map(
     return applied_map
 
 
+@_register_builtin_type_synthesizer(fun_names=["tree_map"])
+def _tree_map(op: TypeSynthesizer) -> TypeSynthesizer:
```
tehrengruber:
```suggestion
@_register_builtin_type_synthesizer
def tree_map(op: TypeSynthesizer) -> TypeSynthesizer:
```
Let's instead remove the import that shadows.

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -633,6 +633,47 @@ def applied_map(
...
+    @type_synthesizer
+    def applied_map(
```
tehrengruber: The `offset_provider_type` needs some care, but `tree_map` should also be a good fit here.
```suggestion
    @utils.tree_map
    def applied_map(
```

src/gt4py/next/ffront/foast_to_gtir.py
tehrengruber: Other reasons: A single `concat_where` is better to digest for optimizations. The lowering would actually be complicated if we would emit tree_map in foast_to_gtir.

src/gt4py/next/iterator/transforms/pass_manager.py
tehrengruber: Note: probably this is also the test case where the loop is required. I'll take a look if another configuration of the pass helps to avoid the loop.

src/gt4py/next/ffront/foast_to_gtir.py
```diff
@@ -412,23 +412,15 @@ def _visit_where(self, node: foast.Call, **kwargs: Any) -> itir.FunCall:
...
-        def create_if(
-            true_: itir.Expr, false_: itir.Expr, arg_types: tuple[ts.TypeSpec, ts.TypeSpec]
-        ) -> itir.FunCall:
-            return _map(
-                "if_",
-                (im.ref(cond_symref_name), true_, false_),
-                (node.args[0].type, *arg_types),
+        result = im.tree_map_tuple(
+            im.lambda_("__a", "__b")(
+                im.op_as_fieldop("if_")(im.ref(cond_symref_name), im.ref("__a"), im.ref("__b"))
```
havogt: **Tuple-`where` over a per-neighbour (local) condition produces invalid IR.**

Hardcoding the per-leaf op to `op_as_fieldop("if_")` drops the local-field handling the non-tuple path keeps: `_lower_and_map("if_", …)` → `_map` wraps the op in `map_list(if_)` and promotes the condition with `make_const_list` when the leaves are local fields. The single, uniform `tree_map_tuple` leaf can't do that.

This is reachable from the frontend and breaks on **roundtrip, gtfn and dace** whenever the `where` condition is itself a local (per-neighbour) field — the mask only has to be a `FieldType`, and a local field qualifies:

```python
@gtx.field_operator
def tup(a: EdgeF, b: EdgeF, c: EdgeF, d: EdgeF) -> tuple[VertF, VertF]:
    cond = a(V2E) > c(V2E)                               # per-neighbour (local) bool field
    t = where(cond, (a(V2E), b(V2E)), (c(V2E), d(V2E)))
    return (neighbor_sum(t[0], axis=V2EDim), neighbor_sum(t[1], axis=V2EDim))
```

The non-tuple equivalent (`neighbor_sum(where(a(V2E)>c(V2E), a(V2E), c(V2E)), …)`) works on all three backends; the tuple version fails on all three with `AssertionError` in the `if_` type synthesizer (`isinstance(pred, ts.ScalarType) and pred.kind == ts.ScalarKind.BOOL`), because after `UnrollTupleMaps` the leaf is `⇑(λ(c,x,y) → if ·c then ·x else ·y)` applied to a **list-typed predicate** — no `map_list`.

Broadcast (non-local) conditions happen to pass, since wholesale `if_(scalar, list, list)` equals element-wise selection and the consuming reduce absorbs the list — only a *varying* condition exposes it.

This is the same per-leaf type-dependent reason given for not migrating `_visit_astype`. It also isn't fixable by "always emit `map_list`": with a non-local cond and mixed leaves (one branch local, one not), no single uniform leaf lambda is correct. Options: keep `where` on `process_elements`, or wrap `if_` in `map_list` post-unroll when the predicate is a list.

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -633,6 +632,53 @@ def applied_map(
...
+                raise TypeError(
+                    f"'{builtin_name}' requires all top-level arguments to be TupleType, "
+                    f"got {[type(a).__name__ for a in args]}."
+                )
+
+            def leaf_op(*leaf_types: ts.TypeSpec) -> ts.TypeSpec:
+                return op(*leaf_types, offset_provider_type=offset_provider_type)  # type: ignore[return-value]
+
+            if recursive:
+                return utils.tree_map(  # type: ignore[return-value]
```
havogt, line 649: **Mismatched tuple structure raises a bare `AssertionError` instead of a clear `TypeError`.**

The check above guarantees all args are `TupleType`, but not that they share arity/nesting. When they differ, `utils.tree_map` trips its internal `assert ... len(args[0]) == len(arg)` (or the all-collection assert during recursion) and raises an `AssertionError` with no message:

```python
im.tree_map_tuple(im.ref("plus"))(
    im.ref("t1", ts.TupleType(types=[int_t, int_t, int_t])),
    im.ref("t2", ts.TupleType(types=[int_t, int_t])),
)   # -> AssertionError (empty message); also for mismatched nesting
```

Not reachable from the frontend today — `where` pre-validates branch structure with a `DSLError` — but it'll bite the upcoming `map_tuple`/tracer producers and anyone building IR directly. A `TypeError` here (and in the matching spot in `UnrollTupleMaps`) would be friendlier.

havogt (commented):
Two findings from reviewing this PR and verifying locally against the branch (simple_mesh, all backends). Details inline.

tehrengruber (commented):
The following lowering strategy does not work right now:
`(2*el for el in (a(V2E), b))` -> `map_tuple(lambda el: op_as_fieldop(*)(2, el))({a_field_with_local_dim, b_field_of_scalars})` since the `el` symref node can only have one type and since we can not distinguish (i.e. no isinstance on GTIR level) between them.
However since we know the length we can use the `process_elements` approach like for `where`. In case we don't know the length the element type is always the same and we can use `map_tuple` in the lowering and then fold it on GTIR level.
```suggestion
(2*el for el in var_length_tuple_scalar) -> map_tuple(lambda el: 2*el)(var_length_tuple)
(2*el for el in var_length_tuple_of_field) -> map_tuple(lambda el: as_fieldop(lambda el_at_grid_point: 2*el_at_grid_point)(el))(var_length_tuple)
(2*el for el in var_length_tuple_field_with_local_dim) -> map_tuple(lambda el: map_list(lambda in_el: 2*in_el)(el))(tuple_of_it_of_list)
```
Conclusion:
We continue / merge this PR (in the previous form and after review). In https://github.com/GridTools/gt4py/pull/2487 we can then use `map_tuple` in the lowering for the variable length cases and only need process elements for the fixed length case. We also need tests to cover all cases (i.e. testing for `where`, `concat_where`, tuple comprehension of elements with mixed type like `(a(V2E), b)`).

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -169,6 +170,23 @@ def apply_common_transforms(
...
+    # arguments had to be let-bound (non-trivial sub-expressions).
+    ir = CollapseTuple.apply(
+        ir,
+        enabled_transformations=(
+            CollapseTuple.Transformation.PROPAGATE_TUPLE_GET
+            | CollapseTuple.Transformation.COLLAPSE_TUPLE_GET_MAKE_TUPLE
+        ),
+        uids=uids,
+        offset_provider_type=offset_provider_type,
+    )  # type: ignore[assignment]  # always an itir.Program
```
```suggestion
    ir = unroll_tuple_maps.UnrollTupleMaps.apply(ir, uids=uids)
```

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -282,6 +300,19 @@ def apply_fieldview_transforms(
...
+    # See note in `apply_common_transforms` about why a single `CollapseTuple` pass suffices.
+    ir = CollapseTuple.apply(
+        ir,
+        enabled_transformations=(
+            CollapseTuple.Transformation.PROPAGATE_TUPLE_GET
+            | CollapseTuple.Transformation.COLLAPSE_TUPLE_GET_MAKE_TUPLE
+        ),
+        uids=uids,
+        offset_provider_type=offset_provider_type,
+    )  # type: ignore[assignment]  # always an itir.Program
```
```suggestion
    ir = unroll_tuple_maps.UnrollTupleMaps.apply(ir, uids=uids)
```
Here the tree is already infered once, but to be 100% clean the pass should call itir type inference when the program has not been inferred yet.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_unroll_tuple_maps.py
```diff
@@ -0,0 +1,255 @@
...
+    return im.lambda_("__a")(im.op_as_fieldop("neg")("__a"))
+
+
+def _plus():
+    return im.lambda_("__a", "__b")(im.op_as_fieldop("plus")("__a", "__b"))
+
+
+def test_tree_map_tuple_multi_arg():
+    uids = utils.IDGeneratorPool()
+    program = _make_program(
```
tehrengruber: Let's try to avoid wrapping everything in a program.

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,122 @@
...
+class UnrollTupleMaps(eve.NodeTranslator):
+    """Unroll tuple-map ITIR builtins (`tree_map_tuple`, `map_tuple`) into `make_tuple`."""
+
+    PRESERVED_ANNEX_ATTRS = ("domain",)
+
+    uids: utils.IDGeneratorPool
+
+    @classmethod
+    def apply(cls, program: itir.Program, *, uids: utils.IDGeneratorPool):
+        return cls(uids=uids).visit(program)
```
tehrengruber:
```suggestion
    def apply(cls, node: itir.Program | itir.Expr, *, uids: utils.IDGeneratorPool | None):
        if not node.type: # double check this signifies no type inference has run yet
          # run type inference
        return cls(uids=uids).visit(node)
```

src/gt4py/next/ffront/foast_to_gtir.py
tehrengruber: Contrary to the tuple comprehension case where we can not type the `el_expr` in `el_expr for ... in ...` and hence have to disallow cases with mixed element types, this is not the case here. The `process_elements` approach works for the case above. Let's add a test (if not already done) and revert to the previous strategy. Additionally a comment that we would like to use `tree_map_tuple` and why this is not possible (or a reference) would be good. If you want to be thorough create an ADR that describes the strategy for all the cases and reference that. https://hackmd.io/PxqTkA4rSGCgCmUAPq15SA could serve as a starting point.

src/gt4py/next/ffront/foast_to_gtir.py
tehrengruber:
```suggestion
    def _visit_concat_where(self, node: foast.Call, **kwargs: Any) -> itir.FunCall:
        # TODO(tehrengruber): Use `tree_map_tuple` when the domain inference is able to handle
        #  lambda functions (with the results domain depending on the caller / args)
```

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -169,6 +170,11 @@ def apply_common_transforms(
     ir = inline_lifts.InlineLifts().visit(ir)
 
     ir = concat_where.expand_tuple_args(ir, offset_provider_type=offset_provider_type)  # type: ignore[assignment]  # always an itir.Program
+    # `UnrollTupleMaps` requires fully-inferred tuple types (relies on `reinfer` to see
+    # nested `TupleType` chains), so the offset_provider is passed for on-demand inference.
```
```suggestion
```

src/gt4py/next/iterator/transforms/pass_manager.py
```diff
@@ -282,6 +288,12 @@ def apply_fieldview_transforms(
     ir = inline_fundefs.prune_unreferenced_fundefs(ir)
     # required for dead-code-elimination and `prune_empty_concat_where` pass
     ir = concat_where.expand_tuple_args(ir, offset_provider_type=offset_provider_type)  # type: ignore[assignment]  # always an itir.Program
+    # `UnrollTupleMaps` requires fully-inferred tuple types, so the offset_provider is passed for
+    # on-demand inference.
```
```suggestion
```

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,140 @@
...
+    PRESERVED_ANNEX_ATTRS = ("domain",)
+
+    uids: utils.IDGeneratorPool
+
+    @classmethod
+    def apply(
+        cls,
+        node: ProgramOrExpr,
+        *,
+        uids: utils.IDGeneratorPool | None = None,
```
tehrengruber:
```suggestion
        uids: utils.IDGeneratorPool | None,
```

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,140 @@
...
+        f = node.fun.args[0]
+        tup_args = node.args
+
+        tup_types: list[ts.TupleType] = []
+        for tup in tup_args:
+            itir_inference.reinfer(tup)
+            assert isinstance(tup.type, ts.TupleType)
+            tup_types.append(tup.type)
+
+        # For trivial args (those that can be duplicated without cost or side effects),
```
tehrengruber: Let's reduce complexity here and rely on tested components: Create the let and then run non-recursive CollapseTuple with only INLINE_TRIVIAL_TUPLE_LET_VAR. I'll create a PR for the non-recursive functionality.

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,140 @@
...
+        if uids is None:
+            uids = utils.IDGeneratorPool()
+        return cls(uids=uids).visit(node)
+
+    def visit_FunCall(self, node: itir.FunCall):
+        node = self.generic_visit(node)
+
+        builtin_name = next((name for name in _UNROLLERS if cpm.is_call_to(node.fun, name)), None)
+        if builtin_name is None:
+            return node
```
tehrengruber: Slop

```suggestion
        if not cpm.is_call_to(node.fun, ("map_tuple", "tree_map_tuple")):
          return node
```

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,140 @@
...
+
+def _tree_map_tuple_body(
+    f: itir.Expr, tup_exprs: list[itir.Expr], tup_types: list[ts.TupleType]
+) -> itir.Expr:
+    """Recursively unroll `tree_map_tuple(f)(t1, ..., tN)` into `make_tuple` calls."""
+
+    def tuple_structure(type_: ts.TypeSpec) -> tuple[object, ...] | None:
+        if isinstance(type_, ts.TupleType):
+            return tuple(tuple_structure(el_type) for el_type in type_.types)
+        return None
```
tehrengruber:
```suggestion
    tree_map_type(lambda _: None)(...)
```
This is a tree map itself.

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -629,6 +628,67 @@ def applied_map(
...
+    builtin_name: str, *, recursive: bool
+) -> Callable[..., TypeOrTypeSynthesizer]:
+    """Shared implementation for `tree_map_tuple` (recursive) and `map_tuple` (top-level)."""
+
+    def tuple_structure(type_: ts.TypeSpec) -> tuple[object, ...] | None:
+        if isinstance(type_, ts.TupleType):
+            return tuple(tuple_structure(el_type) for el_type in type_.types)
+        return None
+
+    def ensure_same_tuple_structure(args: tuple[ts.TupleType, ...]) -> None:
```
tehrengruber: Duplicated code.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_unroll_tuple_maps.py
```diff
@@ -0,0 +1,173 @@
...
+def _apply(expr: itir.Expr) -> itir.Expr:
+    return UnrollTupleMaps.apply(expr, uids=utils.IDGeneratorPool(), offset_provider_type={})
+
+
+def _neg():
+    return im.lambda_("__a")(im.op_as_fieldop("neg")("__a"))
+
+
+def _plus():
+    return im.lambda_("__a", "__b")(im.op_as_fieldop("plus")("__a", "__b"))
```
tehrengruber:
```suggestion
@pytest.fixture
def _unary_fun():
    return im.ref("unary_f", type=ts.FunctionType(...))

@pytest_fixture
def _binary_fun():
    return im.ref("binary_f", type=ts.FunctionType(...))
```

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,140 @@
...
+    def mapper(*args):
+        *_el_types, path = args
+        return im.call(f)(
+            *(functools.reduce(_collapsing_tuple_get, path, tup_expr) for tup_expr in tup_exprs)
+        )
+
+    return mapper(*tup_types)
+
+
+def _map_tuple_body(
```
tehrengruber: Maybe we should make `map_tuple` support multiple args like `tree_map_tuple`.

Open question: Does using multiple args in the lowering improve what collapse tuple can do before the `tree_map_tuple` can do? Then let's use that path even for the `... for tuple in ...` in the frontend.

tehrengruber (commented):
Partial review only.

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
tehrengruber: Here is the PR: https://github.com/GridTools/gt4py/pull/2689

SF-N: This doesn't seem to work with only INLINE_TRIVIAL_TUPLE_LET_VAR and with recursive=False:
```
 tree_map_tuple(neg)({a, b})                                                                                                                                                               
       → let(v, {a,b})( {neg(v[0]), neg(v[1])} )     # always let-bind                                                                                                                         
       → INLINE_TRIVIAL_TUPLE_LET_VAR (non-recursive):                                                                                                                                         
           {neg({a,b}[0]), neg({a,b}[1])}            # v inlined, but v[i] NOT folded   
```
Expected: {neg(a), neg(b)}. Folding {a,b}[0] → I think it would require  INLINE_TRIVIAL_MAKE_TUPLE | COLLAPSE_TUPLE_GET_MAKE_TUPLE and recursive=True.

tehrengruber: Then let's skip this and just run CollapseTuple in the tests. Since we run dead-code elimination after this pass it is not required anyway.

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
tehrengruber: `tuple[0]+tuple[1] for tuple in nested_tuple`
or
`let(tuple, {a, b})(tuple[0]+tuple[1])  for a, b in {nested_tuple[0], nested_tuple[1]}`
Another dimension: `a+b for a, b in nested_tuple`
The interesting parts are what is the advantage / disadvantage in the lowering & unroll pass.

SF-N: All examples single-arg `map_tuple(op)(n)` over eg.g `n : ((Field[[IDim], float64], Field[[IDim], float64]),(Field[[IDim], float64],Field[[IDim], float64]))`, and they collapse to the identical result:

1. `t[0] + t[1] for t in n`: 
lowered:                                                                         
`map_tuple(λ(t) → t[0] + t[1])(n)`
before collapse:                                                                 
`(λ(_utm_0) → {(λ(t) → t[0] + t[1])(_utm_0[0]),(λ(t) → t[0] + t[1])(_utm_0[1])})(n)`
after collapse:                                                                  
`{n[0][0] + n[0][1], n[1][0] + n[1][1]}`

2. `let(tup, {a, b})(tup[0] + tup[1]) for a, b in {n[0], n[1]}`:              
lowered:                                                                         
`map_tuple(λ(t) → (λ(a) → (λ(b) → (λ(tup) → tup[0] + tup[1])({a, b}))(t[1]))(t[0]))({n[0], n[1]})`                                                                  
before collapse:                                                                 
`(λ(_utm_0) → {(λ(t) → (λ(a) → (λ(b) → (λ(tup) → tup[0] + tup[1])({a, b}))(t[1]))(t[0]))(_utm_0[0]), (λ(t) → (λ(a) → (λ(b) → (λ(tup) → tup[0] + tup[1])({a, b}))(t[1]))(t[0]))(_utm_0[1])})({n[0], n[1]})`   
after collapse:                                                                  
`{n[0][0] + n[0][1], n[1][0] + n[1][1]}`

3. `a + b for a, b in n`:                                                     
lowered:                                                                         
`map_tuple(λ(t) → (λ(a) → (λ(b) → a + b)(t[1]))(t[0]))(n)`                    
before collapse:                                                                 
`(λ(_utm_0) → {(λ(t) → (λ(a) → (λ(b) → a + b)(t[1]))(t[0]))(_utm_0[0]), (λ(t) → (λ(a) → (λ(b) → a + b)(t[1]))(t[0]))(_utm_0[1])})(n)`     
after collapse:                                                                  
`{n[0][0] + n[0][1], n[1][0] + n[1][1]}`

All three collapse to the same result, so the choice only affects the intermediate size that the unroll+collapse passes must remove.

There is no use-case for multi-arg map_tuple so far and also not in [#PR2487 ](https://github.com/GridTools/gt4py/pull/2487/), whereas for tree_map_tuple the multi-arg use-case is `where(cond, a, b)`.

Nevertheless, I tried implementing multi-arg map_tuple and no changes in the lowering were necessary, only small changes in the type_synthesizer, ir_makers and in the unroll_tuple_maps replacing

```
def _map_tuple_body(
    f: itir.Expr, tup_exprs: list[itir.Expr], tup_types: list[ts.TupleType]
) -> itir.Expr:
    """Unroll `map_tuple(f)(t)` over top-level elements only (no recursion)."""
    (tup_expr,) = tup_exprs
    (tup_type,) = tup_types
    return im.make_tuple(
        *(im.call(f)(im.tuple_get(i, tup_expr)) for i in range(len(tup_type.types)))
    )
```
by
```
def _map_tuple_body(
    f: itir.Expr, tup_exprs: list[itir.Expr], tup_types: list[ts.TupleType]
) -> itir.Expr:
    """Unroll `map_tuple(f)(t1, ..., tN)` over top-level elements only (no recursion)."""
    return im.make_tuple(
        *(
            im.call(f)(*(im.tuple_get(i, tup_expr) for tup_expr in tup_exprs))
            for i in range(len(tup_types[0].types))
        )
    )
```

The only benefit would be in the case
`map_tuple(λa,b → a+b)(g(), h())` vs `map_tuple(λt → t[0]+t[1])(zip(g(),h()))` where g()/h() would be evaluated 2× in the former and 4x in the latter. But usually a comprehension has a single source, so there is nothing independent to combine.

So, I reverted the changes again and would keep map_tuple single-arg.

tehrengruber: Conclusion: Even if we eventually run `CollapseTuple` before the tuple expansion (or in case we wouldn't expand), the pass would still need to understand the semantics of `map_tuple` in order to reliably remove unused tuples, e.g. in `map_tuple(lambda a: a[0])({{b, c}, {d, e})`.

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,120 @@
...
+        for tup in tup_args:
+            itir_inference.reinfer(tup)
+            assert isinstance(tup.type, ts.TupleType)
+            tup_types.append(tup.type)
+
+        # Introduce a `let` binding for every argument to avoid duplicating (potentially
+        # expensive) sub-expressions across leaf projections. We deliberately do not simplify
+        # the result here: the regular `CollapseTuple` pass that runs later in the pipeline
+        # inlines the trivial bindings back and folds any `tuple_get(i, make_tuple(...))`
+        # patterns, so re-implementing that logic here would only duplicate work.
```
tehrengruber:
```suggestion
        # Let bind tuple args to avoid expression duplication
```

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,120 @@
...
+        let_bindings = [(next(self.uids["_utm"]), tup) for tup in tup_args]
+        substituted_exprs: list[itir.Expr] = [im.ref(ref_name) for ref_name, _ in let_bindings]
```
tehrengruber:
```suggestion
        let_bindings = {(next(self.uids["_utm"]): tup for tup in tup_args}
        substituted_exprs: list[itir.Expr] = [im.ref(ref_name) for ref_name in let_bindings.keys()]
```

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,120 @@
...
+
+        body = _UNROLLERS[node.fun.fun.id](f, substituted_exprs, tup_types)
+
+        result = im.let(*let_bindings)(body)
```
tehrengruber:
```suggestion
        result = im.let(**let_bindings.items())(body)
```

src/gt4py/next/ffront/foast_to_gtir.py
```diff
@@ -413,27 +413,21 @@ def _visit_where(self, node: foast.Call, **kwargs: Any) -> itir.FunCall:
...
         cond_symref_name = f"__cond_{itir.lenient_ir_fingerprinter(cond_)}"
 
-        def create_if(
-            true_: itir.Expr, false_: itir.Expr, arg_types: tuple[ts.TypeSpec, ts.TypeSpec]
-        ) -> itir.FunCall:
-            return _map(
-                "if_",
-                (im.ref(cond_symref_name), true_, false_),
-                (node.args[0].type, *arg_types),
+        result = im.tree_map_tuple(
```
tehrengruber: Since the current approach on main works, we postpone this change until then. Let's leave a todo why this doesn't use `tree_map_tuple`.

src/gt4py/next/ffront/foast_to_gtir.py
tehrengruber: Let's leave a comment why this can not use tree_map right now.

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,120 @@
...
+                functools.reduce(lambda expr, i: im.tuple_get(i, expr), path, tup_expr)
+                for tup_expr in tup_exprs
+            )
+        )
+
+    return mapper(*tup_types)
+
+
+def _map_tuple_body(
+    f: itir.Expr, tup_exprs: list[itir.Expr], tup_types: list[ts.TupleType]
```
tehrengruber:
```suggestion
    f: itir.Expr, tup_exprs: itir.Expr, tup_types: ts.TupleType
```

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,120 @@
...
+) -> itir.Expr:
+    """Unroll `map_tuple(f)(t)` over top-level elements only (no recursion)."""
+    (tup_expr,) = tup_exprs
+    (tup_type,) = tup_types
+    return im.make_tuple(
+        *(im.call(f)(im.tuple_get(i, tup_expr)) for i in range(len(tup_type.types)))
+    )
+
+
+_UNROLLERS = {
```
tehrengruber: Type annotation missing

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -629,6 +628,62 @@ def applied_map(
     return applied_map
 
 
+def _tuple_map_synthesizer(
+    builtin_name: str, *, recursive: bool
+) -> Callable[..., TypeOrTypeSynthesizer]:
+    """Shared implementation for `tree_map_tuple` (recursive) and `map_tuple` (top-level)."""
+
+    def ensure_same_tuple_structure(args: tuple[ts.TupleType, ...]) -> None:
```
tehrengruber: Can be simplified since tuple map takes a single arg only. Most likely the entire function is not needed anymore.

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -629,6 +628,62 @@ def applied_map(
...
+            if not all(isinstance(a, ts.TupleType) for a in args):
+                raise TypeError(
+                    f"'{builtin_name}' requires all top-level arguments to be TupleType, "
+                    f"got {[type(a).__name__ for a in args]}."
+                )
+            if recursive:
+                ensure_same_tuple_structure(args)
+
+            def leaf_op(*leaf_types: ts.TypeSpec) -> ts.TypeSpec:
+                return op(*leaf_types, offset_provider_type=offset_provider_type)  # type: ignore[return-value]
```
tehrengruber:
```suggestion
            op = functools.partial(op, offset_provider_type=offset_provider_type)  # type: ignore[return-value]
```
Or `bound_op = `

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -629,6 +628,62 @@ def applied_map(
     return applied_map
 
 
+def _tuple_map_synthesizer(
```
tehrengruber:
```suggestion
def _make_tuple_map_synthesizer(
```

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -629,6 +628,62 @@ def applied_map(
...
+    """Shared implementation for `tree_map_tuple` (recursive) and `map_tuple` (top-level)."""
+
+    def ensure_same_tuple_structure(args: tuple[ts.TupleType, ...]) -> None:
+        expected_structure = type_info.tuple_structure(args[0])
+        if any(type_info.tuple_structure(arg) != expected_structure for arg in args[1:]):
+            raise TypeError(
+                f"'{builtin_name}' requires all arguments to have the same tuple structure."
+            )
+
+    def factory(op: TypeSynthesizer) -> TypeSynthesizer:
```
tehrengruber:
```suggestion
    def tuple_map_synthesizer(op: TypeSynthesizer) -> TypeSynthesizer:
```

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -629,6 +628,62 @@ def applied_map(
...
+
+            if recursive:
+                return utils.tree_map(  # type: ignore[return-value]
+                    leaf_op,
+                    collection_type=ts.TupleType,
+                    result_collection_constructor=lambda _, elts: ts.TupleType(types=[*elts]),
+                )(*args)
+
+            # Non-recursive: apply `op` once per top-level element.
+            (arg,) = args
```
tehrengruber: Will become simpler with a single arg.

src/gt4py/next/ffront/foast_to_gtir.py
```diff
@@ -409,6 +409,9 @@ def create_cast(expr: itir.Expr, t: tuple[ts.TypeSpec]) -> itir.FunCall:
         )
 
     def _visit_where(self, node: foast.Call, **kwargs: Any) -> itir.FunCall:
+        # TODO: For tuples we unroll over the tuple structure via `process_elements` instead of
+        #  emitting `tree_map_tuple`. `where` would require a multi-argument `tree_map_tuple`, but
+        #  it currently only supports a single argument.
```
tehrengruber:
```suggestion
        # TODO(tehrengruber): For tuples we unroll over the tuple structure via `process_elements`
        #  instead of emitting `tree_map_tuple`.
```
The reason why we are using process elements is missing: Mixed type `(local field, regular field)` arguments.

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def _map_tuple_body(f: itir.Expr, tup_expr: itir.Expr, tup_type: ts.TupleType) -> itir.Expr:
+    """Unroll `map_tuple(f)(t)` over top-level elements only (no recursion)."""
+    return im.make_tuple(
+        *(im.call(f)(im.tuple_get(i, tup_expr)) for i in range(len(tup_type.types)))
+    )
+
+
+_UNROLLERS: dict[str, Callable[[itir.Expr, itir.Expr, ts.TupleType], itir.Expr]] = {
```
tehrengruber:
```suggestion
_MAP_EXPANDER: dict[str, Callable[[itir.Expr, itir.Expr, ts.TupleType], itir.Expr]] = {
```

src/gt4py/next/iterator/transforms/expand_tuple_maps.py
```diff
@@ -0,0 +1,94 @@
+# GT4Py - GridTools Framework
```
tehrengruber, line 1: Let's reword unroll to expand to be consistent with the other passes.

src/gt4py/next/iterator/transforms/unroll_tuple_maps.py
```diff
@@ -0,0 +1,94 @@
...
+            return node
+
+        assert isinstance(node.fun, itir.FunCall)
+        f = node.fun.args[0]
+        (tup,) = node.args
+        itir_inference.reinfer(tup)
+        assert isinstance(tup.type, ts.TupleType)
+
+        # Let bind the tuple arg to avoid expression duplication.
+        ref_name = next(self.uids["_utm"])
```
tehrengruber:
```suggestion
        ref_name = next(self.uids["_etm"])
```

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -629,6 +628,49 @@ def applied_map(
...
+    """Shared implementation for `tree_map_tuple` (recursive) and `map_tuple` (top-level)."""
+
+    def tuple_map_synthesizer(op: TypeSynthesizer) -> TypeSynthesizer:
+        @type_synthesizer
+        def applied_map(
+            *args: ts.TupleType, offset_provider_type: common.OffsetProviderType
+        ) -> ts.TupleType:
+            if len(args) != 1:
+                raise TypeError(f"'{builtin_name}' requires exactly one argument, got {len(args)}.")
+            (arg,) = args
```
tehrengruber:
```suggestion
            arg: ts.TupleType, offset_provider_type: common.OffsetProviderType
        ) -> ts.TupleType:
```

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_unroll_tuple_maps.py
```diff
@@ -0,0 +1,203 @@
...
+        im.make_tuple(
+            im.call(_unary_fun)(im.tuple_get(0, "_utm_0")),
+            im.call(_unary_fun)(im.tuple_get(1, "_utm_0")),
+        )
+    )
+
+
+def test_map_tuple_does_not_recurse():
+    nested = ts.TupleType(types=[i_tuple_field, i_tuple_field])
+    g = im.lambda_("__p")(im.op_as_fieldop("plus")(im.tuple_get(0, "__p"), im.tuple_get(1, "__p")))
```
tehrengruber:
```suggestion
    g = im.ref("g", ts.FunctionType(...))
```

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_unroll_tuple_maps.py
```diff
@@ -0,0 +1,203 @@
...
+                im.make_tuple(im.ref("c", i_field), im.ref("d", i_field)),
+            )
+        )
+    )
+
+    expected = im.make_tuple(
+        im.make_tuple(im.call(_unary_fun)("a"), im.call(_unary_fun)("b")),
+        im.make_tuple(im.call(_unary_fun)("c"), im.call(_unary_fun)("d")),
+    )
+    assert result == expected
```
tehrengruber:
```suggestion
```

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_unroll_tuple_maps.py
```diff
@@ -0,0 +1,203 @@
...
+    )
+
+    expected = im.make_tuple(
+        im.call(_unary_fun)(im.tuple_get(0, "t")),
+        im.call(_unary_fun)(im.tuple_get(1, "t")),
+    )
+    assert result == expected
+
+
+def test_apply_infers_uninferred_expr(_unary_fun):
```
tehrengruber:
```suggestion
def test_tree_map_tuple_two_args(_unary_fun):
```

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_unroll_tuple_maps.py
```diff
@@ -0,0 +1,203 @@
...
+    expr = im.call(im.call("tree_map_tuple")(_unary_fun))(
+        im.make_tuple(im.ref("a", i_field), im.ref("b", i_field))
+    )
+
+    result = UnrollTupleMaps.apply(expr, uids=None, offset_provider_type={})
+
+    assert expr.type is None
+    # `UnrollTupleMaps` let-binds the argument and leaves the `tuple_get(i, make_tuple(...))`
+    # cleanup to the regular `CollapseTuple` pass (see `test_make_tuple_arg_is_collapsed`).
```
tehrengruber:
```suggestion
```

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_expand_tuple_maps.py
```diff
@@ -0,0 +1,203 @@
...
+def test_map_tuple_with_make_tuple_arg_is_collapsed(_unary_fun):
+    """The `make_tuple` collapse also applies for the `map_tuple` builtin."""
+    result = _apply_and_collapse(
+        im.call(im.call("map_tuple")(_unary_fun))(
+            im.make_tuple(im.ref("a", i_field), im.ref("b", i_field))
+        )
+    )
+
+    expected = im.make_tuple(im.call(_unary_fun)("a"), im.call(_unary_fun)("b"))
+    assert result == expected
```
tehrengruber, line 118:
```suggestion
```

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_unroll_tuple_maps.py
```diff
@@ -0,0 +1,203 @@
...
+            im.call(_unary_fun)(im.tuple_get(0, "_utm_0")),
+            im.call(_unary_fun)(im.tuple_get(1, "_utm_0")),
+        )
+    )
+
+    # After `CollapseTuple` the trivial `let` is inlined away.
+    assert _apply_and_collapse(testee) == im.make_tuple(
+        im.call(_unary_fun)(im.tuple_get(0, "t")),
+        im.call(_unary_fun)(im.tuple_get(1, "t")),
+    )
```
tehrengruber:
```suggestion
```

tehrengruber (approved):

PR #2587, opened 2026-04-27
refactor[next]: Split compilation into a actual compilation and loading of the compilation artifact

havogt:
This is a preparation to allow compilation in separate processes for faster jit compilation.

The new workflow: CompilationStep produces a CompilationArtifact which is picklable and has a `load()` method which returns the original ExecutableProgram.


src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -114,19 +119,66 @@ def __call__(self, **kwargs: Any) -> None:
...
+            object.__setattr__(self, "_live_program", program)
+        return gtx_wfddecoration.convert_args(program, device=self.device_type)
+
+    def _load_compiled_program(self) -> CompiledDaceProgram:
+        sdfg = dace.SDFG.from_file(str(self.sdfg_dump))
+        sdfg.build_folder = str(self.build_folder)
+
+        with gtx_wfdcommon.dace_context(device_type=self.device_type):
+            with dace.config.set_temporary("compiler", "use_cache", value=True):
+                sdfg_program = sdfg.compile(validate=False)
```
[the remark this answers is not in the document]
havogt: No, should be fine because we only reach `load()` because we went through a build step first, therefore `load()` is read only.

AGENTS.md
```diff
@@ -65,6 +65,13 @@ If a command above is wrong for your environment, fix `pyproject.toml`,
 ## Do
 
 - Run `uv run pre-commit run` on staged files before claiming a task done.
+- Write docstrings in **Google style** (rendered via Sphinx-napoleon). Use
```
havogt, line 68: I specifically asked for that change. I thought it's not worth opening a separate PR.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -128,19 +135,65 @@ def __call__(self, **kwargs: Any) -> None:
...
+    def _load_compiled_program(self) -> CompiledDaceProgram:
+        # TODO(phimuell): Drop ``sdfg_json`` from the artifact once dace
+        #   exposes a load path that doesn't require an SDFG instance to wrap
+        #   into the returned ``CompiledSDFG``.
+        sdfg = dace.SDFG.from_json(json.loads(self.sdfg_json))
+        folder_version = dace_compiler.get_folder_version(self.build_folder)
+        library_path = dace_compiler.get_binary_name(
+            self.build_folder, sdfg_name=sdfg.name, folder_version=folder_version
+        )
+        sdfg_program = dace_compiler.get_program_handle(library_path, sdfg)
```
egparedes: Question: is it correct to recompute these symbols here using global functions from `dace_compiler`? I mean, can these recomputed values be affected by global configuration flags? Wouldn't it be better to store these symbols inside the `DaCeCompilationArtifact` itself?

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -155,12 +208,12 @@ class DaCeCompiler(
...
         inp: stages.CompilableProject[code_specs.SDFGCodeSpec, code_specs.PythonCodeSpec],
-    ) -> CompiledDaceProgram:
+    ) -> DaCeCompilationArtifact:
         with gtx_wfdcommon.dace_context(
             device_type=self.device_type,
             cmake_build_type=self.cmake_build_type,
         ):
-            sdfg_build_folder = gtx_cache.get_cache_folder(inp, self.cache_lifetime)
+            sdfg_build_folder = pathlib.Path(gtx_cache.get_cache_folder(inp, self.cache_lifetime))
             sdfg_build_folder.mkdir(parents=True, exist_ok=True)
```
egparedes: To avoid the conversion of `sdfg_build_folder` to Path first and back to str
```suggestion
            sdfg_build_folder = gtx_cache.get_cache_folder(inp, self.cache_lifetime)
            pathlib.Path(sdfg_build_folder).mkdir(parents=True, exist_ok=True)
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -169,16 +222,22 @@ def __call__(
             if self.add_gpu_trace_markers and self.device_type == core_defs.CUPY_DEVICE_TYPE:
                 _add_tx_markers(sdfg)
 
-            sdfg.build_folder = sdfg_build_folder
+            sdfg.build_folder = str(sdfg_build_folder)
```
egparedes: If changed above:
```suggestion
            sdfg.build_folder = sdfg_build_folde)
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -128,19 +135,65 @@ def __call__(self, **kwargs: Any) -> None:
         assert result is None
 
 
+_live_program_cache: dict[pathlib.Path, CompiledDaceProgram] = {}
```
egparedes: I don't fully understand the goal of this `cache` , is it really needed? Could you explain and document why?

src/gt4py/next/program_processors/runners/roundtrip.py
```diff
@@ -110,28 +109,20 @@ def visit_Temporary(self, node: itir.Temporary, **kwargs: Any) -> str:
...
-        extract_temporaries: Extract intermediate field values into temporaries.
-        use_embedded: Directly use builtins from embedded backend instead of
-                      generic dispatcher. Gives faster performance and is easier
-                      to debug.
-        offset_provider: A mapping from offset names to offset providers.
-    """
+) -> tuple[str, str]:
+    """Generate the Python source for an ITIR program. Returns ``(source_code, entry_point_name)``."""
     # TODO(tehrengruber): just a temporary solution until we have a proper generic
     #  caching mechanism
```
egparedes, line 127: Could we resolve now this TODO?

egparedes (commented):
First quick round of reviews. In general looks good to me. In the second run I would need to pay more attention to the changes in the roundtrip backend and tests.

tests/next_tests/unit_tests/otf_tests/compilation_tests/test_compiler.py
```diff
@@ -0,0 +1,26 @@
+# GT4Py - GridTools Framework
```
havogt, line 1: double-check that such programs with closures are properly shipped to the compile process:
```
def test_....:
    bla = 1
    @field_operator
    def my_op(...):
        return bla
```

havogt: Checked: the boundary currently is CompilableProgramDef which is gtir+args at which point closures are resolved.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
havogt: Documented in 1a13a63c3 — it's a producer-consumer hand-off, not really a multi-hit cache: `DaCeCompiler.__call__` populates it, the very next `artifact.load()` reads it back, and that's it for the typical flow (only one `load()` call site exists, in `Backend.compile()`).

The reason it's load-bearing rather than a pure optimization: `sdfg.compile()` dlopens the .so internally; the immediately-following `dace_compiler.get_program_handle(library_path, ...)` then hits dace's *"Library `...librun.so` already loaded, renaming file"* path, which renames the .so on disk and dlopens the renamed copy. Without the hand-off, every thread-mode compile would silently invalidate the `library_path` we just stored on the artifact, and litter persistent build caches with `librun_1.so`, `librun_2.so`, …

Quick numbers on a trivial 9.7 KiB SDFG: cache hit 5 µs; cold `_load_compiled_program` 10 ms (5 ms SDFG.from_json + 2 ms get_program_handle + 3 ms binding exec). At icon4py scale (~300 dycore programs) that's ~3 s of avoided cold work per process startup, on top of the correctness concern.

In process-pool mode the cache is just an optimization — the worker's load is genuinely cold and dace has no prior dlopen registered.

src/gt4py/next/program_processors/runners/roundtrip.py
havogt: The "proper generic caching mechanism" exists now: `workflow.CachedStep` (in `src/gt4py/next/otf/workflow.py`). Every other step in the codebase uses it — `func_to_foast`, `past_to_itir`, gtfn's `translation`, dace's `translation`, gtfn's executor `cached` trait. Roundtrip is the last hold-out still rolling its own `_SOURCE_CACHE` / `_MODULE_CACHE` dicts, keyed by `hash((ir, transforms, debug, use_embedded, …))` (process-local, hash-randomization-fragile for nested IR).

Two coverage gaps to be aware of before resolving:

1. **`_SOURCE_CACHE` (codegen-side)** maps cleanly onto `CachedStep`. Extract `_generate_source` into a `Workflow[CompilableProgramDef, …]` step, wrap in `CachedStep.in_memory(input_fingerprinter=…)` with a roundtrip-specific fingerprinter (must include `debug` + `use_embedded` since those alter generated source). Delete `_SOURCE_CACHE`. Optionally expose a `cached_translation` factory Trait so caching is opt-in like gtfn/dace's (today roundtrip's is always-on, which differs).
2. **`_MODULE_CACHE` (load-side)** has *no* infra slot. `CachedStep` operates on workflow steps that produce artifacts; `load()` is post-artifact. It's a process-local amortization of `exec()` cost, conceptually the same shape as the dace `_live_program_cache` we just documented in 1a13a63c3. Best treatment: keep it, add a WHY comment in the same style.

Pragmatic minimum is the gap-1 fix; gap 2 stays with a WHY comment.

The cleaner architectural realignment (matches your layered-architecture proposal) is to make `Roundtrip` an `OTFCompileWorkflow` instance: `translation` = ITIR transforms + EmbeddedDSL codegen → `ProgramSource`; `bindings` = identity; `compilation` = identity (Python source IS its own artifact). Then `CachedStep.persistent` slots onto `translation` uniformly with gtfn/dace and persistent on-disk caching of generated Python comes for free via `cache.get_cache_base_path(...)`. But that's only worth doing inside the layered-architecture work, not as a one-off PR.

Filed for follow-up; out of scope for this PR.

egparedes (approved):
LGTM

src/gt4py/next/otf/stages.py
```diff
@@ -129,6 +129,24 @@ def build(self) -> None: ...
 ExecutableProgram: TypeAlias = Callable
 
 
+class CompilationArtifact(Protocol):
```
tehrengruber, line 141: Can we make this runtime_checkable?

src/gt4py/next/otf/compilation/compiler.py
```diff
@@ -83,17 +111,22 @@ def __call__(
...
+    ) -> CPPCompilationArtifact:
+        return CPPCompilationArtifact(
+            src_dir=src_dir,
+            module=module,
+            entry_point_name=entry_point_name,
+            device_type=self.device_type,
+        )
 
 
 class CompilerFactory(factory.Factory):
```
tehrengruber: Is this class obsolete now?

src/gt4py/next/otf/compilation/compiler.py
```diff
@@ -83,17 +111,22 @@ def __call__(
                     f"On-the-fly compilation unsuccessful for '{inp.program_source.entry_point.name}'."
                 )
```
tehrengruber, line 116: Except for the `builder_factory` this part also feels like loading, the only thing necessary to construct a `CPPCompilationArtifact` appears to be the `src_dir`.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -174,11 +233,18 @@ def __call__(
...
-            inp.binding_source,
+        artifact = DaCeCompilationArtifact(
+            build_folder=pathlib.Path(sdfg_build_folder),
+            library_path=pathlib.Path(sdfg_program.filename),
+            sdfg_json=json.dumps(inp.program_source.source_code),
+            binding_source_code=inp.binding_source.source_code,
+            bind_func_name=self.bind_func_name,
+            device_type=self.device_type,
+        )
+        _live_program_cache[artifact.build_folder] = CompiledDaceProgram(
```
tehrengruber: Feels a little odd that `DaCeCompilationArtifact` and `DaCeCompiler` manage the `_live_program_cache`, but managing everything from `DaCeCompilationArtifact` also has its downsides. Ideally we could compile the sdfg without loading it then this part here wouldn't be needed.

tehrengruber (approved):
Only did a birds eye review. If compilation and loading are cleanly split, it should eventually be possible to only load when compilation already occured in a previous run, which is almost the case in gtfn, but not in dace yet.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
havogt: Yes, in a previous iteration that stuff was on the compilation artifact...

src/gt4py/next/otf/stages.py
havogt: done

src/gt4py/next/otf/compilation/compiler.py
havogt: yes, thanks!

src/gt4py/next/otf/compilation/compiler.py
havogt: I am experimenting to isolate the load to just be an import of a python module. I think will propose as a separate PR after this one.

egparedes (approved):
LGTM

PR #2588, opened 2026-04-28
fix[cartesian]: Consider full stencil when hashing stencil code for caching purposes

romanc:
## Description

Issue https://github.com/NOAA-GFDL/NDSL/issues/445 suggests that there might be a problem with caching in `gt4py.cartesian`. Indeed, caching would not hash the full stencil code, but skip certain nodes, in particular annotations. Those annotation became important with [2d temporaries](https://github.com/GridTools/gt4py/blob/main/docs/development/ADRs/cartesian/experimental/2d-temporaries.md).

The PR suggests to just dump thus hash the full stencil code to work around the issue. The added test is a bit cumbersome. If anyone known something simpler - I'm happy to go with that.

## Requirements

  N/A

### DIFF
--- a/src/gt4py/cartesian/utils/meta.py
+++ b/src/gt4py/cartesian/utils/meta.py
@@ -13,7 +13,7 @@ import copy
 import inspect
 import operator
 import textwrap
-from typing import Callable, Dict, Final, List, Tuple, Type
+from typing import Callable, Final
 
 
 def get_closure(func, *, include_globals=True, included_nonlocals=True, include_builtins=True):
@@ -52,7 +52,7 @@ def get_source(func):
     return source
 
 
-def get_ast(func_or_source_or_ast, *, feature_version: Tuple[int, int]):
+def get_ast(func_or_source_or_ast, *, feature_version: tuple[int, int]):
     if callable(func_or_source_or_ast):
         func_or_source_or_ast = get_source(func_or_source_or_ast)
     if isinstance(func_or_source_or_ast, str):
@@ -216,7 +216,7 @@ class ASTTransformPass(ASTPass):
 
 
 class ASTEvaluator(ASTPass):
-    AST_OP_TO_OP: Final[Dict[Type, Callable]] = {
+    AST_OP_TO_OP: Final[dict[type, Callable]] = {
         # Arithmetic operations
         ast.UAdd: operator.pos,
         ast.USub: operator.neg,
@@ -387,7 +387,7 @@ class QualifiedNameCollector(ASTPass):
             self.name_nodes[node.id].append(node)
 
     def _get_name_components(self, node: ast.AST):
-        components: List
+        components: list
         if isinstance(node, ast.Name):
             components = [node.id]
             valid = self.prefixes is None or node.id in self.prefixes
--- a/src/gt4py/cartesian/utils/meta.py
+++ b/src/gt4py/cartesian/utils/meta.py
@@ -66,25 +66,10 @@ def get_ast(func_or_source_or_ast, *, feature_version: Tuple[int, int]):
     return ast_root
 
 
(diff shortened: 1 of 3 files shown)


src/gt4py/cartesian/utils/meta.py
```diff
@@ -98,28 +80,17 @@ def _dump(node: ast.AST, excluded_names):
...
+        return repr(node)
 
-    dumped_ast = _dump(get_ast(definition, feature_version=feature_version), skip_node_names)
+    dumped_ast = _dump(get_ast(definition, feature_version=feature_version))
 
     return dumped_ast
 
 
-def ast_shash(ast_node, *, skip_decorators=True):
-    return shashed_id(ast_dump(ast_node, skip_decorators=skip_decorators))
```
romanc, line 120: Unused code - no need to keep around

src/gt4py/cartesian/utils/meta.py
```diff
@@ -98,28 +80,17 @@ def _dump(node: ast.AST, excluded_names):
...
-
-    skip_node_names = set()
-    if skip_decorators:
-        skip_node_names.add("decorator_list")
-    if skip_annotations:
-        skip_node_names.add("annotation")
+        return repr(node)
 
-    dumped_ast = _dump(get_ast(definition, feature_version=feature_version), skip_node_names)
+    dumped_ast = _dump(get_ast(definition, feature_version=feature_version))
```
romanc, line 89: this is where we fix the issue of skipping certain nodes

romanc:
cscs-ci run

src/gt4py/cartesian/utils/meta.py
```diff
@@ -98,28 +80,17 @@ def _dump(node: ast.AST, excluded_names):
...
+        if isinstance(node, list):
+            lines = ["[", *[_dump(i) + "," for i in node], "]"]
             return "\n".join(lines)
 
-        else:
-            return repr(node)
-
-    skip_node_names = set()
-    if skip_decorators:
-        skip_node_names.add("decorator_list")
```
twicki, line 110: was there a point to this? Was it that importing `stencil` and using `gtscript.stencil` should not give you a new one? I could not come up with a real example. I think this should be ok to go

tests/cartesian_tests/unit_tests/test_caching.py
```diff
@@ -211,3 +215,54 @@ def test_debug_mode(builder):
...
+
+    # setup "modified stencil"
+    modified_stencil = builder(stencil_tmp_modified)
+
+    # access .definition to trigger stencil annotation
+    modified_stencil.definition
+    assert hasattr(stencil_tmp_modified, "_gtscript_")
+
+    # patch stencil s.t. the only change are inside the stencil
+    stencil_tmp_modified._gtscript_["canonical_ast"] = stencil_tmp_modified._gtscript_[
```
twicki, line 264: the goal of this is to keep the name / line-number the same an check what happens if we only change the AST, correct?

twicki (approved):

romanc:
cscs-ci run

src/gt4py/cartesian/utils/meta.py
romanc: I agree this should be okay to go. I don't think this would filter anything (anymore) at the moment.

As for why this code is here, I can only speculate. Maybe in the past the hashing was different and the whole stencil function (the one with the @stencil decorator) was pushed here? For now, we are only dumping the contents of the stencil function anyway, so it doesn't make a difference.

tests/cartesian_tests/unit_tests/test_caching.py
romanc: Yes

PR #2590, opened 2026-05-03
Infrastructure to test DaCe's codegen (in)deterministic behavior

kotsaloscv:
[dace_determinism_how_it_works.md](https://github.com/user-attachments/files/28554598/dace_determinism_how_it_works.md)
## Infrastructure to test DaCe's codegen (in)deterministic behavior

Adds infrastructure to detect non-determinism in `gt4py.next`'s DaCe codegen. A
parametrized nox session runs gt4py's own DaCe test selection **twice** with isolated
build caches, then compares the generated sources between the two runs. The session
succeeds when the codegen is identical and fails on any difference. Covers the
`dace_cpu`, `dace_gpu` (CUDA), and HIP backends. Driven entirely from gt4py's own
nox sessions — no external repo needed.

### How the comparison works

gt4py.next caches each compiled program at `<root>/<name>_<sha256>/src/...`, where the
`<sha256>` is derived from the generated source. Non-deterministic codegen therefore
produces a *different folder name* each run, so programs are compared by **logical name**
(the folder name without the trailing digest). A single name may be compiled several
times in one run with different baked parameters, so each run yields a **multiset of
generated-source signatures** per name (a signature is the set of `(relpath, sha256)`
pairs of one compiled program).

How many programs a logical name produces is determined by the tests — which ones run,
and with which compile-time parameters — not by codegen. Codegen fixes the *contents* of
each generated program, never the *number* of them. The comparison builds on this:

- A name is compared **only when both runs produced the same number of programs** for it
  (`count1 == count2`). Equal counts mean the same programs were generated in both runs,
  so any difference in their source multisets is codegen non-determinism. The check is
  order-independent (multiset equality), which makes it robust to parallel execution
  under `pytest-xdist`.
- A name whose counts **differ** between the runs is handled according to whether the
  runs were clean. The session reads each run's JUnit result and passes a `runs_healthy`
  flag to the comparator. If a run had test failures, a differing count may simply mean a
  test did not run, so the name is **skipped and reported** — a tolerated coverage caveat,
  not a failure. If **both runs completed cleanly**, a differing count cannot be a missing
  test; it means the *number* of generated programs is itself non-deterministic, and the
  name is reported as a **failure** (`COUNT`). Content differences (`DIFFER`) are failures
  regardless of run health.
- If programs were cached but **no name was comparable** (every name had a differing
  count, e.g. one run failed wholesale or produced nothing) and the mismatches cannot be
  attributed to codegen, the check raises `NoComparableProgramsError` instead of returning
  a vacuous green pass.

Program folders whose `src/` contains no source files are excluded from the multisets
(they carry no codegen), so they cannot manufacture spurious mismatches.

On a mismatch, the per-program diff file lists only the generated files that **actually
differ** between the two runs.

### Layout

- **`scripts/dace_deterministic_codegen.py`** — pure-stdlib comparison library + argparse
  CLI. Walks each `.gt4py_cache/`, hashes every file under each program's `src/`
  (`cpu`, `cuda`, and HIP under `cuda/hip/` are covered by the recursive walk), and
  compares the per-name multisets at equal count. Raises `DeterminismError` on a content
  difference (or, in a clean pair, a non-deterministic program count);
  `NoProgramsObservedError`, `NoSourceFilesObservedError`,
  `NoComparableProgramsError`, and `UnsupportedBackendError` on the corresponding
  conditions.
- **`tests/package_tests/test_dace_deterministic_codegen.py`** — unit tests for the
  comparator: collection, multiset comparison, equal-count match/differ, count-mismatch
  handling (tolerated vs. failed by run health), the "nothing comparable" / one-empty-run
  guards, reporting, and error paths.
- **`noxfile.py`** — `test_next_dace_determinism` session, parametrized over `device`
  (`cpu` / `cuda` / …) and `meshlib` (`nomesh` / `atlas`), tagged
  `next`, `dace`, `determinism`. Runs the DaCe pytest selection twice into isolated
  caches, then invokes the comparator.
- **`ci/cscs-ci-dace-determinism.yml`** — standalone CSCS-CI pipeline entry point.

### Output

- **Report** (`report.txt`): a one-line summary header (`Programs / comparable / matches
  / differs / skipped`), one line per program (`[MATCH] / [DIFFER] / [SKIP] <name>`), a
  note listing any skipped programs with a re-run hint, and a final `RESULT:` line.
- **Diffs** (`diffs/<name>.txt`): for each non-matching program, the name followed by the
  relative paths of the generated files that differ between the two runs.

### Verdict / exit codes

`0` deterministic · `1` codegen differs · `2` bad args / unsupported backend / no source
files / nothing comparable · `3` no programs observed.

### Environment

The session sets, per run:

- `GT4PY_BUILD_CACHE_DIR=<run_dir>` — cache lands at `<run_dir>/.gt4py_cache/`.
- `GT4PY_BUILD_CACHE_LIFETIME=persistent` — keep the cache across the run.
- `DACE_compiler_build_folder_mode=development` — **required**: gt4py defaults DaCe to
  `production`, which strips the generated sources after compilation, leaving nothing to
  compare. The comparator raises `NoSourceFilesObservedError` with a hint if it ever
  encounters caches built without it.

In addition, the session **wipes `.dacecache/` before each of the two runs**. DaCe keeps
its build cache at the repo root, outside the per-run `GT4PY_BUILD_CACHE_DIR`; if it
survived from the first run, the second run could rebuild from it and reproduce the first
run's bytes, masking real non-determinism.

### CI

Standalone CSCS-CI pipeline, independent from `ci/cscs-ci.yml`, registered as
`dace-determinism` and triggered by whitelisted users posting `cscs-ci run dace-determinism`
in a PR comment. `allow_failure: true` while DaCe codegen non-determinism is being fixed —
to be dropped once `main` stays green. Per-cell artifacts (`_dace_deterministic_codegen/`
with `diffs/` and `report.txt`) are retained for 1 month.

### Known limitations

The comparison is exact for the failure mode it targets — a program rendered with
different bytes at equal count — but two situations are out of its reach by construction:

- **Intermittent non-determinism can slip past a two-run comparison.** If a program's
  output varies only sometimes and both runs happen to land identically, it is reported
  as a match. Sampling more than two runs would raise the detection probability.
- **Equal-count, different program *set* would be a false positive.** If a test made the
  *identity* of a compiled program depend on a non-deterministic value (e.g. a random or
  hypothesis-drawn size/dtype baked into the source), the two runs would compile
  different programs at equal count and be flagged as non-deterministic even though
  codegen is deterministic. The selected DaCe tests do not do this (no hypothesis; random
  data only fills runtime values of fixed-shape fields), so it does not occur in practice.

It also compares **generated source under `src/` only** — not the compiled binary, and
not files outside `src/`.

### Local use

```bash
./noxfile.py -s 'test_next_dace_determinism-3.10(cpu, nomesh)'    # one variant
nox -t determinism                                                # all variants
python scripts/dace_deterministic_codegen.py --run1 PATH --run2 PATH \
    [--diffs-dir DIR] [--report FILE]
```


kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
+#!/usr/bin/env python3
```
tehrengruber, line 1: This file is missing a test, otherwise only a manual pipeline run verifies this is runnable at all.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+Supports both gt4py cache layouts:
+
+* ``layout="next"`` (default) — the ``gt4py.next`` cache, a flat
+  ``<cache_root>/<name>_<sha256>/src/{cpu,cuda}/...`` structure
+  written via ``GT4PY_BUILD_CACHE_DIR``. Compares everything dace
+  writes as generated source under each program's ``src/``.
+  Unknown top-level backends (anything other than cpu/cuda, with
+  HIP nesting under cuda/hip) raise :class:`UnsupportedBackendError`.
+
+* ``layout="cartesian"`` — the ``gt4py.cartesian`` cache, a deeply
```
tehrengruber: The script needs special casing for cartesian in many places, but there is no debugging workflow for determinism bugs in cartesian (like the one that is currently developed for next). Additionally cartesian has 1.) another fontend 2.) a different optimization pipeline that only partially shares elements with the next pipeline. Determinism bugs in both components are explicitly no-fix for us people working with next only. Since there have been serious difficulties in locating bugs without a standardized workflow, because the resulting bugs are easily located in places that are marked as no-fix, and because it adds additional complexity, I think we should remove cartesian functionality from this script.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+
+
+def _snapshot_run_next(cache_root: Path) -> dict[str, ProgramSnapshot]:
+    """Snapshot a gt4py.next-layout cache.
+
+    The input directory's name is irrelevant — the function looks for
+    immediate subdirectories matching ``<name>_<64-char-hex-digest>``
+    (gt4py.next's program-folder naming) and reads ``<program>/src/``
+    recursively under each one. HIP files at ``src/cuda/hip/`` are
+    picked up automatically by the recursive walk.
```
tehrengruber: Not sure what this actually means, e.g. The input directory's name is irrelevant, however, writing something like <name>_<64-char-hex-digest> and PROGRAM_FOLDER_RE implicitly couples the caching behavior of next with this script. Refactoring the former will likely break this script with no ide tools to first find the coupling and then the problem that the person doing the refactoring has to understand the mechanics of this script in order to finish a refactoring in a completely different part of the code-base. Do this in a couple of places and you end up with something that breaks often and every refactoring becomes a rabbit-hole. I don't have a suggestion on how to fix this, but this is a sign to me that refinements to the overall design would have been needed.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+        return {}
+
+    out: dict[str, ProgramSnapshot] = {}
+    for folder in sorted(p for p in cache_root.iterdir() if p.is_dir()):
+        m = PROGRAM_FOLDER_RE.match(folder.name)
+        if not m:
+            continue
+        name = m.group("name")
+
+        src_root = folder / CODEGEN_ROOT
```
tehrengruber: `CODEGEN_ROOT` is another constant that introduces an maintenance and refactoring obstacle.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+        for fpath in sorted(src_root.rglob("*")):
+            if not fpath.is_file():
+                continue
+            rel = fpath.relative_to(folder).as_posix()
+            files[rel] = FileEntry(relpath=rel, sha256=_sha256(fpath), disk_relpath=rel)
+        out[name] = ProgramSnapshot(name=name, folder=folder, files=files)
+    return out
+
+
+def _snapshot_run_cartesian(
```
tehrengruber: I did not review this function.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+                    name=name,
+                    match=False,
+                    differing_files=[],
+                    only_in_run1=sorted((s1.files if s1 else {}).keys()),
+                    only_in_run2=sorted((s2.files if s2 else {}).keys()),
+                )
+            )
+            continue
+
+        keys1, keys2 = set(s1.files), set(s2.files)
```
tehrengruber:
```suggestion
        keys1, keys2 = set(s1.files if s1 else []), set(s2.files if s2 else [])
```

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+        only1 = sorted(keys1 - keys2)
+        only2 = sorted(keys2 - keys1)
+        differing = sorted(
+            rel for rel in keys1 & keys2 if s1.files[rel].sha256 != s2.files[rel].sha256
+        )
+        results.append(
+            ProgramResult(
+                name=name,
+                match=not (differing or only1 or only2),
```
tehrengruber:
```suggestion
```
This can be a property of `ProgramResult` avoiding inconsistencies.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+    diffs_dir: Path,
+) -> None:
+    """Emit a unified diff per differing file under ``diffs_dir/<program>/``."""
+    for r in results:
+        if r.match:
+            continue
+        s1, s2 = snap1.get(r.name), snap2.get(r.name)
+        prog_dir = diffs_dir / r.name
+        for rel in r.differing_files:
+            # `rel` is the canonical (normalized) key; the on-disk filename
```
tehrengruber: This looks strange. If the files differ s1 and s2 must both not be none.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+            # may differ from it (cartesian normalizes the 10-hex digest).
+            # Look up the per-side FileEntry to recover the real path.
+            e1 = s1.files.get(rel) if s1 else None
+            e2 = s2.files.get(rel) if s2 else None
+            f1 = (s1.folder / e1.disk_relpath) if (s1 and e1) else None
```
tehrengruber: It looks more convenient to just put absolute paths into a ProgramSnapshot.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+# ---------------------------------------------------------------------------
+
+
+def write_diffs(
+    results: list[ProgramResult],
+    snap1: dict[str, ProgramSnapshot],
+    snap2: dict[str, ProgramSnapshot],
+    diffs_dir: Path,
+) -> None:
+    """Emit a unified diff per differing file under ``diffs_dir/<program>/``."""
```
tehrengruber:
```suggestion
    """Emit a unified diff per differing file under ``<diffs_dir>/<program>/``."""
```

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+
+# ---------------------------------------------------------------------------
+# Diff + report
+# ---------------------------------------------------------------------------
+
+
+def write_diffs(
+    results: list[ProgramResult],
+    snap1: dict[str, ProgramSnapshot],
+    snap2: dict[str, ProgramSnapshot],
```
tehrengruber:
```suggestion
```
This two should be easy to get rid of.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,960 @@
...
+                    fromfile=f"run1/{rel}",
+                    tofile=f"run2/{rel}",
+                    n=3,
+                )
+            )
+            prog_dir.mkdir(parents=True, exist_ok=True)
+            (prog_dir / f"{rel.replace('/', '__')}.diff").write_text(udiff)
+
+
+def render_report(
```
tehrengruber: I'll continue my review here when I find time again.

tehrengruber (commented):
Partial review only.

scripts/dace_deterministic_codegen.py
kotsaloscv: @havogt what do you think about removing cartesian functionality from this PR?
I would also gravitate towards simplifying this testing infrastructure.

havogt: I would not mix it in here for sure. As I think I said last time, I would only add Cartesian testing if it comes for free. I didn't review the PR yet, but it seems there should be a script that you give it 2 program cache dirs and compares both. That should work both on cartesian and next, no?

kotsaloscv: The diffing of the two folders comes for free but the whole infrastructure to identify these two folders automatically is not for free or straightforward, and this is because cartesian has a completely different caching mechanism compared to next.
In any case, I will strip this PR from cartesian and leave only next. That would make it easier to review and digest.

kotsaloscv:
cscs-ci run dace-determinism

scripts/dace_deterministic_codegen.py
kotsaloscv: @havogt @tehrengruber I removed the cartesian from the infrastructure.
Feel free to review the updated code.

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

havogt:
Can you explain that part, I don't follow

>How many programs a logical name produces is determined by the tests — which ones run,
and with which compile-time parameters — not by codegen. Codegen fixes the contents of
each generated program, never the number of them. The comparison builds on this:

>    A name is compared only when both runs produced the same number of programs for it
    (count1 == count2). Equal counts mean the same programs were generated in both runs,
    so any difference in their source multisets is codegen non-determinism. The check is
    order-independent (multiset equality), which makes it robust to parallel execution
    under pytest-xdist.

kotsaloscv:
> Can you explain that part, I don't follow
> 
> > How many programs a logical name produces is determined by the tests — which ones run,
> > and with which compile-time parameters — not by codegen. Codegen fixes the contents of
> > each generated program, never the number of them. The comparison builds on this:
> 
> > A name is compared only when both runs produced the same number of programs for it
> > (count1 == count2). Equal counts mean the same programs were generated in both runs,
> > so any difference in their source multisets is codegen non-determinism. The check is
> > order-independent (multiset equality), which makes it robust to parallel execution
> > under pytest-xdist.

FYI, @havogt

This is a typical view of the cache root:
First level of each cache root — one directory per `(program, parametrization)`, named `<program-name>_<sha256>`:

```text
run1/.gt4py_cache/                 # one directory per (program, parametrization)
  __field_operator_abs_field_operator_4d203261…82653c7/
  __field_operator_arithmetic_2c539886…64c985e/
  __field_operator_simple_if_d599c22d…a5cf350d/      ┐ same program name,
  __field_operator_simple_if_9f88c359…a0057064/      ┘ two parametrizations → two folders
  solve_nonhydro_stencil_52_like_8056ccf7…fd72a9d2/
  ...

run2/.gt4py_cache/
  __field_operator_abs_field_operator_4d203261…82653c7/   ← SAME hash (deterministic → MATCH)
  __field_operator_arithmetic_bddadc61…2a999240/          ← hash CHANGED (non-deterministic → DIFFER)
  __field_operator_simple_if_be34e4a0…c1913df1/
  __field_operator_simple_if_94482a7e…2ed27b5e/
  solve_nonhydro_stencil_52_like_707d6861…78d2fef8/
  ...
```

` __field_operator_simple_if` appears twice in the cache root with a different hash appended. Between run1 & run2 we cannot know which run1-hash corresponds to the run2-hash. Therefore, we create multisets per name (e.g. ` __field_operator_simple_if`) where we collapse/encode these two folders. Regarding, your question about the first paragraph: The multiple cache folders per logical name (e.g. ` __field_operator_simple_if`) are not due to indeterminism but due to different parametrisation of the SDFG.

Does it make sense? Otherwise, we can have a quick call.

havogt:
So now compare the set of programs? What about non-parametrized tests with the same name?

kotsaloscv:
> So now compare the set of programs? What about non-parametrized tests with the same name?

Same approach, but now the multiset has just one element and so the comparison is just the trivial one-to-one correspondence.

havogt:
> > So now compare the set of programs? What about non-parametrized tests with the same name?
> 
> Same approach, but now the multiset has just one element and so the comparison is just the trivial one-to-one correspondence.

I don't get it. Maybe you can explain to me tomorrow.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,408 @@
...
+
+
+class DeterminismError(RuntimeError):
+    def __init__(self, message: str, results: list[NameResult]) -> None:
+        super().__init__(message)
+        self.results = results
+
+
+@dataclasses.dataclass
+class NameResult:
```
tehrengruber:
```suggestion
class ComparisonResult:
    """
    Comparison between two programs.
    
    Note: The same program can be compiled with different options (e.g. different static parameters).
    Since both, the options, as well as in-determinism only result in a change of hash in the cache
    folder path, we can not easily match folders between two runs. Instead this class collects the
    information for all the various options in one comparison result.
```

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,408 @@
...
+            for backend in sorted(d for d in src_root.iterdir() if d.is_dir()):
+                if backend.name not in SUPPORTED_BACKENDS:
+                    raise UnsupportedBackendError(
+                        f"unsupported dace backend `{backend.name}/` under {src_root}; "
+                        f"supported top-level backends are {sorted(SUPPORTED_BACKENDS)} "
+                        f"(HIP lives under cuda/hip/)"
+                    )
+            for path in sorted(src_root.rglob("*")):
+                if path.is_file():
+                    sources.append((path.relative_to(folder).as_posix(), _sha256(path)))
```
tehrengruber: It's hard for me to follow the flow of information. Here `sources` becomes a tuple of the relative path and the hash of the absolute path, whereas the only place where the latter is read at all in the sorting function [here](https://github.com/GridTools/gt4py/pull/2590/changes?diff=unified#diff-3dce8b08e4a8323917aa2ee85d5a3b37f42664ccab70e21400d5ce68914a4854R198) (I believe?). What is `_sha256(path)` used for?

tehrengruber (commented):
Let's have a quick chat before I continue.

scripts/dace_deterministic_codegen.py
```diff
@@ -0,0 +1,408 @@
...
+        super().__init__(message)
+        self.results = results
+
+
+@dataclasses.dataclass
+class NameResult:
+    name: str
+    comparable: bool  # both runs compiled the same number of programs
+    match: bool  # comparable and the two signature multisets are identical
+    only_in_run1: list[tuple[str, str]]
```
tehrengruber: Let's use a type annotation that explains the semantics:
```suggestion
    only_in_run1: list[SourceFileHash]
```
with `SourceFileHash = tuple[str, str] # (source file path, source file contents hash)`. Then `collections.Counter` can also be `collections.Counter[SourceFileHash]` in other places.

kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

tehrengruber (approved):

kotsaloscv:
cscs-ci run dace-determinism

PR #2467, opened 2026-01-30
feat[next-dace]: lowering mixed-GTIR to SDFG

edopao:
We introduce a configuration option in the dace backend to run some transformations on the GTIR representation before lowering it to SDFG. The new configuration (`apply_common_transform=True`) is tested as a new dace backend (`dace_cpu_gt`).

PR #2593, opened 2026-05-06
fix[next-dace]: Map stride of local dimension in nested SDFG

edopao:
This PR contains 3 kinds of changes:

1. A bugfix in mapping the stride of local dimension into the nested SDFG of if expressions.
```
inner_desc = dace.data.Array(
    dtype=arg_desc.dtype,
    shape=(arg_desc.shape[local_dim_pos],),
    strides=(arg_desc.strides[local_dim_pos],),
)
```

2. Addition of connectivities. This is needed to support the pattern with `can_deref` expressions (not part of this PR, see #2467). After GTIR transformations, which unroll the neighbors loops, a common pattern in grids with skip values is `if(can_deref(a, V2E), deref(a(V2E)) + 2, 0)`.

3. Renaming of function parameters.

Blueline performance is in pair with baseline.


src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -948,15 +971,18 @@ def write_output_of_nested_sdfg_to_temporary(inner_value: ValueExpr) -> ValueExp
...
+        used_connectivities = {
+            aname
+            for aname, adesc in nsdfg.arrays.items()
+            if gtx_dace_args.is_connectivity_identifier(aname) and not adesc.transient
+        }
+
         nsdfg_node = self.state.add_nested_sdfg(
             nsdfg,
-            inputs=set(input_memlets.keys()),
+            inputs=(used_connectivities | input_memlets.keys()),
```
philip-paul-mueller:
```suggestion
            inputs=sorted(used_connectivities | input_memlets.keys()),
```

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -948,15 +971,18 @@ def write_output_of_nested_sdfg_to_temporary(inner_value: ValueExpr) -> ValueExp
...
             outputs=outputs,
```
philip-paul-mueller:
```suggestion
            outputs=sorted(outputs),
```

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -919,13 +927,28 @@ def write_output_of_nested_sdfg_to_temporary(inner_value: ValueExpr) -> ValueExp
             if isinstance(sym_type, IteratorExpr)
         }
         direct_deref_iterators = (
```
philip-paul-mueller, line 954: Shouldn't this be sorted too?

philip-paul-mueller (commented):
I am not fully happy with this change because it has potentially big implication for the `IfMover` and the condition fuser (which will stop working, since it can no longer identifying which If-Blocks can be moved together.
Because of these implication we must absolutely make sure to also run MuPhys, because, it will be very much impacted by these changes.

I am actually not sure if the data locality is actually improved, because in the almost everything has a `restrict` keyword, so the compiler should be smart enough tio figuring that out.
But this is just my concern or have you different opinions @iomaganaris

edopao:
> I am not fully happy with this change because it has potentially big implication for the `IfMover` and the condition fuser (which will stop working, since it can no longer identifying which If-Blocks can be moved together. Because of these implication we must absolutely make sure to also run MuPhys, because, it will be very much impacted by these changes.
> 
> I am actually not sure if the data locality is actually improved, because in the almost everything has a `restrict` keyword, so the compiler should be smart enough tio figuring that out. But this is just my concern or have you different opinions @iomaganaris

I see your point, makes sense. This change is not part of the bugfix, so I could revert it.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
edopao: No, not needed.

edopao:
I have applied the review comments from @philip-paul-mueller, I agree totally. I have reverted the modification to the lowering of the condition: this change introduced a performance regression. Below the blueline performance plot:
- gt4py_timers_fix_stride: the current state of the PR, after reverting the modification to the lowering of the condition.
- gt4py_timers_fix_stride_and_move_cond: the previous state of the PR, which moved the evaluation of the if-condition inside the nested SDFG.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -707,37 +709,50 @@ def _visit_if_branch_arg(
...
-        if param_name in if_sdfg.arrays:
+        if param_name in sdfg.arrays:
             # the data desciptor was added by the visitor of the other branch expression
-            assert if_sdfg.data(param_name) == inner_desc
+            assert sdfg.data(param_name) == inner_desc
         else:
-            if_sdfg.add_datadesc(param_name, inner_desc)
-            if_sdfg_input_memlets[param_name] = arg_expr
+            sdfg.add_datadesc(param_name, inner_desc)
+            input_memlets[param_name] = arg_expr
```
philip-paul-mueller, line 742: Question: If this function is called in a uniform way I would expect that the data descriptor is either already added to the sdfg or not in all cases. So this `if` locks a bit fishy to me but maybe I miss something.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -948,14 +981,21 @@ def write_output_of_nested_sdfg_to_temporary(inner_value: ValueExpr) -> ValueExp
 
         outputs = {outval.dc_node.data for outval in gtx_utils.flatten_nested_tuple((result,))}
 
+        # map the connectivities that were used inside the nested SDFG
+        used_connectivities = {
```
philip-paul-mueller: This is also a set, I am not fully sure if it matters, but to promote future safty I would propose to either sort it or modify the loop below such that it does the filtering too as it iterates over a `dict`.

philip-paul-mueller (approved):
LGTM, although I have two smallish points.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
edopao: I see the issue, I have changed `used_connectivities` to a list.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
edopao: The origin of this if is that I am adding globals as I encounter them while visiting the if branches. I am visiting one branch at a time, so it could be that some globals have already been added in the previous branch.

PR #2596, opened 2026-05-08
feat[cartesian] Use default CXX all the time + Intel traceback for debug

FlorianDeconinck:
## Description

The original design of the `cxx_compile_flags` was to feed it in the `GT4PY_EXTRA_COMPILE_ARGS`. This means we can have user override the internal decision.
We propose to swap this strategy - we control flags.

On this PR we also had `-traceback` and `-g` to debug compile in order to gather a better stack on crash or FPE trap.

## Requirements


romanc (commented):
I'm fine with flipping the behavior. We might wanna shoot a message to the PMAP people.

Why are `"apple clang"` and `"clang"` not getting a `-g` flag in case of `-O0`?

FlorianDeconinck:
> I'm fine with flipping the behavior. We might wanna shoot a message to the PMAP people.
> 
> Why are `"apple clang"` and `"clang"` not getting a `-g` flag in case of `-O0`?

Good catch, added.

romanc (approved):
LGTM 👍

PR #2599, opened 2026-05-12
fix[next]: add missing state to cache fingerprints

egparedes:
Add missing state to the fingerprints used as toolchain cache keys, including current gt4py version.


egparedes:
Closing in favor or #2609

PR #2600, opened 2026-05-12
build: Use dace release package from pypi

edopao:
- Use same `dace` version for GT4Py cartesian and next.
- Make `dace` a GT4Py dependency, and remove the development groups `dace-cartesian` and `dace-next`.
- Instead of using the `gridtools` package index, pull the dace pre-release package from PyPI index.

Note: if in the future a custom (non-pypi) version is needed, cartesian and next need to create a version from a common branch (and publish to the gridtools index).


philip-paul-mueller (commented):
I am actually not sure if we should give up the index yet.
Because as soon as we need a special PR or so for DaCe we are back to the same solution as before.
For example we can not use the new GPU codegen until that is merged, which will take a long time.

Another question is what should happen with the GT4Py PyPI index?
This is also a question should we remove the entry from the `pyproject.toml`?

Maybe @egparedes you have some opinions on that.

pyproject.toml
edopao: I have left this example as a reference in case we need to pull from the gridtools index.

pyproject.toml
```diff
@@ -481,13 +472,10 @@ explicit = true
 name = 'gridtools'
 url = 'https://gridtools.github.io/pypi/'
 
-# -- Example: {index = "gridtools", group = "dace-next"}
+# Add the uv source below to pull dace from the gridtools index instead of PyPI:
+# dace = {index = "gridtools"}
 [tool.uv.sources]
 atlas4py = {index = "test.pypi"}
```
havogt, line 478:
```suggestion
[tool.uv.sources]
atlas4py = {index = "test.pypi"}
# Add the uv source below to pull dace from the gridtools index instead of PyPI:
# dace = {index = "gridtools"}
```
nit

havogt (approved):
lgtm

pyproject.toml
edopao: I agree, but the TOML formatter will delete the comment lines inside the `tool.uv.sources` section.

egparedes: Isn't there something like `noqa:` for the toml formatter?

egparedes (approved):
LGTM

pyproject.toml
edopao: I tried but did not succeed.

romanc (approved):
LGTM

PR #2602, opened 2026-05-12
fix[cartesian] No OpenMP on device compile

FlorianDeconinck:
## Description

Do not apply `open_mp` flags on device compilers - we don't do multithreading host code for GPU backends (and device is plain wrong).

This flag usage was discovered in the case of an `icx/nvcc` where the `-qopenmp` intel flag bled into the command line of `nvcc`.

## Requirements


romanc (approved):
Thanks for cleaning this up!

PR #2603, opened 2026-05-15
refactor[next-dace]: Add library node for reduce with skip values

edopao:
DaCe has a library node in its standard library for reduction. However, GT4Py and ICON4Py also need to lower reduction expressions with skip values. The lowering was already done, on main, inside a nested SDFG. All this PR does is to refactor the code that creates the nested SDFG and make an application library node in GT4Py. In this way, the SDFG nodes for regular reduction and the nodes for reduction with skip values are represented in the same way in the SDFG.

No performance regression in ICON4Py blueline:


edopao:
cscs-ci run dace

edopao:
cscs-ci run default

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
```diff
@@ -0,0 +1,176 @@
...
+class ReduceWithSkipValues(dace.sdfg.nodes.LibraryNode):
+    """Implements reduction with skip values."""
+
+    implementations: Final[dict[str, dace_transform.ExpandTransformation]] = {}
+    default_implementation: Final[str | None] = "pure"
+
+    # Properties
+    wcr = dace_properties.LambdaProperty(default="lambda a, b: a")
+    identity = dace_properties.Property(default=0)
+    init = dace_properties.Property(default=0)
```
philip-paul-mueller: I think these properties should not have a default.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
```diff
@@ -0,0 +1,176 @@
...
+    input_conn = dace_properties.Property(default="_in")
+    output_conn = dace_properties.Property(default="_out")
+    mask_conn = dace_properties.Property(default="_mask")
```
philip-paul-mueller: I think it does not make much sense to provide the possibility to use different connector names.
However, if this feature is needed, then there should not be a default and the user should always be forced to make an explicit decision.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
```diff
@@ -0,0 +1,176 @@
...
+            self.mask_conn = mask_conn
+
+        super().__init__(name, inputs={input_conn, mask_conn}, outputs={output_conn})
+        self.wcr = wcr
+        self.identity = identity
+        self.init = init
+        self.debuginfo = debuginfo
+
+    def validate(self, sdfg: dace.SDFG, state: dace.SDFGState) -> None:
+        assert len(list(state.in_edges_by_connector(self, self.input_conn))) == 1
```
philip-paul-mueller: I think this should not be an `assert` but rather a `raise` expression.
Because now if `asserts` are disabled multiple incoming edges with the same connector name is considered valid.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
```diff
@@ -0,0 +1,176 @@
...
+            state.in_edges_by_connector(node, node.mask_conn)
+        )
+        input_desc = sdfg.arrays[inedge.data.data]
+        output_desc = sdfg.arrays[outedge.data.data]
+        mask_desc = sdfg.arrays[maskedge.data.data]
+        assert len(mask_desc.shape) == 2
+        max_neighbors = mask_desc.shape[1]
+        assert isinstance(max_neighbors, int) or str(max_neighbors).isdigit()
+
+        local_dim_index = inedge.data.src_subset.size().index(max_neighbors)
```
philip-paul-mueller, line 124: Might break in cases where the number of {edges, cell, vertices} is the same as local dimensions.
Probably too paranoid and I do not have an alternative.
But you should probably put a TODO or NOTE here explaining what it is doing and what the problems are.

philip-paul-mueller (commented):
There are some small things but nothing big.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
edopao: I tried, but it is not allowed:
`Error importing plugin "gt4py.next.type_system.mypy_plugin": Default not properly defined for property  [misc]`

By the way, the dace standard reduce does the same.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
edopao: As above, I cannot avoid setting the default.
The reason for allowing customization of the connector names is to allow setting up either a standard reduce or a reduce with skip values and then setup connections for the input and output with common logic. Also the standard reduce allows customization of the connector names.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
edopao: That's true. I was considering this case a very extreme one but better to take caution.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
edopao: I have an idea. I can check that the size of the subset is 1 in all dimensions except one, the local dimension. I will push a new commit.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
philip-paul-mueller: Just a random question, did you specify a `dtype` value?
Looking at the error and [Property::__init__()](https://github.com/spcl/dace/blob/a6717b608f497190205d5145b0dd50d1575458c0/dace/properties.py#L86) I would say that `dtype` is missing.
You can also play with `allow_none`.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
philip-paul-mueller: As above adding a `dytpe` should solve your problem.
See also my comment further down.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
```diff
@@ -0,0 +1,187 @@
...
+        init: dace.symbolic.SymbolicType,
+        debuginfo: dace.dtypes.DebugInfo | None = None,
+        input_conn: str | None = None,
+        output_conn: str | None = None,
+        mask_conn: str | None = None,
+    ) -> None:
+        if input_conn is None:
+            input_conn = self.input_conn
+        else:
+            self.input_conn = input_conn
```
philip-paul-mueller:
```suggestion
        if input_conn is not None:
            self.input_conn = input_conn
```

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
```diff
@@ -0,0 +1,187 @@
...
+            output_conn = self.output_conn
+        else:
+            self.output_conn = output_conn
+
+        if mask_conn is None:
+            mask_conn = self.mask_conn
+        else:
+            self.mask_conn = mask_conn
+
+        super().__init__(name, inputs={input_conn, mask_conn}, outputs={output_conn})
```
philip-paul-mueller, line 47:
```suggestion
        super().__init__(name, inputs={self.input_conn, self.mask_conn}, outputs={self.output_conn})
```

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
philip-paul-mueller: This should work (except for the case that there is only a single neighbour, which is probably unlikely).

src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -1530,28 +1401,65 @@ def _visit_reduce(self, node: gtir.FunCall) -> ValueExpr:
         offset_provider_type = self.subgraph_builder.get_offset_provider_type(offset_type.value)
         assert isinstance(offset_provider_type, gtx_common.NeighborConnectivityType)
 
+        inp_conn = "_in"
+        outp_conn = "_out"
+        mask_conn = "_mask"
```
philip-paul-mueller, line 1406: These are exactly the default from above.
That you set them here is indicating that they are not needed as default or here not needed.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -118,7 +121,7 @@ def add_tasklet(
         code: str,
         language: dace.dtypes.Language = dace.dtypes.Language.Python,
         **kwargs: Any,
-    ) -> dace_nodes.Tasklet:
+    ) -> tuple[dace_nodes.Tasklet, dict[str, str]]:
```
philip-paul-mueller, line 124: You should describe what the returned `dict` does.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -369,6 +372,15 @@ def gt_auto_optimize(
                     stacklevel=0,
                 )
 
+        # We now expand all GT4Py specific library nodes.
+        #  We do this such that we have control over all the Maps that are there.
+        # TODO(edopao,phimuell): It is probably the right place, but maybe there is a better one.
```
philip-paul-mueller, line 377: I am actually not sure if it is the best one.
I would maybe also think of moving it at the very end, however, given my experience with the broadcast node, we should probably keep it here for now.

src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py
```diff
@@ -751,8 +763,8 @@ def _gt_auto_process_dataflow_inside_maps(
     # NestedSDFGs inside the ConditionalBlocks it fuses.
     sdfg.apply_transformations_repeated(
         gtx_transformations.FuseHorizontalConditionBlocks(),
-        validate=True,
-        validate_all=True,
+        validate=False,
```
philip-paul-mueller, line 766: Good catch.

philip-paul-mueller (commented):
Some small suggestions.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
edopao: The problem is that `dtype` is not known. Yes, I could play with `allow_none` indeed.

src/gt4py/next/program_processors/runners/dace/library_nodes/reduce_with_skip_values.py
edopao: Yes, in this case adding `dtype` would solve the issue.

philip-paul-mueller (approved):
LGTM

PR #2604, opened 2026-05-20
fix[next]: cache dir construction and translation_cache lifetime

havogt:
- Before the .gt4py_cache dir is constructed on import time of gt4py. Now the directory is created on first write.
- Respect BUILD_CACHE_LIFETIME for the `translation_cache`.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/factory.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/factory.py
@@ -16,6 +16,7 @@ import factory
 from gt4py._core import definitions as core_defs, filecache
 from gt4py.next import config
 from gt4py.next.otf import recipes, stages, workflow
+from gt4py.next.otf.compilation import cache
 from gt4py.next.program_processors.runners.dace.workflow import (
     bindings as bindings_step,
     decoration as decoration_step,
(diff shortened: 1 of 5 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/factory.py
```diff
@@ -47,7 +48,12 @@ class Params:
                 lambda o: workflow.CachedStep(
                     o.bare_translation,
                     hash_function=stages.fingerprint_compilable_program,
-                    cache=filecache.FileCache(str(config.BUILD_CACHE_DIR / "translation_cache")),
+                    cache=filecache.FileCache(
+                        str(
+                            cache.get_cache_base_path(config.BUILD_CACHE_LIFETIME)
+                            / "translation_cache"
```
edopao, line 54: Do we need to use `dace_cache` (similarly to what the gtfn backend does, `gtfn_cache`) to be able to share the cache directory across backends?

Or, I believe, the backend key is already part of the translation hash and therefore different backends can coexist in the same translation folder?

edopao (approved):
I am approving this PR because it fixes a problem. My comment is just a discussion point, how we want to setup the persistent cache for different backends.

tehrengruber (approved):

PR #2608, opened 2026-05-27
fix[next-dace]: Remove Unused Data Descriptor

philip-paul-mueller:
In certain cases the lowering generates SDFGs that contains data descriptors that are not used anywhere. There was a recent change in DaCe which slightly changed how free symbols are computed. The effect was that the symbols of these unused data descriptors (i.e. the shape and size of arrays) must now be provided.
This PR changes the lowering in such a way that these transients are removed.

For more information see [DaCe issue#2382](https://github.com/spcl/dace/issues/2382).

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
+++ b/src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
@@ -677,7 +677,62 @@ class GTIRToSDFG(eve.NodeVisitor, SDFGBuilder):
             if output is not None and (dataname := output.dc_node.data) not in data_args
         }
 
+        connectivity_arrays = {
+            gtx_dace_args.connectivity_identifier(offset)
+            for offset in gtx_dace_args.filter_connectivity_types(self.offset_provider_type)
+        }
+
+        inner_ctx_globals = [
+            dataname
+            for dataname, datadesc in inner_ctx.sdfg.arrays.items()
+            if not datadesc.transient
+        ]
+
+        input_memlets = {}
+        for dataname in inner_ctx_globals:
+            if dataname in data_args:
+                # The global data refers to an argument. However, it might not be used.
+                # If it is used, i.e. `data_args[dataname]` is not `None` then we will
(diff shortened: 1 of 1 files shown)


src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -725,9 +707,8 @@ def add_nested_sdfg(
                 # the number of input connectors, this check is necessary for tuple
                 # arguments, for which domain inference has detected that one or more
                 # of the nested fields is not used. In such cases, the corresponding
-                # argument in the top-level lambda is expected to be None (see how
-                # this case is handled in the if-branch above) and it is not possible
-                # to setup an input edge.
+                # argument in the top-level lambda is expected to be None to setup an
+                # input edge.
```
edopao:
```suggestion
                # argument in the top-level lambda is expected to be None and it is not possible
                # to setup an input edge.
```

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -709,10 +691,10 @@ def add_nested_sdfg(
         input_memlets = {}
         for dataname in inner_ctx_globals:
             if dataname in data_args:
-                # Uninitialized arguments should not be used inside the nested SDFG.
-                if (arg_node := data_args[dataname]) is None:
-                    inner_ctx.sdfg.remove_data(dataname, validate=gtx_config.DEBUG)
-                else:
+                # The global data refers to an argument. However, it might not be used.
+                # If it is used, i.e. `data_args[dataname]` is not `None` then we will
+                # fully map it into the nested SDFG. Otherwise we will remove bellow.
```
edopao:
```suggestion
                # The global data refers to an argument. If the argument is initialized, 
                # i.e. `data_args[dataname]` is not `None`, the argument is used and
                # we fully map it into the nested SDFG. Otherwise we remove it below.
```

edopao (approved):

PR #2609, opened 2026-05-27
fix[next]: extend fingerprinting mechanism to consider all relevant state in caches

egparedes:
## Summary

Generalizes and unifies the fingerprinting / cache-key infrastructure used by the `gt4py.next` workflow caches, so cache keys are **deterministic across interpreter runs** and get **invalidated when any relevant state changes** — the cached step's own configuration and the gt4py version, not just the step input.

The design and rationale are documented in the new ADR `docs/development/ADRs/next/0023-Fingerprinting-and-Custom-Picklers.md`; this description only summarizes the changes.

## Key changes

- **`eve.utils` pickling primitives:** add `content_hash(..., pickler_type=...)`, `pickle_reducer_factory`, `singledispatcher`, `merge_dispatchers`, `get_fully_qualified_name`, and `PurePickler` (pure-Python `pickle._Pickler`); remove the old `custom_pickler*` helpers and `eve.concepts.skipping_fields_node_pickler` / `NodeFingerprinter`.
- **`next` fingerprinters:** add `stable_fingerprinter` (default; sorts `dict`/`set` and reduces types/functions/modules by reference) and `skipping_fields_node_fingerprinter`; rename `MetadataBasedPickling` → `MetadataBasedPicklingMixin` (drops fields marked `gt4py_metadata(pickle=False)`). The ffront/ITIR `semantic_fingerprinter` and the stage `SemanticPickler` (functions fingerprinted by source + closure) are built on these; the old per-node `fingerprint()` methods and the `add_content_to_fingerprint` visitor are removed.
- **Cache keys:** `CachedStep` renames `hash_function` → `key_function`, excludes `key_function`/`cache` from its own fingerprint, and keys on `stable_fingerprinter((BUILD_CACHE_VERSION_ID, self, key_function(inp)))`. New `config.BUILD_CACHE_VERSION_ID` (defaults to the gt4py version, env-overridable) salts the cache. Runners (`gtfn`, `dace`, …) adopt `stable_fingerprinter` as their `key_function`.
- **Docs / tests:** add ADR 0023; update `WorkflowPatterns.md` for the rename; add/update unit tests for the new primitives, fingerprinters, and `CachedStep`.


src/gt4py/next/utils.py
```diff
@@ -166,50 +166,71 @@ class CustomPicklingFingerprinter:
...
-    reduce_dispatcher: xtyping.SingleDispatchCallable[[Any], str]
-    name: str = dataclasses.field(default="CustomPicklingFingerprinter", kw_only=True)
-
-    if TYPE_CHECKING:
-        pickler: pickle.Pickler
 
     def __post_init__(self) -> None:
+        registered_types = set(self.reduce_dispatcher.registry.keys())
+        skip_builtin_types = not ({dict, set, list, tuple} & registered_types)
+        if not skip_builtin_types and types.ModuleType not in registered_types:
```
edopao: I don't follow this code. Maybe Caude does, but for me it contains too many negations.

src/gt4py/eve/utils.py
```diff
@@ -630,10 +630,58 @@ def is_noninstantiable(cls: Type[_T]) -> bool:
...
+) -> type[pickle.Pickler]:
+    """
+    Create a custom pickler class using the provided function as reducer override.
+    """
+    if skip_builtin_types:
+        base_pickler_class = pickle.Pickler
+    else:
+        # The public Pickler (implemented in C) shortcuts some built-in types for performance.
+        # If we use want to override built-in types, we need to the _Pickler pure-Python
+        # implementation.
```
edopao:
```suggestion
        # If we want to override built-in types, we need to use the _Pickler pure-Python
        # implementation.
```

src/gt4py/next/ffront/stages.py
```diff
@@ -22,24 +22,41 @@
...
-from gt4py.next import common
+from gt4py.next import common, utils
 from gt4py.next.ffront import field_operator_ast as foast, program_ast as past, source_utils
 from gt4py.next.otf import arguments, toolchain
 
 
+# Create a custom pickler for the BaseStage `fingerprinter` that handles
+# `types.FunctionType` by using its source code and closure variables.
+# This should be enough for the use case of GT4Py DSL definitions,
+# which are expected to be pure functions without complicated closures.
```
edopao: I don't know at which stage this pickler is invoked, so I don't know if my question makes sense. For compile-time functions, should we also include the compile-time inferred domain?

src/gt4py/next/iterator/ir.py
```diff
@@ -7,30 +7,22 @@
...
-_semantic_node_pickler: type[pickle.Pickler] = concepts.skipping_fields_node_pickler(
-    "location", "type"
-)
-
-#: Generates an unique fingerprint for IR nodes ignoring their location and type attributes.
-semantic_fingerprint: concepts.NodeFingerprinter = functools.partial(
-    eve_utils.content_hash, pickler=_semantic_node_pickler
+#: Generate an unique fingerprint for `eve.Node`s ignoring its "location" attribute.
+semantic_fingerprinter: utils.CustomPicklingFingerprinter = (
+    utils.skipping_fields_node_fingerprinter("location", "type")
```
edopao: The comment only mentions `location`, but you also ignoring `type`. Besides, should we really ignore `type`?

src/gt4py/next/config.py
```diff
@@ -90,6 +90,18 @@ def env_flag_to_int(name: str, default: int) -> int:
...
+    from gt4py import __version__
+
+    return __version__
+
+
+#: Version ID for the build cache. It should only be overridden by advanced users
+#: testing toolchain changes that are expected to break compatibility with previously cached builds.
+BUILD_CACHE_VERSION_ID: str = (
+    os.environ.get("BUILD_CACHE_VERSION_ID") or _get_build_cache_version_id()
+)
```
edopao: I would expect this `BUILD_CACHE_VERSION_ID` to be included in some hash, is it?

edopao (commented):
A first round of review comments.

src/gt4py/next/utils.py
egparedes: Agreed. Changed for clarity.

src/gt4py/next/iterator/ir.py
egparedes: Both location and type were already excluded. This change is preserving the current behavior but the new comment is wrong (copy-paste error). For the actual reason to exclude `type` you need to ask @tehrengruber but IIRC is related to ignore types which hasn't being yet resolved (`deferred`)

src/gt4py/next/ffront/stages.py
egparedes: This comment is related to the exact custom pickler fingerprinter defined for the classes defined in the module. But, to answer your question, yes, compile-time constants should also be included in the fingerprint, and it's done by also hashing the Args classes which contain this information (as part of `ConcreteArtifact` I think)

src/gt4py/next/config.py
egparedes: Yes, exactly, but it will be done in a following PR, so I'll remove this leftover here.

src/gt4py/next/otf/compilation/compiler.py
```diff
@@ -55,6 +55,7 @@ class Compiler(
         stages.ExecutableProgram,
     ],
     definitions.CompilationStep[CPPLikeCodeSpecT, code_specs.PythonCodeSpec],
+    utils.MetadataBasedPicklingMixin,
```
havogt, line 58: Adding `MetadataBasedPicklingMixin` here makes `Compiler` *fingerprintable*, but `Compiler` is never wrapped in a `CachedStep` anywhere I can find — it caches via `_cache_folder_name(source)` directly. So none of `Compiler`'s state (`cache_lifetime`, `builder_factory`, `force_recompile`) actually reaches a cache key today; the mixin is dead weight at this layer until the compilation step is also brought under the new machinery.

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -167,12 +167,12 @@ class Params:
...
         cached = factory.Trait(
             executor=factory.LazyAttribute(
-                lambda o: workflow.CachedStep(o.otf_workflow, hash_function=o.hash_function)
+                lambda o: workflow.CachedStep(o.otf_workflow, key_function=o.key_function)
             ),
             name_cached="_cached",
         )
         device_type = core_defs.DeviceType.CPU
-        hash_function = stages.compilation_hash
+        key_function = utils.stable_fingerprinter
```
havogt, line 175: `cmake_build_type` is a `GTFNCompileWorkflowFactory.Params` field plumbed only into `CompiledbFactory`. It is not a field on `GTFNTranslationStep` (compare to `DaCeCompiler.cmake_build_type`), so toggling `GT4PY_CMAKE_BUILD_TYPE` does not change the translation-step's fingerprint or the on-disk folder. Same hole as the `FIXME[#2447]` in `config.py:95`. Worth surfacing `cmake_build_type` as a field on `GTFNTranslationStep` (or on `Compiler`, once that step is wrapped) so it rides into `self` in `CachedStep.cache_key`.

src/gt4py/next/otf/arguments.py
```diff
@@ -140,7 +140,7 @@ def from_signature(cls, *args: Any, **kwargs: Any) -> Self:
 
 
 @dataclasses.dataclass(frozen=True)
-class CompileTimeArgs:
+class CompileTimeArgs(utils.MetadataBasedPicklingMixin):
```
havogt, line 143: Since `offset_provider` is still the runtime type (per the `TODO(havogt)` on the field below), the mixin pickles the full neighbor-table `ndarray` buffers on every cache lookup. The sorted-`dict` reducer in `stable_fingerprinter` makes it deterministic but doesn't make it cheap — on ICON-class grids this is megabytes of indices hashed per call. Worth either marking `offset_provider` with a custom reducer that hashes by `(connectivity_type, ndarray.shape, ndarray.dtype, ndarray.tobytes())` once and caches, or expediting the `offset_provider_type` switch so only the type, not the data, flows into the fingerprint.

src/gt4py/next/utils.py
```diff
@@ -138,6 +140,146 @@ def __getstate__(self) -> _StandardPickleState:
...
+    def __call__(self, instance: Any) -> str:
+        """Fingerprint the given object."""
+        return eve_utils.content_hash(instance, pickler_type=self.pickler_type)
+
+
+def _stable_container_sort_key(obj: Any) -> Any:
+    """Return a deterministic sort key for container elements/keys."""
+    if isinstance(obj, (type, types.FunctionType, types.ModuleType)):
+        return eve_utils.custom_pickle_reduce_by_reference(obj)
+    return obj
```
havogt: The fallback `return obj` means dicts/sets keyed by arbitrary user types get sorted with `<` on those objects. `common.Dimension` is a frozen dataclass without `__lt__`; a cached step holding `dict[Dimension, ...]` would raise `TypeError: '<' not supported between instances of 'Dimension' and 'Dimension'` inside `stable_fingerprinter`. Either fall back to `repr(obj)` / a content hash, or add reducers for the small set of opaque-but-key-able types you actually expect to see as dict keys (`Dimension`, offset tags, etc.).

havogt (commented):
A few notes on gaps the new fingerprinter machinery doesn't close — mostly pre-existing, flagging since they survive after this PR, not as blockers. (Some of these belonged on lines not touched by the PR, hence in the body rather than inline.)

**`library_deps` is meaningless on the disk-folder side.** `dace/workflow/translation.py:475` sets `library_deps=tuple()`, and `gtfn_module.py:248` hardcodes `LibraryDependency(library_name, "master")`. `_cache_folder_name` does include `library_deps` in its hash, so the channel exists — but bumping the gridtools header revision or the dace package version stays invisible to the on-disk cache folder. Worth feeding real version strings here, or salting `_cache_folder_name` with `dace.__version__` plus a gtfn-header revision.

**`_cache_folder_name` (`otf/compilation/cache.py:45`) sits outside the new machinery.** It hand-rolls `hashlib.sha256` over a serialized string of `code_spec`, entry-point name, params, `library_deps`, and `source_code`. The two cache-key schemes can drift — add a field to `ProgramSource` for the fingerprinter-based caches above and you'd silently *not* invalidate this folder. Worth folding into the same machinery as a follow-up.

**`set_dace_config` (`dace/workflow/common.py:27`) is a pure side channel into codegen + compile.** It mutates thread-local `dace.Config` (`compiler.cpu.args`, `compiler.cuda.args`, `compiler.cuda.hip_args`, `compiler.cuda.max_concurrent_streams`, …) and reads `CXXFLAGS`/`CUDAFLAGS`/`HIPFLAGS` from `os.environ`. None of these reach any cache key. Toggling `CXXFLAGS=-O0` between runs silently reuses the previous `.so`. Pre-existing, but the new machinery makes the asymmetry obvious — worth a `FIXME[#2447]` here too, or, better, capturing these as explicit step fields so they flow into the cache key.

tests/eve_tests/unit_tests/test_utils.py
```diff
@@ -116,6 +116,143 @@ class Derived(Base):
     assert dispatcher.registry.keys() == {object, Base, Derived}
 
 
+def test_singledispatcher_default_from_object_implementation():
+    from gt4py.eve.utils import singledispatcher
```
edopao, line 120: Just to know: what is the reason for importing `singledispatcher` inside the test body?

edopao (commented):
LGTM

src/gt4py/eve/utils.py
```diff
@@ -630,10 +642,104 @@ def is_noninstantiable(cls: Type[_T]) -> bool:
     return "__noninstantiable__" in cls.__dict__
 
 
+def singledispatcher(
+    default: Callable[P, T] | None = None, *, implementations: dict[type, Callable[[Any], Any]]
+) -> xtyping.SingleDispatchCallable[P, T]:
```
havogt, line 647: docstring?

tests/next_tests/unit_tests/otf_tests/test_workflow.py
```diff
@@ -82,3 +95,54 @@ def test_replace():
...
+    # Different modules produce different fingerprints.
+    assert utils.stable_fingerprinter({"mod": os}) != utils.stable_fingerprinter({"mod": sys})
+
+
+def test_cached_step_cache_key_includes_step_config():
+    """Two CachedStep instances with different step configs produce different cache keys."""
+    # Use dataclass steps with different configurations (v=1 vs v=2)
+    cs1 = workflow.CachedStep(step=_StepWithValue(v=1), key_function=lambda x: x)
+    cs2 = workflow.CachedStep(step=_StepWithValue(v=2), key_function=lambda x: x)
+    assert cs1.cache_key(42) != cs2.cache_key(42)
```
havogt, line 149: **`BUILD_CACHE_VERSION_ID` invalidation isn't tested.** The whole point of the version salt is to be the global emergency switch when fingerprints alone can't capture a toolchain change. Without a test, a future refactor could silently drop the salt from `CachedStep.cache_key` and this whole suite would still pass. A sibling test next to `test_cached_step_cache_key_includes_step_config`:

```python
def test_cache_key_changes_with_BUILD_CACHE_VERSION_ID(monkeypatch):
    from gt4py.next import config
    cs = workflow.CachedStep(step=_append_one, key_function=lambda x: tuple(x))
    k1 = cs.cache_key([1, 2])
    monkeypatch.setattr(config, "BUILD_CACHE_VERSION_ID", "different-version")
    k2 = cs.cache_key([1, 2])
    assert k1 != k2
```

tests/next_tests/unit_tests/test_utils.py
```diff
@@ -91,6 +94,97 @@ def test_pickle_roundtrip(self, instance, expected_fields):
             assert getattr(restored, field_name) == expected_value
```
havogt, line 94: Two coverage gaps in this class:

**1. `pickle=False` → fingerprint contract is left implicit.** `test_pickle_roundtrip` confirms the transient field is dropped from `__getstate__`, but the property contributors actually care about — "a `pickle=False` field doesn't change the fingerprint" — is inferred rather than asserted. A direct test pins it:

```python
def test_pickle_false_fields_do_not_change_fingerprint(self):
    a = _DataclassModel(value=1, transient="foo")
    b = _DataclassModel(value=1, transient="bar")
    assert utils.stable_fingerprinter(a) == utils.stable_fingerprinter(b)
```

**2. Missing the `__slots__` + `__dict__` branch of `_get_metadata_based_state_getstate`.** The function has four code paths (no-slots/no-dict, only-dict, only-slots, both). The `@pytest.mark.parametrize` at line 60 exercises three; the `(slots, dict)` tuple state (`return ({dict_names…}, {slot_names…})`) is unexercised. The `(slots, dict)` branch is the one whose default-`__setstate__` interaction is least intuitive, so it's the one worth pinning. Add a model that inherits from a slotted parent without itself declaring `__slots__`:

```python
@dataclasses.dataclass(slots=True)
class _SlotsParent(utils.MetadataBasedPicklingMixin):
    value: int

@dataclasses.dataclass
class _SlotsAndDictChild(_SlotsParent):
    extra: int = 7  # has __dict__ on top of slotted parent
```

and an extra entry in both `parametrize` lists.

src/gt4py/next/otf/workflow.py
```diff
@@ -195,81 +203,87 @@ class StepSequence(ChainableWorkflowMixin[StartT, EndT]):
...
     """
 
     step: Workflow[StartT, EndT]
-    hash_function: Callable[[StartT], HashT] = dataclasses.field(default=hash)  # type: ignore[assignment]
-    cache: OpaqueMutableMapping[HashT, EndT] = dataclasses.field(repr=False, default_factory=dict)
+    key_function: Callable[[StartT], HashT] = dataclasses.field(
+        metadata=utils.gt4py_metadata(pickle=False)
+    )
+    cache: OpaqueMutableMapping[str, EndT] = dataclasses.field(
+        repr=False, default_factory=dict, metadata=utils.gt4py_metadata(pickle=False)
```
havogt, line 264: **`CachedStep` pickle round-trip is unspecified.** `MetadataBasedPicklingMixin` deliberately doesn't implement `__setstate__`, and `pickle`/`copy.deepcopy` bypass `__init__` (so `default_factory=dict` here never fires). After unpickling, `self.cache` is likely missing (`AttributeError` on first call). Today that's fine because pickling is only used for fingerprinting — but the docstring above doesn't say so, and nothing prevents a caller from `copy.deepcopy(cs)`-ing a `CachedStep` and trying to use it. Worth either:

- a test that asserts the contract works (`restored = pickle.loads(pickle.dumps(cs)); restored([1, 2])` succeeds), or
- a test that asserts the contract *doesn't* work (`pytest.raises(AttributeError)`), plus a one-line note in the class docstring saying "pickle is used for fingerprinting only; unpickled instances are not callable".

Either answer is fine; today the behavior isn't documented either way.

src/gt4py/next/utils.py
```diff
@@ -138,6 +140,113 @@ def __getstate__(self) -> _StandardPickleState:
...
+
+class SingleDispatchPickler(pickle.Pickler):
+    reducer_override: xtyping.SingleDispatchCallable[Any, tuple]
+
+
+def _stable_container_sort_key(obj: Any) -> Any:
+    """Return a deterministic sort key for container elements/keys."""
+    if isinstance(obj, (type, types.FunctionType, types.ModuleType)):
+        return eve_utils.get_fully_qualified_name(obj)
+    return obj
```
havogt, line 154: **The `return obj` fallback has no regression test and is a latent footgun.** Any user dataclass without `__lt__` — `common.Dimension`, `common.UnitRange`, `common.Domain`, ITIR's `Sym` — used as a dict key will raise `TypeError: '<' not supported between instances of 'Dimension' and 'Dimension'` inside `stable_fingerprinter`. This is exactly the kind of thing that bites when someone adds a new `dict[Dimension, X]` field to a cached dataclass downstream. Even if you decide not to fix the fallback yet, a regression test would document the boundary:

```python
def test_stable_fingerprinter_handles_dimensions_as_dict_keys():
    from gt4py.next import common
    I = common.Dimension("I")
    J = common.Dimension("J")
    # Currently raises TypeError; the right answer is order-independent equal hashes.
    assert utils.stable_fingerprinter({I: 1, J: 2}) == utils.stable_fingerprinter({J: 2, I: 1})
```

(Alternatively: an `xfail` for now, flipped to passing once the fallback is fixed.)

tests/next_tests/unit_tests/ffront_tests/test_stages.py
```diff
@@ -6,134 +6,42 @@
...
+    first_program = stages.DSLProgramDef(definition=_make_program_definition(1))
+    same_program = stages.DSLProgramDef(definition=_make_program_definition(1))
+    different_program = stages.DSLProgramDef(definition=_make_program_definition(2))
 
-    assert stages.fingerprint_stage(samecode) != stages.fingerprint_stage(past)
-    assert stages.fingerprint_stage(different) != stages.fingerprint_stage(past)
+    assert stages.fingerprinter(first_fieldop) == stages.fingerprinter(same_fieldop)
+    assert stages.fingerprinter(first_fieldop) != stages.fingerprinter(different_fieldop)
+    assert stages.fingerprinter(first_program) == stages.fingerprinter(same_program)
+    assert stages.fingerprinter(first_program) != stages.fingerprinter(different_program)
```
havogt, line 47: **The `debug` flag on `DSL{FieldOperator,Program}Def` is not `pickle=False`, so toggling it must change the fingerprint** — but nothing tests this. If a user runs the same operator first with `debug=True`, then with the default, both calls go through the same `CachedStep` and should land in different cache slots. Add:

```python
def test_definition_stage_debug_flag_affects_fingerprint():
    a = stages.DSLFieldOperatorDef(
        definition=_make_field_operator_definition(1), debug=False)
    b = stages.DSLFieldOperatorDef(
        definition=_make_field_operator_definition(1), debug=True)
    assert stages.fingerprinter(a) != stages.fingerprinter(b)
```

If the intent is the opposite (debug should *not* affect the cache), then the field should be marked `pickle=False` and a test would assert *equal* fingerprints — either way, the current state is ambiguous.

src/gt4py/next/otf/arguments.py
havogt, line 143: **The fingerprint contract for runtime `offset_provider` content isn't pinned.** The `offset_provider` field on the line below still carries runtime `NeighborTable`s (per its `TODO(havogt)`), which means the mixin pickles the full backing `ndarray`s on every cache lookup. Two questions worth asserting:

- Same neighbor-table *content*, different `ndarray` instances → equal fingerprint (correctness of cache hit rate).
- Different content → different fingerprint (correctness of cache invalidation).

Neither direction is currently tested. A single test pins both:

```python
def test_fingerprint_distinguishes_neighbor_table_content():
    arr1 = np.array([[0, 1], [2, 3]], dtype=np.int32)
    arr2 = np.array([[0, 1], [2, 3]], dtype=np.int32)  # equal content, distinct object
    arr3 = np.array([[1, 0], [3, 2]], dtype=np.int32)
    op_a, op_b, op_c = (_make_nb_table(a) for a in (arr1, arr2, arr3))
    cta = lambda op: arguments.CompileTimeArgs(
        args=(), kwargs={}, offset_provider={"E2V": op},
        column_axis=None, argument_descriptor_contexts={})
    assert utils.stable_fingerprinter(cta(op_a)) == utils.stable_fingerprinter(cta(op_b))
    assert utils.stable_fingerprinter(cta(op_a)) != utils.stable_fingerprinter(cta(op_c))
```

This is also the natural anchor for a future perf assertion, if the prior comment about runtime-`OffsetProvider` hashing cost gets addressed.

havogt (commented):
A second pass focused on **test coverage**, after the recent ADR + `BUILD_CACHE_VERSION_ID` work landed. The building blocks are well-tested (the `singledispatcher`/`merge_dispatchers`/`pickle_reducer_factory` suite is impressively thorough, and `test_cached_step_cache_key_includes_step_config` directly pins the cpu↔gpu regression). The gaps below are mostly at the **integration seams** — places where the new mechanism's *behavior* (rather than its parts) is left implicit.

Not blockers; flagging since they're the kinds of properties a future refactor could silently break.

havogt:
**Repro: `RecursionError` in `stable_fingerprinter` at IR depth ~74 (default recursion limit)**

```python
from gt4py.next import utils
from gt4py.next.iterator import ir as itir

# IR shape of `a + a + ... + a` with 80 terms in a field operator
expr = itir.SymRef(id="a")
for _ in range(80):
    expr = itir.FunCall(fun=itir.SymRef(id="plus"), args=[expr, itir.SymRef(id="a")])

utils.stable_fingerprinter(expr)  # RecursionError
```

Measured at this branch head (`1896afbc`, Python 3.10, default `recursionlimit=1000`): the maximum ITIR chain depth before `RecursionError` is **~74** — the pure-Python pickler (`pickle._Pickler`, needed because the C pickler bypasses `reducer_override` for exact `dict`/`set` instances) costs ~13 Python frames per nesting level.

Where it bites: this is exactly what `CachedStep.cache_key` computes for the `run_gtfn_cached` executor (= `gtx.gtfn_cpu`) and for the `cached_translation` `FileCache` — `stable_fingerprinter` over the lowered `CompilableProgram`. The executor key is computed on **every program call, including cache hits**, so a program that was successfully compiled once still fails at lookup time.

End-to-end confirmation: a generated field operator returning an 80-term sum, called with `gtx.gtfn_cpu`, dies with ~985 frames inside `pickle.py`, entered via `cache_key` → `content_hash`, before any compilation starts. (At that size the program would later also hit the pre-existing recursive eve-visitor limit in the gtfn transforms — but that one exists on `main` too, cf. the `sys.setrecursionlimit(100000000)` workaround in `iterator/transforms/trace_shifts.py`. The fingerprinting path adds a much lower ceiling — ~74 vs ~180 levels — in front of it.)

Note that raising the recursion limit is not a safe fix on Python ≤ 3.10 (still supported), where pure-Python recursion consumes C stack and a high limit can turn the `RecursionError` into a segfault.

FWIW: the structural-fingerprinter proposal in egparedes/gt4py#8 traverses iteratively and removes the depth limit entirely (depth-5000 chains fingerprint in ~170 ms).

egparedes:
Closing in favor of #2609

PR #2610, opened 2026-05-27
fix[next-dace]: Avoid referencing subset objects on multiple edges

edopao:
This PR ensures that subset objects are not shared by multiple memlets. This is a potential issue since the subsets of a memlet are mutable objects. An additional check in dace validation has been introduced, as part of the new GPU codegen, which caught this issue.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
+++ b/src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
@@ -422,8 +422,8 @@ class LambdaToDataflow(eve.NodeVisitor):
         dst_node: dace_nodes.Node,
         dst_conn: Optional[str] = None,
     ) -> None:
-        new_subset = copy.deepcopy(src_subset)  # guarantee unique subset objects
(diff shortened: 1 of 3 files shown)


src/gt4py/next/program_processors/runners/dace/lowering/gtir_dataflow.py
```diff
@@ -421,7 +422,8 @@ def _add_input_data_edge(
         dst_node: dace_nodes.Node,
         dst_conn: Optional[str] = None,
     ) -> None:
-        edge = MemletInputEdge(self.state, src, src_subset, dst_node, dst_conn)
+        new_subset = copy.deepcopy(src_subset)  # guarantee unique subset objects
+        edge = MemletInputEdge(self.state, src, new_subset, dst_node, dst_conn)
```
edopao:
```suggestion
        src_subset = copy.deepcopy(src_subset)  # guarantee unique subset objects
        edge = MemletInputEdge(self.state, src, src_subset, dst_node, dst_conn)
```

PR #2612, opened 2026-05-29
feat[next]: Implement embedded as_offset

havogt:


src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -894,6 +894,29 @@ def _concat_where(
...
+        raise ValueError(
+            f"'as_offset' is only supported for Cartesian offsets "
+            f"(target dimension equal to source dimension); "
+            f"got source '{offset_.source.value}' and target ({target_dims})."
+        )
+    source_dim = offset_.source
+    domain = offset_field.domain
+    xp = offset_field.array_ns
+    src_idx = domain.dim_index(source_dim, allow_missing=False)
+    src_range = domain[src_idx].unit_range
```
havogt:
```suggestion
    src_range = domain[source_dim].unit_range
```
does this work?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -894,6 +894,29 @@ def _concat_where(
...
+    shape = tuple(len(src_range) if i == src_idx else 1 for i in range(len(domain)))
+    coords = xp.arange(src_range.start, src_range.stop).reshape(shape)
+    return common._connectivity(
+        xp.asarray(offset_field.ndarray) + coords, codomain=source_dim, domain=domain
```
havogt: Why asarray?

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -508,3 +508,7 @@ def as_connectivity_field(self) -> common.Connectivity:
             self._cache[cache_key] = connectivity
 
         return connectivity
+
+
+def is_cartesian_offset(offset: FieldOffset | ts.OffsetType) -> bool:
+    return len(offset.target) == 1 and offset.source == offset.target[0]
```
havogt: Two defensive gaps in the canonical predicate:

1. **`Dimension.__eq__` ignores `.kind`** (common.py:128-129 compares only `.value`). Trigger: `FieldOffset('X', source=Dimension('K', HORIZONTAL), target=(Dimension('K', VERTICAL),))` is classified as Cartesian and propagates all the way to embedded `_as_offset`, which builds a self-shift along K_horizontal — silently wrong for a cross-kind target.

2. **LOCAL self-loop**: `FieldOffset(source=L, target=(L,))` with `L.kind == LOCAL` passes the check; `_deduce_grid_type` would keep `CARTESIAN`. Pre-existing gap, practically unreachable.

One-line defense-in-depth:
```python
return (
    len(offset.target) == 1
    and offset.source.value == offset.target[0].value
    and offset.source.kind == offset.target[0].kind
    and offset.target[0].kind != common.DimensionKind.LOCAL
)
```

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -894,6 +894,26 @@ def _concat_where(
...
+            f"'as_offset' is only supported for Cartesian offsets "
+            f"(target dimension equal to source dimension); "
+            f"got source '{offset.source.value}' and target ({target_dims})."
+        )
+    source_dim = offset.source
+    domain = offset_field.domain
+    xp = offset_field.array_ns
+    src_range = domain[source_dim].unit_range
+    shape = tuple(len(src_range) if dim == source_dim else 1 for dim in domain.dims)
+    coords = xp.arange(src_range.start, src_range.stop).reshape(shape)
```
havogt: Two things on this construction:

1. **`xp.arange` widens int32 → int64.** `xp.arange(int, int)` returns int64 on 64-bit Linux; if `offset_field.ndarray` is int32 the sum upcasts to int64, silently doubling memory of the resulting connectivity. Fix: `xp.arange(src_range.start, src_range.stop, dtype=offset_field.ndarray.dtype)`.

2. **Reinvents `_identity_index_array`** (this file, line 668). Lines 908–910 collapse into one call:
```python
coords = _identity_index_array(domain, source_dim, xp)
return common._connectivity(offset_field.ndarray + coords, codomain=source_dim, domain=domain)
```
(The helper does an extra `xp.broadcast_to` which is a zero-cost view and doesn't affect the subsequent `+`.)

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -890,6 +890,14 @@ def _visit_as_offset(self, node: foast.Call, **kwargs: Any) -> foast.Call:
         arg_1 = node.args[1].type
         assert isinstance(arg_0, ts.OffsetType)
         assert isinstance(arg_1, ts.FieldType)
+        if not fbuiltins.is_cartesian_offset(arg_0):
+            target_dims = ", ".join(d.value for d in arg_0.target)
+            raise errors.DSLError(
+                node.location,
+                f"'as_offset' is only supported for Cartesian offsets "
```
havogt, line 897: Wording understates the rule: it says `target dimension equal to source dimension` but the predicate also requires `len(target) == 1`. For a 2-target neighbor offset (`V2E`), the user might think renaming target[0] to Edge would fix it — but it's *also* the cardinality.

Suggestion: `(target == (source,))` or `(single target dimension equal to source dimension)`. Same wording at `src/gt4py/next/embedded/nd_array_field.py:902`.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -894,6 +894,26 @@ def _concat_where(
...
+        raise ValueError(
+            f"'as_offset' is only supported for Cartesian offsets "
+            f"(target dimension equal to source dimension); "
+            f"got source '{offset.source.value}' and target ({target_dims})."
+        )
+    source_dim = offset.source
+    domain = offset_field.domain
+    xp = offset_field.array_ns
+    src_range = domain[source_dim].unit_range
+    shape = tuple(len(src_range) if dim == source_dim else 1 for dim in domain.dims)
```
havogt: Same value-only `Dimension.__eq__` issue as the predicate (see comment on `fbuiltins.py:514`): if `domain.dims` ever contained two dims sharing `.value` with different `.kind`, both axes would get `len(src_range)` instead of just the source axis.

Currently shielded by `Domain`'s uniqueness check (`common.py:495`) which uses the same value-only equality — so unreachable today, but coupled to that invariant. If uniqueness is ever relaxed (e.g. to allow same-name dims of different kinds), this shape computation breaks silently.

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -761,6 +762,124 @@ def test_premap_non_contiguous_inverse_image_raises():
...
+        off_arr, domain=common.Domain(dims=(I, J), ranges=(UnitRange(0, 10), UnitRange(0, 3)))
+    )
+
+    result = f.premap(as_offset(Ioff, off))
+
+    assert result.domain == common.Domain(dims=(I, J), ranges=(UnitRange(0, 10), UnitRange(0, 3)))
+    assert np.all(result.ndarray == f.ndarray[np.arange(10)[:, None] + off_arr])
+
+
+def test_as_offset_non_cartesian_offset_raises():
```
havogt, line 875: This test calls `as_offset(...)` directly, hitting only the embedded `ValueError`. The matching DSLError branch in `_visit_as_offset` (type_deduction.py:893-900) has no coverage — a regression that loosens the frontend check (e.g. predicate inversion) wouldn't be caught.

Add a `@gtx.field_operator` test using a non-Cartesian `FieldOffset`, expecting `errors.DSLError` at parse time (e.g. alongside `test_as_offset_dim` / `test_as_offset_dtype` in `tests/next_tests/unit_tests/ffront_tests/test_type_deduction.py`).

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -894,6 +894,26 @@ def _concat_where(
 NdArrayField.register_builtin_func(experimental.concat_where, _concat_where)  # type: ignore[arg-type]
 
 
+def _as_offset(offset: fbuiltins.FieldOffset, offset_field: NdArrayField) -> common.Connectivity:
+    if not fbuiltins.is_cartesian_offset(offset):
+        target_dims = ", ".join(d.value for d in offset.target)
```
havogt, line 902: The full error-message string (lines 899–903) is verbatim-duplicated with `type_deduction.py:894–899`. Unified messaging was intentional, but the contract is currently enforced by hand; a small `fbuiltins._not_cartesian_message(offset) -> str` shared by both raisers would enforce it by construction. Each call site still controls its own exception type and location wrapping. Minor — flag only if you expect future drift.

havogt (commented):
## Self-review notes (via `/code-review` skill, xhigh effort)

9 finder angles → 7 verifiers → sweep. Two candidates were refuted before posting:
- *Missing `source ∈ offset_field.dims` check* — `Domain.__getitem__` already raises a clean `KeyError("No Dimension of type 'X' is present in the Domain.")` (common.py:528-529), so a defensive check would be redundant.
- *Cartesian-only check is a backward-incompatible tightening* — non-Cartesian `as_offset` never compiled on any backend (gtfn/dace/roundtrip all hit downstream asserts; ICON4py only uses `Koff`). The new check just improves diagnostics.

Inline findings below. One finding does not fit any changed line (`_hyperslice` reachability) — recorded here:

### Pre-existing: `inverse_image` raises bare numpy ValueError when offset range is fully disjoint
**`src/gt4py/next/embedded/nd_array_field.py:704`** (in `_hyperslice`, unchanged code). When `offset_field.domain[source_dim]` is fully disjoint from the data field's range (e.g. `f` on `I=[0,10)`, `offset_field` on `I=[100,110)`), `select_mask` is all-False, `xp.nonzero` returns empty, and `xp.min(empty)` raises `ValueError: zero-size array to reduction operation minimum which has no identity`. Pre-existing bug in `_hyperslice`; this PR's new `as_offset` builtin makes it more reachable from user code.

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -890,6 +890,14 @@ def _visit_as_offset(self, node: foast.Call, **kwargs: Any) -> foast.Call:
         arg_1 = node.args[1].type
         assert isinstance(arg_0, ts.OffsetType)
         assert isinstance(arg_1, ts.FieldType)
+        if not fbuiltins.is_cartesian_offset(arg_0):
```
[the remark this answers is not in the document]
havogt: Resolved: added `test_as_offset_non_cartesian` in `tests/next_tests/unit_tests/ffront_tests/test_type_deduction.py` (neighbor + cross-dim cases) asserting `DSLError` at parse time.

src/gt4py/next/ffront/fbuiltins.py
havogt: Resolved: tightened the predicate to also require matching `.kind` and reject `LOCAL` target. Cross-kind / LOCAL-self-loop coverage added in `test_decorator_domain_deduction.py::test_domain_deduction_unstructured`.

src/gt4py/next/embedded/nd_array_field.py
havogt: Resolved: threaded an optional `dtype` through `_identity_index_array` and reused it in `_as_offset` (passing `offset_field.ndarray.dtype`), so the connectivity now preserves the offset field's dtype.

src/gt4py/next/ffront/foast_passes/type_deduction.py
havogt: Resolved: updated wording to `(single target dimension equal to source dimension)` in both raisers (`type_deduction.py` and `nd_array_field.py`).

tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
havogt: Resolved: added `test_as_offset_non_cartesian` in `tests/next_tests/unit_tests/ffront_tests/test_type_deduction.py` next to `test_as_offset_dim` / `test_as_offset_dtype`.

src/gt4py/next/embedded/nd_array_field.py
havogt: Won't fix in this PR. Shielded by `Domain`'s uniqueness invariant (which uses the same value-only equality), so unreachable today. The predicate-level tightening in `fbuiltins.is_cartesian_offset` blocks the construction one layer up. Worth revisiting only if `Domain` uniqueness is ever relaxed.

src/gt4py/next/embedded/nd_array_field.py
havogt: Won't fix in this PR — the hand-unified messages are intentional. Easy follow-up if reviewers later prefer a shared helper, but the contract is only two raisers right now.

havogt:
Re: the pre-existing `_hyperslice` disjoint-range `ValueError` flagged in the self-review body — fix in #2616.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -894,6 +897,26 @@ def _concat_where(
...
+        raise ValueError(
+            f"'as_offset' is only supported for Cartesian offsets "
+            f"(single target dimension equal to source dimension); "
+            f"got source '{offset.source.value}' and target ({target_dims})."
+        )
+    source_dim = offset.source
+    coords = _identity_index_array(
+        offset_field.domain, source_dim, offset_field.array_ns, dtype=offset_field.ndarray.dtype
+    )
```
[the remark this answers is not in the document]
havogt: Good catch. Reverted to using `fbuiltins.IndexType` (= `np.int32`, the codebase's canonical index dtype) for `_identity_index_array`'s `dtype` argument rather than the offset field's dtype. For a narrow offset field (`int8`, `uint8`, etc.) the sum now upcasts to `int32` instead of wrapping; for a wider offset field (`int64`) it upcasts to `int64`. Added `test_as_offset_narrow_offset_dtype_no_wrap` covering an `int8` offset field over a domain of size 200.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -894,6 +897,26 @@ def _concat_where(
 NdArrayField.register_builtin_func(experimental.concat_where, _concat_where)  # type: ignore[arg-type]
 
 
+def _as_offset(offset: fbuiltins.FieldOffset, offset_field: NdArrayField) -> common.Connectivity:
+    if not fbuiltins.is_cartesian_offset(offset):
```
egparedes, line 901: Would it make sense to define a custom protocol / ABC for CartesianOffset fields so it could be used with `TypeGuard` / `TypeIs` checks?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -894,6 +897,26 @@ def _concat_where(
...
+            f"'as_offset' is only supported for Cartesian offsets "
+            f"(single target dimension equal to source dimension); "
+            f"got source '{offset.source.value}' and target ({target_dims})."
+        )
+    source_dim = offset.source
+    coords = _identity_index_array(
+        offset_field.domain, source_dim, offset_field.array_ns, dtype=fbuiltins.IndexType
+    )
+    return common._connectivity(
+        offset_field.ndarray + coords, codomain=source_dim, domain=offset_field.domain
```
egparedes, line 913: What is the dtype of the `offset_field` array (and therefore the dtype of this operation)? Is it checked somewhere? Should the identity index array have the same dtype as the offset_field array?

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -508,3 +508,12 @@ def as_connectivity_field(self) -> common.Connectivity:
             self._cache[cache_key] = connectivity
 
         return connectivity
+
+
+def is_cartesian_offset(offset: FieldOffset | ts.OffsetType) -> bool:
```
egparedes, line 513: As mentioned above, wouldn't it be better to define this function as a TypeGuard? Does it have an unit test?

egparedes (commented):
It looks good overall but I have some questions.

src/gt4py/next/ffront/fbuiltins.py
havogt: Maybe with the changes coming in #2411 and staggering, I would not introduce a separate protocol for cartesian field offsets now.

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -890,6 +890,14 @@ def _visit_as_offset(self, node: foast.Call, **kwargs: Any) -> foast.Call:
...
+            target_dims = ", ".join(d.value for d in arg_0.target)
+            raise errors.DSLError(
+                node.location,
+                f"'as_offset' is only supported for Cartesian offsets "
+                f"(single target dimension equal to source dimension); "
+                f"got source '{arg_0.source.value}' and target ({target_dims}).",
+            )
```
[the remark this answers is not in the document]
havogt: Resolved: error message now uses `str(d)` for source/target rendering (which includes kind, e.g. `K[HORIZONTAL]`) and mentions the non-local requirement. Cross-kind same-name dims and LOCAL self-loops are now disambiguated by the rendered form.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -894,6 +897,26 @@ def _concat_where(
...
 
 
+def _as_offset(offset: fbuiltins.FieldOffset, offset_field: NdArrayField) -> common.Connectivity:
+    if not fbuiltins.is_cartesian_offset(offset):
+        target_dims = ", ".join(d.value for d in offset.target)
+        raise ValueError(
+            f"'as_offset' is only supported for Cartesian offsets "
+            f"(single target dimension equal to source dimension); "
+            f"got source '{offset.source.value}' and target ({target_dims})."
+        )
```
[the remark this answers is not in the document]
havogt: Resolved: matches the frontend message update. Source/target now rendered via `str(d)` (includes kind), wording mentions the non-local requirement.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -894,6 +897,26 @@ def _concat_where(
...
+    source_dim = offset.source
+    coords = _identity_index_array(
+        offset_field.domain, source_dim, offset_field.array_ns, dtype=fbuiltins.IndexType
+    )
+    return common._connectivity(
+        offset_field.ndarray + coords, codomain=source_dim, domain=offset_field.domain
+    )
```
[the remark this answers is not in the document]
havogt: Resolved: `_as_offset` now raises a clear `ValueError("'as_offset' requires an integer offset field; got dtype '...'.")` upfront, before the `_connectivity` assertion can fire. Added `test_as_offset_non_integer_offset_field_raises` covering a float offset field.

src/gt4py/next/embedded/nd_array_field.py
havogt: Offset field dtype is whatever the user provides. Integrality is checked at the frontend via `type_info.is_integral` and now also upfront in embedded `_as_offset` (see Copilot's third thread). The identity index array is built with `fbuiltins.IndexType` (`= np.int32`, the codebase's canonical index dtype) regardless of the offset field's dtype — a narrow `int8` offset can't wrap, a wide `int64` offset upcasts the sum to `int64` and preserves width. Covered by `test_as_offset_narrow_offset_dtype_no_wrap`.

src/gt4py/next/embedded/nd_array_field.py
havogt: Won't fix in this PR. #2411 is moving cartesian shifts off `FieldOffset` entirely onto `CartesianConnectivity` / `itir.CartesianOffset`; adding a `CartesianFieldOffset` protocol here would double down on the deprecated path with a short shelf life. Leaving the thread open for visibility — happy to revisit if it still pays its weight after #2411 lands.

src/gt4py/next/ffront/fbuiltins.py
havogt: Partial: added a direct unit test `test_is_cartesian_offset` in `test_decorator_domain_deduction.py` covering cartesian, multi-target, cross-name, cross-kind, and LOCAL self-loop cases. TypeGuard skipped here — same reasoning as the nd_array_field.py:901 thread (#2411 deprecates the `FieldOffset`-for-cartesian path). Leaving open for the TypeGuard part.

src/gt4py/next/embedded/nd_array_field.py
havogt: Both raisers were updated together in this round and the messages now match verbatim again. Resolving — the drift risk is contained; a shared helper remains a viable follow-up if the contract grows beyond two raisers.

egparedes (approved):
LGTM

PR #2614, opened 2026-05-29
fix[next-dace]: Reject MoveDataflowIntoIfBody when tasklet free symbol clashes with inner data descriptor

edopao:
## Summary

- `MoveDataflowIntoIfBody` could relocate a tasklet whose code references an outer SDFG **symbol** as a free variable (no input connector), even when the nested SDFG already has a **data descriptor** with the same name. After relocation the tasklet would silently read the inner descriptor instead of the outer symbol, producing incorrect results.
- Fix: `_check_for_data_and_symbol_conflicts` now returns `False` when any name in `required_symbols` matches an existing data descriptor in `if_block.sdfg`.
- Regression test `test_if_mover_symbol_clashes_with_inner_data` added to `test_move_dataflow_into_if_body.py`.

## Test plan

- [x] New test `test_if_mover_symbol_clashes_with_inner_data` fails before the fix and passes after.
- [x] All 17 existing tests in `test_move_dataflow_into_if_body.py` continue to pass.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
@@ -646,6 +646,13 @@ class MoveDataflowIntoIfBody(dace_transformation.SingleStateTransformation):
             if conflicting_symbol != str(symbol_mapping[conflicting_symbol]):
                 return False
 
+        # A required symbol must not share its name with a data descriptor already present
+        #  in the nested SDFG. If it did, after relocation the tasklet code would silently
+        #  read the inner data descriptor instead of the outer symbol.
+        inner_arrays = if_block.sdfg.arrays
+        if any(sym in inner_arrays for sym in required_symbols):
+            return False
+
         return True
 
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -646,6 +646,13 @@ def _check_for_data_and_symbol_conflicts(
             if conflicting_symbol != str(symbol_mapping[conflicting_symbol]):
                 return False
 
+        # A required symbol must not share its name with a data descriptor already present
+        #  in the nested SDFG. If it did, after relocation the tasklet code would silently
+        #  read the inner data descriptor instead of the outer symbol.
+        inner_arrays = if_block.sdfg.arrays
+        if any(sym in inner_arrays for sym in required_symbols):
```
philip-paul-mueller: It is probably better to compute the intersection here.

philip-paul-mueller (approved):
LGTM.

Rejecting it is okay, because we do not do symbol renaming.

PR #2615, opened 2026-05-29
refactor[next-dace]: Avoid symbol to scalar conversion in nested sdfg

edopao:


edopao:
cscs-ci run default

edopao:
Not needed, indeed.

PR #2623, opened 2026-06-05
fix[next]: Enable the translation cache in compiled backends

edopao:
Considering a compiled backend (`dace` or `gtfn`, it doesn't matter), the `CachedStep` framework is used in two places:
- to cache (in memory) the executor OTF workflow in program call
- to cache (in memory or on disk, configurable) the translation stage of the executor OTF workflow

For the former, this PR removes the executor cache since a compiled program is already cached in the program pool.
For the latter, this PR enables the translation cache, by default, in both the dace and gtfn backends.

Note that this PR removes `test_offset_provider_cache` since this cache was removed in a previous PR, and the test fails with the new backend behavior. The test was added in #2044, later the offset provider cache was deleted in #2299 .


egparedes (approved):
LGTM

PR #1570, opened 2024-07-03
Refactor tree_map and replace apply_to_primitive_constituents

SF-N:
## Description
- refactors tree_map to be able to handle e.g. tree_map(lambda x: x + 1, ((1, 2), 3)) and adds respective tests
- renames `result_collection_type` to `result_collection_constructor`
- introduces new alias for utils.tree_map(collection_type=ts.TupleType, result_collection_constructor=lambda x: ts.TupleType(types=[*x]))
- replaces apply_to_primitive_constituents by calls to tree_map adds respective tests

## TODOs

- [ ] Handle apply_to_primitive_constituents with `with_path_arg=True`
- [x] Handle apply_to_primitive_constituents with `tuple_constructor`
- [ ] pre-commit still failing:src/gt4py/next/type_system/type_info.py:200:17 and src/gt4py/next/iterator/type_system/type_synthesizer.py:142:16: error: No overload variant of "tree_map" matches argument types "type[TupleType]", "Callable[[Any], TupleType]"  [call-overload]
- [ ] pre-commit still failing: src/gt4py/next/utils.py:87:1: error: Overloaded function implementation does not accept all possible arguments of signature 1  [misc]
- [x] merge main and resolve conflicts, especially check license in new files

PR #2626, opened 2026-06-05
refactor[next]: Replace apply_to_primitive_constituents with tree_map_type

havogt:
Closes #1570

## Description

Removes `type_info.apply_to_primitive_constituents` and migrates all call sites to `tree_map_type`, completing the long-standing TODO in `ffront/type_info.py`.

- The canonical `tree_map_type` / `_tree_map_type_constructor` are moved from `ffront.type_info` down to the low-level `type_system.type_info`, so iterator- and codegen-layer call sites can use them without an upward dependency on `ffront`. All call sites (including the ones in `ffront`) reference `type_system.type_info` directly — there is no re-export.
- Unlike `apply_to_primitive_constituents` (which recursed only on `ts.TupleType` and treated `ts.NamedCollectionType` as a leaf), `tree_map_type` recurses over `COLLECTION_TYPE_SPECS` and preserves named-collection structure. This changes behavior only for `NamedCollectionType` inputs, which are lowered to plain tuples (`named_collections_to_tuple_types`) before any iterator-/codegen-stage call site — so the migrated sites are unaffected.
- `tree_map_type` is overloaded so its result type reflects the constructor's return (`_T | ts.CollectionTypeSpec` for the default, `_T | _C` for a custom `result_collection_constructor`) instead of `tree_map`'s intentionally weak return type, avoiding casts at the call sites.

Pure refactoring, no behavior change at any call site.

## Requirements

- [x] All fixes and/or new features come with corresponding tests. — behavior-preserving refactor, covered by existing suites (type system, iterator type inference, `global_tmps`, gtfn codegen, ffront type deduction).
- [ ] Important design decisions have been documented in the appropriate ADR — N/A, no user-facing design change.

### DIFF
--- a/src/gt4py/next/ffront/type_info.py
+++ b/src/gt4py/next/ffront/type_info.py
@@ -157,7 +157,7 @@ def _tree_map_type_constructor_drop_python_type(
     value: ts.CollectionTypeSpecT,
     elems: Iterable[ts.DataType],
 ) -> ts.CollectionTypeSpecT:
-    result = type_info._tree_map_type_constructor(value, elems)
+    result = type_info.tree_map_type_constructor(value, elems)
     if isinstance(value, ts.NamedCollectionType):
         result = datamodels.evolve(result, original_python_type=ts.ANY_PYTHON_TYPE_NAME)
     return result
--- a/src/gt4py/next/ffront/type_info.py
+++ b/src/gt4py/next/ffront/type_info.py
@@ -7,20 +7,16 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import functools
 import inspect
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from typing import Any, Iterator, Sequence, cast
 
 import gt4py.next.ffront.type_specifications as ts_ffront
 import gt4py.next.type_system.type_specifications as ts
 from gt4py.eve import datamodels
-from gt4py.eve.extended_typing import NestedTuple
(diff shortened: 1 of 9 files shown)


src/gt4py/next/ffront/type_info.py
```diff
@@ -19,26 +19,8 @@
...
-
-# TODO: Replace all occurrences of `apply_to_primitive_constituents` with this function,
-#   which also works with NamedCollections.
-tree_map_type = functools.partial(
-    utils.tree_map,
-    collection_type=ts.COLLECTION_TYPE_SPECS,
-    result_collection_constructor=_tree_map_type_constructor,
-)
+_tree_map_type_constructor = type_info._tree_map_type_constructor
+tree_map_type = type_info.tree_map_type
```
havogt: Let's update the call sites instead of re-exporting.

src/gt4py/next/iterator/transforms/global_tmps.py
```diff
@@ -169,24 +169,24 @@ def _transform_by_pattern(
     if extracted_fields:
         tmp_stmts: list[itir.Stmt] = []
 
+        into_tuple = lambda _, elts: tuple(elts)  # noqa: E731
```
havogt:
```suggestion
        as_tuple = lambda _, elts: tuple(elts)  # noqa: E731
```

tehrengruber (approved):

PR #2627, opened 2026-06-05
fix[next]: canonical order for gather output dims

havogt:
## Description

`return_type_field` (frontend type deduction) and `_gather_output_domain`
(embedded runtime) built the gather output domain in insertion order. When a
connectivity introduces a dimension of a different kind than the codomain it
replaced (e.g. a vertical iteration axis while a local dimension survives),
insertion order is non-canonical: the result is rejected by `check_dims` /
mismatches the out field, and the runtime- and frontend-deduced dimensions can
disagree.

Both now order their dimensions through `common.order_dimensions`, so runtime
and frontend agree on a canonical, valid field domain.

Tests: vertical-axis canonicalization regressions (frontend + embedded), the
raw-premap tests updated to expect canonical order, and a guard that same-dim
connectivities (V2V) keep the source dimension.

## Requirements

### DIFF
--- a/src/gt4py/next/embedded/nd_array_field.py
+++ b/src/gt4py/next/embedded/nd_array_field.py
@@ -593,7 +593,8 @@ def _gather_output_domain(
     field_domain: common.Domain, connectivities: Sequence[common.GatherConnectivity]
 ) -> common.Domain:
     """Output domain of a simultaneous gather: each codomain is replaced by the dimensions of its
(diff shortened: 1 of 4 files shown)


tehrengruber:
As discussed let's also fix https://github.com/GridTools/gt4py/issues/2692 in this PR.

tehrengruber (commented):
From the discussion: In the toolchain we ensure ordering by always doing operations in a way that preserve the order instead of ordering again as done here.

PR #2628, opened 2026-06-07
feat[cartesian]: DaCe optimal for/map schedule

FlorianDeconinck:
## Description

Reworked the based scheduling of the loops in OIR -> TreeIR to simplify the code and introduce an optimal scheduling for parallelism:

 - merge all map/for schedule code into a single `_resolve_loop_schedule`
 - respect syntax for sequential loops
 - default parallelism to maximize local parallelization (`omp for parallel` on CPU and plain kernel for CUDA)

This should be covered by the current unit tests


romanc:
I think the failing gpu-tests are due to a cuda-codgen issue in DaCe. From what I understand, this test case triggers [the codepath here](https://github.com/spcl/dace/blob/a6717b608f497190205d5145b0dd50d1575458c0/dace/codegen/targets/cuda.py#L2777-L2788) which inserts a global grid sync into a merged kernel. The grid sync insertion in successful, but the `__gbar` symbol isn't defined because it's not passed to the arguments of the nested SDFG. Similar, the (recreated) conditions inserted after the grid sync might (and in this case do) use undefined symbols that would need to be passed as arguments of the nested SDFG function. Both doesn't happen and this is why `nvcc` fails.

We could work around by slightly changing the test scenario.

twicki (approved):
Looks generally ok. @romanc and I quickly discussed that at some point the sequential concept that turns into loops should not have an (ignored) scheduleType anymore, but that can wait for a follow-up. Happy to merge this one as is

romanc:
looks like the new test needs to mark test running on gpu to only run on gpu, e.g.

```py
pytest.param("dace:gpu", marks=[pytest.mark.requires_gpu])
```

romanc:
cscs-ci run default

romanc (approved):
> discussed that at some point the sequential concept that turns into loops should not have an (ignored) scheduleType anymore

- Follow-up issue: https://github.com/GridTools/gt4py/issues/2681
- PR to update the vertical loop representation in the treeir: https://github.com/GridTools/gt4py/pull/2688

romanc:
cscs-ci run default

PR #2634, opened 2026-06-08
docs: Update PR review instructions

havogt:
Added guidelines for labeling PRs as 'good to merge'. This is a experiment.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -41,6 +41,6 @@ and never AGENTS.md, so a redirect here would propagate nothing. Keep short.
 
 ## Summary
 
-- If the review didn't find any issues and the PR is small (\<30 lines of code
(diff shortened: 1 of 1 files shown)


egparedes (commented):
The new instructions look good to me but, if I understand this file correctly, it only affects github copilot. I think we should also add similar instructions for other agents (maybe in AGENTS.md or maybe in another markdown file referenced from AGENTS), since they could also interact with the PR on github trough GitHub MCP or directly with API calls if configured properly at the user side.

It might seem a bit overkill, but for example I've already run out of copilot credits this month, and I can't request more reviews, so I guess the new instructions would never affect my PRs until the next billing cycle

havogt:
My idea with this was to have one accepted authority (github copilot) to do the judgement, but indeed the new usage limits will not allow to do this for long, I will soon run out as well...

PR #2636, opened 2026-06-08
feat[cartesian]: Better macro of sequential vertical loop when targeting parallel hardware

FlorianDeconinck:
## Description

In the case of a `with computation(FORWARD)` under a `KJI` backend, the ongoing strategy in `dace` backends is to always match the layout, e.g. 
```
for K
  map J
     map I
```  
but that means that for parallel hardware you end up with a sequential loop of 2D parallel kernel! 

We introduce, right after the layout-optimal flip, another flip to get have
```
map J
   map I
       for K
```
while conserving the optimal
```
map K
   map J
      map I
``` 

This code triggers only for parallel targets, e.g. GPU or threaded CPU.

## Requirements


tests/cartesian_tests/unit_tests/backend_tests/test_dace_backend.py
```diff
@@ -153,3 +174,33 @@ def test_dace_cpu_KJI_loop_structure():
     assert [node.map.params for node in state.nodes() if isinstance(node, nodes.MapEntry)] == [
         ["__j", "__i"]
     ]
+
+    # Case #2 - sequential K is inside map
+    os.environ["OMP_NUM_THREADS"] = "2"
```
romanc: I'd prefer if we had two separate test cases, e.g.

```py
def test_dace_cpu_KJI_loop_structure():
    """
    Testing that non-parallel target (e.g. CPU without threading) will respect layout.
    """
    pass

def test_dace_cpu_KJI_loop_structure_parallel():
    """
    Testing that parallel target (e.g. GPU or CPU with threading) will push the sequential
    loop inside the parallel kernel.
    """
    pass
```

since the whole test is anyway duplicated. That way, if the test fails, one doesn't have to first check which case is actually failing (and if the serial test case fails, the parallel one still runs).

If your intention is to group tests together, we could use a test class to keep the context that those two belong together.

src/gt4py/cartesian/gtc/dace/passes/push_vertical_map_down.py
```diff
@@ -43,6 +43,9 @@ class PushVerticalMapDown(tn.ScheduleNodeVisitor):
             // computation here (2)
     """
 
+    def __init__(self, forscope_only: bool = False) -> None:
```
romanc:
```suggestion
    def __init__(self, *, forscope_only: bool = False) -> None:
```

just to be sure that people don't go and write `PushVerticalMapDown(True)`

romanc (approved):
I think it's awesome how easy this gets with the schedule tree.

tests/cartesian_tests/unit_tests/backend_tests/test_dace_backend.py
FlorianDeconinck: Split them - much better

FlorianDeconinck:
cscs-ci run default

romanc (approved):
Nice, thanks for splitting the tests and adding the context manager. That reads very clean now in my eyes.

FlorianDeconinck:
Merging after a few more tests on the HPC

PR #2640, opened 2026-06-09
fix[next-dace]: Fix filtering of nodes in `all_nodes_recursive` call

edopao:
Fixes a bug introduced in #2621. The previous PR changed the filtering of the nodes returned by `all_nodes_recursive` call. It erroneously assumed that the `predicate` argument would filter the nodes, but it just checks whether to continue searching in nested SDFG levels.


iomaganaris (approved):
LGTM. Thanks for the quick fix!

PR #2642, opened 2026-06-09
refactor[next]: relocate shared fixture module out of ffront_tests

havogt:
`ffront_test_utils.py` lived under `feature_tests/ffront_tests/` but is actually the **shared** backend/mesh/dimension fixture module — it defines `exec_alloc_descriptor` (the backend-matrix fixture), `mesh_descriptor`, the grid descriptors, and the dimension/offset/field aliases, and is imported by **42 files** across unit / integration / regression — including `cases.py` itself.


egparedes (approved):
LGTM

PR #2645, opened 2026-06-10
test[next]: drop two redundant if-statement tests

havogt:


havogt:
I think we can keep them.

PR #2648, opened 2026-06-10
fix[next]: extend fingerprinting to cover all relevant cache state via structural hashing

egparedes:
## Summary

Generalizes and unifies the fingerprinting / cache-key infrastructure used by the `gt4py.next` workflow caches, so cache keys are **deterministic across interpreter runs** and get **invalidated when any relevant state changes** — the cached step's own configuration and the gt4py version, not just the step input.

Supersedes #2609, keeping its caching *policy* (`cache_key = H(H(BUILD_CACHE_VERSION_ID, step), input_fingerprint(input))`, default-include with explicit field opt-out) while replacing the pickle-based *mechanism* with **structural Merkle-style hashing**, addressing the review issues with the old approach (`RecursionError` on realistic IR depths, local-enum regression, non-orderable dict keys, `OrderedDict` false cache hits, identity-sensitivity, conflated pickling contracts). 

The design and rationale are documented in the new ADR [`docs/development/ADRs/next/0023-Fingerprinting.md`](docs/development/ADRs/next/0023-Fingerprinting.md).

## Key changes

- **Structural fingerprinter as a layered catamorphism** (`gt4py.next.fingerprinting`), separating traversal from reduction:
  - *Deconstruction*: per-type deconstructors peel one level into `EmptyDeconstruction` / `Deconstruction` / `OrderInsensitiveDeconstruction`, dispatched on the MRO (`eve.utils.singledispatcher`, composable via `merge_dispatchers`). Defaults cover class-tagged primitives, ordered tuples/lists, digest-sorted dicts/sets (unsortable keys just work), enums by member content, types/functions/modules by qualified name, dataclasses/datamodels by fields honoring `gt4py_metadata(fingerprint=False)`, and a `__reduce_ex__` fallback honoring `copyreg.dispatch_table` (e.g. NumPy ufuncs).
  - *Traversal*: `catabolize`, a generic **iterative** post-order fold with identity-based memoization, cycle back-references (`allow_cycles`) and canonical ordering of order-insensitive pieces — no recursion limit, and fingerprints are a pure function of value, not object identity.
  - *Reduction*: a single `fingerprint_aggregator` (the algebra) hashes deconstructions uniformly with xxh64 and domain separation. A fingerprinter is `catabolize` partially applied to a deconstructor and an aggregator (`make_fingerprinter`).
- **Strict vs. lenient durability**: two interchangeable fingerprinters that agree on any graph without non-importable objects. `strict_fingerprinter` rejects non-importable callables/types/modules so a key stays reproducible across processes; `lenient_fingerprinter` instead falls back to a structural identity (functions by code object + closure cells, dynamic types/modules by unverified qualified name).
- **Cache keys**: `CachedStep` takes an explicit `input_fingerprinter` plus a `step_fingerprinter` (default `lenient_fingerprinter`); both, along with `cache`, are excluded from its own fingerprint. It keys on `step_fingerprinter((step_fingerprint, input_fingerprinter(inp)))`, with `step_fingerprint = step_fingerprinter((BUILD_CACHE_VERSION_ID, self))` computed once per instance. The `in_memory` / `persistent` constructors pair the store with `lenient_fingerprinter` for the step half. New `config.BUILD_CACHE_VERSION_ID` (defaults to the gt4py version, env-overridable via `GT4PY_BUILD_CACHE_VERSION_ID`) salts every key.
- **Per-call-site input fingerprinters**:
  - ffront in-memory steps use `ffront_stages.semantic_fingerprinter` (lenient deconstructor that keeps `SourceLocation`s on AST nodes and hashes DSL definition functions by source + closure variables).
  - the executor's in-memory translation cache uses `fast_compilable_program_fingerprinter` (ITIR `lenient_ir_fingerprinter`, which skips `location`/`type`, plus offset providers by connectivity identity in iteration order — cheap, per-process).
  - the **persistent** translation caches (`gtfn`, `dace`) use `compilable_program_fingerprinter` = `strict_fingerprinter`, so the on-disk program key is reproducible across processes.
- **Fingerprinting no longer interferes with real pickling**: the `MetadataBasedPicklingMixin` / custom-pickler machinery is removed; classes keep standard serialization.

## Testing

- `tests/next_tests/unit_tests` + `tests/eve_tests` green (only pre-existing GPU/CuPy environment parametrizations skip without CuPy), including unit tests for the `catabolize` driver, the deconstructors, and `CachedStep`.
- `tests/next_tests/integration_tests -k roundtrip` green, including `test_constant_closure_vars_with_enums` (local enums in closure variables) and the `cbrt`/`test_roots` ufunc cases.
- The deep-ITIR `RecursionError` repro from the #2609 review passes; cross-process determinism verified.
- `pre-commit` (ruff, mypy, tach) green on all changed files.


src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -123,7 +123,7 @@ class Params:
             translation=factory.LazyAttribute(
                 lambda o: workflow.CachedStep(
                     o.bare_translation,
-                    hash_function=stages.fingerprint_compilable_program,
+                    key_function=utils.stable_fingerprinter,
```
havogt: **[1/10] Persistent translation cache key now includes `location` and `type` node fields**

This `CachedStep` (backed by a persistent `FileCache`) switched from the location/type-skipping `fingerprint_compilable_program` to plain `utils.stable_fingerprinter`, which hashes every ITIR node's `SourceLocation` (absolute file path, line, column) and lazily-inferred `type` field. Same issue in `dace/workflow/factory.py:50`.

Verified empirically: `stable_fingerprinter(SymRef(id='x', location=SourceLocation(1, 1, '/path/a.py')))` differs from the same node with a different path, while the old key was equal.

**Impact:** with `GT4PY_BUILD_CACHE_LIFETIME=persistent` (e.g. icon4py), inserting a blank line above a stencil, moving the checkout, or sharing the cache dir across paths changes every key → guaranteed miss, full C++/SDFG recompilation, unbounded cache growth. The `skipping_fields_node_fingerprinter('location', 'type')` deconstructors already exist (via `return_deconstructors=True`) and could be composed into this key function.

src/gt4py/next/ffront/stages.py
```diff
@@ -22,24 +22,58 @@
...
+        source_utils.make_source_definition_from_function(func),
+        source_utils.get_closure_vars_from_function(func),
+        state=b"definition_function",
+    )
+
+
+#: Fingerprinter for the frontend stages: skips source locations on AST nodes
+#: and fingerprints DSL definition functions by their source code and closure
+#: variables (instead of by qualified name).
+semantic_fingerprinter: utils.Fingerprinter = functools.partial(
```
havogt: **[2/10] FOAST/PAST cache keys became location-insensitive → cached lowering shared across files with wrong `SourceLocation`s**

`semantic_fingerprinter` skips `location` on all FOAST/PAST nodes. This inverts the invariant the deleted `test_stages.py` tests asserted (`test_fingerprint_stage_foast_op_def` / `test_fingerprint_stage_past_def`: same code at different locations must fingerprint *differently*).

Verified empirically: two field operators with byte-identical source in `mod_one.py` and `mod_two.py` produce equal FOAST-stage fingerprints. The process-wide `foast_to_gtir`/`past_to_itir` `CachedStep`s then return the first operator's cached lowering — with `mod_one.py` `SourceLocation`s baked in by `PreserveLocationVisitor` — for the second operator, so its error messages, warnings, and dace debug info point at the wrong file/line.

havogt (commented):
Automated code review (Claude, high effort: 7 finder angles, per-finding adversarial verification with reproduction scripts). 10 findings, ranked most severe first as inline comments.

The two most consequential findings pull in opposite directions and suggest the location-handling changed by accident, not design: the ffront-stage cache keys became location-**insensitive** (cross-file cache aliasing) while the ITIR-level persistent cache keys became location-**sensitive** (permanent cache misses on unrelated edits). upstream/main had exactly the opposite arrangement in both places.

src/gt4py/next/utils.py
```diff
@@ -29,113 +34,537 @@
...
+    if not module or not qualname:
+        return False
+    target: Any = sys.modules.get(module, None)
+    for part in qualname.split("."):
+        if (target := getattr(target, part, None)) is None:
+            return False
+    return target is obj
+
+
+def _reference_by_fully_qualified_name(obj: Any) -> str:
```
havogt: **[3/10] Regression: locally-defined `FrozenNamespace` subclass instances in closure vars now crash at decoration**

The old fingerprint path (`add_content_to_fingerprint`) had a never-failing `str(obj)` fallback. Now `_reference_by_fully_qualified_name` raises `TypeError` for locally-defined classes reachable via `__reduce_ex__`.

Verified empirically against both venvs: a field operator whose closure vars hold a locally-defined `eve.utils.FrozenNamespace` subclass instance — a frontend-valid `ConstantPythonNamespaceObject` (`type_translation.py:262`) and the icon4py constants idiom — crashes at `@field_operator` decoration with `TypeError: Objects which are not importable under their qualified name ('__main__.....<locals>.Consts') ... cannot be safely referenced`, while upstream/main decorates and compiles fine. Module-level namespaces, enums, and scalars are unaffected — narrow, but a hard crash before compilation.

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -167,12 +167,12 @@ class Params:
...
         cached = factory.Trait(
             executor=factory.LazyAttribute(
-                lambda o: workflow.CachedStep(o.otf_workflow, hash_function=o.hash_function)
+                lambda o: workflow.CachedStep(o.otf_workflow, key_function=o.key_function)
             ),
             name_cached="_cached",
         )
         device_type = core_defs.DeviceType.CPU
-        hash_function = stages.compilation_hash
+        key_function = utils.stable_fingerprinter
```
havogt: **[4/10] Executor cache key content-hashes full neighbor tables (D2H copy for CuPy)**

The replaced `stages.compilation_hash` hashed `offset_provider` by identity via `common.hash_offset_provider_items_by_id` (~1.2 µs, O(1) in mesh size). `stable_fingerprinter` has no ndarray deconstructor, so connectivity arrays inside `CompileTimeArgs.offset_provider` fall through to `__reduce_ex__(2)` = full data bytes: measured ~2.9 ms per 1M-entry int32 table (~2400× slower). For CuPy arrays `__reduce__` calls `.get()`, i.e. a synchronous device-to-host copy. Same in `dace/workflow/backend.py:52`.

**Impact:** once per program variant on the main `CompiledProgramsPool` path (compile/startup cost), but per-invocation on the `iterator/runtime.py:92` `fendef` dispatcher and `FieldOperatorFromFoast.__call__` (`decorator.py:715`) paths — every call pays a multi-MB mesh re-hash plus a D2H transfer on GPU.

src/gt4py/next/config.py
```diff
@@ -82,6 +82,18 @@ def env_flag_to_int(name: str, default: int) -> int:
...
 
+def _get_build_cache_version_id() -> str:
+    return __import__("gt4py").__version__
+
+
+#: Version ID for the build cache. It should only be overridden by gt4py
+#: developers or advanced users when making changes requiring forced
+#: compatibility with previously cached builds.
+BUILD_CACHE_VERSION_ID: str = (
+    os.environ.get("BUILD_CACHE_VERSION_ID") or _get_build_cache_version_id()
```
havogt: **[5/10] Env var missing the `GT4PY_` prefix**

Every other env var in this file uses the `GT4PY_` prefix (`GT4PY_BUILD_CACHE_DIR`, `GT4PY_BUILD_CACHE_LIFETIME`, `GT4PY_BUILD_JOBS`, ...); this one reads bare `BUILD_CACHE_VERSION_ID`.

A user following the convention sets `GT4PY_BUILD_CACHE_VERSION_ID` and is silently ignored; conversely an unrelated CI/HPC tool exporting the generic name `BUILD_CACHE_VERSION_ID` silently pins or perturbs every gt4py cache key — including reusing stale binaries across gt4py upgrades, exactly what this salt exists to prevent. (The ADR documents the unprefixed name too and would need updating.)

src/gt4py/eve/datamodels/core.py
```diff
@@ -100,6 +101,17 @@ def __init__(self, *args: Any, **kwargs: Any) -> None: ...
...
 
 
+class DataModelABC(abc.ABC):
+    """ABC for data models."""
+
+    __datamodel_fields__: ClassVar[utils.FrozenNamespace[Attribute]]
+    __datamodel_params__: ClassVar[utils.FrozenNamespace[Any]]
+
+    @classmethod
+    def __subclasshook__(cls, subclass: type) -> bool:
```
havogt, line 111: **[6/10] `__subclasshook__` returns `False` instead of `NotImplemented`, breaking the ABC hook contract**

The hook ignores `cls` and never returns `NotImplemented`. Verified empirically:
- for `class MyBase(DataModelABC)`, `issubclass(AnyUnrelatedDatamodel, MyBase)` returns `True` (hook inherited, `cls` ignored);
- a genuine direct subclass `class X(DataModelABC)` fails `issubclass(X, DataModelABC)` because returning `False` suppresses the normal MRO fallback that `NotImplemented` would allow.

Correct pattern:
```python
if cls is DataModelABC:
    return is_datamodel(subclass)
return NotImplemented
```
(Note: `DataModelABC` currently has no production callers — see the dead-infrastructure comment on `eve/utils.py`.)

src/gt4py/next/utils.py
```diff
@@ -29,113 +34,537 @@
...
+        type(obj), struct.pack(">dd", obj.real, obj.imag)
+    ),
+    str: lambda obj: EmptyDeconstruction.from_typed_value(
+        type(obj), obj.encode("utf-8", "surrogatepass")
+    ),
+    bytes: lambda obj: EmptyDeconstruction.from_typed_value(type(obj), bytes(obj)),
+    bytearray: lambda obj: EmptyDeconstruction.from_typed_value(type(obj), bytes(obj)),
+    tuple: lambda obj: Deconstruction.from_pieces(*obj, state=_class_obj_tag(obj)),
+    list: lambda obj: Deconstruction.from_pieces(*obj, state=_class_obj_tag(obj)),
+    dict: lambda obj: Deconstruction.from_pieces(
```
havogt: **[7/10] `dict`/`list` subclass instance state silently dropped by MRO dispatch**

`singledispatch` walks the MRO, so this `dict` entry also captures all dict subclasses, hashing only the class tag + `items()` and bypassing the `__reduce_ex__` fallback that would capture extra instance state.

Verified empirically: two instances of a dict subclass with equal items but different extra-attribute values produce identical fingerprints (same for list subclasses). Reachable via `closure_vars: dict[str, Any]` in ffront stages or a user-supplied `offset_provider` mapping → false cache hit returns an artifact compiled for the other object.

src/gt4py/next/otf/workflow.py
```diff
@@ -232,44 +228,57 @@ class CachedStep(
...
-    1
+    A different argument is computed and cached separately:
+    >>> cached_step([4, 5]) is result
+    False
     """
 
     step: Workflow[StartT, EndT]
-    hash_function: Callable[[StartT], HashT] = dataclasses.field(default=hash)  # type: ignore[assignment]
-    cache: OpaqueMutableMapping[HashT, EndT] = dataclasses.field(repr=False, default_factory=dict)
+    key_function: Callable[[StartT], HashT] = dataclasses.field(
```
havogt: **[8/10] Breaking `CachedStep` API change with no deprecation path**

`hash_function` (`default=hash`) was renamed to `key_function` and is now required, and `cache_key` fingerprints `self`, so steps containing lambdas/closures raise at first call (a test removed by this PR explicitly used `step=lambda inp: [*inp, 1]`).

Verified empirically:
- `CachedStep(step=tuple)` → `TypeError: missing ... 'key_function'`
- `CachedStep(step=tuple, hash_function=str)` → unexpected keyword
- `CachedStep(step=lambda inp: [*inp, 1], key_function=str)([1])` → `TypeError: not importable under their qualified name`

The new docstring and `WorkflowPatterns.md` document the restriction, so this is clearly intentional — but downstream code (icon4py-style custom workflows, the pre-PR `WorkflowPatterns.md` recipe) breaks at construction or first call with no deprecation period.

src/gt4py/eve/utils.py
```diff
@@ -630,81 +641,125 @@ def is_noninstantiable(cls: Type[_T]) -> bool:
...
+            )
+
+    assert callable(default)  # for mypy
+    result = functools.singledispatch(default)
+    for cls, func in implementations.items():
+        result.register(cls)(func)
+    return cast(xtyping.SingleDispatchCallable[P, T], result)
+
+
+def merge_dispatchers(
```
havogt, line 710: **[9/10] Dead-on-arrival eve infrastructure (3 items)**

Verified by grep — each is referenced only by its own new unit tests, with zero production callers:
- `merge_dispatchers` (here, ~30 lines + 3 tests) — `ffront/stages.py:59` composes deconstructors via plain dict merging into `make_deconstructor({**foast.semantic_fingerprint_deconstructors, ...})` instead;
- `DataModelABC` (`datamodels/core.py:104`) — `next/utils.py` deliberately avoids it with an explicit comment that virtual-ABC dispatch breaks `singledispatch`, and calls `datamodels.is_datamodel` directly;
- `@runtime_checkable` on `SingleDispatchCallable` (`extended_typing.py:228`) — no `isinstance`/`issubclass` check against the protocol exists anywhere; the pre-existing `is_single_dispatch_callable` typeguard uses getattr duck-typing.

Same theme, smaller: the empty `BaseStage` marker class (`ffront/stages.py:37`) and the `fingerprinter = semantic_fingerprinter` alias (`stages.py:72`) are referenced nowhere / tests-only.

src/gt4py/next/utils.py
```diff
@@ -29,113 +34,537 @@
...
+}
+
+
+# Pickle protocol used when deconstructing unknown objects via
+# `__reduce_ex__`. Pinned so the deconstruction (and therefore the
+# fingerprint) does not depend on the running Python's default protocol.
+_DECONSTRUCT_PICKLE_REDUCE_PROTOCOL: Final[int] = 2
+
+
+def _dataclass_fields(cls: type) -> Optional[tuple[Any, ...]]:
```
havogt: **[10/10] Avoidable per-node costs on the fingerprinting path**

Measured 8.2 ms per `stable_fingerprinter` call on a 2404-node ITIR tree (~3.4 µs/node), re-paid several times across the chained `CachedStep`s per program even on warm-disk-cache startup — noticeable for sessions compiling many programs (icon4py). Profile attribution:
- `_dataclass_fields` recomputes `dataclasses.fields()` / `datamodels.get_fields()` per *object* instead of caching per *type*, and `fingerprint_fallback` re-filters field metadata per instance (~27% — the old code cached this with `@functools.cache` on `_get_metadata_based_state_getstate`);
- digests round-trip through `hexdigest()` strings re-encoded with `.encode('ascii')` per edge (~13%; raw `digest()` bytes sort fine for the order-insensitive case, only the root needs hex);
- `catabolize` uses `dataclasses.replace` per internal node (~8%; direct construction skips the field-introspection machinery);
- `CachedStep.cache_key` re-fingerprints the constant frozen `self` on every call instead of once per instance (`catabolize`'s memo dict is call-local).

egparedes:
Thanks for the thorough automated pass — it caught a real location-handling inversion I'd half-mirrored. Addressed in two commits (`48027da` and `c7f9b1d`). Disposition below.

### Fixed

- **[1/10] persistent ITIR key includes `location`/`type`** — the gtfn/dace translation `FileCache` now keys on a restored `stages.fingerprint_compilable_program` built on `itir.semantic_fingerprinter` (skips `location`/`type`), so unrelated edits / checkout moves no longer bust the persistent cache.
- **[2/10] ffront keys location-*insensitive* → cross-file aliasing** — confirmed against `main`'s deleted `test_stages.py` (`samecode_fieldop != fieldop`). FOAST/PAST node fingerprints now **include** `location` again, and the DSL definition is fingerprinted by its full `SourceDefinition`. Location-agnostic fingerprinting stays where it belongs (ITIR persistent cache + the `__val_`/`__cond_` var-name generation, which use `itir.semantic_fingerprinter`). Added a `test_fingerprinter_is_location_sensitive` regression test.
- **[3/10] local `FrozenNamespace` subclass crashes at decoration** — registered a content-based deconstructor for `eve.utils.Namespace`, so constant namespaces in closure vars (the icon4py idiom) are fingerprinted by their items instead of falling through to `__reduce_ex__` and referencing the non-importable local class.
- **[4/10] executor content-hashes neighbor tables** — restored `stages.compilation_hash` for the in-memory executor key: location-agnostic program fingerprint + offset providers by identity/order via `hash_offset_provider_items_by_id` (O(1), no D2H copy).
- **[5/10] env var prefix** — now `GT4PY_BUILD_CACHE_VERSION_ID`.
- **[6/10] `__subclasshook__`** — now `if cls is DataModelABC and is_datamodel(subclass): return True; return NotImplemented` (genuine subclasses and derived-ABC checks defer to the MRO).
- **[7/10] container-subclass instance state dropped** — `dict`/`list`/`set`/… subclasses now contribute their instance `__dict__` to the fingerprint; exact builtins are unchanged.
- **[10/10] per-node overhead** — applied the cheap, safe wins: `_dataclass_fields` is now cached per type, and `catabolize`'s COMBINE constructs the recombined `Deconstruction` directly instead of via `dataclasses.replace`. (Left the hex/`encode('ascii')` digest round-trip alone for now — it changes the collapser format and the order-insensitive sort, so it felt out of scope for a fix-up.)

Each fix has a unit test; all affected suites are green, mypy/pre-commit clean, and an end-to-end `run_gtfn_cached` compile + cache-hit checks out.

### Left as-is (flagging for your call)

- **[8/10] breaking `CachedStep` API** — this is the intent of the PR (fingerprinting the step *is* the new "step state in caches" behavior), so I didn't add a deprecation shim. I did fix the doc recipe and cache the per-instance step fingerprint so it isn't re-walked on every call. Happy to add a transitional `hash_function` alias if you want a deprecation window.
- **[9/10] dead infrastructure** — `merge_dispatchers` was deliberately restored earlier in the branch, and `DataModelABC` is new public API (now with a correct hook), so removing them felt like your decision rather than a review fix-up. Easy to drop them + `@runtime_checkable`/`BaseStage`/the `fingerprinter` alias if you'd prefer.

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -114,6 +122,11 @@ class NdArrayField(
 
     array_ns: ClassVar[ModuleType]  # TODO(havogt): introduce a NDArrayNamespace protocol
 
+    def __getstate__(self) -> dict[str, Any]:
+        # Serialize only the dataclass fields, excluding cached properties
+        # stored in `__dict__` (which may not be picklable).
+        return {f.name: getattr(self, f.name) for f in dataclasses.fields(self)}
+
```
havogt, line 129: This was introduced as a fix by claude. Is it needed @egparedes

src/gt4py/next/utils.py
```diff
@@ -8,17 +8,22 @@
 
 from __future__ import annotations
 
+import collections
```
havogt: Maybe the fingerprinting related utils deserve their own module?

src/gt4py/next/program_processors/runners/dace/workflow/factory.py
```diff
@@ -47,7 +47,7 @@ class Params:
             translation=factory.LazyAttribute(
                 lambda o: workflow.CachedStep(
                     o.bare_translation,
-                    hash_function=stages.fingerprint_compilable_program,
+                    key_function=stages.fingerprint_compilable_program,
```
edopao: The current `FileCache` key does not include the stage fingerprint. I think it should be added, question is whether to do it now or in a separate PR.

edopao: Another comment about this aspect:

> [1/10] persistent ITIR key includes location/type — the gtfn/dace translation FileCache now keys on a restored stages.fingerprint_compilable_program built on itir.semantic_fingerprinter (skips location/type), so unrelated edits / checkout moves no longer bust the persistent cache.

In my opinion, the `FileCache` key should include the `type`: the translated program should not clash in disk cache for different types.

docs/development/ADRs/next/0023-Fingerprinting.md
```diff
@@ -0,0 +1,349 @@
...
+## Context
+
+A *fingerprint* is a stable, content-based identifier of an object: two objects
+with the same semantically relevant content must produce the same fingerprint,
+and the value must be reproducible across interpreter restarts (so it can key a
+persistent build cache). GT4Py needs fingerprints in several places:
+
+- The OTF build cache (`CachedStep`) keys compiled artifacts by a fingerprint
+  of the workflow step **and** its input, so a cached compiled program can be
+  reused only when both are unchanged.
```
edopao, line 34: This text does not match the key of the disk cache in translation stage (see my other comments).

src/gt4py/next/embedded/nd_array_field.py
egparedes: Not anymore, thanks.

egparedes: I've changed my mind: even if it's not strictly related to the fingerprinting, it is still needed for pickling the object properly. I'll leave it here.

src/gt4py/next/program_processors/runners/dace/workflow/factory.py
havogt: What about this open question?

src/gt4py/next/fingerprinting.py
```diff
@@ -0,0 +1,714 @@
...
+    Per-type deconstructors are dispatched on the object's MRO; `overrides`
+    take precedence over the default rules. Objects without a matching
+    deconstructor are handled by `fallback` (by default deconstructed through
+    their fields for dataclasses and datamodels, or via the standard
+    ``__reduce_ex__`` protocol).
+    """
+
+    return eve_utils.singledispatcher(
+        fallback,
+        implementations={**_BUILTIN_DECONSTRUCTORS, **(overrides or {})},
```
havogt: **The `fallback=lenient_*` composition doesn't actually make the composed fingerprinters lenient for functions/modules/dynamic types.**

`make_deconstructor` builds `singledispatcher(fallback, implementations={**_BUILTIN_DECONSTRUCTORS, **overrides})`. The `fallback` is only the singledispatch *default* — consulted only for types absent from the registry. But `_BUILTIN_DECONSTRUCTORS` registers the **strict** `from_reference` for exactly `type` / `FunctionType` / `BuiltinFunctionType` / `ModuleType`. So when `skipping_fields_node_deconstructor(..., fallback=lenient_fingerprint_deconstructor)` (and the ffront `semantic_fingerprinter`) pass the lenient deconstructor as `fallback`, it is **shadowed** for precisely the types lenience was meant to cover — the lenient overrides inside `lenient_fingerprint_deconstructor` never get a chance for those types.

Result: the composed `itir.semantic_fingerprinter` / `ffront.semantic_fingerprinter` are *not* lenient about non-importable callables/modules, even though they were wired to be:

```python
>>> lenient_fingerprinter({"f": lambda: 1})        # OK -> b7a105b3003316b7
>>> itir.semantic_fingerprinter({"f": lambda: 1})  # TypeError: Objects which are not importable
>>> itir.semantic_fingerprinter({"m": types.ModuleType("dyn")})  # TypeError
```

`compilable_program_fingerprinter` (the persistent gtfn/dace translation input key) *is* `itir.semantic_fingerprinter`, and ADR 0023 states this lenient path tolerates "Mock-backed value / dynamically-created connectivity type" — but it raises on them instead.

**Severity: latent, not a blocker.** In normal use no lambda/dynamic-module/non-importable-type is reachable inside an ITIR program or a real `offset_provider`, so this doesn't bite production programs today — it's mostly Mock/test territory. But (a) it's a hard crash where the design promises graceful degradation, and (b) the ADR's tolerance claim is currently false at these call sites. The `lenient_fingerprinter` used *directly* (e.g. `CachedStep.in_memory`) is fine — overrides win there; only the `fallback=lenient_*` composition is affected.

**Fix:** compose the lenient per-type handlers into the `overrides` of `make_deconstructor` (so they outrank `_BUILTIN_DECONSTRUCTORS`) rather than passing `lenient_fingerprint_deconstructor` as `fallback` — or have `skipping_fields_node_deconstructor` take the lenient overrides and merge them in. Either way the lenient `FunctionType`/`type`/`ModuleType` rules need to land in the registry, not behind it.

_(Automated review, verified by probe against `1c898f6f`. Separately: my earlier draft finding about the `type`-skip dropping `Literal.type` does **not** hold — those types are recoverable from the keyed structure+args, so the skip is safe; disregard if you saw it.)_

src/gt4py/next/program_processors/runners/dace/workflow/factory.py
egparedes: I think the original point is outdated or wrong: AFAIK the cache contains the fingerprint of both the input stage and the state of the cached workflow.

egparedes: The second question about ignoring the type of ITIR nodes is reasonable. I think ignoring this type is needed in several stages, but it makes sense to make sure that for steps dealing with the final version of the IR (without deferred types) the type is also fingerprinted. I'll take a look

edopao: I agree, I will resolve this conversation.

edopao: Just maybe a clarification. In which case would the front-end generate an IR with different `type` node attributes? Is that even expected? Otherwise we should not exclude it from the fingerprint.

src/gt4py/next/fingerprinting.py
egparedes: Good point. I've tried to address it in the last commit.

src/gt4py/next/program_processors/runners/dace/workflow/factory.py
egparedes: You need to ask @tehrengruber for more details, but IIRC the problem is that the fingerprinter is also used internally in some passes where the nodes have not been fully typed yet, so some nodes have deferred type and later change to the concrete type, but for semantic comparisons is still the same node. The solution might be to define two different fingerprints for the two use cases, which is now trivial to do, but if it's not breaking anything so far, I'd postpone it as a follow-up after merging this PR.

src/gt4py/next/otf/stages.py
```diff
@@ -10,48 +10,48 @@
...
-            # As the frontend types contain lists they are not hashable. As a workaround we just
-            # use content_hash here.
-            utils.content_hash(tuple(arg for arg in program_def.args.args)),
-            common.hash_offset_provider_items_by_id(offset_provider) if offset_provider else None,
-            program_def.args.column_axis,
-        )
-    )
+if TYPE_CHECKING:
+    # Imported only for typing to avoid the import cycle with `otf.definitions`,
+    # which imports this module.
```
havogt, line 23:
```suggestion
```

havogt (approved):
lgtm, let's see how it will do in the real world! (don't fix the comment if you don't have to re-run anyway)

Please wait with merge until 1.1.11 is released.

src/gt4py/next/iterator/ir.py
```diff
@@ -23,9 +23,9 @@
 # Generate an unique fingerprint for `eve.Node`s ignoring the "location" and "type" attribute.
 # TODO(tehrengruber): this is a workaround for the fact that `eve.Node`s type can be
 # set after creation, in the type inference passes.
-semantic_fingerprinter: fingerprinting.Fingerprinter = fingerprinting.make_fingerprinter(
-    deconstructor=fingerprinting.skipping_fields_node_deconstructor(
-        "location", "type", fallback=fingerprinting.lenient_fingerprint_deconstructor
+lenient_ir_fingerprinter: fingerprinting.Fingerprinter = fingerprinting.make_fingerprinter(
```
edopao, line 26: We discussed yesterday whether it would be possible or not to use strict fingerprinter on the IR. I see that you are using the lenient fingerprinter. Didn't the other work?
```
semantic_ir_fingerprinter = fingerprinting.Fingerprinter = fingerprinting.make_fingerprinter(
    deconstructor=fingerprinting.make_strict_data_deconstructor(
        {concepts.Node: fingerprinting.skipping_fields_node_deconstructor("location", "type")}
```

edopao: Besides, I preferred `semantic_fingerprinter` or `semantic_ir_fingerprinter`, but I do not have a big opinion.

egparedes: Actually, the caches are using the strict_fingerprinter for the IR (check current state of otf/stages.py). This lenient fingerprinter is used only in internal IR passes and in the fast approximate hash. I've changed the name to make this explicit and also because I thought `semantic` is misleading here since it's actually ignoring the types too, but it's fine for me to change it back.

edopao: Thanks, I understand now. It is easy to get confused and think that this fingerprinter is the one used for the input IR in the translation stage.
OK, in this case I agree with keeping the name `lenient_ir_fingerprinter`.

edopao (approved):
LGTM

PR #2649, opened 2026-06-10
feat[cartesian]: Make `dace` gpu backend unit-aligned

FlorianDeconinck:
## Description

Previous decision was to make the GPU backend 32-bit aligned to help with auto-vectorization. With more modern hardware it's not as crucial _and_ it gets in the way of mapping to Fortran-style memory allocation.

We propose to make `dace:gpu` and `dace:gpu_IJK` unit aligned to allow memory mapping but also have a small reduction on memory impact.

Current tests should cover this.


romanc (approved):
Fine by me.

twicki (approved):
All good from my side as well!

PR #2655, opened 2026-06-12
feat[next]: Friendly error messages

havogt:
Makes gt4py.next DSL errors user-friendly: structured diagnostics that point at the offending source and suggest a fix, instead of plain message strings.

- Extends DSLError in place with optional structured fields (label, related spans, notes, hints, per-subclass code).
- A single renderer owns the on-screen shape: line-number gutter, carets, same-line related-span merge, multi-line capping, note/hint wrapping, and a graceful fallback when source text is unavailable.
- Adds a catalogue of unsupported Python constructs, mapping each to a user-facing name and an actionable hint, with a fallback for uncatalogued nodes.
- Upgrades the highest-frequency frontend errors: undefined symbol with "did you mean", boolean and mixed-precision operand mismatches, and and/or on fields.
- Attaches toolchain context ("While processing the definition of ...") via PEP 678 add_note.
- Adds an authoring guide at docs/development/next/error-messages.md, linked from CODING_GUIDELINES.md and src/gt4py/next/AGENTS.md.
- Regression tests pin the rendered diagnostic text.

### DIFF
--- a/src/gt4py/next/errors/exceptions.py
+++ b/src/gt4py/next/errors/exceptions.py
@@ -20,6 +20,7 @@ in that submodule as opposed to being in this file.
 from __future__ import annotations
 
 import difflib
+import sys
 from typing import Any, ClassVar, Iterable, Optional, Sequence
 
 from gt4py.eve import SourceLocation
@@ -94,26 +95,18 @@ class DSLError(GT4PyError):
         self.location = location
         return self
 
-    # TODO(havogt): on Python >=3.11 this shadows 'BaseException.add_note' (PEP 678);
-    #  on 3.10 it is a plain new method, so 'add_note' on GT4Py exceptions that are
-    #  not 'DSLError's raises 'AttributeError' on 3.10 but silently goes to
-    #  '__notes__' on 3.11+. Remove this caveat once the Python floor is >=3.12.
-    def add_note(self, note: str) -> None:
-        """
-        Add a note to the diagnostic, using the standard 'BaseException.add_note' API.
+    # TODO(havogt): drop this shim and the matching '__notes__' fold-in in
+    #  '__str__' once the Python floor is >=3.11, where 'BaseException.add_note'
+    #  (PEP 678) and its automatic '__notes__' traceback rendering are built in.
+    if sys.version_info < (3, 11):
 
-        Toolchain stages use this to attach context to an in-flight error
-        ("While processing ...") without touching the message.
-
-        The note is routed into the structured 'notes' field instead of
-        '__notes__': the traceback machinery (and IPython/pytest) renders this
-        exception through 'str()', which already includes the structured notes,
-        so storing them in '__notes__' as well would print them twice.
-        """
-        self.notes.append(note)
+        def add_note(self, note: str) -> None:
+            if not hasattr(self, "__notes__"):
+                self.__notes__ = []
+            self.__notes__.append(note)
 
     def __str__(self) -> str:
-        return formatting.format_diagnostic_parts(
+        body = formatting.format_diagnostic_parts(
             self.message,
             self.location,
             label=self.label,
@@ -121,6 +114,12 @@ class DSLError(GT4PyError):
             notes=self.notes,
             hints=self.hints,
         )
+        if sys.version_info < (3, 11):
+            # On 3.10 the traceback machinery doesn't render '__notes__'; fold
+            # them in so they surface through 'str()' (pytest, IPython, logging)
+            # the way the >=3.11 machinery does automatically.
+            body += "".join(f"\n{note}" for note in getattr(self, "__notes__", []))
+        return body
 
 
 class UnsupportedPythonFeatureError(DSLError):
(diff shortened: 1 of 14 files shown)


edopao:
LGTM. Very nice, I don't have any comment.

edopao (approved):

src/gt4py/next/errors/exceptions.py
```diff
@@ -32,37 +35,131 @@ def message(self) -> str:
...
+
+        Toolchain stages use this to attach context to an in-flight error
+        ("While processing ...") without touching the message.
+
+        The note is routed into the structured 'notes' field instead of
+        '__notes__': the traceback machinery (and IPython/pytest) renders this
+        exception through 'str()', which already includes the structured notes,
+        so storing them in '__notes__' as well would print them twice.
+        """
+        self.notes.append(note)
```
egparedes: Question: shouldn't this method be added only for python < 3.11? Example:
```python
if sys.version_info < (3, 11):
    ....
```

src/gt4py/next/errors/exceptions.py
```diff
@@ -32,37 +35,131 @@ def message(self) -> str:
...
+            self.location,
+            label=self.label,
+            related=self.related,
+            notes=self.notes,
+            hints=self.hints,
+        )
 
 
 class UnsupportedPythonFeatureError(DSLError):
+    code = "unsupported-syntax"
```
egparedes: Question: shouldn't `code` be declared with the `Final` type hint in all DSLError subclasses?

src/gt4py/next/errors/formatting.py
```diff
@@ -14,53 +14,208 @@
...
 if TYPE_CHECKING:
     from gt4py.next.errors import exceptions
 
+#: Maximum number of source lines rendered for a multi-line location.
+_MAX_SNIPPET_LINES = 3
 
-def get_source_from_location(location: SourceLocation) -> str:
+
+#: An underline row below the first snippet line: (column, end_column, label).
+_UnderlineRow = tuple[int, Optional[int], str]
```
egparedes: Isn't the formatting wrong here?
```suggestion
_underline_row = tuple[int, Optional[int], str]
```

egparedes (approved):
LGTM (with a couple of style nitpicks).

src/gt4py/next/errors/exceptions.py
havogt: Not exactly: because this writes to `.notes`, but the python >= 3.11 `add_note` appends to `__notes__`. So should we actually add `__notest__` for Python 3.10? Or should we merge the removal of 3.10 first?

src/gt4py/next/errors/formatting.py
havogt: Should be `_UNDERLINE_ROW`, no?

src/gt4py/next/errors/exceptions.py
egparedes: Is there any drawback for writing always to `__notes__`  and add only this method to 3.10? It feels like the natural solution to me, but maybe I'm missing something

src/gt4py/next/errors/formatting.py
egparedes: Ah, no, I get it now. This is a just a private `TypeAlias`, that's why is using PascalCasing, so it's correct. I would the `TypeAlias` annotation then to make it clear, and also the `Final` to the (private or public) constants

src/gt4py/next/errors/exceptions.py
havogt: Variant (b) — `add_note` writing to native `__notes__` instead of our `.notes` field — is workable, but on our 3.10 floor it's strictly more code than it looks, not less.

The subtlety isn't really double-printing. For a `DSLError` our excepthook formats and returns without ever calling the fallback (`excepthook.py`), so it bypasses CPython's `__notes__` display entirely. Going native therefore isn't "delete code", it's a 3.10/3.11 split across three places:

- `add_note` → native on 3.11+, shim on 3.10;
- the formatter must stop emitting the breadcrumb on 3.11+ (else pytest/IPython, which *do* render `__notes__`, would double-print it);
- the excepthook must itself read `__notes__` on 3.11+.

Plus `.notes` currently carries two kinds of note — construction-time factual ones (rendered inside the diagnostic) and the `add_note` breadcrumb — and only the breadcrumb is a `__notes__` candidate, so native also means either splitting the two channels or detaching the factual notes to the end of the output.

All of that collapses to a clean, branch-free change once the floor is ≥3.11, so my preference is to wait and switch wholesale then rather than carry the split now.

havogt: Update — went ahead and did this now (e133275c9), since reviewing the final state is easier than reviewing the override + a promise.

`add_note` now uses native `__notes__`. On 3.10 a small shim populates `__notes__`, and `__str__` folds it into `str()` to reproduce what the >=3.11 traceback machinery does automatically; both are guarded by `if sys.version_info < (3, 11)` and delete cleanly when the floor moves up. Structured `notes` stays reserved for raise-site content; breadcrumbs live in `__notes__`.

The one permanent addition is the excepthook rendering `__notes__` itself — it replaces the default hook, so the machinery never gets the chance.

PR #2658, opened 2026-06-16
fix[next]: scalar dtype in math builtins + implement embedded gamma (#2277)

### LINKED ISSUE #2277: gtx.sqrt result type on scalars is Python float by default
For scalars, the gtx.sqrt function does not return the same type as the one it is called with:
```
four = gtx.float32(4)
type(gtx.sqrt(four)) => float
type(four) => numpy.float32
```

havogt:
Two related changes to gt4py.next math builtins, both centered on gamma and scalar dtype.

1. Preserve scalar dtype in unary math builtins (fixes #2277)

Unary math builtins on scalars (gtx.sqrt, gtx.sin, ...) returned a Python float/int/bool instead of preserving the input scalar type, e.g. type(gtx.sqrt(gtx.float32(4))) was float instead of numpy.float32. The scalar fallback in ffront/fbuiltins.py used Python's math module, which widens to a Python scalar. It now uses the numpy equivalents, which keep the input dtype, matching the field path (nd_array_field.py) and the iterator-embedded path (iterator/embedded.py). gamma has no numpy equivalent, so it uses scipy.special.gamma when available (as in gt4py.cartesian), falling back to math.gamma with a cast back to the input scalar type.

2. Implement gamma for embedded fields (numpy/cupy/jax)

gtx.gamma on an embedded field previously raised AssertionError on every backend: the field path had a '# TODO gamma' and skipped registration because numpy has no gamma ufunc. It is now registered via a per-namespace resolver: scipy.special.gamma (numpy), cupyx.scipy.special.gamma (cupy), jax.scipy.special.gamma (jax), with a np.vectorize(math.gamma) fallback when scipy is absent. _make_builtin was extended to accept a namespace-to-op resolver.

No new dependencies: scipy is already pulled in by gt4py[next] via the standard extra; cupyx.scipy and jax.scipy ship with cupy and jax.

Tests
- New backend-free unit test for scalar dtype preservation across all unary math builtins (float32/float64) and bool for the predicates.
- New gamma field-operator test exercising the full backend matrix; verified locally green on embedded numpy, embedded cupy (GPU), embedded jax, gtfn (cpu+gpu), dace (cpu+gpu), and roundtrip.

Fixes #2277.

### DIFF
--- a/src/gt4py/next/embedded/nd_array_field.py
+++ b/src/gt4py/next/embedded/nd_array_field.py
@@ -56,18 +56,12 @@ def _get_nd_array_class(*fields: common.Field | core_defs.Scalar) -> type[NdArra
 
 
 def _make_builtin(
-    builtin_name: str,
-    array_builtin_name: str | Callable[[ModuleType], Callable],
-    reverse: bool = False,
+    builtin_name: str, array_builtin_name: str, reverse: bool = False
 ) -> Callable[..., NdArrayField]:
     def _builtin_op(*fields: common.Field | core_defs.Scalar) -> NdArrayField:
         cls_ = _get_nd_array_class(*fields)
         xp = cls_.array_ns
-        op = (
-            array_builtin_name(xp)
-            if callable(array_builtin_name)
-            else getattr(xp, array_builtin_name)
-        )
+        op = _get_builtin(xp, array_builtin_name)
 
         domain_intersection = embedded_common.domain_intersection(
             *[f.domain for f in fields if isinstance(f, common.Field)]
@@ -104,18 +98,24 @@ except ImportError:
         return np.vectorize(math.gamma, otypes=[a.dtype])(a)
 
 
-def _gamma_op(xp: ModuleType) -> Callable:
-    if xp is np:
-        return _np_gamma
-    if cp is not None and xp is cp:
-        import cupyx.scipy.special
+def _get_builtin(xp: ModuleType, name: str) -> Callable:
+    match name:
+        case "gamma":
+            if xp is np:
+                return _np_gamma
+            if cp is not None and xp is cp:
+                import cupyx.scipy.special
 
-        return cupyx.scipy.special.gamma
-    if jnp is not None and xp is jnp:
-        import jax.scipy.special
+                return cupyx.scipy.special.gamma
+            if jnp is not None and xp is jnp:
+                import jax.scipy.special
 
-        return jax.scipy.special.gamma
-    raise NotImplementedError(f"'gamma' is not implemented for array namespace '{xp.__name__}'.")
+                return jax.scipy.special.gamma
+            raise NotImplementedError(
+                f"'gamma' is not implemented for array namespace '{xp.__name__}'."
+            )
+        case _:
+            return getattr(xp, name)
 
 
 _Value: TypeAlias = common.Field | core_defs.ScalarT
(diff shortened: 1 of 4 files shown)


src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -55,12 +56,18 @@ def _get_nd_array_class(*fields: common.Field | core_defs.Scalar) -> type[NdArra
 
 
 def _make_builtin(
-    builtin_name: str, array_builtin_name: str, reverse: bool = False
+    builtin_name: str,
+    array_builtin_name: str | Callable[[ModuleType], Callable],
```
havogt:
```suggestion
    array_builtin: str | Callable[[ModuleType], Callable],
```
?

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -55,12 +56,18 @@ def _get_nd_array_class(*fields: common.Field | core_defs.Scalar) -> type[NdArra
...
 ) -> Callable[..., NdArrayField]:
     def _builtin_op(*fields: common.Field | core_defs.Scalar) -> NdArrayField:
         cls_ = _get_nd_array_class(*fields)
         xp = cls_.array_ns
-        op = getattr(xp, array_builtin_name)
+        op = (
+            array_builtin_name(xp)
+            if callable(array_builtin_name)
+            else getattr(xp, array_builtin_name)
+        )
```
havogt: move this into a function
```
def _get_builtin(xp, name):
    match name:
        case "gamma":
             ....
        default:
             getattr(xp, name)
```

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -313,34 +313,44 @@ def astype(
...
 UNARY_MATH_NUMBER_BUILTIN_NAMES: Final = [*_UNARY_MATH_NUMBER_BUILTIN_IMPL.keys()]
 
+
+try:
+    from scipy.special import gamma as _gamma  # dtype-preserving ufunc
+except ImportError:
+
+    def _gamma(value: core_defs.ScalarT) -> core_defs.ScalarT:
+        # restore the input scalar type, which `math.gamma` widens to `float`
+        return cast(core_defs.ScalarT, type(value)(math.gamma(value)))
```
egparedes, line 323: I assume the `cast()` has been added here for a reason, but I would've assumed that mypy is able to infer that correctly....

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -89,6 +90,34 @@ def _builtin_op(*fields: common.Field | core_defs.Scalar) -> NdArrayField:
...
+    def _np_gamma(a: core_defs.NDArrayObject) -> core_defs.NDArrayObject:
+        return np.vectorize(math.gamma, otypes=[a.dtype])(a)
+
+
+def _get_builtin(xp: ModuleType, name: str) -> Callable:
+    match name:
+        case "gamma":
+            if xp is np:
+                return _np_gamma
+            if cp is not None and xp is cp:
```
egparedes: Just a nitpick, not a blocker, is the `is not None` check here really required? Shouldn't the type checker already guarantee that `xp` is not `None`?
```suggestion
            if xp is cp:
```

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -89,6 +90,34 @@ def _builtin_op(*fields: common.Field | core_defs.Scalar) -> NdArrayField:
...
+                import cupyx.scipy.special
+
+                return cupyx.scipy.special.gamma
+            if jnp is not None and xp is jnp:
```
egparedes: Same question as above?

egparedes (approved):
LGTM, just nipicks.

src/gt4py/next/ffront/fbuiltins.py
havogt: It's required I think because `type(value)(some_type)`  resolves to `Scalar` not the concrete `ScalarT`.

PR #2662, opened 2026-06-16
fix[eve]: use identity check for None in infer_type

havogt:
- infer_type used `value in (None, type(None))`, which invokes `__eq__` and raises for values with non-boolean equality such as NumPy arrays ("The truth value of an array ... is ambiguous").
- Switch to an identity check (`value is None or value is type(None)`).


egparedes (approved):
LGTM

PR #2666, opened 2026-06-17
refactor[scripts]: migrate DaCe determinism check to dev-scripts conventions

egparedes:
## What & why

PR #2590 added the DaCe codegen determinism-check infrastructure at the old
`scripts/` layout, landing just before #2653 restructured the dev-scripts
toolbox. This branch migrates that infrastructure onto the new conventions and
tightens its design so it is a self-contained dev-scripts CLI.

- **Migrate to dev-scripts conventions** — move
  `scripts/dace_deterministic_codegen.py` → `scripts/python/dace_determinism.py`
  (uv shebang, Typer `cli`, auto-discovered by `./scripts/run`); move its tests
  under `scripts/tests/python/`; drop the orphaned `scripts/tests/__init__.py`.
- **Absorb the orchestration** — the "run pytest twice + compare" loop that
  lived in `noxfile.py` now lives in the script. `noxfile.py` no longer imports
  the module: the `test_next_dace_determinism` session installs the `scripts`
  dependency group into its venv and *executes* the script with that venv's own
  (gt4py-capable) python.
- **Two CLI subcommands** — `check` (compare two existing build caches) and
  `ci-check` (run the suite twice, then compare).
- Drop the lazy/guarded `typer` import — the module now always runs in an
  environment where `typer` is available.

## Code-review fixes

- **Exit codes (`ci-check`)**: all comparison exceptions subclass `RuntimeError`,
  so a bare `except RuntimeError` shadowed the specific handler — a determinism
  diff exited 2 instead of 1, "no programs" exited 2 instead of 3, and the
  "see report.txt / diffs/" pointer was unreachable. Added a dedicated
  `PytestRunError` for infrastructure failures and made the handlers disjoint.
- **Interpreter limitation**: documented that `ci-check` needs a gt4py-capable
  interpreter and added `--python` to point at one (the default is correct for
  the nox session but not for a bare `./scripts/run`).
- **`.dacecache` location**: now derived from the process CWD (where DaCe
  writes it) rather than `workdir.parent`, and exposed as `--dacecache`.

## Testing

- `./scripts/test python` → 37 passed (comparison library, orchestration driven
  by a stub interpreter, and CLI exit-code regression tests).
- `pre-commit run` clean (ruff, mypy, tach, license headers).
- The CSCS `dace-determinism` pipeline and the on-disk `_dace_deterministic_codegen/`
  artifact directory name are unchanged.


kotsaloscv:
cscs-ci run dace-determinism

scripts/python/dace_determinism.py
```diff
@@ -1,4 +1,5 @@
-#!/usr/bin/env python3
+#!/usr/bin/env -S uv run -q --frozen --isolated --python 3.12 --group scripts python3
```
kotsaloscv, line 1: Could you please explain to me this line and why 3.12?
Thanks!

kotsaloscv (approved):
LGTM

scripts/python/dace_determinism.py
egparedes: This shebang runs the script in an isolated frozen environment with the right dependencies and the right python version. It is shortly mentioned in the [scripts/README](https://github.com/GridTools/gt4py/blob/main/scripts/README.md).

edopao (approved):
LGTM

PR #2667, opened 2026-06-18
feat[next]: staggered fields

tehrengruber:
A **staggered dimension** is a dimension sitting at the **half-integer** positions of a base dimension. For example in a cell centered 2d cartesian grid with cells located at `I`, `J` the edges are located at `I - ½`, `J` and `I`, `J - ½`. Differentiating between staggered and non-staggered dimensions avoids accidental errors when fields defined on different entities are combined and improves readability for users by staying close to the mathematical formulations.

Let `i_field` for simplicity be a 1d field of cell values defined on `I` and `IHalf` the corresponding staggered dimension of edges between those cells then

- `i_field(IHalf + 0.5)` → maps edges to the cell value above `(i−½)+½ = i`
- `i_field(IHalf − 0.5)` → maps edges to the cell value below `(i−½)−½ = i−1`

```
           ┊      ●      ┊       ●      ┊      ●      ┊
 I               -1              0             1
 IHalf    -1½           -½             +½            +1½
```

Similary, let `ihalf_field` be defined on `I - ½` then

- `ihalf_field(I + 0.5)` → maps cells to the edge above `i+½`
- `ihalf_field(I − 0.5)` → maps cells to the edge below `i−½`

__Storage__

`ihalf_field.ndarray[idx]` holds the value for logical index `i = domain.start + idx`, which is *interpreted* as sitting at `i − ½`. In the example above, `IHalf(0)` therefore sits at `0 − ½ = −½`, i.e. it is the index of the edge just below the cell `I(0)`. The memory layout is identical to any normal field; the "half" is purely how the index is interpreted geometrically.

Replaces #2339


src/gt4py/next/ffront/foast_to_gtir.py
```diff
@@ -318,19 +318,27 @@ def _visit_shift(self, node: foast.Call, **kwargs: Any) -> itir.Expr:
...
+                    left=foast.Name(),  # TODO(tehrengruber): use type instead
                     right=foast.Constant(value=offset_index),
                 ):
                     if arg.op == dialect_ast_enums.BinaryOperator.SUB:
                         offset_index *= -1
-                    assert isinstance(dim_name.type, ts.DimensionType)
-                    dim = dim_name.type.dim
+                    # `connectivity_for_cartesian_shift` resolves an integer shift to a translation
+                    # (`domain == codomain`) and a half (±0.5) shift to a relocation onto the
+                    # staggered dimension (`domain != codomain`).
```
tehrengruber:
```suggestion
```

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -173,36 +173,53 @@ def translate(
...
-                cod = misc.dim_from_axis_literal(off.codomain)
-                assert dom == cod  # relocation (staggering) is not supported here
-                new_ranges[dom] = SymbolicRange.translate(self.ranges[dom], val.value)
-                return SymbolicDomain(self.grid_type, new_ranges)
-            assert isinstance(off, itir.OffsetLiteral) and isinstance(off.value, str)
-            connectivity_type = common.get_offset_type(offset_provider_type, off.value)
-
-            if isinstance(connectivity_type, common.NeighborConnectivityType):
+                assert len(connectivity.domain.dims) == 1
+                # cartesian offset
```
tehrengruber:
```suggestion
```

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -660,24 +697,26 @@ def apply_shift(
...
-                            found = True
-                    assert found
+                    source_dim, target_dim = offset_axis.domain, offset_axis.codomain
                 else:
-                    raise NotImplementedError(f"{type_} is not a supported Connectivity type.")
+                    assert isinstance(offset_axis, it_ts.OffsetLiteralType)
+                    assert isinstance(offset_axis.value, str)
+                    type_ = common.get_offset_type(offset_provider_type, offset_axis.value)
+                    assert isinstance(
+                        type_, (common.CartesianConnectivityType, common.NeighborConnectivityType)
```
tehrengruber:
```suggestion
                        type_, common.NeighborConnectivityType
```

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -500,23 +511,49 @@ def _resolve_dimensions(
...
+        [Dimension(value='JDim', kind=<DimensionKind.HORIZONTAL: 'horizontal'>), Dimension(value='IDim', kind=<DimensionKind.HORIZONTAL: 'horizontal'>)]
+
     """
     resolved_dims = []
     for input_dim in input_dims:
+        resolved_dim = input_dim
         for off_literal in reversed(
             shift_tuple[::2]
-        ):  # Only OffsetLiterals are processed, located at even indices in shift_tuple. Shifts are applied in reverse order: the last shift in the tuple is applied first.
+        ):  # Only OffsetLiterals/CartesianOffsets are processed, located at even indices in shift_tuple. Shifts are applied in reverse order: the last shift in the tuple is applied first.
```
tehrengruber: Fix line length

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -500,23 +511,49 @@ def _resolve_dimensions(
...
+            else:
+                assert isinstance(off_literal, itir.OffsetLiteral) and isinstance(
+                    off_literal.value, str
+                )
+                offset_type = common.get_offset_type(offset_provider_type, off_literal.value)
+                if isinstance(offset_type, common.Dimension) and resolved_dim == offset_type:
+                    continue  # No shift applied
+                if isinstance(offset_type, common.NeighborConnectivityType):
+                    if resolved_dim == offset_type.codomain:  # Check if input fits to offset
+                        resolved_dim = offset_type.domain[0]  # Update input_dim for next iteration
```
tehrengruber:
```suggestion
                if isinstance(offset_type, common.NeighborConnectivityType):
                    if resolved_dim == offset_type.codomain:  # Check if input fits to offset
                        resolved_dim = offset_type.domain[0]  # Update input_dim for next iteration
```
For everything else this should error.

src/gt4py/next/iterator/embedded.py
```diff
@@ -588,15 +588,15 @@ def execute_shift(
         return cast(IncompletePosition, pos) | {tag: new_entry}
 
     # a `CartesianConnectivity` tag is a self-describing cartesian shift (e.g. from a
-    # `CartesianOffset` IR node); named offsets are resolved through the offset provider
+    # `CartesianOffset` IR node); named offsets are resolved through the offset provider.
+    # The position moves from the connectivity's `domain_dim` to its `codomain`: when they
+    # coincide this is a plain translation, otherwise a relocation onto another (e.g.
+    # staggered) dimension.
```
tehrengruber: The entire comment looks superfluous.

src/gt4py/next/program_processors/codegens/gtfn/gtfn_module.py
```diff
@@ -134,6 +134,8 @@ def _process_connectivity_args(
                 arg_exprs.append(
                     f"gridtools::hymap::keys<generated::{name}_t>::make_values({nbtbl})"
                 )
+            elif isinstance(connectivity_type, common.CartesianConnectivityType):
+                pass
```
tehrengruber:
```suggestion
```

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -151,39 +159,25 @@ def _collect_offset_definitions(
...
+            for v in (cart_offset.domain, cart_offset.codomain)
+        ]
+        for dim in dims:
+            if common.is_staggered(dim):
+                dim = common.flip_staggered(dim)
+            offset_definitions[dim.value] = TagDefinition(name=Sym(id=dim.value))
 
     for offset_name, connectivity_type in offset_provider_type.items():
+        # Cartesian shifts are self-describing `CartesianOffset` nodes (handled above) and do not
+        # appear in the offset provider, which only holds (gather) neighbor connectivities.
```
tehrengruber:
```suggestion
```

src/gt4py/next/common.py
```diff
@@ -1424,3 +1442,32 @@ def __gt_builtin_func__(cls, /, func: fbuiltins.BuiltInFunction[_R, _P]) -> Call
...
+
+
+def flip_staggered(dim: Dimension) -> Dimension:
+    if is_staggered(dim):
+        return Dimension(dim.value[len(_STAGGERED_PREFIX) :], dim.kind)
+    else:
+        return Dimension(f"{_STAGGERED_PREFIX}{dim.value}", dim.kind)
+
+
+def as_non_staggered(dim: Dimension) -> Dimension:
```
tehrengruber, line 1461: In some places in the PR we do `if is_staggered(...): flip_staggered(...)` instead of just `as_non_staggered`. This should be fixed.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_basic.py
```diff
@@ -71,3 +72,11 @@ def testee() -> cases.IFloatField:
...
     except ZeroDivisionError:
         pass
+
+
+def test_undefined_symbols(cartesian_case):
+    with pytest.raises(errors.DSLError, match="Undeclared symbol"):
+
+        @gtx.field_operator(backend=cartesian_case.backend)
+        def return_undefined():
+            return undefined_symbol  # noqa: F821 [undefined-name]
```
tehrengruber:
```suggestion
```
Broken merge?

src/gt4py/next/ffront/foast_to_gtir.py
tehrengruber-ai: Done, removed the comment.

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber-ai: Done, removed the stray `# cartesian offset` comment.

src/gt4py/next/iterator/type_system/type_synthesizer.py
tehrengruber-ai: Done, narrowed the assert to `common.NeighborConnectivityType`.

src/gt4py/next/iterator/type_system/type_synthesizer.py
tehrengruber-ai: Done, moved the comment onto its own (wrapped) lines to fix the line length.

src/gt4py/next/iterator/type_system/type_synthesizer.py
tehrengruber-ai: Done. Dropped the `common.Dimension` special case; now only `NeighborConnectivityType` is handled and everything else raises `NotImplementedError`.

src/gt4py/next/iterator/embedded.py
tehrengruber-ai: Done, removed the comment.

src/gt4py/next/program_processors/codegens/gtfn/gtfn_module.py
tehrengruber-ai: Done, removed the `CartesianConnectivityType` branch.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
tehrengruber-ai: Done, removed the comment.

src/gt4py/next/common.py
tehrengruber-ai: Done. Replaced the `if is_staggered(...): flip_staggered(...)` occurrences with `as_non_staggered` (in `order_dimensions`, `itir_to_gtfn_ir`, and `nanobind`).

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_basic.py
tehrengruber-ai: Yes, broken merge — `test_undefined_symbols` already exists in `test_type_deduction.py`. Removed the duplicate here (and the now-unused `pytest`/`errors` imports).

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -173,36 +173,54 @@ def translate(
...
+                domain = common.Dimension(value=off.domain.value, kind=off.domain.kind)
+                codomain = common.Dimension(value=off.codomain.value, kind=off.codomain.kind)
+                connectivity = common.CartesianConnectivity(domain, codomain=codomain)
+            elif isinstance(off, itir.OffsetLiteral):
+                assert isinstance(off.value, str)
+                # `get_offset` accepts both `OffsetProvider` and `OffsetProviderType`, but is typed
+                # for the former only (the two overlap, so an overload is not possible).
+                connectivity = common.get_offset(offset_provider, off.value)  # type: ignore[arg-type]
+            else:
+                raise AssertionError()
```
tehrengruber:
```suggestion
            else:
                assert isinstance(off, itir.OffsetLiteral) and isinstance(off.value, str)
                # `get_offset` accepts both `OffsetProvider` and `OffsetProviderType`, but is typed
                # for the former only (the two overlap, so an overload is not possible).
                connectivity = common.get_offset(offset_provider, off.value)  # type: ignore[arg-type]
```

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -679,6 +672,22 @@ def _deduce_binop_type(
...
+        elif (
+            node.op in (dialect_ast_enums.BinaryOperator.ADD, dialect_ast_enums.BinaryOperator.SUB)
+            and isinstance(left.type, ts.DimensionType)
+            and isinstance(right.type, ts.ScalarType)
+            and type_info.is_arithmetic(right.type)
+        ):
+            # e.g. `IDim+1` or `IDim+0.5`
+            if not isinstance(right, foast.Constant):
+                raise NotImplementedError(
+                    "Cartesian offsets are only supported with literal rhs, e.g. `IDim + 1`, but not `IDim + expr`."
```
havogt: what about as_offset shifts between Dim and staggered dim? Is it possible in this PR?

src/gt4py/next/ffront/foast_to_gtir.py
```diff
@@ -320,19 +320,23 @@ def _visit_shift(self, node: foast.Call, **kwargs: Any) -> itir.Expr:
                     current_expr = im.as_fieldop(
                         im.lambda_("__it")(im.deref(im.shift(shift_offset, new_index)("__it")))
                     )(current_expr)
-                # `field(Dim + idx)`
+                # `field(Dim + idx)` (and the staggered/relocation `field(Dim ± 0.5)`)
```
havogt: Do we support `Dim + 1.5`? then 
```suggestion
                # `field(Dim + idx)` (where `idx` is integer or half integers)
```
or something similar

src/gt4py/next/ffront/foast_to_gtir.py
```diff
@@ -320,19 +320,23 @@ def _visit_shift(self, node: foast.Call, **kwargs: Any) -> itir.Expr:
...
                 case foast.BinOp(
                     op=dialect_ast_enums.BinaryOperator.ADD | dialect_ast_enums.BinaryOperator.SUB,
-                    left=foast.Name() as dim_name,
+                    left=foast.Name(),  # TODO(tehrengruber): use type instead
```
havogt: can you add to the TODO why it's not done here (i.e. the unblocker of the todo).

src/gt4py/next/iterator/ir_utils/misc.py
```diff
@@ -235,6 +235,16 @@ def dim_from_axis_literal(axis_literal: itir.AxisLiteral) -> common.Dimension:
     return common.Dimension(value=axis_literal.value, kind=axis_literal.kind)
 
 
+def connectivity_from_cartesian_offset(
```
havogt: used once in dace, maybe the make_cartesian_shift there should be changed to work on CartesianOffset? Didn't have a close look if possible.

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -173,36 +173,52 @@ def translate(
...
+                assert new_dim not in new_ranges or old_dim == new_dim
+
+                new_range = SymbolicRange.translate(
+                    self.ranges[old_dim], connectivity.offset + val.value
+                )
+                new_ranges = dict(
+                    (dim, range_) if dim != old_dim else (new_dim, new_range)
+                    for dim, range_ in new_ranges.items()
+                )
+            elif common.is_neighbor_table(connectivity):
```
havogt: **`translate` no longer supports type-only offset providers.** A `NeighborConnectivityType` fails both the `CartesianConnectivity` isinstance check and `common.is_neighbor_table` and falls through to the bare `raise AssertionError()` below.

Reproduced against `main`: `dom.translate((OffsetLiteral("V2E"), OffsetLiteral(0)), {"V2E": NeighborConnectivityType(...)}, symbolic_domain_sizes={"Edge": "num_edges"})` works on `main` but raises a bare `AssertionError` on this branch — this breaks `infer_domain` with `offset_provider_type` + `symbolic_domain_sizes` (the AOT compile path) for every unstructured shift. `main` normalized via `offset_provider_to_type` and dispatched on `NeighborConnectivityType`.

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -482,41 +482,80 @@ def _resolve_dimensions(
...
+            else:
+                assert isinstance(off_literal, itir.OffsetLiteral) and isinstance(
+                    off_literal.value, str
+                )
+                offset_type = common.get_offset_type(offset_provider_type, off_literal.value)
+                if isinstance(offset_type, common.NeighborConnectivityType):
+                    if resolved_dim == offset_type.codomain:  # Check if input fits to offset
+                        resolved_dim = offset_type.domain[0]  # Update input_dim for next iteration
+                else:
+                    raise NotImplementedError(
```
havogt, line 555: **`_resolve_dimensions` now rejects plain `Dimension`-valued offset providers.** Previously such entries fell through leaving the dimension unchanged; now anything that is not a `NeighborConnectivityType` raises.

Reproduced: `_resolve_dimensions([IDim, KDim], (OffsetLiteral("Koff"), OffsetLiteral(1)), {"Koff": KDim})` returns `[IDim, KDim]` on `main` but raises `NotImplementedError("'KDim[vertical]' is not a supported connectivity type.")` here — ITIR programs using named cartesian offsets with `{"Koff": KDim}`-style providers now crash during type inference.

src/gt4py/next/ffront/fbuiltins.py
```diff
@@ -488,9 +488,7 @@ def __getitem__(self, offset: int) -> common.Connectivity:
         offset_definition = common.get_offset(current_offset_provider, self.value)
 
         connectivity: common.Connectivity
-        if isinstance(offset_definition, common.Dimension):
-            connectivity = common.CartesianConnectivity(offset_definition, offset)
-        elif isinstance(offset_definition, common.Connectivity):
+        if isinstance(offset_definition, common.Connectivity):
             assert common.is_neighbor_table(offset_definition)
             named_index = common.NamedIndex(self.target[-1], offset)
             connectivity = offset_definition[named_index]
```
havogt: **Embedded regression for `field(Off[idx])`.** `__getitem__` dropped the `isinstance(offset_definition, common.Dimension)` branch (which built a `CartesianConnectivity`), so the deprecated-but-still-supported cartesian subscript syntax now dies in this `else` with a message-less `NotImplementedError` in embedded execution, while compiled backends still accept it.

Reproduced: with `Ioff = FieldOffset("Ioff", source=IDim, target=(IDim,))` and `offset_provider={"Ioff": IDim}`, `return a(Ioff[1])` returns the shifted field on `main`, raises here on this branch.

src/gt4py/next/common.py
```diff
@@ -1336,7 +1336,13 @@ def order_dimensions(dims: Iterable[Dimension]) -> list[Dimension]:
     """Find the canonical ordering of the dimensions in `dims`."""
     if sum(1 for dim in dims if dim.kind == DimensionKind.LOCAL) > 1:
         raise ValueError("There are more than one dimension with DimensionKind 'LOCAL'.")
-    return sorted(dims, key=lambda dim: (_DIM_KIND_ORDER[dim.kind], dim.value))
+    return sorted(
```
havogt, line 1339: **`order_dimensions` is no longer a total order for staggered pairs.** Sorting by `as_non_staggered(dim).value` gives a dimension and its staggered counterpart identical sort keys, so the "canonical" order of `[IDim, IHalfDim]` depends on input order (stable-sort tie).

Reproduced: fields over `(IDim, IHalfDim)` and `(IHalfDim, IDim)` both pass `check_dims`, and adding them crashes with an `AssertionError` in embedded `nd_array_field._broadcast` because `promote_dims` picks an arbitrary order. The same tie makes dace's `order_dimensions`-derived stride/layout ordering input-order dependent. A secondary sort key on `is_staggered` would restore totality.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -151,36 +159,29 @@ def _collect_offset_definitions(
...
         node.walk_values().if_isinstance(itir.CartesianOffset)
     ).to_set()
     for cart_offset in cartesian_offsets:
-        for axis in (cart_offset.domain, cart_offset.codomain):
+        dims = [
+            common.Dimension(value=v.value, kind=v.kind)
+            for v in (cart_offset.domain, cart_offset.codomain)
+        ]
+        for dim in dims:
+            dim = common.as_non_staggered(dim)
```
havogt: **The staggered→base flip is missing in `_collect_dimensions_from_params` (line 153 above).** It is applied here in `_collect_offset_definitions`, in `FlipStaggeredDims`, and in `visit_AxisLiteral`, but params emit their raw `_StaggeredIDim` tag. Two reproduced consequences:

1. In this PR's own working staggered path, the stray `_StaggeredIDim_t` tag comes first in the dict merge, so `_block_sizes` assigns block size 32 to the dead tag and demotes the real loop dim `IDim_t` to 8 — a silent performance bug in every staggered gtfn program.
2. A staggered field param whose base dim appears in no domain/offset generates bindings referencing `generated::IDim_t`, which is never defined → C++ compile error `'IDim_t' is not a member of 'generated'`.

src/gt4py/next/common.py
```diff
@@ -1424,3 +1430,32 @@ def __gt_builtin_func__(cls, /, func: fbuiltins.BuiltInFunction[_R, _P]) -> Call
...
+def as_non_staggered(dim: Dimension) -> Dimension:
+    if is_staggered(dim):
+        return flip_staggered(dim)
+    return dim
+
+
+def connectivity_for_cartesian_shift(dim: Dimension, offset: int | float) -> CartesianConnectivity:
+    if isinstance(offset, float):
+        integral_offset, half = divmod(offset, 1)
+        assert half == 0.5
```
havogt: **User-reachable input validated by a bare `assert`.** `_deduce_binop_type` now accepts any arithmetic constant, so this `assert half == 0.5` is the only guard on fractional offsets.

Reproduced: `a(IDim + 1.0)` or `a(IDim + 0.25)` in a field operator raised a located `DSLError` on `main`; here it raises a message-less `AssertionError` from `common.py` with no source location. Under `python -O` the assert is stripped and the `divmod` truncation silently produces a wrong staggered connectivity (`IDim + 1.0` → shift by 2 into the staggered dim). This should be a `DSLError` at the type-deduction boundary.

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -173,36 +173,52 @@ def translate(
...
-            assert isinstance(off, itir.OffsetLiteral) and isinstance(off.value, str)
-            connectivity_type = common.get_offset_type(offset_provider_type, off.value)
-
-            if isinstance(connectivity_type, common.NeighborConnectivityType):
+                assert len(connectivity.domain.dims) == 1
+
+                old_dim = connectivity.domain_dim
+                new_dim = connectivity.codomain
+
+                assert new_dim not in new_ranges or old_dim == new_dim
```
havogt, line 197: **This assert forbids domains containing both a dimension and its staggered counterpart** — which the frontend type system explicitly permits (see `order_dimensions` tie).

Reproduced end-to-end: `return a(IDim + 0.5) + b` with `a: Field[[IHalfDim]]`, `b: Field[[IHalfDim, IDim]]` passes frontend type deduction, then crashes with a bare `AssertionError` here inside `infer_domain` during gtfn compilation. Even without the assert, the rename comprehension below would produce a duplicate key and `dict()` would silently drop one range.

src/gt4py/next/otf/binding/nanobind.py
```diff
@@ -209,7 +209,7 @@ def make_argument(
             source_buffer=name,
             dimensions=[
                 DimensionSpec(
-                    name=dim.value,
+                    name=common.as_non_staggered(dim).value,
```
havogt: **Duplicate SID tags for fields mixing a base dim and its staggered counterpart.** Emitting `as_non_staggered(dim).value` per dimension means a field over `[_StaggeredIDim, IDim]` gets two identical tags.

Reproduced end-to-end: `a + b` with `a: Field[IDim]`, `b: Field[IHalfDim]` type-deduces to `Field[[_StaggeredIDim, IDim]]`, and the generated binding calls `rename_numbered_dimensions<generated::IDim_t, generated::IDim_t>`, failing C++ compilation with `static assertion failed: wrong key` in gridtools hymap — an opaque error for a program the frontend accepted. (Either such fields should be rejected earlier, or the tags must stay distinct.)

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -91,15 +91,10 @@ def extract_connectivity_args(
     # the keys' order is taken into account. Any modification to the hashing
     # of offset providers may break this assumption here.
     args: list[tuple[core_defs.NDArrayObject, tuple[int, ...]]] = [
-        (ndarray, zero_origin)
+        (conn.ndarray, zero_origin)
         for conn in offset_provider.values()
-        if (ndarray := getattr(conn, "ndarray", None)) is not None
+        if common.is_neighbor_table(conn)
```
havogt, line 101: **Dropped device validation.** The old `assert all(is_neighbor_table(conn) and field_utils.verify_device_field_type(conn, device) ...)` is gone and nothing re-establishes it on the gtfn path (`verify_device_field_type` now only runs on dace paths); the `device` parameter is now unused.

A host NumPy neighbor table passed to a CUDA-compiled gtfn program was previously rejected with an `AssertionError`; now its raw pointer is handed to the kernel → invalid memory access or garbage neighbor indices. (The old guard was already inert under `python -O`, but the non-optimized fail-fast is lost.)

src/gt4py/next/iterator/type_system/type_synthesizer.py
```diff
@@ -660,24 +699,24 @@ def apply_shift(
...
+                    assert isinstance(type_, common.NeighborConnectivityType)
+                    source_dim, target_dim = type_.domain[0], type_.codomain
+
+                found = False
+                for i, dim in enumerate(new_position_dims):
+                    if dim == source_dim:
+                        assert not found
+                        new_position_dims[i] = target_dim
+                        found = True
+                assert found
```
havogt, line 719: **`assert found` is stricter than the old behavior for cartesian translations.** The previous code `continue`d for `CartesianOffsetType` ("translation leaves position dims unchanged"); now every shift requires its source dim to be present in `position_dims`.

Reproduced on hand-built ITIR: a K-shift inside an `as_fieldop` over an I-only cartesian domain type-checks on `main` but raises a bare `AssertionError` here. I could not demonstrate toolchain-produced IR hitting this (the frontend rejects shifting a field in a dim it lacks), so this may be latent — but any pass that prunes domains re-exposes it.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -91,10 +91,18 @@ def _name_from_named_range(named_range_call: itir.FunCall) -> str:
     return named_range_call.args[0].value
 
 
+class FlipStaggeredDims(eve.NodeTranslator):
+    def visit_AxisLiteral(self, node: itir.AxisLiteral) -> itir.AxisLiteral:
+        dim = ir_utils_misc.dim_from_axis_literal(node)
+        if common.is_staggered(dim):
+            return im.axis_literal(common.as_non_staggered(dim))
+        return node
```
havogt: Why does this flip?

src/gt4py/next/program_processors/runners/gtfn.py
```diff
@@ -91,15 +91,10 @@ def extract_connectivity_args(
...
     ]
```
havogt, line 102: Dow we still need this function or do we only have neighbor tables now anyway? Then this function might still be useful, but with a different name

src/gt4py/next/common.py
```diff
@@ -1424,3 +1430,32 @@ def __gt_builtin_func__(cls, /, func: fbuiltins.BuiltInFunction[_R, _P]) -> Call
...
+        if not is_staggered(dim):
+            integral_offset += 1
```
havogt, line 1485: In the ADR, we should clearly describe this hard-coded convention of where the staggered lives with respect to non-staggered.

src/gt4py/next/ffront/foast_to_gtir.py
tehrengruber-ai: Yes — `connectivity_for_cartesian_shift` only asserts `half == 0.5`, so any `int ± 0.5` shift works (e.g. `Dim + 1.5`). Will update the comment accordingly.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
tehrengruber-ai: A staggered dimension shares the tag of its base dimension in the generated gtfn code — the staggering has already been resolved to integer `CartesianConnectivity` offsets earlier in the toolchain. So when collecting the domain's dimension tags we emit the base name here rather than a separate `_StaggeredIDim` axis.

src/gt4py/next/ffront/foast_passes/type_deduction.py
tehrengruber: @tehrengruber-ai What is needed to support this?

src/gt4py/next/ffront/foast_to_gtir.py
tehrengruber: @tehrengruber-ai Let's implement it then, use `foast.Node(type=ts.DimensionType(dim=common.Dimension() as dim))`

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: True. @tehrengruber-ai  Let's fix the other issues first.

src/gt4py/next/iterator/type_system/type_synthesizer.py
tehrengruber: @tehrengruber-ai Please elaborate on this. `{"Koff": KDim}` is not a valid offset provider anymore and hence should also not be used. Is it still?

src/gt4py/next/ffront/fbuiltins.py
tehrengruber: @tehrengruber-ai I thought we had removed support instead of deprecating it?

src/gt4py/next/common.py
tehrengruber: @tehrengruber-ai Mixing staggered and non staggered should not be allowed. Let's address this in a later step.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
tehrengruber: @tehrengruber-ai Later.

src/gt4py/next/common.py
tehrengruber: Agreed.

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: As mentioned above this should not be allowed, but will be resolved later iteration of this PR.

src/gt4py/next/otf/binding/nanobind.py
tehrengruber: Same as above.

src/gt4py/next/program_processors/runners/gtfn.py
tehrengruber: The `verify_device_field_type` check should still exist. @tehrengruber-ai Is there no test for this?

src/gt4py/next/iterator/type_system/type_synthesizer.py
tehrengruber: The new behavior is the correct one.

src/gt4py/next/program_processors/runners/gtfn.py
tehrengruber: Right now we only have neighbor tables. Renaming would be inconsistent unless we also remove the wireing for the offset provider (which is also only a neighbor table). Let's postpone this design decision.

src/gt4py/next/common.py
tehrengruber: @tehrengruber-ai Let's do this in a later iteration.

src/gt4py/next/ffront/foast_to_gtir.py
tehrengruber-ai: Done (c7ad10812). Implemented the type-based match:
```python
left=foast.Node(type=ts.DimensionType(dim=common.Dimension() as dim)),
```
and dropped the now-redundant `assert isinstance(arg.left.type, ts.DimensionType)` — the dimension is bound directly from the type.

src/gt4py/next/ffront/foast_passes/type_deduction.py
tehrengruber-ai: Not supported yet. `as_offset` currently lowers to a pure translation `im.shift(im.cartesian_offset(dim), im.deref("__offset"))` (i.e. `domain == codomain`), so it cannot relocate onto the staggered dimension. To support an `as_offset` across a `Dim`/staggered pair we would need to:
1. let the offsets `OffsetType` express the relocation (`source != target` base dim, the same way the `Dim ± 0.5` binop already does via `connectivity_for_cartesian_shift`);
2. lower to `im.shift(im.cartesian_offset(conn.domain_dim, conn.codomain), ...)` and fold the `±0.5 → integer` convention into the runtime `__offset` field;
3. extend `domain_utils.translate` (inverse image) and `_resolve_dimensions` to handle a *relocating dynamic* shift.

Out of scope for this PR.

src/gt4py/next/iterator/type_system/type_synthesizer.py
tehrengruber-ai: It is no longer a valid offset provider. `OffsetProviderElem` / `OffsetProviderTypeElem` are now restricted to `NeighborTable` / `NeighborConnectivityType` (`is_offset_provider` requires every value to be a neighbor table), and named cartesian offsets are self-describing `CartesianOffset` IR nodes that do not go through the offset provider at all. The only remaining reference to the `{str: Dimension}` form is a commented-out line in `test_program.py`.

So `_resolve_dimensions` handling only `NeighborConnectivityType` is consistent — the repro passes an offset-provider shape that is no longer valid. If you want a friendlier failure I can raise a clearer error for non-`NeighborConnectivityType` values, but there is no valid caller to keep working.

src/gt4py/next/ffront/fbuiltins.py
tehrengruber-ai: It is currently inconsistent between the two execution paths:

- **Embedded** (`fbuiltins.FieldOffset.__getitem__`) is effectively **removed**: `main`s `isinstance(offset_definition, common.Dimension)` branch is gone, so it hits `raise NotImplementedError()`. That branch relied on the offset provider returning a plain `Dimension` — i.e. the `{str: Dimension}` form that was removed (see the `_resolve_dimensions` thread) — so embedded genuinely cannot support it anymore.
- **Compiled** (`foast_to_gtir`) still **deprecates** it: a `DeprecationWarning` branch lowers `field(Off[i])` to a `CartesianOffset` deduced from the offset *type* (it does not need the provider).

So we removed it in one place and only deprecated it in the other. To make it consistent I would drop the deprecated `foast_to_gtir` branch too (uniform removal). Shall I?

src/gt4py/next/program_processors/runners/gtfn.py
tehrengruber-ai: Done (c7ad10812) — restored the device guard in `extract_connectivity_args`:
```python
assert all(
    common.is_neighbor_table(conn) and field_utils.verify_device_field_type(conn, device)
    for conn in offset_provider.values()
    if hasattr(conn, "ndarray")
)
```
(re-added the `field_utils` import; `device` is used again).

Re: a test — `verify_device_field_type` itself is unit-tested in `tests/next_tests/unit_tests/test_field_utils.py::test_verify_device_field_type`. There is **no** test for this gtfn-path guard specifically (a host neighbor table handed to a CUDA-compiled gtfn program), since it needs a GPU + a compiled program. Happy to add a GPU-marked test if you want it covered.

src/gt4py/next/iterator/type_system/type_synthesizer.py
tehrengruber: Conclusion: The initial comment didn't make sense.

src/gt4py/next/ffront/foast_passes/type_deduction.py
tehrengruber: The answer above is not really helpful.
The main problem is that `as_offset` takes a field of relative integer offsets, whereas staggered offsets are non-integral. A simple workaround that I believe is totally sufficient is to do a half shift and then use an offset field.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
tehrengruber: This is just a simple way of removing all notions of a dimension being staggered since we only care about offsets. Backend support would be better, but gtfn is a mess in that regard.

havogt: Ah, I think now I get the name: flips staggered dimensions but leaves non-staggered, right? I read it as flip between staggered and non-staggered. Maybe change the name to something like `ToNonStaggeredDims`?

tehrengruber: After I saw the block sizes bug I changed the approach and now instead of flipping dimensions we use an alias for the tag. This leaves the generated code in a more readable state and centralizes how staggered dimensions are handled.

docs/development/ADRs/next/0024-Staggered_Dimension_Encoding.md
```diff
@@ -0,0 +1,34 @@
+---
```
tehrengruber: I'm good with this AI reformulated doc.

docs/development/ADRs/next/0024-Staggered_Dimension_Encoding.md
```diff
@@ -0,0 +1,34 @@
...
+tags: []
+---
+
+# Encoding Staggered Dimensions via a Name Prefix
```
havogt: Can we also cover the convention of where staggering lives, i.e. is KStaggered(0) left or right of K(0).

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: Fixed. Tested in `test_unstructured_translate_with_symbolic_domain_sizes` now.

src/gt4py/next/ffront/fbuiltins.py
tehrengruber: Support is removed now: https://github.com/GridTools/gt4py/pull/2695

src/gt4py/next/common.py
tehrengruber: Addressed in `test_premap_cartesian_invalid_fractional_offset`

src/gt4py/next/common.py
tehrengruber: Addressed in `check_dims`

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
tehrengruber: Addressed see my comment here: https://github.com/GridTools/gt4py/pull/2667#discussion_r3542785857

docs/development/ADRs/next/0024-Staggered_Dimension_Encoding.md
tehrengruber: ADR has been updated

src/gt4py/next/common.py
```diff
@@ -1424,3 +1442,32 @@ def __gt_builtin_func__(cls, /, func: fbuiltins.BuiltInFunction[_R, _P]) -> Call
...
+def as_non_staggered(dim: Dimension) -> Dimension:
+    if is_staggered(dim):
+        return flip_staggered(dim)
+    return dim
+
+
+def connectivity_for_cartesian_shift(dim: Dimension, offset: int | float) -> CartesianConnectivity:
+    if isinstance(offset, float):
+        integral_offset, half = divmod(offset, 1)
+        assert half == 0.5
```
havogt: GLM 5.2: The `assert half == 0.5` is back at this head. This was reported earlier and `tehrengruber-ai` replied "Done (c7ad10812)", but that fix is not present in the rebased tree.

`Dim + 1.0` (and `0.0`, `2.0`, … — any float literal whose value is integral) passes the frontend gate at `type_deduction.py:689` (`offset_index % 1 not in (0, 0.5)`, and `1.0 % 1 == 0.0`), then reaches this line where `divmod(1.0, 1) == (1.0, 0.0)` trips the assert with a message-less `AssertionError`.

Under `python -O` (which the project recommends — a `UserWarning` is emitted without it) the assert is stripped: `Dim + 1.0` on a field defined on `_StaggeredIDim` is silently accepted and produces `CartesianConnectivity(IDim, 2, codomain=_StaggeredIDim)` — i.e. a staggered shift by 2 instead of the intended integer shift by 1. Wrong results, no error.

Suggest replacing the bare assert with a non-strippable `ValueError`/`DSLError`, and reconciling the `isinstance(offset, float)` discriminator here with the `% 1 not in (0, 0.5)` check in `type_deduction.py` so integral floats are either rejected or coerced to `int`.

src/gt4py/next/common.py
```diff
@@ -1424,3 +1442,32 @@ def __gt_builtin_func__(cls, /, func: fbuiltins.BuiltInFunction[_R, _P]) -> Call
...
+        return Dimension(f"{_STAGGERED_PREFIX}{dim.value}", dim.kind)
+
+
+def as_non_staggered(dim: Dimension) -> Dimension:
+    if is_staggered(dim):
+        return flip_staggered(dim)
+    return dim
+
+
+def connectivity_for_cartesian_shift(dim: Dimension, offset: int | float) -> CartesianConnectivity:
```
havogt, line 1468: GLM 5.2: Missing docstrings on the four new helpers (`is_staggered`, `flip_staggered`, `as_non_staggered` at lines 1448–1463, and this one). `connectivity_for_cartesian_shift` is the one that needs it most — the `integral_offset += 1` (non-staggered branch) and the half-integer→integer+offset convention are non-obvious and currently only explained in the ADR.

Also: are these intended to be part of the user-facing API? Tests reach into `common.flip_staggered` to construct `IHalfDim`, so the feature isn't usable without them, but they're not in `gt4py.next.__init__.py`'s `__all__`. If they're experimental, the ADR should say so; otherwise they should be exported.

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -679,6 +672,31 @@ def _deduce_binop_type(
...
             return ts.DomainType(dims=promote_dims(left.type.dims, right.type.dims))
+        elif (
+            node.op in (dialect_ast_enums.BinaryOperator.ADD, dialect_ast_enums.BinaryOperator.SUB)
+            and isinstance(left.type, ts.DimensionType)
+            and isinstance(right.type, ts.ScalarType)
+            and type_info.is_arithmetic(right.type)
+        ):
+            # e.g. `IDim+1` or `IDim+0.5`
+            if not isinstance(right, foast.Constant):
+                raise NotImplementedError(
```
havogt: GLM 5.2: `IDim + expr` (non-literal rhs) is a user-facing DSL rejection, but this raises a bare `NotImplementedError` without a `SourceLocation` — `right.location` is available right here. Per `docs/development/next/error-messages.md`, DSL errors should be `DSLError` with a location. (The message also uses backticks around `` `IDim + 1` `` / `` `IDim + expr` `` where the guidelines ask for single quotes.)

src/gt4py/next/ffront/foast_passes/type_deduction.py
```diff
@@ -679,6 +672,31 @@ def _deduce_binop_type(
...
+                )
+            offset_index = right.value
+            if node.op == dialect_ast_enums.BinaryOperator.SUB:
+                offset_index *= -1
+            if offset_index % 1 not in (0, 0.5):
+                raise errors.DSLError(
+                    right.location,
+                    f"Invalid offset '{right.value}' for a Cartesian shift of dimension "
+                    f"'{left.type.dim.value}'.",
+                    label="expected an integer offset (to shift within the dimension) or a "
```
havogt: GLM 5.2: The actionable guidance ("expected an integer offset … or a half-integer offset such as '0.5' or '-1.5'") is passed as `label`, but `label` is meant for a short fragment shown after the caret. Per error-messages.md, "what to do instead" guidance belongs in `hints=` so the renderer formats it as a `Hint:` line; the current `label` is also too long for a label.

Separately, this diagnostic isn't pinned by a rendered-text test in `tests/next_tests/unit_tests/ffront_tests/test_diagnostic_messages.py` — only `pytest.raises(match="Invalid offset")` in `test_foast_to_gtir.py:119`, which checks the raw message rather than the rendered output (spans, carets, hint placement).

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -181,55 +179,70 @@ def translate(
...
-        offset_provider_type = common.offset_provider_to_type(offset_provider)
-
         dims = list(self.ranges.keys())
         new_ranges = {dim: self.ranges[dim] for dim in dims}
         if len(shift) == 0:
             return self
         if len(shift) == 2:
             off, val = shift
+
             if isinstance(off, itir.CartesianOffset):
```
havogt, line 189: GLM 5.2: This new relocation branch (`CartesianOffset` with `domain != codomain`, the dict key-remap at 201–204, and the `assert new_dim not in new_ranges or old_dim == new_dim` at 198) has no unit test — `test_domain_utils.py` contains no `CartesianOffset` references. It's only exercised end-to-end via `test_staggered.py`. Given the non-trivial invariant, a direct unit test for `SymbolicDomain.translate` with a staggered `CartesianOffset` (both `Dim + 0.5` and `Dim - 0.5`) would be worth adding.

(nit) The comment on line 190 ("cartesian shift: extract the (co)domain directly") restates what the next two lines do rather than a non-obvious WHY — the coding guidelines restrict comments to hidden constraints/invariants.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -151,36 +151,28 @@ def _collect_offset_definitions(
...
 
-    # cartesian shifts (`field(Dim + n)`) are encoded as `CartesianOffset` nodes and don't
-    # occur in the `offset_provider_type`; define a tag for each of their dimensions
     cartesian_offsets: set[itir.CartesianOffset] = (
         node.walk_values().if_isinstance(itir.CartesianOffset)
     ).to_set()
     for cart_offset in cartesian_offsets:
-        for axis in (cart_offset.domain, cart_offset.codomain):
+        dims = [
+            common.Dimension(value=v.value, kind=v.kind)
```
havogt: GLM 5.2: This inlines `common.Dimension(value=v.value, kind=v.kind)` even though the file already imports `misc as ir_utils_misc` (line 20) and `ir_utils_misc.dim_from_axis_literal` does exactly this. The same PR uses the helper correctly in `type_synthesizer.py:544-545` and `gtir_dataflow.py:1503-1504`, and inlines it again at `domain_utils.py:195-196`. Trivial one-liner for consistency.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_staggered.py
```diff
@@ -0,0 +1,141 @@
+# GT4Py - GridTools Framework
+#
+# Copyright (c) 2014-2024, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+import functools
```
havogt: GLM 5.2: This import block looks copy-pasted from a template — 32 of the ~40 imported names are unused (`functools`, `math`, `reduce`, `TypeAlias`, `numpy`, `astype`, `broadcast`, `common`, `errors`, `float32/64`, `int32/64`, `minimum`, `neighbor_sum`, `gt_utils`, `as_offset`, `C2E`, `E2V`, `V2E`, `Edge`, `JDim`, `KDim`, `V2EDim`, `unstructured_case`, `unstructured_case_3d`, `exec_alloc_descriptor`, `mesh_descriptor`). Ruff excludes `tests/**` so pre-commit doesn't catch it, but it's worth cleaning up.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_staggered.py
```diff
@@ -0,0 +1,141 @@
...
+
+    size = cartesian_case.default_sizes[IDim]
+    a = cases.allocate(cartesian_case, testee, "a", domain={IDim: (-1, size - 1)})()
+    out = cases.allocate(cartesian_case, testee, cases.RETURN, domain={IHalfDim: (0, size)})()
+
+    cases.verify(cartesian_case, testee, a, out=out, ref=a[:], offset_provider={})
+
+
+@pytest.mark.uses_cartesian_shift
+def test_cartesian_half_shift_half2center(cartesian_case):
```
havogt, line 102: GLM 5.2: A few scenarios the implementation supports but aren't tested here (all run on CPU backends):

- **Multi-dimensional staggered field** (e.g. `Field[[IHalfDim, JDim]]` with `a(IHalfDim + 0.5)`) — the ADR's headline use case; exercised at the codegen layer by `_add_staggered_aliases`/`_block_sizes` but with no integration test.
- **Vertical (K) half-shift** (`a(KDim + 0.5)` or `a(KHalfDim + 0.5)`). `KHalfDim` is defined in `cases_utils.py:153` and in `default_sizes` but never used with an actual half-shift; vertical staggering is the primary atmospheric use case.
- **Direct unit test for `connectivity_for_cartesian_shift`** with negative half-integers (`Dim - 0.5`, `Dim - 1.5`) — these take a distinct code path from the tested `+0.5` and the (staggered-field) `-0.5`.

docs/development/ADRs/next/0026-Staggered_Dimensions.md
```diff
@@ -0,0 +1,74 @@
...
+ I               -1              0             1
+ IHalf    -1½           -½             +½            +1½
+```
+
+Symmetrically, let `ihalf_field` be defined on `I − ½` (i.e. on `IHalf`). Then:
+
+- `ihalf_field(I + 0.5)` → maps cells to the edge above: `i+½`
+- `ihalf_field(I − 0.5)` → maps cells to the edge below: `i−½`
+
+### Storage
```
havogt, line 50: GLM 5.2: The ADR doesn't mention that a dimension and its staggered counterpart must not appear together in the same field/domain. This is a first-class constraint enforced in `check_dims` (`common.py:1348-1364`, with a detailed code comment) and tested in `test_common.py:591-595`, but a reader of the ADR alone wouldn't know it exists until they hit the `ValueError`. Worth a short paragraph alongside the storage/encoding discussion.

(Also not documented: the single-staggered-counterpart-per-dimension limitation. Sufficient for an Arakawa C-grid — u staggered in I, v in J — but two distinct staggered locations on the *same* axis aren't expressible. Worth stating as an explicit limitation.)

docs/development/ADRs/next/0024-Staggered_Dimensions.md
```diff
@@ -0,0 +1,74 @@
...
+on `Dimension`. The helpers `is_staggered`, `flip_staggered` and
+`as_non_staggered` operate purely on that prefix.
+
+Dimensions are identified by their **name** and appear throughout the toolchain
+in more than one form: as a `common.Dimension` instance, but also as an
+`itir.AxisLiteral` (and as plain strings in generated backend code). Adding a
+"staggered" flag to `Dimension` would have required threading it through every one
+of these representations and every place that constructs, compares, or lowers a
+dimension — a large, invasive change touching the frontend, the IR, and all
+backends. Carrying the marker in the name instead means it round-trips through all
```
havogt: GLM 5.2: The "round-trips through all representations for free" claim isn't quite accurate — backends actively handle the prefix:

- **gtfn** (`itir_to_gtfn_ir.py:_add_staggered_aliases`, ~line 200) turns every staggered tag into a C++ `alias` of its base dimension's `TagDefinition`, and `visit_AxisLiteral` (~line 401) emits `node.type.dim.value` instead of `node.value`.
- **nanobind** (`nanobind.py:212`) uses `common.as_non_staggered(dim).value` so the generated binding references the base dimension name.

A contributor adding a new backend needs to know the prefix must be handled at code-generation boundaries. Suggest adding a short "Backend handling" subsection.

havogt (commented):
GLM 5.2: Review of the staggered-fields feature, run against the current head (`df11778e2`). Most of the earlier round (mixing restriction, `_collect_dimensions_from_params` block sizes, `extract_connectivity_args` device guard, `field(Off[idx])` removal) is resolved in the rebased tree — thanks.

One item that was reported earlier and marked resolved (`connectivity_for_cartesian_shift`'s bare `assert half == 0.5`, `tehrengruber-ai` "Done (c7ad10812)") does **not** appear at this head, so the fix looks to have been lost during the rebase. Details in the inline comment at `common.py:1468`.

Remaining items are mostly ADR completeness, error-message formatting, and a few test gaps for the new code paths. No other correctness issues found — the half-integer offset arithmetic is correct for all sign/direction combinations I checked, and the dace/gtfn/embedded backends handle relocation consistently.

src/gt4py/next/ffront/foast_passes/type_deduction.py
tehrengruber: @tehrengruber-ai Make this a DSLError.

src/gt4py/next/common.py
tehrengruber: @tehrengruber-ai: Add brief docstrings with doctest examples. Export is_staggered, flip_staggered, as_non_staggered.

src/gt4py/next/common.py
tehrengruber: @tehrengruber-ai Let's not rely on the type
```
integral_offset, remainder = divmod(offset, 1)
if remainder == 0.5:
  ...
else:
  assert remainder == 0
  ...
```

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: @tehrengruber-ai Let's add unit tests. The comment in 190 is unnecessary.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
tehrengruber: @tehrengruber-ai Use the helper

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_staggered.py
tehrengruber: @tehrengruber-ai Let's clean this up.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_staggered.py
tehrengruber: 1. Let's add those.
2. Let's also add those, paramterized on the grid type.
3. Let's add Dim -0.5.

docs/development/ADRs/next/0026-Staggered_Dimensions.md
tehrengruber: @tehrengruber-ai Add a single sentence.

> two distinct staggered locations on the same axis aren't expressible. Worth stating as an explicit limitation

This doesn't make sense to me.

docs/development/ADRs/next/0024-Staggered_Dimensions.md
tehrengruber: @tehrengruber-ai Let's remove the `common.as_non_staggered(dim)` in the bindings unless there is a reason this is needed.

src/gt4py/next/ffront/foast_passes/type_deduction.py
tehrengruber-ai: Addressed in 544232d2b: the guidance now goes through `hints=` (rendered as a `Hint:` line) instead of `label`, and a rendered-text test was added — `test_diagnostic_messages.py::test_invalid_cartesian_offset_suggests_valid_offsets`.

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber-ai: Addressed in 544232d2b: added `test_domain_utils.py::test_translate_staggered_cartesian_offset` (both `+0.5` and `-0.5`), and removed the restating comment.

src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
tehrengruber-ai: Addressed in 544232d2b: now uses `ir_utils_misc.dim_from_axis_literal` (also switched the `domain_utils.py` inline construction to the same helper).

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_staggered.py
tehrengruber-ai: Addressed in 544232d2b: trimmed the import block to the names actually used. (`exec_alloc_descriptor` is kept — it is a required fixture dependency of `cartesian_case`.)

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_staggered.py
tehrengruber-ai: Addressed in 544232d2b: added `test_cartesian_half_shift_vertical` (K half-shift) and `test_cartesian_half_shift_multi_dim` (`Field[[IHalfDim, JDim]]`), plus a parametrized `connectivity_for_cartesian_shift` unit test in `test_common.py` covering negative half-integers.

docs/development/ADRs/next/0026-Staggered_Dimensions.md
tehrengruber-ai: Addressed in 544232d2b: added a 'Dimension Constraints' section documenting that a dimension and its staggered counterpart must not appear together (enforced by `check_dims`).

docs/development/ADRs/next/0024-Staggered_Dimensions.md
tehrengruber-ai: Addressed in 544232d2b: softened the 'round-trips for free' wording and added a 'gtfn backend' section describing the alias handling.

docs/development/ADRs/next/0026-Staggered_Dimensions.md
```diff
@@ -0,0 +1,101 @@
+---
```
havogt, line 1: Index 24 and 25 is taken in the meantime.

docs/development/ADRs/next/README.md
```diff
@@ -18,6 +18,7 @@ Writing a new ADR is simple:
 
 - [0005 - Extending Iterator IR](0005-Extending_Iterator_IR.md)
 - [0023 - Fingerprinting](0023-Fingerprinting.md)
+- [0024 - Staggered Dimensions](0024-Staggered_Dimensions.md)
```
havogt:
```suggestion
- [0026 - Staggered Dimensions](0026-Staggered_Dimensions.md)
```

src/gt4py/next/ffront/foast_to_gtir.py
```diff
@@ -304,19 +304,22 @@ def _visit_shift(self, node: foast.Call, **kwargs: Any) -> itir.Expr:
                     current_expr = im.as_fieldop(
                         im.lambda_("__it")(im.deref(im.shift(offset_name.id, new_index)("__it")))
                     )(current_expr)
-                # `field(Dim + idx)`
+                # `field(Dim + idx)` (where `idx` is integer or half integer)
                 case foast.BinOp(
                     op=dialect_ast_enums.BinaryOperator.ADD | dialect_ast_enums.BinaryOperator.SUB,
-                    left=foast.Name() as dim_name,
+                    left=foast.LocatedNode(type=ts.DimensionType(dim=common.Dimension() as dim)),
```
havogt, line 310: `LocatedNode` makes sense?

src/gt4py/next/__init__.py
```diff
@@ -122,6 +125,9 @@
     "Domain",
     "unit_range",
     "UnitRange",
+    "is_staggered",
```
havogt, line 128: Should `is_staggered` and `as_non_staggered` be part of the public API? Do you have a use-case?

havogt (approved):
Please check my questions. No need to do anything if you think current state makes sense.

src/gt4py/next/__init__.py
tehrengruber: Yes, for example when we construct the various domains for staggered and non-staggered index spaces in PMAP their size depends on staggeredness of a dimension so we use this to distinguish them. There are other properties where staggeredness does not matter, but only the geometric dimension is relevant. In those cases `as_non_staggered` is used.

src/gt4py/next/ffront/foast_to_gtir.py
tehrengruber: Somewhat, the located property does not matter, but the root node in `foast` is `LocatedNode` and we only care about the type here.

Issue #2205, opened 2025-08-05
next: issue with domain inference with concat_where at domain boundaries


havogt:
```
testee_program(a, b, out, edges, levels) {
  __tmp_1 = temporary(domain=u⟨ Edgeₕ: [0, edges[, KDimᵥ: [0, 11[ ⟩, dtype=int32);
  __tmp_1 @ u⟨ Edgeₕ: [0, edges[, KDimᵥ: [0, 11[ ⟩
       ← as_fieldop(λ(a, b) → ·⟪E2Vₒ, 0ₒ⟫(a) + ·b, u⟨ Edgeₕ: [0, edges[, KDimᵥ: [0, 11[ ⟩)(a, b);
  out @ u⟨ Edgeₕ: [0, edges[, KDimᵥ: [0, 10[ ⟩
       ← as_fieldop(
           λ(__ct_flat_el_0___tcw_pos, tmp0ᐞ0) → 
             (λ(_cs_1, _icdlv_6) → 
                if True ∧ _cs_1 < 9 then _icdlv_6() else if 10 <= _cs_1 ∧ True then _icdlv_6() else ·tmp0ᐞ0)(
               ·__ct_flat_el_0___tcw_pos, λ() → ·tmp0ᐞ0 - ·⟪Koffₒ, 1ₒ⟫(tmp0ᐞ0)
             ), u⟨ Edgeₕ: [0, edges[, KDimᵥ: [0, 10[ ⟩
         )(index(KDimᵥ), __tmp_1);
}
```

In this example `__tmp_1` is only consumed in `KDim: [0, 10[`, but computed in `[0, 11[`. The reason is that the domain inference takes both branches of the `if` (from concat_where) into account and therefore concludes that K domain needs to have the extra level.

havogt:
```
def test_nested_reduction(unstructured_case_3d):
    @gtx.field_operator
    def testee(
        a: gtx.Field[[Vertex, KDim], np.int32],
        b: gtx.Field[[Edge, KDim], np.int32],
        levels: np.int32,
    ) -> gtx.Field[[Edge, KDim], np.int32]:
        tmp0 = a(E2V[0]) + b
        tmp1 = concat_where(
            KDim == levels - 1,
            tmp0,
            tmp0 - tmp0(Koff[1]),
        )
        return tmp1

    @gtx.program(backend=unstructured_case_3d.backend)
    def testee_program(
        a: gtx.Field[[Vertex, KDim], np.int32],
[...]

PR #2677, opened 2026-06-29
fix[next]: symbolic domain union for empty domains (static case)

tehrengruber:
Domain inference computes the union of two domains as the per-dimension convex hull:

```
union(A, B) = [ min(A.start, B.start), max(A.stop, B.stop) [
```

This is wrong when the inputs are empty (`start >= stop`). The union of two empty domains should stay empty, but `min`/`max` produces a non-empty range:

- `A = [10, 10[` (empty)
- `B = [11, 11[` (empty)

```
union(A, B) = [ min(10, 11), max(10, 11) [ = [10, 11[
```

so two empty domains wrongly union to the non-empty `[10, 11[`.

This shows up in practice with `concat_where`: a branch that is never selected produces an empty domain, and previously the empty range leaked into the `min`/`max` and over-approximated the inferred domain (#2205).

**Fix**

An empty domain is the identity element of union, so empty ranges are dropped before folding rather than fed into `min`/`max` (`_reduce_ranges` in `domain_utils.py`). The union of only empty domains then correctly stays empty.

**Scope**

This PR only handles static domains — concrete integer bounds, where emptiness is decidable. The symbolic case (runtime-sized bounds like `[0, n[`, where emptiness can't be decided at compile time and empty ranges must instead be guarded with conditional expressions) is considerably more involved and is left to a follow-up. A prototype implementation can be found in #2673.


tests/next_tests/unit_tests/iterator_tests/ir_utils_tests/test_domain_utils.py
```diff
@@ -69,53 +61,86 @@ def test_domain_op_preconditions():
 
 
 def test_domain_union():
```
tehrengruber, line 63: These tests have a static domain now. In https://github.com/GridTools/gt4py/pull/2673 both cases are covered again.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_global_tmps.py
```diff
@@ -379,6 +380,9 @@ def add_shifted(domain: itir.FunCall | None = None):
 
     actual = global_tmps.create_global_tmps(testee, offset_provider, uids=uids)
     actual = ct.CollapseTuple.apply(actual, uids=uids)
+    actual = cf.ConstantFolding.apply(
```
tehrengruber, line 383: This is a detail originating from the fact that empty domains are now returned as is, without constant folding.

src/gt4py/next/iterator/ir_utils/domain_utils.py
```diff
@@ -236,16 +244,24 @@ def _reduce_ranges(
...
-        lambda current_expr, el_expr: stop_reduce_op(current_expr, el_expr),
-        [range_.stop for range_ in ranges],
-    )
-    # constant fold expression to keep the tree small
+    """
+    Uses start_op and stop_op to fold the start and stop of a list of ranges.
+
+    This function only computes the correct value if the (non-empty) ranges are either
+    overlapping / adjacent or empty, as calculation is by means of the convex hull. Non-static
+    ranges may not be empty for now.
```
havogt, line 252: What does "may not be empty" mean? Does it mean they are not detected as empty or that something breaks if they are not empty?

havogt (approved):
lgtm

src/gt4py/next/iterator/ir_utils/domain_utils.py
tehrengruber: It breaks (in the sense that the bug that this PR addresses for static ranges occurs).

PR #2679, opened 2026-06-30
feat[next]: Compilation with process pool

havogt:
Compile programs in worker processes instead of threads. Threads cannot hide the CPython-side cost of translation/codegen and dace's compile path does not parallelize well under the GIL; processes can.

- New `gt4py.next.otf.runners` with a `Runner` protocol and three implementations: `SerialRunner`, `ThreadRunner`, `ProcessRunner` (a `ProcessPoolExecutor` with `spawn` start method). Selected via `GT4PY_BUILD_JOBS_MODE=serial|thread|process`; default is `process`.
- `gt4py.next.otf.compilation_tasks` prepares the runner-ready `CompilationTask`: frontend lowering happens main-side (the raw user function does not cross the process boundary); the `ProcessRunner` ships the pickled executor and the lowered program to a worker, which runs the artifact-producing step. Runners return a future of the `CompilationArtifact`; the pool loads it in the main process on first use.
- Connectivity tables are not pickled into the tasks: they are dumped once per run to a temporary file (lazily, on first shipping) and cross the process boundary as memory-mapped file references, shared by all workers through the page cache.
- Worker processes never touch the parent's GPU: the CUDA architecture is detected once main-side (`CUDAARCHS` takes precedence) and exported to the workers, which run with hidden devices.
- Only stdlib `pickle` is used. Executors are picklable now; this required giving the fingerprinting deconstructor singletons their own pickle identity (`name`/`module` arguments on the maker functions, like `collections.namedtuple`).
- Backends that customize `compile` or whose executor is not picklable are compiled in the calling thread instead, with a warning.
- The default runner is created lazily (importing gt4py does not spin up multiprocessing machinery) and shut down atexit; `wait_for_compilation()` waits for all runners and raises if a compilation failed.
- `spawn`, not `fork`: the parent may hold threads, CUDA contexts and MPI state; fork would not relax the picklability requirement anyway since `ProcessPoolExecutor` pickles every submitted task.

Tested with ICON ("blueline", dace GPU backend) on santis: startup-to-first-timestep dropped from ~28.5 min to ~10.4 min on a clean cache, and compilation can now overlap the timeloop.


src/gt4py/next/backend.py
```diff
@@ -164,3 +165,25 @@ def __gt_allocator__(
...
+    executor: workflow.Workflow[definitions.CompilableProgramDef, stages.CompilationArtifact],
+) -> bytes:
+    """Pickle ``executor`` for shipment to a worker process.
+
+    Strips ``CachedStep`` wrappers (outer + on ``OTFCompileWorkflow.translation``):
+    the cache is process-local, the worker is one-shot, and the strip avoids
+    ``functools.singledispatch`` dispatch tables that stdlib pickle can't ship.
+    """
+    if isinstance(executor, workflow.CachedStep):
+        executor = executor.step
```
edopao: This first if-branch should never be hit, after #2623, I guess.

src/gt4py/next/backend.py
```diff
@@ -164,3 +165,25 @@ def __gt_allocator__(
...
+    if dataclasses.is_dataclass(executor):
+        unwrapped_fields = {
+            f.name: getattr(executor, f.name).step
+            for f in dataclasses.fields(executor)
+            if isinstance(getattr(executor, f.name), workflow.CachedStep)
```
edopao: The translation step is the only `CachedStep`, currently. Does this code disable the translation cache, when using the worker?

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,229 @@
...
+    """Compiles in a :class:`concurrent.futures.ProcessPoolExecutor` (``spawn``).
+
+    The worker runs ``backend.executor`` (post-translation compile) and returns a
+    picklable :class:`~gt4py.next.otf.stages.CompilationArtifact`; the main
+    process rehydrates it via ``artifact.load()`` (in a done-callback) so the
+    returned future yields a live ``ExecutableProgram``.
+    """
+
+    def __init__(self, max_workers: int, shared_session_cache_dir: str) -> None:
+        # spawn (not fork): fork-after-threads / fork-with-CUDA is unsafe.
```
havogt: Why can't we fork?

src/gt4py/next/fingerprinting.py
```diff
@@ -662,6 +676,13 @@ def catabolize(
 #: rejecting non-importable ones, and honors the
 #: ``gt4py_metadata(fingerprint=False)`` opt-out.
 strict_fingerprint_deconstructor: Final[Deconstructor] = make_strict_data_deconstructor()
+# ``functools.singledispatch`` (inside ``singledispatcher``) copies the wrapped
+# fallback's ``__qualname__``/``__name__`` onto the dispatcher, so pickle-by-reference
+# resolves to the *original* plain function instead of this singleton. Restore the
+# dispatcher's own identity so it round-trips through stdlib ``pickle`` (e.g. when an
+# executor carrying this deconstructor is shipped to a process-pool worker).
+strict_fingerprint_deconstructor.__qualname__ = "strict_fingerprint_deconstructor"
+strict_fingerprint_deconstructor.__name__ = "strict_fingerprint_deconstructor"
```
havogt: @egparedes does this make sense?

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,228 @@
...
+        definition_stage: Any,
+        compile_time_args: arguments.CompileTimeArgs,
+    ) -> concurrent.futures.Future[stages.ExecutableProgram]:
+        # Frontend lowering stays main-side: decorators rebind the user's
+        # function module attribute, so the raw ``types.FunctionType`` does not
+        # cross the process boundary. The worker receives the lowered,
+        # pickle-safe ``CompilableProgramDef`` and runs only ``backend.executor``.
+        compilable = backend.transforms(
+            definitions.ConcreteProgramDef(data=definition_stage, args=compile_time_args)
+        )
```
havogt: this is kind of a hack, bypassing the standard workflow

src/gt4py/next/fingerprinting.py
egparedes: yes, it makes sense. Actually it might be a better idea to add optional "name" and "module" keyword only arguments to the deconstructor maker functions to optionally specify this in the make call, exactly like "collections.namedtuple" does with the "module" argument.

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,228 @@
...
+        return ThreadRunner(max_workers=config.BUILD_JOBS)
+    if mode is config.BuildJobsMode.PROCESS:
+        return ProcessRunner(
+            max_workers=config.BUILD_JOBS,
+            shared_session_cache_dir=str(_cache._session_cache_dir_path),
+        )
+    raise ValueError(f"Unsupported BUILD_JOBS_MODE={mode!r}.")
+
+
+_default_runner: CompilationRunner | None = None
```
egparedes: Question: isn't a lock needed to protect concurrent access to this global variable?

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,228 @@
...
+        compilable = backend.transforms(
+            definitions.ConcreteProgramDef(data=definition_stage, args=compile_time_args)
+        )
+        executor_blob = pickle.dumps(backend.executor)
+        artifact_future = self._pool.submit(
+            _process_pool_compile_job,
+            executor_blob,
+            compilable,
+            _config_snapshot(),
+        )
```
egparedes: It would probably make the core more readable to pass the arguments to the submitted function as keyword arguments here.

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,228 @@
...
+        loaded: concurrent.futures.Future[stages.ExecutableProgram] = (
+            concurrent.futures.Future()
+        )
+
+        def _load(
+            artifact_future: concurrent.futures.Future[stages.CompilationArtifact],
+        ) -> None:
+            try:
+                loaded.set_result(artifact_future.result().load())
+            except BaseException as exception:  # noqa: BLE001 - re-raised via the future
```
egparedes:
```suggestion
            except BaseException as exception:  # noqa: BLE001 [blind-except] re-raised via the future
```

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,228 @@
...
+        compile_time_args: arguments.CompileTimeArgs,
+    ) -> concurrent.futures.Future[stages.ExecutableProgram]:
+        future: concurrent.futures.Future[stages.ExecutableProgram] = (
+            concurrent.futures.Future()
+        )
+        try:
+            future.set_result(
+                backend.compile(definition_stage, compile_time_args=compile_time_args)
+            )
+        except BaseException as exception:  # noqa: BLE001 - re-raised via the future
```
egparedes:
```suggestion
        except BaseException as exception:  # noqa: BLE001 [blind-except] re-raised via the future
```

egparedes (commented):
Looks good in general, just a few questions and comments, besides the previous reply to your mention about the changes in the fingerprinting module

src/gt4py/next/otf/compilation_runner.py
havogt: I think it's not strictly needed because this should only be accessed in the main thread, but fable said it's easier to do then argue against it. ;)

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,268 @@
...
+        self._pool.shutdown(wait=wait)
+
+
+def _pool_worker_initializer(shared_session_cache_dir: str) -> None:
+    """Point the worker's session-lifetime build cache at the main process's temp dir.
+
+    Each worker would otherwise create its own ``TemporaryDirectory`` and scrub it
+    on exit — taking the compiled artifacts with it before main can ``dlopen`` them.
+    """
+    import pathlib
```
havogt: why local?

src/gt4py/next/fingerprinting.py
```diff
@@ -451,42 +451,82 @@ def make_deconstructor(
...
     """
-
-    return eve_utils.singledispatcher(
+    deconstructor = eve_utils.singledispatcher(
         fallback,
         implementations={**_COMMON_DECONSTRUCTORS, **(overrides or {})},
     )
+    if name is not None:
+        deconstructor.__name__ = deconstructor.__qualname__ = name
+    if module is not None:
```
egparedes, line 501: I was literally suggesting to copy the `collections.namedtuple()` implementation here (https://github.com/python/cpython/blob/17b086917326ba36d36ffe20d04a9995c39a0778/Lib/collections/__init__.py#L522) unless the caller's context is not usually expected to be the final module of the deconstructor object

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,272 @@
...
+        )
+
+    @staticmethod
+    def _executor_blob_for_offload(job: CompileJob) -> tuple[bytes | None, str]:
+        if job.offload is None:
+            return None, "it does not use the standard compilation workflow (customized 'compile')"
+        try:
+            return pickle.dumps(job.offload.executor), ""
+        except Exception as error:  # pickling arbitrary object graphs raises arbitrary errors
+            return None, f"its executor is not picklable ({error!s})"
```
egparedes: AFAIK this function is overengineering the implementation for no real reason since it's only called once. I would inline it in the caller side. Additionally, staticmethod is deprecated in our coding guidelines

egparedes (commented):
A couple of comments

egparedes (approved):
LGTM

src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -133,6 +133,13 @@ def set_dace_config(
         dace.Config.set("compiler.cuda.backend", value="hip")
     elif device_type == core_defs.DeviceType.CUDA:
         dace.Config.set("compiler.cuda.backend", value="cuda")
+        if cuda_archs := os.environ.get("CUDAARCHS", "").strip():
+            # DaCe's CMake otherwise compiles and runs a probe on the local GPU —
+            # creating a CUDA context in the compiling process — to determine the
+            # architecture; when no device is visible (e.g. in a process-pool
+            # worker) the probe falls back to this list.
+            # TODO(havogt): we should move DaCe config to accept a cuda_arch without detection.
+            dace.Config.set("compiler.cuda.cuda_arch", value=cuda_archs.replace(";", ","))
```
edopao: Should be fixed by #2685

edopao: #2685 is now merged

havogt: updated

src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -74,15 +74,11 @@ def set_dace_config(
...
-        if device_arch:
-            dace.Config.set(
-                "compiler.extra_cmake_args",
-                value=f"-DLOCAL_CUDA_ARCHITECTURES={device_arch}",
-            )
+    if device_arch := gtx_compilation_common.get_device_arch():
+        dace.Config.set(
+            "compiler.extra_cmake_args",
+            value=f"-DLOCAL_CUDA_ARCHITECTURES={device_arch}",
+        )
```
edopao, line 81: Nice, I was about to suggest it!

src/gt4py/next/otf/compilation_tasks.py
```diff
@@ -0,0 +1,148 @@
+# GT4Py - GridTools Framework
```
havogt, line 1: The current solution is cleaner for this codebase — I'd keep it. Reasoning:

The custom-pickler route has a practical blocker and an architectural cost. You can't hand ProcessPoolExecutor a custom pickler: its queue uses multiprocessing.reduction.ForkingPickler, so interception means either monkeypatching that globally (process-wide side effect on all multiprocessing traffic — a non-starter) or pre-pickling the compilable to bytes ourselves with a Pickler subclass and shipping the blob, like we already do for the executor. The latter works, but then either the runner regains domain knowledge (it must know which pickler to use — we just spent effort making it domain-free), or OffloadableWork carries opaque bytes instead of an inspectable compilable. There's also a fiddly detail: reducer_override intercepts by type, and Connectivity is a non-runtime-checkable Protocol, so the check needs concrete-class knowledge anyway.

What the pickler would genuinely buy — and why it's not worth it today. Its one real advantage is structural completeness: it catches a connectivity anywhere in the object graph, while the substitution only knows about offset_provider. But offset_provider is currently the only sanctioned place for buffers in a compile job, and the failure mode for a stray buffer elsewhere is loud (the exact cudaErrorNoDevice we saw, or a pickle error), not silent corruption. Paying the pickler's architecture for insurance against a case that fails loudly is a bad trade.

The substitution matches the layering we built. "What may cross the process boundary" is a policy decision, and compile_jobs.py is precisely the layer that owns it (that's the module's stated purpose). The policy is expressed declaratively in ~60 lines, visible in one place, unit-testable by round-tripping a ref — and trivially deletable, which matters because this whole mechanism is a stopgap until CompileTimeArgs.offset_provider becomes type-only.

The trigger to revisit: if connectivity values start appearing outside offset_provider — e.g. static connectivity-typed program arguments — the graph-walking pickler becomes the right tool, because per-location substitution stops scaling. I'd note that in the PR discussion rather than build it now.

One aesthetic wrinkle in the current design worth acknowledging if a reviewer pokes at it: the offload copy's offset_provider holds _ConnectivityFileRefs where the type says connectivities — a deliberate lie, contained because nothing main-side reads the offload copy. If that ever bothers us, the fix is typing OffloadableWork.compilable as "pickle-ready payload" rather than switching to a pickler.

ci/cscs-ci.yml
```diff
@@ -99,6 +101,11 @@ test_cscs_gh200:
         # TODO: investigate why the dace tests seem to hang with multiple jobs
         GT4PY_BUILD_JOBS: 1
         SLURM_TIMELIMIT: "00:15:00"
+    - if: $SUBPACKAGE == 'next' && $VARIANT == 'internal' && $DETAIL == 'nomesh'
+      variables:
+        # runs within seconds of the 5-minute default since compilation moved
+        # to worker processes
```
havogt:
```suggestion
```

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,294 @@
...
+from typing import Any, Callable, Protocol, runtime_checkable
+
+from gt4py._core import definitions as core_defs
+from gt4py.next import config
+from gt4py.next.otf import definitions, stages, workflow
+from gt4py.next.otf.compilation import cache as _cache, common as compilation_common
+
+
+@dataclasses.dataclass(frozen=True)
+class OffloadableWork:
```
egparedes: Just a suggestion to align the naming of related concepts in this module:
```suggestion
class CompilationTask:
```

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,294 @@
...
+
+    #: The already-lowered program; pickle-safe by construction.
+    compilable: definitions.CompilableProgramDef
+    #: The backend's artifact-producing step. A runner may execute it anywhere,
+    #: including another process; its picklability is the runner's concern.
+    executor: workflow.Workflow[definitions.CompilableProgramDef, stages.CompilationArtifact]
+
+
+@dataclasses.dataclass(frozen=True)
+class CompileJob:
```
egparedes: Just a suggestion to align the naming of related concepts in this module:
```suggestion
class LoadTask:
```
and then:
```
offload: OfffloadableWork  -> compilation: CompilationTask
```
Since I think this task is basically loading the compilation artifact generated by the internal compilation task.

src/gt4py/next/otf/compilation_runner.py
```diff
@@ -0,0 +1,294 @@
...
+    #: Compiles in the current thread; always valid.
+    run: Callable[[], stages.ExecutableProgram]
+    #: The decomposed standard workflow, for runners that can execute it in
+    #: another process; ``None`` when the backend customizes ``compile`` and
+    #: the job can only run as-is.
+    offload: OffloadableWork | None = None
+
+
+@runtime_checkable
+class CompilationRunner(Protocol):
```
egparedes:
```suggestion
class Runner(Protocol):
```

src/gt4py/next/otf/runners.py
```diff
@@ -0,0 +1,294 @@
...
+    Points the worker's session-lifetime build cache at the main process's temp
+    dir: each worker would otherwise create its own ``TemporaryDirectory`` and
+    scrub it on exit — taking the compiled artifacts with it before main can
+    ``dlopen`` them.
+
+    When the CUDA architectures are known, hides all GPUs from the worker and
+    exports ``CUDAARCHS`` instead: GPU builds otherwise probe the device for its
+    architecture (gt4py's cmake build system via cupy, dace's CMake via a
+    ``try_run`` binary), each probe creating a CUDA context on the GPU of the
+    parent — which may be running kernels concurrently.
```
egparedes, line 130: Maybe add a TODO if it makes to remove part of these workarounds after merging some PRs in  DaCe

src/gt4py/next/otf/runners.py
egparedes, line 1: Revise the name of the file if the naming suggestions in my review are accepted

src/gt4py/next/otf/compilation_tasks.py
egparedes, line 1: Revise the name of the file if the naming suggestions in my review are accepted. Even in the opposite case, I think this should renamed anyway (maybe something like `...utils.py`).

src/gt4py/next/otf/compilation_tasks.py
```diff
@@ -0,0 +1,156 @@
...
+        return (
+            _connectivity_from_file,
+            (self.path, self.domain, self.codomain, self.skip_value),
+        )
+
+
+#: Files already written for a connectivity, keyed by its id (validated against
+#: a weakref, since ids can be reused): the same mesh is typically shared by all
+#: programs of a run and must be dumped only once.
+_connectivity_files: dict[int, tuple[weakref.ref, str]] = {}
```
egparedes, line 77: When is this per-process cache used? Only by the thread pool? Or is it possible to require the same connectivity several times in the same compilation job?

egparedes (commented):
Looks good. Just need some clarifications

src/gt4py/next/otf/runners.py
havogt: So you would go with runner.py? Or keep to have `compilation_runner.Runner` instead of a generic `runner.Runner`?

src/gt4py/next/otf/compile_jobs.py
```diff
@@ -0,0 +1,156 @@
...
+#: Files already written for a connectivity, keyed by its id (validated against
+#: a weakref, since ids can be reused): the same mesh is typically shared by all
+#: programs of a run and must be dumped only once.
+_connectivity_files: dict[int, tuple[weakref.ref, str]] = {}
+
+
+def _connectivity_file_ref(value: common.Connectivity) -> _ConnectivityFileRef:
+    entry = _connectivity_files.get(id(value))
+    if entry is None or entry[0]() is not value:
+        dump_dir = compilation_cache._session_cache_dir_path / "connectivities"
```
havogt: Is this a good strategy? To use the path of a potential session cache for this offloading? Why not create a separate tmp directory here where we can control the location?

src/gt4py/next/otf/compilation_tasks.py
havogt: see extended documentation

src/gt4py/next/otf/runners.py
havogt: done

src/gt4py/next/otf/load_tasks.py
```diff
@@ -0,0 +1,171 @@
...
+        # pickle queue would serialize the full tables once per task. The lazy
+        # file references keep this pure — nothing is dumped unless a runner
+        # actually pickles the task.
+        offload_compilable = dataclasses.replace(
+            compilable,
+            args=dataclasses.replace(
+                compilable.args,
+                offset_provider=_offset_provider_with_file_refs(compilable.args.offset_provider),
+            ),
+        )
```
havogt: can we make this a function `def construct_offload_compilable(with_refs: bool)` then the pool's submit can select mode.

src/gt4py/next/otf/load_tasks.py
```diff
@@ -0,0 +1,171 @@
...
+    return runners.LoadTask(
+        name=name,
+        compile_and_load=lambda: backend.executor(compilable).load(),
+        compilation=runners.CompilationTask(
+            compilable=offload_compilable, executor=backend.executor
+        ),
```
havogt: can we ship `compilation` and `load` in some way which is used in all runner implementations.

src/gt4py/next/otf/runners.py
```diff
@@ -0,0 +1,304 @@
...
+
+        def _load(
+            artifact_future: concurrent.futures.Future[stages.CompilationArtifact],
+        ) -> None:
+            try:
+                loaded.set_result(artifact_future.result().load())
+            except BaseException as exception:  # re-raised via the future
+                loaded.set_exception(exception)
+
+        artifact_future.add_done_callback(_load)
```
havogt: Can we do the load in all cases outside of the runner's submit? i.e. the runner doesn't return the ExecutableProgram, but the Artifact.

egparedes (approved):
LGTM

src/gt4py/next/otf/compilation/common.py
```diff
@@ -10,8 +10,56 @@
...
+    # an early one succeeded.
+    if core_defs.CUPY_DEVICE_TYPE == core_defs.DeviceType.CUDA:
+        # use `cp` from core_defs to avoid trying to re-import cupy
+        try:
+            return core_defs.cp.cuda.Device(0).compute_capability  # type: ignore[attr-defined]
+        except core_defs.cp.cuda.runtime.CUDARuntimeError as e:  # type: ignore[attr-defined]
+            warnings.warn(
+                UserWarning(f"Could not determine the CUDA compute capability: {e}"), stacklevel=2
+            )
+            return None
```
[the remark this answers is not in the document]
havogt: No, https://docs.cupy.dev/en/stable/reference/generated/cupy.cuda.Device.html#cupy.cuda.Device.compute_capability

havogt:
last change was only docstring cleanup. Will merge once code quality passed.

PR #2595, opened 2026-05-08
feat/fix[cartesian]: Integrate `dace.` derived math functions to match Fortran

romanc:
## Description

**Update** This branch is clean except for the original math functions. DaCe math functions require https://github.com/spcl/dace/pull/2418 to be merged first and then DaCe to be released such that it can be updated in GT4Py (to the newest alpha).

:bricks: In short, this is blocked by https://github.com/spcl/dace/pull/2418 and a subsequent DaCe release.

---

:dragon: :warning:
Temporary working branch of `gt4py.cartesian` for the NASA crew. To be properly integrated in July.

Several math functions, e.g. `log10(x)`, do not respect `GT4Py_LITERAL_FLOAT_PRECISION` in the dace backend.

The issue is caused by a combination of dace's code generation and the C/C++ standard library history. In C, `log10(x)` is defined to operate in double precision. The single-precision equivalent is called `log10f(x)`, [see here](https://en.cppreference.com/c/numeric/math/log10).

In C++, there is `std::log10(x)`, which is a templated version that computes in either single- or double-precision depending on the argument. DaCe exposes access to those templated versions through `dace/math.h`. Some function were missing and are being patched with https://github.com/spcl/dace/pull/2364. Building on that PR in DaCe, this PR uses the newly exposed fuctions in the cartesian DaCe backends.

## Requirements

- [x] All fixes and/or new features come with corresponding tests.
- [ ] Important design decisions have been documented in the appropriate ADR inside the [docs/development/ADRs/](docs/development/ADRs/README.md) folder.
  N/A

PR #2682, opened 2026-07-02
perf[cartesian]: avoid `x + 0` and `0 + x` expression in index bounds

romanc:
## Description

In oir -> treeir, avoid inserting artificial `x + 0` or `0 + x` calculations when translating axis bounds. In the end, this makes for cleaner generated code, which is easier to read and doesn't rely on general compilers to remove the calculation as part of their optimization pipeline.

This PR is part of cleaning up our "working branch", i.e. https://github.com/GridTools/gt4py/pull/2595, which accumulated a couple of changes over the last month.

## Requirements

  N/A


twicki (approved):
That works! Even covering for the niece end=0 case

PR #2684, opened 2026-07-02
fix[cartesian]: actually freeze origin & domain in K-axis

romanc:
## Description

This fix comes from running GFDL_1M in KJI-layout where - in NDSL - there's no extra point allocated in the vertical (to be able to directly copy the Fortran data).

This commit freezes origin and `domain` (not `shape`) in the K-axis of the stencil. `shape` and `domain` agree most of the time, but can differ if - for example - the stencil is configured on `K_INTERFACE_DIM` and the field of size `K_DIM`. Depending on the combination, this can lead to out of bound reads where we were previously saved by the extra point allocated in the back (but in KJI since we don't have this point, we crash).

I am not touching the horizontal axes in this commit. We probably should, but we haven't seen any crashes (probably due to the fact that we can't currently run dynamics in KJI (i.e. without the extra point)). I'd argue we should tackle this in a follow-up when we also look at the "potential out of bounds" memlet warnings that we get from dace (mainly when running dynamics).

This PR is part of cleaning up our "working branch", i.e. https://github.com/GridTools/gt4py/pull/2595, which accumulated a couple of changes over the last month.

## Requirements

  N/A
  N/A

### DIFF
--- a/src/gt4py/cartesian/backend/dace_backend.py
+++ b/src/gt4py/cartesian/backend/dace_backend.py
@@ -95,6 +95,7 @@ def _sdfg_add_arrays_and_edges(
     inputs: set[str] | dict[str, dtypes.typeclass],
     outputs: set[str] | dict[str, dtypes.typeclass],
     origins: dict[str, tuple[int, ...]],
+    domain: tuple[int, ...],
 ) -> None:
     for name, array in inner_sdfg.arrays.items():
         if array.transient:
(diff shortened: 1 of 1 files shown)


twicki (approved):
This is hard to parse, and we will most likely go back to clean this up when we touch the horizontal anyways. 

And we know it works, so let's merge this!

PR #2685, opened 2026-07-02
fix[next-dace]: Set cmake argument for local cuda architecture

edopao:
Skip auto-discovery of cuda device architecture, by default enabled in the dace cmake configuration, and rely on the device architecture provided by GT4Py.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/common.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/common.py
@@ -72,10 +72,11 @@ def set_dace_config(
     dace.Config.set("cache", value="single")
 
     # Disable auto-detection of the CUDA architecture, and instead use the one provided by GT4Py.
-    dace.Config.set(
-        "compiler.extra_cmake_args",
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -70,6 +71,13 @@ def set_dace_config(
     #   creating any further sub-folder to compile the SDFG.
     dace.Config.set("cache", value="single")
 
+    # Disable auto-detection of the CUDA architecture, and instead use the one provided by GT4Py.
```
edopao:
```suggestion
    # Workaround to disable detection of the CUDA architecture in DaCe, and instead use the one provided by GT4Py.
    # TODO(edopao): revisit this workaround once it is possible to disable GPU detection in DaCe.
    # (see https://github.com/spcl/dace/pull/2424)
```

philip-paul-mueller (approved):
As a temporary fix LGTM.

PR #2691, opened 2026-07-03
fix[next]: make OTF build caches crash-consistent

havogt:
An interrupted pipeline (Ctrl-C, SIGKILL/OOM, full disk) can leave a cache entry that a lookup reports as a hit but whose payload is truncated or missing; the next run then fails on load instead of rebuilding. Fingerprinting (#2648) made cache keys correct but says nothing about payload completeness.

This PR makes cache writes atomic (temp sibling + `os.replace`, new helper `gt4py._core.file_utils`) and cache reads validating + self-healing (any load failure is treated as a miss and rebuilt). No fsync: process kill is covered by the atomic rename, a power-loss torn write by validate-on-read. Rationale and alternatives in ADR 0024 (part of this PR).

Fixed, each with a test:

- truncated translation pickle (FileCache): evicted and recomputed instead of aborting with an unpickling error
- truncated gt4py.json (gtfn): treated as "not built" and rebuilt
- status COMPILED but module artifact deleted (gtfn): rebuilt instead of raising CompilationError
- truncated compiled SDFG library (dace): the compile step now writes a completion marker last; a library without the marker is dropped so dace rebuilds it
- truncated compiledb template (shared by all programs of a configuration): published atomically and validated on lookup

The existing partial-build resume behavior of the gtfn status machine is preserved.

Known gap (documented in the ADR): a gtfn module truncated after a completed build (power-loss torn write) is not self-healed.


src/gt4py/_core/file_utils.py
egparedes, line 1: Maybe it would be more appropriate to rename it as file_utils.py or atomic_io.py, ...

egparedes (commented):
Looks good to me. How does this PR interact with #2679 ?

src/gt4py/_core/file_utils.py
havogt: Renamed to `file_utils.py` in 811dec48a.

havogt:
claude: @egparedes regarding #2679: they are disjoint and complementary. No file overlap (verified with `git merge-tree`: clean merge in either order). Semantically, #2679 moves compilation into worker processes, but all mechanisms here are file-based (locks, atomic publish, completion marker) and therefore process-agnostic — the compile step behaves identically in a worker. If anything, #2679 makes this PR more relevant: a worker killed mid-compile is exactly the interruption this PR recovers from. The pool's own temp files (connectivity shipping buffers) are write-once IPC, not a persistent cache, and need no such treatment.

src/gt4py/_core/file_utils.py
edopao, line 1: This file is so generic that could belong to `eve`

egparedes (approved):

src/gt4py/_core/file_utils.py
```diff
@@ -0,0 +1,46 @@
...
+"""Helpers for atomic file writes."""
+
+from __future__ import annotations
+
+import os
+import pathlib
+import uuid
+
+
+def atomic_write_bytes(target: str | os.PathLike, data: bytes) -> None:
```
havogt, line 18: Can we unit test this?

src/gt4py/_core/file_utils.py
havogt: Let's keep it here. It's next to locking.py and filecache.py. We could move all 3 together to eve (or go with the structure that Enrique presented recently).

PR #2698, opened 2026-07-10
fix[next]: propagate recursion limit to compilation worker processes

havogt:
## Description

`sys.setrecursionlimit` is per-process: a limit the parent raised for deeply nested IR silently resets to the interpreter default in a `spawn` worker, so compiles that succeed in-process (or under the thread/serial runners) fail with `RecursionError` under the process runner introduced in #2679.

Observed in icon4py CI after bumping to gt4py main (C2SM/icon4py#1350): the large advection stencils set `sys.setrecursionlimit(5500)` at module import, which never reaches the workers; the deep eve traversals then blow past the default 1000 frames worker-side, surfacing as `RecursionError` re-raised from the compilation future.

Fix: capture the parent's recursion limit at submit time and adopt it in the worker entry point, never lowering the worker's own limit. Captured per task (not at pool creation) because user code may raise the limit after the pool exists.

The unit test reproduces the failure signature exactly before the fix (recursion succeeding in the parent, `RecursionError` from the worker future).

## Requirements

- [ ] Important design decisions have been documented in the appropriate ADR inside the [docs/development/ADRs/](docs/development/ADRs/README.md) folder. (No new design decision: restores the "compile behaves identically in a worker" contract of ADR 0024.)

### DIFF
--- a/src/gt4py/next/otf/runners.py
+++ b/src/gt4py/next/otf/runners.py
@@ -17,6 +17,7 @@ import multiprocessing
 import os
 import pathlib
 import pickle
+import sys
 import threading
 import warnings
 from typing import Any, Callable, Protocol, runtime_checkable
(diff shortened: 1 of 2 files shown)


havogt:
I guess if we find more things like this we are re-considering the fork @tehrengruber ;)

tehrengruber (approved):

src/gt4py/next/otf/runners.py
```diff
@@ -167,8 +168,13 @@ def _run_compilation_task_in_worker(
     executor_blob: bytes,
     compilable: Any,
     config_overrides: dict[str, Any],
+    recursion_limit: int,
 ) -> stages.CompilationArtifact:
     """Worker entry point: deserialize the executor and run it."""
+    # `sys.setrecursionlimit` is per-process: a limit the parent raised for
+    # deeply nested IR would silently reset to the interpreter default in a
+    # `spawn` worker. Adopt the parent's limit, but never lower the worker's.
```
edopao, line 176: > Adopt the parent's limit, but never lower the worker's.

I was thinking about "never lower the worker's". What if it was in the parent intention (e.g. a unit test) to lower that limit?

havogt: I'll merge now, if we have a concrete case we can drop it later.

PR #2703, opened 2026-07-10
tests[cartesian]: add parsing test for ddims size with numpy int type

### LINKED ISSUE #2705: cartesian: allow data dimension sizes of type other than just plain `int`
## Description

GT4Py IR validation (e.g. nodes, gtir, ...) expect the size of data dimensions to be of type `int`. This is a bit too restrictive and will fail in case of e.g. `numpy.int32` types. We _should_ be able to remedy this by checking for `numbers.Integral` as the base type of all integral numbers in python. This will work with both, standard `int` and `numpy.int32` types. Before we can go there, we'll need to go down to DaCe and do something similar for the validation of the `shape` property of a `Data` class.

The current workaround is simply to manually cast to `int` when defining data dimensions.

PR https://github.com/GridTools/gt4py/pull/2703 adds a simple test case which is marked as `xfail` for now until we have proper support for this.

Related issues:

- upstream: https://github.com/NOAA-GFDL/NDSL/issues/500
- downstream: https://github.com/spcl/dace/issues/2431

romanc:
## Description

GT4Py IR validation (e.g. nodes, gtir, ...) expect the size of data dimensions to be of type `int`. This is a bit too restrictive and will fail in case of e.g. `numpy.int32` types. We _should_ be able to remedy this by checking for `numbers.Integral` as the base type of all integral numbers in python. This will work with both, standard `int` and `numpy.int32` types. Before we can go there, we'll need to go down to DaCe and do something similar for the validation of the `shape` property of a `Data` class.

Related issues:

- tracking issue for the real fix: https://github.com/GridTools/gt4py/issues/2705
- upstream: https://github.com/NOAA-GFDL/NDSL/issues/500
- downstream: https://github.com/spcl/dace/issues/2431

## Requirements

  N/A

### DIFF
--- a/tests/cartesian_tests/unit_tests/frontend_tests/test_gtscript_frontend.py
+++ b/tests/cartesian_tests/unit_tests/frontend_tests/test_gtscript_frontend.py
@@ -1765,8 +1765,9 @@ class TestAssignmentSyntax:
 
             parse_definition(func, name=inspect.stack()[0][3], module=self.__class__.__name__)
 
-    def test_datadims_direct_access(self):
-        # Check classic data dimensions are working
+
+class TestGlobalTablesWithDataDimensions:
+    def test_classic_data_dimensions_access(self) -> None:
         def data_dims(
             out_field: gtscript.Field[gtscript.IJK, np.int32],
             global_field: gtscript.Field[(np.int32, (3, 3, 3))],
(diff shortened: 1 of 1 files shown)


twicki (approved):
Looks good, thanks for extending / cleaning up the test-suite

PR #2704, opened 2026-07-10
fix[next-dace]: Ignore `guid` in SDFG fingerprint

edopao:
The source code generated by the translation stage should be byte identical, for the same input IR program. In case of lowered SDFGs, the JSON representation produced by the translation stage contains `guid`s which are not deterministic. These IDs should be excluded from the string containing the JSON source code, in order not to influence the DaCe compilation cache key.
The random `guid`s could be a problem in multi-rank deployment: multiple ranks could try compiling the same gt4py program. This PR allows the ranks to produce exactly the same compilable program, thus using the same cache key and compilation artifact.

Additional change: remove `build_folder` from `DaCeCompilationArtifact`, because not used.

Co-authored-by: Hannes Vogt

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/translation.py
@@ -474,7 +474,17 @@ class DaCeTranslator(
 
 
 def _drop_element_ids(json_obj: Any) -> Any:
-    """Recursively remove `guid` keys from a serialized SDFG."""
+    """
+    Remove ``guid`` keys from a serialized SDFG with recursion.
+
+    We remove ``guid`` keys because of indeterminism of the ``guid`` values, which
+    would cause two identical programs to result in different compiled SDFGs, since
+    the cache key contains the hash of the serialized SDFG (a JSON string).
+    # FIXME(edopao): remove this workaround once the SDFG lowering is stable.
+
+    Note that this recursive implementation is 2-3x faster than the iterative one,
+    on a large SDFG, and it is also simpler to read.
+    """
     if isinstance(json_obj, dict):
         return {k: _drop_element_ids(v) for k, v in json_obj.items() if k != "guid"}
     if isinstance(json_obj, list):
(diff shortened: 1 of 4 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -462,13 +462,38 @@ def __call__(
             # Set 'hash=True' to compute the SDFG hash and store it in the JSON.
             #   We compute the hash in order to refresh `cfg_list` on the SDFG,
             #   which makes the JSON serialization stable.
-            source_code=sdfg.to_json(hash=True),
+            # `guid` is a per-element identity token: it does not affect code generation, and
+            #   `SDFG.from_json()` assigns fresh ids anyway. Keeping it would make the compile
+            #   cache key depend on element creation order, so two structurally identical
+            #   lowerings of the same program would not share a cached build.
+            source_code=_drop_element_ids(sdfg.to_json(hash=True)),
```
tehrengruber, line 469: Maybe the problem is not that the "source code" contains the indeterministic UUIDs, but more that we try to force the SDFG into an `ProgramSource` object and rely on the string representation of the `source_code` to determine the cache folder path. Since the computation of the sdfg hash already contains the functionality to remove the GUID it feels like the hash is the right way to determine the path instead of duplicating part of the functionality here. Maybe we should investigate refactoring `stages.ProgramSource` into something that allows us to:
1. store object other than strings, since it doesn't really make sense to serialize and de-serialize the sdfg
2. compute a stable hash based on the sdfg hash

tehrengruber (approved):
Fine with me, but for a real cleanup we might want to take a different direction. In particular the reliance on yet another mechanism to create a stable-across-runs-representation of an SDFG doesn't feel quite right.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
edopao: I agree. I quickly tried to achieve 2 by specifying the source fingerprinter in `SourceCodeSpec` but I see that it also requires 1 (I cannot call `hash_sdfg` on a serialized SDFG). On the other hand, we will no longer have this issue once the SDFG lowering is stable. We can have a discussion.

PR #2706, opened 2026-07-14
refactor[cartesian]: use f-strings instead of `.format()`

romanc:
## Description

Use f-strings instead of `.format()` for simple text concatenation where e.g. the template is not shared. f-strings are shorter and offer almost the same amount of flexibility. In these simple cases, it's imo much more readable.

This all started with Charles hitting a non-boolean `if` condition yesterday, which gives a very cryptic and non-informative error message. While expanding the error message, I noticed in how many place we use `.format()` when simple f-strings would be sufficient. So I changed that too ... 

## Requirements

  N/A
  N/A

### DIFF
--- a/src/gt4py/cartesian/gtc/common.py
+++ b/src/gt4py/cartesian/gtc/common.py
@@ -279,7 +279,9 @@ class Stmt(LocNode):
 
 def verify_condition_is_boolean(parent_node_cls: datamodels.DataModel, cond: Expr) -> None:
     if cond.dtype and cond.dtype is not DataType.BOOL:
-        raise ValueError("Condition in `{}` must be boolean.".format(type(parent_node_cls)))
+        raise ValueError(
+            f"Condition in `{type(parent_node_cls)}` must be boolean, got {cond.dtype} instead. Condition: {cond}."
+        )
 
 
 def verify_and_get_common_dtype(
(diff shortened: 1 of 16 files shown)


src/gt4py/cartesian/gtc/common.py
```diff
@@ -279,7 +279,9 @@ class Stmt(LocNode):
 
 def verify_condition_is_boolean(parent_node_cls: datamodels.DataModel, cond: Expr) -> None:
     if cond.dtype and cond.dtype is not DataType.BOOL:
-        raise ValueError("Condition in `{}` must be boolean.".format(type(parent_node_cls)))
+        raise ValueError(
+            f"Condition in `{type(parent_node_cls)}` must be boolean, got {cond.dtype} instead. Condition: {cond}."
+        )
```
romanc, line 284: This is where it all started: Non-boolean conditions would return a rather cryptic error message with little information. While the error message is still cryptic, at last now it contains more information such that you can actually understand what's wrong.

src/gt4py/cartesian/utils/meta.py
```diff
@@ -74,9 +74,7 @@ def _dump(node: ast.AST) -> str:
             return "".join(
                 [
                     node.__class__.__name__,
-                    "({content})".format(
-                        content=", ".join("{}={}".format(name, value) for name, value in fields)
-                    ),
+                    f"({', '.join('{}={}'.format(name, value) for name, value in fields)})",
```
twicki: is there a reason for this nested `.format` to be kept?

twicki (commented):
Looks good, just one question on one

src/gt4py/cartesian/utils/meta.py
romanc: I don't think so. I've added a new commit to fix this.

twicki (approved):
Looks good, thanks

PR #2709, opened 2026-07-15
fix[next]: keep scan compute domain within the statement colum

SF-N:
A scan computes its entire vertical column sequentially and cannot read its inputs outside that column. When a scan's result is read at a vertical offset — e.g. a backward scan reading a forward scan's result at `Koff[-2]` — domain inference back propagated that offset and extended the scan's compute domain (to `[-2, 11)` instead of `[0, 11)`). The scan then reads its inputs out of bounds.

This was fixed by making `_infer_stmt` now derive the statement's vertical column and thread it (`scan_vertical_bounds`) through `infer_expr`. When a scan's vertical dimension is re-added to its domain, the range is taken from that column rather than from the (offset-extended) back-propagated domain, so every scan in a `SetAt` is pinned to the statement column. All fields written by a `SetAt` share one vertical column, so any concrete target domain is representative of it.

A new unit test `test_scan_result_vertical_offset`  with a fused forward/backward scan program where the backward scan reads the forward result at `Koff[-2]` was added.

### DIFF
--- a/src/gt4py/next/iterator/transforms/infer_domain.py
+++ b/src/gt4py/next/iterator/transforms/infer_domain.py
@@ -12,7 +12,7 @@ import typing
 
 from gt4py import eve
 from gt4py.eve import utils as eve_utils
-from gt4py.eve.extended_typing import Callable, Optional, TypeAlias, Unpack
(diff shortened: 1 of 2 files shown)


SF-N:
Closing for now and opening issue https://github.com/GridTools/gt4py/issues/2713 instead.

PR #2650, opened 2026-06-11
fix[next]: Add build configuration to the compilation fingerprint

edopao:
**Description**
The compiled artifacts (build folders, SDFGs) were keyed only by a hash of the program source. As a result, changing the backend configuration — in particular DaCe's `dace.Config` or the builder factory for `gtfn`backends — did not invalidate the cache, so a stale artifact could be reused even though the configuration that produced it had changed.

This PR includes the relevant build-time configuration into the cache-folder fingerprint so artifacts are correctly rebuilt when the configuration changes.

To make it possible, we added an optional `build_context_id` to `get_cache_folder()` which allows to distinguish different build contexts that produce different artifacts for the same extension source. 

**Changes to DaCe backend**
`DaCeCompiler` captures the non-default `dace.Config `values at construction and includes them in the build context fingerprint, so an SDFG is rebuilt when the user changes the backend configuration.

**Changes to GTFN backend**
The `builder_factory` is fingerprinted into the above-mentioned `build_context_id`. Note that `builder_factory`  specifies the cmake configuration.

PR #2714, opened 2026-07-22
fix[next]: make the DaCe determinism check follow gt4py's cache folder naming

kotsaloscv:
## Description

The format of the SDFG build folder name changed in #2650: the fingerprint became a 16-hex xxhash64 and the name gained `_{version_id}` and `_{build_context_id}` suffixes (e.g. `__field_operator_solve_tridiag_pyext_ebf83e9be18232fd_1.1.12_db5ed39d9f2c69fb`). The dace-determinism script still expected `<name>_<64-hex-sha256>`, so the standalone `dace-determinism` CI pipeline failed every run with "no programs observed" (exit 3).

This fixes the breakage and makes the coupling robust against future renames by single-sourcing the naming scheme in gt4py:

- `otf/compilation/cache.py` now exposes `CACHE_FOLDER_NAME_PATTERN` right next to `get_cache_folder`; a new round-trip unit test builds folder names with `get_cache_folder` and asserts the pattern parses them back, so a future change to the naming scheme fails the test in the same PR that makes it.
- `scripts/python/dace_determinism.py` always reads that pattern from a gt4py installation at runtime: `ci-check` asks the interpreter it drives, `check` asks the interpreter running the script (which therefore needs gt4py installed, e.g. run it via `uv run python scripts/python/dace_determinism.py check ...`). There is deliberately no fallback — if no gt4py can provide the pattern, the check aborts with a clear error (exit 2) instead of guessing with a possibly stale copy.
- The script's orchestration tests drive it against a stub interpreter that advertises its own folder-naming scheme and writes its fake caches in that scheme, so they fail immediately if the pattern fetch is ever broken.

Verified locally: `ci-check` on a single dace test (`test_basic.py::test_copy[dace.run_dace_cpu]`) reproduces the CI failure on `main` (exit 3, "no subdirectory matches") and reports `codegen deterministic` with this change (exit 0).

### DIFF
--- a/scripts/python/dace_determinism.py
+++ b/scripts/python/dace_determinism.py
@@ -14,10 +14,11 @@ gt4py.next caches each compiled program at ``<root>/<folder>/src/...``, where
 the folder name embeds the program name and a fingerprint of the generated
 source. The folder-name layout is owned by gt4py — ``CACHE_FOLDER_NAME_PATTERN``
 next to ``get_cache_folder`` in ``gt4py/next/otf/compilation/cache.py``, kept in
-sync by a round-trip test there — and the ``ci-check`` subcommand reads it from
-the gt4py installation it drives, so this script follows folder-naming changes
-automatically. The ``check`` subcommand uses a built-in copy of the pattern,
-overridable with ``--folder-pattern``.
+sync by a round-trip test there — and this script always reads it from a gt4py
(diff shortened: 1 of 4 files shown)


kotsaloscv:
cscs-ci run dace-determinism

kotsaloscv:
cscs-ci run dace-determinism

edopao (approved):
LGTM

PR #2721, opened 2026-07-27
DO NOT MERGE feat[next-dace]: Use `nanobind` Interface of DaCe

philip-paul-mueller:
It is not fast only to see if it works.


src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1018,7 +1018,11 @@ def _add_sdfg_params(
                 transient=True,
             )
 
-        # the list of all sdfg arguments (aka non-transient arrays) which include tuple-element fields
+        # The list of the SDFG "arguments", i.e. the non transient arrays. Note that
+        #  tuple arguments are flatten and their name is mangled and no longer matches
+        #  the name in the signature of the field operator / program. Furthermore,
+        #  after optimization they might no longer be needed and are removed from
```
edopao: > after optimization they might no longer be needed and are removed

This only applies to scalar arguments, because they are represented as SDFG symbols. Unused symbols are pruned.
There were some changes over time, back and forward, whether to use scalars or symbols. The last reason for using symbols was that the gt4py toolchain was (is?) producing tuples of scalars for field domain range `[prefix]_range_0, [prefix]_range_1`. These scalars are used in domain expressions, therefore in memlet subsets and map ranges.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -105,91 +84,39 @@ def __init__(
...
 
     def __call__(self, **kwargs: Any) -> None:
         """Call the compiled SDFG with the given arguments.
 
-        Note that this function will not update the argument vectors stored inside
-        `self`. Furthermore, it is not recommended to use this function as it is
-        very slow.
+        Note that this function will not use the user argument entry point. Furthermore,
+        using it requires exact knowledge of the SDFG argument names. In short: using it
+        is most certainly an error.
```
edopao: This text is not very clear. Besides, the line below makes me wonder why the function is needed?

> In short: using it is most certainly an error.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -217,7 +144,8 @@ class DaCeCompiler(
     cmake_build_type: config.CMakeBuildType = dataclasses.field(
         default_factory=lambda: config.CMAKE_BUILD_TYPE
     )
-    # we store the non-default values of `dace.Config` in order to include it in the stage fingerprint
+    # We store the non-default values of `dace.Config` in order to include it in the stage fingerprint
+    # NOTE: They do not include the non default keys set through environment variables.
```
edopao:
```suggestion
    # NOTE: They do not include the non default keys set through DaCe environment variables.
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -232,6 +160,7 @@ def __call__(self, inp: SDFGExtensionSource) -> DaCeCompilationArtifact:
             device_type=self.device_type,
             cmake_build_type=self.cmake_build_type,
         ):
+            # TODO(phimuell): Extraction of the binding code is missing.
```
edopao: Is this TODO still needed?

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -48,33 +43,13 @@ def decorated_program(
...
-            )
-            this_call_args |= {
-                gtx_wfdcommon.SDFG_ARG_METRIC_LEVEL: metrics.get_current_level(),
-                gtx_wfdcommon.SDFG_ARG_METRIC_COMPUTE_TIME: collect_time_arg,
-            }
-            fun.construct_arguments(**this_call_args)
-
-        # Perform the call to the SDFG.
-        fun.fast_call()
+        proc_arc = argument_preprocessing_function(
```
edopao: why `arc`? a typo? `proc_args`

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -435,6 +588,14 @@ def _generate_sdfg_without_configuring_dace(
...
             add_instrumentation(sdfg, on_gpu)
 
+        # This function must be called _after_ the optimization of the SDFG has been finished.
+        sdfg.user_args = make_user_args(
+            prog=ir,
+            sdfg=sdfg,
+            offset_provider_type=offset_provider_type,
+            use_metrics=self.use_metrics,
+        )
+
```
edopao: It seems to me that `user_args` is not used.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -8,281 +8,122 @@
...
+    backend: Literal["gtfn", "dace"],
+) -> str:
+    """
+    Creates a Python function that translates GT4Py arguments into arguments suitable for a SDFG call.
 
-    For scalar arguments, a single field is set in the SDFG arglist.
+    The returned string can be passed to exec to then dynamically create a function to
+    perform the translation. The function will have the same signature as
+    `argument_processing_function()` and have the name `bind_func_name`.
+    The translation function targets the `user_args` interface of the SDFG.
```
edopao: > The translation function targets the `user_args` interface of the SDFG.

Is this still the case?

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -292,9 +133,210 @@ def bind_sdfg(
...
+    backend: str,
+    use_metrics: bool,
+) -> Union[str, Any]:
+
+    if not use_metrics:
+        return tuple() if eval_mode else ""
+
+    # This is currently different and should probably be unified somehow.
+    if backend == "gtfn":
+        raise NotImplementedError("Implement me")
```
edopao:
```suggestion
        raise NotImplementedError()
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -292,9 +133,210 @@ def bind_sdfg(
...
+        command = f"(offset_provider['{table_name}'].ndarray, (0, 0)), "
+        processed_offset_providers += eval(command) if eval_mode else command  # type: ignore[operator]
+    return processed_offset_providers
+
+
+def _process_metric_arguments(
+    metric_level: Union[str, int],
+    compute_time_argument: Union[str, Any],
+    eval_mode: bool,
+    backend: str,
```
edopao:
```suggestion
    backend: Literal["dace", "gtfn"],
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -292,9 +133,210 @@ def bind_sdfg(
...
+    # This is currently different and should probably be unified somehow.
+    if backend == "gtfn":
+        raise NotImplementedError("Implement me")
+    else:
+        if eval_mode:
+            assert not (isinstance(metric_level, str) or isinstance(compute_time_argument, str))
+            return (metric_level, compute_time_argument)
+        else:
+            assert isinstance(metric_level, str) and isinstance(compute_time_argument, str)
+            return f"{metric_level}, {compute_time_argument}, "
```
edopao:
```suggestion
    if eval_mode:
        assert not (isinstance(metric_level, str) or isinstance(compute_time_argument, str))
        return (metric_level, compute_time_argument)
    else:
        assert isinstance(metric_level, str) and isinstance(compute_time_argument, str)
        return f"{metric_level}, {compute_time_argument}, "
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -292,9 +133,210 @@ def bind_sdfg(
...
+    processed_offset_providers: Union[str, tuple[Any, ...]] = tuple() if eval_mode else ""
+    for table_name in offset_provider:
+        command = f"(offset_provider['{table_name}'].ndarray, (0, 0)), "
+        processed_offset_providers += eval(command) if eval_mode else command  # type: ignore[operator]
+    return processed_offset_providers
+
+
+def _process_metric_arguments(
+    metric_level: Union[str, int],
+    compute_time_argument: Union[str, Any],
```
edopao: `Union[str, Any]` is equivalent to `Any`

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -292,9 +133,210 @@ def bind_sdfg(
...
+
+    if not use_metrics:
+        return tuple() if eval_mode else ""
+
+    # This is currently different and should probably be unified somehow.
+    if backend == "gtfn":
+        raise NotImplementedError("Implement me")
+    else:
+        if eval_mode:
+            assert not (isinstance(metric_level, str) or isinstance(compute_time_argument, str))
```
edopao:
```suggestion
            assert isinstance(metric_level, int)
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
edopao: Maybe a bit more clear, not so important.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -292,9 +133,210 @@ def bind_sdfg(
...
+    else:
+        assert isinstance(argument, str)
+        raise ValueError(f"Parameter `{argument}` had unexpected type `{param_type}`")
+
+    command += ", "  # Comma is intentional.
+    return eval(command) if eval_mode else command
+
+
+def _process_offset_providers(
+    offset_provider: Union[gtx_common.OffsetProvider, gtx_common.OffsetProviderType],
```
edopao: Here I am very confused. Do you need `gtx_common.OffsetProvider` or `gtx_common.OffsetProviderType`? The first has `ndarray`, the second not. Tome it seems that you need `gtx_common.OffsetProvider`.

Also, I do not understand why you are passing all connectivities, and not only the connectivities which are used (non transient arrays). This is related to another comment I have above.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -8,281 +8,122 @@
...
-    code: codegen.TextBlock,
-    sdfg_arglist: dict[str, dace.data.Data],
-) -> None:
-    """Emit Python code to parse a program argument and set the required fields in the SDFG arglist.
+def _create_sdfg_bindings(
+    prog: ir.Program,
+    offset_provider_type: gtx_common.OffsetProviderType,
+    bind_func_name: str,
+    use_metrics: bool,
+    eval_mode: bool,
```
edopao: Can we hard-code `eval_mode=True`and have a wrapper function that evaluates the entire generated string instead of evaluating piece by piece?

edopao: Also, am I wrong to say that `eval_mode=True` would allow to defer passing the `offset_provider` to call time, thus we could generate the bindings in the binding stage? You do not need to know the keys in the `offset_provider` table, all you need is the list of non-transient arrays used for connectivities. See in baseline:
```
def _parse_gt_connectivities(
    code: codegen.TextBlock, sdfg_arglist: dict[str, dace.data.Data]
) -> None:
    for sdfg_arg_index, (arg_name, arg_desc) in enumerate(sdfg_arglist.items()):
        if gtx_dace_args.is_connectivity_identifier(arg_name):
            assert isinstance(arg_desc, dace.data.Array)
            assert len(arg_desc.shape) == 2
            assert isinstance(arg_desc.shape[1], int) or str(arg_desc.shape[1]).isdigit()
```

edopao (commented):
An initial round of review. I think that the main point to discuss is the `offset_provider` argument.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
philip-paul-mueller: I reformulated it.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
philip-paul-mueller: Agreed this function does not make sense.
I ported it because it was there, but I remember that I thought about the old one "this function does not make sense".
And I also think that it can be made to work and we should simply turn it into an error.

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
philip-paul-mueller: I am not sure, looking at it I think it is outdated.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
philip-paul-mueller: Yes a typo.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
philip-paul-mueller: What do you mean by that?
`user_args` is used by the DaCe code generator, if it is defined a new function is exposed in the binding code.
Or did you mean something differently?

edopao: Oh I see, sorry. Then we need to add a comment above saying that the `user_args` is the SDFG signature required to generate the `nanobind` interface.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
philip-paul-mueller: I am actually not sure what you mean by that.
However, I have reformulated it a bit to make it clearer.

But I think where your confusion comes from.
You can think of `SDFG.user_args` as something similar to `arg_names`, but you can defines tuples.
So assigning `[("B", "C"), "A"]` to `user_args`  adds the following signature to the SDFG:
```python
def fun(arg1, arg2):
    pass
```
This tells the SDFG that `arg1` is a tuple and the first element supplies the SDFG non-transient input `A` and the second element is `B` and `A` will be taken from `arg1`.

I hope this helps.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
philip-paul-mueller: I kind of agree that it is _now_ would be nicer.
However, it would not look nice if we implement GTFN.
So I would keep it the way it is.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
philip-paul-mueller: Slightly modified it.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
philip-paul-mueller: To answer your first question that I do not fully understand myself either.
However, there are two things here. In eval mode you do not need the `OffsetProviders` but not in generation mode, where you only need the structure of the offset providers.
However, I think that we should remove the eval mode entierly, it was funny to implement but it does not really serves a purpose.
The second part to your first question is, in `build_sdfg_from_gtir()` the type annotation is `OffsetProviderType`, however, for `generate_sdfg()` the type annotation is `OffsetProvider`.
But I will remove the eval mode, this will also remove some complexity.

Your second question.
I do not need to pass them, in the generation of the `user_args` the unused are filtered out, but must be passed (or at least something).
I think I will change that such that `None` is passed in that case.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
philip-paul-mueller: No, the idea of `eval_mode` was that you could actually debug the code, because the Python debugger could not handle dynamically generates code.
However, this idea was stupid.

I am actually not sure if you can move it to call time.
Because you essentially need `Program.params` _and_ the offset providers.

Furthermore, now the offset providers are also passed by their position.
Thus you must know their order and not just which exists and which not.
The `user_args` interface does not support `**kwargs`.

I hoped this helped.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
philip-paul-mueller: I removed that thing in total.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -44,7 +45,7 @@ def _create_sdfg_bindings(
     offset_provider_type: gtx_common.OffsetProviderType,
```
edopao: I think you can avoid passing `offset_provider_type`, for dace, see below.
As for GTFN backend, it does not have `offset_provider` either in the binding stage.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -8,281 +8,114 @@
...
+    assert process_arg_stmt.strip().endswith(",")
+
+    needed_offset_providers: Optional[Container[str]] = None
+    if backend == "dace" and sdfg is not None:
+        sdfg_arglist = sdfg.arglist()
+        needed_offset_providers = {
+            table_name
+            for table_name in offset_provider_type
+            if gtx_sdfg_args.connectivity_identifier(table_name) in sdfg_arglist
+        }
```
edopao:
```suggestion
    if sdfg is not None:
        needed_offset_providers = {aname for aname, _ in sdfg_arglist.items() if gtx_dace_args.is_connectivity_identifier(aname)}
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -292,9 +125,118 @@ def bind_sdfg(
...
+    # Assumes that the order of the offset providers is stable.
+    processed_offset_providers = ""
+    for table_name in offset_provider_type:
+        if table_name in needed_offset_providers:
+            processed_offset_providers += f"(offset_provider['{table_name}'].ndarray, (0, 0))"
+        else:
+            processed_offset_providers += "None"
+        processed_offset_providers += ", "
+
+    return processed_offset_providers
```
edopao:
```suggestion
def _process_offset_providers(
    needed_offset_providers: List[str],
) -> str:
    # Assumes that the order of the offset providers is stable.
    processed_offset_providers: list[str] = []
    for table_name in needed_offset_providers:
            processed_offset_providers.append(f"(offset_provider['{table_name}'].ndarray, (0, 0))")

    return ", ".join(processed_offset_providers=
```

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1018,7 +1018,12 @@ def _add_sdfg_params(
...
             )
 
-        # the list of all sdfg arguments (aka non-transient arrays) which include tuple-element fields
+        # The list of the SDFG arguments", i.e. the non transient arrays. Note that
+        #  tuple arguments are flatten and their name is mangled and no longer matches
+        #  the name in the signature of the field operator / program. Also note that
+        #  scalar arguments after optimization might no longer be needed and have been
+        #  pruned from the SDFG. Also the domain related symbols are not included.
+        # NOTE: The dispatch code does not use it, instead the `user_args` are used.
         return [arg_name for arg_name, _ in sdfg_args]
```
edopao:
```suggestion
        # The list of the SDFG arguments, i.e. the global arrays, scalars and free symbols.
        #  Note that tuple arguments are flattened and their name is mangled and no longer
        #  matches the name in the signature of the field operator / program.
        #  Also note that unused SDFG symbols are pruned, so some scalar arguments
        #  used in symbolic domain expressions might not appear in this signature.
        #  NOTE: The dispatch code does not use it, instead the `user_args` are used.
        return [arg_name for arg_name, _ in sdfg_args]
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
philip-paul-mueller: I agree this is nicer.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
philip-paul-mueller: Are you sure about the last sentence?
I would say that symbols might appear in this signature that are not actually needed (and because of that are removed from the C-API).

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
philip-paul-mueller: This does not work.
The reason is because the order in `needed_offset_providers` is not stable and we have to pass the offset provders by their position.
So you must iterate over something that has the same order as the offset provider that is finally used as call time.
Furthermore, GTFN passes always all offset providers.

I looked at GTFN and there they "ship" all offset providers through the [`entry_point` property](https://github.com/GridTools/gt4py/blob/6327c4db63a943736726095537830ca2bd4ffa61/src/gt4py/next/program_processors/codegens/gtfn/gtfn_module.py#L218) .
I agree there is a way, it is just not super nice.

philip-paul-mueller:
@edopao I have now updated the code such that the bindings are generated during the bindings stage.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1018,7 +1018,13 @@ def _add_sdfg_params(
...
             )
 
-        # the list of all sdfg arguments (aka non-transient arrays) which include tuple-element fields
+        # The list of the SDFG arguments, i.e. the global arrays, scalars and free symbols.
+        #  Note that tuple arguments are flattened and their name is mangled and no longer
+        #  matches the name in the signature of the field operator / program.
+        #  Also note that some scalar arguments (which are lowered to symbols) listed in
+        #  this signature are might not be part of the generated C-API, as unused symbols
+        #  are exluded from it and might have been pruned from the SDFG during optimization.
+        # NOTE: The dispatch code does not use it, instead the `user_args` are used.
```
edopao, line 1027:
```suggestion
        #  Also note that some scalar arguments (which are lowered to symbols) listed in
        #  this signature might not be part of the generated C-API, as unused symbols
        #  are pruned from the SDFG during optimization.
        # NOTE: The dispatch code does not use it, instead the `user_args` are used.
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -26,7 +25,7 @@ def argument_processing_function(
     offset_provider: gtx_common.OffsetProvider,
     metrics_level: int,
     runtime_return_value: Any,
-) -> Any:
+) -> tuple[tuple[Any, ...], dict[str, Any]]:
     """Signature generated by `_create_sdfg_bindings()` and used to process the arguments for an SDFG call.
 
     See `_create_sdfg_bindings()` and `bind_sdfg()` for more.
```
edopao: The description of the return values would be useful here.

`_create_sdfg_bindings()` does not provide much information.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -41,8 +40,7 @@ def argument_processing_function(
 
 
 def _create_sdfg_bindings(
```
edopao: This function contains `_sdfg_` in the name, but it tries to support also `gtfn`.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -72,48 +69,77 @@ def _create_sdfg_bindings(
...
+            positional_arguments += _process_argument(
+                argument=arg_name,
+                param_type=param.type_,
+                unpacked_variables=unpacked_variables,
+                backend=backend,
+            )
+        assert positional_arguments.strip().endswith(",")
+    assert positional_arguments.strip().endswith(",")
+
+    positional_arguments, kwargs_arguments = _process_metric_arguments(
```
edopao: Maybe I would check `use_metrics` here, guarding the call to `_process_metric_arguments`

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -127,23 +153,22 @@ def bind_sdfg(
...
-    binding_code_generated_during_translation, jsdfg = source_code
-    assert f"def {bind_func_name}(" in binding_code_generated_during_translation
-
-    program_source = dataclasses.replace(inp, source_code=jsdfg)
-    binding_source = stages.BindingSource(
-        binding_code_generated_during_translation, library_deps=tuple()
+    assert isinstance(inp.source_code, dict)
+
+    sdfg = dace.SDFG.from_json(inp.source_code)
+    use_metrics = gtx_wfdcommon.SDFG_ARG_METRIC_COMPUTE_TIME in sdfg.arrays
```
edopao, line 166: `use_metrics` could be computed inside `_create_sdfg_bindings`

edopao (commented):
I think it looks better now. I am not sure whether it is worth sharing code with `gtfn`, as in the current implementation.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
philip-paul-mueller: Yes indeed.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
philip-paul-mueller: I am not sure if GTFN could do that, so I will keep it for now until I decided to either drop GTFN support or not.

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -8,281 +8,147 @@
...
     bind_func_name: str,
-) -> stages.BindingSource[code_specs.SDFGCodeSpec, code_specs.PythonCodeSpec]:
+    use_metrics: bool,
+    sdfg: Optional[dace.SDFG],
+    backend: Literal["gtfn", "dace"],
+) -> str:
     """
-    Creates a Python translation function to convert the GT4Py arguments list
-    to the SDFG calling convention.
+    Creates a Python function that translates GT4Py arguments into arguments suitable for a SDFG call.
```
edopao, line 55:
```suggestion
    Creates a Python function that translates GT4Py arguments into arguments suitable for the `nanobind` call.
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -8,281 +8,147 @@
...
-    sdfg_arglist = sdfg.arglist()
+    Args:
+        prog: The program definition.
+        bind_func_name: The name of the function to perform the translation.
+        use_metrics: If metric support was added to the underlying compiled code.
+            In that case the last two arguments of the generated function signature
+            are ignored and also not included in the output.
+        sdfg: If provided and `backend` is `dace` then only the offset providers that
+            are used in the SDFG are processed, otherwise `None` is passed.
+            This is compatible with the `user_args` signature.
```
edopao, line 73:
```suggestion
            are used in the SDFG are processed, otherwise all offset providers
            are passed. This is compatible with the `user_args` signature.
```

src/gt4py/next/program_processors/runners/dace/workflow/compilation.py
```diff
@@ -192,77 +200,33 @@ def _configure_external_workspace(self, **kwargs: Any) -> None:
...
-    ``program.sdfg(z)`` dump under the upcoming minimal-build-dir mode.
-
-    The SDFG we store here is the one on which we called `SDFG.compile(return_program_handle=False)`.
-    Note that the `compile()` call has side effects, because it applies transformations
-    to the SDFG, in order to enable code generation for the target platform.
-    Since we pass `return_program_handle=False`, the `compile()` method does not
-    return a `CompiledSDFG` instance, therefore we cannot access `CompiledSDFG.sdfg`,
-    which would be the modified SDFG from which DaCe generates the C++/CUDA/HIP code.
-    """
+    """Result of a DaCe compilation: library path + SDFG bindings + the SDFG itself."""
```
edopao, line 220:
```suggestion
    """Result of a DaCe compilation: SDFG build folder + SDFG bindings + the SDFG itself."""
```

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -44,14 +46,14 @@ def __init__(
     ) -> None:
         self._fun = fun
         # Retrieve metrics level from GT4Py environment variable.
+        # TODO(phimuell): Should the check be in the call or at construction.
```
edopao, line 49: I think it is OK to have it here, at construction, because the metrics level is not supposed to change:
```suggestion
```

On the other hand, it should also be possible to remove `self._collect_time` and just call `metrics.is_level_enabled(metrics.PERFORMANCE)`, without performance regression.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1064,10 +1070,10 @@ def visit_Program(self, node: gtir.Program) -> dace.SDFG:
                 assert isinstance(nsdfg.arrays[data], dace.data.Array)
                 nsdfg.arrays.pop(data)
 
-        # Create the call signature for the SDFG.
-        #  Only the arguments required by the GT4Py program, i.e. `node.params`, are added
-        #  as positional arguments. The implicit arguments, such as the offset providers or
-        #  the arguments created by the translation process, must be passed as keyword arguments.
+        # NOTE: A program uses the `user_args` mechanism to perform the call. So,
+        #   technically these arguments is not needed. However, the orchestrator needs
```
edopao, line 1074:
```suggestion
        #   technically `sdfg.arg_names` is not needed. However, the orchestrator needs
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -8,281 +8,147 @@
...
-        raise ValueError("Cannot unpack argument with length zero.")
-    else:
-        # The trailing comma is needed to unpack single-element tuples
-        code.append(f"{', '.join(tuple_args)}, = {arg_name}")
-    return tuple_args
+    See `_create_binding_function()` and `bind_sdfg()` for more.
 
+    Returns:
+        The function returns a `tuple` containing the processed positional arguments for
+        the call and a `dict` containing the keyword arguments of the call.
```
edopao, line 36:
```suggestion
        The function returns a `tuple` with the processed positional arguments and
        a `dict` with the keyword arguments to use in the `nanobind` call.
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -8,281 +8,147 @@
...
-
-    For scalar arguments, a single field is set in the SDFG arglist.
-
-    For array arguments, in addition to the data pointer, the fields for array shape
-    and strides are also set in SDFG arglist. This results in nested calls to
-    `_parse_gt_param()` with the scalar values of array shape and strides.
-
-    For tuple arguments, this function is recursively called on all elements of the tuple.
+    Args:
+        args: The arguments that are passed to the SDFG. A tuple of fields and nested tuples.
```
edopao, line 39:
```suggestion
        args: The arguments passed to the GT4Py program call. A tuple of fields and nested tuples.
```

src/gt4py/next/program_processors/runners/dace/workflow/bindings.py
```diff
@@ -8,281 +8,147 @@
...
+        offset_provider: The offset providers (order is used verbatim).
+        metrics_level: The level of the metric that is active.
+        runtime_return_value: NumPy array used to write back the runtime spend in the call.
```
edopao, line 42:
```suggestion
        runtime_return_value: NumPy array used to write back the runtime spent in the call.
```

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -450,21 +597,74 @@ def __call__(
...
+            interface.Parameter(param.id, param.type)  # type: ignore[arg-type]
+            for param in program.params
+        ]
+
+        # Offset provider arguments
+        # NOTE: The prefixing is important for the binding stages and used as the sole
+        #   criteria to determine if a parameter is a connection or not. Furthermore,
+        #   the name, i.e. the thing after `@`, is used to find out where to find the
+        #   table in the offset provider. This means we are no longer dependent on the
+        #   ordering of `offset_providers`.
```
edopao, line 649:
```suggestion
        #   the name, i.e. the string after `@`, follows the same order as the
        #   connectivities in the `offset_providers` table.
```

edopao (commented):
LGTM

PR #2725, opened 2026-07-29
fix[next]: add missing side condition to EtaReduction

havogt:
`EtaReduction` can turn a well-formed expression into one with a free (escaped)
symbol, which `gt4py.eve`'s symbol-reference validator then rejects with
`EveValueError: Symbols {...} not found.`


havogt:
or https://github.com/GridTools/gt4py/pull/2729?

havogt:
deleted the passes in #2729

PR #2727, opened 2026-07-29
fix[next]: validate rhs in comparison type synthesizer

havogt:
In synthesize_binary_math_comparison_builtins the assertion iterates over (lhs, rhs) but tests lhs in the body, so the right operand is never validated. As a result comparisons such as scalar < FieldType, scalar < DomainType and scalar < TupleType are accepted and typed as bool instead of being rejected. The fix tests arg instead of lhs.

### DIFF
--- a/tests/next_tests/unit_tests/iterator_tests/test_type_inference.py
+++ b/tests/next_tests/unit_tests/iterator_tests/test_type_inference.py
@@ -89,6 +89,8 @@ def expression_test_cases():
         (im.call("power")(2.0, 2), float64_type),
         (im.plus(1, 2), int_type),
         (im.eq(1, 2), bool_type),
+        (im.less(im.ref("a", int_type), im.axis_literal(IDim)), ts.DomainType(dims=[IDim])),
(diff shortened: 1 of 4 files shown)


tests/next_tests/unit_tests/iterator_tests/test_type_inference.py
```diff
@@ -89,6 +89,8 @@ def expression_test_cases():
         (im.call("power")(2.0, 2), float64_type),
         (im.plus(1, 2), int_type),
         (im.eq(1, 2), bool_type),
+        (im.less(im.ref("a", int_type), im.axis_literal(IDim)), ts.DomainType(dims=[IDim])),
+        (im.less(im.axis_literal(IDim), im.ref("a", int_type)), ts.DomainType(dims=[IDim])),
```
havogt, line 93: these cases are added because the original question was is IDim < JDim a valid case in gtir and then noticed that this is not covered.

tehrengruber (approved):

PR #2730, opened 2026-07-29
refactor[next]: carry the connectivity name in `OffsetType`

havogt:
## Description

The unstructured shift lowering took the ITIR offset tag from the source-level identifier (`foast.Name.id`, and after #2724 also `foast.Attribute.attr`), so `field(Off)` had to pattern-match the FOAST node shape, and resolved to the wrong tag whenever the identifier was not the connectivity name:

```python
from next_tests.integration_tests.cases import V2E as RenamedV2E

@gtx.field_operator
def testee(a: cases.EField) -> cases.VField:
    return neighbor_sum(a(RenamedV2E), axis=V2EDim)
```

```
embedded: [1 3]
gtfn:     KeyError: "Offset 'RenamedV2E' not found in offset provider."
```

Embedded resolves through `FieldOffset.value` (`fbuiltins.py:485-488`) and was unaffected, so the two execution paths disagreed. This dates back to the first FOAST shift lowering (#625, 2022-01-26) and became observable when embedded remap landed (#1309).

## Change

`ts.OffsetType` gains `name`, set from `FieldOffset.value` in `__gt_type__`. Both shift arms in `_visit_shift` now match on the type rather than the node shape — mirroring the Cartesian arm, which already reads the `Dimension` out of `ts.DimensionType`. The `foast.Attribute` special case from #2724 is dropped; it is subsumed.

`name` is `None` for the `Dim + idx` offsets synthesized in `_deduce_binop_type`. Those are Cartesian, resolved from `source`/`target` alone, and never looked up in the offset provider. `FieldOffset.__gt_type__` always sets it, so Cartesian *`FieldOffset`s* keep a name and the `field(Off[idx])` arm is unaffected.

The `field(Off)` arm moved below the `as_offset` arm: `as_offset(...)` propagates its argument's `OffsetType`, so it would otherwise be caught by the now type-driven pattern. (It also has a single target, so the `target=[_, _]` guard would reject it — the move makes the ordering intentional rather than incidental.)

This is a step toward the `# TODO(havogt): replace by ConnectivityType` already on `OffsetType`, not the whole thing — I did not add an ADR since it does not change the public API shape.

## Not addressed

A separate name coupling is still live in the backends: a sparse field's LOCAL dimension is looked up in the offset provider by its `value`, so the local dimension must be named exactly like its offset.

- `codegens/gtfn/gtfn_module.py:93-96` — `dim_name = dim.value` → `common.get_offset_type(offset_provider_type, dim_name)`
- `runners/dace/lowering/gtir_to_sdfg.py:553` — raises `ValueError("The provided local dimension … does not match any offset provider type.")`
- `runners/dace/lowering/gtir_dataflow.py:733`

That is the direction of `common.py:977-979` (`TODO(havogt): refactor towards encoding this information in the local dimensions of the ConnectivityType.domain`) and is untouched here.

## Tests

Added to `test_import_from_mod.py`, across the backend matrix:

- `test_import_renamed_offset_unstructured_shift` — `neighbor_sum(a(RenamedV2E), axis=V2EDim)`
- `test_import_renamed_offset_sparse_shift` — `a(RenamedE2V[0])`, covering the `field(Off[idx])` arm

Both fail on #2724 and pass here.

Local runs (full backend matrix incl. GPU, `mypy src/` and `pre-commit` clean):

- `tests/next_tests/unit_tests` — 2151 passed, 9 skipped, 15 xfailed
- `tests/next_tests/integration_tests` — 5316 passed, 197 skipped, 22 deselected, 612 xfailed

The 22 deselected are `test_orchestration.py::test_sdfgConvertible_connectivities[*-dace.run_dace_gpu]` and its module siblings: that test fails identically with and without this change on my machine (`TypeError: Illegal copy! (from gt_conn_E2V to tlet_0_deref)`), verified by stashing and re-running on the base commit. Pre-existing, unrelated.

## Requirements


egparedes:
Close in favor of #2898

PR #2732, opened 2026-07-29
feat[next]: register JaxArrayField as a JAX pytree node

havogt:
Flatten a JaxArrayField into its backing array (traced child) plus its domain (static aux data), so gt4py fields can be passed to and returned from jax.jit-ed functions outside of field operators.

### DIFF
--- a/src/gt4py/next/embedded/nd_array_field.py
+++ b/src/gt4py/next/embedded/nd_array_field.py
@@ -1090,6 +1090,22 @@ if jnp:
     common._field.register(jnp.ndarray, JaxArrayField.from_array)
     common._connectivity.register(jnp.ndarray, JaxArrayConnectivityField.from_array)
 
+    def _flatten_jax_field(
+        field: JaxArrayField,
+    ) -> tuple[tuple[core_defs.NDArrayObject], common.Domain]:
+        return (field.ndarray,), field.domain
+
+    def _unflatten_jax_field(
+        domain: common.Domain, children: tuple[core_defs.NDArrayObject]
+    ) -> JaxArrayField:
+        return JaxArrayField(domain, children[0])  # type: ignore[abstract] # mypy does not see '__gt_builtin_func__' as implemented by 'FieldBuiltinFuncRegistry'
+
+    jax.tree_util.register_pytree_node(
+        JaxArrayField,  # type: ignore[type-abstract, unused-ignore] # only reported when 'jax' is installed, see '_unflatten_jax_field'
+        _flatten_jax_field,
+        _unflatten_jax_field,
+    )
+
 
 def _broadcast(field: common.Field, new_dimensions: Sequence[common.Dimension]) -> common.Field:
     if field.domain.dims == new_dimensions:
--- a/tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
+++ b/tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
@@ -1758,3 +1758,44 @@ def test_concat(fields_data, dim, expected_data, expect_error):
 
         assert result.domain == expected_domain
         np.testing.assert_allclose(result.asnumpy(), expected_array)
+
+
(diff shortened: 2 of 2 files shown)


tests/next_tests/unit_tests/embedded_tests/test_nd_array_field.py
```diff
@@ -1758,3 +1758,44 @@ def test_concat(fields_data, dim, expected_data, expect_error):
...
+    traced_domains = []
+
+    @jax.jit
+    def identity(f):
+        traced_domains.append(f.domain)
+        return f
+
+    result = identity(field)
+
+    assert traced_domains == [domain]  # the domain is static metadata, not a traced value
```
havogt: What are we actually testing here?

havogt: Not much, honestly — replaced in 2d1173e87.

The assertion escaped the trace through a side-effecting list append, so it only held because `identity` was called exactly once; `jax.jit` caches by pytree structure and shape/dtype, so a second call or a parametrization would have left `traced_domains` stale and the failure would have been baffling. Beyond that it mostly duplicated `test_jax_jit_field_arguments`.

It now flattens and unflattens directly, and asserts the three things the registration is actually promising:

- the single child is the backing array, not the field itself (an unregistered type flattens to one leaf too, so `len(children) == 1` alone does not discriminate — this was the accidental part of the old test)
- the domain survives `tree_unflatten`
- a differing domain yields a differing `treedef`, which is what makes the domain static and forces a retrace rather than letting a field come back silently mislabelled

Checked that it fails without the registration:

```
>       assert children[0] is field.ndarray
E       AssertionError: assert JaxArrayField(...) is Array([[1., 1., 1.], [1., 1., 1.]], dtype=float64)
```

src/gt4py/next/embedded/nd_array_field.py
```diff
@@ -1090,6 +1090,22 @@ class JaxArrayConnectivityField(NdArrayConnectivityField):
...
 
+    def _flatten_jax_field(
+        field: JaxArrayField,
+    ) -> tuple[tuple[core_defs.NDArrayObject], common.Domain]:
+        return (field.ndarray,), field.domain
+
+    def _unflatten_jax_field(
+        domain: common.Domain, children: tuple[core_defs.NDArrayObject]
+    ) -> JaxArrayField:
+        return JaxArrayField(domain, children[0])  # type: ignore[abstract] # mypy does not see '__gt_builtin_func__' as implemented by 'FieldBuiltinFuncRegistry'
```
egparedes, line 1101: Just for my own understanding:  shouldn't this distinguish between `JaxArrayField` and `JaxArrayConnectivityField`?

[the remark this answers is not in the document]
havogt: Deliberate — routing through `from_array` would break more than it fixes.

JAX calls `unflatten` with things that are not arrays. Measured on this exact registration: `object()` sentinels and `None` (vmap, jacfwd), `ShapeDtypeStruct` (eval_shape), plus four tracer types. `from_array` rejects all of those, so `eval_shape` / `vmap` / `jacfwd` would start raising.

For the shape-preserving transforms (`jit`, `grad`, `lax.scan`) there is nothing left to validate: the domain is static aux data, resolved at trace time and baked into the very slice constants that produced the array. The mismatch you describe is reachable only under shape-*changing* transforms, which cannot work with a static domain regardless — validating here would turn "silently wrong" into "raises", not into "works".

havogt: I looked a bit into it, but it's slightly more complicated to the the premap working (that's what we expected). Just registering will not make it work. Let's talk about it. I'd leave it for a separate PR, this will already allow jiting cartesian problems.

egparedes (approved):
After discussing offline, this looks good to me as a first step. There are more things to do for connectivities, but we'll do them in different PRs.

PR #2719, opened 2026-07-24
feat[next-dace]: Support external memory for transient arrays

edopao:
This pull request introduces a new "external workspace" memory mode for DaCe transients in the `gt4py.next` DaCe backend, allowing users to provide their own workspace buffers and thereby avoid per-call GPU memory allocation overhead. The change is documented in a new ADR and is implemented by extending the transient memory mode enum, updating backend and compilation workflows, and making the memory management API more flexible and explicit. The pull request also cleans up and unifies transient memory handling throughout the codebase.

The most important changes are:

**External Workspace Memory Mode (EXTERNAL) Implementation:**
* Introduced a new `TransientMemoryMode` enum with an `EXTERNAL` option, allowing users to supply their own workspace buffers for DaCe transients. This is documented in a detailed ADR (`docs/development/ADRs/next/0027-External_Workspace_Memory.md`) and referenced in the ADR index. [[1]](diffhunk://#diff-a6ad8c8541d8f8d6aa0f3fb050c209f047fc5e4007263aee59afc9a978debf15R1-R171) [[2]](diffhunk://#diff-f082dca95a0995ab2ef5db5d9bf6f9a89e70ddf74bf02640ab84656407f17cf5R51) [[3]](diffhunk://#diff-ddbc0e8ab1d5b89929dfbd12477936ae3acad531911fa9afb125066b6fe74706R114-R147)
[...]

PR #2733, opened 2026-07-29
fix[next]: stop KeyError from masking compiled-program failures

havogt:
When a compiled program cannot be obtained, the traceback leads with a KeyError on the dispatch
dict, even though the actual cause is something else. 

The reason is that `CompiledProgramsPool.__call__` did all cache-miss handling inside the
except KeyError handler of the dispatch lookup, so everything raised there was chained to the
KeyError.

The dispatch lookup now uses compiled_programs.get(key) and an is-None check, so a miss raises
nothing and there is no exception context left to chain to. The miss branch itself is unchanged.
A new _load_artifact wraps load failures as
`<SomeError> from <cause>`, so `<SomeError>` is
what the reader sees. Worker exceptions from `future.result()` are deliberately not wrapped and
propagate as themselves.

### DIFF
--- a/src/gt4py/next/otf/compiled_program.py
+++ b/src/gt4py/next/otf/compiled_program.py
@@ -598,6 +598,9 @@ class CompiledProgramsPool(Generic[ffront_stages.DSLDefinitionT]):
 
         artifact_future = self._compilation_jobs.pop(key)
         assert key not in self.compiled_programs
+        # A failing 'result()' keeps the future alive through its traceback, so the
+        # weak entry would survive until the next GC pass and be reported twice.
+        _ongoing_compilations.pop(artifact_future, None)
         self.compiled_programs[key] = self._load_artifact(artifact_future)
         return True
 
--- a/src/gt4py/next/otf/compiled_program.py
+++ b/src/gt4py/next/otf/compiled_program.py
@@ -432,10 +432,10 @@ class CompiledProgramsPool(Generic[ffront_stages.DSLDefinitionT]):
             arg_specialization_key,
         )
 
-        try:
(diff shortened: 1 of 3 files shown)


src/gt4py/next/otf/compiled_program.py
```diff
@@ -427,10 +427,12 @@ def __call__(
             arg_specialization_key,
         )
 
-        try:
-            compiled_program = self.compiled_programs[key]
+        # Note: a plain lookup, without a `KeyError` handler: everything raised while such a
+        # handler is active gets chained to it, and the dispatch miss with its opaque key then
+        # dominates the traceback and hides the actual failure.
```
egparedes: I find this comment too verbose at least, and maybe completely unnecessary.

egparedes (commented):
LGTM, I can approve as soon as it not longer in draft state.

src/gt4py/next/otf/compiled_program.py
havogt: Trimmed to one line:

```python
# Note: no `KeyError` handler, as anything raised inside one gets chained to it.
compiled_program = self.compiled_programs.get(key)
```

I'd keep that much rather than nothing — without it the natural "optimization" is back to `d[key]` + `except KeyError`, which is exactly what this PR undoes, and the reason isn't visible from the code.

Also rebased onto current `main`, which conflicted for real: #2719 moved the load path to `self.backend.load_artifact(artifact)`, so `_load_artifact` now goes through that seam instead of calling `artifact.load()` directly.

egparedes (approved):
LGTM

PR #2734, opened 2026-07-29
feat[next]: name program and static args in "no program compiled" error

havogt:
Before:

    RuntimeError: No program compiled for this set of static arguments.

After:

    RuntimeError: No program compiled for this set of static arguments of 'prog': cond=False.
    Note that a variant is also selected by the identity of the 'offset_provider' entries and,
    for generic programs, by the argument types.

Neither the program nor the offending arguments were named, and the offset provider taking part
in the selection was not mentioned at all, which is the other common cause of an unexpected miss.

### DIFF
--- a/src/gt4py/next/otf/compiled_program.py
+++ b/src/gt4py/next/otf/compiled_program.py
@@ -458,10 +458,30 @@ class CompiledProgramsPool(Generic[ffront_stages.DSLDefinitionT]):
                 )  # passing `enable_jit=False` because a cache miss should be a hard-error in this call
 
             else:
-                raise RuntimeError("No program compiled for this set of static arguments.")
+                raise RuntimeError(
+                    f"No program compiled for this set of static arguments of "
+                    f"'{self.definition.__name__}'{self._describe_argument_descriptors(key[0])}."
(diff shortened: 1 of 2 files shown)


src/gt4py/next/otf/compiled_program.py
```diff
@@ -456,11 +456,31 @@ def __call__(
...
+                    " Note that a variant is also selected by the identity of the"
+                    " 'offset_provider' entries and, for generic programs, by the argument types."
+                )
 
         with compiled_program_call_context(self, key, args, kwargs, offset_provider):
             compiled_program(*args, **kwargs, offset_provider=offset_provider)
 
+    def _describe_argument_descriptors(self, descriptor_values: tuple[Hashable, ...]) -> str:
+        # Note: not `zip(..., strict=True)`: this runs while building an error message, so a
+        # mismatch must degrade to a shorter description instead of replacing the error.
```
egparedes: I don't get the reference to `zip` here. Make the comment shorter and mention only the existing code.

egparedes (approved):
LGTM, just a minor nitpick.

PR #2739, opened 2026-07-30
fix[next-dace]: Include modifed SDFG with GPU TX markers in build folder fingerprint

edopao:
This pull request refactors and improves the GPU trace marker logic in the DaCe compilation workflow, enhances test coverage, and clarifies test parameterization. The main functional change is to ensure that the fingerprint of the build folder includes the modified SDFG after applying the GPU transaction markers. This way, enabling or not the GPU transaction markers results in different build artifacts. Tests are updated to reflect the new marker application logic and to ensure that build artifacts change when relevant compilation settings differ.

**DaCe compilation logic improvements:**

* Refactored `_add_tx_markers` to take and return an `ExtensionSource`, only modifying the SDFG if GPU scheduling is detected, and returning a new `ExtensionSource` with updated SDFG JSON.
* Updated the main DaCe compiler logic to call `_add_tx_markers` only when GPU trace markers are requested and the device type is GPU, ensuring markers are not redundantly applied. [[1]](diffhunk://#diff-89b749ff9b80be5fdb8a7e0d05411d2b992b6a5892f819c7a3e001c4241a6c0dL200-R222) [[2]](diffhunk://#diff-89b749ff9b80be5fdb8a7e0d05411d2b992b6a5892f819c7a3e001c4241a6c0dL218-R235)

**Test suite improvements:**

* Refactored tests to use a real `program_source` fixture and removed unnecessary mocks and spies, directly verifying the presence or absence of GPU TX markers on the compiled SDFG. [[1]](diffhunk://#diff-57ff279a6de9ce815a6bde8ae126007687da1d07f6ad28626e998541fc9d89b3L82-R83) [[2]](diffhunk://#diff-57ff279a6de9ce815a6bde8ae126007687da1d07f6ad28626e998541fc9d89b3R100-L109) [[3]](diffhunk://#diff-57ff279a6de9ce815a6bde8ae126007687da1d07f6ad28626e998541fc9d89b3L119-L123) [[4]](diffhunk://#diff-57ff279a6de9ce815a6bde8ae126007687da1d07f6ad28626e998541fc9d89b3L138-L173)
* Added new tests to assert that identical compilation inputs produce the same artifact, and that changing instrumentation or compiler flags results in different artifacts. [[1]](diffhunk://#diff-57ff279a6de9ce815a6bde8ae126007687da1d07f6ad28626e998541fc9d89b3R181-R203) [[2]](diffhunk://#diff-57ff279a6de9ce815a6bde8ae126007687da1d07f6ad28626e998541fc9d89b3R213-R248)

**Test parameterization and device naming:**

* Updated device type parameterization in both compilation and translation tests to use consistent and descriptive IDs (`CPU`, `GPU`, `CUDA`, `ROCM`) and to use the correct device type constants. [[1]](diffhunk://#diff-19fed38498781ceef82337d097e43458274d2ebe7e0bae4578839e9fbed15194L31-R33) [[2]](diffhunk://#diff-8ba502dd24c52188c94e85f0bcd66c31fbe3093077168033f99c14eb19ac1cb4L54-R55)

These changes improve the correctness, maintainability, and clarity of both the DaCe compilation workflow and its associated tests.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/workflow/common.py
+++ b/src/gt4py/next/program_processors/runners/dace/workflow/common.py
@@ -17,37 +17,6 @@ from gt4py.next import config as gtx_config
 from gt4py.next.otf.compilation import common as gtx_compilation_common
 
 
-def serialize_sdfg_as_json(sdfg: dace.SDFG) -> dict[str, Any]:
-    """
-    Serialize an SDFG to JSON while removing ``guid`` keys.
-
-    `guid` is a per-element identity token: it does not affect code generation, and
-    `SDFG.from_json()` assigns fresh ids anyway. Keeping it would make the compile
-    cache key depend on element creation order, so two structurally identical
-    lowerings of the same program would not share a cached build.
-    # FIXME(edopao): remove this workaround once the SDFG lowering is stable.
-
-    Note that the recursive implementation to drop ``guid`` keys is 2-3x faster than
-    the iterative one, on a large SDFG, and it is also simpler to read.
-
-    Note that we set 'hash=True' to compute the SDFG hash and store it in the JSON
-    object. We compute the hash in order to refresh `cfg_list` on the SDFG, which
-    makes the JSON serialization stable.
-    """
-    json_obj = sdfg.to_json(hash=True)
-
-    def _drop_guids(obj: Any) -> Any:
-        if isinstance(obj, dict):
-            return {k: _drop_guids(v) for k, v in obj.items() if k != "guid"}
-        if isinstance(obj, list):
-            return [_drop_guids(v) for v in obj]
-        if isinstance(obj, tuple):
-            return tuple(_drop_guids(v) for v in obj)
-        return obj
-
-    return _drop_guids(json_obj)
-
-
 SDFG_ARG_METRIC_LEVEL: Final[str] = "gt_metrics_level"
 """Name of SDFG argument to input the GT4Py metrics level."""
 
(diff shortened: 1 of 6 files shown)


src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -17,6 +17,37 @@
 from gt4py.next.otf.compilation import common as gtx_compilation_common
 
 
+def serialize_sdfg_as_json(sdfg: dace.SDFG) -> dict[str, Any]:
```
philip-paul-mueller: You should define the function below the definition of the constants.

philip-paul-mueller (commented):
LGTM.
Maybe you should also add a note to say that `compile()` might modify the hash.

philip-paul-mueller (approved):
LGTM

src/gt4py/next/program_processors/runners/dace/workflow/common.py
edopao: You are right. I asked the agent to move the function into this module and I simply trusted it.

edopao:
> LGTM. Maybe you should also add a note to say that `compile()` might modify the hash.

Good idea, I have added a note to `DaCeCompilationArtifact`. However, we should have captured most of the context both in the translation cache (which fingerprints all arguments passed to auto optimizer) and in the `library_path` (which includes the build folder, which in turn fingerprints the non-default dace configuration).

PR #2740, opened 2026-07-30
fix[next]: unique sys.modules key per compiled program variant

tehrengruber:
#2431 keeps a nanobind module alive by registering it in `sys.modules`, but the key was just the `entry_point_name`. Static variants of the same program share that name (and the `.so` base name), so they collided and evicted one another:

```python
# variant_a / variant_b: same program, different static args
variant_a: CompilationArtifact
variant_b: CompilationArtifact
func_a = variant_a.load()  # sys.modules["gt4py.__compiled_programs__.my_fo"] = module_a
func_b = variant_b.load()  # overwrites the key; module_a now only reachable via func_a

del variant_a
# on nanobind >=2.10 module_a can be GC'd (nb_module_clear) while func_a
# is still in use -> segfault when func_a is called
```

Solution here is to not rely on `sys.modules`, but just keep a reference to the module in a global variable.

Text and code generated with the help of LLMs and then verified by me.

### DIFF
--- a/src/gt4py/next/otf/compilation/importer.py
+++ b/src/gt4py/next/otf/compilation/importer.py
@@ -12,8 +12,7 @@ from types import ModuleType
 
 
 #: Keeps references to dynamically loaded modules alive for the lifetime of the
-#: process. Required for nanobind extension modules, which must outlive the
-#: functions imported from them (see PR #2431). Append-only.
+#: process.
 _loaded_modules: list[ModuleType] = []
 
 
--- a/src/gt4py/next/otf/compilation/importer.py
+++ b/src/gt4py/next/otf/compilation/importer.py
@@ -8,28 +8,29 @@
 
 import importlib.util
 import pathlib
-import sys
 from types import ModuleType
 
 
-def import_from_path(
-    module_file: pathlib.Path, *, add_to_sys_modules: bool = True, sys_modules_prefix: str = ""
-) -> ModuleType:
+#: Keeps references to dynamically loaded modules alive for the lifetime of the
+#: process. Required for nanobind extension modules, which must outlive the
+#: functions imported from them (see PR #2431). Append-only.
+_loaded_modules: list[ModuleType] = []
+
+
+def import_from_path(module_file: pathlib.Path, *, keep_reference: bool = False) -> ModuleType:
     """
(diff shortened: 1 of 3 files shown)


src/gt4py/next/otf/compilation/importer.py
```diff
@@ -8,28 +8,36 @@
...
+            return the already loaded module on subsequent calls for the same
+            file.
 
     Returns:
         The loaded module.
     """
-    module_name = module_file.name.split(".")[0]
+    key = str(module_file)
+    if keep_reference and (cached := _loaded_modules.get(key)) is not None:
+        return cached
```
egparedes: I'm not sure if using the `_loaded_modules` variable as a cache is a good idea, I think it can add hidden bugs in weird corner cases which might be very hard to track. Are you aware of any scenario where this would matter for performance? Otherwise, I think I'd prefer to define it something similar to:

```python
_loaded_modules: list[tuple[str, Module]]
...
_loaded_modules.append((key, loaded_module))
```

egparedes (commented):
I'm not sure about using the holding variable as a cache, so proposed an alternative.

tests/next_tests/unit_tests/otf_tests/compilation_tests/test_importer.py
```diff
@@ -24,22 +21,14 @@ def function(a, b):
...
-    imported_modules = set(sys.modules.keys())
-    module = importer.import_from_path(file, add_to_sys_modules=False)
-    assert hasattr(module, "function")
-    assert "module" not in sys.modules
-    assert set(sys.modules.keys()) == imported_modules
-
-    module = importer.import_from_path(
-        file, add_to_sys_modules=True, sys_modules_prefix="_temp_test_prefix_"
-    )
+    # Without keeping a reference, the module is loaded but not retained.
```
tehrengruber, line 24: Not sure if these tests still make sense. I'm leaning to remove them.

egparedes: Those are testing if the `keep_reference` option works as intended, right? I think they still make sense, but feel free to remove them since in the current trivial this is trivially true.

egparedes (approved):
LGTM

PR #2744, opened 2026-07-31
feat[next-dace]: Add synchronization for multi-stream scheduling

edopao:
## Summary

This PR makes DaCe's multi-stream scheduling opt-in and safe for GT4Py's external workspace mode. When `max_concurrent_gpu_streams >= 1`, the DaCe backend injects CUDA/HIP event-based synchronization around SDFG execution, ordering the internal DaCe streams with respect to a caller-provided external stream (or the default stream).

## Motivation

External workspace mode (`transient_memory_mode=EXTERNAL`) avoids per-call GPU allocations by reusing a caller-provided workspace. This is only safe when no other GPU work touches the workspace between two consecutive calls. A concrete use case that breaks this assumption is overlapping GT4Py program execution with MPI halo exchange: halo exchange runs on one CUDA/HIP stream while the GT4Py program runs on DaCe's internal streams. Without explicit ordering, the external workspace can be reused before previous kernels or the halo exchange finish, causing data races.

## Changes

- Introduce `max_concurrent_gpu_streams` as a backend parameter:
  - Added `max_concurrent_gpu_streams` to `make_dace_backend`, `DaCeBackendFactory`, `DaCeWorkflowFactory`, `DaCeTranslationStepFactory`, `DaCeCompiler`, and `DaCeTranslator`.
  - Negative values are rejected in `DaCeTranslator.__post_init__`.
  - `0` preserves the historical default-stream-only behavior (sets DaCe's `compiler.cuda.max_concurrent_streams` to `-1`).
  - `>= 1` enables DaCe's internal stream pool.
  - Ignored on CPU targets.

- Added optional `external_sync_stream` support:
  - Added `external_sync_stream: cupy.cuda.Stream | None = None` to `make_dace_backend`.
  - The stream is stored on `DaCeBackend` (not the picklable workflow) and injected at artifact load time.
  - When no stream is provided and multi-stream scheduling is enabled, the default stream is used implicitly.
  - When `max_concurrent_gpu_streams == 0`, `external_sync_stream` is ignored and no validation occurs.

- Implemented event-based SDFG synchronization:
  - Added `add_synchronization` in the translation step.
  - When enabled on GPU SDFGs, injects `sync_entry` and `sync_exit` tasklets that record per-stream done events on internal DaCe streams and make the external sync stream wait on them.
  - Added the SDFG scalar symbol `__external_sync_stream`, passed as a `uint64` scalar.
  - When `max_concurrent_gpu_streams == 0` and a blocking call is requested, emits tasklets that synchronize on the default stream.
  - When `max_concurrent_gpu_streams == 0` and an asynchronous call is requested, emits DaCe init code that forces all internal streams to the default stream and skips the tasklets.

- Added `async_sdfg_call` parameter:
  - Controls whether the SDFG call blocks until GPU execution has completed.
  - Default is `True` (asynchronous).

- Added runtime stream validation:
  - Rejected on CPU targets.
  - Must be a `cupy.cuda.Stream`.
  - Must be on the current GPU device.
  - Must pass `cudaStreamQuery` (accepting both `cudaSuccess` and `cudaErrorNotReady`).

- Added safeguards:
  - `transient_memory_mode=POOL` combined with `max_concurrent_gpu_streams >= 1` is rejected, because the memory-pool pass assumes default-stream ordering.

- Added / updated unit tests:
  - Stream synchronization behavior in translation.
  - Stream pointer forwarding and default-stream fallback in compilation.
  - Backend factory parameter propagation and validation.

## Default behavior

No behavioral change unless explicitly requested: `max_concurrent_gpu_streams=0` (default) produces the same SDFG and execution as before.

## Testing

```bash
uv run pytest tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_backend.py
uv run pytest tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_compilation.py
uv run pytest tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_streams.py
uv run pytest tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/test_dace_translation.py
```

## Out of scope

- Per-call stream selection.
- Non-cupy stream support (e.g. torch, raw CUDA, HIP handles).
- Device-wide synchronization.
- Multi-stream safety for SCOPED/POOL modes.
- GT4Py-owned cross-call events; the caller is responsible for ordering work on the external stream across calls.

## Related documents

- docs/development/ADRs/0028-External_Stream_Synchronization.md


edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

docs/development/ADRs/0028-External_Stream_Synchronization.md
```diff
@@ -0,0 +1,135 @@
...
+## Decision
+
+We introduce an optional synchronization layer with the following design.
+
+### Configuration
+
+- Add `max_concurrent_gpu_streams` as a parameter to `DaCeWorkflowFactory` and
+  the `make_dace_backend` helper. It maps to DaCe's
+  `compiler.cuda.max_concurrent_streams`. A value of `0` keeps the historical
+  default-stream-only behavior (DaCe value `-1`); values `>= 1` enable DaCe's
```
philip-paul-mueller, line 41: This is just my personal opinion: I think it is confusing to say in DaCe `-1` mean `0` in GT4Py.
However, you can keep it if you want.

It also means that there is no way to say "use as mean stream as you think you should use".

docs/development/ADRs/0028-External_Stream_Synchronization.md
```diff
@@ -0,0 +1,135 @@
...
+  internal stream pool. Negative values are rejected at backend construction
+  time (the validation lives in `DaCeTranslator.__post_init__`, which is reached
+  by both direct translator creation and the backend factory).
+- Add `async_sdfg_call` as a parameter to `make_dace_backend`. When `True`
```
philip-paul-mueller, line 45: This will probably not work.
The reason is our asynchronous launch relies on the fact that when you use only the default stream DaCe does not add _any_ synchronization call.
Am actually not sure if you could use `SDFGState.nosync` to achieve something similar, but I am not fully sure if it will work (I forgot, but am skeptical).
However, setting that value is also not so good, because then you do not have synchronization over multiple states, i.e. you could read some memory before it is computed (I agree it is most likely not a big problem for us as we mostly have one state, but it is not guaranteed).

In the new GPU code generator, there is a way to kind of emulate that behaviour, as far as I know in a safe way, by suppressing the _final_ synchronization.

docs/development/ADRs/0028-External_Stream_Synchronization.md
```diff
@@ -0,0 +1,135 @@
...
+  (default), the SDFG call returns immediately and GPU kernels run concurrently
+  with the Python driver; when `False`, the SDFG call blocks until the GPU work
+  has completed.
+- The settings are ignored on CPU targets.
+
+### External sync stream
+
+- Add an optional `external_sync_stream: cupy.cuda.Stream | None = None`
```
philip-paul-mueller, line 53: Also personal opinion: I would not make it a backend property, because if you want to use multiple streams you need multiple backend.
In my view this is a call property and the only thing that the backend needs to know is, if there is an external stream or not.

docs/development/ADRs/0028-External_Stream_Synchronization.md
```diff
@@ -0,0 +1,135 @@
...
+  `cupy.cuda.Stream` is not picklable. The stream is injected at compiled-program
+  load time via `DaCeCompilationArtifact.load()`.
+- If no stream is provided and `max_concurrent_gpu_streams >= 1`, the CUDA/HIP
+  default stream (`0`) is used implicitly as the synchronization anchor.
+- When `max_concurrent_gpu_streams == 0`, `external_sync_stream` is ignored and
+  the SDFG executes asynchronously on the default stream.
+
+### Synchronization mechanism
+
+When `max_concurrent_gpu_streams >= 1` and the SDFG contains GPU schedules,
```
philip-paul-mueller, line 66: I think this should be done through an SDFG based API.

docs/development/ADRs/0028-External_Stream_Synchronization.md
```diff
@@ -0,0 +1,135 @@
...
+- `external_sync_stream` is validated during compiled-program construction when
+  it is not `None`:
+  - Rejected on CPU targets.
+  - Must be a `cupy.cuda.Stream`.
+  - Must be on the current GPU device.
+  - Must pass `cudaStreamQuery` (accepting both `cudaSuccess` and
+    `cudaErrorNotReady`).
+- When `max_concurrent_gpu_streams == 0`, `external_sync_stream` is ignored even
+  if provided, so no validation occurs.
+- The combination `transient_memory_mode=POOL` and
```
philip-paul-mueller, line 101: This point I do not understand.

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -27,16 +27,25 @@
...
+
+    These resources are intentionally kept out of the executor workflow: the workflow
+    (and its `DaCeCompiler`) is pickled when compilation is offloaded to a worker
+    process, while a `cupy.cuda.Stream` cannot be pickled. These resource are instead
+    stored on the backend object, which lives in the calling process, and passed
+    through to `DaCeCompilationArtifact.load()`.
+    """
 
     external_workspace: gtx_wfdcommon.ExternalWorkspace | None = None
+    external_sync_stream: Any | None = None
```
philip-paul-mueller, line 41: You can find a stream protocol here: https://github.com/C2SM/icon4py/blob/a25f35c7f9f67de96d6b70a0d5b6fd413c72ded4/model/common/src/icon4py/model/common/decomposition/definitions.py#L60

src/gt4py/next/program_processors/runners/dace/workflow/backend.py
```diff
@@ -27,16 +27,25 @@
...
 
     def load_artifact(self, artifact: stages.CompilationArtifact) -> stages.ExecutableProgram:
         program = super().load_artifact(artifact)
         assert isinstance(program, gtx_wfddecoration.DaCeDecoratedProgram)
         # Inject the backend-level workspace so it is used when arguments are constructed.
         program.set_external_workspace(self.external_workspace or {})
```
philip-paul-mueller, line 47: When the `nanobind` interface is used this is no longer possible as `set_external_workspace()` needs the symbols.
Also when using the `ctypes` interface you need to make sure that `CompiledSDFG.__call__()` (`construct_arguments()` is not enough) has been called.

src/gt4py/next/program_processors/runners/dace/workflow/common.py
```diff
@@ -34,6 +34,10 @@
 """DaCe datatype of `SDFG_ARG_METRIC_COMPUTE_TIME` argument."""
 
 
+#: SDFG scalar argument name for the external synchronization stream pointer.
+SDFG_ARG_EXTERNAL_SYNC_STREAM: Final[str] = "__external_sync_stream"
```
philip-paul-mueller, line 38: I do not like to pass it that way, there should be a DaCe API to do that.
However, I am kind of fine to do that.

src/gt4py/next/program_processors/runners/dace/workflow/decoration.py
```diff
@@ -101,3 +133,17 @@ def set_external_workspace(self, external_workspace: gtx_wfdcommon.ExternalWorks
         This method should be called before the first call to the program.
         """
         self._fun.external_workspace = external_workspace
+
+    def set_external_sync_stream(self, external_sync_stream: Any | None) -> None:
```
philip-paul-mueller, line 137: This function should also invalidate `self._fun.csdfg_argv` such that a stream that is changed is also passed to the underlying call.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -95,51 +95,176 @@ def find_constant_symbols(
...
+        for node, _ in sdfg.all_nodes_recursive()
+    )
 
-    This means that `CompiledSDFG.fast_call()` will return immediately after all
-    computations have been _scheduled_ on the device. This function only has an effect
-    for work that runs on the GPU. Furthermore, all work is scheduled on the
-    default stream.
 
-    Todo: Revisit this function once DaCe changes its behaviour in this regard.
+def add_synchronization(sdfg: dace.SDFG, *, gpu: bool, blocking: bool, n_streams: int) -> None:
```
philip-paul-mueller, line 106: I would name this function `add_synchronization_with_external_stream()` it is supper long but you know what it is doing.
Furthermore, you will not get into trouble when you later add another function that "synchronizes with something".

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -95,51 +95,176 @@ def find_constant_symbols(
...
+                f"for (int i = 0; i < {n_streams}; ++i) {{",
+                f"    {dace_gpu_backend}StreamWaitEvent(__state->gpu_context->streams[i], "
+                f"__state->gpu_context->events[i], 0);",
+                "}",
+            ]
+        )
+
+        # Asynchronous exit barrier: record per-stream done events and make the
+        # external stream wait on them.
+        exit_code = "\n".join(
```
philip-paul-mueller, line 210: Has the same loop bound issues as the one above.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -95,51 +95,176 @@ def find_constant_symbols(
...
+                f"__state->gpu_context->streams[i]);",
+                "}",
+                f"for (int i = 0; i < {n_streams}; ++i) {{",
+                f"    {dace_gpu_backend}StreamWaitEvent(({dace_gpu_backend}Stream_t){stream_arg}, "
+                f"__state->gpu_context->events[i], 0);",
+                "}",
+            ]
+        )
+
+    if blocking:
```
philip-paul-mueller, line 221: As outlined above I am not sure if this will have an effect, i.e. if `blocking` is `False` as I suspect that DaCe will ad some synchronization even when `nosync` is specified, but I might be wrong.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -95,51 +95,176 @@ def find_constant_symbols(
...
+    #   we can disable synchronization for the SDFG state.
+    entry_state.nosync = blocking
+    exit_state.nosync = blocking
+
+    # Wire the entry state before the original start block and the exit state
+    # after all original sink nodes.
+    sdfg.add_edge(entry_state, original_start, dace.InterstateEdge())
+    sdfg.start_block = sdfg.node_id(entry_state)
+
+    for sink_node in sdfg.sink_nodes():
```
philip-paul-mueller, line 259:
```suggestion
    for sink_node in list(sdfg.sink_nodes()):
```
Not sure if needed.

docs/development/ADRs/0028-External_Stream_Synchronization.md
```diff
@@ -0,0 +1,135 @@
...
+
+- **Entry tasklet** (`sync_entry_tlet`): records per-stream done events on each
+  internal DaCe stream and makes the external sync stream wait on them before
+  touching the external workspace.
+- **Exit tasklet** (`sync_exit_tlet`): records per-stream done events on each
+  internal stream and makes the external sync stream wait on them before the SDFG
+  call returns.
+
+When `async_sdfg_call=False`, both tasklets additionally call
+`cudaStreamSynchronize` on the external sync stream, so the SDFG call blocks
```
philip-paul-mueller, line 77: What is the motivation that `sync_exit_tlet` also syncs with the external stream.
I would say only the `sync_enter_tlet` needs to synchronize with the external stream.

docs/development/ADRs/0028-External_Stream_Synchronization.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Decision
+
+We introduce an optional synchronization layer with the following design.
+
+### Configuration
+
+- Add `max_concurrent_gpu_streams` as a parameter to `DaCeWorkflowFactory` and
+  the `make_dace_backend` helper. It maps to DaCe's
+  `compiler.cuda.max_concurrent_streams`. A value of `0` keeps the historical
```
philip-paul-mueller, line 40: I think any value `!= 0` is meaningless since it will hit bugs in the stream generation thing.
I mean even with a stream pool of size 1 it fails.

Thus multiple streams are essentially useless, except we explicitly target the new code generator from the beginning.

docs/development/ADRs/0028-External_Stream_Synchronization.md
```diff
@@ -0,0 +1,135 @@
...
+- The feature introduces additional CUDA/HIP event overhead per call, but only
+  when multi-stream scheduling is explicitly requested.
+- The `cupy` dependency for stream validation is isolated to GPU execution paths;
+  CPU-only installations are unaffected.
+
+## Alternatives considered
+
+- **Device-wide synchronize**: simpler, but serializes the entire device and
+  defeats the purpose of multi-stream scheduling.
+- **Per-call stream argument**: more flexible, but requires changing the
```
philip-paul-mueller, line 128: I mean _technically_ you changed the signature, the only difference is that instead of passing it at call you pass it differently.
What you essentially have done is `d = functool.partial(decorated_function, external_stream)` with the exception that `d` and `decorated_function` are the _same_ object.

src/gt4py/next/program_processors/runners/dace/workflow/translation.py
```diff
@@ -95,51 +95,176 @@ def find_constant_symbols(
...
+        # Add a scalar symbol for the stream handle. It is passed as a 64-bit integer
+        # pointer from Python and cast to the appropriate CUDA/HIP handle type inside
+        # the generated tasklets.
+        stream_arg = gtx_wfdcommon.SDFG_ARG_EXTERNAL_SYNC_STREAM
+        sdfg.add_symbol(stream_arg, dace.uint64)
+
+        # Asynchronous entry barrier: make the internal streams wait on the external stream.
+        entry_code = "\n".join(
+            [
+                f"for (int i = 0; i < {n_streams}; ++i) {{",
```
philip-paul-mueller: This has an implicit assumption that you have always more events than streams (which is likely bit not a given).
Furthermore, you abuse the internal DaCe event here, which is not a problem since once an event is enqueued it can be overwritten, but still.
Also I am not sure, but I think you only need one event, but you have to check.

In my opinion this is indicating that you need support from within DaCe and that this should belong into DaCe itself.

philip-paul-mueller (commented):
There are multiple things here.
- I think that "synchronization" with an external stream, cannot be done in GT4Py and you essentially need DaCe support for that.
- Then "not using the default stream (`compiler.cuda.max_concurrent_streams != -1`)" and "making it asynchronous" will not work.
    Maybe it is possible to set `nosync` of the _inner_ states (i.e. the ones that do computations) to `False`, but this is only safe if you have a single state.
- Also concerning not using the default stream, i.e. `compiler.cuda.max_concurrent_streams != -1`, is essentially not possible with the current GPU code generator.
    Thus targeting it is in my opinion a futile endeavor and you should target the new codegen Yakup is currently making directly.

edopao:
> There are multiple things here.
> 
>     * I think that "synchronization" with an external stream, cannot be done in GT4Py and you essentially need DaCe support for that.

I don't really understand why. On the other hand, as you commented and I explain below, probably synchronization is not needed for `compiler.cuda.max_concurrent_streams != -1`, because the code generated by DaCe is already synchronized.

>     * Then "not using the default stream (`compiler.cuda.max_concurrent_streams != -1`)" and "making it asynchronous" will not work.
>       Maybe it is possible to set `nosync` of the _inner_ states (i.e. the ones that do computations) to `False`, but this is only safe if you have a single state.

__will not work__ because DaCe does not implement it, right? Yes, it is probably useless and I should remove the support. I also realize now that what we need is `compiler.cuda.max_concurrent_streams == -1` but using a different stream than the default one. This would allow us to use the same stream and the same memory pool as the one used in the ICON OpenACC driver (or the Python-CuPy code in ICON-greenline).


>     * Also concerning not using the default stream, i.e. `compiler.cuda.max_concurrent_streams != -1`, is essentially not possible with the current GPU code generator.
>       Thus targeting it is in my opinion a futile endeavor and you should target the new codegen Yakup is currently making directly.

The case `compiler.cuda.max_concurrent_streams != -1` in this PR is targeting the new code generator. The current code generator only supports `compiler.cuda.max_concurrent_streams == -1`. I can make it more explicit.

docs/development/ADRs/0028-External_Stream_Synchronization.md
edopao: I see your point, but I don't know how to make it a call property.

Also consider that creating backends should be *cheap*, and I think that the ICON4Py dycore/diffusion granules actually create one _custom_ backend for each program, see https://github.com/C2SM/icon4py/blob/main/model/common/src/icon4py/model/common/model_options.py

edopao:
I converted this PR to draft because the code generator in DaCe does not support multi-stream GPU scheduling yet.

edopao:
This change is not planned.

PR #2746, opened 2026-08-04
build: update uv lock file

havogt:
Refresh all locked dependencies via `uv lock --upgrade`: 110 packages updated, 4 added, 6 removed.

Notable moves: `mypy 1.19.1 -> 2.3.0`, `ruff 0.15.7 -> 0.16.1`, `numpy 2.4.3 -> 2.5.1`, `jax 0.9.2 -> 0.11.0`, `setuptools 82 -> 83`, `scipy 1.17.1 -> 1.18.0`, `cupy 14.0.1 -> 14.1.1`, `xarray 2026.2.0 -> 2026.7.0`. `dace` and `pybind11` are unchanged.

The two QA tools are major bumps and turned `pre-commit run -a` red, so the second commit adapts the sources.

### ruff 0.16

New/expanded rules produced 31 errors on a previously clean tree:

- 19x `Union[None, X]` reordered to put `None` last
- 9x implicit string concatenation inside a collection literal, parenthesized explicitly
- `FieldOffset` was listed twice in `gt4py.next.__all__`
- `gtir_to_sdfg` was imported both at module level and under `TYPE_CHECKING` in `gtir_to_sdfg_primitives`
- a trailing comma in `iterator/type_system/inference` made an assert message a 1-tuple instead of a string

ruff 0.16 also formats code blocks inside markdown, which reformats six ADRs. CI runs `pre-commit --all-files`, so this is required rather than optional.

### mypy 2.3

18 errors:

- 7 `type: ignore`s are now unused, including four in `instrumentation/metrics.py` that explicitly waited on python/mypy#20573, which 2.3 ships
- 5 casts are redundant now that mypy narrows properly after the preceding `isinstance` (three of them already carried a comment saying the cast shouldn't be needed)
- `Metric.mean` / `Metric.std` return `float` to match the numpy 2.5 stubs; at runtime these were and remain `np.float64`, and the only callers format them or wrap them in `float()`
- `foast_passes/type_deduction`: the narrowing assert is split off the declared-type assignment
- `next/utils.tree_map`: `isinstance(x, collection_type)` with a variable class object now narrows `Any` to `object`, so the length checks needed restructuring
- `cartesian/utils/compiler`: `compiler_cxx` exists at runtime but is not declared on the `CCompiler` base class, so it gets a targeted ignore — the surrounding `except AttributeError` already handles the non-Unix case


src/gt4py/next/instrumentation/metrics.py
```diff
@@ -89,13 +89,13 @@ def __post_init__(self) -> None:
             object.__setattr__(self, "name", sys.intern(self.name))
 
     @property
-    def mean(self) -> np.floating:
+    def mean(self) -> float:
         if len(self.samples) == 0:
             raise ValueError("Cannot compute mean of empty sample list.")
         return np.mean(self.samples)
```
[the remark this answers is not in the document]
havogt: Not taking this one — the premise doesn't hold. `np.float64` is a subclass of Python `float`:

```python
>>> issubclass(np.float64, float)
True
>>> type(np.mean([1.0, 2.0]))
<class 'numpy.float64'>
```

So the value returned at runtime already satisfies the `float` annotation; there is no divergence between the contract and the implementation. `-> float` is also exactly what the numpy 2.5 stubs declare for `np.mean` over a `list[float]`, which is why mypy 2.3 rejected the previous `np.floating` annotation in the first place.

The `float(metric.mean)` calls in the tests are defensive, not a workaround for a mismatch.

The one case that would break the annotation is a `np.float32` sample, since that is *not* a `float` subclass. But no such path exists: every producer passes a Python float — `runners/gtfn.py` passes `exec_info["run_cpp_duration"]`, dace's `workflow/decoration.py` passes `collect_time_arg[0].item()`, and `compute_metric` has no implementations beyond the timing collectors. Adding a conversion here would be dead weight in a dependency-bump PR.

pyproject.toml
```diff
@@ -203,7 +203,12 @@ plugins = ['gt4py.next.type_system.mypy_plugin']
 show_column_numbers = true
 show_error_codes = true
 warn_redundant_casts = true
-warn_unused_configs = true
+# `warn_unused_configs` reports every override that the current invocation did not
+# reach, so it fires on any partial run: `mypy src/` never reaches the
+# `typing_tests.*` sections, and the `typing_tests` suite never reaches the
+# cartesian ones. The note is part of mypy's output, which breaks the exact-output
+# assertions in `typing_tests/`.
+warn_unused_configs = false
```
havogt: Can you double-check this @egparedes (it's responsible for the failure in the typing tests before this was added). All the other changes look good to me.

havogt:
Mainly did this version bump because we lock a package with security issue (according to dependabot), so I bumped everything.

pyproject.toml
egparedes: It seems reasonable but I think I would just remove the comment or make it shorter.

egparedes (approved):
LGTM, just a minor non-blocking complaint about a verbose comment.

pyproject.toml
```diff
@@ -203,7 +203,12 @@ plugins = ['gt4py.next.type_system.mypy_plugin']
 show_column_numbers = true
 show_error_codes = true
 warn_redundant_casts = true
-warn_unused_configs = true
+# `warn_unused_configs` reports every override that the current invocation did not
+# reach, so it fires on any partial run: `mypy src/` never reaches the
+# `typing_tests.*` sections, and the `typing_tests` suite never reaches the
+# cartesian ones. The note is part of mypy's output, which breaks the exact-output
+# assertions in `typing_tests/`.
```
havogt:
```suggestion
```

PR #2754, opened 2026-08-06
feat[eve,next]: runtime PEP 695 type alias support and located annotation diagnostics

egparedes:
## Description

Python 3.12 (PEP 695) `type X = ...` binds a `typing.TypeAliasType` rather than
the annotation itself, and nothing unwraps it at runtime: `get_type_hints`
returns it as is, `__value__` is lazy and raises `NameError` until the names it
mentions exist, and a plain `isinstance(x, xtyping.TypeAliasType)` misses
native aliases because `eve.extended_typing` star-imports both `typing` and
`typing_extensions`. Now that the floor is 3.12, users write these in datamodel
fields and DSL signatures.

**`eve`.** `extended_typing` exposes `is_type_alias()` and `eval_type_alias()`,
and resolution happens at the three annotation-dispatch funnels
(`type_validation.SimpleTypeValidatorFactory.__call__`,
`datamodels.core._make_type_converter`, `get_represented_types`) rather than at
annotation-storage time, so nesting (`list[MyAlias]`) is covered for free and
the stored annotation keeps the alias name for reprs and signatures.

Failures split on whether the annotation can still become valid. `NameError`
defers to first instantiation via `ForwardRefValidator`, or the new
`DeferredTypeConverter` for coerced fields, exactly as string forward
references already do. Anything else is permanent — an alias value is arbitrary
user code, so a typo'd dtype raises `AttributeError` — and is wrapped into a
`TypeError` naming the alias and the cause. Cycles are detected by the set of
visited aliases, with the depth cap (64) kept as the backstop for aliases that
grow on every step (`type G[T] = G[Tuple[T]]`).

**`next`.** `from_type_hint` resolves aliases, and annotation failures reach
the user as a located `InvalidAnnotationError` instead of a bare `ValueError`
or a raw traceback, through a shared `type_from_annotation()` in
`dialect_parser`. Two span defects are fixed: `StringifyAnnotationsPass` left
rebuilt annotations with no location at all, crashing `visit_AnnAssign` on
every annotated assignment inside a field operator; and both return-annotation
diagnostics pointed at the whole function definition. This reorders one case —
a bad return annotation is now reported before body type-deduction errors.

### For reviewers

- `get_represented_types` fails *silently*: an unresolved alias matches none of
  its branches and falls through to its empty-tuple return, which would quietly
  empty the import-time isinstance tuples in `next/common.py` and
  `next/named_collections.py`. It had no test coverage before this PR.
- `DSLTypeError` loses its `__init__`. That override was a pass-through with a
  *narrower* signature than `DSLError`, so removing it only widens what is
  accepted; it is required for subclasses to pass `label` / `notes` / `hints`.
- Aliases are supported *as annotations*. One also used as a runtime value
  (`common.Tag`, subclassed at `iterator/embedded.py:102`) cannot be rewritten
  to `type X = ...` at all, so ruff's `UP040` stays off repo-wide. PEP 695
  generics already worked and are unchanged.

## Requirements

  - By request, the rationale lives in comments next to the code in
    `eve/extended_typing.py` instead; the `ADRs/eve/` hierarchy added earlier
    on this branch was removed.


src/gt4py/next/errors/exceptions.py
```diff
@@ -183,8 +183,9 @@ def __init__(self, location: Optional[SourceLocation], arg_name: str, is_kwarg:
 
 
 class DSLTypeError(DSLError):
-    def __init__(self, location: Optional[SourceLocation], message: str) -> None:
-        super().__init__(location, message)
+    # Note: no '__init__' override, so that subclasses inherit the full structured
+    # signature of 'DSLError' ('label', 'related', 'notes', 'hints').
+    pass
```
egparedes, line 188: Doesn't this change break the already existing custom Exception classes defined in this module?

src/gt4py/eve/extended_typing.py
```diff
@@ -408,6 +411,70 @@ def is_actual_type(obj: Any) -> TypeGuard[Type]:
...
+#: ``type X = ...`` alias is not an instance of the ``typing_extensions`` backport (nor
+#: the other way round), so both of them have to be checked.
+_TypeAliasTypes: Final[tuple[type, ...]] = (
+    _typing.TypeAliasType,
+    _typing_extensions.TypeAliasType,
+)
+
+#: Upper bound for the number of resolution steps in `eval_type_alias`, to avoid
+#: hanging on recursive aliases like ``type A = A``.
+_MAX_TYPE_ALIAS_DEPTH: Final = 64
```
egparedes: Is this the only or the best way to detect definition cycles in recursive aliases?

egparedes (commented):
A couple of questions

src/gt4py/next/errors/exceptions.py
egparedes: No — it widens the accepted signature rather than narrowing it, so nothing that worked before stops working.

The override I removed was a pure pass-through with a *narrower* signature than its parent:

```python
def __init__(self, location: Optional[SourceLocation], message: str) -> None:
    super().__init__(location, message)
```

`DSLError.__init__` is `(location, message, *, label=None, related=(), notes=(), hints=())`. Every call that satisfied the old two-argument override still satisfies the inherited one, so all three subclasses (`MissingParameterAnnotationError`, `InvalidParameterAnnotationError`, `InvalidAnnotationError`) and the four direct `DSLTypeError(loc, msg)` call sites in `func_to_foast.py`, `otf/arguments.py` and `otf/compiled_program.py` are unaffected. Verified by instantiating each.

The removal was also required rather than incidental: the new `InvalidAnnotationError` passes `label` / `notes` / `hints` up to `DSLError`, and the old narrow override rejected those with `TypeError: got an unexpected keyword argument 'label'`. That is what the comment left in its place is trying to record — keeping the override would silently cap every future `DSLTypeError` subclass at the unstructured two-argument form.

src/gt4py/eve/extended_typing.py
egparedes: Good catch — it was neither, and I have changed it in 6edd8dd10.

A depth bound alone cannot tell a cycle from a legitimately deep chain, so `type A = A` and a mutual `A -> B -> A` were both reported as *"recursive or nested too deeply"*. Keeping the set of aliases already walked through detects a real cycle exactly, as soon as one alias is seen twice, however long the cycle is:

```python
seen: set[TypeAliasType] = set()
...
if is_type_alias(annotation):
    if annotation in seen:
        recursive = True
        break
    seen.add(annotation)
    annotation = annotation.__value__
```

It cannot *replace* the bound, though, which is why both are now there. A parametrized alias that grows on every step builds a bigger annotation each time and never repeats one:

```python
type G[T] = G[Tuple[T]]
# G[int] -> G[Tuple[int]] -> G[Tuple[Tuple[int]]] -> ...
```

The visited-alias check never fires for that, so the bound is what stops it. The message now distinguishes the two cases (`recursive definition` vs `nested too deeply`), and there is a test for each.

One trap worth recording: tracking `id()` instead of the objects is unsound here. The intermediate annotations are temporaries, so a freed id gets reused and a cycle is reported where there is none — I hit exactly that at step 3 of the `G[T]` chain above while trying it.

src/gt4py/eve/extended_typing.py
```diff
@@ -408,6 +411,136 @@ def is_actual_type(obj: Any) -> TypeGuard[Type]:
...
+# constructor, an `isinstance()` argument) cannot be written this way at all, since a
+# `TypeAliasType` is not the class it stands for. The latter is why ruff's `UP040`
+# cannot be enabled repo-wide.
+
+#: Both implementations of PEP 695 type aliases. They are distinct classes, and a native
+#: ``type X = ...`` alias is not an instance of the ``typing_extensions`` backport (nor
+#: the other way round), so both of them have to be checked. Note that a `hasattr(obj,
+#: "__value__")` check is not equivalent: ``MyGenericAlias[int]`` is a `types.GenericAlias`
+#: which proxies attribute lookups to its origin, so it passes that check but is not an
+#: alias itself.
```
egparedes: Try to make it concise

egparedes: Done in fb6e8a5b8 — cut roughly in half (46 lines out, 23 in), across all four comment blocks from that commit, not just this one.

Dropped: the opening paragraph on `get_type_hints` not unwrapping aliases (restates PEP 695), "nesting covered for free", and the restatement of what `datamodels` does with the deferral.

Kept the parts that are not evident from the code:

- resolution happens at the dispatch funnels, and why not eagerly (breaks aliases defined before their target)
- a funnel left out fails *silently* via the empty-tuple branch — the one that actually bites
- `hasattr(obj, "__value__")` is not equivalent to the `isinstance` check
- `NameError` defers, everything else is permanent and fails on the spot
- the two unsupported cases, including why `UP040` stays off

The module-level block is now 12 lines and the two `except` comments 4-5 each.

src/gt4py/eve/extended_typing.py
```diff
@@ -408,6 +411,113 @@ def is_actual_type(obj: Any) -> TypeGuard[Type]:
...
+        # resolve, so they fail here rather than being deferred. Wrapping them into one
+        # type lets consumers keep a single narrow 'except' and propagate this message
+        # verbatim, as it is the only text naming the actual cause.
+        raise TypeError(
+            f"Type alias '{original_annotation}' cannot be resolved ({error})."
+        ) from error
+
+    raise TypeError(
+        f"Type alias '{original_annotation}' cannot be resolved "
+        f"({'recursive definition' if recursive else 'nested too deeply'})."
```
havogt, line 517: Are the 2 cases the only way we can reach here?

src/gt4py/eve/type_validation.py
```diff
@@ -199,6 +199,20 @@ def __call__(
             ):  # see https://github.com/python/cpython/issues/105499
                 type_annotation = typing.Union[type_annotation.__args__]
 
+            # PEP 695 type aliases ('type X = ...') stand for another annotation whose
+            # value is only evaluated on demand. A 'NameError' raised here means the
+            # alias references a name which does not exist yet, and is deliberately not
+            # converted: 'datamodels' defers the creation of the validator when it sees
+            # one, in the same way it does for forward references.
```
havogt: this comment is duplicated with minor modifications in many places. I think it's good to have this information, but I can't judge if it's repeated to often or if it should be in one location and then referenced.

src/gt4py/next/errors/exceptions.py
```diff
@@ -183,8 +183,9 @@ def __init__(self, location: Optional[SourceLocation], arg_name: str, is_kwarg:
```
havogt, line 183: I assume the TODOs in this file above about Python >=3.12 are left out on purpose to be resolved in a later PR? because scope here is really only about the `type` alias?

src/gt4py/next/ffront/func_to_foast.py
```diff
@@ -182,18 +182,34 @@ def _postprocess_dialect_ast(
...
 
+    @staticmethod
+    def _returned_value_location(foast_node: foast.FunctionDefinition) -> eve.SourceLocation:
+        """Locate the expression the deduced return type comes from.
+
+        Falls back to the whole function definition when the type cannot be pinned on
+        a single expression, i.e. when the function returns from more than one place.
+        """
+        returns = foast_node.walk_values().if_isinstance(foast.Return).to_list()
+        return returns[0].value.location if len(returns) == 1 else foast_node.location
```
havogt: free function? or is this the style of the current file?

src/gt4py/next/ffront/func_to_foast.py
```diff
@@ -215,6 +231,18 @@ def visit_FunctionDef(self, node: ast.FunctionDef, **kwargs: Any) -> foast.Funct
...
+        # and could point no further than the whole function definition. It compares
+        # the annotated and deduced types, which is why the check is split rather than
+        # moved: the deduced type does not exist until the body has been typed.
+        if node.returns is not None and "return" in self.annotations:
+            type_from_annotation(
+                self.annotations["return"],
+                self.get_location(node.returns),
+                description="return type annotation",
+            )
+
```
havogt: we are checking that it's a valid annotation, but we are are not doing anything with it? Not sure how the code around here works.

src/gt4py/next/ffront/func_to_foast.py
```diff
@@ -182,18 +182,34 @@ def _postprocess_dialect_ast(
 
         # check deduced matches annotated return type
         if "return" in annotations:
-            annotated_return_type = type_translation.from_type_hint(annotations["return"])
+            # The annotation itself was already validated in 'visit_FunctionDef', where
+            # its source location was still available, so this cannot raise.
```
havogt, line 196: I don't understand this and the corresponding check/comment around lines R240 to R245.

src/gt4py/next/errors/exceptions.py
egparedes: Yes, exactly. They are fixed in the other PR #2755

src/gt4py/eve/extended_typing.py
egparedes: Yes — the loop has exactly three exits, and two of them land here:

- `return annotation` when the annotation is not (or no longer) an alias — the only success path;
- `break` when an alias is seen twice, setting `recursive = True`;
- the `range(_MAX_TYPE_ALIAS_DEPTH)` running out.

Both `except` clauses re-raise, so nothing falls through from them either. So reaching this line means the second or third case, which is what the message distinguishes.

src/gt4py/eve/type_validation.py
egparedes: Agreed, and fixed in 7b4181b3c — it was in four places with small variations.

The full explanation now lives once in `eval_type_alias`, at the `except NameError` branch where the signal originates, and the sites that react to it carry a one-liner pointing there:

- `type_validation.py`: *"A `NameError` is deliberately left to propagate here, so that `datamodels` can defer; see `xtyping.eval_type_alias` for both failure modes."*
- `datamodels.field_type_validator_factory`: *"Deferral signal from `xtyping.eval_type_alias` (see its `except NameError` branch); handled like a forward reference."*
- `datamodels._make_type_converter`: *"Deferral signal, as in `field_type_validator_factory` above."*

Also trimmed the restatement out of the `DeferredTypeConverter` docstring.

src/gt4py/next/ffront/func_to_foast.py
egparedes: Free function is right — it used neither `cls` nor `self`. Moved to module level in 7b4181b3c, next to the other module-level helpers, which also matches `dialect_parser._describe_unsupported_feature`. It was the only `@staticmethod` in the file.

src/gt4py/next/ffront/func_to_foast.py
egparedes: You read it correctly, and the result really was discarded — the call was there only to raise. That is too subtle, so 7b4181b3c gives it a name that says so, `_reject_invalid_return_annotation(node)`, returning `None`.

Why the annotation is translated twice: `_postprocess_dialect_ast` needs the *value* to compare against the deduced return type, and that comparison can only happen after the body has been typed. But by then the `ast` is gone, so if the translation fails there it can point no further than the whole function definition — which is what produced the 63-caret diagnostic this PR is fixing. So the invalid case is rejected in `visit_FunctionDef`, where `node.returns` still carries the source location, and the second translation no longer raises.

The alternative would be threading the annotation's `SourceLocation` through `apply` into `_postprocess_dialect_ast`, which widens the shared `DialectParser` signature and reaches `func_to_past.py` too. Happy to go that way if you prefer it, but for one location it seemed the heavier option.

src/gt4py/next/ffront/func_to_foast.py
egparedes: Same construct as [the R240-R245 thread](https://github.com/GridTools/gt4py/pull/2754#discussion_r3750351137) — the split confused you in both places, so it needed fixing rather than explaining. See that reply for the reasoning; 7b4181b3c names the earlier half `_reject_invalid_return_annotation` and documents the split in its docstring, so this line should now read as "translate again for the comparison, which cannot fail".

havogt (approved):
lgtm

Issue #2406, opened 2025-11-27
cartesian: drop support for `field[...]` syntax with ellipsis in brackets


romanc:
# Description

In https://github.com/GridTools/gt4py/pull/2399, we realized that the following is legal gt4py syntax

```py
with computation(PARALLEL), interval(...):
    out_field[...] = 42
```

Notice the ellipsis in the square brackets `[...]`, which are verbatim and not to abbreviate something. This is redundant with

```py
with computation(PARALLEL), interval(...):
    out_field = 42
```

To get ride of this redundancy, we thus decided to remove support for `out_field[...]`, especially since it starts to make problems with supporting python >=3.14 and <3.14.

In PR https://github.com/GridTools/gt4py/pull/2399, Rico just removed the problematic test case. This issue is a follow-up to fully remove support from the parser of `gt4py.cartesian`.

romanc:
Dropped support in #2820

PR #2755, opened 2026-08-06
fix[cartesian, eve, next]: latent bugs and deprecated typing shims surfaced by the Python 3.12 floor

egparedes:
## Description

Auditing the codebase after the floor moved to 3.12 (#2326) turned up several real
bugs, plus deprecated compatibility code, configuration and documentation that still
assumed an older floor. Merged up to date with `main`.

> [!IMPORTANT]
> @romanc — `main` is currently broken by #2772, and this PR contains the fix
> (still reproducible on `main` as of `0d8fb34d3`).
> That PR replaced the ellipsis check with `isinstance(cn, types.EllipsisType)`, but
> that is the type of the `...` *object*, not of its AST node (an `ast.Constant` whose
> value is `Ellipsis`), so the condition is never true. On current `main`:
> ```
> field_b[...] = field_a
> → GTScriptSyntaxError: Unexpected type found <class 'ellipsis'>.
>   Expected one of: int, AxisIndex, string (var ref), or None.
> ```
> The merge here keeps this branch's `_is_ellipsis_node` helper, which matches
> correctly on every supported version, and takes #2772's `visit_Expr` (better than
> what this branch had: it drops only *string* constant statements, not every
> constant). Nothing else from #2772 is reverted.

### Bugs fixed

- **`cartesian`: `'...'` index detection was version-dependent.** Originally
  `getattr(ast, "Ellipsis", types.EllipsisType)`, which resolved to the deprecated
  `ast.Ellipsis` on 3.12/3.13 (a `DeprecationWarning` on every import) and, on 3.14,
  fell back to a type that never matches. Now matched via `_is_ellipsis_node`, reused
  by the vertical-interval parser. See the note above for the interaction with #2772.

- **`next`: `_type_conversion_helper` rejected PEP 604 unions.** `A | B` is a
  `types.UnionType` with no `__origin__`, so the `__origin__ is Union` branch never
  matched and control reached `raise AssertionError("Illegal type encountered.")`;
  the builtin `tuple` spelling was missed the same way. Now dispatches on
  `get_origin()`. This also unblocks ruff's `UP006` (relevant to #2224).

- **`eve`: `frozen="strict"` was completely broken.** It called a non-existent
  `xtyping.is_hashable_type`, so *every* strict-frozen datamodel raised
  `AttributeError` at class definition — dead since #1146 (2023), unnoticed because
  the flag is used nowhere, had no tests, and the star-import plus module
  `__getattr__` in `extended_typing` hides a missing name from both ruff and mypy.
  Renaming was not enough: the check now decomposes composite annotations, so
  wrapping a mutable type in a union, a container (`tuple[list[int], ...]`), or a
  `TypeVar` bound is no longer a way around it, and bare `tuple`/`frozenset` are
  rejected for the same reason as `tuple[Any, ...]`. PEP 695 aliases are resolved and
  checked through what they stand for, so `type Pair = tuple[int, int]` is accepted
  like the plain annotation while one hiding a mutable type is still rejected; an
  alias that cannot be evaluated proves nothing and is rejected rather than escaping
  as a raw `NameError`. The error names the offending fields and their datamodel.

- **`next`: the typing-export tests were not really parametrized.**
  `test_typing_exports` runs on 3.12/3.13/3.14, but pytest-mypy-plugins passes no
  `--config-file`, so mypy walked up out of its temp directory and picked up the
  project `[tool.mypy]` table — pinning every run to the `python_version` floor and
  making the 3.13/3.14 runs duplicates of 3.12. The session now has its own
  `typing_tests/mypy.ini`, which sets no `python_version` and, since these tests
  model *downstream client* code, carries over only `implicit_reexport` (the setting
  that makes them check exports at all).

- **`next`: multiple compilation failures lost all but one traceback.**
  `wait_for_compilation` flattened them into a single `RuntimeError`, leaving
  failures 2..n as `repr` text in the message. They are now raised as a group, so
  each keeps its own traceback. `BaseExceptionGroup` is the constructor, since
  `Future.exception()` is typed `BaseException`; it yields a plain `ExceptionGroup`
  whenever every member is an `Exception`, so `except*` on `Exception` still catches
  it. The multi-failure path had no test before.

### Deprecated and dead code removed

- **`eve`: the PEP 585 `typing` aliases are no longer re-exported.** `Dict`,
  `FrozenSet`, `List`, `Set`, `Tuple` and `Type` are deprecated spellings of the
  builtin generics on a 3.12 floor; the use sites move to `dict`, `list`, … Note they
  are *not* created by the re-export block — the `typing` / `typing_extensions` star
  imports bind them and the module `__getattr__` forwards anything missing — so
  deleting the block alone would have left all six in place, silently rebound from
  the builtins to the deprecated `typing` objects. They are therefore dropped from
  the namespace and rejected explicitly, with a message naming the replacement. A
  bare `typing.Type` means `type[Any]`, so unsubscripted occurrences become that
  rather than the stricter bare `type`.
  Forward-reference resolution is unchanged: these names stay valid in user-written
  annotations, so `List[int]` and `typing.List[int]` still resolve to `list[int]`.
  The re-exports from `collections.abc`, `collections`, `contextlib` and `re` are
  deliberately kept — there the name is already the modern one, only its `typing`
  home is deprecated.

- **Dead version guards.** An `if sys.version_info >= (3, 9)` block, an always-true
  `isinstance(typing.Any, type)` guard, and two branches for
  `typing_extensions.Any` being a distinct object from `typing.Any` — true only
  below 3.11, and `typing_extensions` re-exports `typing.Any` on 3.12, 3.13 and 3.14
  alike. In `next/errors/exceptions.py`, the `add_note` override and `__notes__`
  fold-in, both dead since PEP 678 landed in 3.11; `DSLError` now inherits
  `add_note`, and a test that ran on no supported version is deleted. Also fixes
  `dir()` on `extended_typing`, which listed only `typing` names and hid the
  module's own API.

### Diagnostics

`try*` is named as a construct instead of falling back to a generic `ast` class name.
Constructs *newer* than the supported floor can now be catalogued too: naming
`ast.TemplateStr` (PEP 750 t-strings, 3.14) directly would break the import on
3.12/3.13, so they are registered by name and skipped where the interpreter lacks
them — a t-string now reports `Unsupported Python syntax: t-string` on 3.14.

### Build and docs

`.gitpod.Dockerfile` still based the workspace on Python 3.11, so `.gitpod.yml` would
build a venv the project cannot be installed into. The noxfile's PEP 723 block
declared `requires-python = ">=3.11"`. `[tool.mypy]` declared no `python_version`, so
the type-check floor followed whichever interpreter ran mypy. Drops a mypy override
and an isort `known-third-party` entry that reference things which do not exist.
`docs/development/next/error-messages.md` was advising contributors to write
needlessly compatible code and misdescribed where `add_note` breadcrumbs go.

## Follow-ups

Filed rather than fixed here, to keep this PR scoped:

- #2758 — unsupported-syntax diagnostics are preempted by closure-variable
  resolution, so `try` / `except ValueError` reports `DSLTypeError: Unexpected
  object 'ValueError' of type '<class 'type'>' encountered.` instead of its
  catalogued message. This is why the `ast.TryStar` entry is pinned at the catalogue
  level rather than end-to-end: `try*` grammatically requires an `except*` handler
  naming an exception type, so it can never reach `generic_visit`. TODOs at the root
  cause, the catalogue and the test point here.
- #2759 — `test_unlisted_construct_falls_back_to_ast_name` does not exercise the
  fallback it names, because `ast.JoinedStr` is in the catalogue.

## Requirements

- [ ] Important design decisions have been documented in the appropriate ADR inside the [docs/development/ADRs/](docs/development/ADRs/README.md) folder. — n/a per review, no design decisions here


src/gt4py/eve/datamodels/core.py
```diff
@@ -1019,8 +1015,57 @@ def _type_converter(value: Any) -> _T:
...
+    if origin_type is not None:
+        # Parametrized generic alias ('tuple[int, ...]', 'list[int]', ...). The alias
+        # itself is a hashable object, so it has to be decomposed: both the container
+        # type and every type argument must be strictly immutable, since the hash of
+        # a container folds in the hashes of the items it holds.
+        return _is_strictly_immutable_type(origin_type) and all(
+            _is_strictly_immutable_type(arg) for arg in type_args if arg is not Ellipsis
+        )
+
+    if xtyping.is_actual_type(type_annotation):  # plain type, already known not to be a datamodel
```
egparedes, line 1133: An **unparametrized** hashable container passes the check, so `hash()` can still blow up — which is the exact hole the parametrized branch above was added to close.

```python
@datamodels.datamodel(frozen="strict")
class M:
    values: tuple          # accepted
hash(M(values=([1, 2],)))  # TypeError: unhashable type: 'list'
```

`tuple` / `frozenset` / bare `typing.Tuple` reach this branch with `is_type_with_custom_hash(...) is True`, while the equivalent `tuple[Any, ...]` **is** rejected (via `Any`) — so the same annotation passes or fails depending only on whether it was spelled with type arguments. Suggest rejecting item-hashing containers here unless they arrived as the origin of a parametrized alias, and treating an alias with no `__args__` (`typing.Tuple`) as its bare origin.

src/gt4py/eve/datamodels/core.py
```diff
@@ -1019,8 +1015,57 @@ def _type_converter(value: Any) -> _T:
...
+
+    # Anything else (type variables, forward references, ...) is checked through the
+    # concrete types it stands for, if any. Forward references that cannot be resolved
+    # at this point do not prove anything, so they are rejected.
+    try:
+        represented_types = xtyping.get_represented_types(type_annotation)
+    except NameError:
+        return False
+
+    return bool(represented_types) and all(map(_is_strictly_immutable_type, represented_types))
```
egparedes: This path flattens the annotation to its origins and loses the type arguments, so a mutable payload hidden behind a `TypeVar` bound survives the check:

```python
MutableBoundT = TypeVar("MutableBoundT", bound=Tuple[List[int], ...])

@datamodels.datamodel(frozen="strict")
class M:
    value: MutableBoundT   # accepted -> _is_strictly_immutable_type(...) is True
```

`get_represented_types(Tuple[List[int], ...])` returns `(tuple,)`, and `tuple` alone is hashable, so the `List[int]` argument is never inspected — even though the same annotation written directly on the field is correctly rejected. Recursing on the bound / constraints **as annotations** (and on the resolved forward reference) keeps the decomposition the docstring promises.

src/gt4py/eve/datamodels/core.py
```diff
@@ -1019,8 +1015,57 @@ def _type_converter(value: Any) -> _T:
...
+
+    if xtyping.is_actual_type(type_annotation):  # plain type, already known not to be a datamodel
+        return xtyping.is_type_with_custom_hash(type_annotation)
+
+    # Anything else (type variables, forward references, ...) is checked through the
+    # concrete types it stands for, if any. Forward references that cannot be resolved
+    # at this point do not prove anything, so they are rejected.
+    try:
+        represented_types = xtyping.get_represented_types(type_annotation)
+    except NameError:
```
egparedes: Narrow `except`: `get_represented_types` calls `eval_forward_ref`, which can also raise `TypeError` (non-subscriptable resolved name), `SyntaxError` (malformed annotation string), or trip the `assert not isinstance(actual_type, ForwardRef)` inside `eval_forward_ref`. Any of those escapes `_make_datamodel` raw instead of producing the documented `EveTypeError`, contradicting the docstring's "conservatively rejected".

src/gt4py/eve/extended_typing.py
```diff
@@ -34,53 +34,91 @@
...
+_DEPRECATED_TYPING_ALIASES: Final[Mapping[str, str]] = {
+    "Dict": "dict",
+    "FrozenSet": "frozenset",
+    "List": "list",
+    "Set": "set",
+    "Tuple": "tuple",
+    "Type": "type",
+}
+
+for _alias in _DEPRECATED_TYPING_ALIASES:
```
egparedes, line 101: Popping the names from `globals()` also removes them from the namespace that forward references are evaluated in, which is a silent behaviour change for the public API:

```python
xtyping.eval_forward_ref("List[int]")   # before: list[int]   now: NameError: name 'List' is not defined
```

`eval_forward_ref` builds its dummy `f` inside this module, so `f.__globals__` (this module's `globals()`) is the default `globalns` whenever the caller passes `globalns=None` — which is the case for `get_represented_types`, and for `type_validation.SimpleTypeValidatorFactory` when it resolves a `ForwardRef` field annotation. A downstream model that keeps `from typing import List` under `if TYPE_CHECKING:` used to resolve through this fallback and now raises. Worth either documenting the drop in the release notes or keeping the names reachable for forward-ref evaluation only.

src/gt4py/eve/extended_typing.py
```diff
@@ -34,53 +34,91 @@
...
+    Annotations are resolved through this module, so that 'typing_extensions'
+    definitions take priority and the standard collection types are used. The
+    deprecated builtin aliases are not re-exported here, but they remain perfectly
+    valid in user-written annotations, so 'typing.List[int]' and friends fall back
+    to the real 'typing' module instead of raising.
+    """
+
+    def __getattr__(self, name: str) -> Any:
+        if name in _DEPRECATED_TYPING_ALIASES:
+            return getattr(_typing, name)
```
egparedes: This resolves `typing.List[int]` to the **deprecated** `typing.List[int]` object, whereas before this PR the same forward reference produced `list[int]` (the module bound `List` to `builtins.list`). So resolved annotations coming out of `eval_forward_ref` now differ in identity and `type()` from those produced by every other path, and the module reintroduces exactly the objects it is otherwise rejecting — the class docstring above even says "the standard collection types are used".

`return getattr(_builtins, _DEPRECATED_TYPING_ALIASES[name])` would keep user-written `typing.List[int]` valid **and** preserve the pre-PR result; the new `test_deprecated_typing_alias_still_resolves_in_forward_refs` assertions (`get_origin(...) is list`, `get_args(...) == (int,)`) hold either way.

src/gt4py/eve/extended_typing.py
```diff
@@ -106,15 +150,17 @@ def __getattr__(name: str) -> Any:
...
 
 
-def __dir__() -> List[str]:
+def __dir__() -> list[str]:
     if not hasattr(self_func := (globals()["__dir__"]), "__cached_dir"):
         import typing
 
         import typing_extensions
 
         orig_dir = typing.__dir__()
```
egparedes: `__dir__` is built purely from `typing` + `typing_extensions`, so it never reflects this module's own namespace: `dir(xtyping)` omits `is_actual_type`, `eval_forward_ref`, `infer_type`, `NestedList`, … That also makes the new `assert name not in dir(xtyping)` a weak oracle — it would pass for any name this module simply never had.

Adding `*globals()` to the set (and using set membership instead of the `n not in orig_dir` list scan, which is O(n²) over ~450 names) fixes both.

src/gt4py/eve/extended_typing.py
```diff
@@ -395,7 +441,7 @@ def __pretty__(
     _ArtefactTypes = (*_ArtefactTypes, typing_exts_any)
 
 
-def is_actual_type(obj: Any) -> TypeGuard[Type]:
+def is_actual_type(obj: Any) -> TypeGuard[type[Any]]:
```
egparedes, line 445: Dead sub-floor guard just above (lines 431-441): `isinstance(_typing.Any, type)` is unconditionally true from Python 3.11 on, so on the new 3.12 floor `_ArtefactTypes` can be built unconditionally. This PR removes the `sys.version_info >= (3, 9)` block and the 3.10 diagnostics shims but leaves this one behind, so the "no version shims left over" claim added to `docs/development/next/error-messages.md` is not quite true for `eve`.

(The `typing_extensions.Any` checks right below are guards on the *typing_extensions* version rather than the Python version, so those should stay.)

src/gt4py/eve/extended_typing.py
```diff
@@ -34,53 +34,91 @@
...
+
+
+# The 'typing' aliases of the builtin generics are deprecated since PEP 585 and are no
+# longer re-exported: use the builtin spelling instead. They are not simply absent from
+# this module -- the star imports above bind them, and '__getattr__' below would happily
+# forward them to 'typing' -- so they are dropped from the namespace here and rejected
+# explicitly. Without this, removing them would silently downgrade every use site from
+# the builtin to the deprecated 'typing' object. Note that this is a runtime guarantee
+# only: a type checker still resolves the names through the star imports.
+_DEPRECATED_TYPING_ALIASES: Final[Mapping[str, str]] = {
```
egparedes, line 92: Removing six names that were deliberately re-exported with explicit `X as X` aliases is a public-API-shape change for `gt4py.eve.extended_typing`, and it is enforced at runtime (`AttributeError` / `ImportError`) for downstream users too.

`AGENTS.md` asks for a record of that kind of decision:

> For a non-trivial architectural choice (new dependency, IR change, backend interface, **public-API shape**), add an ADR under the appropriate `docs/development/ADRs/<subsystem>/` directory.

No ADR is added in this PR. A short one under `docs/development/ADRs/eve/` capturing "builtin generics only, `typing` aliases rejected at runtime" would give the rule a citable home (the module comment currently carries all of the rationale).

typing_tests/mypy.ini
```diff
@@ -0,0 +1,21 @@
...
+; Deliberately no 'python_version' here: it must follow the interpreter running the
+; session, so each parametrized run checks the version it claims to check.
+
+[mypy]
+; The plugin is what downstream users are told to enable; see the module docstring of
+; 'gt4py.next.type_system.mypy_plugin'.
+plugins = gt4py.next.type_system.mypy_plugin
+ignore_missing_imports = True
+show_column_numbers = True
+show_error_codes = True
```
egparedes, line 21: This standalone config silently drops `implicit_reexport = false`, which the session used to inherit from the project `[tool.mypy]` table — and that is precisely the setting that makes a *typing-exports* test an export test. Without it, mypy stops reporting `Module "gt4py.next" does not explicitly export attribute "X"`, so a regression where `gtx.<symbol>` is only implicitly re-exported would no longer be caught by the one session whose stated purpose is to catch it.

The rest of the gt4py-internal strictness is rightly left out (it would apply to the client snippets), but this one belongs to the library side. I verified the suite still passes with it enabled:

```suggestion
show_error_codes = True
; Kept from the project table (which this file no longer inherits): it is what makes
; these tests check *exports*, by rejecting names GT4Py only imports but does not
; explicitly re-export.
implicit_reexport = False
```

src/gt4py/next/ffront/dialect_parser.py
```diff
@@ -57,7 +55,19 @@
...
+    # TODO(egparedes): make these two entries reachable for the common 'except <Type>:'
+    # shape. Naming an exception type turns it into a closure variable, and
+    # 'func_to_foast.FieldOperatorParser.visit_FunctionDef' types closure variables
+    # *before* visiting the body, so 'try: ... except ValueError: ...' fails first with
+    # "Unexpected object 'ValueError' of type '<class 'type'>' encountered." Only
+    # 'try/finally' and bare 'try/except:' reach this catalogue; 'try*' never does,
+    # since 'except*' always names a type. Fixing it means running the
+    # unsupported-syntax scan ahead of closure-variable type deduction.
     ast.Try: ("'try' statement", ("Exception handling is not available inside GT4Py functions.",)),
+    ast.TryStar: (
```
egparedes, line 72: The new `ast.TryStar` entry is unreachable through the frontend, and the PR's own TODO right above explains why: `except*` always names an exception type, and closure-variable typing runs before the AST walk. The accompanying test has to reach around the frontend and call the private `dialect_parser._describe_unsupported_feature` directly to exercise it.

That makes this a catalogue entry with no end-to-end coverage, layered on top of a known ordering defect rather than fixing it — moving the unsupported-syntax scan ahead of closure-variable type deduction in `func_to_foast.visit_FunctionDef` would make this entry (and the far more common `try: ... except ValueError:` shape, which today reports "Unexpected object 'ValueError' …") reachable, and would let the test go through `parse_error` like every other case in that file. Worth an issue link next to the TODO so it does not become permanent.

egparedes (commented):
Drop the whole `eve`s ADR folder, it's not needed for this trivial change.

egparedes:
Dropped the `docs/development/ADRs/eve/` folder in 4350c055f (revert of 7e7d399e) — the ADR and its index README are both gone, so `ADRs/` is back to `cartesian` + `next`.

For the record, the rationale it carried isn't lost: the non-obvious part (the six names come from the `typing`/`typing_extensions` star imports, not the re-export block, and `__getattr__` forwards anything missing — so deleting the block alone would have silently rebound them to the deprecated `typing` objects) is in the module comment next to `_DEPRECATED_TYPING_ALIASES` and in the commit message of a86bfb92.

src/gt4py/eve/datamodels/core.py
egparedes: Fixed in 203d6102d. Bare `tuple`/`frozenset` (and an unparametrized `typing.Tuple`, treated as its origin) are now rejected, since used bare they say nothing about the items whose hashes they fold in — the same reason `tuple[Any, ...]` is rejected. Covered by `test_strict_frozen_rejects_unparametrized_container_fields`.

src/gt4py/eve/datamodels/core.py
egparedes: Fixed in 203d6102d. The `get_represented_types` fallback is replaced by explicit `TypeVar` (bound / constraints / default) and `ForwardRef` handling that recurses on the *annotation* rather than its flattened origins, so a `tuple[list[int], ...]` bound stays decomposed. Covered by `test_strict_frozen_rejects_type_var_with_mutable_bound`.

src/gt4py/eve/datamodels/core.py
egparedes: Fixed in 203d6102d. Forward references are now resolved in a `try`/`except Exception` that rejects rather than propagating, so a reference resolving to something non-subscriptable, a malformed annotation, or one resolving to another `ForwardRef` yields the documented `EveTypeError` instead of a raw exception.

src/gt4py/eve/extended_typing.py
egparedes: Confirmed and fixed in 6ca810006 — verified against `9ce5be59b`, `eval_forward_ref("List[int]")` did return `list[int]` before and raised `NameError` after. When `globalns` is `None` the module namespace used as the default now re-adds the six names bound to their builtins, so both spellings resolve to the builtin generic again. An explicit `globalns` is still left untouched (also pinned in the test).

src/gt4py/eve/extended_typing.py
egparedes: Good catch, fixed in 6ca810006 using your suggestion. `_ForwardRefTypingNamespace` now returns the builtin, so `typing.Dict[str, int]` resolves to `dict[str, int]` (`types.GenericAlias`) exactly as before this PR, instead of the deprecated `typing._GenericAlias` the module rejects elsewhere. The test asserts the resulting object type, not just the origin.

src/gt4py/eve/extended_typing.py
egparedes: Fixed in 6ca810006. `__dir__` now unions `typing`, `typing_extensions` and `globals()` (minus the rejected aliases) using set membership, so the module own API is listed again and the O(n^2) scan is gone.

src/gt4py/eve/extended_typing.py
egparedes: Fixed in 6ca810006 — the `isinstance(_typing.Any, type)` guard around `_ArtefactTypes` is unconditionally true on the 3.12 floor and is now inlined.

typing_tests/mypy.ini
egparedes: Fixed in cdefb90b4. You are right that this is the setting that makes the session check *exports* at all; it is carried back over with a comment explaining why that one and not the rest of the project table. Re-ran `test_typing_exports` with it enabled: 16 passed.

src/gt4py/eve/extended_typing.py
egparedes: An ADR was added in 7e7d399e, then dropped again in 4350c055f at the maintainer request ("not needed for this trivial change"). The rationale that is not obvious from the diff — that the names come from the `typing`/`typing_extensions` star imports rather than the re-export block, so deleting the block alone would silently rebind them to the deprecated `typing` objects — is recorded in the comment next to `_DEPRECATED_TYPING_ALIASES` and in the a86bfb92 commit message.

src/gt4py/next/ffront/dialect_parser.py
egparedes: Confirmed, and deliberately left as is in this PR. Reproduced: `try: ... except ValueError: ...` reports "Unexpected object 'ValueError' of type '<class type>'" because `visit_FunctionDef` types closure variables before walking the body, so only `try/finally` and bare `try/except:` reach the catalogue and `try*` never does. The fix is to run the unsupported-syntax scan ahead of closure-variable type deduction, which changes user-visible error messages and belongs in its own PR. db6e0ac6b records this as TODOs at the root cause (`func_to_foast.visit_FunctionDef`), at the catalogue entries, and on the test, so it is not mistaken for general `try` coverage. Leaving this thread open to track it.

egparedes:
@romanc heads-up: `main` is currently broken by #2772, and the fix is in this PR.

That PR swapped the ellipsis check for `isinstance(cn, types.EllipsisType)`, but that is the type of the `...` *object*, not of its AST node — an `ast.Constant` whose value is `Ellipsis` — so the condition is never true on any version:

```python
>>> node = ast.parse('a[...]').body[0].value.slice
>>> isinstance(node, types.EllipsisType)                       # #2772
False
>>> isinstance(node, ast.Constant) and node.value is Ellipsis  # this PR
True
```

Effect on current `main` — a stencil using an ellipsis index no longer parses:

```
field_b[...] = field_a
→ GTScriptSyntaxError: Unexpected type found <class 'ellipsis'>.
  Expected one of: int, AxisIndex, string (var ref), or None.
```

This branch already had `_is_ellipsis_node` for exactly this (it was the 3.14 half of the same bug: the old `getattr(ast, 'Ellipsis', types.EllipsisType)` fell back to the same non-matching type once `ast.Ellipsis` is removed). The merge in bac34fe94 keeps that helper and takes your `visit_Expr` — yours is better than what this branch had, since it drops only *string* constant statements rather than every constant. Nothing else from #2772 is reverted, and `TestEllipsisNodeDetection` pins both the helper and an end-to-end parse.

If you would rather not wait on this PR's review, the one-line fix is happy to go to `main` on its own and I will rebase.

src/gt4py/cartesian/frontend/gtscript_frontend.py
```diff
@@ -52,6 +52,16 @@
 PYTHON_AST_VERSION: Final = (3, 12)
 
 
+def _is_ellipsis_node(node: ast.AST) -> bool:
+    """Check whether an AST node is the '...' literal.
+
+    'ast.Ellipsis' is a deprecated alias scheduled for removal in Python 3.14, and
+    'types.EllipsisType' is the type of the '...' object itself, not of its AST node,
+    so neither is usable as an 'isinstance()' target here.
```
romanc: While this part of the docstring is helpful in the context of the PR, I think it won't be helpful in the future. I suggest to drop it, leaving the function as

```py
def _is_ellipsis_node(node: ast.AST) -> bool:
    """Check whether an AST node is the '...' literal."""
    return isinstance(node, ast.Constant) and node.value is Ellipsis
```

The PR contains all the context and explains in detail what was wrong. (Future) User of this functions don't need to know what the history was.

romanc (approved):
🧹  Thanks for the cleanup!

> @romanc heads-up: `main` is currently broken by #2772, and the fix is in this PR.

oh oh ... thanks for catching and fixing this. 

> If you would rather not wait on this PR's review, the one-line fix is happy to go to `main` on its own and I will rebase.

I think it's fine to wait. I didn't even know that syntax existed and since there was no test coverage, I can only assume it's something old/odd. Seems redundant to me ... I mean why would you write `field_b[...] = field_a`  when you can just write `field_b = field_a` instead?

I'll bring it to the team. Maybe we can just get rid of the syntax (in subsequent PRs) if nobody cares about it.

docs/development/next/error-messages.md
```diff
@@ -119,13 +119,15 @@ except errors.DSLError as err:
...
-context.
+This is the stock `BaseException.add_note`, so the breadcrumb lands in
+`__notes__`, not in the structured `notes` field — `notes` is reserved for
+content authored at the raise site. The two have different renderers:
+`DSLError.__str__` emits only the structured parts, while `__notes__` is
+printed by the traceback machinery (and therefore by pytest and
+IPython/Jupyter). The excepthook in `errors/excepthook.py` replaces that
+machinery, so it appends `__notes__` itself. The seam is wired at
+`func_to_foast` (`ffront/func_to_foast.py`); add it at later stages as they
+gain useful context.
```
havogt: If I understand correctly this note is obsolete. To my understanding only documenting pre-3.12 history. But please let it double-check this.
```suggestion
```

docs/development/next/error-messages.md
```diff
@@ -201,14 +203,11 @@ Unsupported operand type(s) for +: 'Field[[IDim], float64]' and 'Field[[IDim], b
...
+678\) and every `ast` node up to 3.12 can be used directly.
 
-- Import `Self` from `gt4py.eve.extended_typing`, not `typing` (3.11+ only).
-- `DSLError.add_note` works on every Python because `DSLError` defines it;
-  don't rely on `add_note` for *other* `GT4PyError`s — it is a builtin only on
-  3.11+.
-- The catalogue must not reference `ast` nodes added after 3.10 (e.g.
-  `ast.TryStar`) unconditionally — that breaks import on 3.10.
+The diagnostics code carries no version shims left over from the old 3.10
+floor; do not add new ones.
```
havogt:
```suggestion
```

docs/development/next/error-messages.md
```diff
@@ -201,14 +203,11 @@ Unsupported operand type(s) for +: 'Field[[IDim], float64]' and 'Field[[IDim], b
...
 
-These spots are flagged with `TODO(havogt)`.
+Nodes introduced *after* 3.12 (for example `ast.TemplateStr` for PEP 750
+t-strings, 3.14) still cannot be referenced unconditionally in the catalogue.
```
havogt: Can this be easily fixed in this PR or a small follow-up? I don't exactly understand what it means either.

src/gt4py/cartesian/frontend/gtscript_frontend.py
egparedes: Agreed.

src/gt4py/eve/datamodels/core.py
```diff
@@ -1056,9 +1052,113 @@ def _type_converter(value: Any) -> _T:
...
+_ITEM_HASHING_CONTAINER_TYPES: Final = (tuple, frozenset)
+
+
+def _is_strictly_immutable_type(
+    type_annotation: TypeAnnotation, *, _as_container_origin: bool = False
+) -> bool:
+    """Check whether an annotation only admits strictly immutable (hashable) values.
+
+    A datamodel qualifies only if it is itself defined with ``frozen="strict"``;
+    any other plain type qualifies if it defines a custom ``__hash__``. Composite
```
havogt: > any other plain type qualifies if it defines a custom ``__hash__`

Don't understand that part.

src/gt4py/eve/datamodels/core.py
```diff
@@ -1056,9 +1052,113 @@ def _type_converter(value: Any) -> _T:
 
 _KNOWN_MUTABLE_TYPES: Final = (list, dict, set)
 
+#: Hashable container types folding the hashes of the items they hold into their own.
+#: Used unparametrized they say nothing about those items, so they are only accepted
+#: as the origin of a parametrized alias (``tuple[int, ...]``), never on their own.
+_ITEM_HASHING_CONTAINER_TYPES: Final = (tuple, frozenset)
+
+
+def _is_strictly_immutable_type(
```
havogt, line 1061: Was this a bugfix for a general problem or exposed by Python 3.12+ features

havogt: In case you didn't already, it might be worth to ask an agent for a targeted review on this function.

src/gt4py/eve/extended_typing.py
```diff
@@ -385,11 +439,8 @@ def __pretty__(
 
 
 # -- Added functionality --
-_ArtefactTypes: tuple[type, ...] = (_types.GenericAlias,)
-
-# `Any` is a class since Python 3.11
-if isinstance(_typing.Any, type):  # Python >= 3.11
-    _ArtefactTypes = (*_ArtefactTypes, _typing.Any)
+# `Any` is a class since Python 3.11, which is below the supported floor.
```
havogt:
```suggestion
```

src/gt4py/eve/extended_typing.py
```diff
@@ -398,7 +449,7 @@ def __pretty__(
     _ArtefactTypes = (*_ArtefactTypes, typing_exts_any)
```
havogt: Is this still needed.

src/gt4py/next/otf/compiled_program.py
```diff
@@ -190,7 +190,12 @@ def wait_for_compilation() -> None:
     if len(failures) == 1:
         raise failures[0][1]
     if failures:
-        # TODO(havogt): raise ExceptionGroup once Python 3.10 is dropped.
+        # TODO(havogt): raise an ExceptionGroup here. The 3.10 floor that originally
+        # blocked this is gone (PEP 654 is available on the 3.12 floor), so only the
+        # flattening below still loses information: failures 2..n survive as 'repr'
+        # text and '__cause__'/'__traceback__' carry the first failure alone. Left as
+        # is because it changes the documented 'Raises:' contract of this public
+        # function from 'RuntimeError' to 'ExceptionGroup', which callers may catch.
```
havogt: I think it's ok to fix here.

havogt (changes requested):

docs/development/next/error-messages.md
egparedes: Double-checked as asked, and this one is not history — it describes current, version-independent behaviour, so I kept it but reworded the part that made it read as history (532fdb2b3).

Verified against the code on this branch:
- `DSLError.__str__` (`errors/exceptions.py:94`) never touches `__notes__`, so it emits only the structured parts.
- `errors/excepthook.py:53-55` explicitly re-renders them: `exc_strs += [f"\n{note}" for note in getattr(value, "__notes__", [])]`, with the comment "Our hook replaces the default, so we render PEP 678 `__notes__`".

So `add_note` really does land somewhere `str(err)` will not show you, and the excepthook has to compensate — worth keeping documented. What made it look like history was the opening "This is the stock `BaseException.add_note`", which reads as a contrast with the 3.10 override this PR deletes. That framing is gone now.

docs/development/next/error-messages.md
egparedes: Dropped in 532fdb2b3. Agreed — "the floor is 3.12" and "carries no shims" are meta-commentary that goes stale rather than guidance, and the second sentence was already load-bearing in a way it should not have been (this PR had to fix two more dead guards in `extended_typing` that contradicted it).

docs/development/next/error-messages.md
egparedes: Fixed here, it was small (3bd102d72).

What it meant: the catalogue is a dict literal keyed by `ast` classes, evaluated at import. Naming `ast.TemplateStr` (PEP 750 t-strings, 3.14) in it would raise `AttributeError` on 3.12/3.13 and break the import there — so post-floor constructs could not be catalogued at all and fell back to the generic `ast`-class-name message.

There is now a companion `_NEWER_UNSUPPORTED_FEATURE_HINTS` keyed by node *name*, merged in via `getattr(ast, name, None)`; entries the interpreter does not have are skipped, which is harmless since the construct is unparseable there anyway. Added `TemplateStr` as the first entry, so on 3.14 a t-string now reports `Unsupported Python syntax: t-string` instead of `ast.TemplateStr`. Test is skipif-guarded on `hasattr(ast, "TemplateStr")`.

src/gt4py/eve/datamodels/core.py
egparedes: Reworded in 069b9422e — it stated the rule without saying why it holds.

The reasoning: `__hash__` is the only signal the language gives here. A type that defines its own is claiming its values hash by content, which is only sound if they do not change; the mutable builtins opt out explicitly with `__hash__ = None`; and the inherited `object.__hash__` hashes by identity, which says nothing at all about the value. So the check (`is_type_with_custom_hash`) excludes exactly those two — `type_.__hash__ not in (None, object.__hash__)` — and accepts everything else. The docstring now says that.

src/gt4py/eve/extended_typing.py
egparedes: Dropped in 069b9422e, along with the guard below it — see the next thread.

src/gt4py/eve/extended_typing.py
egparedes: No — dead, removed in 069b9422e. Checked on all three supported versions:

```
3.12 -> typing_extensions.Any is typing.Any: True
3.13 -> typing_extensions.Any is typing.Any: True
3.14 -> typing_extensions.Any is typing.Any: True
```

So `_ArtefactTypes` never grew its third entry. There was a second instance of the same guard further down (`is_Any`, whose own comment said "When using Python < 3.11 and typing_extensions >= 4.4"), which always took the `else` branch — that one is gone too. Both collapse to the branch that actually ran.

src/gt4py/next/otf/compiled_program.py
egparedes: Done in 3bd102d72.

One wrinkle worth flagging: `Future.exception()` is typed as `BaseException`, and `ExceptionGroup` only accepts `Exception` (mypy catches this). So the raise uses `BaseExceptionGroup`, which returns a plain `ExceptionGroup` whenever every member is an `Exception` — i.e. the normal case — so `except*` on `Exception` still works. The test asserts `type(...) is ExceptionGroup` to pin that.

The `Raises:` docstring is updated, and the multi-failure path now has a test; it had none before. Note the test has to hold strong references to the futures, since tracking is weak and a collected future silently degrades it to the single-failure path.

src/gt4py/eve/datamodels/core.py
egparedes: **On the origin:** a general bug, nothing to do with 3.12. The function called `xtyping.is_hashable_type`, which has never existed, so *every* `frozen="strict"` datamodel raised `AttributeError` at class-definition time on every Python version. `git log -S` puts the broken call in b3677258b (#1146, 2023-02-01), so it had been dead for ~3.5 years.

It survived that long because three things lined up: `frozen="strict"` is used nowhere in `src/` (so nothing ever executed the path), `tests/eve_tests` had zero coverage of it before this PR, and the star-import plus module `__getattr__` in `extended_typing` means a missing name is invisible to both ruff and mypy — `xtyping.<anything>` type-checks. That last one is also why the same module needed the explicit rejection list for the removed `typing` aliases: absence there is silent by construction.

So the 3.12 audit is what surfaced it, but nothing about 3.12 caused it.

**On the agent review:** yes, twice, and it was worth it — most of the hardening in this PR came out of it rather than from me. Between the two passes it found:

- an unparametrized container (`values: tuple`) passing the check, so `hash()` still raised `TypeError` — the exact invariant the flag exists to guarantee;
- the `get_represented_types` fallback flattening annotations to their origins, so a `tuple[list[int], ...]` `TypeVar` bound came back as a plain hashable `tuple`;
- `except NameError` being too narrow around forward-reference resolution;
- a `Tuple[List[int], ...]` field passing because only the origin was inspected;
- forward references escaping as raw `NameError` instead of the documented `EveTypeError`.

All fixed, each with a test. Fair warning that the function has changed again since that last pass — ac1dd6c69 added PEP 695 alias resolution on top, after #2754 landed, because `_is_strictly_immutable_type` turned out to be a fourth annotation-dispatch funnel that did not know about aliases (a `type Pair = tuple[int, int]` field was rejected while the identical plain annotation was accepted). Happy to run another targeted pass over the current version if you want one before merging.

havogt (approved):
lgtm

src/gt4py/cartesian/frontend/gtscript_frontend.py
romanc: haha I found https://github.com/GridTools/gt4py/issues/2406 after searching a bit. Let's merge this PR and then I'll go and do the work to drop that syntax once and for all, giving a nice user error.

PR #2767, opened 2026-08-10
fix[next]: decide concat_where pruning from the concat_where's domain

havogt:
`prune_empty_concat_where` decided whether a branch is ever selected from **the branch's own domain**, filtered to the dimensions of the condition. That is the wrong quantity, and it is wrong in both directions.

## Too weak

A branch's domain is restricted to the dimensions of the branch's own type, so for a branch that does not have the concat dimension the filtered domain has no ranges at all — and an empty mapping is not an empty domain. A branch selected nowhere survives.

```
concat_where(u⟨ Kᵥ: [-∞, 0[ ⟩, a, b)        a: Field[[Vertex]]
                                            b: Field[[Vertex, K]]
```

over `Vertex: [0, 10)`, `K: [0, 10)`. `a` is selected where `K < 0`, i.e. nowhere, so the expression is `b` — but `a`'s domain is `Vertex: [0, 10)` with no `K` range to be empty, so it survived.

## Too strong

Nothing checked that the surviving branch spans the same dimensions as the `concat_where`, so with `a: Field[[Vertex, K]]` and `b: Field[[K]]` the same expression is rewritten to `b`, replacing a two dimensional expression by a one dimensional one. This one reproduces on `main` today, independent of the case above.

## The fix

Both follow from deriving the selected region from the `concat_where`'s **own** domain — which is exactly what `infer_domain._infer_concat_where` already computes per branch. That computation moves to `domain_utils.concat_where_branch_domain` and both callers use it, so the requirement to take the complement *before* promoting — `domain_complement` is undefined on a range infinite on both sides — exists in one place rather than being duplicated.

A branch is pruned only when its domain spans every dimension of the `concat_where`. Comparing domains rather than types also avoids depending on dimension order, and on a type the pipeline does not set: `_infer_concat_where` rebuilds the node through `im.call`, which produces an untyped `FunCall`, so `node.type` is `None` at every `concat_where` this pass sees.

The pass body goes from 90 lines to 46; the three overlapping mechanisms it had — an `empty()` early return, the branch domain filter, and a type comparison — collapse to one.

## Testing

Three regression tests, each verified to fail against the previous implementation: the never selected branch that lacks the concat dimension, the dimension dropping prune, and a `NEVER` domain that previously crashed. Plus the complement before promote ordering, tested in `test_domain_utils.py` where the helper now lives.

504 `iterator_tests` pass, 39 doctests, and 304 CPU integration tests across `test_concat_where.py` / `test_where.py` / `test_named_collections.py`. `ruff` and `mypy` clean.

## Not behaviour preserving, deliberately

Two intentional differences: pruning is declined where the old code would drop a dimension, and input with no `annex.domain` together with an empty condition now raises rather than pruning — the pass documents that it requires inferred domains, and the doctest was the only such caller. The doctest now runs `infer_expr` first.

### DIFF
--- a/tests/next_tests/integration_tests/feature_tests/ffront_tests/test_concat_where.py
+++ b/tests/next_tests/integration_tests/feature_tests/ffront_tests/test_concat_where.py
@@ -407,3 +407,28 @@ def test_with_tuples_different_domain(cartesian_case):
         out=out,
         ref=(ref0, ref1),
     )
+
+
+def test_concat_where_never_selected_branch_keeps_dimensions(cartesian_case):
+    # The bounds have to be literals: a branch can only be shown to be never selected
+    #  when the domain is known at compile time.
+    isize, ksize = 3, 4
+
+    @gtx.field_operator
+    def testee(interior: cases.IKField, profile: cases.KField) -> cases.IKField:
+        return concat_where(KDim < 0, interior, profile)
+
+    @gtx.program
+    def prog(interior: cases.IKField, profile: cases.KField, out: cases.IKField):
+        testee(interior, profile, out=out, domain={IDim: (0, 3), KDim: (0, 4)})
+
+    interior = cases.allocate(
+        cartesian_case, testee, "interior", domain=gtx.domain({IDim: isize, KDim: ksize})
+    )()
+    profile = cases.allocate(cartesian_case, testee, "profile", domain=gtx.domain({KDim: ksize}))()
+    out = cases.allocate(
+        cartesian_case, testee, cases.RETURN, domain=gtx.domain({IDim: isize, KDim: ksize})
+    )()
+
+    ref = np.broadcast_to(profile.asnumpy(), (isize, ksize))
+    cases.verify(cartesian_case, prog, interior, profile, out=out, ref=ref)
--- a/src/gt4py/next/iterator/ir_utils/domain_utils.py
+++ b/src/gt4py/next/iterator/ir_utils/domain_utils.py
@@ -353,6 +353,20 @@ def promote_domain(
     return SymbolicDomain(domain.grid_type, dims_dict)
 
 
+def concat_where_branch_domain(
+    domain: SymbolicDomain, cond: SymbolicDomain, is_true_branch: bool
+) -> SymbolicDomain:
+    """
+    Return the part of `domain` on which one branch of a `concat_where` is selected.
+
+    Note: the complement is taken before promoting to the dimensions of `domain`, since
+    `domain_complement` requires each range to be infinite on exactly one side, which a
+    promoted dimension is not.
+    """
+    region = cond if is_true_branch else domain_complement(cond)
+    return domain_intersection(domain, promote_domain(region, domain.ranges.keys()))
+
+
 def is_finite(range_or_domain: SymbolicRange | SymbolicDomain) -> bool:
     """
     Return whether a range is unbounded in (at least) one direction.
--- a/src/gt4py/next/iterator/transforms/infer_domain.py
+++ b/src/gt4py/next/iterator/transforms/infer_domain.py
@@ -364,19 +364,14 @@ def _infer_concat_where(
     actual_domains: AccessedDomains = {}
     cond, true_field, false_field = expr.args
     symbolic_cond = domain_utils.SymbolicDomain.from_expr(cond)
-    cond_complement = domain_utils.domain_complement(symbolic_cond)
 
-    for arg in [true_field, false_field]:
+    for is_true_branch, arg in [(True, true_field), (False, false_field)]:
 
         @tree_map
         def mapper(d: NonTupleDomainAccess):
             if isinstance(d, DomainAccessDescriptor):
                 return d
-            promoted_cond = domain_utils.promote_domain(
-                symbolic_cond if arg == true_field else cond_complement,  # noqa: B023 # function is never used outside the loop
-                d.ranges.keys(),
-            )
-            return domain_utils.domain_intersection(d, promoted_cond)
+            return domain_utils.concat_where_branch_domain(d, symbolic_cond, is_true_branch)  # noqa: B023 # function is never used outside the loop
 
         domain_ = mapper(domain)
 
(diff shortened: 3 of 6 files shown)


tehrengruber:
I haven't read the code and essentially stopped at
> Both follow from deriving the selected region from the concat_where's own domain

I think it is not needed to couple the `prune_empty_concat_where` pass to the domain inference which propagates in reverse direction (from an expression to its arguments). Instead it is totally sufficient to prune based on the arguments / branches as the semantics of `concat_where` is to implicitly boardcast the dimensions that are not given in a branch, but only in another. In other words in the example
```
concat_where(u⟨ Kᵥ: [-∞, 0[ ⟩, a, b)        a: Field[[Vertex]]
                                                               b: Field[[Vertex, K]]
```
is the same as `concat_where(u⟨ Kᵥ: [-∞, 0[ ⟩, broadcast(a, (Vertex, K)), b)` and then the problem described in the PR does not occur. Without further research it would be ideal to have a central function that promotes the domain of the concat where arguments (one pass already has this, but I don't recall from the top of my head) and then use this function in `prune_empty_concat_where`.

I'll add further comments after reading the code.

havogt:
> *Written by Claude, not reviewed by @havogt — treat the reasoning as needing a second pair of eyes, though the measurements below are reproducible from the scripts mentioned.*

Thanks — I implemented your suggestion to see where it leads, and it uncovered something worth reporting before the design is settled: **taken literally, pruning from the branch domains cannot fix the first of the two bugs, because the information is not there.**

## The counterexample

Two expressions, same condition, same branch types, and **byte-identical branch domains**, that require opposite answers:

| | accessed domain | `cond` | `a.annex.domain` | `b.annex.domain` | correct |
|---|---|---|---|---|---|
| A | `V:[0,10[, K:[0,10[` | `K < 0` | `V:[0,10[` | `V:[0,10[, K:[0,10[` | prune to `b` |
| B | `V:[0,10[, **K:[-5,10[**` | `K < 0` | `V:[0,10[` | `V:[0,10[, K:[0,10[` | **must not prune** |

with `a: Field[[Vertex]]`, `b: Field[[Vertex, K]]`. The two differ *only* in `node.annex.domain`. A pass that does not read it must return the same result for both, so it must not prune either.

The reason is `infer_expr`: it filters each branch's domain to the dimensions of the branch's own type (`infer_domain.py`, `_filter_domain_dimensions`). For the branch that lacks the concat dimension, the one range that would be empty is exactly the one that gets dropped.

## What that leaves

Three ways to complete the design, as far as I can see:

1. feed the promotion the `concat_where`'s own domain — which is what #2767 already does;
2. materialise the broadcast in the IR, so the branch node genuinely has the dimensions;
3. leave that bug unfixed.

Option 2 is what your own comment writes out (`concat_where(cond, broadcast(a, (Vertex, K)), b)`), so that is what I implemented, on a branch off `main`: **havogt/gt4py#76**, staged on the fork for discussion rather than proposed as a replacement. It is two commits: one that decides pruning from the branch domains, and one that adds a `concat_where.broadcast_branches` pass making the implicit broadcast explicit before `canonicalize_domain_argument`. After it, every branch spans the full dimensions and the branch-domain test is sufficient.

## One thing that argues against the framing

The central promotion helper you half-remembered does exist: `gtir_to_sdfg_concat_where.py`, in `_translate_concat_where_branch()`. It fills in the branch's missing concat range, promotes to the output dimensions, intersects, and builds an `as_fieldop("deref", …)` — a literal broadcast. It is called with `output_domain=node.annex.domain`.

So the one pass that already promotes `concat_where` argument domains needs the `concat_where`'s own domain to do it. That is not conclusive about what `prune_empty_concat_where` should do, but it is evidence that the domain is the natural input for this operation rather than an accident of #2767's approach.

## Severity, since it affects how much this is worth

The two bugs are not equally serious.

- The **dimension dropping** one is a hard compile failure, not a pessimisation: `inference.visit_SetAt`'s `assert expr_type.dims == target_type.dims` fires, reproduced from plain user code on gtfn. Under `python -O` the assert disappears and the mismatch reaches the backend. It needs statically known bounds, which is why the existing ffront tests never hit it — `cases` produces `get_domain_range(out, …)` bounds so nothing prunes — and it is exactly icon4py's compilation mode.
- The **never selected branch** one is only a missed optimisation; the numbers come out right.

## Cost, which is the part I am least sure about

`broadcast_branches` runs type inference and rewrites IR for every `concat_where` with a lower-dimensional branch, which is a common icon4py shape. Measured: gtfn output byte-identical for a non-prunable case (`fuse_as_fieldop` absorbs the `deref`), the DaCe path produces exactly the `as_fieldop(deref, D ∩ cond)` its own lowering was building internally, and 2030 integration tests unchanged across roundtrip, numpy and gtfn. But "no difference in the cases I measured" is not "no difference at scale", and this touches code that is currently being performance tuned.

The two commits are split so the second can be dropped, in which case the never-selected-branch bug stays open and cannot be closed without either `node.annex.domain` or the explicit broadcast.

The branch is at havogt/gt4py#76 if you want to look at it; happy to fold whichever shape you prefer into #2767.

tehrengruber:
Do you have an end-to-end example that fails without this fix? I'm currious what happens in that case.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_concat_where.py
```diff
@@ -407,3 +407,28 @@ def testee(
         out=out,
         ref=(ref0, ref1),
     )
+
+
+def test_concat_where_never_selected_branch_keeps_dimensions(cartesian_case):
```
havogt, line 412: wondering if we should make the cases to parametrize over static and runtime domains (or at least make it selectable). I only realize now that probably we only have very limited tests with static domains, but that's our production use-case.

havogt:
Asked it to add a frontend test for the failing case. Note that I initially worked on a fix for the second case, when the prune is not done but could be done. Do you want an example for that as well?

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_concat_where.py
tehrengruber: Done in https://github.com/GridTools/gt4py/pull/2795/changes

tehrengruber:
Superseeded by #2795. Should cover all cases and more.

PR #2778, opened 2026-08-11
fix[next]: more dace transformation determinism

tehrengruber:
Remaining changes from #2568 not in #2635.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
@@ -842,7 +842,11 @@ class MoveDataflowIntoIfBody(dace_transformation.SingleStateTransformation):
         sdfg: dace.SDFG,
         if_block: dace_nodes.NestedSDFG,
     ) -> Optional[
-        tuple[list[str], list[str], dict[str, tuple[dace.SDFGState, dace_nodes.AccessNode]]]
+        tuple[
+            OrderedSet[str],
+            OrderedSet[str],
+            dict[str, tuple[dace.SDFGState, dace_nodes.AccessNode]],
+        ]
     ]:
         """Check if `if_block` can be processed and partition the input connectors.
 
@@ -946,6 +950,6 @@ class MoveDataflowIntoIfBody(dace_transformation.SingleStateTransformation):
         if len(non_relocatable_connectors) == 0:
             return None
 
-        relocatable_connectors = list(connector_usage_location.keys())
+        relocatable_connectors = OrderedSet(connector_usage_location.keys())
 
-        return relocatable_connectors, list(non_relocatable_connectors), connector_usage_location
+        return relocatable_connectors, non_relocatable_connectors, connector_usage_location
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -895,7 +894,7 @@ def _partition_if_block(
         connector_usage_location: dict[str, tuple[dace.SDFGState, dace_nodes.AccessNode]] = {}
 
         # This is the dataflow that can not be relocated.
-        non_relocatable_connectors: set[str] = set()
+        non_relocatable_connectors: OrderedSet[str] = OrderedSet()
```
edopao, line 895: The function is returning a list. I was thinking whether we could keep `non_relocatable_connectors` a regular set, but in the return statement sort the connectors alphabetically: `sorted(list(non_relocatable_connectors))`

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -874,8 +873,8 @@ def _partition_if_block(
         if len(if_block.out_connectors.keys()) == 0:
             return None
 
-        input_names: set[str] = set(if_block.in_connectors.keys())
-        output_names: set[str] = set(if_block.out_connectors.keys())
+        input_names: OrderedSet[str] = OrderedSet(if_block.in_connectors.keys())
+        output_names: OrderedSet[str] = OrderedSet(if_block.out_connectors.keys())
```
edopao, line 875: If applying the solution suggested below, these `_names` variables can stay as regular sets.

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
tehrengruber: I would propose the exact opposite: remove the sorting and return an ordered set:
- Sorting statements usually hinder readability and they amplify performance indeterminism from unstable symbols, whereas relying on insertion order makes unstable symbols mostly harmless as they only matter for comparability.
- The sorting approach only works for sortable entities, whereas insertion order always works so just relying on insertion order gives a simple recipe.

edopao: You are right, I like in particular the argument "Sorting statements amplify performance indeterminism from unstable symbols".

OK, agreed. Let's return an `OrderetSet` instead of a list.

tehrengruber:
cscs-ci run dace-determinism

tehrengruber:
cscs-ci run dace-determinism

src/gt4py/next/program_processors/runners/dace/transformations/move_dataflow_into_if_body.py
```diff
@@ -953,8 +946,6 @@ def _partition_if_block(
         if len(non_relocatable_connectors) == 0:
             return None
 
-        # We only guarantee that `relocatable_connectors` has an stable order,
-        #  everything else has no guaranteed order, even `connector_usage_location`.
-        relocatable_connectors = sorted(connector_usage_location.keys())
+        relocatable_connectors = list(connector_usage_location.keys())
 
         return relocatable_connectors, list(non_relocatable_connectors), connector_usage_location
```
edopao: Optional: `relocatable_connectors` and `non_relocatable_connectors` could be returned as `OrderedSet`s.

edopao (approved):
LGTM

PR #2779, opened 2026-08-12
fix[next]: deterministic declaration order in gtfn codegen

havogt:
The gtfn lowering collected dimensions, cartesian offsets and shift offsets into sets and iterated them while building the generated source.

The fix replaces the set by an OrderedSet at the three collectors. Only the one in _get_domains is backed by observed diffs; the other two are the same defect on inputs the test runs did not reach, namely several cartesian dimensions and more than one connectivity offset in a stencil.

### DIFF
--- a/src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
+++ b/src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
@@ -77,10 +77,15 @@ def _is_tuple_of_ref_or_literal(expr: itir.Expr) -> bool:
 
 
 def _get_domains(nodes: Iterable[itir.Stmt]) -> Iterable[itir.FunCall]:
-    result = set()
+    # The order the domains are returned in fixes the order of the generated dimension
+    # tag declarations, so it has to be derived from the IR rather than from set
+    # iteration, which varies with `PYTHONHASHSEED` across processes.
+    result: dict[itir.FunCall, None] = {}
     for node in nodes:
-        result |= node.walk_values().if_isinstance(itir.SetAt).getattr("domain").to_set()
-    return result
+        result.update(
+            dict.fromkeys(node.walk_values().if_isinstance(itir.SetAt).getattr("domain").to_list())
+        )
+    return result.keys()
 
 
 def _name_from_named_range(named_range_call: itir.FunCall) -> str:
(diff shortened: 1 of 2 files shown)


src/gt4py/next/program_processors/codegens/gtfn/itir_to_gtfn_ir.py
```diff
@@ -424,19 +431,23 @@ def _make_domain(self, node: itir.FunCall) -> tuple[TaggedValues, TaggedValues]:
...
+                dict.fromkeys(
+                    n.pre_walk_values()
+                    .if_isinstance(node_type)
+                    .getattr("value")
+                    .if_isinstance(str)
+                    .to_list()
+                )
             )
-        return result
+        return list(result)
```
edopao: What about returning just an iterable (`result.keys()`) as in `_get_domains()`?

edopao (approved):
LGTM

tehrengruber (commented):
I would just use `OrderedSet` which we have as a requirement anyway and is a drop-in replacement for set.

havogt:
Switched all three sites to `OrderedSet` — it is a genuine drop-in (`add` / `remove` /
`discard` / `update` all behave), and it states the intent far more directly than
threading order through a `dict`. The diff is now `set` → `OrderedSet` and `.to_set()` →
`.to_list()`, which reads as the minimal change it should have been.

@edopao on returning an iterable: `_collect_offset_or_axis_node` now returns the
`OrderedSet` itself rather than a materialized `list`, which is the `OrderedSet`
analogue of your `result.keys()` suggestion and keeps it symmetric with `_get_domains`.

Verified after the rewrite: the two regression tests still fail against `main` and pass
with the fix, and `pre-commit` (ruff, mypy, tach) is clean.

havogt:
Comments tightened to a single one on `_get_domains`, noting that it covers the other
collectors too, and no longer mentioning `PYTHONHASHSEED`. With `OrderedSet` the intent
is legible from the type, so the other two sites are now bare.

On whether this over-orders: I checked, and all three do reach the generated code.

| site | order flows to |
|---|---|
| `_get_domains` | `offset_definitions` insertion order → dimension tag declarations |
| `_collect_offset_definitions` | the same dict → the same declarations |
| `_collect_offset_or_axis_node` | `connectivities.append(...)` → the emitted connectivities list |

Worth being explicit about the evidence, though: only `_get_domains` is backed by
observed diffs. The other two are the same defect on inputs the icon4py dycore /
diffusion workload does not reach — `_collect_offset_definitions` needs several
cartesian dimensions (unstructured only ever inserts the vertical one there), and
`_collect_offset_or_axis_node` needs a stencil with more than one connectivity offset.

So: correct as far as I can tell, but two thirds of it is reasoned rather than measured.
Happy to split those two out if you would rather they came with their own evidence.

havogt:
Dropped the tests. An ordering assertion only fails when the unordered iteration happens
to differ from the expected order, so it is flaky by construction and buys little
confidence — the same reason #2780 ships without one.

The PR is now source-only: `set` → `OrderedSet` at the three collectors, one comment.

tehrengruber:
The description is not correct:
> The build cache is keyed on a fingerprint of the generated source
(strict_fingerprinter(ExtensionSource) in otf/compilation/cache.py), so affected
programs miss the cache on every fresh process.

The translation cache side-steps this.

I'm noticing wrong descriptions quite a bit recently and I fear that this deteriorate human and agentic performance. At least I frequently resort to commit messages (let aside their importance for review).

havogt:
I usually cleanup the description before merge, but I am not sure what's better for review. The details (with maybe partially wrong reasoning, which might highlight also wrong implementation) or strip to the very concise high level description that I would usually write. Happy to discuss. (And with the last sentence you don't know who wrote this. ;) )

PR #2780, opened 2026-08-13
fix[next-dace]: deterministic access node order in concat_where mapper

havogt:
The helper _setup_initial_producer_description_in_nested_state collected the data names it still had to create access nodes for into a set and then iterated it. That iteration decides the order the access nodes are inserted into the state so the same program could produce SDFGs differing in node order between processes.

The fix uses an OrderedSet, so the access nodes follow the order of the producer specs.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/concat_where_mapper.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/concat_where_mapper.py
@@ -22,6 +22,7 @@ from dace import (
 )
 from dace.sdfg import graph as dace_graph, nodes as dace_nodes
 from dace.transformation.passes import analysis as dace_analysis
+from ordered_set import OrderedSet
 
 from gt4py.next import config as gtx_config
 from gt4py.next.program_processors.runners.dace import transformations as gtx_transformations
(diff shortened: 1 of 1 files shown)


tehrengruber (approved):
I would propose the `OrderedSet` again. Sorting of indeterministic strings lead to performance indeterminism, whereas ordered sets don't care.

havogt:
Switched to `OrderedSet`, keeping the insertion order of the producer specs.

Your reasoning is better than mine and I have dropped the argument I made in the
description. I had reached for `sorted` on the grounds that a canonical order is robust
even if the upstream order were itself unstable. But these names embed generated
numbering, so sorting imposes an order unrelated to the dataflow — deterministic output
bought at the cost of a schedule that no longer follows the natural order, which is
exactly the performance non-determinism you point at. `OrderedSet` preserves the natural
order and does not make that trade.

`pre-commit` (ruff, mypy, tach) clean. The end-to-end evidence is unchanged: 45/45
pairwise comparisons identical across 10 hash seeds on `test_graupel_only` (R2B05),
against 4 distinct fingerprints from 2 runs before the fix.

PR #2785, opened 2026-08-13
fix[next-dace]: deterministic GT4PyStateFusion global sink merging

### LINKED ISSUE #2784: GT4PyStateFusion: surviving global sink node is chosen with set.pop(), making the fused state's node order non-deterministic
When `GT4PyStateFusion` merges two states that both write the same global, the combined
state ends up with several sink `AccessNode`s for that data. The ADR-18 fixup that
merges them picks the survivor out of a `set`:

`src/gt4py/next/program_processors/runners/dace/transformations/state_fusion.py`

```python
global_sink_nodes: dict[str, set[dace_nodes.AccessNode]] = {}
...
    global_sink_nodes[sink_node.data] = set()
    global_sink_nodes[sink_node.data].add(sink_node)

for sink_nodes in global_sink_nodes.values():
    if len(sink_nodes) <= 1:
        continue
    final_sink_node = sink_nodes.pop()          # arbitrary
    for sink_node in sink_nodes:                # set order
        for iedge in first_state.in_edges(sink_node):
            first_state.add_edge(iedge.src, iedge.src_conn, final_sink_node,
                                 iedge.dst_conn, iedge.data)
        first_state.remove_node(sink_node)
```

`AccessNode` does not override `__hash__`, so the set is ordered by `id()`. Which node
survives — and hence where it sits in the state's node list — varies between processes,
as does the order the redirected edges are added. Note this is independent of
`PYTHONHASHSEED`.

### Reproduction

Both states write the same global `a`. Six runs, two different node orders (the
surviving `a` moves):

```
node_order: ['MapEntry:comp1_map', 'Tasklet:comp1', 'MapExit:comp1_map', 'AccessNode:b', 'AccessNode:a', 'MapEntry:comp2_map', ...]
node_order: ['MapEntry:comp1_map', 'Tasklet:comp1', 'MapExit:comp1_map', 'AccessNode:b', 'MapEntry:comp2_map', 'Tasklet:comp2', 'MapExit:comp2_map', 'AccessNode:c', ...]
```

```python
import dace
from dace.sdfg import nodes as dace_nodes
from gt4py.next.program_processors.runners.dace import transformations as gtx_transformations

sdfg = dace.SDFG("global_write_in_both_states")
state1 = sdfg.add_state(is_start_block=True)
state2 = sdfg.add_state_after(state1)
for name in ["a", "b", "c"]:
    sdfg.add_array(name, shape=((20,) if name == "a" else (10,)), dtype=dace.float64,
                   transient=False)
state1.add_mapped_tasklet("comp1", map_ranges={"__i": "0:10"},
    inputs={"__in1": dace.Memlet("b[__i]")}, code="__out = __in1 + 10.",
    outputs={"__out": dace.Memlet("a[__i]")}, external_edges=True)
state2.add_mapped_tasklet("comp2", map_ranges={"__i": "0:10"},
    inputs={"__in1": dace.Memlet("c[__i]")}, code="__out = __in1 + 12.",
    outputs={"__out": dace.Memlet("a[__i + 10]")}, external_edges=True)
sdfg.validate()

sdfg.apply_transformations_repeated(gtx_transformations.GT4PyStateFusion(),
                                    validate=True, validate_all=True)

st = sdfg.states()[0]
print([type(n).__name__ + ":" + getattr(n, "label", getattr(n, "data", "?"))
       for n in st.nodes()])
```

(The builder is `_make_global_both_state_write` from `test_state_fusion.py`, inlined.)

### Impact

The fused SDFGs are semantically equivalent — in every run both producers end up feeding
the surviving sink — so this is a stable-serialization problem rather than a correctness
one: the SDFG JSON, its hash, and therefore the compile-cache key differ between
processes.

An `OrderedSet` (or picking the survivor deterministically, e.g. the first in the state's
node order) would fix it, in line with #2779 / #2780 and spcl/dace#2445.

Found while auditing the dace transformations for non-deterministic ordering. Verified on
`main` @ 6ec5244e0.

edopao:
Closes #2784 

With this fix, the selection of the sink node is stable, because it relies on the order of nodes in the state, which in turns follows the order of the IR representation.

As commented in the issue report, this is currently not an issue because no debug information is attached to the access nodes, so the serialization of any of the access nodes (to a given data) in the state produces the same result. However, this is a potential source of nondeterminism in case the lowering changes, and each node gets some unique debug information.

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/state_fusion.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/state_fusion.py
@@ -12,6 +12,7 @@ from typing import Any
 import dace
 from dace import transformation as dace_transformation
 from dace.sdfg import nodes as dace_nodes, utils as dace_sdutils
+from ordered_set import OrderedSet
 
 
 # Conditional import because `gt4py.cartesian` uses an older DaCe version without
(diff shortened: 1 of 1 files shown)


iomaganaris (approved):
LGTM

PR #2795, opened 2026-08-14
fix[next]: Improve concat_where pruning

tehrengruber:
## Problem

A `concat_where` implicitly broadcasts each branch to the dimensions the branch does not have itself: in `concat_where(K < 0, a, b)` with `a: Field[[Vertex]]` and `b: Field[[Vertex, K]]`, accessed on `Vertex: [0, 10), K: [0, 10)`, the true branch is effectively `broadcast(a, (Vertex, K))`. This broadcast exists only implicitly — nothing in the IR represents it, and consequently the inferred domain of a branch is restricted to the dimensions of the branch's own type: the domain of the true_branch argument `a` is just `Vertex: [0, 10)`. `prune_empty_concat_where` decided whether a branch is ever selected from exactly these branch domains (filtered to the dimensions of the condition), which fails in two ways:

- **Never-selected branches survived.** In the example, `a` is selected where `K < 0`, i.e. nowhere within `K: [0, 10)` — but the empty range `K: [0, 0)` proving this is exactly the one missing from `a`'s domain, so `a` was never pruned.
- **Pruning dropped dimensions.** With the condition flipped, `concat_where(0 <= K, a, b)` selects `a` everywhere and was pruned to plain `a` — replacing the two-dimensional expression by a one-dimensional one, as nothing accounted for the surviving branch relying on the implicit broadcast. This is a hard failure with statically known domain bounds: GTIR type inference later hits `assert expr_type.dims == target_type.dims` in `visit_SetAt` (and under `python -O`, the mismatched IR reaches the backend).

## Design

To decide whether a branch is ever selected, the pass makes the implicit broadcasts explicit (`broadcast(a, (Vertex, K))`) and reruns domain inference on the resulting `concat_where`. Since an explicit `broadcast` spans all dimensions of the expression, nothing is dropped from its inferred domain, and the pruning decision reduces to a plain `branch.annex.domain.empty()` check on that IR expression. Domain inference is used instead of computing the branch domains in the pass itself for simplicity and to not duplicate the domain semantics of `concat_where`. A surviving `broadcast` branch stays in the IR and is lowered by the subsequent `RemoveBroadcast` pass via the domain annex populated from the pruned `concat_where` — so both bugs disappear with one mechanism. Branches that are equal (up to the implicit broadcast) are pruned directly; the surviving branch is reinferred on the full domain of the `concat_where`, since its domains stem from the region where only its branch instance was selected.

To make the re-inference cheap, `infer_domain.infer_expr` gains a `revisit_already_inferred` option (default `True`, preserving behavior): with `False`, subexpressions that already carry a domain are neither descended into nor altered, so only the newly created `broadcast` nodes are populated.

Also in this PR:

- new ir_maker `im.broadcast(expr, dims)`
- `infer_expr` typed with a TypeVar so input and output expression types match

## New test cases

Notation: `concat_where(<cond>, <tb>: <tb dims>, <fb>: <fb dims>): <accessed domain> → <result>`

Behavior changes (fail or crash against `main`'s pass):

```
concat_where(K < 0,   a: [Vertex, K], b: [K]):      {Vertex: [0, 10), K: [0, 10)} → broadcast(b, (Vertex, K))
concat_where(K < 0,   a: [Vertex],    b: [Vertex]): {Vertex: [0, 10), K: [0, 10)} → broadcast(b, (Vertex, K))
concat_where(2 <= K,  a: [Vertex],    a: [Vertex]): {Vertex: [0, 10), K: [0, 10)} → broadcast(a, (Vertex, K))
concat_where(2 <= K,  a: [K],         broadcast(a, (Vertex, K))): {Vertex: [0, 10), K: [0, 10)} → broadcast(a, (Vertex, K))
```

Guard against the inverse mistake (a branch lacking the concat dimension is still selected via the implicit broadcast and must not be treated as never-selected):

```
concat_where(K < 5,   a: [Vertex],    b: [Vertex, K]): {Vertex: [0, 10), K: [0, 10)} → unchanged
```

Added coverage (already worked on `main`):

```
concat_where(K < 0,   a: [Vertex, K], b: [Vertex, K]): {Vertex: [0, 10), K: [0, 10)} → b
concat_where(10 <= K, a: [Vertex, K], b: [Vertex, K]): {Vertex: [0, 10), K: [0, 10)} → b
concat_where(K < 10,  a: [Vertex, K], b: [Vertex, K]): {Vertex: [0, 10), K: [0, 10)} → a
concat_where(0 <= K,  a: [Vertex, K], b: [Vertex, K]): {Vertex: [0, 10), K: [0, 10)} → a
```

The unit test additionally asserts, for every case, that the result's domain annex equals the accessed domain (`RemoveBroadcast` relies on it), and, for pruned `broadcast` results, that the domains inside the branch are inferred from the full domain of the `concat_where`.

Integration: a test for the field branch with fewer dimensions than the expression was added to `test_concat_where.py`:

```
concat_where(K < 0,   a: [I, J, K],   b: [K]):         full output domain → broadcast(b, (I, J, K))
```

Since statically known domain bounds are required for the pruning to trigger — before this PR no test compiled with static bounds, so the pass had no end-to-end coverage at all — the test is parametrized on a new `static_domains` fixture: `dynamic_domains` runs with runtime bounds as before, while `static_domains` creates the field operator with `@gtx.field_operator(static_domains=True)`, making all field domains known at compile time so the pruning fires. The parametrization is extended to all tests in the file as it also improves coverage in general — in their static variant, the existing tests with never-selected branches now exercise the pass as well.

**AI disclaimer**: This PR was written with the help of AI tools. The description and pass implementation were reviewed in detail. Tests were only briefly reviewed, by putting them into a simple unified format and working with them makes me confident for them to be meaningful.


tests/next_tests/integration_tests/feature_tests/ffront_tests/test_concat_where.py
```diff
@@ -407,3 +407,32 @@ def testee(
         out=out,
         ref=(ref0, ref1),
     )
+
+
+def test_concat_where_prune_to_lower_dimensional_branch(cartesian_case):
```
tehrengruber: @tehrengruber-ai Is there no existing test for this already? Reformulate things here to be a generic test for this implicit broadcast behavior and add a note that this also tests pruning.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_prune_empty_concat_where.py
```diff
@@ -67,3 +86,111 @@ def test_prune_concat_where(accessed_domain, cond_domain, expected):
     actual = prune_empty_concat_where(testee)
     actual = InlineLambdas.apply(actual)
     assert actual == expected
+
+
+def _broadcast(expr):
```
tehrengruber: @tehrengruber-ai Make this a general ir_maker function.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_prune_empty_concat_where.py
```diff
@@ -67,3 +86,111 @@ def test_prune_concat_where(accessed_domain, cond_domain, expected):
...
+    return im.call("broadcast")(
+        expr,
+        im.make_tuple(*(itir.AxisLiteral(value=dim.value, kind=dim.kind) for dim in (Vertex, K))),
+    )
+
+
+def _concat_where(cond_range, true_branch_type, false_branch_type, accessed_domain):
```
tehrengruber: @tehrengruber-ai Add type annotations.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_prune_empty_concat_where.py
```diff
@@ -67,3 +86,111 @@ def test_prune_concat_where(accessed_domain, cond_domain, expected):
...
+    """A `concat_where` on `K` with domains inferred from `accessed_domain`."""
+    testee = im.concat_where(
+        im.domain(common.GridType.UNSTRUCTURED, {K: cond_range}),
+        im.ref("a", true_branch_type),
+        im.ref("b", false_branch_type),
+    )
+    return _infer(testee, accessed_domain)
+
+
+@pytest.mark.parametrize(
```
tehrengruber: @tehrengruber-ai Why not integrate it into the `test_prune_concat_where` above and make the test names and param comments here comments in the parameter list there.

tests/next_tests/integration_tests/feature_tests/ffront_tests/test_concat_where.py
tehrengruber-ai: The closest existing tests are `test_concat_where_scalar_broadcast_on_empty_branch` (scalar branch, but runtime bounds, so nothing prunes) and `test_concat_where_single_level_broadcast` (lower-dimensional field branch, but both branches selected). None covers a lower-dimensional *field* branch whose sibling is statically never selected. Reformulated as `test_concat_where_field_broadcast_on_empty_branch`, a generic implicit-broadcast test, with a docstring note that the statically known bounds make it also exercise `prune_empty_concat_where`'s broadcast reintroduction.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_prune_empty_concat_where.py
tehrengruber-ai: Added `im.broadcast(expr, dims)` to `ir_makers.py` and use it here and in `_prune_to`.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_prune_empty_concat_where.py
tehrengruber-ai: Done.

tests/next_tests/unit_tests/iterator_tests/transforms_tests/test_prune_empty_concat_where.py
tehrengruber-ai: Done — the four cases are now parameters of `test_prune_concat_where` under "cond disjoint from the accessed domain" / "cond covers the accessed domain" comments, and the separate test is gone.

havogt:
One case from #2767 is not covered by this PR, and since #2767 is now closed I want to flag it before it gets lost.

`_infer_concat_where` decides which region a branch is inferred on with

```python
symbolic_cond if arg == true_field else cond_complement
```

`infer_domain.py:382`. That is value equality on IR nodes, so when the two branches are structurally equal, the second iteration also compares equal to `true_field` and the false branch is inferred on the condition instead of its complement. Both branches end up with the true-branch domain.

The result is a silent wrong answer, not a crash. Minimal reproducer, plain Cartesian, no connectivity:

```python
@gtx.field_operator
def testee(a: IKField) -> IKField:
    t = a + 1.0
    return concat_where(KDim < 2, t, t)   # both branches equal, so the result is `t` everywhere

@gtx.program
def prog(a: IKField, out: IKField):
    testee(a, out=out, domain={IDim: (0, 3), KDim: (0, 4)})
```

with `a` all ones and `out` prefilled with `-1`, on `run_gtfn`:

```
[[ 2.  2. -1. -1.]
 [ 2.  2. -1. -1.]
 [ 2.  2. -1. -1.]]
```

`K >= 2` is never written. The `as_fieldop` is emitted with domain `KDim: [0, 2)` while the `SetAt` target is `KDim: [0, 4)`. Statically known bounds are required, as for the other two bugs.

| | result |
|---|---|
| `main` `0d8fb34d3` | wrong |
| this PR `808ed8f73` | wrong |
| #2767 | correct |

So this is pre-existing on `main` rather than introduced here, but #2767 fixed it as part of the same change and this PR does not. The unit tests cannot catch it because they compare IR rather than execute; the closest case, `concat_where(2 <= K, a, broadcast(a, (Vertex, K)))`, checks the structure but not the resulting domain.

The fix is to select by position instead of by value:

```python
for is_true_branch, arg in [(True, true_field), (False, false_field)]:
    ...
    symbolic_cond if is_true_branch else cond_complement,
```

Applied to `808ed8f73` this makes the reproducer correct, and `tests/next_tests/unit_tests/iterator_tests` plus `test_concat_where.py` stay green, 737 passed on the non-GPU backends.

Otherwise the PR does what #2767 did, checked rather than read: #2767's regression test passes here on every backend and fails on `main` on all four compiled CPU backends, and both end-to-end examples behave. Handling the lower-dimensional branch by materialising the broadcast is also strictly more than #2767 did, which only declined to prune in that case.

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
```diff
@@ -53,26 +67,62 @@ def visit_FunCall(self, node: itir.FunCall) -> itir.Expr:
         node = self.generic_visit(node)
 
         if cpm.is_call_to(node, "concat_where"):
+            type_inference.reinfer(node)
+            if not isinstance(node.type, (ts.FieldType, ts.ScalarType)):
+                return node
+
+            def make_broadcast_explicit(branch: itir.Expr) -> itir.Expr:
```
havogt: maybe a free function with the following signature would be nicer?

```
def make_broadcast_explicit(expr: itir.Expr, target_dimensions) -> itir.Expr:
```

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
```diff
@@ -53,26 +67,62 @@ def visit_FunCall(self, node: itir.FunCall) -> itir.Expr:
...
+            # semantics of `concat_where`.
+            node_with_explicit_broadcast, _ = infer_domain.infer_expr(
+                im.concat_where(
+                    cond_expr, make_broadcast_explicit(tb), make_broadcast_explicit(fb)
+                ),
+                node.annex.domain,
+                offset_provider={},  # not needed on field-view level expressions
+                revisit_already_inferred=False,
+            )
+            cond_expr, tb, fb = node_with_explicit_broadcast.args
```
havogt: you know my standard comment for these kind of things: make it a function to hide the details from someone reading the code.
```suggestion
            cond_expr, tb, fb = node_with_explicit_broadcast(node).args
```

havogt (commented):
lgtm, 2 structural comments. feel free to disagree. Might post an agent review as well in a bit.

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
```diff
@@ -53,26 +67,62 @@ def visit_FunCall(self, node: itir.FunCall) -> itir.Expr:
...
+                #  the domain `K: [0, 2)` of the true branch.
                 # note: as long as we visited the args we have a copy here, so no need to copy again
-                tb.annex.domain = node.annex.domain
+                new_node, _ = infer_domain.infer_expr(
+                    tb,
+                    node.annex.domain,
+                    keep_existing_domains=False,
+                    revisit_already_inferred=True,
+                    offset_provider={},
+                )
```
havogt, line 150: **`offset_provider={}` crashes here on unstructured shifts.**

Unlike the first `infer_expr` call above, which is safe *only* because `revisit_already_inferred=False` stops it from descending into anything, this call re-infers the whole surviving branch with `revisit_already_inferred=True`. Any `as_fieldop` containing an unstructured shift reaches `SymbolicDomain.translate` -> `common.get_offset({}, "V2E")`:

```
KeyError: "Offset 'V2E' not found in offset provider."
```

Repro: `concat_where(2 <= K, X, X)` with `X = as_fieldop(lambda it: deref(shift(V2E, 0)(it)))(e)`, `e: Field[[Edge]]`, inferred on `{Vertex: [0, 10), K: [0, 10)}`. `X` has no `K`, so both branch domains come out identical, `tb == fb` fires, and the re-inference blows up. This is reachable from the pipeline whenever a branch lacking the concat dimension appears twice - including the new case where an inner `concat_where` prunes to a `broadcast(a, dims)` matching its sibling.

The same call also drops `symbolic_domain_sizes`, which `infer_program` was given, so even where it does not crash the recomputed domains can disagree with the rest of the program.

Both `prune_empty_concat_where` call sites in `pass_manager.py` have the real `offset_provider` and `symbolic_domain_sizes` at hand; suggest threading them into the pass. (The `# not needed on field-view level expressions` comment above is also not correct.)

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
```diff
@@ -53,26 +67,62 @@ def visit_FunCall(self, node: itir.FunCall) -> itir.Expr:
...
+                return new_node
+
+            if not isinstance(node.annex.domain, domain_utils.SymbolicDomain):
+                return node
```
havogt, line 154: **This guard needs to move above the `tb == fb` block.**

When `node.annex.domain` is a `DomainAccessDescriptor` (e.g. `NEVER`), the equal-branch path above runs first, re-infers with the descriptor as target domain, and returns a `broadcast` node whose `annex.domain` is that descriptor. `RemoveBroadcast`, which runs immediately after in both pass-manager pipelines, then trips:

```
File ".../remove_broadcast.py", line 53, in visit_FunCall
    assert isinstance(node.annex.domain, domain_utils.SymbolicDomain)
AssertionError
```

Repro: `tuple_get(1, make_tuple(concat_where(2 <= K, a, a), c))` with `a: Field[[Vertex]]`, inferred on `{Vertex: [0, 10), K: [0, 10)}` - the unused tuple element gets `NEVER`. On `main` this returned the bare `SymRef` and did not crash.

Moving the check above `if tb == fb:` fixes it, and also avoids feeding a non-domain to `infer_expr` in the first place (with an `as_fieldop` branch that path raises `ValueError("'target_domain' cannot be 'NEVER' ...")`).

havogt (commented):
Two issues in the new `prune_empty_concat_where` logic, both reproduced locally. The rest of the diff checks out - in particular the `_infer_concat_where` fix is a genuine latent bug (`arg == true_field` was true on *both* iterations when the branches were structurally equal, so identical branches both got the true-branch domain).

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
tehrengruber: @tehrengruber-ai True, add a test case and pass the offset provider in common transforms.

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
tehrengruber: This is not an issue. When `concat_where(cond, tb, fb)` can be domain inferred, so can `tb`. The `tuple_get(1, make_tuple(concat_where(2 <= K, a, a), c))` is invalid anyway an will be removed by the collapse tuple pass.

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
tehrengruber-ai: Done — extracted as a free function, named `_broadcast_to(expr, target_dims)` (mirroring `np.broadcast_to`, incl. being a no-op when the dimensions already match).

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
tehrengruber-ai: Done — extracted as a free function `_concat_where_with_explicit_broadcast(node, offset_provider=..., symbolic_domain_sizes=...)`; the explanatory comment moved into its docstring.

src/gt4py/next/iterator/transforms/prune_empty_concat_where.py
tehrengruber-ai: Done — the pass now takes `offset_provider` and `symbolic_domain_sizes`, threaded in from both `pass_manager` call sites, and forwards them to both `infer_expr` calls (the incorrect comment is gone with it). Regression test: `test_prune_equal_branches_containing_unstructured_shift` (equal branches containing a `V2E` shift, pruned and reinferred with the actual offset provider).

havogt (approved):
lgtm

havogt:
Unrelated to this PR, but noticed while looking at `concat_where` pruning: `concat_where(dim != i, a, b)` silently evaluates to `b` everywhere.

`InferDomainOps` lowers `!=` to an intersection where the semantics is a union — `infer_domain_ops.py:91-99`:

```python
elif cpm.is_call_to(node, "not_eq"):
    # `IDim != a` -> `IDim < a & IDim > a`
    return self.visit(
        im.call("and_")(
```

`{i : i != a}` is `{i < a} ∪ {i > a}`, but `{i < a} ∩ {i > a}` is empty. `canonicalize_domain_argument` then faithfully expands the `and_`, and the true branch can never be selected:

```
out ← concat_where(c⟨ Iₕ: [-∞, 2[ ⟩, concat_where(c⟨ Iₕ: [3, ∞[ ⟩, a, b), b)
```

With `a = ones`, `b = zeros` on `I: [0, 6)`:

```
concat_where(I != 2, a, b)      -> [0. 0. 0. 0. 0. 0.]
concat_where((I<2)|(I>2), a, b) -> [1. 1. 0. 1. 1. 1.]   # expected
concat_where(I == 2, a, b)      -> [0. 0. 1. 0. 0. 0.]   # control, correct
```

Only the compiled path is affected; `Dimension.__ne__` with an integer raises `NotImplementedError` pointing at ADR 22.

This reproduces identically on `main` and on this branch — no regression from this PR, and the pruning changes here do not affect it either way. Filing it here only because it is in the same neighbourhood.

The fix looks like one token, `im.call("and_")` -> `im.call("or_")` at `infer_domain_ops.py:94` together with the comment above it; note that `test_infer_domain_ops.py:56-61` currently asserts the `and_` shape, so it pins the current behaviour.

Issue #2694, opened 2026-07-07
next: clarify who is responsible for gpu arch detection


havogt:
In #2679 we add protection for CUDAARCHs check in worker processes (inject CUDAARCHS and hide the CUDA_VISIBLE_DEVICES) such that we don't clash on CUDA context. The clean solution is probably to change the contract: each gpu backend gets a gpu architecture param and must not use it's own query (or at least not via a mechnaism that creats a CUDA context).

We didn't do the same mechnism for ROCm/HIP (which might be required as an intermediate solution if it breaks there).

PR #2796, opened 2026-08-14
fix[next]: Detect ROCm device architecture instead of hard-coding gfx942

tehrengruber:
## Problem

`_query_device_arch()` in `otf/compilation/common.py` returns a hard-coded `"gfx942"` (MI300A) for ROCm devices — an old TODO. Every GPU binary is therefore compiled for gfx942 only (`get_device_arch()` feeds both the DaCe and the gtfn/CMake build); on any other AMD GPU the fat binary contains no matching code object and the first kernel launch segfaults inside `libamdhip64` (the stub-based `hipLaunchKernel` does not return a clean error for this case). Observed on a gfx1103 APU; setting `HIPARCHS` explicitly works around it.

## Fix

Query the architecture from the device via `cupy.cuda.runtime.getDeviceProperties(0)["gcnArchName"]`, stripping the target-feature suffix (e.g. `gfx90a:sramecc+:xnack-` → `gfx90a`), with the same error handling as the CUDA branch. The `CUDAARCHS`/`HIPARCHS` environment override precedence is unchanged. Verified on a gfx1103 APU (ROCm 7.2): detection returns `gfx1103`.

**AI disclaimer**: Description and code were developed with the help of LLMs. Code was reviewed in detail, description briefly read, too verbose, but doesn't matter.

### DIFF
--- a/src/gt4py/next/otf/compilation/common.py
+++ b/src/gt4py/next/otf/compilation/common.py
@@ -38,10 +38,15 @@ def _query_device_arch() -> str | None:
             )
             return None
     elif core_defs.CUPY_DEVICE_TYPE == core_defs.DeviceType.ROCM:
-        # TODO(egparedes): Implement this properly, either parsing the output of `$ rocminfo`
-        # or using the HIP low level bindings.
-        # Check: https://rocm.docs.amd.com/projects/hip-python/en/latest/user_guide/1_usage.html
-        return "gfx942"  # MI300A
+        try:
+            arch = core_defs.cp.cuda.runtime.getDeviceProperties(0)["gcnArchName"]  # type: ignore[attr-defined]
+        except core_defs.cp.cuda.runtime.CUDARuntimeError as e:  # type: ignore[attr-defined]
+            warnings.warn(
+                UserWarning(f"Could not determine the HIP device architecture: {e}"), stacklevel=2
+            )
+            return None
+        # strip the target feature suffix, e.g. `b"gfx90a:sramecc+:xnack-"` -> `"gfx90a"`
+        return arch.decode().split(":")[0]
 
     return None
 


src/gt4py/next/otf/compilation/common.py
```diff
@@ -38,10 +38,15 @@ def _query_device_arch() -> str | None:
             )
             return None
     elif core_defs.CUPY_DEVICE_TYPE == core_defs.DeviceType.ROCM:
-        # TODO(egparedes): Implement this properly, either parsing the output of `$ rocminfo`
-        # or using the HIP low level bindings.
-        # Check: https://rocm.docs.amd.com/projects/hip-python/en/latest/user_guide/1_usage.html
-        return "gfx942"  # MI300A
+        try:
+            arch = core_defs.cp.cuda.runtime.getDeviceProperties(0)["gcnArchName"]  # type: ignore[attr-defined]
+        except core_defs.cp.cuda.runtime.CUDARuntimeError as e:  # type: ignore[attr-defined]
```
havogt, line 43: The `except` only covers `CUDARuntimeError`, but the two other realistic failure modes are unguarded, and both escape the warn-and-return-`None` contract:

- `getDeviceProperties()` returns a plain dict whose HIP-only keys depend on the cupy/ROCm build. On a cupy built against an older ROCm — where the property was `gcnArch` (an int) and `gcnArchName` was not exported — this raises `KeyError`.
- If a cupy version returns the name as `str` rather than `bytes`, `arch.decode()` raises `AttributeError`.

Either one propagates out of `_query_device_arch()` → `get_device_arch()` → `get_cmake_device_arch_option()`, so instead of the previous wrong-but-working arch you get a hard crash during compilation of every program, with no warning path. Catching `(KeyError, AttributeError)` alongside `CUDARuntimeError` — or using `.get(...)` plus a bytes/str-tolerant decode — restores the intended behaviour.

src/gt4py/next/otf/compilation/common.py
```diff
@@ -38,10 +38,15 @@ def _query_device_arch() -> str | None:
             )
             return None
     elif core_defs.CUPY_DEVICE_TYPE == core_defs.DeviceType.ROCM:
-        # TODO(egparedes): Implement this properly, either parsing the output of `$ rocminfo`
-        # or using the HIP low level bindings.
-        # Check: https://rocm.docs.amd.com/projects/hip-python/en/latest/user_guide/1_usage.html
-        return "gfx942"  # MI300A
+        try:
+            arch = core_defs.cp.cuda.runtime.getDeviceProperties(0)["gcnArchName"]  # type: ignore[attr-defined]
```
havogt, line 42: Behavioural regression in process-pool workers on ROCm: `_detect_cuda_archs()` (`src/gt4py/next/otf/runners.py:108`) deliberately returns `None` for non-CUDA, so ROCm workers get neither a `HIPARCHS` env var nor `CUDA_VISIBLE_DEVICES=""`.

Previously the ROCm branch was a constant and needed no device at all. Now every pool worker calls `getDeviceProperties(0)`, which creates a HIP context on the parent's GPU — exactly the interference the CUDA worker workaround exists to avoid — and where the device is not visible/allocatable in the worker it returns `None`, so CMake is invoked without `-DCMAKE_HIP_ARCHITECTURES` and the build falls back to the compiler default arch. That is the same class of "no matching code object" failure this PR is fixing.

Worth at least exporting `HIPARCHS` from the main process in `_pool_worker_initializer` the way `CUDAARCHS` is, or noting the gap explicitly.

havogt: Not sure if this should be fixed here, because more work and the change in itself is correct, but might hit us later...

src/gt4py/next/otf/compilation/common.py
tehrengruber-ai: I checked cupy's source history (v8 through current master) to see for which versions these two failure modes can actually occur:

**`KeyError: 'gcnArchName'`** — the key is added in `cupy_backends/cuda/api/runtime.pyx` under a *compile-time* guard that has been unchanged since it was introduced in v9.0.0b3:

```cython
IF CUPY_HIP_VERSION >= 310:  # HIP >= 3.10
    properties['gcnArchName'] = props.gcnArchName
```

so what matters is the ROCm version cupy was **built** against. The key is missing only for cupy ≤ 8.x (whose HIP builds expose only the int `gcnArch`) or cupy ≥ 9 built against ROCm 3.5–3.9 (≤ Oct 2020). On HIP ≥ 6.0, `gcnArchName` is actually the only option, since `gcnArch` was removed there (`IF 0 < CUPY_HIP_VERSION < 60000000`). With our ROCm extras (`cupy>=13.4.1,<14` for ROCm 6, `cupy-rocm-7-0>=14.0` for ROCm 7) the key is therefore always present.

**`AttributeError` from `.decode()`** — occurs in no cupy release: the field is declared `char gcnArchName[256]` in `_runtime_typedef.pxi` and assigned to the dict without post-processing in every version from 9.0.0b3 through current master, and Cython coerces `char[N]` to `bytes` (same as the well-known `bytes`-typed `'name'` key). Also verified empirically on a ROCm cupy 14.1.1 build (returns e.g. `b'gfx1103'`).

So the `CUDARuntimeError`-only `except` is sufficient for every cupy version gt4py can run with on ROCm: catching `KeyError` would only defend against pre-2021 builds far below our `>=13.4.1` floor (which don't support Python 3.12 anyway), and a bytes/str-tolerant decode defends against behavior that has never existed in any release. Happy to add `(KeyError, AttributeError)` as cheap belt-and-suspenders if you prefer regardless.

src/gt4py/next/otf/compilation/common.py
tehrengruber-ai: Tracked in #2802 (with the failure mode spelled out); the long-term contract change is #2694.

havogt (approved):

PR #2797, opened 2026-08-14
fix[next-dace]: Do not cut GPU block size down to symbolic map sizes

tehrengruber:
## Description

`GPUSetBlockSize` cuts the GPU thread block size down to the map size when the map is smaller than the requested block, so no threads are wasted. The check used SymPy's fancy comparison `(map_size < block_size) == True`, which can also be *provably* true for a symbolic map size — e.g. `5 - Max(0, N)` arising from a statically bounded domain combined with a runtime scalar `N` (`concat_where` with static domains). In that case the symbolic expression was assigned as the block size, and code generation later failed with `TypeError: Cannot convert symbols to int` (see `test_static_domains_with_runtime_scalar_condition` on `main`).

A GPU block size must be a compile-time integer, so the block size is now only cut down when the map size is a concrete integer. Following review, the check is `isinstance(map_dim_size, (int, sympy.Integer))`, which is O(1) — roughly an order of magnitude faster than `dace.symbolic.issymbolic()` (traverses the expression tree) and `str(...).isdigit()` (renders it through the SymPy printer); benchmark in [this thread](https://github.com/GridTools/gt4py/pull/2797#discussion_r3793996390).

The regression is covered by parametrizing the `concat_where` tests on `static_domains`, which reproduces the failure on GPU backends.

---

**AI disclaimer**: Fully AI-generated (implementation and description).

### DIFF
--- a/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
@@ -757,9 +757,15 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
         # TODO(phimuell): Also think of how to connect this with the loop blocking.
         assert dims_to_inspect <= 3
         for i in range(dims_to_inspect):
-            map_dim_idx_to_inspect = len(gpu_map.params) - 1 - i
-            if (map_size[map_dim_idx_to_inspect] < block_size[i]) == True:  # noqa: E712 [true-false-comparison]  # SymPy Fancy comparison.
-                block_size[i] = map_size[map_dim_idx_to_inspect]
+            map_dim_size = map_size[len(gpu_map.params) - 1 - i]
+            # Note that the comparison can be provably true for a symbolic map size, e.g.
+            #  `5 - Max(0, N)` for a statically bounded domain with a runtime scalar `N`,
+            #  but a block size must be a compile-time integer, so only cut down to
+            #  concrete sizes.
+            if not dace.symbolic.issymbolic(map_dim_size) and (
+                (map_dim_size < block_size[i]) == True  # noqa: E712 [true-false-comparison]  # SymPy Fancy comparison.
+            ):
+                block_size[i] = map_dim_size
 
         gpu_map.gpu_block_size = tuple(block_size)
         # Only set `gpu_maxnreg` if it has not been set already (default is 0),
--- a/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
+++ b/src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
@@ -15,7 +15,6 @@ import warnings
 from typing import Any, Callable, Final, Optional, Sequence, Union
 
 import dace
-import sympy
 from dace import (
     dtypes as dace_dtypes,
     properties as dace_properties,
@@ -762,8 +761,8 @@ class GPUSetBlockSize(dace_transformation.SingleStateTransformation):
             # Note that the comparison can be provably true for a symbolic map size, e.g.
             #  `5 - Max(0, N)` for a statically bounded domain with a runtime scalar `N`,
             #  but a block size must be a compile-time integer, so only cut down to
-            #  concrete sizes. The type check is much cheaper than
-            #  `dace.symbolic.issymbolic()`, which traverses the expression tree.
+            #  concrete sizes. The string check is much cheaper than
+            #  `dace.symbolic.issymbolic()`.
             if str(map_dim_size).isdigit() and map_dim_size < block_size[i]:
                 block_size[i] = int(map_dim_size)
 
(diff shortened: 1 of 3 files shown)


src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -757,9 +757,15 @@ def apply(
...
-            if (map_size[map_dim_idx_to_inspect] < block_size[i]) == True:  # noqa: E712 [true-false-comparison]  # SymPy Fancy comparison.
-                block_size[i] = map_size[map_dim_idx_to_inspect]
+            map_dim_size = map_size[len(gpu_map.params) - 1 - i]
+            # Note that the comparison can be provably true for a symbolic map size, e.g.
+            #  `5 - Max(0, N)` for a statically bounded domain with a runtime scalar `N`,
+            #  but a block size must be a compile-time integer, so only cut down to
+            #  concrete sizes.
+            if not dace.symbolic.issymbolic(map_dim_size) and (
+                (map_dim_size < block_size[i]) == True  # noqa: E712 [true-false-comparison]  # SymPy Fancy comparison.
+            ):
```
philip-paul-mueller:
```suggestion
            if str(map_dim_size).isdigit() and map_dim_size < block_size[i]:
```

While `not issymbolic(x)` can be used to test if `x` is a digit, it is super slow.
The much faster way is to convert it to a string and check if it only contains digit.
The added benefit is, that it also excludes floats (which is not an issue here).

philip-paul-mueller (commented):
From a DaCe perspective it looks okay.
But I would like to ask @iomaganaris for his opinion on that as well.

iomaganaris (commented):
Apart from Philip's comment looks good to me

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
tehrengruber-ai: I benchmarked the three variants on the expression from the failing case (`5 - Max(0, N)`, dace 2.0.0a6):

| check | concrete `Integer` | symbolic `Add` |
|---|---|---|
| `issymbolic(x)` | ~1.0 µs | ~3.2 µs |
| `str(x).isdigit()` | ~1.6 µs | ~19 µs |
| `isinstance(x, (int, sympy.Integer))` | ~80 ns | ~90 ns |

`str(x)` renders the whole expression tree through the sympy printer, so at least on current dace/sympy it is slower than `issymbolic`. Since `Range.size()` returns sympy objects where a concrete size is a `sympy.Integer`, a plain `isinstance` check is O(1) and also excludes floats/rationals. Changed to that.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
```diff
@@ -761,11 +762,10 @@ def apply(
...
             #  `5 - Max(0, N)` for a statically bounded domain with a runtime scalar `N`,
             #  but a block size must be a compile-time integer, so only cut down to
-            #  concrete sizes.
-            if not dace.symbolic.issymbolic(map_dim_size) and (
-                (map_dim_size < block_size[i]) == True  # noqa: E712 [true-false-comparison]  # SymPy Fancy comparison.
-            ):
-                block_size[i] = map_dim_size
+            #  concrete sizes. The type check is much cheaper than
+            #  `dace.symbolic.issymbolic()`, which traverses the expression tree.
+            if isinstance(map_dim_size, (int, sympy.Integer)) and map_dim_size < block_size[i]:
```
philip-paul-mueller:
```suggestion
            if str(map_dim_size).isdigit() and map_dim_size < block_size[i]:
```
Should be used as `sympy.Integer` is derived from rationals and thus this loop would also work for `3/2` which is not realistic, but it opens the trap for a lot of strange SymPy behaviour.
Please use the conversion to string, which is the _proven_ way of doing it, something you need if you are in DaCe land.

philip-paul-mueller (approved):
Beside my comment it looks okay, but please apply it.

src/gt4py/next/program_processors/runners/dace/transformations/gpu_utils.py
tehrengruber: String conversion has to traverse the expression, just like `issymbolic` which merely does `isinstance` checks, so either I am missing something or your statement is not correct. If you insist I'll take your proposal, but conversion to a string to check if something is symbolic instead of using a function designated for that purpose doesn't make sense to me.

philip-paul-mueller: If you can see [here](https://github.com/sympy/sympy/blob/16fa855354eb7bcabd3fe10993841e03b1382692/sympy/core/numbers.py#L1770) `Integer` is derived from `Rational`, thus `isinstance(x, sympy.Integer)` returns `True` also if `x` is for example `1/2` which is not what you want.
However, convertig it to string and check if it only consists of number will work.
I tested a lot of different ways and after a day or so I concluded that this is the _fastest_ (before we were switching to `str().isdigit()` we spend so much time in `issmybolic()` that it was noticable in the tracing) and most reliable way of doing it.

So please change it to `str(map_dim_size).isdigit()`.

edopao:
I propose to create a helper function to check the integer type: https://github.com/GridTools/gt4py/pull/2806

My PR can go after this one.

PR #2799, opened 2026-08-17
feat[next-dace]: fold single iteration Map dimensions

havogt:
## Description

A Map dimension such as `i = c:c+1:1` can only ever take the value `c`, but the Memlets inside its scope still refer to `i` symbolically. `MapFusionVertical` compares producer and consumer subsets with `Range.covers()`, which does not know the Map range, so a producer writing `a[c]` is not seen to cover a consumer reading `a[i - o]` even though both denote the same element, and a legal fusion is rejected.

`TrivialMapDimensionFolding` replaces the parameter of such a dimension by its value inside the Map scope. On a Map with range `1:17, 80` over the surface level of an icon4py program:

```
                       before                                    after
MapEntry -> Tasklet    gtir_tmp_32[i_Cell - 1, i_K - 80]         gtir_tmp_32[i_Cell - 1, 0]
MapEntry -> Tasklet    gtir_tmp_38[i_Cell - 1, i_K - 80]         gtir_tmp_38[i_Cell - 1, 0]
Tasklet  -> MapExit    gtir_tmp_40[i_Cell - 1, i_K - 80]         gtir_tmp_40[i_Cell - 1, 0]
```

The dimension itself is kept, unlike DaCe's native `TrivialMapElimination` which removes it. The transformation runs in `_gt_auto_process_top_level_maps` immediately before `vertical_map_fusion`, restricted to top level Maps.

On `compute_perturbed_quantities_and_interpolation` the two Maps over the surface level then fuse into one, and nothing else in the program changes. It also removes a source of non-determinism: the same program compiled to a different number of kernels depending on what else was compiled in the same session.

`ConstantPropagation` does not cover this case, it propagates symbols assigned on interstate edges and a Map parameter is not one of those.

## Testing

`test_trivial_map_dimension_folding.py`: folding rewrites the Memlets while the Map keeps its parameters and range, the SDFG computes the same result before and after, applying repeatedly yields a single application, and a Map without a single iteration dimension is left alone.

## Requirements

Reviewed on the fork first, at havogt/gt4py#73.


src/gt4py/next/program_processors/runners/dace/transformations/trivial_map_dimension_folding.py
```diff
@@ -0,0 +1,105 @@
...
+    Args:
+        only_toplevel_maps: Only process Maps that are on the top level.
+    """
+
+    map_entry = dace_transformation.PatternNode(dace_nodes.MapEntry)
+
+    only_toplevel_maps = dace_properties.Property(
+        dtype=bool,
+        default=False,
+        desc="Only process Maps that are on the top level.",
```
philip-paul-mueller:
```suggestion
        desc="See docs.",
```

src/gt4py/next/program_processors/runners/dace/transformations/trivial_map_dimension_folding.py
```diff
@@ -0,0 +1,105 @@
...
+        if self.only_toplevel_maps and graph.entry_node(map_entry) is not None:
+            return False
+
+        replacements = self._single_iteration_parameters(map_entry.map)
+        if not replacements:
+            return False
+
+        # Only apply if a parameter is still referenced, otherwise the transformation
+        #  would apply again and again on the same Map.
+        # NOTE: `include_entry` is needed because the uses are on the out edges of
```
philip-paul-mueller: The same can be said for `include_exit`, I would drop the node entirely.

philip-paul-mueller (approved):
LGTM

src/gt4py/next/program_processors/runners/dace/transformations/trivial_map_dimension_folding.py
havogt: Applied in 586c5b4d0.

src/gt4py/next/program_processors/runners/dace/transformations/trivial_map_dimension_folding.py
havogt: Right, dropped it in 586c5b4d0.

PR #2801, opened 2026-08-19
test: self-enforce `requires_*` markers, rename `requires_dace` to `uses_dace`

### LINKED ISSUE #2800: test[next]: direct `pytest` runs on CPU-only machines fail with confusing CUDA errors instead of skipping GPU tests
## Summary

On a machine without a CUDA toolkit, running the dev-loop command documented in
[`src/gt4py/next/AGENTS.md`](https://github.com/GridTools/gt4py/blob/main/src/gt4py/next/AGENTS.md)

```bash
uv run pytest tests/next_tests/ -x -q
```

produces hundreds of failures whose error messages point at `nvcc` and `cupy`, giving no
hint that the intended invocation simply excludes those tests. The `requires_gpu` markers
are correct and complete — the deselection that makes them effective just lives only in
`noxfile.py`, so it is not applied when `pytest` is invoked directly.

## Evidence

Same file, same machine (no CUDA toolkit, no `cupy`), only the marker filter differs:

| Command | Result |
| --- | --- |
| `pytest .../test_compiled_program.py -n 4` | **68 failed**, 260 passed, 6 errors |
| `pytest .../test_compiled_program.py -n 4 -m "not requires_gpu"` | **250 passed**, 80 skipped, 0 failed |

Extrapolated to the full `tests/next_tests/integration_tests` tree, the unfiltered run
reports **557 failed / 152 errors**.

## Why the failures are hard to read

Two distinct symptoms, neither of which names the real cause:

1. **Allocation** — `RuntimeError: Missing 'cupy' dependency for GPU allocation`
2. **Compilation** — `CMakeCUDAFindToolkit.cmake:104 (message): Failed to find nvcc`,
   surfaced as `gt4py.next.errors.exceptions.CompilationError: See attached compilation log.`

Symptom 2 then propagates: failed GPU compilations accumulate in the process-global
registry behind `gtx.wait_for_compilation()`, which drains *all* pending compilations.
Whichever **non-GPU** test calls it next inherits those foreign failures and dies with

```
ExceptionGroup: Multiple compilations failed: 'testee (run_gtfn_gpu)',
'testee (run_dace_gpu_opt)', ... (36 sub-exceptions)
```

So e.g. `test_wait_for_compilation[roundtrip.default-program-serial]` — a pure-CPU
roundtrip parametrization — fails with a wall of CUDA errors. Which CPU test becomes the
victim varies between runs, since it depends on which one happens to call the global drain
first, so the failure set looks non-deterministic on top of everything else.

This part is by design (the docstring documents that it waits for *all* ongoing
compilations) and it does not occur once GPU tests are correctly deselected — it just
compounds the confusion when they are not.

## Suggested fixes

Either would resolve it; the first is the more robust:

1. **Auto-skip in `conftest.py`** — when `cupy` is unimportable or no CUDA toolkit is
   detectable, skip `requires_gpu`-marked tests instead of letting them run and fail.
   This makes the marker self-enforcing regardless of how `pytest` is invoked, matching
   how `requires_dace` / `requires_jax` behave from a user's point of view.
2. **Document the filter** — note in `src/gt4py/next/AGENTS.md` (and wherever the direct
   `pytest` invocation is suggested) that CPU-only machines want
   `-m "not requires_gpu"`, mirroring `noxfile.py`.

## Notes

- Nothing here is a defect in the test matrix: the markers at
  `tests/next_tests/integration_tests/cases_utils.py:95,100,112` and the deselection at
  `noxfile.py:81` are correct, and `nox -s test_next` behaves properly on a CPU-only box.
- This is purely about the direct-`pytest` path that the contributor docs recommend for
  targeted runs.
- Found while validating an unrelated refactor; the ~90 minutes spent tracing the CUDA
  errors back to a missing marker filter is the cost this issue is about.

egparedes:
Closes #2800.

## Problem

The suite has two kinds of markers, enforced very differently:

- `uses_*` — **feature** markers, enforced in-process by `BACKEND_SKIP_TEST_MATRIX`, so they work however `pytest` is invoked.
- `requires_*` — **environment** markers, enforced *only* by the `-m` filter in `noxfile.py`.

So `requires_*` correctness depended on the caller remembering a flag — and the contributor docs recommend `uv run pytest tests/next_tests/ -x -q`, which passes none. On a CPU-only machine that produced hundreds of failures pointing at `nvcc` and `cupy`, none of which named the real cause.

Two related problems surfaced while fixing it:

- **`requires_jax` failed green.** `EmbeddedDummyBackend` holds `nd_array_field.jnp`, which is `None` without jax, and `constructors.py` treats `allocator=None` as "device default", i.e. NumPy. So `requires_jax` parametrizations *passed* without jax — on NumPy. 27 items in `test_arg_call_interface.py` alone.
- **`requires_dace` was a lie.** `dace` is an unconditional `project.dependencies` entry, so the marker never gated anything.

## Changes

**Self-enforcing markers.** `tests/conftest.py` maps each `requires_*` marker to a probe and skips unsatisfied items at collection, covering `next_tests`, `cartesian_tests` and `storage_tests` uniformly.

**`--require-optional-deps`** disables that skipping, and every nox session whose marker selection is paired with its installed extras passes it — otherwise a broken GPU environment would turn `-m requires_gpu` into an all-skipped, green run. For the same reason the main `test_next` run no longer accepts "0 collected" as success.

**`requires_dace` → `uses_dace`.** Renamed rather than dropped: the marker still splits the nox matrix into `internal` (`not uses_dace`) and `dace` jobs. It moves to the `uses_*` group, and the 55 `pytest.importorskip("dace")` calls that can never skip become plain imports — which also removes the inert guard in `test_transient_memory_mode.py` that sat *after* the imports it was meant to protect (thanks @edopao).

**`requires_atlas` joins the club.** `fvm_nabla_setup.py` guards its `atlas4py` import, so the nabla modules import without it and the marker does the skipping. `StructuredGrid("O32")` moves out of the `nabla_setup` signature, since default arguments are evaluated at import.

**`gt4py._core.definitions.gpu_device_count()`** replaces the duplicated `cupy.cuda.Device()` probes, so an importable `cupy` no longer counts as a usable GPU.

**Tests and docs.** Canaries that fail if the skipping stops, a drift test tying `pyproject.toml` markers to conftest probes, and a `pytester`-style test driving the hook with a stubbed probe so the contract holds whatever is installed. `CONTRIBUTING.md` and `src/gt4py/next/AGENTS.md` updated; the documented `pytest` invocation is now correct as written.

## Verification

`test_compiled_program.py` unfiltered, on a machine with no CUDA toolkit and no `cupy`:

| | passed | failed | errors |
| --- | --- | --- | --- |
| Before (#2800) | 260 | **68** | 6 |
| After | **248** | **0** | **0** |

The rename selects a byte-identical test set: `-m uses_dace` collects the same node IDs as `-m requires_dace` did (1830 in next, 1098 in cartesian, `diff` = 0), with complements summing to the full suites.

`atlas4py` present → 55 passed; absent → 72 skipped, 0 collection errors; installed-but-unloadable → its own `ImportError` rather than a `None` sentinel.

Local suites all green (next unit + package + core + storage + eve: 2682 passed; cartesian unit: 761 passed), `pre-commit` green, and CI green including the CSCS GPU pipeline on **both** `cuda12` and `rocm7`.

## Follow-up

#2800 also notes that failed compilations accumulate in the process-global `_ongoing_compilations` registry and get inherited by whichever test next calls `wait_for_compilation()`. That is fixed separately in #2803.


src/gt4py/next/AGENTS.md
```diff
@@ -88,4 +88,8 @@ Next tests run across a **backend matrix** (embedded NumPy/CuPy/JAX, `gtfn`,
   finds it regardless of which file it lives in).
 - Run with `uv run pytest tests/next_tests/ -x -q`; matrix-level confidence
   comes from `uv run nox -s "test_next-<py>(...)"`. GPU sessions
-  may be unavailable locally and will skip.
+  may be unavailable locally and will skip. Tests carrying a `requires_*`
+  marker (`requires_gpu`, `requires_dace`, `requires_jax`, `requires_atlas`)
+  are skipped automatically when that dependency is missing, so a CPU-only
+  machine needs no extra `-m` filter; pass `--require-optional-deps` to turn
+  those skips back into real failures.
```
egparedes:
```suggestion
  may be unavailable locally and will skip. Tests carrying a `requires_*`
  marker (`requires_gpu`, `requires_dace`, `requires_jax`, `requires_atlas`)
  are skipped automatically when that dependency is missing; pass
  `--require-optional-deps` to turn those skips back into real failures.
```

tests/conftest.py
```diff
@@ -39,13 +40,76 @@ def _get_pkg_marks(module_name: str) -> list[pytest.Mark | str]:
...
+
+        cp.cuda.Device()
+    except Exception:
+        return False
+    return True
+
+
+# Unlike the `uses_*` feature markers, the `requires_*` markers describe the
+# *environment* a test needs, so they can be enforced here instead of relying on
+# the caller passing the matching `-m` filter (which only `noxfile.py` does).
```
egparedes:
```suggestion
# `requires_*` markers describe the *environment* a test needs, so they can
# be enforced here instead of relying on the caller passing the matching
# `-m` filter (which only `noxfile.py` does).
```

CONTRIBUTING.md
```diff
@@ -136,6 +136,17 @@ pytest --ff tests/
 pytest -v -l -s tests/
 ```
 
+Tests which need an optional dependency carry a `requires_*` marker
+(`requires_gpu`, `requires_dace`, `requires_jax`, `requires_atlas`) and are
+skipped automatically when that dependency is unavailable. This keeps a plain
+`pytest tests/` usable on, for example, a machine without a GPU. Use
+`--require-optional-deps` to disable the automatic skipping and see the actual
+error instead:
```
egparedes, line 143:
```suggestion
Tests which need an optional dependency carry a `requires_*` marker
(`requires_gpu`, `requires_dace`, `requires_jax`, `requires_atlas`) and are
skipped automatically when that dependency is unavailable. Use
`--require-optional-deps` to disable the automatic skipping and see the actual
error instead:
```

noxfile.py
```diff
@@ -164,6 +164,10 @@ def test_cartesian(
 
     session.run(
         *"pytest --cache-clear -sv -n auto --dist loadgroup".split(),
+        # The marker selection below is paired with the extras installed above, so an
+        # unmet `requires_*` marker means a broken environment. Fail loudly instead of
+        # letting the auto-skip in 'tests/conftest.py' turn the session into all-skips.
+        "--require-optional-deps",
```
egparedes:
```suggestion
        # An unmet `requires_*` marker means a broken environment. Fail loudly
        # instead of skipping all tests due the auto-skip in 'tests/conftest.py'.
        "--require-optional-deps",
```

noxfile.py
```diff
@@ -255,6 +259,10 @@ def test_next(
...
```
egparedes:
```suggestion
        # An unmet `requires_*` marker means a broken environment. Fail loudly
        # instead of skipping all tests due the auto-skip in 'tests/conftest.py'.
        "--require-optional-deps",
```

noxfile.py
```diff
@@ -305,6 +313,10 @@ def test_storage(
...
```
egparedes:
```suggestion
        # An unmet `requires_*` marker means a broken environment. Fail loudly
        # instead of skipping all tests due the auto-skip in 'tests/conftest.py'.
        "--require-optional-deps",
```

src/gt4py/next/AGENTS.md
egparedes: Applied in 21f91b5c8.

tests/conftest.py
egparedes: Applied in 21f91b5c8.

CONTRIBUTING.md
egparedes: Applied in 21f91b5c8.

noxfile.py
egparedes: Applied in 21f91b5c8, at all three sites (this one plus `test_next` and `test_storage`). One deviation from the suggestion verbatim: I wrote "due **to** the auto-skip" — the suggested text reads "due the auto-skip". Say the word if you'd rather keep it exactly as suggested.

noxfile.py
egparedes: Applied in 21f91b5c8 (same wording as the `test_cartesian` site, with the "due **to**" fix noted there).

noxfile.py
egparedes: Applied in 21f91b5c8 (same wording as the `test_cartesian` site, with the "due **to**" fix noted there).

tests/conftest.py
```diff
@@ -39,13 +40,76 @@ def _get_pkg_marks(module_name: str) -> list[pytest.Mark | str]:
...
+
+    Note:
+        Same probe as in `tests/cartesian_tests/definitions.py`. Importing
+        `cupy` alone is not enough, since its wheels install fine on machines
+        without a driver or a visible device.
+    """
+    try:
+        import cupy as cp
+
+        cp.cuda.Device()
```
havogt: Let's double-check if this is creating a CUDA context. If yes, this could possible crash. As we don't see CI failures it's probably ok.

tests/conftest.py
```diff
@@ -39,13 +40,76 @@ def _get_pkg_marks(module_name: str) -> list[pytest.Mark | str]:
...
+        return False
+    return True
+
+
+# `requires_*` markers describe the *environment* a test needs, so they can
+# be enforced here instead of relying on the caller passing the matching
+# `-m` filter (which only `noxfile.py` does).
+_REQUIREMENT_PROBES: Final[dict[str, tuple[Callable[[], bool], str]]] = {
+    "requires_atlas": (lambda: _is_importable("atlas4py"), "the `atlas4py` package"),
+    "requires_dace": (lambda: _is_importable("dace"), "the `dace` package"),
```
havogt: This should probably be removed as dace is non-optional since a while. Should we drop the requires_dace marker?

havogt: Maybe before merging this PR?

tests/conftest.py
```diff
@@ -39,13 +40,76 @@ def _get_pkg_marks(module_name: str) -> list[pytest.Mark | str]:
...
+        cp.cuda.Device()
+    except Exception:
+        return False
+    return True
+
+
+# `requires_*` markers describe the *environment* a test needs, so they can
+# be enforced here instead of relying on the caller passing the matching
+# `-m` filter (which only `noxfile.py` does).
+_REQUIREMENT_PROBES: Final[dict[str, tuple[Callable[[], bool], str]]] = {
```
havogt: I think we still have some pytest.import_skip markers. We should probably delete the ones covered by these markers.

tests/conftest.py
egparedes: Maybe it would be better to rename `requires_dace` to `uses_dace`, what do you think?

tests/conftest.py
egparedes: I guess you mean the `import_or_skip()` statements in tests, right? It makes sense, yes

tests/conftest.py
egparedes: Good catch — fixed in d98f5630c by not going through `Device()` at all.

`cupy.cuda.Device()` with no argument resolves the *current* device via `cudaGetDevice`, so whether it materializes a primary context depends on CUDA version and prior runtime state. I don't have a GPU here, so rather than argue about which CUDA versions lazily initialize what, I switched to the query that provably needs no context:

```python
return cp.cuda.runtime.getDeviceCount() > 0
```

Enumerating devices is exactly the pre-context operation (`cudaGetDeviceCount`) — it is how you discover devices *before* choosing one. It is also more precise for our purpose: it distinguishes "driver present but zero visible devices" from "device usable", which `Device()` did not.

Two things worth knowing:

- Your reasoning from CI was right but for a slightly lucky reason. A probe that raises now returns `False`, which would *skip* GPU tests rather than crash — that could have gone green vacuously. It cannot, because the nox GPU sessions pass `--require-optional-deps`, which disables auto-skipping entirely. So a broken probe surfaces as real failures there, not silence.
- `tests/cartesian_tests/definitions.py:9-13` still uses the `Device()` form at import time. That is pre-existing and outside this PR, but the same concern applies — happy to fix it here or in a follow-up, your call.

tests/conftest.py
egparedes: Done in d98f5630c, for the four that this PR actually makes redundant — but not all of them, and the distinction matters:

**Removed (4).** `jax` ×2 in `test_nd_array_field.py` and `cupy` ×2 in `test_dace_backend.py`. All four sit *inside* test/fixture bodies that only run after `requires_jax` / `requires_gpu` has decided not to skip, so a plain `import` is enough now.

**Kept: `atlas4py` ×2.** These are not redundant, and removing them would be a regression. Markers are applied in `pytest_collection_modifyitems` and take effect at *setup*; these `importorskip` calls run at *collection*, strictly earlier. `fvm_nabla_setup.py:12` does an unguarded module-level `from atlas4py import (...)`, so without the guard a machine without `atlas4py` gets a collection **error** instead of a skip — which no marker can prevent.

**Kept: `dace` ×57.** These are redundant, but not because of this PR — they are redundant because `dace` is a core dependency (`project.dependencies`), so `importorskip("dace")` can never skip. That is the same observation as your `requires_dace` comment, so I have left them for whatever comes out of that thread rather than pre-empting it.

tests/conftest.py
egparedes: Some evidence for the decision, since I think **rename and drop have very different consequences** and the thread is currently mixing them.

**Confirmed: `dace` is non-optional.** It is an unconditional entry in `project.dependencies` (`dace>=2.0.0a6`) — not an extra, no environment marker. So the probe I added for `requires_dace` can never fire, and the 57 `pytest.importorskip("dace")` calls in the suite can never skip. havogt is right that the *dependency* meaning of the marker is dead.

**But the marker is still load-bearing — for selection, not for availability.** `noxfile.py` splits the test matrix with it:

```python
CodeGenTestSettings     = {"internal": {"extras": ["jax"], "markers": ["not requires_dace"]}}
CodeGenDaceTestSettings = CodeGenTestSettings | {"dace": {"extras": [], "markers": ["requires_dace"]}}
```

So `test_next` / `test_cartesian` run once as `(internal, ...)` and once as `(dace, ...)`. **Dropping** the marker collapses those into one job — a CI matrix change, not a cleanup. **Renaming** it to `uses_dace` keeps the split intact and only changes the marker string. That makes @egparedes' proposal strictly the safer of the two, and it also fixes the taxonomy this PR leans on: `uses_*` = what a test exercises, `requires_*` = what the environment must provide. `dace` clearly belongs in the first group now.

One wrinkle worth naming: in `gt4py.next` the `uses_*` family currently means *DSL features* indexed by `BACKEND_SKIP_TEST_MATRIX`, whereas `uses_dace` would name a codegen backend. Still a much better fit than `requires_*`, just not a perfect one.

**Blast radius of the rename**, if you want it here: `pyproject.toml` marker declaration, `noxfile.py` (2 settings dicts), ~57 test modules with `pytestmark = pytest.mark.requires_dace` or package-level marks, plus this PR's probe table, drift test and canary. Mechanical, but it touches a lot and would put this PR's diff well past its current scope.

I have deliberately not guessed which way to go — happy to do the rename in this PR, or land the marker work as-is and do the rename in a focused follow-up. Whichever you two settle on.

havogt: uses_dace makes sense

tests/conftest.py
havogt: I guess we can fix the cartesian case as well here. Or even go through a shared function from _core?

edopao:
It is not related to this PR, but I found that this `importskip` call should appear before importing dace transformations:
```
--- a/tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_transient_memory_mode.py
+++ b/tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests/transformation_tests/test_transient_memory_mode.py
@@ -8,10 +8,10 @@
 
 import pytest
 
-from gt4py.next.program_processors.runners.dace import transformations as gtx_transformations
-
 dace = pytest.importorskip("dace")
 
+from gt4py.next.program_processors.runners.dace import transformations as gtx_transformations
+
 
 def _make_sdfg_with_top_level_transient(storage):
     sdfg = dace.SDFG("external_mode_scalar_test")
```

I also asked the model whether this PR made it possible to avoid the `importskip` call in `tests/next_tests/unit_tests/program_processor_tests/runners_tests/dace_tests` but the answer is no. Unless we remove it since `dace` is not optional.

tests/conftest.py
egparedes: Done in efb6e7482 — renamed across 23 files / 44 occurrences, including `noxfile.py` and the marker declaration in `pyproject.toml` (reworded to *tests that exercise the `dace` backend* and moved into the alphabetical `uses_*` group).

Since the marker drives the nox matrix, I checked that the split is genuinely unchanged rather than just plausible — comparing collected node IDs before and after:

| suite | `-m requires_dace` (before) | `-m uses_dace` (after) | diff |
| --- | --- | --- | --- |
| next_tests | 1830 | 1830 | **0** |
| cartesian_tests | 1098 | 1098 | **0** |

Byte-identical lists, and the complements sum to the full suites (8214 and 2777). So the `internal` / `dace` jobs select exactly what they selected before.

`uses_dace` is also no longer in `_REQUIREMENT_PROBES` — it is a selection marker now, not an environment one — so its probe and canary are gone and the docs no longer list it among the auto-skipped markers.

tests/conftest.py
egparedes: Went with the shared function, in efb6e7482: `gt4py._core.definitions.gpu_device_count()`.

`_core.definitions` already owns the `cupy` import and `CUPY_DEVICE_TYPE`, so it was the natural home. Three call sites now go through it — the root `tests/conftest.py`, `tests/cartesian_tests/definitions.py`, and the `requires_gpu` canary — and `cupy.cuda.Device()` no longer appears as a probe anywhere in `src/` or `tests/`.

That felt worth doing beyond the immediate fix: my probe was a copy of the `Device()` idiom from `cartesian_tests/definitions.py`, so the duplication is what propagated the problem in the first place. One function means the next person cannot copy the wrong one.

Left alone: `cartesian_tests/utils.py`, `test_field_layouts.py`, `storage_tests/test_interface.py` and `benchmarks/benchmark_program_call.py` each bind `cp` on import success alone, with no device check. That is a weaker guarantee, but they are only read from `requires_gpu` tests, which now auto-skip — so they are not affected by the context concern. Happy to fold them in if you would rather have one pattern everywhere.

egparedes:
@edopao thanks — that one is fixed in efb6e7482, though by deletion rather than by reordering.

Following your own suggestion ("unless we remove it since `dace` is not optional") and @havogt's point above, `requires_dace` is now `uses_dace` and every `pytest.importorskip("dace")` is gone: 52 `dace = pytest.importorskip("dace")` became a plain `import dace`, and 3 bare guards that bound nothing were dropped. `test_transient_memory_mode.py` now imports `dace` at the top with everything else, so the ordering bug you found cannot recur there.

Two cases the sweep would have broken if applied mechanically, in case they are interesting:

- `test_dace_parsing.py` was the only *indented* occurrence — the guard was the entire body of an `else:` under `if TYPE_CHECKING:`. Removing it left a syntax error. With `dace` non-optional the whole construct collapses to one `import dace`, and `TYPE_CHECKING` became unused.
- The 3 bare guards never bound `dace`, so inserting `import dace` for them was wrong; in the two `__init__.py` files `import pytest` also went unused, leaving just the license header.

Note the `atlas4py` guards are deliberately kept. They are not equivalent: markers apply at *setup*, whereas those run at *collection*, and `fvm_nabla_setup.py` imports `atlas4py` unguarded at module level — so removing them turns a skip into a collection error.

Verified on a CPU-only machine, all green: cartesian/package/core/storage/eve 1370 passed / 62 skipped, next unit 2067 passed / 51 skipped / 11 xfailed, plus unchanged collection counts across all six suites (next 8214, cartesian 2777).

egparedes:
Pushed `96a5da3c0` + `238e00dbd`: `requires_atlas` now behaves like `requires_jax`, plus fixes from an adversarial review pass.

## `requires_atlas` is now self-enforcing

`fvm_nabla_setup.py` guards its `atlas4py` import the way `atlas_utils` does, so both nabla test modules import without atlas and the marker does the skipping. The two `pytest.importorskip("atlas4py")` calls are gone.

Two things that a naive guard would have gotten wrong:

- `grid=StructuredGrid("O32")` was a **default argument**, evaluated at class-body execution i.e. at import. Guarding only the import would have left `StructuredGrid = None` and raised `TypeError: 'NoneType' object is not callable` at import — trading one failure for another. The default is now resolved inside `__init__` (verified unchanged: 5440 nodes / 16167 edges).
- `requires_*` is probed with `importlib.util.find_spec`, which does **not** execute the module. So an `atlas4py` that is *installed but not loadable* (missing `libatlas`, ABI mismatch) is reported available and never skipped, and the `None` sentinels would surface as `NoneType is not callable`. The `ImportError` is now kept and re-raised from `nabla_setup`, so that case fails with its own message.

| `atlas4py` state | before | after |
| --- | --- | --- |
| absent | 2 module-level skips | **72 item-level skips**, each naming the package |
| present | 55 passed / 2 skipped / 9 xfailed | unchanged |
| installed but broken | `ImportError` at collection | `ImportError` with its own message, at call |

## Other fixes from the review

- **`item.keywords` was the wrong lookup.** `NodeKeywords.__init__` seeds `{node.name: True}`, so `"requires_gpu" in item.keywords` matches any node *named* after a marker, skipping it without it carrying one — a silent-skip vector inside the code meant to prevent silent skips. Now `get_closest_marker`.
- **`_is_importable` no longer lets `find_spec` escape.** It raises for a partially broken install, and this runs in a collection hook where that aborts the entire session.
- **`test_next`'s main run no longer treats "0 collected" as success.** It carried `success_codes=[0, NO_TESTS_COLLECTED_EXIT_CODE]` while `test_cartesian`/`test_storage` did not, so a broken `-m` selection would have gone green — on the very PR that renames the marker driving it. Verified all 8 combinations select tests first (5553 / 57 / 772 / 4 / 1402 / 9 / 415 / 2). **Behaviour change worth knowing:** `nox -s "test_next-3.x(...)" -- -k <no-match>` now fails with exit 5 instead of passing, which matches the other two sessions.
- **`uses_dace` documented as the exception** to "index-only markers are safe to add" in `definitions.py` and `AGENTS.md` — tagging a test with it removes that test from every `internal` session.
- **Corrected my own `gpu_device_count` docstring.** It claimed counting devices needs no CUDA context while `Device()` can create one. `cudaGetDeviceCount` also initialises the runtime, so that distinction is not established; the honest justification is that enumerating is what makes "installed but no visible device" answerable. @havogt — this walks back part of what I told you [above](https://github.com/GridTools/gt4py/pull/2801#discussion_r3821173642); the behavioural improvement stands, the rationale was overstated.
- **The GPU canary was circular** — it asserted `gpu_device_count() > 0`, the very probe that decided not to skip it. It now touches the device.
- **Added `test_auto_skip_hook_is_wired`**: runs a nested pytest session with a stubbed probe, so the contract is verified whatever is installed. Mutation-tested — replacing the skip condition with `if False` fails it; reverting passes.

## Verification

CPU-only machine, all exit 0: next unit **2067 passed**, cartesian unit **761 passed**, package+core+storage+eve **610 passed / 37 skipped**, atlas tests **55 passed** with atlas and **72 skipped** without. `pre-commit` green (ruff, format, mypy, tach, mdformat).

tests/next_tests/integration_tests/multi_feature_tests/fvm_nabla_setup.py
```diff
@@ -9,16 +9,31 @@
...
+        build_node_to_edge_connectivity,
+        functionspace,
+    )
+except ImportError as error:
+    # Keep this module importable without `atlas4py`, so that collection succeeds and
+    # the `requires_atlas` marker can skip the tests. The error is kept and re-raised
+    # from `nabla_setup`: `requires_atlas` is probed with `importlib.util.find_spec`,
+    # which does not execute the module, so an `atlas4py` that is installed but not
+    # loadable (a missing `libatlas`, an ABI mismatch) is not skipped and has to fail
+    # with its own message rather than with a `None` sentinel.
```
egparedes: Too verbose, make it more concise

egparedes: Trimmed in `9d5276b6a` — that block is 6 lines down to 3:

```python
except ImportError as error:
    # Stay importable so collection succeeds and `requires_atlas` can skip the tests.
    # `nabla_setup` re-raises the error, since the marker is probed with `find_spec`
    # and would not skip an `atlas4py` that is installed but not loadable.
```

I took it as a general signal rather than a one-off and swept the rest of what I had added: **54 added comment lines down to 37**.

| location | before | after |
| --- | --- | --- |
| this block | 6 | 3 |
| `noxfile.py` flag rationale (x3) | 5 | 3 |
| `conftest.py` `_is_importable` | 4 | 2 |
| `conftest.py` marker match | 3 | 1 |
| four others | 2-3 | 1-2 |

One of those was my own regression: the `--require-optional-deps` comment is the one you had **already trimmed to two lines** in the first review round. Answering a later comment about the flag being all-or-nothing, I regrew it to five across all three sites — undoing your edit. It is three lines now, with the extra rationale compressed to one clause. Sorry for the churn.

Also merged `upstream/main` (`d69168195`) before pushing, as requested. Clean, no file overlap; I checked specifically that none of the four incoming commits reintroduces `requires_dace` or a dace `importorskip`, since a conflict-free merge would have reverted those silently. Post-merge: `next_tests/unit_tests` + package + core + storage + eve = **2682 passed, 88 skipped, 11 xfailed**, and all six suites collect (next 8214 -> 8219, matching the tests those commits add). `pre-commit` green.

src/gt4py/_core/definitions.py
```diff
@@ -420,6 +420,30 @@ class DeviceType(enum.IntEnum):
 )
 
 
+def gpu_device_count() -> int:
```
havogt, line 423: nit: maybe the function should return `None` for no cupy installed?

tests/cartesian_tests/definitions.py
```diff
@@ -6,28 +6,25 @@
...
 
 import numpy as np
 import pytest
 
 from gt4py import cartesian as gt4pyc
+from gt4py._core import definitions as core_defs
 from gt4py.cartesian import utils as gt_utils
 
+# Only bound when a GPU is genuinely usable: an importable `cupy` is not enough.
+cp = core_defs.cp if core_defs.gpu_device_count() > 0 else None
```
havogt, line 19:
```suggestion
cp = import_cupy_if_device()
```
not sure if worth it.

havogt (approved):
lgtm, 2 minor comments feel free to skip

src/gt4py/_core/definitions.py
egparedes: I'd argue for keeping `int`, but happy to switch if you feel otherwise.

`CUPY_DEVICE_TYPE` (right above, line 418) already answers "is `cupy` there at all" — it is `None` exactly when `cupy` is absent, and `CUDA`/`ROCM` otherwise. So the "no cupy" case is not lost; it just lives in the constant that was already there.

That leaves `gpu_device_count` answering one question — *how many usable devices* — where 0 is the honest answer for both "no cupy" and "cupy but no visible device". Callers genuinely cannot act differently on those two: both mean "do not run GPU tests". Making it `int | None` would turn today's two call sites (`tests/conftest.py:56`, `tests/cartesian_tests/definitions.py:19`, both plain `> 0`) into None-handling, and `None > 0` is a `TypeError` rather than a quiet wrong answer — a sharp edge on a probe whose whole job is to be total and never raise.

Say the word and I'll change it.

tests/cartesian_tests/definitions.py
egparedes: Happy to add it if you want it — one caveat and one honest count first.

The suggestion as written would not apply: `import_cupy_if_device` does not exist anywhere in the tree, so committing it verbatim is a `NameError`. It would need a new helper, presumably next to `gpu_device_count` in `_core.definitions`:

```python
def cupy_if_gpu_available() -> ModuleType | None:
    """Return the `cupy` module when a usable GPU is present, else `None`."""
    return cp if gpu_device_count() > 0 else None
```

On "worth it": there is exactly **one** call site for the bind-`cp`-if-usable idiom (`tests/cartesian_tests/definitions.py:19`). `tests/conftest.py` only needs the boolean, and the canary needs neither. So it is a named helper in `src/` for a single line in a test module.

The argument *for* is real though, and it is the one that caused this thread in the first place: my original probe was a copy of the old `cupy.cuda.Device()` idiom from that very file, so the failure mode here is people copying a subtly wrong pattern. A named helper makes that harder than a two-part expression does.

I've left it out for now since you flagged it as uncertain and the PR is approved — but it is a two-line addition, so just say and I'll push it.

src/gt4py/_core/definitions.py
egparedes: I think both the name of the function and the docstring express correctly  the semantics, and returning 0 when no GPU device is visible (for whatever reason) is correct and simpler, in my opinion. I will leave it as it is now.

tests/cartesian_tests/definitions.py
egparedes: If this check (or similar) appears in several places in the codebase then I agree it makes sense to create a function to avoid duplication. Otherwise I think it's ok for now

egparedes: Checked, and by that criterion it stays as-is.

**The exact check — bind `cp` only when a device is usable — appears once**, at `tests/cartesian_tests/definitions.py:19`. `tests/conftest.py` needs only the boolean; the canary needs neither.

The looser "`import cupy` or `None`" pattern does appear 9 times, but a device-gated helper is the wrong tool for almost all of them:

| where | count | why a device gate would be wrong |
| --- | --- | --- |
| `src/gt4py/storage/{allocators,cartesian/utils,cartesian/interface}.py`, `src/gt4py/next/field_utils.py` | 4 | library code needs the *module* for dispatch and `isinstance` checks regardless of whether a device is currently visible |
| `src/gt4py/_core/definitions.py:45` | 1 | that is the `cp` `gpu_device_count()` itself reads — gating it would be circular |
| `tests/cartesian_tests/utils.py`, `.../test_field_layouts.py`, `tests/storage_tests/unit_tests/test_interface.py`, `tests/next_tests/benchmarks/benchmark_program_call.py` | 4 | only read from `requires_gpu`-marked tests, which now auto-skip, so the weaker guard is already sufficient |

So there is one site a helper would serve, not several — "ok for now" it is. Leaving the code unchanged and treating this thread as settled; easy to revisit if a second device-gated site ever shows up.

romanc:
Thanks for the cleanup!

PR #2803, opened 2026-08-19
test[next]: isolate the global compilation registry between tests

egparedes:
Closes the test-isolation half of #2800. Tests-only: 3 files, +83 lines, no `src/` changes.

## Problem

`compiled_program._ongoing_compilations` is process-global and `gtx.wait_for_compilation()` drains **all** of it. A compilation that fails in one test therefore stays tracked until some later test calls `wait_for_compilation()` and inherits the foreign failure.

This is not a local-invocation artefact — it has already broken CI. On #2733's `11b2802b` the [CSCS pipeline went red](https://cicd-ext-mw.cscs.ch/ci/pipeline/results/4525297225819146/42923301/2775636266?iid=22612&type=gitlab) with the correct marker filter on a properly configured runner, two tests back to back on the same xdist worker (excerpt via @havogt):

```
[gw9] PASSED  test_compiled_program.py::test_dispatch_miss_propagates_compilation_error
[gw9] FAILED  test_runners.py::test_wait_for_compilation_groups_multiple_failures
              At index 0 diff: _WorkerError('compilation failed in the worker') != ValueError('first boom')
```

`_WorkerError` is defined inside the *first* test.

The same mechanism produces the noisier local symptom that motivated #2800 — an unrelated pure-CPU parametrization drowning in another test's GPU failures:

```
ExceptionGroup: Multiple compilations failed: 'testee (run_gtfn_gpu)',
'testee (run_dace_gpu_opt)', ... (36 sub-exceptions)
```

Which test gets hit depends on which one drains first, so the failure set looks non-deterministic on top of being misattributed.

## Why this is still needed now that #2733 has merged

#2733 (`dfed778d`) fixed the **consumed** future: `_finish_compilation_job` now calls `_ongoing_compilations.pop(...)` instead of relying on the weak key, which on the failure path was defeated by the traceback keeping the future alive through a reference cycle.

That is a different path from the one here. This PR covers the **never-consumed** future — a variant that is compiled, fails, and is simply never called. Nothing pops it, the weak key never fires while the pool holds its reference, and it sits in the global registry waiting for whichever test next calls `wait_for_compilation()`. No production-side change reaches that case, because from the library's point of view nothing has gone wrong yet: the failure is still pending delivery to a caller that never arrives. Only the test harness knows the test is over.

The two are complementary; #2733 is merged into this branch and both are exercised by the runs below.

## Fix

An autouse fixture in `tests/next_tests/` clears the registry around each test.

What makes it safe to drop entries: the global drain is **not** the only reference to a compilation. `CompiledProgramsPool._compile_variant()` stores `self._compilation_jobs[key] = future` right next to the `_ongoing_compilations[future] = ...` registration, so clearing the global registry loses no result — a failed variant still raises its original error the next time it is called.

Only the *next* time, and the fixture docstring says so: the pool pops the job before re-raising and never reaches `compiled_programs`, so a further call reports a cache miss instead. Measured:

```
FIRST CALL : ValueError: original boom
SECOND CALL: RuntimeError: No program compiled for this set of static arguments ...
```

`wait_for_compilation()`'s own semantics are deliberate and unchanged. Draining every ongoing compilation is documented behaviour and correct for the API's stated purpose; the defect is that tests shared one process-global registry.

Scoped to `next_tests` because the registry lives in `gt4py.next.otf.compiled_program` and nothing outside `gt4py.next` uses it.

### Known limitation, documented in the fixture

Being function-scoped, the teardown also drops the failure of a compilation that is never awaited inside the test that submitted it. That is harmless while every compiling fixture is function-scoped — as it is today, verified by AST scan — but a module- or session-scoped one would have its failures silently swallowed. The docstring says to give such a fixture its own explicit wait.

The cleaner long-term shape is to scope the registry itself (per pool, or a `contextvar`) rather than clear a global one, which would remove this limitation instead of documenting it. That is a `src/` change with an open API question — what a process-wide `gtx.wait_for_compilation()` means once the registry is not process-wide — so it is deliberately not in this PR.

## Regression tests

Two tests beside the existing `wait_for_compilation` unit tests in `test_runners.py`, both order-independent so neither can pass vacuously depending on ordering:

- `test_wait_for_compilation_drains_foreign_entries` — pins that the drain is process-global and reports entries the caller never created, which is the mechanism the fixture contains.
- `test_reset_ongoing_compilations_untracks_pending_failures` — a reset test must not see a failure a previous one left behind.

## Verification

**A/B proof that it fixes the reported contamination.** The same two-test module (test 1 leaks a failed future, test 2 calls `wait_for_compilation()`) placed once inside `next_tests` and once outside it, so the only variable is whether the fixture applies:

| Location | Fixture applies | Result |
| --- | --- | --- |
| `tests/next_tests/unit_tests/` | yes | **2 passed** |
| `tests/cartesian_tests/unit_tests/` | no | **1 failed** — test 2 inherits test 1's `ValueError` |

Fixture ordering confirmed via `--setup-show`: `isolate_ongoing_compilations` sets up *before* the compiling fixtures and tears down after them, so the pre-`yield` clear cannot wipe a registration made during setup.

**No regressions**, on the merge commit (CPU-only machine, `-m "not requires_gpu"`, since this branch does not carry #2801's auto-skip):

- `tests/next_tests/unit_tests` — 2077 passed, 6 skipped, 11 xfailed
- `test_compiled_program.py` (integration) — 250 passed, 80 skipped, 6 xfailed, exactly the baseline recorded in #2800
- `uv run pre-commit run --files <changed>` green

**Not run locally:** GPU sessions (no device available). CI covers them.

## Note on the ruff warning in `tests/next_tests/conftest.py`

The added `from .fixtures.compilation import isolate_ongoing_compilations` is an unused import by ruff's reckoning, as is the pre-existing `uids` re-export on the line above — that is the pytest idiom for exposing fixtures from a conftest. `ruff check` excludes `^(tests/|docs/|examples/)` in `.pre-commit-config.yaml`, so neither is flagged in CI, and the new line is kept consistent with its neighbour rather than carrying a `noqa` the existing one does not have.

### DIFF
--- a/tests/next_tests/conftest.py
+++ b/tests/next_tests/conftest.py
@@ -7,3 +7,4 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 from .fixtures.common import uids
+from .fixtures.compilation import isolate_ongoing_compilations
--- a/tests/next_tests/fixtures/compilation.py
+++ b/tests/next_tests/fixtures/compilation.py
@@ -1,0 +1,41 @@
+# GT4Py - GridTools Framework
+#
+# Copyright (c) 2014-2024, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Fixtures keeping the process-global compilation state from leaking between tests."""
+
+from collections.abc import Iterator
+
+import pytest
+
+from gt4py.next.otf import compiled_program
+
+
+def reset_ongoing_compilations() -> None:
+    """Forget every in-flight compilation tracked for `gtx.wait_for_compilation()`.
+
+    `compiled_program._ongoing_compilations` is process-global, and
+    `wait_for_compilation()` drains *all* of it, which is the documented
+    behaviour. A compiled program keeps its own reference to the future in
+    `CompiledProgramsPool._compilation_jobs`, so dropping the entry here only
+    removes the program from that global drain; a failed variant still raises
+    its original error when it is called.
+    """
+    compiled_program._ongoing_compilations.clear()
+
+
+@pytest.fixture(autouse=True)
+def isolate_ongoing_compilations() -> Iterator[None]:
+    """Stop a test from inheriting compilations left behind by an earlier one.
+
+    Without this, a compilation that fails in one test stays tracked until the
+    next `gtx.wait_for_compilation()` anywhere in the process, so an unrelated
+    test drains it and fails with a foreign error. Which test is hit depends on
+    ordering, so the failures look non-deterministic. See GitHub issue #2800.
+    """
+    yield
+    reset_ongoing_compilations()
--- a/tests/next_tests/fixtures/compilation.py
+++ b/tests/next_tests/fixtures/compilation.py
@@ -23,7 +23,9 @@ def reset_ongoing_compilations() -> None:
     behaviour. A compiled program keeps its own reference to the future in
     `CompiledProgramsPool._compilation_jobs`, so dropping the entry here only
     removes the program from that global drain; a failed variant still raises
-    its original error when it is called.
+    its original error the next time it is called. Only the next one: the pool
+    pops the job before re-raising and never reaches `compiled_programs`, so a
+    further call reports a plain cache miss instead.
     """
     compiled_program._ongoing_compilations.clear()
 
@@ -36,6 +38,14 @@ def isolate_ongoing_compilations() -> Iterator[None]:
     next `gtx.wait_for_compilation()` anywhere in the process, so an unrelated
     test drains it and fails with a foreign error. Which test is hit depends on
     ordering, so the failures look non-deterministic. See GitHub issue #2800.
+
+    Note: being function-scoped, this also drops the failure of a compilation
+    that is never awaited within the test that submitted it. That is harmless
+    while every compiling fixture is function-scoped too, as it is today, but a
+    module- or session-scoped one would have its failures silently swallowed by
+    any later `wait_for_compilation()`. Give such a fixture its own explicit
+    wait instead of relying on the global drain.
     """
+    reset_ongoing_compilations()
     yield
     reset_ongoing_compilations()
--- a/tests/next_tests/unit_tests/otf_tests/test_runners.py
+++ b/tests/next_tests/unit_tests/otf_tests/test_runners.py
@@ -237,8 +237,12 @@ def test_wait_for_compilation_groups_multiple_failures():
     compiled_program.wait_for_compilation()
 
 
-def test_wait_for_compilation_inherits_failures_left_by_other_code():
-    """Document the leak the `isolate_ongoing_compilations` fixture exists to contain."""
+def test_wait_for_compilation_drains_foreign_entries():
+    """Pin the leak the `isolate_ongoing_compilations` fixture exists to contain.
+
+    The drain is process-global, so it reports entries this test never created.
+    Across tests that is what makes one test fail with another one's error.
+    """
     # The future has to stay referenced: tracking is weak, so a collected future
     # is not reported at all.
     foreign = concurrent.futures.Future()
--- a/tests/next_tests/unit_tests/otf_tests/test_runners.py
+++ b/tests/next_tests/unit_tests/otf_tests/test_runners.py
@@ -22,6 +22,8 @@ import gt4py.next as gtx
 from gt4py.next import backend as next_backend, config
 from gt4py.next.otf import arguments, compilation_tasks, compiled_program, runners
 
+from next_tests.fixtures import compilation as fixtures_compilation
+
 
 @pytest.fixture
 def process_runner(tmp_path):
@@ -233,6 +235,31 @@ def test_wait_for_compilation_groups_multiple_failures():
 
     # each failure is reported only once
     compiled_program.wait_for_compilation()
+
+
+def test_wait_for_compilation_inherits_failures_left_by_other_code():
+    """Document the leak the `isolate_ongoing_compilations` fixture exists to contain."""
+    # The future has to stay referenced: tracking is weak, so a collected future
+    # is not reported at all.
+    foreign = concurrent.futures.Future()
+    foreign.set_exception(ValueError("foreign boom"))
+    compiled_program._ongoing_compilations[foreign] = "testee (run_gtfn_gpu)"
+
+    # Nothing in this test compiled anything, yet it drains someone else's failure.
+    with pytest.raises(ValueError, match="foreign boom"):
+        compiled_program.wait_for_compilation()
+
+
+def test_reset_ongoing_compilations_untracks_pending_failures():
+    """A reset test must not see a failure a previous one left behind."""
+    foreign = concurrent.futures.Future()
+    foreign.set_exception(ValueError("foreign boom"))
+    compiled_program._ongoing_compilations[foreign] = "testee (run_gtfn_gpu)"
+
+    fixtures_compilation.reset_ongoing_compilations()
+
+    assert foreign not in compiled_program._ongoing_compilations
+    compiled_program.wait_for_compilation()  # must not raise
 
 
 def test_detect_cuda_archs_prefers_cudaarchs_env():


havogt (commented):
Reviewed this against #2733, whose CI failure turns out to be the same leak. The fixture is the right thing and I want it merged — but the framing needs correcting on two points, and the second one is load-bearing for the PR's own safety argument.

Verified by cherry-picking #2733 onto this branch: the fixture alone makes the failing ordering pass, and there is no file overlap between the two PRs (they apply in either order).

## 1. "Purely test isolation — no `src/` changes" is not accurate

There is also a production defect on the *consumed-future* path that no test-side fixture can fix.

`_finish_compilation_job` pops the future from `self._compilation_jobs` but never from `_ongoing_compilations`, relying purely on the weak key. That works on the success path. On the failure path it does not: `future.result()` re-raises, and the traceback of the exception it raises pins the `_load_artifact` frame, which holds `artifact_future`. Traceback -> frame is a reference cycle, so the entry survives until the next cyclic-GC pass rather than dying by refcount:

```
after scenario, tracked: 1
after dropping exception, tracked: 1
after gc.collect, tracked: 0
```

Consequence for users, not just tests: `wait_for_compilation()` reports a failure that was already raised at call time, which contradicts its own docstring ("Each failure is raised only once"). I pushed the one-line fix to #2733 — untrack explicitly where the pool consumes the future.

The two PRs are complementary, not redundant: this one covers the never-consumed future, #2733 covers the consumed-and-failed one. Suggest dropping the "no `src/` changes" claim or cross-referencing #2733 for that path.

## 2. The docstring of `reset_ongoing_compilations()` is wrong in a way that matters

> a failed variant still raises its original error when it is called

True only for the **first** call. Measured:

```
FIRST CALL:  original error re-raised          OK
SECOND CALL: RuntimeError: No program compiled for this set of static arguments.
```

`_finish_compilation_job` pops the job and `_load_artifact` raises before anything lands in `compiled_programs`, so the key ends up in neither dict. Since this sentence is exactly what justifies dropping the global entry, please make it exact — "raises its original error the next time it is called".

## 3. The motivation is undersold — use #2733 instead

The PR body and #2800 both motivate this with running the suite without `-m "not requires_gpu"` on a CPU-only machine, and #2800 explicitly says that part is "by design ... does not occur once GPU tests are correctly deselected". Read together, that makes the fixture look like it papers over a wrong invocation, which undersells it.

It is stronger than that. #2733 failed **in CI**, with the correct marker filter, on a properly configured runner:

```
[gw9] PASSED  test_compiled_program.py::test_dispatch_miss_propagates_compilation_error
[gw9] FAILED  test_runners.py::test_wait_for_compilation_groups_multiple_failures
              At index 0 diff: _WorkerError('compilation failed in the worker') != ValueError('first boom')
```

`_WorkerError` is defined inside the first test. Same worker, back to back, sanctioned invocation. That is real contamination independent of #2800's scenario, and it is the best evidence this PR has — worth putting in the body.

## 4. Latent masking risk — worth one sentence in the fixture docstring

Function-scoped autouse teardown means a compilation submitted in test A and never called again has its failure silently dropped by any later `wait_for_compilation()`. Latent today, not live: every `.compile(` under `tests/next_tests` is inside a test body, none in a fixture. But the day someone adds a module-scoped compiling fixture, the error disappears with no trace. A line in the docstring saves that debugging session.

## 5. `test_wait_for_compilation_inherits_failures_left_by_other_code` is misnamed

It does not exercise inheritance across tests — it hand-registers a future and drains it within the same test. It cannot test cross-test inheritance, since the autouse fixture guarantees an empty registry at entry. What it actually pins is "`wait_for_compilation` drains entries it did not create", which is worth pinning. `test_wait_for_compilation_drains_foreign_entries` would match what it does.

## 6. Nit — clear before the `yield` as well

The fixture clears only on teardown, so anything registered during collection or during a session-fixture setup that runs before the first test leaks into that test. Marginal, but the extra `clear()` is free.

## Checked, no action needed

- `tests/next_tests/fixtures/` has no `__init__.py`, but it is a PEP 420 namespace package and the existing `common.py` import proves the layout works.
- The ruff note in the PR body is right: `.pre-commit-config.yaml` excludes `^(tests/|docs/|examples/)` from `ruff-check`, so the fixture re-export idiom is consistent with the existing `uids` line and needs no `noqa`.
- `test_runners.py` on this branch: 18 passed.

egparedes:
Thanks @havogt — I reproduced both measurements before acting, and all six points hold. Addressed in f70399a (code) and a body rewrite.

### 2 & 6 — `reset_ongoing_compilations()` docstring and the pre-`yield` clear

Confirmed, and you are right that this one is load-bearing. Reproduced by turning a compiled variant back into a pending-but-failed job and calling twice with `enable_jit=False`:

```
FIRST CALL : ValueError: original boom
SECOND CALL: RuntimeError: No program compiled for this set of static arguments.
```

Docstring now says "the next time it is called", plus a sentence naming the consequence (`_finish_compilation_job` pops the job before re-raising and never reaches `compiled_programs`, so a further call is a plain cache miss). Added the `reset_ongoing_compilations()` before the `yield` too.

### 5 — misnamed test

Agreed; renamed to `test_wait_for_compilation_drains_foreign_entries` and gave it a docstring stating what it actually pins.

### 4 — masking risk

Added, with one correction that does not change your conclusion: there are already two `.compile()` calls inside fixtures — `compile_variants_testee` and `compile_variants_testee_tuple` in `integration_tests/feature_tests/ffront_tests/test_compiled_program.py`. Both are function-scoped, so the compile is still bounded by the test that requested them and the teardown is correct. The hazard therefore keys on *scope*, not on fixture-vs-test-body, so I wrote the note that way: a module- or session-scoped compiling fixture is what would go silent.

### 1 — the consumed-future path

Reproduced:

```
call raised: ValueError: original boom
key still in _compilation_jobs: False
after dropping exception, tracked: 1
after gc.collect, tracked: 0
```

One detail for the record: on this branch the frame that pins the future is `_finish_compilation_job`, not `_load_artifact` — `artifact_future.result()` is evaluated before `self.backend.load_artifact(...)` is entered, so on the failure path `load_artifact` never runs. The cycle is the same one you describe (future → exception → traceback → frame → future) and the conclusion is unchanged.

Dropped the "purely test isolation — no `src/` changes" framing. The body now states that the tests-only diff is a fact about this PR, not a claim that the production side is clean, and cross-references #2733 for the consumed-and-failed path.

### 3 — motivation

Agreed, and I found the run so it can be linked rather than quoted: the failure is on #2733's `11b2802b`, where [`cscs/default` is red](https://cicd-ext-mw.cscs.ch/ci/pipeline/results/4525297225819146/42923301/2775636266?iid=22612&type=gitlab). That is why it does not show up in `gh run list` — it was the CSCS pipeline, not GitHub Actions. Promoted to the top of the Problem section as the primary motivation, with #2800's CPU-only symptom kept as the secondary one.

### Validation

- `tests/next_tests/unit_tests` — 2072 passed, 6 skipped, 11 xfailed, unchanged from the baseline
- `test_runners.py` — 18 passed
- `--setup-show` still shows `SETUP F` / `TEARDOWN F isolate_ongoing_compilations`
- `uv run pre-commit run --files ...` green

Merge order is still free either way — no file overlap with #2733.

havogt:
Just a random thought and not saying it's better: a config flag to disable keeping track of ongoing compilations could also be an option?

egparedes:
Thought it through — I don't think it replaces the fixture here, but there is a stronger version of the idea worth keeping.

**A flag can't just be off for `next_tests`.** `test_wait_for_compilation`, `test_wait_for_compilation_raises_on_failed_compilation` and the `test_runners.py` unit tests all need tracking to work, so it would have to be toggled per test. And toggling does not isolate on its own: entries registered while it was on stay in the dict when you turn it off, so you would still need to clear on disable — the fixture, plus a new public knob on top.

**It also has a user-facing cost.** `gtx.wait_for_compilation()` documents that it raises the failure. A flag that makes it silently report nothing is a quiet footgun for whoever sets it and then never hears about a broken compile — a worse failure mode than the test-ordering one. Per AGENTS.md a public config knob is ADR-shaped too, which is heavy for a test-isolation fix.

**What it would genuinely buy** is an opt-out for suites that never call `wait_for_compilation()`, which would sidestep the masking risk I noted in the fixture docstring. Real, but narrow.

**The stronger version.** The root cause is not that tracking exists, it is that the registry is process-global. Scoping it — per pool, or a `contextvar` — makes isolation fall out for free and removes the masking risk entirely rather than documenting it. That is a `src/` change with a genuine open question (what a process-wide `gtx.wait_for_compilation()` means once the registry is not process-wide), so I would rather not smuggle it into this PR. Happy to open an issue for it if you think it is worth pursuing — it would supersede both this fixture and the `_ongoing_compilations` half of #2733.

egparedes:
@havogt ready for another round.

**Merged current `main`** (`f3210d1`), which brought in #2733 and #2734 — both land right on this PR's subject matter, so I re-checked rather than assuming:

- **#2733 (`dfed778d`)** is the production-side fix for the consumed-future path. `_finish_compilation_job` now does `_ongoing_compilations.pop(artifact_future, None)`. The body no longer describes it as an open sibling that can merge in either order.
- **#2734 (`b1cdb475`)** rewrote the "No program compiled for this set of static arguments" error to name the program and its static args. That is precisely the error the corrected `reset_ongoing_compilations()` docstring points at, so I confirmed the wording still holds: the docstring says a further call "reports a plain cache miss instead", and only the message text got richer — the behaviour is unchanged.

No conflicts, and the fixture's safety argument is untouched: the pool still keeps its own `_compilation_jobs` reference, so dropping the global entry still loses no result.

**Since your review**, in `f70399a8`: docstring corrected to "the next time it is called" plus the cache-miss consequence, masking-risk note added, `test_wait_for_compilation_drains_foreign_entries` renamed, clear added before the `yield`. Details and the two measurements are in the comment above; my take on the config-flag idea is in the one after it.

**Validation on the merge commit:**

- `tests/next_tests/unit_tests` — 2077 passed, 6 skipped, 11 xfailed (2072 -> 2077 is purely the tests #2733 and #2734 added)
- `test_compiled_program.py` (integration) — 250 passed, 80 skipped, 6 xfailed, still exactly the #2800 baseline
- `test_runners.py` + unit `test_compiled_program.py` — 37 passed
- `uv run pre-commit run --files ...` green

One note on the previous push: `cscs/default` went red on `f70399a8` and then green on a retry of the same pipeline without anyone touching it. Worth knowing it is flaky — it is also red on 3 of the last 12 `main` commits, including the v1.2.1 release commit.

havogt (approved):
lgtm

PR #2806, opened 2026-08-20
refactor[next-dace]: add utility to replace str().isdigit() checks

edopao:
Introduce a new utility function `is_compile_time_size` in `sdfg_utils.py` to check if an expression is a non-negative integer literal. Replace repeated `str(x).isdigit()` checks across DaCe-related files (library nodes, lowering, transformations, workflow bindings) with this utility for consistency.


src/gt4py/next/program_processors/runners/dace/sdfg_utils.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+from __future__ import annotations
+
+from typing import Any
+
+
+def is_compile_time_integer(expr: Any) -> bool:
```
philip-paul-mueller: The name does not convey the fact that only non negative integers are checked.
So we should change it to `is_non_neg_int_literal()` (which after I write it does not look greate).
However, I have no strong opinion on it and am technically file to accept it also with that name, it is just confusing.

to check all integer we could also check `stringified_input[0] == '-'`, but I think that non negative integers _are_ more useful at least considering the current usage.

src/gt4py/next/program_processors/runners/dace/sdfg_utils.py
```diff
@@ -0,0 +1,41 @@
+# GT4Py - GridTools Framework
+#
+# Copyright (c) 2014-2024, ETH Zurich
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
```
philip-paul-mueller, line 6: About the name, I am totally fine with it but have you considered naming it just `utils.py`?
I am asking because currently there is nothing SDFG specific in that file, it is more a DaCe utility function file.
But as I said, I am fully fine with it.

src/gt4py/next/program_processors/runners/dace/sdfg_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        True
+        >>> is_compile_time_integer(1 / 3)
+        False
+        >>> import sympy
+        >>> is_compile_time_integer(sympy.Rational(1, 3))
+        False
+        >>> is_compile_time_integer(sympy.Symbol("N"))
+        False
+    """
+    return str(expr).isdigit()
```
philip-paul-mueller, line 47:
```suggestion
    # NOTE: This approach was selected because benchmarking on the full pipeline showed that it was the fastest and most robust way of doing it.
    return str(expr).isdigit()
```

philip-paul-mueller (approved):
LGTM.

src/gt4py/next/program_processors/runners/dace/sdfg_utils.py
edopao: what about `is_size_literal()` ?

edopao: `is_compile_time_size()` ?

philip-paul-mueller: This is a very good name.

PR #2807, opened 2026-08-20
fix[dace]: avoid `set` for dace node connectors, use `dict` to ensure deterministic lowering

edopao:
This PR fixes the warnings:
`Using sets as inputs is discouraged as it leads to indeterministic behavior`

All warnings in GT4Py-next are false alarms, since the set is either empty or contains one element. The corresponding warning won't be emitted in dace after https://github.com/spcl/dace/pull/2511

The warning from GT4Py cartesian is real, and this PR fixes the issue.

### DIFF
--- a/paseo.json
+++ b/paseo.json
@@ -1,0 +1,5 @@
+{
+  "worktree": {
+    "setup": "uv sync"
+  }
+}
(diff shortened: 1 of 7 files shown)


philip-paul-mueller (approved):
LGTM

romanc (commented):
Thank you!

FlorianDeconinck:
Thank you!

PR #2814, opened 2026-08-22
fix[eve]: support type aliases recursing through a container

tehrengruber:
`NestedTuple` and its siblings in `gt4py.eve.extended_typing` were defined as plain assignments holding a string forward reference to themselves:

    NestedTuple = tuple[Union[_T_co, "NestedTuple[_T_co]"], ...]

Subscripting such an alias does not substitute the type parameter inside the forward reference, so `NestedTuple[Foo]` expands to `tuple[Foo | ForwardRef('NestedTuple[_T_co]'), ...]` and `get_type_hints()` fails with `NameError: name '_T_co' is not defined` when resolving it in the namespace of the annotated object. These aliases could therefore not be used as datamodel/node field annotations at all, which is why #2487 had to fall back to `target: Any` instead of `NestedTuple[DataSymbol]` in `foast.TupleComprehensionMapper`.

Changes:

- The `Nested*` / `MaybeNested*` aliases are now defined with the PEP 695 `type` statement, which substitutes type parameters properly across the recursion.
- `SimpleTypeValidatorFactory` breaks the resulting cycle. Such an alias is well founded, unlike the `type A = A` cycle `eval_type_alias()` rejects: a single resolution step already yields an annotation which is not an alias itself. What it does not yield is a *finite* one, so the alias occurring inside its own definition is now handed a deferred validator, filled in as soon as the definition has been processed.

`test_recursive_type_alias_is_not_supported` was renamed to `test_cyclic_type_alias_is_not_supported`, since only degenerate cycles remain unsupported.

Once this is merged, the `target: Any` workaround in #2487 can be removed.

**Disclaimer**: This PR and its description were written largely with the help of AI. Code was reviewed briefly by me and in more detail by the reviewer.

### DIFF
--- a/src/gt4py/eve/extended_typing.py
+++ b/src/gt4py/eve/extended_typing.py
@@ -187,19 +187,14 @@ class ArgsOnlyCallable(Protocol[_A, _R]):
 
 _T_co = TypeVar("_T_co", covariant=True)
 
-# Note: these recursive aliases have to be defined with the PEP 695 `type` statement. Written as
-# plain assignments (`NestedTuple = tuple[Union[_T_co, "NestedTuple[_T_co]"], ...]`), subscription
-# does not substitute the type parameter inside the string annotation, so `NestedTuple[int]` keeps
-# an unparametrized `ForwardRef("NestedTuple[_T_co]")` which cannot be resolved in the namespace
-# of the annotated object (see `get_partial_type_hints`).
-type NestedSequence[T] = Sequence[T | NestedSequence[T]]
-type NestedList[T] = list[T | NestedList[T]]
-type NestedTuple[T] = tuple[T | NestedTuple[T], ...]
+type NestedSequence[_T_co] = Sequence[_T_co | NestedSequence[_T_co]]
+type NestedList[_T_co] = list[_T_co | NestedList[_T_co]]
+type NestedTuple[_T_co] = tuple[_T_co | NestedTuple[_T_co], ...]
 
-type MaybeNested[T] = T | NestedSequence[T]
-type MaybeNestedInSequence[T] = T | NestedSequence[T]
-type MaybeNestedInList[T] = T | NestedList[T]
-type MaybeNestedInTuple[T] = T | NestedTuple[T]
+type MaybeNested[_T_co] = _T_co | NestedSequence[_T_co]
+type MaybeNestedInSequence[_T_co] = _T_co | NestedSequence[_T_co]
+type MaybeNestedInList[_T_co] = _T_co | NestedList[_T_co]
+type MaybeNestedInTuple[_T_co] = _T_co | NestedTuple[_T_co]
 
 
 def is_nested_tuple_of(value: object, type_: type[_T_co]) -> TypeIs[NestedTuple[_T_co]]:
--- a/src/gt4py/eve/extended_typing.py
+++ b/src/gt4py/eve/extended_typing.py
@@ -186,14 +186,20 @@ class ArgsOnlyCallable(Protocol[_A, _R]):
 
 
 _T_co = TypeVar("_T_co", covariant=True)
-NestedSequence = Sequence[Union[_T_co, "NestedSequence[_T_co]"]]
-NestedList = list[Union[_T_co, "NestedList[_T_co]"]]
-NestedTuple = tuple[Union[_T_co, "NestedTuple[_T_co]"], ...]
 
-MaybeNested = Union[_T_co, NestedSequence[_T_co]]
-MaybeNestedInSequence = Union[_T_co, NestedSequence[_T_co]]
-MaybeNestedInList = Union[_T_co, NestedList[_T_co]]
-MaybeNestedInTuple = Union[_T_co, NestedTuple[_T_co]]
+# Note: these recursive aliases have to be defined with the PEP 695 `type` statement. Written as
+# plain assignments (`NestedTuple = tuple[Union[_T_co, "NestedTuple[_T_co]"], ...]`), subscription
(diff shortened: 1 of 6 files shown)


src/gt4py/eve/extended_typing.py
```diff
@@ -186,14 +186,20 @@ def __call__(self, *args: _A) -> _R: ...
...
-MaybeNested = Union[_T_co, NestedSequence[_T_co]]
-MaybeNestedInSequence = Union[_T_co, NestedSequence[_T_co]]
-MaybeNestedInList = Union[_T_co, NestedList[_T_co]]
-MaybeNestedInTuple = Union[_T_co, NestedTuple[_T_co]]
+
+# Note: these recursive aliases have to be defined with the PEP 695 `type` statement. Written as
+# plain assignments (`NestedTuple = tuple[Union[_T_co, "NestedTuple[_T_co]"], ...]`), subscription
+# does not substitute the type parameter inside the string annotation, so `NestedTuple[int]` keeps
+# an unparametrized `ForwardRef("NestedTuple[_T_co]")` which cannot be resolved in the namespace
+# of the annotated object (see `get_partial_type_hints`).
```
egparedes: I find this comment verbose and basically unnecessary since the codebase should be upgraded soon to >= 3.12 only, so all aliases will become PEP 695 `type` statements anyway.

src/gt4py/eve/extended_typing.py
```diff
@@ -186,14 +186,20 @@ def __call__(self, *args: _A) -> _R: ...
...
+type NestedSequence[T] = Sequence[T | NestedSequence[T]]
+type NestedList[T] = list[T | NestedList[T]]
+type NestedTuple[T] = tuple[T | NestedTuple[T], ...]
+
+type MaybeNested[T] = T | NestedSequence[T]
+type MaybeNestedInSequence[T] = T | NestedSequence[T]
+type MaybeNestedInList[T] = T | NestedList[T]
+type MaybeNestedInTuple[T] = T | NestedTuple[T]
```
egparedes: Why changing from `_T_co` to `T` as type variable here? Shouldn't the covariant type var be used here since a tuple is an immutable container?

src/gt4py/eve/type_validation.py
```diff
@@ -135,6 +135,26 @@ def __call__(
 
 
 # Implementations
+@dataclasses.dataclass
+class _DeferredTypeValidator:
```
egparedes, line 139: I might be wrong but isn't this class a duplication of the `datamodels.core.ForwardRefValidator`? Would it be worth to refactor the code to define it only once in the proper place (probably defined here and used from datamodels)?

egparedes (commented):
I have a couple of comments to make the PR cleaner.

src/gt4py/eve/extended_typing.py
tehrengruber: @tehrengruber-ai Remove comment.

src/gt4py/eve/extended_typing.py
tehrengruber: @tehrengruber-ai: Keep the previous style as requested.

src/gt4py/eve/type_validation.py
tehrengruber: @tehrengruber-ai Investigate.

src/gt4py/eve/extended_typing.py
tehrengruber-ai: Removed.

src/gt4py/eve/extended_typing.py
tehrengruber-ai: No good reason — that was collateral from switching to the PEP 695 syntax, not intentional. Reverted: the aliases take `_T_co` again.

Note that PEP 695 has no syntax for declaring variance on a type parameter (it is inferred at the use site), so unlike the old `TypeVar("_T_co", covariant=True)` the name is now the only thing carrying the intent. But you are right that it is the correct one for these containers, and keeping it also keeps the aliases consistent with `is_nested_tuple_of` / `is_maybe_nested_in_tuple_of` below, which still use the module-level `_T_co`.

src/gt4py/eve/type_validation.py
tehrengruber-ai: Investigated. They look alike, but the contract differs in both directions, so I kept them separate:

- **Different protocol.** `ForwardRefValidator` is an `attrs` *field* validator, `(instance, attribute, value)` — that signature is what `field_type_validator_factory` has to hand back to `attrs`. `_DeferredTypeValidator` is a plain `FixedTypeValidator`, `(value, **kwargs)`, which is what gets composed into the surrounding validators here. `type_validation` also has no notion of datamodels/`attrs`, and the dependency only runs the other way.
- **Different filling mechanism.** `ForwardRefValidator` is *lazy and self-resolving*: on its first call it goes through `instance` to the model class, runs `update_forward_refs(model_cls)` and only then can look up the field annotation and build the real validator — it cannot do any of that at construction time. `_DeferredTypeValidator` resolves nothing: the annotation is already in hand, and the factory assigns `.validator` on the very next line, before the composed validator is returned and therefore before any value can reach it. Hence the `assert` where the other one has a resolution step.

A shared class would have to carry both call signatures and both filling modes, which is more machinery than the ~8 lines each costs today. The one unification that would type-check is `ForwardRefValidator` *holding* a `_DeferredTypeValidator` instead of its own `validator` field, but that is pure indirection — it removes no code and hides the lazy resolution one level deeper.

I did add a sentence to the `_DeferredTypeValidator` docstring pointing at the distinction, so the next reader does not have to redo this comparison. Happy to go the other way if you still prefer a single place.

egparedes (approved):
LGTM

PR #2821, opened 2026-08-25
fix[next]: make the ITIR pretty printer and parser actually round-trip

havogt:
`pparse` is meant to invert `pformat`, but nothing in the test suite composed the
two: `test_pretty_printer.py` never imports `pparse` and `test_pretty_parser.py`
never imports `pformat`. Two tokens the printer emits were therefore never
parseable.

What changed:

- New `test_pretty_roundtrip.py` asserting `pparse(pformat(t)) == t`, parametrised
  over the union of the terms the two existing test files use. The cases that fail
  on `main` go in as `xfail(strict=True)` and are cleared by the commits below.
- `INFINITY_LITERAL: "∞" | "-∞"` added to the grammar, with its transformer method.
  `InfinityLiteral` is a distinct node and `-∞` is one node, not `minus(∞)`, so one
  terminal covers both signs. `a - ∞`, `a-∞` and `a - -∞` still parse as `minus`.
- `<=` and `>=` added as `prec4` alternatives, with their transformer methods.
  `BINARY_OPS` had them; the grammar had only `==`, `<`, `>`.
- `pparse` re-raises `exc.orig_exc` instead of letting lark's `VisitError` reach the
  caller.

One xfail group remains, `Literal.type` never reaching the output. It is cleared by
the follow-up PR.

### DIFF
--- a/src/gt4py/next/iterator/pretty_parser.py
+++ b/src/gt4py/next/iterator/pretty_parser.py
@@ -8,7 +8,13 @@
 
 from typing import Union
 
-from lark import lark, lexer as lark_lexer, tree as lark_tree, visitors as lark_visitors
+from lark import (
+    exceptions as lark_exceptions,
+    lark,
+    lexer as lark_lexer,
+    tree as lark_tree,
+    visitors as lark_visitors,
+)
 
 from gt4py.next.iterator import ir
 from gt4py.next.iterator.ir_utils import ir_makers as im
@@ -296,4 +302,7 @@ class ToIrTransformer(lark_visitors.Transformer):
 def pparse(pretty_str: str) -> ir.Node:
     parser = lark.Lark(GRAMMAR, parser="earley")
     tree = parser.parse(pretty_str)
-    return ToIrTransformer(visit_tokens=True).transform(tree)
+    try:
+        return ToIrTransformer(visit_tokens=True).transform(tree)
+    except lark_exceptions.VisitError as e:
+        raise e.orig_exc from None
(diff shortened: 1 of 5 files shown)


src/gt4py/next/iterator/pretty_parser.py
```diff
@@ -130,6 +141,11 @@ def OFFSET_LITERAL(self, value: lark_lexer.Token) -> ir.OffsetLiteral:
     def ID_NAME(self, value: lark_lexer.Token) -> str:
         return value.value
 
+    def INFINITY_LITERAL(self, value: lark_lexer.Token) -> ir.InfinityLiteral:
+        if value.value.startswith("-"):
+            return ir.InfinityLiteral.NEGATIVE
+        return ir.InfinityLiteral.POSITIVE
```
tehrengruber:
```suggestion
        if value.value == "-∞":
            return ir.InfinityLiteral.NEGATIVE
        assert value.value == "∞":
        return ir.InfinityLiteral.POSITIVE
```

tehrengruber (approved):
I would use ir_makers instead of "hand" crafting the IR, but not essential.

src/gt4py/next/iterator/pretty_parser.py
```diff
@@ -103,9 +112,11 @@ class ToIrTransformer(lark_visitors.Transformer):
     def SYM(self, value: lark_lexer.Token) -> ir.Sym:
         return ir.Sym(id=value.value)
 
-    def SYM_REF(self, value: lark_lexer.Token) -> Union[ir.SymRef, ir.Literal]:
+    def SYM_REF(self, value: lark_lexer.Token) -> Union[ir.SymRef, ir.Literal, ir.NoneLiteral]:
         if value.value in ("True", "False"):
             return im.literal(value.value, "bool")
+        if value.value == "None":
```
tehrengruber: Removed in https://github.com/GridTools/gt4py/pull/2823

PR #2822, opened 2026-08-25
feat[next]: type annotations in the ITIR pretty-printed syntax

havogt:
`itir.Literal.type` participates in equality but was never printed, so `pparse`
re-typed the lexeme from scratch and any literal whose type differs from the one
implied by its spelling came back as a different term:

    pformat(im.literal("1.0", "float32"))   # '1.0' -> reparses as float64
    pformat(im.literal("1", "int64"))       # '1'   -> reparses as int32

What changed:

- A literal may carry a postfix type annotation: `1.0:f32`, `1:i64`, `True:i1`. No
  space, as in `named_range`'s `KDimᵥ:`. It binds at `prec9`, tighter than every
  operator. `:` was previously used only by `named_range`, guarded on both sides.
- The annotation takes a bare scalar name and nothing else. Admitting bracketed
  forms would make `1:i16[2]` ambiguous with `tuple_get`, which is also
  `prec8 "[" prec0 "]"`; `1:i16[2]` is therefore a tuple index. Only a literal may
  be annotated, not a symbol.
- Type names are LLVM/MLIR spellings: `bool` is `i1`, `int32` is `i32`, `float64`
  is `f64`. `SCALAR_TYPE_NAMES` in `pretty_printer` is the single forward mapping
  and `pretty_parser` imports the derived reverse map, so respelling a type is a
  one-line edit. `ts.ScalarType.__str__` is deliberately unchanged; it serves
  `foast_pretty_printer` and error messages.
- `ScalarKind.STRING` has no LLVM spelling and is excluded. The table covers
  exactly `builtins.TYPE_BUILTINS`, asserted in a test so the two cannot drift.
- The annotation is written only where it is needed to reconstruct the literal, so
  `1`, `1.0` and `True` print as before. `implied_literal_type` reads the lexeme
  back to a Python value and types it with `type_translation.from_value`, which is
  what the frontend applies to a constant, and the parser calls the same function
  rather than mirroring it.
- A literal whose value is not a boolean, integer or floating point lexeme is
  rejected with a `ValueError` rather than printed; neither `hello` nor `hello:i32`
  would parse back.
- `Temporary.dtype` with a shape or a tuple printed but did not parse; the
  `declaration` rule took only `TYPE_LITERAL: CNAME`. It now takes a `type_expr`
  sub-grammar, and a tuple dtype is spelled `{f64, f64}`, as `make_tuple` already
  prints.

Breaking:

- `pformat` emits `dtype=f64` where it emitted `dtype=float64`.
- `pformat` emits `dtype={f64, f64}` where it emitted `dtype=tuple[float64, float64]`.
- `pparse` no longer accepts `dtype=float64` or `1.0:float32`. One spelling, both
  directions, deliberately not a compatibility shim.
- `pformat`, and with it `str()` on any `itir` node, raises `ValueError` on a
  `Literal` whose value is not a boolean, integer or floating point lexeme, where it
  printed the value verbatim.

### DIFF
--- a/src/gt4py/next/iterator/pretty_parser.py
+++ b/src/gt4py/next/iterator/pretty_parser.py
@@ -114,9 +114,7 @@ GRAMMAR = """
 
 def _bare_literal(value: str) -> ir.Literal:
     """A literal written without a type annotation."""
-    type_ = pretty_printer.implied_literal_type(value)
-    assert type_ is not None, f"'{value}' is not a literal lexeme."
-    return ir.Literal(value=value, type=type_)
+    return ir.Literal(value=value, type=pretty_printer.implied_literal_type(value))
 
 
 @lark_visitors.v_args(inline=True)
--- a/src/gt4py/next/iterator/pretty_parser.py
+++ b/src/gt4py/next/iterator/pretty_parser.py
@@ -16,8 +16,7 @@ from lark import (
     visitors as lark_visitors,
 )
 
-from gt4py.next.iterator import ir
-from gt4py.next.iterator.pretty_printer import SCALAR_TYPE_KINDS, implied_literal_type
+from gt4py.next.iterator import ir, pretty_printer
 from gt4py.next.type_system import type_specifications as ts
 
 
@@ -115,7 +114,7 @@ GRAMMAR = """
 
 def _bare_literal(value: str) -> ir.Literal:
     """A literal written without a type annotation."""
-    type_ = implied_literal_type(value)
+    type_ = pretty_printer.implied_literal_type(value)
     assert type_ is not None, f"'{value}' is not a literal lexeme."
     return ir.Literal(value=value, type=type_)
 
@@ -137,11 +136,12 @@ class ToIrTransformer(lark_visitors.Transformer):
         return _bare_literal(value.value)
 
     def TYPE_LITERAL(self, value: lark_lexer.Token) -> ts.ScalarType:
-        if (kind := SCALAR_TYPE_KINDS.get(value.value)) is not None:
-            return ts.ScalarType(kind=kind)
-        raise ValueError(
-            f"Invalid type '{value}'; expected one of {', '.join(sorted(SCALAR_TYPE_KINDS))}."
-        )
+        if (kind := pretty_printer.SCALAR_TYPE_KINDS.get(value.value)) is None:
+            raise ValueError(
+                f"Invalid type '{value}'; expected one of "
+                f"{', '.join(sorted(pretty_printer.SCALAR_TYPE_KINDS))}."
+            )
+        return ts.ScalarType(kind=kind)
 
     def shaped_scalar_type(self, type_: ts.ScalarType, *shape: ir.Literal) -> ts.ScalarType:
         return ts.ScalarType(kind=type_.kind, shape=[int(s.value) for s in shape])
--- a/src/gt4py/next/iterator/pretty_parser.py
+++ b/src/gt4py/next/iterator/pretty_parser.py
@@ -17,8 +17,7 @@ from lark import (
 )
 
 from gt4py.next.iterator import ir
-from gt4py.next.iterator.ir_utils import ir_makers as im
-from gt4py.next.iterator.pretty_printer import SCALAR_TYPE_KINDS
+from gt4py.next.iterator.pretty_printer import SCALAR_TYPE_KINDS, implied_literal_type
 from gt4py.next.type_system import type_specifications as ts
 
 
@@ -85,6 +84,7 @@ GRAMMAR = """
         | "c⟨" ( prec0 "," )* prec0? "⟩" -> cartesian_domain
 
     ?prec9: _literal
+        | typed_literal
         | SYM_REF
         | named_range
         | cartesian_offset
@@ -95,6 +95,10 @@ GRAMMAR = """
     else_branch_seperator: "else"
     if_stmt: "if" "(" prec0 ")" "{" ( stmt )* "}" else_branch_seperator "{" ( stmt )* "}"
 
+    // Only a bare scalar name: `Literal.type` is a `ts.ScalarType`, and admitting
+    // the bracketed forms here would make `1:i16[2]` and `1:tuple[f32]` ambiguous
+    // with `tuple_get`, which is also `prec8 "[" prec0 "]"`.
+    typed_literal: ( INT_LITERAL | FLOAT_LITERAL | SYM_REF ) ":" TYPE_LITERAL
     ?type_expr: TYPE_LITERAL
         | TYPE_LITERAL "[" INT_LITERAL ("," INT_LITERAL)* "]" -> shaped_scalar_type
         | "tuple" "[" type_expr ("," type_expr)* "]" -> tuple_type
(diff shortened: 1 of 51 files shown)


src/gt4py/next/iterator/pretty_printer.py
```diff
@@ -63,15 +65,88 @@
 }
 
 
+#: Surface spelling of the scalar types, LLVM/MLIR style.  Sole definition of the
+#: ITIR type vocabulary: `pretty_parser` inverts this table instead of carrying its
+#: own, so respelling a type means editing here and nowhere else.  Deliberately not
+#: `ts.ScalarType.__str__`, which serves the frontend printer and error messages
+#: and keeps the `ScalarKind` names.  Covers exactly `builtins.TYPE_BUILTINS`;
+#: `ScalarKind.STRING` has no LLVM spelling and is excluded, so printing or parsing
+#: one is an error.
```
havogt:
```suggestion
#: Surface spelling of the scalar types, LLVM/MLIR style.
```

src/gt4py/next/iterator/pretty_printer.py
```diff
@@ -63,15 +65,82 @@
 }
 
 
+#: Surface spelling of the scalar types, LLVM/MLIR style.
```
havogt, line 68: I asked for this because of conciseness, but no strong opinion. Happy to switch back to the name of the enums.

src/gt4py/next/iterator/transforms/fuse_as_fieldop.py
```diff
@@ -261,8 +261,8 @@ class FuseAsFieldOp(
     ...     im.ref("inp3", field_type),
     ... )
     >>> print(nested_as_fieldop)
-    as_fieldop(λ(__arg0, __arg1) → ·__arg0 + ·__arg1, c⟨ IDimₕ: [0, 1[ ⟩)(
-      as_fieldop(λ(__arg0, __arg1) → ·__arg0 × ·__arg1, c⟨ IDimₕ: [0, 1[ ⟩)(inp1, inp2), inp3
+    as_fieldop(λ(__arg0, __arg1) → ·__arg0 + ·__arg1, c⟨ IDimₕ: [0:i32, 1:i32[ ⟩)(
```
havogt: Should we drop printing the type for domains? Because it's the index type which I believe we fix to 32 bit.

tehrengruber (commented):
Before I read the code here: I would rather skip the type string in case it is not needed (following the same logic as type_translation.from_value). Booleans are certainly nicer as True and False not 1:f1 for our purposes.

havogt:
~For bool I had already done it and I was considering it also for domain, but for plain literals I am not so sure.~ It's done as discussed: only use types when they disagree with the defaults.

src/gt4py/next/iterator/pretty_printer.py
```diff
@@ -63,12 +65,74 @@
...
+SCALAR_TYPE_NAMES: Final[Mapping[ts.ScalarKind, str]] = _types.MappingProxyType(
+    {
+        ts.ScalarKind.BOOL: "i1",
```
havogt, line 71: fyi, bool is still used for the dtype of temporaries. I was wondering if we should use `bool` here instead of `i1`, because it's the least intuitive, but then I thought it doesn't harm to learn this convention instead of mixing styles.

src/gt4py/next/iterator/pretty_parser.py
```diff
@@ -17,7 +17,7 @@
 )
 
 from gt4py.next.iterator import ir
-from gt4py.next.iterator.ir_utils import ir_makers as im
+from gt4py.next.iterator.pretty_printer import SCALAR_TYPE_KINDS, implied_literal_type
```
egparedes: Direct symbol imports are forbidden in our coding conventions (Google Python Conventions)

src/gt4py/next/iterator/pretty_parser.py
```diff
@@ -107,26 +113,48 @@
...
     def FLOAT_LITERAL(self, value: lark_lexer.Token) -> ir.Literal:
-        return im.literal(value.value, "float64")
+        return _bare_literal(value.value)
+
+    def TYPE_LITERAL(self, value: lark_lexer.Token) -> ts.ScalarType:
+        if (kind := SCALAR_TYPE_KINDS.get(value.value)) is not None:
+            return ts.ScalarType(kind=kind)
+        raise ValueError(
+            f"Invalid type '{value}'; expected one of {', '.join(sorted(SCALAR_TYPE_KINDS))}."
+        )
```
egparedes: Minor coding style nitpick:
```suggestion
        if (kind := SCALAR_TYPE_KINDS.get(value.value)) is None:
            raise ValueError(
                f"Invalid type '{value}'; expected one of {', '.join(sorted(SCALAR_TYPE_KINDS))}."
            )
        return ts.ScalarType(kind=kind)
        
```

src/gt4py/next/iterator/pretty_printer.py
egparedes: Another alternative could be to use Numpy dtype abbreviations, where boolean would be `b1` and the rest very similar (although the number would indicate number of bytes, not bits) : https://readmedium.com/numpy-typecodes-cheatsheet-1c4cd8fd2318

src/gt4py/next/iterator/pretty_printer.py
```diff
@@ -131,7 +195,9 @@ def visit_Sym(self, node: ir.Sym, *, prec: int) -> list[str]:
         return [node.id]
 
     def visit_Literal(self, node: ir.Literal, *, prec: int) -> list[str]:
-        return [str(node.value)]
+        if implied_literal_type(node.value) == node.type:
+            return [str(node.value)]
+        return [f"{node.value}:{format_type(node.type)}"]
```
egparedes: Optional style suggestion:
```suggestion
        return [
            f"{node.value}
            if implied_literal_type(node.value) == node.type
            else f"{node.value}:{format_type(node.type)}"
        ]
```

src/gt4py/next/iterator/pretty_printer.py
```diff
@@ -63,12 +65,74 @@
...
+    if value in ("True", "False"):
+        py_value = value == "True"
+    else:
+        try:
+            py_value = int(value)
+        except ValueError:
+            try:
+                py_value = float(value)
+            except ValueError:
+                return None
```
egparedes: Question: why is this case allowed? When is it normal behavior to return None?

egparedes (commented):
Only minor code style nitpicks and questions.

src/gt4py/next/iterator/pretty_printer.py
havogt: Addressed in https://github.com/GridTools/gt4py/pull/2822/commits/f3ca41762f637e9c1405b2cb28e676beb8277d07

src/gt4py/next/iterator/pretty_printer.py
havogt: You mean the array interface one without the `<`?
I think I prefer the llvm style, but of course any python style would make more sense. If you don't have a strong opinion let's keep it for now.

egparedes (approved):
LGTM

PR #2824, opened 2026-08-25
refactor[next-dace]: refactor visitor dispatch in SDFG lowering

edopao:
## Description

Refactoring of the DaCe GTIR→SDFG lowering. No functional change intended.

**Module renames** — clarify the two lowering levels:
- `lowering/gtir_dataflow.py` → `lowering/gtir_to_sdfg_lambda.py` — lambda/dataflow-level lowering inside field operators (`LambdaToDataflow`, `translate_lambda_to_dataflow`, `MemletExpr`/`ValueExpr` and the dataflow edge classes)
- `lowering/gtir_to_sdfg.py` keeps its name — program/fieldview-level driver (`GTIRToSDFG`, `SDFGBuilder`, `SubgraphContext`), so the class name matches its module again
- public entry point `build_sdfg_from_gtir()` → `lower_program_to_sdfg()`

**Visitor dispatch refactoring** — replace the long `cpm.is_call_to(...)` if/elif chains in both `visit_FunCall` implementations with a two-step dispatch:
1. Structural patterns first (let-lambda, `as_fieldop`, applied `map_`/`reduce`/`shift`)
2. Name-based dispatch through lookup tables for the remaining builtins:
   - new `BUILTIN_TRANSLATORS: dict[str, PrimitiveTranslator]` in `gtir_to_sdfg_primitives.py`
   - in `LambdaToDataflow`, two typed tables built in `__post_init__`: `_builtin_dispatch: dict[str, Callable[[gtir.FunCall], VisitResult]]` for name-matched builtins and `_pattern_dispatch: list[tuple[matcher, handler]]`, which makes the precedence between the structural patterns explicit and keeps the dispatch mypy-checked (no `getattr`)

The let-lambda handling is extracted into a dedicated `_visit_let()` method with a docstring in both visitors.

**Incidental cleanups:**
- Narrow `FunCall` arguments through the `cpm.is_applied_as_fieldop` type guard in `translate_as_fieldop`/`translate_scan`, and raise `ValueError` for a missing fieldop domain.
- Remove dead `origin is None` handling in `FieldopData.get_local_view()` — `origin` is a mandatory field.
- Replace a `type: ignore` with an `assert` on `offset_type` in `_create_field_operator_impl()`.
- Rename `translate_lambda_to_dataflow`'s parameter `sdfg_builder` → `subgraph_builder` for consistency with the `LambdaToDataflow` attribute.
- Docstring fixes: stale `origin=None` claim in `SDFGBuilder.make_field`, "type ineference" typo.


src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
```diff
@@ -1151,60 +1150,69 @@ def _visit_target(
...
-                    else self.visit(arg, ctx=ctx)
-                )
-                for param, arg in zip(node.fun.params, node.args, strict=True)
-            }
-            return self.visit(node.fun, ctx=ctx, args=args)
-        elif isinstance(node.type, ts.ScalarType):
+        # Name-matched builtins
+        if isinstance(node.fun, gtir.SymRef):
+            name = str(node.fun.id)
+            if translator := gtir_to_sdfg_primitives.BUILTIN_TRANSLATORS.get(name):
```
iomaganaris: Just a suggestion. It would be a bit more compact when looking at the code if this was defined in the same file since it's only used here but not a strong opinion on that

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_lambda.py
```diff
@@ -190,6 +191,10 @@ def get_memlet_subset(self, sdfg: dace.SDFG) -> dace_subsets.Range:
         )
 
 
+VisitResult: TypeAlias = MaybeNestedInTuple[IteratorExpr | DataExpr]
+"""Result of visiting an expression in a lambda body (i.e. in the iterator view of GTIR)."""
```
iomaganaris, line 195: Just of curiosity, is this a defined code style to add the docstring after the definition of a type/variable/etc?

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_types.py
```diff
@@ -79,14 +79,11 @@ def get_local_view(
...
-                    dim: gtir_dataflow.SymbolExpr(index, INDEX_DTYPE)
+                    dim: gtir_to_sdfg_lambda.SymbolExpr(index, INDEX_DTYPE)
                     for dim, index in zip(domain_dims, domain_indices)
                 }
-            field_origin = [
-                (dim, dace.symbolic.SymExpr(0) if self.origin is None else self.origin[i])
-                for i, dim in enumerate(self.gt_type.dims)
-            ]
-            return gtir_dataflow.IteratorExpr(
+            field_origin = [(dim, self.origin[i]) for i, dim in enumerate(self.gt_type.dims)]
```
iomaganaris, line 85: In the documentation of the functions it mentions:
```
Pass an empty tuple for `ScalarType` data or zero-dimensional fields.
```
so it seems to me (without having the larger picture) that if `self.origin` is `None` it actually has a meaning

iomaganaris (commented):
Overall LGTM. I only have some questions to verify that some changes are valid

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
edopao: It is true that `BUILTIN_TRANSLATORS` is only used here, so I could define it here. However, in order to define it, I would have to import multiple modules (`gtir_to_sdfg_primitives, translate_concat_where, translate_scan`) here, since the implementation of some primitives is located in separate files. These module are already imported in `gtir_to_sdfg_primitives` for type checking.

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_lambda.py
edopao: I think so. According to our coding conventions (https://github.com/GridTools/gt4py/blob/main/CODING_GUIDELINES.md#docstrings) docstrings should use Google Style Python docstrings (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google).

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg_types.py
edopao: By _empty tuple_ it means `origin=()`.

iomaganaris (approved):
LGTM

src/gt4py/next/program_processors/runners/dace/lowering/gtir_to_sdfg.py
edopao: On a second thought, I have applied this review comment. I agree that the suggested change is good.

PR #2828, opened 2026-08-26
fix[next]: lower `as_offset` applied to a staggered access

tehrengruber:
Applying `as_offset` to a staggered (half-integer) access — e.g. `theta_v_ic(E2C[0])(KDim - 0.5)(as_offset(Koff, ikoffset))`, reduced from icon4py's `compute_hydrostatic_correction_term` — hits two independent defects. The composition order is forced, since `as_offset` requires the offset field to carry the dimension being shifted, so the half-integer shift has to come first.

**`InlineDynamicShifts` did not fuse to a fixpoint.** The pass fused a single producer per `as_fieldop` node and returned. When an `as_fieldop` is nested more than one level deep behind a dynamic shift, the remaining producer is still hidden behind that shift after one fusion step, so `infer_domain` marks it `UNKNOWN` and emits it without a domain. This is not dace-specific — it affects every pipeline ending at `infer_domain` (`apply_fieldview_transforms`), and `roundtrip.gtir` fails the new test on `main` too; `apply_common_transforms` masked it via its later `fuse_as_fieldop` loop, which is why gtfn was unaffected. Now fuses until no dynamically shifted argument is left. This terminates because `fuse_as_fieldop` replaces every fused argument by the arguments of the producer it inlines, i.e. by strict subterms of that argument.

**The dace lowering leaked a `SymbolExpr` into tasklet code.** `_make_cartesian_shift` interpolated the whole dataclass instead of its `value` when a static cartesian shift landed on an already dynamically shifted iterator, emitting `shifted_index = index + SymbolExpr(value=0, dc_dtype=int)`. dace then picks up `int` as a free symbol and `sdfg.arglist()` fails with `KeyError: 'int'`. The branch is reachable without staggering, but a half-integer shift always lowers to a static cartesian shift, so staggering turns a corner case into the normal path.

Tests: a backend-free unit test for the fusion fixpoint, plus cartesian and unstructured `as_offset`-on-staggered-access tests, value-verified across the backend matrix. Each fails on `main`. GPU is unverified — no working GPU in my environment — but the changed code paths are not device-specific.

---

Extracted from https://github.com/havogt/gt4py/pull/77, which additionally turns the domain `assert` in `translate_as_fieldop` into an explicit error; that part is left out here.

### DIFF
--- a/src/gt4py/next/iterator/transforms/inline_dynamic_shifts.py
+++ b/src/gt4py/next/iterator/transforms/inline_dynamic_shifts.py
@@ -64,19 +64,26 @@ class InlineDynamicShifts(eve.NodeTranslator, eve.VisitorWithSymbolTableTrait):
                     node, eligible_params=list(inline_let_params.values())
                 )
 
-        if dynamic_shift_args := _dynamic_shift_args(node):
-            assert len(node.fun.args) in [1, 2]  # type: ignore[attr-defined]  # ensured by is_applied_as_fieldop in _dynamic_shift_args
+        # Fusing one producer can expose another one behind it (e.g. a chain of shifts split
+        # across multiple `as_fieldop`s), so repeat until no dynamically shifted argument is left.
+        # This terminates: `fuse_as_fieldop` replaces every fused argument by the arguments of the
+        # producer it inlines, i.e. by strict subterms of that argument, so the combined size of
+        # the arguments strictly decreases in each iteration.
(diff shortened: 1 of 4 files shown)


tehrengruber:
Superseded by https://github.com/GridTools/gt4py/pull/2829 and https://github.com/GridTools/gt4py/pull/2830, which split this into a stack.

PR #2829, opened 2026-08-26
fix[next]: fuse dynamic shifts to a fixpoint in `InlineDynamicShifts`

tehrengruber:
`InlineDynamicShifts` fused a single producer per `as_fieldop` node and returned immediately. When an `as_fieldop` is nested more than one level deep behind a dynamic shift, the remaining producer is still hidden behind that shift after one fusion step, so `infer_domain` marks it `UNKNOWN` and emits it without a domain.

This is not backend-specific — it affects every pipeline ending at `infer_domain` (`apply_fieldview_transforms`), and shows up on `roundtrip.gtir` and dace alike. `apply_common_transforms` masked it via its later `fuse_as_fieldop` loop, which is why gtfn was unaffected.

Now fuses until no dynamically shifted argument that is not a `SymRef` is left. This terminates because `fuse_as_fieldop` replaces every fused argument by the arguments of the producer it inlines, i.e. by strict subterms of that argument. The `SymRef` exclusion means a let-bound producer shared between two dynamically shifted consumers is still not inlined; that case is pre-existing (it fails identically on `main`) and tracked in #2839.

Tests: a backend-free unit test for the fusion fixpoint, plus a minimal cartesian integration test — a dynamic offset applied on top of a chain of two operations, `(a + 1) * 2`. No staggering and no unstructured shift are needed to trigger this. Both fail on `main`.

**AI Disclaimer**: This PR was written with the help of AI tools. I reviewed the code (tests and implementation) in detail.

---

Found while extracting https://github.com/havogt/gt4py/pull/77.

### DIFF
--- a/src/gt4py/next/iterator/transforms/inline_dynamic_shifts.py
+++ b/src/gt4py/next/iterator/transforms/inline_dynamic_shifts.py
@@ -64,19 +64,26 @@ class InlineDynamicShifts(eve.NodeTranslator, eve.VisitorWithSymbolTableTrait):
                     node, eligible_params=list(inline_let_params.values())
                 )
 
-        if dynamic_shift_args := _dynamic_shift_args(node):
-            assert len(node.fun.args) in [1, 2]  # type: ignore[attr-defined]  # ensured by is_applied_as_fieldop in _dynamic_shift_args
+        # Fusing one producer can expose another one behind it (e.g. a chain of shifts split
+        # across multiple `as_fieldop`s), so repeat until no dynamically shifted argument is left.
+        # This terminates: `fuse_as_fieldop` replaces every fused argument by the arguments of the
+        # producer it inlines, i.e. by strict subterms of that argument, so the combined size of
+        # the arguments strictly decreases in each iteration.
+        expr: itir.Expr = node
+        while dynamic_shift_args := _dynamic_shift_args(expr):
+            assert isinstance(expr, itir.FunCall) and len(expr.fun.args) in [1, 2]  # type: ignore[attr-defined]  # ensured by is_applied_as_fieldop in _dynamic_shift_args
             fuse_args = [
                 not isinstance(inp, itir.SymRef) and dynamic_shift_arg
-                for inp, dynamic_shift_arg in zip(node.args, dynamic_shift_args, strict=True)
+                for inp, dynamic_shift_arg in zip(expr.args, dynamic_shift_args, strict=True)
             ]
-            if any(fuse_args):
-                return fuse_as_fieldop.fuse_as_fieldop(
-                    node,
-                    fuse_args,
-                    uids=self.uids,
-                    offset_provider_type=self.offset_provider_type,
-                    enable_cse=True,
-                )
+            if not any(fuse_args):
+                break
+            expr = fuse_as_fieldop.fuse_as_fieldop(
+                expr,
+                fuse_args,
+                uids=self.uids,
+                offset_provider_type=self.offset_provider_type,
+                enable_cse=True,
+            )
 
-        return node
+        return expr
--- a/src/gt4py/next/iterator/transforms/inline_dynamic_shifts.py
+++ b/src/gt4py/next/iterator/transforms/inline_dynamic_shifts.py
@@ -65,10 +65,11 @@ class InlineDynamicShifts(eve.NodeTranslator, eve.VisitorWithSymbolTableTrait):
                 )
 
         # Fusing one producer can expose another one behind it (e.g. a chain of shifts split
-        # across multiple `as_fieldop`s), so repeat until no dynamically shifted argument is left.
-        # This terminates: `fuse_as_fieldop` replaces every fused argument by the arguments of the
-        # producer it inlines, i.e. by strict subterms of that argument, so the combined size of
-        # the arguments strictly decreases in each iteration.
+        # across multiple `as_fieldop`s), so repeat until no dynamically shifted argument that is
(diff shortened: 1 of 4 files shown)


tests/next_tests/integration_tests/feature_tests/ffront_tests/test_cartesian_shifts.py
```diff
@@ -88,3 +88,30 @@ def testee(a: cases.IKField, offset_field: cases.IKField) -> gtx.Field[[IDim, KD
         ref=ref,
         comparison=lambda out, ref: np.all(out == ref),
     )
+
+
+@pytest.mark.uses_dynamic_offsets
+def test_offset_field_of_chained_ops(cartesian_case):
+    """A dynamic offset on top of a chain of operations must be fused past all of them."""
+    Koff = gtx.FieldOffset("Koff", source=KDim, target=(KDim,))
```
tehrengruber: @tehrengruber-ai Don't redefine it here, this should already exist in some test module and can be imported.

tehrengruber-ai: There was no importable `Koff` yet — the only definitions were local ones in this file. Added `Ioff`/`Koff` to `cases_utils.py` next to `EdgeOffset` and imported them here; `test_offset_field` now uses them too.

havogt:
> Now fuses until no dynamically shifted argument is left.

The stopping rule is really *no dynamically shifted argument that is not a `SymRef`* — `fuse_args` starts with `not isinstance(inp, itir.SymRef)`. A let-bound producer is a `SymRef`, so a shifted producer that is **shared between two consumers** is skipped:

```python
Koff = gtx.FieldOffset("Koff", source=KDim, target=(KDim,))

@gtx.field_operator
def testee(a: IKField, b: IKField, off: IKFieldInt32) -> IKField:
    x = b * 3.0
    p = (a + x) * 2.0
    q = (b + x) * 4.0
    return p(as_offset(Koff, off)) + q(as_offset(Koff, off))
```

This fails with `TypeError: as_fieldop() missing 1 required positional argument: 'domain'` on `roundtrip.gtir`, and on `dace_cpu` with `assert len(fun_node.args) == 2` at `gtir_to_sdfg_primitives.py:239` — the same undomained-`as_fieldop` symptom the stack fixes for the staggered case. Under `PYTHONOPTIMIZE`, which ICON's generated `setting` script exports, that assert is stripped and it surfaces as `ValueError: not enough values to unpack (expected 2, got 1)`.

To be explicit: **this fails identically on `main`** — a pre-existing limitation, not a regression from this PR. But this PR is what establishes the fixpoint, so it seems like the right place to either widen it or scope the claim.

The IR reaching `InlineDynamicShifts` is

```
(λ(xᐞ0) → as_fieldop(+)( <shift p>, <shift q> ))( as_fieldop(×)(b, 3.0) )
```

The let-inlining pre-pass at the top of `visit_FunCall` doesn't catch it: it only fires when the let's *body* is itself the dynamically shifted `as_fieldop`, and here the body is the `+` node one level up, so `_dynamic_shift_args(let_body)` is all-`False`. Bottom-up, each shift node then fuses its `×` and `+` producers and stops at `if not any(fuse_args): break` with args `a`, `xᐞ0`, `off` — all `SymRef`s. `infer_domain` marks `xᐞ0` `UNKNOWN` and the producer is emitted without a domain.

Generalising the pre-pass to inline let params feeding a dynamic shift at any depth would cover it. This is also the shape most likely to show up in icon4py, where reused intermediates under an offset are common.

tehrengruber-ai:
Thanks, you are right — filed as #2839.

Confirmed pre-existing: your reproducer fails the same way on `main`, so this PR narrows the gap (`roundtrip.default` goes from failing to passing) without closing it. I did implement the widening — searching the whole let body via `node.fun.expr.pre_walk_values()` and revisiting after inlining — and it makes the reproducer pass with the existing suites green, but I backed it out again on cost grounds: `visit_FunCall` runs at every let node and `_dynamic_shift_args` calls `trace_shifts.trace_stencil` at every `FunCall` it walks past, so each let would re-traverse its entire body with a stencil trace per node. On the deep let chains CSE produces that is too expensive to take on speculatively. The issue records that, and sketches the direction a viable fix would have to take instead (collect once which symbols are consumed in a dynamically shifted argument position, then consult that when the let node is reached), plus the observation that a graph representation — where a reference is an edge to its defining node rather than a name — makes the question the pass is trying to answer trivial and removes the need for the `SymRef` guard altogether.

I will scope the sentence in the description to *no dynamically shifted argument that is not a `SymRef`* and link #2839 from it.

src/gt4py/next/iterator/transforms/inline_dynamic_shifts.py
```diff
@@ -64,19 +64,26 @@ def visit_FunCall(self, node: itir.FunCall, **kwargs):
                     node, eligible_params=list(inline_let_params.values())
                 )
 
-        if dynamic_shift_args := _dynamic_shift_args(node):
-            assert len(node.fun.args) in [1, 2]  # type: ignore[attr-defined]  # ensured by is_applied_as_fieldop in _dynamic_shift_args
+        # Fusing one producer can expose another one behind it (e.g. a chain of shifts split
+        # across multiple `as_fieldop`s), so repeat until no dynamically shifted argument is left.
```
havogt: maybe we should link to https://github.com/GridTools/gt4py/issues/2839 and correct the limitation of the statement "repeat until no dynamically shifted argument is left."

havogt (approved):
lgtm. See the minor request for comment change.

PR #2830, opened 2026-08-26
fix[next-dace]: lower `as_offset` applied to a staggered access

tehrengruber:
The dace lowering of a dynamic offset — a shift whose distance is only known at runtime, i.e. `as_offset(Koff, offset_field)` in the frontend — interpolated the whole `SymbolExpr` dataclass into the tasklet code instead of its `value`, emitting `shifted_index = index + SymbolExpr(value=0, dc_dtype=int)`. dace then picks up `int` as a free symbol and `sdfg.arglist()` fails with `KeyError: 'int'`.

The affected branch of `_make_cartesian_shift` is the one where the offset is a compile-time constant but the index it is added to is not, i.e. a static shift on top of an already dynamically shifted iterator. Nothing produced that combination before: the only integration test using `as_offset`, `test_offset_field`, shifts dynamically twice but in two different dimensions, and since the shift is lowered per dimension each of them still sees a symbolic index and takes the branch where only the offset arrives through a connector; a static shift alone is constant-folded and never reaches a tasklet. Staggering makes the combination the normal case, because a half-integer shift always lowers to a static cartesian shift — so `field(KDim - 0.5)(as_offset(Koff, offset_field))` hits it every time.

Tests: three `as_offset`-on-staggered-access tests, value-verified across the backend matrix. `test_cartesian_half_shift_as_offset` isolates the defect fixed here. The other two put a further operation between the staggered shift and the dynamic offset, which additionally runs into the unrelated fixpoint defect of the PR below, so those need the whole stack to go green.

**AI disclaimer**: This PR was written with the help of AI tools. Code was reviewed in detail, PR description briefly.

---

Stacked on-top of: #2829

Extracted from https://github.com/havogt/gt4py/pull/77, which additionally turns the domain `assert` in `translate_as_fieldop` into an explicit error; that part is left out here.

### DIFF
--- a/tests/next_tests/integration_tests/feature_tests/ffront_tests/test_staggered.py
+++ b/tests/next_tests/integration_tests/feature_tests/ffront_tests/test_staggered.py
@@ -157,17 +157,12 @@ def test_cartesian_half_shift_as_offset(cartesian_case):
     ) -> cases.IKField:
         return a(KDim - 0.5)(as_offset(Koff, offset_field))
 
-    isize = cartesian_case.default_sizes[IDim]
     ksize = cartesian_case.default_sizes[KDim]
-    a = cases.allocate(cartesian_case, testee, "a", sizes={IDim: isize, KHalfDim: ksize + 1})()
+    a = cases.allocate(cartesian_case, testee, "a", sizes={KHalfDim: ksize + 1})()
     offset_field = cases.allocate(
-        cartesian_case,
-        testee,
-        "offset_field",
-        sizes={IDim: isize, KDim: ksize},
-        strategy=cases.ConstInitializer(1),
+        cartesian_case, testee, "offset_field", strategy=cases.ConstInitializer(1)
     )()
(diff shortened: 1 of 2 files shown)


tests/next_tests/integration_tests/feature_tests/ffront_tests/test_staggered.py
```diff
@@ -140,3 +146,88 @@ def testee(a: cases.IJField) -> gtx.Field[[IHalfDim, JDim], np.int32]:
...
+        offset_field,
+        out=out,
+        ref=a.asnumpy()[:, 1:] + 1,
+        offset_provider={},
+    )
+
+
+@pytest.mark.uses_unstructured_shift
+@pytest.mark.uses_dynamic_offsets
+def test_unstructured_shift_half_shift_as_offset(unstructured_case_3d):
```
tehrengruber, line 204: This test is just to increase coverage as there are no unstructured & staggering tests. The tests above are sufficient to trigger the bug.

havogt (approved):

havogt:
cscs-ci run

havogt:
cscs-ci run default

PR #2832, opened 2026-08-27
perf[next-dace]: Added automatic setting of chiplet number using amdsmi

iomaganaris:
- Set `compiler.cuda.chiplet_number` automatically using the `amdsmi` python package on AMD GPUs
- Give priority to `DACE_compiler_cuda_chiplet_number` environment variable if it set
- If `amdsmi` python package doesn't exist or the env variable is not set generate a warning
- Needs https://github.com/spcl/dace/pull/2527


iomaganaris:
The logic above was added to DaCe so this PR can be closed

